PR #2173, opened 2022-12-20
fix: Presenter dependencies

thure:
A change in the cjs/esm build system might have broken these links. This PR fixes them.

### DIFF
--- a/tools/presenter/vite.config.ts
+++ b/tools/presenter/vite.config.ts
@@ -13,7 +13,7 @@ import rehypeHighlight from 'rehype-highlight';
 import { defineConfig } from 'vite';
 import { VitePluginFonts } from 'vite-plugin-fonts';
 
-import { ConfigPlugin } from '@dxos/config/vite-plugin';
+import { ConfigPlugin } from '@dxos/config/dist/plugin/node/vite-plugin.cjs';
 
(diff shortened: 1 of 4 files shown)


tools/presenter/vite.config.ts
```diff
@@ -13,7 +13,7 @@ import rehypeHighlight from 'rehype-highlight';
 import { defineConfig } from 'vite';
 import { VitePluginFonts } from 'vite-plugin-fonts';
 
-import { ConfigPlugin } from '@dxos/config/vite-plugin';
+import { ConfigPlugin } from '@dxos/config/dist/plugin/node/vite-plugin.cjs';
```
wittjosiah, line 16: Is this because of this being a module?

thure: Seems like it just isn’t exported anywhere, and AFAIK there isn’t an alias that would resolve to this, though maybe you know of one?

wittjosiah: Well, there is something there

https://github.com/dxos/dxos/blob/main/packages/sdk/config/vite-plugin.js
https://github.com/dxos/dxos/blob/main/packages/sdk/config/package.json#L21

it's just not a module

PR #2183, opened 2022-12-22
test: re-enable playwright testing

wittjosiah:


wittjosiah:
* main
  * **PR #2183** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2183" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2183?utm_source=stack-comment).

packages/sdk/react-client/playwright/tests/smoke.test.ts
```diff
@@ -6,14 +6,13 @@ import { test, expect } from '@playwright/test';
...
-    await page.goto('#/__story/stories-client-stories-tsx/Primary');
+  // NOTE: This test depends on connecting to the default production deployed HALO vault.
+  test('Renders remote client info', async ({ page }) => {
+    await page.goto('http://localhost:9009/iframe.html?id=react-client-clientcontext--primary&viewMode=story');
 
     await waitForExpect(async () => {
       const isVisible = await page.isVisible(':has-text("initialized")');
       expect(isVisible).toBeTruthy();
-    });
+    }, 15_000); // It takes the storybook a minute to boot up.
```
thure, line 16: Maybe it's too early for this optimization, but if this is about Storybook’s boot time should this instead become a `before` hook or similar?

Personally I'm accustomed to a setup where Storybook booting etc is in a `beforeAll` (https://github.com/OfficeDev/fluent-blocks/blob/next/packages/react/src/setup.test.ts#L24-L33), and each test battery as a `before` that awaits the particular test stories (https://github.com/OfficeDev/fluent-blocks/blob/next/packages/react/src/blocks/Card/Card.test.tsx#L8-L10). Would something like that work here?

thure (approved):

packages/sdk/react-client/playwright/tests/smoke.test.ts
wittjosiah: Yes! I included the comment specifically because I wasn't happy about it haha. I was used to playwright handling this for us but it seems that the storybook server responds to the healthcheck from playwright before the stories are ready to serve, then the tests start and time out. You're probably already aware of this because your solution looks like it would help. I like the idea of using mocha for everything, but we're currently still relying on playwright's ability to start the dev server alongside the test, I was already considering adding a similar feature to our test executor, though not sure if it's worth doing now or not.

thure (approved):
Neat! lgtm :)

wittjosiah:
⏳ This pull request is set to merge as part of a Graphite merge job
Stack job ID: UWDuAwWbBMCT2ACPK8ZX.
See details on [graphite.dev](https://app.graphite.dev/github/pr/dxos/dxos/2183)

PR #2187, opened 2022-12-23
feat: HOC to automatically track subscribed objects

dmaretskyi:


packages/experimental/kai/src/hooks/selection.ts
```diff
@@ -65,3 +65,26 @@ export const useSelection = (selection: Selection): SelectionHandle => {
   handle.update(selection);
   return handle;
 };
+
+export const makeReactive = <P>(comp: React.FC<P>): React.FC<P> => props => {
+  const db = useDatabase();
```
richburdon, line 70: Can't assume there is a context for this hook.
I think passing in db is a bit cumbersome and not generic (e.g., edge case of supporting multiple dbs, or selecting the space/db inside the component).

PR #2191, opened 2022-12-26
feat: Factor out ECHO type generator.

richburdon:
- [x] New `echo-typegen` package for codegen with tests.
- [x] Rename `echo-db2` to `echo-schema`. 
- [x] Implement `Object.keys()`
- [x] Implement `type.json()` 
- [x] Remove user `id` property.


packages/core/echo/echo-db2/src/object.ts
```diff
@@ -76,24 +80,34 @@ export class EchoObjectBase {
...
+    return this[unproxy]?._schemaType?.name ?? 'EchoObject';
+  }
 
-      set: (target, property, value, receiver) => {
-        if (!isValidKey(property)) {
-          return Reflect.set(target, property, value, receiver);
+  /**
+   * Convert to JSON object.
+   */
+  json() {
```
dmaretskyi, line 94: rename to `toJSON` to follow the convention: https://javascript.info/json

richburdon: Done and added tests.

PR #2193, opened 2022-12-28
fix(devtools-extension): Make Devtools Extension work

mykola-vrmchk:
https://user-images.githubusercontent.com/36420699/210776461-6e55b24b-4c40-4217-a4c7-b79054fe1ecd.mov

### DIFF
--- a/packages/devtools/devtools-extension/src/devtools.ts
+++ b/packages/devtools/devtools-extension/src/devtools.ts
@@ -5,8 +5,6 @@
 import debug from 'debug';
 import browser from 'webextension-polyfill';
 
-import { waitForDXOS } from './utils';
-
 const log = debug('dxos:extension:devtools');
 
 let panelCreated = false;
@@ -18,7 +16,6 @@ const createPanel = async () => {
   }
 
   log('Attempting to create panel...');
-  await waitForDXOS();
   await browser.devtools.panels.create('DXOS', '', 'panel.html');
   panelCreated = true;
   log('Panel created.');
--- a/packages/devtools/devtools/src/initialize.tsx
+++ b/packages/devtools/devtools/src/initialize.tsx
@@ -19,9 +19,9 @@ import { PanelsContainer } from './containers';
 import { sections } from './sections';
 import { theme } from './theme';
 
-type ClientReadyEvent = Event<{ client: Client; services: ClientServices }>;
+export type ClientAndServices = { client: Client; services: ClientServices }
 
-const Devtools = ({ clientReady }: { clientReady: ClientReadyEvent }) => {
+const Devtools = ({ clientReady }: { clientReady: Event<ClientAndServices> }) => {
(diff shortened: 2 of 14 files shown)


mykola-vrmchk:
* main
  * **PR #2193** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2193" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2193?utm_source=stack-comment).

packages/devtools/devtools/src/initialize.tsx
```diff
@@ -18,30 +19,32 @@ import { PanelsContainer } from './containers';
...
     <ErrorBoundary>
       <ThemeProvider theme={theme}>
         <CssBaseline />
         <FullScreen>
-          <Loader loading={!client} label='Loading DXOS Client...' />
-          {client && (
-            <ClientProvider client={client}>
+          <Loader loading={!value} label='Loading DXOS Client...' />
+          {value && (
+            <ClientContext.Provider value={value}>
```
wittjosiah, line 38: Realizing that nothing inside `react-client` is actually setting the `services` value in the context...I think that either we want the `ClientProvider` to set that or we should make a separate context for it. Given the comment that `services` is specifically for devtools I think it'd be reasonable to pull that code out here.

@zhenyasav what do you think? The `useClientServices` and `useDevtools` hooks probably don't need to be a part of the sdk, do they?

packages/sdk/client/src/packlets/devtools/devtools.ts
```diff
@@ -12,7 +12,6 @@ import { Client } from '../client';
  * A hook bound to window.__DXOS__.
  */
 export interface DevtoolsHook {
-  client: Client;
```
wittjosiah, line 15: What the reason for removing this? I'll say that it's been a useful fallback for debugging occasionally, but as devtools stabilizes and expands it's functionality hopefully it should become less and less needed.

mykola-vrmchk: It is not used anywhere, so I deleted it. Do you mean you were using `window.__DXOS__.client` in the console to debug something?

packages/devtools/devtools-extension/src/devtools.ts
```diff
@@ -18,7 +16,6 @@ const createPanel = async () => {
   }
 
   log('Attempting to create panel...');
-  await waitForDXOS();
```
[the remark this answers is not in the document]
mykola-vrmchk: It was breaking the extension (something about using `eval` before we have a panel is causing the error), and we are waiting for the hook in `panel.ts`, so it is fine to delete it here.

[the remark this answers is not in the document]
wittjosiah: That's funny that that is now what's breaking it. Before this is what helped make it more stable by ensuring that it didn't try to initialize a client before it could connect.

PR #2207, opened 2023-01-05
refactor: low-level UI packaging

thure:
This PR removes `react-uikit`, moving some of its functionality and all of the package formerly knows as `react-ui` into a package now called `react-components`.

This PR also creates a new package for OS UI called `react-ui`.

The package formerly identified as `@dxos/react-components` is now `@dxos/react-components-deprecated`.

These changes should add no new functionality nor cause any visual or functional regressions.


packages/apps/composer-app/src/App.tsx
```diff
@@ -44,7 +44,7 @@ export const App = () => {
   });
 
   return (
-    <UiKitProvider
+    <UiProvider
```
wittjosiah, line 47: Will eventually come from appkit?

packages/apps/composer-app/src/components/Main.tsx
```diff
@@ -4,7 +4,7 @@
 
 import React, { ComponentProps } from 'react';
 
-import { mx } from '@dxos/react-uikit';
+import { mx } from '@dxos/react-ui';
```
wittjosiah, line 7: Will this also come from appkit? My assumption based on what we discussed yesterday was that outside of halo, the apps shouldn't need to access the OS patterns directly once a steady state is reached

packages/apps/composer-app/src/pages/SpacePage.tsx
```diff
@@ -8,7 +8,7 @@ import { useOutletContext } from 'react-router-dom';
 import type { Item, Space } from '@dxos/client';
 import { useSelection } from '@dxos/react-client';
 import { Composer, DOCUMENT_TYPE } from '@dxos/react-composer';
-import { Loading } from '@dxos/react-uikit';
+import { Loading } from '@dxos/react-ui';
```
wittjosiah, line 11: Why not `react-components`?

packages/apps/composer-app/vite.config.ts
```diff
@@ -72,7 +71,6 @@ export default defineConfig({
         resolve(__dirname, './index.html'),
         resolve(__dirname, './src/**/*.{js,ts,jsx,tsx}'),
         resolve(__dirname, './node_modules/@dxos/react-ui/dist/**/*.mjs'),
-        resolve(__dirname, './node_modules/@dxos/react-uikit/dist/**/*.mjs'),
```
wittjosiah, line 75: I think react-components will probably be needed here as well

packages/apps/patterns/react-ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@dxos/react-uikit",
```
wittjosiah, line 2:
```suggestion
  "name": "@dxos/react-ui",
```

packages/common/react-components/package.json
```diff
@@ -31,10 +31,12 @@
     "@tailwindcss/forms": "^0.5.3",
```
wittjosiah, line 31: update package name to `@dxos/react-components`

packages/apps/halo-app/src/components/AlertDialog/AlertDialog.tsx
```diff
@@ -6,9 +6,7 @@ import { Transition } from '@headlessui/react';
 import * as AlertDialogPrimitive from '@radix-ui/react-alert-dialog';
 import React, { cloneElement, ComponentProps, Fragment, ReactHTMLElement, ReactNode, useEffect, useState } from 'react';
 
-import { defaultDescription, defaultFocus } from '../../styles';
-import { mx } from '../../util';
-import { Input, InputProps } from '../Input';
+import { Input, InputProps, defaultDescription, defaultFocus, mx } from '@dxos/react-ui';
```
wittjosiah, line 9: Alert dialog can't live in `@dxos/react-components`?

packages/apps/composer-app/src/App.tsx
thure: No, `react-components` needs to provide the provider. We can rename it as needed, but it's necessary for several of the components it offers.

packages/apps/composer-app/src/pages/SpacePage.tsx
thure: I'm not done yet, it will be renamed.

packages/apps/composer-app/src/components/Main.tsx
thure: No, `mx` needs to come from `react-components` because that is where the Tailwind stack is defined.

packages/apps/composer-app/vite.config.ts
thure: `react-ui` will be replaced by `react-components`.

packages/apps/halo-app/src/components/AlertDialog/AlertDialog.tsx
thure: I have a very complicated answer for this simple question 😅 — I want to redevelop this to be less specific to HALO’s use-case, so for now I gave the component to HALO.

packages/apps/patterns/react-ui/package.json
thure: Rather, `@dxos/react-components`, I will make that change.

packages/apps/composer-app/src/components/Main.tsx
thure: Or, do you mean `Main`? I think since it isn't about something DXOS is doing, it's probably best for apps to define a component like that on their own. The only component in appkit that isn't about what DXOS is doing (at the moment) is `AppLayout`, though that may later come in handy when we have OS patterns for apps that want to coordinate layout between their content and OS panels.

packages/apps/composer-app/docs/README.md
```diff
@@ -189,13 +187,11 @@ dxos/rpc-tunnel --> dxos/rpc
...
 dxos/react-client --> dxos/react-async
-dxos/react-appkit --> dxos/react-uikit
-dxos/react-uikit --> dxos/react-client
-dxos/react-uikit --> dxos/react-ui
+dxos/react-appkit --> @dxos/react-components
 dxos/react-appkit --> dxos/telemetry
 dxos/telemetry --> dxos/sentry
 dxos/composer-app --> dxos/react-composer
-dxos/react-composer --> dxos/react-uikit
+dxos/react-composer --> @dxos/react-componentskit
```
wittjosiah, line 194: I know this is generated but where did "componentskit" come from?

packages/apps/halo-app/package.json
```diff
@@ -20,13 +20,14 @@
     "@dxos/react-appkit": "workspace:*",
     "@dxos/react-async": "workspace:*",
     "@dxos/react-client": "workspace:*",
-    "@dxos/react-ui": "workspace:*",
```
wittjosiah, line 23: If halo isn't using react-ui, what is?

packages/apps/halo-app/src/pages/JoinIdentityPage.tsx
```diff
@@ -6,14 +6,15 @@ import React, { useCallback } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import type { Invitation } from '@dxos/client';
+import { JoinPanel } from '@dxos/react-appkit';
```
wittjosiah, line 9: So the existing "os" patterns have moved into appkit? I know in general apps shouldn't be using those patterns but halo would be the exception to that because it would be providing them. Presumably the successor to these will live in `@dxos/react-ui` though, right? We wouldn't want these to live in appkit because then it would give all of our apps access to them.

packages/deprecated/react-components/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@dxos/react-components",
+  "name": "@dxos/react-components-deprecated",
```
wittjosiah, line 2: rename the parent directory to match the name for sanity

packages/apps/composer-app/docs/README.md
thure: Oh, it's not generated, I may have made a goof. Will fix.

packages/apps/halo-app/package.json
thure: Nothing is right now, since react-ui is currently empty. That won't be the case for long, however.

packages/apps/halo-app/src/pages/JoinIdentityPage.tsx
thure: Right, there's still nontrivial tech debt in `JoinPanel` and some other OS patterns that landed in `appkit`, so I've moved them there for the time being.

Apps may still want something like `JoinPanel`, even if the OS might have its own version, so this is one pattern where having the two styles available could be helpful.

PR #2209, opened 2023-01-05
feat(object-model): Add Reference

mykola-vrmchk:


packages/core/echo/object-model/src/object-model.test.ts
```diff
@@ -109,4 +110,14 @@ describe('ObjectModel', () => {
     // Peer states have converged to a single value.
     expect(peer1.model.get('title')).toEqual(peer2.model.get('title'));
   });
+
+  test.only('reference', async () => {
```
dmaretskyi, line 114: . only

packages/core/echo/object-model/src/object-model.test.ts
```diff
@@ -109,4 +110,14 @@ describe('ObjectModel', () => {
     // Peer states have converged to a single value.
     expect(peer1.model.get('title')).toEqual(peer2.model.get('title'));
   });
+
+  test('reference', async () => {
+    const rig = new TestBuilder(new ModelFactory().registerModel(ObjectModel), ObjectModel);
```
dmaretskyi, line 115: add a test with two peers

dmaretskyi (approved):

PR #2220, opened 2023-01-06
feat: Configure ts-node for mocha executor

dmaretskyi:


patches/ts-node@10.9.1.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/dist/transpilers/swc.js b/dist/transpilers/swc.js
+index eeddd4fbb2f18a7e8b20bb628d5f5c806ae20594..882d10888a2d064a183da36351a0df4fd79d24db 100644
+--- a/dist/transpilers/swc.js
++++ b/dist/transpilers/swc.js
+@@ -191,6 +191,7 @@ function createSwcOptions(compilerOptions, nodeModuleEmitKind, swcInstance, swcD
+                 keepClassNames,
+                 experimental: {
+                     keepImportAssertions: true,
++                    plugins: JSON.parse(process.env.SWC_PLUGINS || '[]')
```
wittjosiah, line 9: Suggestion: open an issue w/ ts-node to expose plugins somehow so that this patch isn't necessary.

wittjosiah (approved):

patches/ts-node@10.9.1.patch
dmaretskyi: yep

dmaretskyi: https://github.com/TypeStrong/ts-node/issues/1937

PR #2222, opened 2023-01-07
feat: simplified, cleaned up app templates 

zhenyasav:


packages/apps/samples/bare-sample/project.json
```diff
@@ -1,17 +1,17 @@
...
       "options": {
         "format": "unix",
         "lintFilePatterns": [
-          "packages/apps/samples/bare/**/*.{ts,js,tsx,jsx}"
+          "packages/apps/samples/bare-sample/**/*.{ts,js,tsx,jsx}"
         ]
       },
       "outputs": [
-        "{options.outputFile}"
+        "{options.lintFilePatterns.0}"
```
zhenyasav, line 14: @wittjosiah i verified that nx actually supports `.0.` but i am not sure if this is necessary? idk ...

wittjosiah: What's the goal here? I think this would cache all the source code which seem like a bad idea to me

PR #2230, opened 2023-01-10
feat: OS patterns

thure:


packages/apps/patterns/react-ui/src/translations/locales/en-US.ts
```diff
@@ -2,4 +2,7 @@
 // Copyright 2022 DXOS.org
 //
 
-export const os = {};
+export const os = {
```
wittjosiah, line 5: I understand why you named this `os` but I'm conflicted because it breaks the pattern of it following the package name

packages/apps/tasks-app/src/App.tsx
```diff
@@ -48,7 +49,7 @@ export const App = () => {
   return (
     <ThemeProvider
       appNs='halo'
-      resourceExtensions={[appkitTranslations, tasksTranslations]}
+      resourceExtensions={[appkitTranslations, osTranslations, tasksTranslations]}
```
wittjosiah, line 52: is order important? if there's a collision which takes precedence?

packages/common/react-components/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -13,7 +13,15 @@ export default {
 
 export const Default = {
   args: {
-    fallbackValue: '20970b563fc49b5bb194a6ffdff376031a3a11f9481360c071c3fed87874106b',
-    label: <span className='sr-only'>Hello</span>
+    fallbackValue: 'squirrel',
```
wittjosiah, line 16: if possible we've tried to avoid hardcoding foo/bar-like example values in favour of slightly more realistic values, potentially indicating a practical use case

wittjosiah (approved):

packages/apps/tasks-app/src/App.tsx
thure: Order isn't important unless the resources define translations under the same namespace, in which case i18n will overwrite any keys within that namespace that collide. We’ve been avoiding that as a practice though, so order doesn’t currently matter.

packages/apps/patterns/react-ui/src/translations/locales/en-US.ts
thure: `ui` just seemed… relatively more-incorrect to me, I suppose? I'm amenable to any better name we might prefer.

packages/common/react-components/src/components/Avatar/Avatar.stories.tsx
thure: Agreed, I think I might’ve committed this particular value by mistake.

packages/apps/patterns/react-ui/src/translations/locales/en-US.ts
thure: Or maybe if we prefix all of our namespaces with `dxos` it'd be fine, which is why the package names are the same way. Maybe we could do `dxos-ui`, `dxos-appkit`, etc.

PR #2250, opened 2023-01-11
feat: Import/export

richburdon:


packages/core/echo/echo-schema/src/serializer.ts
```diff
@@ -0,0 +1,45 @@
...
+// TODO(burdon): Sort JSON keys (npm canonical serialize util).
+export class Serializer {
+  async export(database: EchoDatabase): Promise<SerializedSpace> {
+    const result = database.query();
+    const objects = result.getObjects();
+    const data = {
+      objects: objects.map((object) => {
+        return strip({
+          '@id': object[id],
+          '@type': object[schema] ? object[schema].name : undefined,
```
dmaretskyi, line 29: How about obj[type] getter that reads the internal @type field

PR #2252, opened 2023-01-12
feat: OS invitations

thure:
This PR adds OS patterns for
- lists of members, each with status badges
- compact lists of invitations

https://user-images.githubusercontent.com/855039/212211866-c0280cda-2a4c-421e-9538-908d446cf311.mp4


richburdon (commented):
Just to comment i think this is now stale since we have the functionality in kai now.

thure:
This isn't stale since it introduces the OS pattern for a list of members with status badges and invitations that can fit in compact spaces. #2294 is based on this branch — both have lots of OS feature work in them.

Kai doesn’t have to consume these if invitations and join flows should remain in their own pages there, I can remove those uses.

thure:
I'll close this in favor of merging #2294 instead, where I'll remove integrations with kai so those can be considered for a follow-up.

PR #2256, opened 2023-01-12
fix: Fix non-deterministic test in ObjectModel

dmaretskyi:


packages/core/echo/object-model/src/object-model.test.ts
```diff
@@ -238,10 +237,11 @@ describe('ObjectModel', () => {
       testBuilder.configureReplication(true);
       await testBuilder.waitForReplication();
 
-      const expectedArray = [1, 2.2, 2.1, 3.2];
-
-      expect(peer1.model.get('tags').toArray()).toEqual(expectedArray);
-      expect(peer2.model.get('tags').toArray()).toEqual(expectedArray);
+      // The output will either be [1, 2.2, 2.1, 3.2] or [ 1, 2.1, 2.2, 3.2 ].
```
richburdon, line 240: can we predict this based on use comparing the lexical order of our feed keys (so that the test is more precise?)

PR #2257, opened 2023-01-12
feat: Integrate arrays and refs into the experimental db API

dmaretskyi:


packages/core/echo/echo-schema/src/database.test.ts
```diff
@@ -72,33 +72,6 @@ describe('EchoDatabase', () => {
     expect(task.details.deadline).toEqual('2021-01-01');
```
richburdon, line 72: prefer that we commit to using chai everywhere (since we're mostly doing this now).

packages/core/echo/echo-schema/src/database.test.ts
```diff
@@ -153,6 +126,88 @@ describe('EchoDatabase', () => {
...
+      expect(task.tags.slice()).toEqual(['red', 'yellow', 'magenta', 'blue']);
+    });
+
+    test('array of sub documents', async () => {
+      const db = await createDatabase();
+
+      const task = new Document({ title: 'Main task' });
+      await db.save(task);
+
+      task.subtasks = new EchoArray();
```
richburdon, line 163: Nice. So we can do `= []` in the proxy `set` method as syntax sugar here?

packages/core/echo/echo-schema/src/database.ts
```diff
@@ -184,8 +189,11 @@ export class EchoDatabase {
   private _createObjectInstance(item: Item<any>): EchoObject | undefined {
     if (item.model instanceof ObjectModel) {
       const type = item.model.get('@type');
-      const Proto = this._router.schema?.getPrototype(type);
+      if (!type) {
+        return new Document();
+      }
 
+      const Proto = this._router.schema?.getPrototype(type);
```
richburdon, line 196: `Type` or `TypeConstructor`?

richburdon: can you look at toJSON to see how these should be serialized

packages/core/echo/echo-schema/src/document.ts
```diff
@@ -137,7 +137,7 @@ export class DocumentBase extends EchoObject<ObjectModel> {
   }
 
   private _getModelProp(prop: string): any {
-    let type = this._item!.model.get(`${prop}$type`);
+    let type;
```
richburdon, line 140: For serialization (both toJSON and import/export) we need to be able to inspect (recursively) the object model (Document) properties.

packages/core/echo/echo-schema/src/echo-array.ts
```diff
@@ -116,7 +116,7 @@ export class OrderedSet<T extends DocumentBase> implements Array<T> {
   }
 
   slice(start?: number | undefined, end?: number | undefined): T[] {
-    throw new Error('Method not implemented.');
+    return Array.from(this.values()).slice(start, end);
```
richburdon, line 119: Is this working? It seems to just return a sliced JS object -- this doesn't affect that actual model?

packages/core/echo/echo-schema/src/index.ts
```diff
@@ -6,7 +6,7 @@ export * from './database';
 export * from './database-router';
 export * from './defs';
 export * from './object';
-export * from './ordered-set';
+export * from './echo-array';
```
richburdon, line 9: please sort import/exports

packages/core/echo/object-model/src/index.ts
```diff
@@ -10,3 +10,5 @@ export * from './object-model';
 export * from './ordered-list';
 export * from './text-index';
 export * from './util';
+export * from './ordered-array';
```
richburdon, line 13: sort

packages/experimental/kai/src/containers/ProjectList.tsx
```diff
@@ -66,9 +66,9 @@ export const ProjectItem: FC<{ project: Project }> = withReactor(({ project }) =
 
   // TODO(burdon): Pass in Task1, Task2.
   const handleDrag = (active: number, over: number) => {
-    const task1 = project.tasks[active];
-    const task2 = project.tasks[over];
-    project.tasks.splice(project.tasks.indexOf(task2), 0, task1);
+    const task = project.tasks[active];
+    project.tasks.splice(active, 1);
+    project.tasks.splice(over, 0, task);
```
richburdon, line 71: will this be atomic? should you use your new batch API?

packages/core/echo/echo-schema/src/echo-array.ts
dmaretskyi: yes, slice should return a copy https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

packages/core/echo/echo-schema/src/index.ts
dmaretskyi: isn't the linter doing it?

packages/experimental/kai/src/containers/ProjectList.tsx
dmaretskyi: won't be atomic. I think we should have some sort of deffered batching by default. Same as react batches state updates automatically.

All mutations that follow one another should go into one batch.

This, together with cross-item batches (needed to achieve strong-reference consistency), we should do this on the database level.

packages/core/echo/echo-schema/src/database.test.ts
dmaretskyi: yes

dmaretskyi: see test below

packages/core/echo/echo-schema/src/database.test.ts
dmaretskyi: sure

packages/core/echo/echo-schema/src/echo-array.ts
richburdon: Add TODOs?

packages/core/echo/echo-schema/src/index.ts
richburdon: can you try to run lint before submitting your PR?

PR #2278, opened 2023-01-14
fix: Taks keyboard access in `kai`

thure:
Note that arrows up and down shouldn't change focus because they change cursor position and tracking whether the cursor is at the beginning or end of the text is nontrivial (afaik, though we do have a precedent in the `PinInput`).

https://user-images.githubusercontent.com/855039/212799254-0552ecb1-bbee-4501-b9a0-835e56d91b7f.mp4


packages/experimental/kai/src/containers/TaskList.tsx
```diff
@@ -154,17 +162,45 @@ export const TaskItem: FC<{
...
+              (document.querySelector('input#new-task') as HTMLElement | undefined)?.focus();
+            } else {
+              (
+                document.querySelector(`input[data-orderindex="${orderIndex + 1}"]`) as HTMLElement | undefined
+              )?.focus();
+            }
+          }
+          break;
+        case 'PageDown':
+          e.preventDefault();
```
wittjosiah, line 192: this didn't seem to be working for me, it shifted the focus but if there was scroll possible on the page it would also trigger that as well

wittjosiah: also, it's not currently possible to pagedown into the new task input which felt a little awkward in practice

packages/experimental/kai/src/containers/TaskList.tsx
```diff
@@ -156,14 +170,58 @@ export const TaskItem: FC<{
...
   const { debug } = useAppState();
   useReactorContext({
     onChange: () => {
       onSave?.(task);
     }
   });
 
+  const onKeyDown = useCallback(
+    (e: KeyboardEvent<Element>) => {
+      switch (e.key) {
```
richburdon, line 185: Hey Will, thanks for fixing this.
Small note: we avoid single letter var names and try to consistently use `event` in these callback patterns, which should be consistent with our docs.

thure: Got it, will do!

Issue #2270, opened 2023-01-13
switch app create git clone to https


zhenyasav:
right now git ssh is being used unnecessarily

PR #2281, opened 2023-01-16
feat: convert templates to self-contained npm packages

zhenyasav:
This also creates packages `@dxos/create` to enable `npm init @dxos` and `@dxos/create-bare` to enable `npm init @dxos/bare` 

@wittjosiah  would be good to get a sanity check on `packages/apps/templates/hello-template/project.json` and the `project.json` in `bare-template` to make sure I set that up correctly (the build requires a two step thing: tsc, followed by basically a file copy)


packages/devtools/cli/package.json
```diff
@@ -31,16 +31,19 @@
   },
   "dependencies": {
     "@dxos/async": "workspace:*",
+    "@dxos/bare-template": "workspace:*",
```
wittjosiah, line 34: Did you consider `npm install`-ing the requested template as needed rather than including all the templates as cli deps?

The template name to cli could fit into a `@dxos/${template}-template` template string, then it could be installed globally or in a temp directory so it could be imported to execute.

This is probably simpler for now, but can't help but consider the possibility of support community templates like we've played with in previous iterations.

packages/devtools/cli/src/commands/app/create.ts
```diff
@@ -52,65 +52,44 @@ export default class Create extends BaseCommand {
...
     }
     try {
-      await promisify(exec)('which pnpm');
+      await exec('which pnpm');
     } catch {
       this.error('pnpm not found. Please run "npm i -g pnpm" first.', { exit: 1 });
     }
     try {
-      this.log('Cloning template from Github...');
-      await promisify(exec)(`
```
wittjosiah, line 80: I think removing this resolves #2270 by removing git clone

packages/devtools/cli/src/commands/app/create.ts
```diff
@@ -125,16 +104,18 @@ export default class Create extends BaseCommand {
...
+          if (verbose) {
+            console.log(saved ? 'wrote' : 'skip', file.shortDescription());
+          }
+        })
+      );
     } catch (err: any) {
-      this.log(`Unable to create: ${err.message}`);
+      this.log('Unable to create application');
       this.log(err?.stack);
       this.error(err, { exit: 1 });
```
wittjosiah, line 118:
```suggestion
      this.error(err);
      this.error(err?.stack);
      this.error('Unable to create application', { exit: 1 });
```
wondering if something like this might make more sense when reading the output?

packages/devtools/cli/src/util/exec.ts
```diff
@@ -0,0 +1,42 @@
...
+//
+
+import { spawn } from 'node:child_process';
+
+export type ExecOptions = {
+  verbose: boolean;
+  cwd: string;
+};
+
+export const exec = (command: string, options?: Partial<ExecOptions>): Promise<string> => {
```
wittjosiah, line 12: note to compare with `execTool` in the test executor

packages/devtools/initializers/create/src/main.ts
```diff
@@ -0,0 +1,20 @@
...
+    boolean: ['interactive', 'verbose', 'monorepo']
+  });
+  const { monorepo, interactive, verbose } = { interactive: false, verbose: false, monorepo: false, ...args };
+  const files = await template.execute({
+    verbose,
+    input: interactive
+      ? { monorepo }
+      : { pwa: true, dxosUi: true, tailwind: true, react: true, monorepo, storybook: false }
+  });
+  await Promise.all(files.map((file) => file.save()));
```
wittjosiah, line 19: - What directory does this output in by default?
- Is there a way to change that?
- Does `npm install` automatically get run on that directory after this executes?
- How is the vite patch being dealt with in that case?

packages/devtools/cli/package.json
zhenyasav: oh yeah totally did consider it and opted to leave that for later yep.

zhenyasav: we can even `exec(`npx init @dxos`)` and not worry about this problem

packages/devtools/initializers/create/src/main.ts
zhenyasav: `cwd` by default, no sign of any npm convention besides that, but there is argument forwarding which can be used to special case an argument or flag or something
npm install does not run after this, that's always a separate step (which I think should not be automated without some prompt or flag) and i think that's why the patch works fine

Issue #2277, opened 2023-01-14
Mobile usability issues


zhenyasav:
kai, os shell
- [x] keyboard does not come up during pin entry
- [x] cannot dismiss pin entry when joining a device

thure:
> keyboard does not come up during pin entry
As it turns out, this is unexpectedly nontrivial: https://gist.github.com/cathyxz/73739c1bdea7d7011abb236541dc9aaa

thure:
Having done some more research and tried a few things, I'm at a loss.

The best workaround for now is for the user (I know :/) to tap outside the input, then tap back in. We're encountering weirdness where iOS Safari is doing everything _but_ opening the keyboard when calling `.focus()` outside of a click handler, and annoyingly Apple sees this as by-design.

iOS _requires_ the user to tap something in order to open the keyboard. It's not something that can be faked, though [we could listen for any tapping and focus the input then](https://stackoverflow.com/a/35368807/392113).

Perhaps the best solution would be to only autofocus on non-iOS browsers, though UA sniffing is always a bit of a smell…

thure:
I think this is reasonably resolved by #2284, please reopen if there are further issues.

PR #2284, opened 2023-01-16
fix: Don’t autofocus in iOS Safari

thure:
This tries to resolve part of #2277.


packages/common/react-components/src/components/Input/BarePinInput.tsx
```diff
@@ -29,6 +29,14 @@ export const BarePinInput = forwardRef<HTMLInputElement, BarePinInputProps>(
     const inputRef = useForwardedRef(ref);
     const inputFocused = useIsFocused(inputRef);
 
+    // todo (thure): Validate this workaround for iOS: when `autoFocus` is true, iOS focuses the input but does not open the keyboard; this attempts to re-focus the input on touch events.
```
wittjosiah, line 32: unfortunately doesn't seem to work for me

PR #2288, opened 2023-01-16
feat: vault compatibility mode

wittjosiah:
Introduces compatibility mode for hosting client services. This will allow browsers without support for `SharedWorker` to still use a remote client via iframe in a diminished fashion.

The limitation placed on the client in compatibility mode is that only one tab can be active at a time. We expect this is a reasonable trade-off as the main browser without `SharedWorker` support is Android Chrome and mobile browsers only support viewing a single tab at a time anyways.

Currently this introduces compatibility mode for _all_ mobile browsers via user agent sniffing. This is because we have not found a good way yet to debug `SharedWorker`s on iOS despite Safari supporting them.


wittjosiah:
* main
  * **PR #2288** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2288" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2288?utm_source=stack-comment).

packages/sdk/client-services/src/packlets/vault/localstorage-resource-manager.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  hasAcquired() {
+    const id = this._currentId();
+    return id === this.id;
+  }
+
+  async acquire() {
+    log('acquiring...', { key: this.key, id: this.id });
+    localStorage.setItem(this.key, JSON.stringify({ id: this.id, time: Date.now() }));
```
dmaretskyi, line 43: this should check whether some other tab is holding the lock

wittjosiah: This is setup like this along the lines of what we discussed, where any tab would be able to resume usage (aka acquire access to the vault). That being said, I think you're right that it needs to be more aware of the tabs access. Maybe something like the `acquire` method doesn't resolve until an existing tab has finished releasing? So they would need to communicate via a second key likely.

packages/apps/halo-app/src/util/browser-check.ts
```diff
@@ -0,0 +1,12 @@
...
..._|a|g|p|s|t)|tp)|hu(aw|tc)|i\-(20|go|ma)|i230|iac( |\-|\/)|ibro|idea|ig01|ikom|im1k|inno|ipaq|iris|ja(t|v)a|jbro|jemu|jigs|kddi|keji|kgt( |\/)|klon|kpt |kwc\-|kyo(c|k)|le(no|xi)|lg( g|\/(k|l|u)|50|54|\-[a-w])|libw|lynx|m1\-w|m3ga|m50\/|ma(te|ui|xo)|mc(01|21|ca)|m\-cr|me(rc|ri)|mi(o8|oa|ts)|mmef|mo(01|02|bi|de|do|t(\-| |o|v)|zz)|mt(50|p1|v )|mwbp|mywa|n10[0-2]|n20[2-3]|n30(0|2)|n50(0|2|5)|n7(0(0|1)|10)|ne((c|m)\-|on|tf|wf|wg|wt)|nok(6|i)|nzph|o2im|op(ti|wv)|oran|owg1|p800|pan(a|d|t)|pdxg|pg(13|\-([1-8]|c))|phil|pire|pl(ay|uc)|pn\-2|po(ck|rt|se)|prox|psio|pt\-g|qa\-a|qc(07|12|21|32|60|\-[2-7]|i\-)|qtek|r380|r600|raks|rim9|ro(ve|zo)|s55\/|sa(ge|ma|mm|ms|ny|va)|sc(01|h\-|oo|p\-)|sdk\/|se(c(\-|0|1)|47|mc|nd|ri)|sgh\-|shar|sie(\-|m)|sk\-0|sl(45|id)|sm(al|ar|b3|it|t5)|so(ft|ny)|sp(01|h\-|v\-|v )|sy(01|mb)|t2(18|50)|t6(00|10|18)|ta(gt|lk)|tcl\-|tdg\-|tel(i|m)|tim\-|t\-mo|to(pl|sh)|ts(70|m\-|m3|m5)|tx\-9|up(\.b|g1|si)|utst|v400|v750|veri|vi(rg|te)|vk(40|5[0-3]|\-v)|vm40|voda|vulc|vx(52|53|60|61|70|80|81|83|85|98)|w3c(\-| )|webc|whit|wi(g |nc|nw)|wmlb|wonu|x700|yas\-|your|zeto|zte\-/i.test(a.substr(0,4))) { check = true; } })(navigator.userAgent || navigator.vendor || (window as any).opera);
```
dmaretskyi, line 10: 🤯

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,61 @@
...
+  private _onMessage(event: MessageEvent<any>) {
+    if (event.data.message === Message.ACQUIRING) {
+      this._releaseTrigger.wake();
+    }
+  }
+
+  private async _requestLock(steal = false) {
+    log('requesting lock...', { steal });
+    const acquired = new Trigger();
+    void navigator.locks.request(LOCK_KEY, { steal }, async () => {
```
dmaretskyi, line 49: if you get AbortError thrown by lock.request then this means that lock was stolen and you should close the client

packages/sdk/client-services/src/packlets/deprecated/system.ts
```diff
@@ -3,33 +3,47 @@
...
 
 /**
  * @deprecated
  */
 export class SystemServiceImpl implements SystemService {
-  constructor(private readonly _config: Config, private readonly _serviceContext: ServiceContext) {}
+  private readonly _resourceManager?: VaultResourceManager;
+
+  constructor(private readonly _config: Config, private readonly _serviceHost: ClientServicesHost) {
+    if (this._serviceHost.compatibilityMode) {
```
dmaretskyi, line 18: i'd change `compatibilityMode` to `exclusive` (uses locks to make sure only a single instance is running)

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -47,18 +48,27 @@ export class ClientServicesHost implements ClientServicesProvider {
...
     networkManager,
-    storage = createStorageObjects(config.get('runtime.client.storage', {})!).storage
+    storage = createStorageObjects(config.get('runtime.client.storage', {})!).storage,
+    compatibilityMode = false
   }: ClientServicesHostParams) {
     this._config = config;
     this._modelFactory = modelFactory;
     this._networkManager = networkManager;
     this._storage = storage;
+    this.compatibilityMode = compatibilityMode;
```
dmaretskyi, line 67: I'd initialize service registry with systemService here in the constructor so they are always available

then `open` and `close` control only other services

packages/sdk/client-services/src/packlets/vault/iframe-compatibility-runtime.ts
```diff
@@ -0,0 +1,93 @@
...
+  private _config!: Config;
+
+  @logInfo
+  public origin?: string;
+
+  constructor({ configProvider, appPort }: IFrameCompatibilityRuntimeParams) {
+    this._configProvider = configProvider;
+    this._appPort = appPort;
+    this._getServices = async (checkContext = true) => {
+      if (checkContext && !this._clientServices.isOpen) {
```
dmaretskyi, line 37: move this check into the ClientServiceHost:

```
clientServiceHost.getService('ServiceName') // does the check and returns the service
```

packages/sdk/client-services/src/packlets/services/service-host.ts
dmaretskyi: Service registry might need to become mutable for this

packages/sdk/client-services/src/packlets/deprecated/system.ts
dmaretskyi: or `lockKey`

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+
+enum Message {
+  ACQUIRING = 'acquiring'
+}
+
+export class VaultResourceManager {
+  private readonly _broadcastChannel = new BroadcastChannel('vault-resource-manager');
+  private _releaseTrigger = new Trigger();
+
+  constructor(private readonly _serviceHost: ClientServicesHost) {
```
dmaretskyi, line 20: pass a pair of callbacks

packages/sdk/client-services/src/packlets/deprecated/system.ts
```diff
@@ -3,33 +3,47 @@
...
+      this._resourceManager = new VaultResourceManager(this._serviceHost);
```
dmaretskyi, line 19: try moving to service-host

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { asyncTimeout, Trigger } from '@dxos/async';
+import { log } from '@dxos/log';
+
+import { ClientServicesHost } from '../services';
+
+const LOCK_KEY = 'DXOS_COMPATIBILITY_LOCK';
```
dmaretskyi, line 10:
```suggestion
const LOCK_KEY = 'DXOS_RESOURCE_LOCK';
```

packages/sdk/client-services/src/packlets/vault/iframe-runtime.ts
```diff
@@ -25,7 +25,7 @@ export class IFrameRuntime {
   private readonly _systemRpc: ProtoRpcPeer<WorkerServiceBundle>;
   private readonly _transportService = new WebRTCTransportService();
 
-  constructor({ systemPort, workerAppPort, windowAppPort }: IframeRuntimeParams) {
+  constructor({ systemPort, workerAppPort, windowAppPort }: IFrameRuntimeParams) {
```
dmaretskyi, line 28: IFrameProxyRuntime

packages/sdk/client-services/src/packlets/vault/iframe-compatibility-runtime.ts
```diff
@@ -0,0 +1,93 @@
...
+import { MaybePromise } from '@dxos/util';
+
+import { clientServiceBundle, ClientServices, ClientServicesHost } from '../services';
+
+export type IFrameCompatibilityRuntimeParams = {
+  configProvider: () => MaybePromise<Config>;
+  appPort: RpcPort;
+};
+
+export class IFrameCompatibilityRuntime {
```
dmaretskyi, line 20: IFrameHostRuntime

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+
+enum Message {
+  ACQUIRING = 'acquiring'
+}
+
+export class VaultResourceManager {
```
dmaretskyi, line 16:
```suggestion
export class VaultResourceLock {
```

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+  private readonly _broadcastChannel = new BroadcastChannel('vault-resource-manager');
```
dmaretskyi, line 17: extract to const

packages/sdk/client/src/packlets/client/client.ts
```diff
@@ -173,10 +173,14 @@ export class Client {
   /**
    * Get system status.
    */
-  async getStatus(): Promise<Status> {
+  async getStatus(): Promise<StatusResponse> {
     return this._services.services?.SystemService.getStatus();
   }
 
+  async compatibilityModeReconnect(): Promise<void> {
```
dmaretskyi, line 180: add jsdoc

packages/core/protocols/src/proto/dxos/client.proto
```diff
@@ -19,15 +19,21 @@ package dxos.client;
 // SYSTEM
 //
 
-message Status {
-  string message = 1;
+enum Status {
+  OK = 0;
```
dmaretskyi, line 23: ACTIVE, INACTIVE

packages/core/protocols/src/proto/dxos/client.proto
```diff
@@ -19,15 +19,21 @@ package dxos.client;
...
+  CLOSED = 1;
+}
+
+message StatusResponse {
+  optional string message = 1;
```
dmaretskyi, line 28: swap them

packages/apps/patterns/react-appkit/src/components/Fallback/Fallback.tsx
```diff
@@ -19,3 +21,35 @@ export const GenericFallback = () => {
   const { t } = useTranslation('appkit');
   return <Fallback message={t('generic loading label', { ns: 'appkit' })} />;
 };
+
+export const ClientFallback = ({ client, status }: Partial<ClientContextProps>) => {
```
richburdon, line 25: One line comment.
I assume we can't use the hook since this is used before the context is set-up. In which case pass in needed properties (compatReconnect) rather than the client object.

packages/core/protocols/src/proto/dxos/client.proto
```diff
@@ -19,15 +19,21 @@ package dxos.client;
...
+  optional Status status = 2;
 }
 
 service SystemService {
   /// Called at the start of the session. Service should block until it is fully initialized.
   rpc InitSession (google.protobuf.Empty) returns (google.protobuf.Empty);
   rpc GetConfig(google.protobuf.Empty) returns (dxos.config.Config);
-  rpc GetStatus (google.protobuf.Empty) returns (Status);
+  rpc GetStatus (google.protobuf.Empty) returns (StatusResponse);
```
richburdon, line 36: Stream?

packages/sdk/client-services/src/packlets/vault/iframe-compatibility-runtime.ts
richburdon: Comment please. What, how, when? Really important and complex class.

packages/sdk/client-services/src/packlets/vault/iframe-compatibility-runtime.ts
```diff
@@ -0,0 +1,93 @@
...
+            ? new WebsocketSignalManager([signalServer])
+            : new MemorySignalManager(new MemorySignalManagerContext()), // TODO(dmaretskyi): Inject this context.
+          transportFactory: this._transportFactory
+        })
+      });
+
+      this._clientRpc = createProtoRpcPeer({
+        exposed: clientServiceBundle,
+        handlers: {
+          DataService: async () => (await this._getServices()).services.DataService,
```
richburdon, line 68: Try something like this to avoid boilerplate?

```ts
const getService = async (cb) => cb(await this._getServices();

DataService: getService(services => services.DataService),
...
```

packages/sdk/client-services/src/packlets/vault/iframe-compatibility-runtime.ts
```diff
@@ -0,0 +1,93 @@
...
+      const error = await this._ready.wait();
+      if (error !== undefined) {
+        throw error;
+      }
+      return this._clientServices;
+    };
+  }
+
+  async start() {
+    try {
```
richburdon, line 50: log('starting...') then stopping below.
Ask @dmaretskyi about using context objects also?

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+        this._releaseTrigger = new Trigger();
+        await this._releaseTrigger.wait();
+        log('releasing lock...');
+        await this._serviceHost.close();
+        log('released lock');
+      })
+      .catch(async () => {
+        await this._serviceHost.close();
+      });
+    await acquired.wait();
```
richburdon, line 62: newline

packages/sdk/react-client/src/client/ClientContext.tsx
```diff
@@ -78,55 +80,81 @@ export const ClientProvider = ({
...
     throw error;
   }
 
-  useAsyncEffect(async () => {
+  useEffect(() => {
+    if (!client) {
+      return;
+    }
+
+    const interval = setInterval(async () => {
```
richburdon, line 99: generally to be safe i'd move interval outside of effect and clear it each time effect is run. realize in this case client isn't likely to change.

packages/sdk/client-services/src/packlets/vault/vault-resource-manager.ts
wittjosiah: I'm unsure if this is a good idea since the message handler for is it assigned in the constructor?

packages/sdk/react-client/src/client/ClientContext.tsx
wittjosiah: Not sure I understand, why outside the effect? Having it inside the effect with the clear on unmount ensures that the setInterval and clearInterval are paired

packages/apps/halo-app/src/util/browser-check.ts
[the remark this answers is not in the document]
wittjosiah: including tablets because they also run ios/android which is the relevant thing here as opposed to screen size

PR #2309, opened 2023-01-17
fix: cosmetic improvements in kai

zhenyasav:
- [x] sidebar border
- [x] table row height, row borders, cell padding, full width
- [x] table frame search bar bottom border
- [x] no pointer events on the fade `:after` element in cards to allow clicking through it
- [x] do not initialize another identity in demo mode if one exists

### DIFF
--- a/packages/common/react-components/src/config/tailwind.ts
+++ b/packages/common/react-components/src/config/tailwind.ts
@@ -246,5 +246,8 @@ export const tailwindConfig = ({
   },
   plugins: [tailwindcssLogical, tailwindcssForms, tailwindcssRadix()],
   ...(env === 'development' && { mode: 'jit' }),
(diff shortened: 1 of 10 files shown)


packages/common/react-components/src/config/tailwind.ts
```diff
@@ -246,5 +246,8 @@ export const tailwindConfig = ({
   },
   plugins: [tailwindcssLogical, tailwindcssForms, tailwindcssRadix()],
   ...(env === 'development' && { mode: 'jit' }),
```
richburdon, line 248: what is jit?

thure: @richburdon https://v2.tailwindcss.com/docs/just-in-time-mode

thure (approved):
lgtm!

PR #2315, opened 2023-01-19
feat(config): inject build info and environment variables

wittjosiah:


wittjosiah:
* main
  * **PR #2315** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2315" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2315?utm_source=stack-comment).

packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -83,7 +84,12 @@ message Runtime {
       optional string path = 4;                   // TODO(burdon): Rename data_path.
     }
 
-    optional string debug = 1 [(env) = "DEBUG"];  // TODO(burdon): Change to LOG.
+    message Log {
+      optional string filter = 1 [(env) = "LOG_FILTER"];
+      optional string prefix = 2 [(env) = "LOG_PREFIX"];
+    }
+
+    optional Log log = 1;
```
wittjosiah, line 92: I realize this isn't backward compatible but figured we might be able to get away with it since it's not really being used currently.

packages/sdk/client/src/packlets/client/client.ts
```diff
@@ -68,11 +68,10 @@ export class Client {
     this._echo = new EchoProxy(this._services, this._modelFactory, this._halo);
     this._mesh = new MeshProxy(this._services);
 
-    // TODO(burdon): Reconcile with Config.sanitizer.
-    if (Object.keys(this._config.values).length > 0 && this._config.values.version !== EXPECTED_CONFIG_VERSION) {
```
wittjosiah, line 72: This seems unnecessary since we require a `Config` instance to be passed in

packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -95,15 +101,24 @@ message Runtime {
...
     optional int32 invitation_expiration = 5;
 
-    // TODO(burdon): Remove?
     optional string remote_source = 6;
   }
 
   message App {
+    message BuildInfo {
+      /// RFC-3339 datetime string.
+      optional string build_time = 1;
```
richburdon, line 110: just `timestamp`
document version from package.json
move commit hash after timestamp
perhaps add tag?

packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -95,15 +101,24 @@ message Runtime {
...
+      optional string version = 2;
+      optional string commit_hash = 3;
+    }
+
```
richburdon, line 114: need `app_id`, `org_id`

theme feels like part of a sub `options` message

similarly website, public_url as part of a deployment message

packages/devtools/devtools/vite.config.ts
```diff
@@ -62,7 +49,14 @@ export default defineConfig({
     }
   },
   plugins: [
-    ConfigPlugin(),
+    ConfigPlugin({
+      env: [
+        'DX_ENVIRONMENT',
+        'IPDATA_API_KEY',
+        'SENTRY_DESTINATION',
+        'TELEMETRY_API_KEY'
```
richburdon, line 57: unless these are standard env keys then prefix with KUBE_ or DX_?

packages/core/protocols/src/proto/dxos/config.proto
wittjosiah: afaik none of those are being used currently so we could probably redo that section later

packages/devtools/devtools/vite.config.ts
wittjosiah: yeah, I think if we're going to integrate telemetry calls into client when we do that we should update the name of the variable it's looking for to be more dxos-specific :+1:

PR #2325, opened 2023-01-19
fix: enable kube cross signaling in Gravity

quaisx:
move kubes to VMs and enable cross signaling
Gravity launches agents that rely on different instances of kubes to obtain each others ICE candidates, exchange them via kubes, establish a p2p connect and perform info exchange.


egorgripasov:
closing as non relevant

PR #2330, opened 2023-01-20
feat: Restructure devtools with router.

richburdon:
Begins work on https://github.com/dxos/dxos/issues/2331

@mykola-vrmchk please note standardizing the different app bootstrap sequences: `html => main.tsx => App.tsx` and the use of hooks to handle the async bootstrap in a consistent manner.

PR #2337, opened 2023-01-23
fix(devtools-extension): Fix Devtools extension after App reafactor

mykola-vrmchk:


mykola-vrmchk:
* main
  * **PR #2330** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2330" target="_blank">Graphite</a> 
    * **PR #2337** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2337" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2337?utm_source=stack-comment).

packages/apps/patterns/react-appkit/src/telemetry/base-properties.ts
```diff
@@ -3,13 +3,22 @@
...
 import { humanize } from '@dxos/util';
 
 const IPDATA_API_KEY = process.env.IPDATA_API_KEY;
 
-export const DX_TELEMETRY = localStorage.getItem('halo-app:telemetry-disabled');
-export const DX_GROUP = localStorage.getItem('halo-app:telemetry-group');
+let DX_TELEMETRY: string | null = null;
+let DX_GROUP: string | null = null;
+try {
+  DX_TELEMETRY = localStorage.getItem('halo-app:telemetry-disabled');
```
wittjosiah, line 15: was this crashing? my impression is the localStorage.getItem returns null if the item isn't found? is this because localstorage isn't available in the extension?

packages/devtools/devtools-extension/vite.config.ts
```diff
@@ -0,0 +1,111 @@
...
+import { ThemePlugin } from '@dxos/react-components/plugin';
+import { kaiThemeExtension } from '@dxos/kai/theme-extensions';
+import { osThemeExtension } from '@dxos/react-ui/theme-extensions';
+
+import manifest from './manifest.json';
+
+const env = (value?: string) => (value ? `"${value}"` : undefined);
+
+// https://vitejs.dev/config/
+export default defineConfig({
```
wittjosiah, line 21: what's the motivation behind building devtools w/ vite?

packages/apps/patterns/react-appkit/src/telemetry/base-properties.ts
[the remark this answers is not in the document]
wittjosiah: I think that if these variables aren't available that we should just omit the telemetry hook from the extension until we have a proper place to pull them from

mykola-vrmchk: localStorage is throwing an error in sandboxed window in the extension

packages/devtools/devtools-extension/vite.config.ts
mykola-vrmchk: for consistency across repo, and it was easier to do than upgrade build.js with all required plugins

wittjosiah: makes sense :+1:

packages/devtools/devtools-extension/vite.config.ts
```diff
@@ -0,0 +1,111 @@
...
+  base: '', // Ensures relative path to assets.
+  optimizeDeps: {
```
wittjosiah, line 23: optimize deps can be omitted because it's only relevant for the dev server which I don't think is being used here?

packages/devtools/devtools-extension/vite.config.ts
```diff
@@ -0,0 +1,111 @@
...
+      '@dxos/rpc'
+    ]
+  },
+  build: {
+    commonjsOptions: {
+      include: [/packages/, /node_modules/]
+    },
+    rollupOptions: {
+      input: {
+        panel: resolve(__dirname, 'panel.html')
```
wittjosiah, line 52: other html files are also needed I assume?

wittjosiah (commented):
why did you remove the public folder? I think img should still be in there and maybe manifest unless that plugin outputs a manifest file as well

packages/devtools/devtools-extension/vite.config.ts
mykola-vrmchk: Other files are mentioned in the manifest and vite extension plugin handles them

packages/devtools/devtools-extension/vite.config.ts
mykola-vrmchk: Yeah, you are right

mykola-vrmchk:
> why did you remove the public folder? I think img should still be in there and maybe manifest unless that plugin outputs a manifest file as well

@wittjosiah plugin outputs everything including manifest and images

wittjosiah:
> > why did you remove the public folder? I think img should still be in there and maybe manifest unless that plugin outputs a manifest file as well
> 
> @wittjosiah plugin outputs everything including manifest and images

how does is know about the images? is `assets` a special path?

packages/devtools/devtools-extension/vite.config.ts
wittjosiah: in this case, is it necessary to import the manifest config from `manifest.json`? could you just have it as an object inline?

mykola-vrmchk: We can do inline object

PR #2401, opened 2023-01-30
feat(react-appkit): Generic Tiles

wittjosiah:
Introduces a generic `Tile` type which provides a `render` function as well as `canRender` validator.

Starting to play with this generic card as a POC in storybook.

### DIFF
--- a/packages/apps/patterns/react-appkit/src/components/Card/Card.tsx
+++ b/packages/apps/patterns/react-appkit/src/components/Card/Card.tsx
@@ -1,0 +1,58 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import React from 'react';
+
+import { DocumentBase, id, type } from '@dxos/echo-schema';
+import { Group } from '@dxos/react-components';
+
+import { Task, TaskList } from '../../proto';
+
+export type Card<T = any> = {
+  canRender: (data: DocumentBase) => boolean;
+  render: (props: { data: T }) => JSX.Element;
+};
+
+export const BaseCard = Group;
+
+export const DefaultCard: Card = {
+  canRender: (_data) => true,
+  render: ({ data }) => (
+    <BaseCard label={{ children: data ? data.title : 'Unknown object' }}>
+      <pre>{JSON.stringify(data, null, 2)}</pre>
+    </BaseCard>
+  )
+};
+
+export const TaskListCard: Card<TaskList> = {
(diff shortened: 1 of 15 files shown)


packages/apps/patterns/react-appkit/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,71 @@
...
+//
+
+import React from 'react';
+
+import { DocumentBase, id, type } from '@dxos/echo-schema';
+import { Group } from '@dxos/react-components';
+
+import { Task, TaskList } from '../../proto';
+
+export type Card<T = any> = {
```
wittjosiah, line 12:
```suggestion
export type Card<T = DocumentBase> = {
```

@dmaretskyi I think ideally it wouldn't use `any` however I had issues reading any properties out of `DocumentBase` because typescript gets angry (e.g., https://github.com/dxos/dxos/pull/2401/files#diff-ccfe78b5591675c1b7ef0aaca22664f18ec216651705ebf1a87abb8074432f73R22)

packages/apps/patterns/react-appkit/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,71 @@
...
+  canRender: (_data) => true,
+  render: ({ data }) => (
+    <BaseCard label={{ children: data ? data.title : 'Unknown object' }}>
+      <pre>{JSON.stringify(data, null, 2)}</pre>
+    </BaseCard>
+  )
+};
+
+export const TaskListCard: Card<TaskList> = {
+  // TODO(wittjosiah): Ideally the type string would be exported from the schema as a constant.
```
wittjosiah, line 29: @dmaretskyi I assume this would be a fairly trivial change?

thure (commented):
While this is fresh, I think we should consider alternatives to using ‘card’ to refer to this, since that already has strong referents in UI already that only partially overlap with what’s going on here.

I think we need a less overloaded term. ‘Tile’ was mentioned once before and I think it’s much more functional for this concept.

packages/apps/patterns/react-appkit/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,71 @@
...
+  canRender: (data) => data[type] === new TaskList()[type],
```
zhenyasav, line 30:
```suggestion
  canRender: (data) => data[type] === TaskList.type,
```

PR #2411, opened 2023-01-31
feat: agent enable gravity kibana logging

quaisx:


packages/common/log/src/processors/browser-processor.ts
```diff
@@ -62,10 +62,8 @@ const APP_BROWSER_PROCESSOR: LogProcessor = (config, entry) => {
 
   let args = [];
   args.push(entry.message);
-
-  const context = getContextFromEntry(entry);
-  if (context) {
-    args.push(context);
+  if (entry.context && Object.keys(entry.context).length > 0) {
```
dmaretskyi, line 65: Was this change intentional?

PR #2436, opened 2023-02-06
refactor: Refactor database protobufs

dmaretskyi:


packages/core/protocols/src/proto/dxos/echo/object.proto
```diff
@@ -15,66 +15,60 @@ package dxos.echo.object;
...
-    NOOP = 0;
-    DELETE = 1;
-    RESTORE = 2;
-  }
-
-  optional string parent_id = 1;  // TODO(burdon): Define Action type?
-  optional Action action = 2;
-}
+  optional timeframe.TimeframeVector timeframe = 4;
+}
```
mykola-vrmchk, line 74: new line

mykola-vrmchk, line 74:
```suggestion
}

```

PR #2454, opened 2023-02-08
feat: Factor out components from Kai

mykola-vrmchk:


mykola-vrmchk:
* main
  * **PR #2454** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2454" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2454?utm_source=stack-comment).

packages/devtools/devtools/vite.config.ts
```diff
@@ -26,6 +26,8 @@ export default defineConfig({
       '@dxos/config',
       '@dxos/keys',
       '@dxos/log',
+      '@dxos/mosaic',
+      '@dxos/plexus',
```
wittjosiah, line 30: are these not build with esm versions?

mykola-vrmchk: It is, but for some reason, testbench utilizing cjs entry point and fails with an error. 
image

wittjosiah: huh, interesting

mykola-vrmchk: I guess the issue is inside Kai, because I copied those from Kai

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
```diff
@@ -25,7 +24,20 @@ export const Searchbar: FC<{
   return (
     <div className='flex flex-1 flex-col bg-white'>
       <div className={mx('flex flex-1 items-center p-2', border && 'border rounded')}>
-        <Input className='w-full outline-0' spellCheck={false} onChange={handleChange} placeholder='Search...' />
+        <Input
+          label={undefined}
```
thure, line 28: `label` should never be undefined when required — use the same value as `placeholder` and set `slots: { label: { className: 'sr-only'}}}`

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
```diff
@@ -25,7 +24,20 @@ export const Searchbar: FC<{
...
+          onChange={(event) => handleChange(event.target.value)}
+          placeholder='Search...'
```
thure, line 30: This should be a call to `t`, e.g. `t('search placeholder')`, and the string should go in `en-US.ts`.

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
```diff
@@ -25,7 +24,20 @@ export const Searchbar: FC<{
...
+          slots={{
+            root: {
+              className: 'flex flex-1 mlb-0'
+            },
+            input: {
```
thure, line 35: This should probably also have `type: 'search'` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search)

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
```diff
@@ -25,7 +24,20 @@ export const Searchbar: FC<{
   return (
     <div className='flex flex-1 flex-col bg-white'>
       <div className={mx('flex flex-1 items-center p-2', border && 'border rounded')}>
-        <Input className='w-full outline-0' spellCheck={false} onChange={handleChange} placeholder='Search...' />
+        <Input
```
thure, line 27: `react-components` generally don't use each other since they’re meant to stand on their own. Since this is combining `Input` and `Button` elements, I think it probably belongs in `react-appkit`.

packages/common/react-components/src/components/Selector/Selector.tsx
```diff
@@ -5,9 +5,8 @@
 import { CaretDown, CaretUp } from 'phosphor-react';
 import React, { FC, KeyboardEvent, useState } from 'react';
 
-import { mx } from '@dxos/react-components';
-
-import { Input } from './Input';
+import { mx } from '../../util';
+import { Input } from '../Input';
```
thure, line 9: If this component does what I think it does, it needs to be reimplemented using Radix’s `Select` component which does a ton of the heavy lifting for us: https://www.radix-ui.com/docs/primitives/components/select

It can be styled using `tailwindcss-radix` which is part of our stack: https://tailwindcss-radix.vercel.app/#select

packages/common/react-components/src/components/Table/Table.tsx
```diff
@@ -5,7 +5,7 @@
 import React, { useMemo } from 'react';
 import { Column, useFlexLayout, useResizeColumns, useTable } from 'react-table';
 
-import { mx } from '@dxos/react-components';
+import { mx } from '../../util';
 
 // https://github.com/TanStack/table/blob/v7/examples/full-width-resizable-table/src/App.js
```
thure, line 11: This component is missing a ton of keyboard access and accessibility features, but we can address that later. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role

packages/common/react-components/src/components/TreeView/TreeView.tsx
```diff
@@ -5,7 +5,8 @@
 import { CaretDown, CaretRight } from 'phosphor-react';
 import React, { FC, ReactNode, useState } from 'react';
 
-import { getSize, mx } from '@dxos/react-components';
+import { getSize } from '../../styles';
+import { mx } from '../../util';
```
thure, line 10: I believe @richburdon was going to make this move, make sure these changes will gel with the ones he intends to make.

thure: Radix provides a `Collapsible` component which would do a decent job of patching-in good interactivity here. This component should eventually be refactored to use it: https://www.radix-ui.com/docs/primitives/components/collapsible

thure (changes requested):
The only big important things that I think need fixing ahead of merging this PR in my view are these:
- [ ] `Select` should use Radix’s `Select` component, we shouldn’t reinvent such a nuanced & complex wheel
- [ ] `Selector` should just be called `Select`
- [ ] `Searchbar` probably belongs in `react-appkit`

There are a few would-be-nice items and can-do-later items but I'd say they’re less important.

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
thure: I've been meaning to improve this API so it’s not so clunky, so apologies for the inconvenience here.

packages/common/react-components/src/components/Searchbar/Searchbar.tsx
thure: If you’re unfamiliar with `t`, look for `useTranslation` and how components are using it, or feel free to DM me.

thure (approved):
I can convert my feedback into follow-up items for the UXers.

PR #2457, opened 2023-02-08
feat: `EditableList`

thure:


packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export interface CrudListSlots {
+  root?: Omit<ComponentPropsWithoutRef<'div'>, 'children'>;
+  listHeading?: InputProps['slots'];
+  list?: ListProps['slots'];
+  addItem?: InputProps['slots'];
+}
+
+export interface CrudListProps {
```
richburdon, line 50: I don't love this name.
EditableList?

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+            event.preventDefault();
+            inputsInScope[targetIndex - 1].focus();
+          }
+          break;
+        }
+      }
+    },
+    [hostId]
+  );
+  return { onListItemInputKeyDown, hostAttrs, itemAttrs };
```
richburdon, line 106: style comment: would prefer newline before return statements. i prefer "code paragraphs" delimitered by newlines)

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export const CrudList = ({
+  children,
+  labelId,
+  nextItemTitle,
+  defaultNextItemTitle,
+  onNextItemTitleChange,
+  defaultItemIdOrder,
+  itemIdOrder,
+  onItemIdOrderChange,
```
richburdon: nice to group params in order: required, optional, callbacks (at least group callbacks)

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+          const oldIndex = itemIds.findIndex((id) => id === active.id);
+          const newIndex = itemIds.findIndex((id) => id === over.id);
+          return arrayMove(itemIds, oldIndex, newIndex);
+        } else {
+          return itemIds;
+        }
+      });
+    }
+  };
+  return (
```
richburdon, line 142: again white space between: hooks, callbacks, and return

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+        />
+        <ListItemEndcap className='invisible' />
+      </div>
+    </div>
+  );
+};
+
+export const CrudListItem = forwardRef<HTMLLIElement, CrudListItemProps>(
+  (
+    {
```
richburdon, line 195: to me it's clearer to have:

```ts
const Components = ({ a, b }) => {
  const { x, y } = b;
}
```

just easier to grok what are the params to the function when there are complex destructs.
just an opinion.

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
```diff
@@ -0,0 +1,226 @@
...
+      title,
+      onTitleChange,
+      slots = {}
+    }: CrudListItemProps,
+    forwardedRef
+  ) => {
+    return (
+      <ListItem
+        ref={forwardedRef}
+        {...{
```
richburdon, line 210: why `...`?

packages/common/react-components/src/components/List/List.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+// LIST ITEM
+
+const [createListItemContext, createListItemScope] = createContextScope(LIST_ITEM_NAME, []);
+
+type ListItemContextValue = { headingId: string };
+
+const [ListItemProvider, useListItemContext] = createListItemContext<ListItemContextValue>(LIST_ITEM_NAME);
+
+const ListItemEndcap = ({ children, className, ...props }: ComponentPropsWithoutRef<'div'>) => {
```
richburdon, line 141: this isn't an obvious name to me? vs, say, ListItemCheckbox, ListItemIcon and ListItemMenu (i.e., a bit more specialize)

packages/common/react-components/src/components/List/List.tsx
```diff
@@ -0,0 +1,270 @@
...
+  const listItemId = useId('listItem');
+
+  if (variant === 'ordered-draggable') {
+    return <DraggableListItem {...props} ref={forwardedRef} id={props.id ?? listItemId} />;
+  } else {
+    return <PureListItem {...props} ref={forwardedRef} id={props.id ?? listItemId} />;
+  }
+});
+
+export {
```
richburdon, line 259: sort

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
thure: That does sound better, will do

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
thure: Sure, I can do that. If this is a strong preference, we should add it to our formatting rules.

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
thure: I’m not grouping callbacks here because e.g. `itemIdOrder` comes in three since it is a controllable state:

- `defaultItemIdOrder` (uncontrolled value)
- `itemIdOrder` (controlled value)
- `onChangeItemIdOrder` (callback)

I feel these props should come together so it's easier to be sure I got all three for each prop that is part of a controllable state.

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
thure: There is not a complex destruct here, it is flat as in your first line. Maybe the `forwardRef` structure is confusing the matter?

packages/apps/patterns/react-appkit/src/components/CrudList/CrudList.tsx
thure: I frequently use object spreads instead of the equivalent `propName={propValue}` to take advantage of the same-key-name-as-variable-name shorthands, though I treat both syntaxes as equivalent for values that are not `key` (which I prefer to always use the usual JSX syntax because `key` is special).

I don't feel strongly about this pattern. If it’s causing ergonomic issues, let’s find a formatter rule that we think works best.

packages/common/react-components/src/components/List/List.tsx
thure: I’m concerned about specializing too early if for instance we need an element with all the same styles but the content happens to be a Dialog (i.e. it would host the Dialog’s trigger).

I notice Radix usually names element that just provide a surface like this something abstract, e.g. `Item` (taken in this context), `Action`, `Content`. I hoped `Endcap` would get the point across that it should just be used at the ends, but I'm amenable to a different name that doesn’t imply a specific content.

packages/common/react-components/src/components/List/List.tsx
thure: I’ve (tried to) meaningfully group these here, similar to your suggestion about grouping props:

1. Collection component
2. Collection hooks / management functions
3. Item component
4. Components of Items
5. Item hooks

packages/common/react-components/src/components/List/List.tsx
thure: Come to think of it, you’d been calling this ‘gutter’, so maybe `Gutter` would be more apt?

PR #2483, opened 2023-02-12
feat: Restructure containers (separate folders).

richburdon:


packages/apps/patterns/react-ui/src/layouts/PanelSidebar/PanelSidebar.tsx
```diff
@@ -76,17 +76,22 @@ export const PanelSidebarProvider = ({ children, slots }: PropsWithChildren<Pane
   return (
     <PanelSidebarContext.Provider value={{ setDisplayState, displayState }}>
       <DialogPrimitive.Root open={domShow} modal={!isLg}>
+        {/* Sidebar. */}
         <DialogPrimitive.Content
           {...slots?.content}
           className={mx(
-            'fixed block-start-0 block-end-0 is-sidebar z-50 overscroll-contain transition-[inset-inline-start,inset-inline-end] duration-200 ease-in-out overflow-x-hidden overflow-y-auto',
+            // TODO(burdon): Fix '__block-start-0'
+            'fixed __block-start-0 block-end-0 is-sidebar z-50 overscroll-contain overflow-x-hidden overflow-y-auto',
```
thure, line 84: I know there is an issue somewhere else here, but I don't think this solution is safe to commit.

Until we figure out the bug, you can use [the `!` modifier](https://tailwindcss.com/docs/configuration#important-modifier) in downstream. If even that doesn’t work, inline styles or arbitrary values should still work.

PR #2487, opened 2023-02-13
feat: Add Root `Document` with Space meta data

mykola-vrmchk:


mykola-vrmchk:
* main
  * **PR #2487** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2487" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2487?utm_source=stack-comment).

packages/experimental/kai/src/containers/SpaceList/SpaceList.tsx
```diff
@@ -30,6 +31,7 @@ export type SpaceItemProps = {
 // TODO(burdon): Action menu.
 // TODO(burdon): Full width mobile.
 const SpaceItem = ({ space, selected, children, onAction }: SpaceItemProps) => {
+  const [meta] = useQuery(space, SpaceMeta.filter());
```
dmaretskyi, line 34: expose the document with `space.data`

it should always be present: on init wait for the item to be loaded into the database

in components you should be able to just write `{space.data.title ?? humanzie(space.key)}`. Wrapping the component with `withReactor` should add reactivity as you're accessing "title" on an ECHO object

dmaretskyi: Also, please override the type of `space.data` so that typescript doesnt error if you attach other props to it:

```
space.data.counter ??= 0;

space.data.counter++;
```

dmaretskyi: cc @zhenyasav

PR #2534, opened 2023-02-18
feat: Fix space truncation

richburdon:


packages/experimental/kai/src/containers/FrameRegistry/FrameRegistry.tsx
```diff
@@ -71,9 +69,9 @@ export const FrameRegistry: FC<{ slots?: FrameRegistrySlots }> = ({ slots = {} }
     switch (type) {
       case ExtensionType.FRAME: {
         const active = !activeFrames.find((frameId) => frameId === id);
-        setActiveFrame(id, active);
+        setActiveFrame(id, active); // TODO(burdon): Reconcile with navigation.
```
wittjosiah, line 72: I can try to reconcile the app navigation when I update shell to remove space selection from it

packages/sdk/react-client/src/echo/SpaceContext.tsx
```diff
@@ -34,7 +40,9 @@ export const SpaceProvider = ({ initialSpaceKey, onSpaceChange, children }: Spac
   });
 
   useEffect(() => {
-    onSpaceChange?.(spaceKey);
+    if (spaceKey) {
+      onSpaceChange?.(spaceKey);
+    }
```
wittjosiah, line 45: Why should deselection not fire this event?

packages/experimental/kai/src/pages/SpacePage.tsx
```diff
@@ -46,17 +45,19 @@ const Toolbar = () => {
...
 const SpacePage = () => {
-  useDevDataGenerator();
+  useGenerator();
   const { t } = useTranslation('kai');
   const { active } = useFrames();
   const { section } = useParams();
-  const { space, frame } = useFrameState();
-
+  const { space, frame } = useAppRouter();
+  if (!space) {
```
wittjosiah, line 53: Undefined space is handling below already in a way the allows a space to be selected

packages/experimental/kai/src/pages/IdentityPage.tsx
```diff
@@ -48,7 +47,7 @@ const IdentityPage = () => {
             </Button>
             <Button
               variant='primary'
-              onClick={() => navigate(createSpacePath(space?.key, defaultFrameId))}
+              onClick={() => navigate(createPath({ spaceKey: space!.key, frame: defaultFrameId }))}
```
wittjosiah, line 50: Can be guarantee this returns a space? If not this seems like it will break sooner or later

packages/experimental/kai/src/hooks/useAppRouter.ts
```diff
@@ -0,0 +1,105 @@
...
+  const identity = useIdentity();
+  const { spaceKey, section, frame, objectId } = useParams();
+
+  const space = spaceKey ? findSpace(spaces, spaceKey) : undefined;
+
+  useEffect(() => {
+    let t: ReturnType<typeof setTimeout>;
+    if (identity && !space) {
+      t = setTimeout(
+        oncePerWindow('echo/first-space', async () => {
```
wittjosiah, line 86: What do you prefer about this over shell creating the first space as the identity is created which was happening already? It seems to me like this may end up creating spaces more than just the first time the app is opened

packages/experimental/kai/src/hooks/useAppRouter.ts
```diff
@@ -0,0 +1,105 @@
...
+import { oncePerWindow } from './once';
+import { defaultFrameId, FrameDef, useFrames } from './useFrames';
+// TODO(burdon): Create defs/helpers for other routes.
+export enum Section {
+  FRAME = 'frame',
+  REGISTRY = 'registry',
+  SETTINGS = 'settings'
+}
+
+const truncateKey = (key: PublicKey) => key.toHex().slice(0, 8);
```
wittjosiah, line 21: If this is going to be the common way to use keys in paths, should we add a method to `PublicKey` for it? `PublicKey.short(length = 8)`?

packages/experimental/kai/src/hooks/useAppRouter.ts
```diff
@@ -0,0 +1,105 @@
...
+import { useEffect } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
+
+import { Invitation, InvitationEncoder, Space } from '@dxos/client';
+import { PublicKey } from '@dxos/keys';
+import { useClient, useIdentity, useSpaces } from '@dxos/react-client';
+
+import { oncePerWindow } from './once';
+import { defaultFrameId, FrameDef, useFrames } from './useFrames';
+// TODO(burdon): Create defs/helpers for other routes.
```
wittjosiah, line 14: newline

packages/experimental/kai/src/hooks/once.ts
```diff
@@ -0,0 +1,17 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+export const once = (storage: any) => (operationKey: string, fn: Function) => async () => {
```
wittjosiah, line 5: this doesn't look like a hook, so it should probably be moved elsewhere

packages/experimental/kai/src/hooks/useAppRouter.ts
```diff
@@ -0,0 +1,105 @@
...
+export const createPath = ({
+  spaceKey,
+  section,
+  frame,
+  objectId
+}: {
+  spaceKey?: PublicKey;
+  section?: string;
+  frame?: string;
+  objectId?: string;
```
wittjosiah, line 38: this should probably be renamed `documentId`, right?

packages/experimental/kai/src/containers/Sidebar/Sidebar.tsx
```diff
@@ -55,16 +55,15 @@ export const Sidebar = () => {
 
   const handleCreateSpace = async () => {
     const space = await client.echo.createSpace();
-    // await space.properties.set('title', 'XXX'); // TODO(burdon): Not implemented.
-    setSpace(space.key);
+    navigate(createPath({ spaceKey: space.key }));
```
wittjosiah, line 58: probably want to pass the default frame here?

packages/experimental/kai/src/containers/Sidebar/Actions.tsx
```diff
@@ -49,7 +48,7 @@ export const Actions = () => {
     const json = new TextDecoder('utf-8').decode(data);
     const space = await client.echo.createSpace();
     await serializer.import(space.db, JSON.parse(json));
-    navigate(createSpacePath(space.key));
+    navigate(createPath({ spaceKey: space.key }));
```
wittjosiah, line 51: probably want to pass the default frame here?

packages/sdk/react-client/src/echo/SpaceContext.tsx
richburdon: it probably should but i had some downstream issues -- e.g., that this would trigger creating a new space.
the app route/space selection/auto-creation needs a bit of a think.

packages/experimental/kai/src/pages/SpacePage.tsx
richburdon: I wasn't keen on there being naked text in the frame
i think after we solve the flickering frame issue, i want to think about a more aesthetic loading and selection ux

packages/experimental/kai/src/pages/IdentityPage.tsx
richburdon: i've changed this to have a handler, but we should discuss the onboarding flow -- esp. post shell integration.

packages/experimental/kai/src/hooks/useAppRouter.ts
richburdon: this is @zhenyasav's fix to get the demo working. i think we need to workshop this -- see above.

packages/experimental/kai/src/hooks/once.ts
richburdon: see above.

packages/experimental/kai/src/hooks/useAppRouter.ts
richburdon: I ECHO Objects are by default Documents.

PR #2546, opened 2023-02-21
ci: update release please git mechanics

wittjosiah:
The way this job is currently setup, `production` and `main` diverge because of merge commits. This causes the job to fail when doing a hotfix after there have been a number of releases. This change stops using merge commits and only ever fast-forwards `main` and `production` from the release branches; this ensures that they maintain the same history. `main` continues to fast-forward on every commit, ensuring that updates to the release are ported back to `main`, while `production` is only fast-forwarded once the release is created. By using fast-forward only, `production` is essentially just a time-delayed version of `main`. One caveat to this flow is that if there is ever a conflict the CI jobs will fail and need to be resolved manually, however this was also the case before as well.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -84,22 +84,14 @@ jobs:
           git add .
           git commit -m "chore: apidoc update for ${{ steps.release.outputs.tag_name }}"
 
-      - name: Merge with production
+      - name: Fast-forward production
         if: ${{ steps.release.outputs.release_created }}
         run: |
           git switch production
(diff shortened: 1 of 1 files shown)


wittjosiah:
* main
  * **PR #2546** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2546" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2546?utm_source=stack-comment).

zhenyasav (approved):

PR #2556, opened 2023-02-22
feat: Density

thure:
This PR adds `DensityProvider` and the concept of density to r-c.

Density values are currently based on the [CSS `pointer` media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer): they are either `coarse` or `fine`.

For Button and Input, coarse is always 40px and fine is always 32px. These are minimums, so developers will need to add rules to handle truncation.

e.g.:
```tsx
<DensityProvider density='coarse'>
    <Button>Text</Button>
    <Input
      label='excluded'
      variant='subdued'
      labelVisuallyHidden
      placeholder='Type something'
      defaultValue='Editable text'
    />
    <Button>
      <PaperPlane className={getSize(6)} />
    </Button>
</DensityProvider>
```

Screenshot 2023-02-21 at 18 44 08


packages/common/react-components/src/components/DensityProvider/DensityProvider.tsx
```diff
@@ -0,0 +1,21 @@
...
+  density?: Density;
+}
+
+export type DensityProviderProps = PropsWithChildren<{
+  density?: Density;
+}>;
+
+export const DensityContext = createContext<DensityContextValue>({ density: 'coarse' });
+
+export const DensityProvider = ({ density, children }: DensityProviderProps) => (
```
dmaretskyi, line 19: css props might be an alternative to implement this https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#basic_usage

thure: I know, I've considered it. I think with something like Density, this might need to scale to affect several CSS properties, e.g. line-height, padding, margin; and I think even behavior may change, e.g. lists aren’t draggable in `fine` density.

Even with `Elevation` which would reasonably be a good target for CSS properties, Tailwind does a lot of magic with its own CSS custom properties (see [box-shadow color](https://tailwindcss.com/docs/box-shadow-color)) and trying to carry that in a custom property the design system manages proved to be unworkable.

I think for now, keeping styles and things-that-are-partly-styles in JS is our best bet.

packages/common/react-components/src/props/Density.ts
```diff
@@ -0,0 +1,5 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+export type Density = 'coarse' | 'fine';
```
richburdon, line 5: I like that these are literal with respect to the quality of density :). Although I think `compact` and `full` (default) are a little less "abstract".

packages/common/react-components/src/components/DensityProvider/DensityProvider.tsx
```diff
@@ -0,0 +1,21 @@
...
+  <DensityContext.Provider value={{ density }}>{children}</DensityContext.Provider>
```
richburdon, line 20: While this technically is a provider from the react framework POV, to a user, this is a container so perhaps there's another abstraction like `<ControlGroup>` where one might specify density and elevation, etc. to influence a more horizontal provider (rather than multiple).

```tsx
<Group density='coarse' elevation={2} variant='ghost'>
  <Button>1</Button>
  <Button>2</Button>
  <Button variant='outline'>3</Button>
</Group>
```

richburdon (approved):

packages/common/react-components/src/props/Density.ts
thure: That's fair, though I wanted to pick a continuum that 1) indicates a connection to interactive constraints and 2) can potentially map to Tailwind-style “t-shirt sizes”, e.g. if we were to introduce ‘extra’ values, to me ‘extra-coarse’ is more readily understood than ‘extra-full’.

packages/common/react-components/src/components/DensityProvider/DensityProvider.tsx
thure: Sure — I don’t anticipate many will want to use `DensityProvider` directly. `List` and others already use it internally to communicate density to their children, and I’d imagine other grouping components could offer `density` as a prop and do the same like you illustrate here.

PR #2560, opened 2023-02-22
feat: Notes Board bug fixes and content document.

richburdon:
Bugs:

 - [ ] on being invited currently the joiner doesn't go to the right space
 - [ ] the mosaic Grid layout currently doesn't get updated when a Tile is moved (since it's a mutation within the Board's array and I don't know how to make that reactive)
 - [ ] seeing a bug with the textobject not synching, but not happening 100% of the time.
`Cannot read properties of undefined (reading doc).`

https://user-images.githubusercontent.com/3523355/220754947-0761dbfc-6347-47b8-bcdc-7c096b9d2e87.mov


packages/common/react-components/src/components/Input/Input.tsx
```diff
@@ -56,7 +57,7 @@ export const Input = ({
     ...(autoFocus && !hasIosKeyboard && { autoFocus: true }),
     disabled,
     placeholder,
-    value,
+    value: value ?? '',
```
thure, line 60: In some situations this has locked the input’s value at empty for me, I'd just make sure that isn’t happening here.

packages/common/react-components/src/components/Dialog/Dialog.tsx
```diff
@@ -30,13 +30,12 @@ export interface DialogProps {
   description?: ReactNode;
   children?: ReactNode;
   closeLabel?: string;
-  initiallyOpen?: boolean; // TODO(burdon): Change to just 'open'.
+  initiallyOpen?: boolean; // TODO(burdon): Change to just 'open' (i.e., controlled and non-controlled?)
```
thure, line 33: [Radix’s Dialog](https://www.radix-ui.com/docs/primitives/components/dialog)
> Can be controlled or uncontrolled.

Other aspects of r-c’s Dialog need to be refactored too, so I'll be sure to include this if not already addressed.

packages/experimental/kai/src/containers/FrameContainer/FrameContainer.tsx
```diff
@@ -16,5 +16,9 @@ export const FrameContainer: FC<{ frame: FrameDef }> = ({ frame }) => {
   }
 
   // TODO(burdon): Standardize container (bg, padding, centered, size, etc.)
-  return <Component />;
+  return (
+    <Suspense>
```
thure, line 20: It’s usually good practice to provide a fallback, e.g. `<Suspense fallback={<Loading/>}>…`

packages/experimental/kai/src/containers/SpaceList/SpaceList.tsx
```diff
@@ -86,43 +91,31 @@ const SpaceItem = withReactor(({ space, selected, children, onAction }: SpaceIte
...
       {spaces.map((space) => (
         <SpaceItem
           key={space.key.toHex()}
           space={space}
           selected={selected && space.key.equals(selected)}
-          onAction={(action) => handleAction(space.key, action)}
+          onAction={(intent) => onAction?.(intent)}
         >
-          {Tile && <Tile />}
+          <Suspense>{Tile && <Tile />}</Suspense>
```
thure, line 118: Ditto here, though it seems like the best fallback wouldn’t be a spinning loader but a shimmering text skeleton. We should add that to r-c.

packages/experimental/kai/src/frames/Note/NoteFrame.tsx
```diff
@@ -42,17 +66,24 @@ export const TileContent = withReactor(({ item, selected, onDelete }: TileConten
...
+            }}
             placeholder='Title'
-            autoFocus={selected} // TODO(burdon): Not working.
-            value={item.label}
-            onChange={({ target: { value } }) => {
-              // TODO(burdon): Breaks abstraction: bubble change event instead.
-              const note = item.data! as Note;
-              note.title = value;
+            slots={{
+              label: { className: 'sr-only' },
```
thure, line 79: I keep forgetting, but if the `label` slot is just the `sr-only` classname, you can instead use the boolean prop `labelVisuallyHidden` which does this for you.

packages/experimental/kai/src/frames/Note/NoteFrame.tsx
```diff
@@ -42,17 +66,24 @@ export const TileContent = withReactor(({ item, selected, onDelete }: TileConten
...
+              root: {
+                className: 'm-0 w-full'
+              },
+              input: {
+                className: 'p-1 w-full border-0 text-xl',
+                autoFocus: true // TODO(burdon): Apply selectively?
```
thure, line 85: This already is applied somewhat selectively; `autoFocus` isn’t applied in iOS to prevent the mixed focus state there.

packages/experimental/mosaic/src/layout/layout.tsx
```diff
@@ -31,7 +31,7 @@ export interface Layout {
 export type Item<T extends {} = {}> = {
   id: string;
   data?: T;
-  label: string;
+  title: string;
```
thure, line 34: I've always waffled around between `label` and `title` personally… here’s what I know with some certainty:

- Objects with `label` should have an optional `description`, these are first-class UI concepts that go together
- `label` is a UI idea rather than a kind of user content
- Objects with `title` could also have `body` for the main text content and `abstract` which summarizes the body (can be derived directly from `body` or not)
- `title` is a kind of user content rather than part of the software (with certain rare exceptions)

packages/experimental/mosaic/src/testing/testing.ts
```diff
@@ -8,7 +8,7 @@ import { Item, Location } from '../layout';
 
 export const createItem = (location?: Location): Item => ({
   id: faker.datatype.uuid(),
-  label: faker.lorem.words(3),
+  title: faker.lorem.words(3),
   content: faker.lorem.sentences(faker.datatype.number(3)),
```
thure, line 12: On the above topic, I’d reconsider `content` since it’s possibly too vague. Maybe `body` is a good candidate.

thure (approved):
lgtm, just had some non-blocking thoughts.

packages/common/react-components/src/components/Input/Input.tsx
richburdon: I've used this pattern for a long time and not seen that but i will watch out for that.

packages/experimental/kai/src/containers/FrameContainer/FrameContainer.tsx
richburdon: Unless there's a really long block (>1s) then this results in flickering and as a user that means to me a broken app.
What I propose is a fallback that is silent for the first 200ms but then if triggered is visible for at lest 500ms. OR is much more subtle -- e.g., a 2px linear line at the bottom of the screen that is overlaid and familiar.
Anyway to discuss.

packages/experimental/kai/src/containers/SpaceList/SpaceList.tsx
richburdon: ok -- you just resolved my issue above :).

packages/experimental/kai/src/frames/Note/NoteFrame.tsx
richburdon: previously i used a ref to programmatically set the focus. i assume i can do that with Input via forwardrefs?

packages/experimental/mosaic/src/layout/layout.tsx
richburdon: Ah nice.
I was in a slightly different place -- but have also been switching back and forth.
I think I prefer title and body as UI concepts since they shadow concepts in HTML; this way there are less likely to be confused with user data structures. BUT, as I think we've been in agreement from the outset, components should define their own data semantics and either provide accessors or some more advanced "model" interface that clients resolve.

packages/experimental/mosaic/src/testing/testing.ts
richburdon: However this is a test class -- so it's representative of some end user concept not the framework. so on that basis it would be better to feel like a real world data type.

packages/experimental/kai/src/frames/Note/NoteFrame.tsx
richburdon: is the label useful for any reason (e.g., documentation) in SR?

PR #2573, opened 2023-02-24
feat: react-components/select

zhenyasav:


packages/apps/halo-app/src/components/AlertDialog/AlertDialog.tsx
```diff
@@ -2,7 +2,7 @@
 // Copyright 2022 DXOS.org
 //
 
-import { Transition } from '@headlessui/react';
+import { Transition } from '@headlessui/react';s
```
thure, line 5: Typo?

packages/common/react-components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,96 @@
...
+        </Button>
+      </SelectPrimitive.Trigger>
+      <SelectPrimitive.Content>
+        <SelectPrimitive.ScrollUpButton className='flex items-center justify-center text-gray-700 dark:text-gray-300'>
+          <div className=''>
+            <CaretUp />
+          </div>
+        </SelectPrimitive.ScrollUpButton>
+        <SelectPrimitive.Viewport className='bg-white dark:bg-gray-800 p-2 rounded-lg shadow-lg'>
+          <SelectPrimitive.Group>
```
thure, line 70: I was imagining we’d export this separately, I think the developer should assemble groups and options together rather than using props.

packages/common/react-components/src/components/index.ts
```diff
@@ -26,4 +26,5 @@ export * from './Tooltip';
 export * from './ThemeProvider';
 export * from './TreeView';
 export * from './Selector';
+export * from './Select';
```
thure, line 29: Can we delete Selector with this PR?

zhenyasav: yeah I think i'll replace the other usages first

packages/common/react-components/src/components/Select/Select.tsx
thure: I say this mainly because the mental model for this package is in-flux, and for ergonomic reasons we should settle on at least a rule of thumb: if it’s a collection, the collection items are components you render as children.

zhenyasav: for now i'm on the side of doing that when a new usage pops up requiring groups (and we can also make a minor effort to avoid needing that in the product for a while ;) )

zhenyasav: we agreed to use children and it looks nicer, but i witheld adding some slots people might want in the Select.Items for later when someone needs it

packages/devtools/devtools/src/panels/mesh/SignalMessages.tsx
```diff
@@ -104,67 +105,49 @@ const types: ColumnType<SignalResponse>[] = [
...
+export const SignalMessages = (props: SignalMessagesProps) => {
+  const { messages } = { messages: [], ...props };
+  const [viewType, setViewType] = useState<ViewType>('swarm-event');
+  const [search, setSearch] = useState('');
+  const view = viewType ? getView(viewType) : undefined;
+  const filteredMessages = getFilteredData(messages, view, search);
   return (
     <div className='flex flex-col flex-1 overflow-hidden'>
       <div className='flex p-3 border-b border-slate-200 border-solid'>
-        <div className='flex'>
```
zhenyasav, line 154: @mykola-vrmchk check out all these useless divs we had

packages/devtools/devtools/src/panels/mesh/SignalMessages.tsx
```diff
@@ -104,67 +105,49 @@ const types: ColumnType<SignalResponse>[] = [
...
-
-    return () => {
-      signalOutput.close();
-    };
-  }, []);
+export type SignalMessagesProps = {
+  messages?: SignalResponse[];
+};
 
+export const SignalMessages = (props: SignalMessagesProps) => {
```
zhenyasav, line 118: @mykola-vrmchk check out this "pure" version of SignalMessages and it's counterpart `SignalMessagesContainer` which does 'just the data loading' for it

packages/devtools/devtools/src/panels/mesh/SignalMessages.tsx
```diff
@@ -104,67 +105,49 @@ const types: ColumnType<SignalResponse>[] = [
       }
     ]
   }
-];
+] as const; // this is ok because getView below will fail typecheck if this array is misdefined
```
zhenyasav, line 108: @mykola-vrmchk check out this use of `as const` to be able to pull types out of the array below as `ViewType`

packages/devtools/devtools/src/panels/mesh/SignalMessages.stories.tsx
```diff
@@ -0,0 +1,17 @@
+//
```
zhenyasav, line 1: @mykola-vrmchk i made snippets for this and react components, try writing `story` or `component` in a `tsx` file

zhenyasav: fyi @wittjosiah

packages/devtools/devtools/src/panels/mesh/SignalMessages.tsx
zhenyasav: this makes it easy to build inside a storybook story as opposed to in-app

packages/devtools/devtools/src/panels/mesh/SignalMessagesContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+//
+
+import React, { useEffect, useState } from 'react';
+
+import { SignalResponse } from '@dxos/protocols/proto/dxos/devtools/host';
+import { useDevtools } from '@dxos/react-client';
+
+import { SignalMessages } from './SignalMessages';
+
+const SignalMessagesContainer = () => {
```
zhenyasav, line 12: @mykola-vrmchk this is `SignalMessagesContainer` the data-loading wrapper for the pure `SignalMessages` (note pure doesn't mean it can't encapsulate private state, just means it's separate from the "app-specific" data environment)

PR #2580, opened 2023-02-24
feat: `withReactorRef` HOC for elements that forward refs that also need withReactor

thure:


zhenyasav:
should we just make this ref forwarding the default behavior in `withReactor`? @wittjosiah @dmaretskyi

PR #2631, opened 2023-03-02
chore: Add error in EchoArray for regular objects 

mykola-vrmchk:


mykola-vrmchk:
* main
  * **PR #2631** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2631" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2631?utm_source=stack-comment).

packages/core/echo/echo-schema/src/echo-array.ts
```diff
@@ -330,6 +330,14 @@ export class EchoArray<T> implements Array<T> {
     ) {
       return new Reference((value as any)['@id']);
     } else {
+      assert(
+        value === null ||
+          value === undefined ||
+          typeof value === 'boolean' ||
+          typeof value === 'number' ||
+          typeof value === 'string',
+        'Not EchoObject or primitive value'
```
richburdon, line 339: `Invalid type: ${JSON.stringify(value)}`

Short errors with data.

packages/core/echo/echo-typegen/src/codegen.ts
```diff
@@ -125,20 +125,10 @@ export const generate = (root: pb.NamespaceBase): string => {
...
-    yield text`
-      export namespace ${ns.name} {
-        ${ns.nestedArray.flatMap((nested) => Array.from(emitDeclarations(nested)))}
-      }
-      
-    `;
-  }
-
   if (ns instanceof pb.Type) {
     if (ns.name === 'Text') {
```
richburdon, line 133: extract any string literals that are used more than once?

richburdon (approved):

packages/core/echo/echo-schema/src/echo-array.ts
```diff
@@ -330,6 +330,14 @@ export class EchoArray<T> implements Array<T> {
     ) {
       return new Reference((value as any)['@id']);
     } else {
+      assert(
```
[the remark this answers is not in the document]
mykola-vrmchk: > above `=== object should mean !== null`?
There is a check for that, or I am misunderstanding the comment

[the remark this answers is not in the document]
mykola-vrmchk: > also why do we test the length of the number of properties?

I don't know for sure, I guess, it is some kind of Proxy detection (@dmaretskyi ?)

packages/core/echo/echo-typegen/src/codegen.ts
mykola-vrmchk: Sure

PR #2670, opened 2023-03-07
feat: Add specialized RPC implementation for client-services

dmaretskyi:


packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -117,11 +131,38 @@ export class ClientServicesHost {
     return this._serviceRegistry.services;
   }
 
+  /**
+   * Initialize the service host with the config.
+   * Config can also be provided in the constructor.
+   * Can only be called once.
+   */
+  initialize({ config, networkManager }: InitializeOptions) {
```
wittjosiah, line 139: why can this only be called once? seems like it would be simpler if it could just be a part of `open`?

packages/sdk/client-services/src/packlets/vault/worker-session.ts
```diff
@@ -48,34 +45,49 @@ export class WorkerSession {
...
-      IdentityService: () => this._getService((services) => services.IdentityService),
-      SpaceInvitationsService: () => this._getService((services) => services.SpaceInvitationsService),
-      SpacesService: () => this._getService((services) => services.SpacesService),
-      SystemService: () => this._getService((services) => services.SystemService),
-      TracingService: () => this._getService((services) => services.TracingService)
+    this._serviceHost = serviceHost;
+
+    const middleware: Pick<ClientRpcServerParams, 'handleCall' | 'handleStream'> = {
+      handleCall: async (method, params, handler) => {
+        const error = await readySignal.wait({ timeout: 3_000 });
```
wittjosiah, line 64: could we create a single file inside `@dxos/client` that defines our various timeout lengths and then start to use those everywhere?

packages/sdk/client-services/src/packlets/vault/worker-session.ts
```diff
@@ -48,34 +45,49 @@ export class WorkerSession {
...
-      DevtoolsHost: () => this._getService((services) => services.DevtoolsHost),
-      NetworkService: () => this._getService((services) => services.NetworkService),
-      IdentityService: () => this._getService((services) => services.IdentityService),
-      SpaceInvitationsService: () => this._getService((services) => services.SpaceInvitationsService),
-      SpacesService: () => this._getService((services) => services.SpacesService),
-      SystemService: () => this._getService((services) => services.SystemService),
-      TracingService: () => this._getService((services) => services.TracingService)
+    this._serviceHost = serviceHost;
+
+    const middleware: Pick<ClientRpcServerParams, 'handleCall' | 'handleStream'> = {
```
wittjosiah, line 62: `middleware` feels like the wrong word here... `handlers`?

packages/sdk/client-services/src/packlets/services/client-rpc-server.ts
```diff
@@ -0,0 +1,92 @@
...
+    const { serviceRegistry, handleCall, handleStream, ...rpcOptions } = params;
+    this._handleCall = handleCall;
+    this._handleStream = handleStream;
+
+    this._serviceRegistry = serviceRegistry;
+    this._rpcPeer = new RpcPeer({
+      ...rpcOptions,
+      callHandler: (method, params) => {
+        const [serviceName, methodName] = parseMethodName(method);
+        const service = this._getServiceHandler(serviceName);
```
wittjosiah, line 44: should this be called inside the handler? what if the service isn't available here but then is after the wait inside `this._handleCall`? perhaps worth testing missing service + dynamic service?

packages/sdk/client-services/src/packlets/vault/worker-session.ts
dmaretskyi: 👍

packages/sdk/client-services/src/packlets/services/service-host.ts
dmaretskyi: Passing in config, and networkManager over and over again when `open` and `close` are called multiple times feels kinda wrong. Im thinking `initialize` is to setup/configure the host and then `open` and `close` are the lifecycle.

PR #2713, opened 2023-03-10
feat(react-composer): StackEdit-style Markdown editing experience

thure:
https://user-images.githubusercontent.com/855039/224458960-369f2c68-0c9d-4321-8e2c-307ca323b8d6.mp4


wittjosiah (commented):
Questions based on the screenshot you sent:
- could the url part of links be dimmed as well? I think that's nice because it also is not visible text in the rendered document
- do we want the current line to be highlighted?
- does it understand task list syntax `- [ ]`?

packages/apps/patterns/react-composer/src/components/Markdown/codeMirrorDark.ts
```diff
@@ -0,0 +1,158 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { tags } from '@lezer/highlight';
+import get from 'lodash.get';
+
+import { tokens } from '../../styles';
+
+export const chalky = '#e5c07b';
```
richburdon, line 10: i realize this editor is rather unique -- and potentially user customizable, but i'm just curious how you think about families of colors.
should we have a package-global file to define colors?
example: i want at some point to adapt my "presenter" frame and may want to share a theme.

packages/apps/patterns/react-composer/src/styles/markdown.ts
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { mx } from '@dxos/react-components';
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
+
+export const heading: Record<HeadingLevel, string> = {
```
richburdon, line 9: related to question about colors above -- re handling custom themes (colors, sizes) for a possible presentation view. this may not be necessary since in my demo i just scale the container so the geometry didn't change.

richburdon (approved):
Fantastic.

packages/apps/patterns/react-composer/src/components/Markdown/codeMirrorDark.ts
thure: Yeah, I was wondering about that myself… I think our Tailwind config should export its own palettes based on LAB space so these shade numbers actually mean something, then we'd also have these values available at runtime for situations like this.

packages/apps/patterns/react-composer/src/styles/markdown.ts
thure: The sizes are currently available in our Tailwind config, so if you need to use those at runtime they can be retrieved.

PR #2719, opened 2023-03-11
feat: Scroll Container and Experimental cards.

richburdon:
Screenshot 2023-03-11 at 1 39 22 PM
Screenshot 2023-03-11 at 1 40 53 PM

### DIFF
--- a/packages/experimental/mosaic/src/components/Card/Card.stories.tsx
+++ b/packages/experimental/mosaic/src/components/Card/Card.stories.tsx
@@ -1,0 +1,112 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import faker from 'faker';
+import { Buildings, DotsThree, Envelope, User } from 'phosphor-react';
+import React from 'react';
+
+// TODO(burdon): Add note to @dxos/react-components/plugin
+// storybook config (files and modules)
+// dxosTheme import
+
+import { Button, DensityProvider, getSize } from '@dxos/react-components';
+import '@dxosTheme';
+
+import { Card } from './Card';
+import { CardActions } from './CardActions';
+import { CardContent } from './CardContent';
+import { CardHeader } from './CardHeader';
+
+// TODO(burdon): Copy storybook structure from @dxos/react-components.
+
+// TODO(burdon): IconButton.
+// TODO(burdon): Density: Input, Checkbox.
+// TODO(burdon): ListItem as CardHeader.
+
+// TODO(burdon): Column on mobile; no rounded corner if mobile column.
+
+// TODO(burdon): Factor out default semantic colors.
+
+const styles = {
+  header: 'bg-blue-300 dark:bg-zinc-800'
+};
+
+// TODO(burdon): Test embedded table/lists (e.g., checkbox in side gutter).
+
+const Test = () => {
+  return (
+    <div className='flex flex-col w-full md:w-[390px] m-4 space-y-8'>
+      <DensityProvider density='fine'>
+        <Card slots={{ root: { className: 'text-sm' } }}>
+          <CardHeader
+            slots={{ root: { className: styles.header } }}
+            icon={<User className={getSize(6)} />}
+            action={<DotsThree className={getSize(6)} />}
+          >
+            {faker.lorem.sentence()}
+          </CardHeader>
+          {/* TODO(burdon): Fade bottom. */}
+          <CardContent
+            icon={<Buildings className={getSize(6)} />}
+            scrollbar
+            slots={{
+              root: { className: 'h-[160px]' },
+              body: { className: 'pt-2.5' }
+            }}
+          >
+            {faker.lorem.sentences(32)}
+          </CardContent>
+          <CardContent
+            icon={<Envelope className={getSize(6)} />}
+            slots={{
+              body: { className: 'pt-2.5' }
+            }}
+          >
+            {faker.internet.email()}
+          </CardContent>
+          <CardContent gutter>{faker.lorem.sentences(3)}</CardContent>
+          <CardActions gutter slots={{ root: { className: 'bg-zinc-100' } }}>
+            <Button variant='ghost'>OK</Button>
+            <Button variant='ghost'>Cancel</Button>
+          </CardActions>
+        </Card>
+
+        <Card>
+          <CardHeader icon={<User className={getSize(6)} />} action={<DotsThree className={getSize(6)} />}>
+            {faker.lorem.sentence()}
+          </CardHeader>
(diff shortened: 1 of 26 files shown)


packages/experimental/mosaic/src/components/Card/util.tsx
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import React, { FC, ReactNode } from 'react';
+
+import { mx } from '@dxos/react-components';
+
+// TODO(burdon): Factor out as hook.
```
thure, line 9: Why would this need to be a hook? It’s okay if they’re just string constants IMO.

packages/experimental/mosaic/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  header: 'bg-blue-300 dark:bg-zinc-800'
+};
+
+// TODO(burdon): Test embedded table/lists (e.g., checkbox in side gutter).
+
+const Test = () => {
+  return (
+    <div className='flex flex-col w-full md:w-[390px] m-4 space-y-8'>
+      <DensityProvider density='fine'>
+        <Card slots={{ root: { className: 'text-sm' } }}>
```
thure, line 41: To your point about aiming to surface an API that minimizes the need for style interventions, perhaps the story here should demonstrate what Cards look like with zero intervention.

packages/experimental/mosaic/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <CardHeader
+            slots={{ root: { className: styles.header } }}
+            icon={<User className={getSize(6)} />}
+            action={<DotsThree className={getSize(6)} />}
```
thure, line 45: Here’s where things get dicey, generally the Radix-style subcomponents are _for_ avoiding props which are literal `React.ReactNode[]`s, which should either be just `props.children` or their own subcomponent so we’re not proliferating standards 3 ways (since `slots.{slot}.children` could also be used for this).

I think these should probably use a similar structure to list item endcaps, where endcaps (or whatever else we should call square layout parts) are used to defer semantics to the developer or other libraries so this package can focus on layout and a coherent API surface.

packages/experimental/mosaic/src/components/Card/CardActions.tsx
```diff
@@ -0,0 +1,30 @@
...
+  slots?: CardActionsSlots;
+  children?: ReactNode;
+  gutter?: boolean;
+};
+
+export const CardActions = ({ slots = {}, gutter, children }: CardActionsProps) => {
+  return (
+    <div className={mx('flex h-[40px] px-2', slots.root?.className)}>
+      {gutter && <Icon />}
+      <div className={mx('flex h-[40px] space-x-4', slots.root?.className)}>{children}</div>
```
thure, line 27: Why is the same slot’s `className` applied here as well?

packages/experimental/mosaic/src/components/Card/CardActions.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export type CardActionsProps = {
+  slots?: CardActionsSlots;
+  children?: ReactNode;
+  gutter?: boolean;
+};
+
+export const CardActions = ({ slots = {}, gutter, children }: CardActionsProps) => {
+  return (
+    <div className={mx('flex h-[40px] px-2', slots.root?.className)}>
```
thure, line 25: The other slot props should be spread before `className`, `{...slots.root}`

thure: Ah, I see only `className` is permitted… is there a reason `ComponentPropsWithoutRef<'div'>` won’t do?

packages/experimental/mosaic/src/components/Card/util.tsx
```diff
@@ -0,0 +1,23 @@
...
+  divide: 'divide-y divide-y-reverse __divide-neutral-300',
+  heading: 'text-black dark:text-white',
+  body: 'text-neutral-600'
+};
+
+// TODO(burdon): ListItemEndcap density.
+export const Icon: FC<{ className?: string; children?: ReactNode }> = ({ className, children }) => (
+  <div className={mx('flex flex-col shrink-0 items-center justify-center w-[40px] h-[40px]', className)}>
+    {children}
+  </div>
```
thure, line 22: This should probably live in its own file or alongside another component where it is most often used.

packages/experimental/mosaic/src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+    if (modified) {
+      this.onChange.emit();
+    }
+  }
+
+  toggleZoom() {
+    this.setAspect({ zoom: this.zoom === 1 ? this._options.zoomOut : 1 });
+  }
+}
```
thure, line 70: Usually it’s more ergonomic to put class definitions in their own `.ts` file.

packages/experimental/mosaic/src/components/ScrollContainer/ScrollContainer.tsx
```diff
@@ -0,0 +1,96 @@
...
+  viewport?: {
+    className?: string;
+  };
+  scrollbar?: {
+    className?: string;
+  };
+  thumb?: {
+    className?: string;
+  };
+};
```
thure, line 23: Awesome.

packages/experimental/mosaic/src/components/ScrollContainer/ScrollContainer.tsx
```diff
@@ -0,0 +1,96 @@
...
+// TODO(burdon): System colors.
+// TODO(burdon): Is this right? Merge?
+export const defaultSlots: ScrollContainerSlots = {
+  scrollbar: {
+    className: 'bg-neutral-50 hover:bg-neutral-100'
+  },
+  thumb: {
+    className: 'bg-neutral-400'
+  }
+};
```
thure, line 34: I’d prefer to put these in a `scrollContainerSyles.ts` file if other components might need them, otherwise just use them as literals in the component itself.

thure (commented):
Generally it looks good, I think it’s worth considering whether `Card` wants to be a member of our low-level unopinionated design system (in which case we should groom its API surface so it isn’t conflating semantics with layout as much), or if it’s a walled garden in which case it can do what it wants and might want to enforce stronger opinions about which subcomponents are used where (i.e. not export subcomponents).

PR #2730, opened 2023-03-14
feat: Kai search results

richburdon:


packages/experimental/kai/src/containers/SearchPanel/SearchPanel.tsx
```diff
@@ -159,8 +160,12 @@ export const SearchPanel = ({ onSelect }: SearchPanelProps) => {
...
 
   const { space } = useAppRouter();
-  const results = useQuery(space).map(matchFilter(text)).filter(Boolean) as SearchResult[];
-  const sorted = results.sort(sort);
+  const results = useQuery(space).map(matchFilter(text));
+  const sorted = (results.filter(Boolean) as SearchResult[]).sort(sort);
+  useEffect(() => {
+    // TODO(burdon): Causes recursion.
+    // onResults?.(sorted);
+  }, [results]); // TODO(burdon): The `sorted` array would always be updated.
```
dmaretskyi, line 168:
```suggestion
  }, [...getIdsFromSelection(result)]); // TODO(burdon): The `sorted` array would always be updated.
```

packages/experimental/kai/src/containers/SearchPanel/SearchPanel.tsx
```diff
@@ -159,8 +160,12 @@ export const SearchPanel = ({ onSelect }: SearchPanelProps) => {
   };
 
   const { space } = useAppRouter();
-  const results = useQuery(space).map(matchFilter(text)).filter(Boolean) as SearchResult[];
-  const sorted = results.sort(sort);
+  const results = useQuery(space).map(matchFilter(text));
```
dmaretskyi, line 163:
```suggestion
  const results = useQuery(space, matchFilter(text), [text]);
```

packages/common/react-components/src/components/Dialog/Dialog.tsx
```diff
@@ -89,6 +89,7 @@ export const Dialog = ({
           onCloseAutoFocus={(event) => event.preventDefault()}
           {...slots.content}
           className={mx(
+            'flex flex-col',
```
thure, line 92: `flex-col` by default can cause very subtle bugs except in those very specific cases where the element has an extrinsic size and one of its children should grow to fill it. Only part of that scenario is present here, so there is no cause to have `flex-col`.

packages/common/react-components/src/components/Dialog/Dialog.tsx
```diff
@@ -101,10 +102,11 @@ export const Dialog = ({
             <DialogPrimitive.Title
               {...slots.title}
               className={mx(
+                'shrink-0',
                 'text-xl font-system-medium text-neutral-900 dark:text-neutral-100 rounded-md',
                 titleVisuallyHidden && 'sr-only',
                 defaultFocus,
-                slots.content?.className
+                slots?.title?.className
```
thure, line 109: Ah, thank you for fixing this

packages/common/react-components/src/components/Dialog/Dialog.tsx
thure: If the Dialog that Kai is consuming often has an extrinsic size and `flex-col`, kai can reduce classname repetition in slots by defining its own Dialog component that sets those defaults.

packages/common/react-components/src/components/Table/Table.tsx
```diff
@@ -74,11 +74,14 @@ export const Table = <T extends {}>({ columns, data = [], slots = {}, onSelect,
         <div className='thead sticky top-0'>
           {headerGroups.map((headerGroup) => (
             // eslint-disable-next-line react/jsx-key
-            <div {...headerGroup.getHeaderGroupProps()} className='tr h-[2.5rem] border-b border-zinc-300'>
+            <div {...headerGroup.getHeaderGroupProps()} className='tr h-[2rem] border-b border-zinc-300'>
```
thure, line 77: 2rem is 32px, which is one of the density sizes… should this use one of those or remain arbitrary?

`import { fineBlockSize } from '../../styles/density'`

PR #2735, opened 2023-03-14
fix: Creating space on new devices & space properties init

### LINKED ISSUE #2493: Multiple device space creation fails
image

Reproduction:
1) Invite second device
2) Try to create space on second device

dmaretskyi:
Closes #2493


packages/sdk/client/src/packlets/proxies/space-proxy.ts
```diff
@@ -166,6 +172,7 @@ export class SpaceProxy implements Space {
       }
     }
 
+    this._initialized = true;
```
mykola-vrmchk, line 175:
```suggestion
    this._initialized = true;
    this._initializing = false;
```

PR #2753, opened 2023-03-16
fix: dx halo credential

egorgripasov:


packages/devtools/cli/src/base-command.ts
```diff
@@ -80,6 +82,34 @@ export abstract class BaseCommand extends Command {
     this.log('ok');
   }
 
+  async getStdin() {
```
richburdon, line 85: I thins https://nodejs.org/api/readline.html is the best way to read from stdin.
Move to util/readline.ts?

```
import readline from 'readline';

const rl = readline.createInterface({
  input: process.stdin,
  terminal: false
});

rl.on('line', (line) => {
    console.log(line);
});

rl.once('close', () => {
  // Done.
});
```

packages/devtools/cli/src/util/util.ts
```diff
@@ -7,6 +7,7 @@ import { CliUx } from '@oclif/core';
 import { Space, SpaceMember } from '@dxos/client';
 import { truncateKey } from '@dxos/debug';
 import { PublicKey } from '@dxos/keys';
+import { Credential } from '@dxos/protocols/proto/dxos/halo/credentials';
 
 const maybeTruncateKey = (key: PublicKey, truncate = false) => (truncate ? truncateKey(key.toHex(), 8) : key.toHex());
```
richburdon, line 12: truncateKey just takes a key

packages/devtools/cli/src/util/util.ts
```diff
@@ -84,3 +85,35 @@ export const printMembers = (members: SpaceMember[], flags = {}) => {
     }
   );
 };
+
+export const mapCredentials = (credentials: Credential[], truncateKeys = false) => {
```
richburdon, line 89: Refactor util file to co-locate all "print" related methods (always factor out "util" as it grows)

packages/devtools/cli/src/util/index.ts
```diff
@@ -2,6 +2,7 @@
 // Copyright 2022 DXOS.org
 //
 
+export * from './credential';
```
richburdon, line 5: sort imports

packages/devtools/cli/src/util/credential.ts
```diff
@@ -0,0 +1,38 @@
...
+        result = credentials.filter(predicate);
+      } else {
+        result = credentials;
+      }
+    },
+    onError: (err) => {
+      throw err;
+    }
+  });
+  setTimeout(() => {
```
richburdon, line 31: Why is this required?

richburdon: @wittjosiah should this be halo.credentials.get?

richburdon (changes requested):

packages/devtools/cli/src/util/credential.ts
egorgripasov: this is required because we are reading creds from feed, and they are coming progressively (not all at once).
e.g. if you have 5 creds in your halo, you will get 5 `onUpdate` events:
1 - with 1 cred
2 - with 2 creds
...
5 - with all 5 creds.
Since we don't really know how many credentials will be there, we could only wait for some time till we think all of them are returned

packages/devtools/cli/src/util/util.ts
egorgripasov: +1

packages/devtools/cli/src/util/credential.ts
mykola-vrmchk: We are working on this with @dmaretskyi, so for now timeout is ok

PR #2760, opened 2023-03-17
feat: Add options to fields in EchoSchema

mykola-vrmchk:


mykola-vrmchk:
* main
  * **PR #2760** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2760" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2760?utm_source=stack-comment).

packages/core/echo/echo-typegen/test/proto/test.proto
```diff
@@ -23,7 +23,7 @@ message Contact {
 
   option (object) = true;
 
-  string name = 1;
+  string name = 1 [primary = true, required = true];
```
dmaretskyi, line 26: default="Anonymous" is a good example.

I don't want primary to be confused with primary keys in sql.

Cc @wittjosiah: can we follow the same default= pattern for text?

dmaretskyi (approved):

PR #2789, opened 2023-03-20
feat: Rename Document to TypedObject (and Expando)

richburdon:
Notes on docs

- [ ] Need to mention `Expando` (some snippets now include it).
- [ ] `snippets/schema.ts` is out of date (e.g., missing TaskProps type def need to regen)? (Where is the associated proto def?)
- [ ] Examples should use `example` as the namespace (not `dxos`).


packages/apps/quickstart/src/Counter.tsx
```diff
@@ -13,13 +13,14 @@ export const Counter = () => {
   const [counter] = useQuery(space, { type: 'counter' });
   useEffect(() => {
     if (!counter && space) {
-      const c = new Document({ type: 'counter' });
-      void space.db.add(c);
+      const counter = new Expando({ type: 'counter' });
+      void space.db.add(counter);
```
dmaretskyi, line 17:
```suggestion
      space.db.add(counter);
```

dmaretskyi (approved):

PR #2815, opened 2023-03-23
fix(client-services): break infinite loop when starting with resource lock

wittjosiah:
ios env is testing compatibility mode, android is hitting this https://github.com/dxos/dxos/issues/2821


wittjosiah:
* main
  * **PR #2815** <a href="https://app.graphite.dev/github/pr/dxos/dxos/2815" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/2815?utm_source=stack-comment).

PR #2829, opened 2023-03-24
feat: Don't block UI on pending spaces

dmaretskyi:


packages/apps/composer-app/src/layouts/DocumentLayout/DocumentLayout.tsx
```diff
@@ -19,8 +19,8 @@ export const DocumentLayout = () => {
   const shadow = useButtonShadow('base');
 
   const { spaceKey } = useParams();
-  const spaces = useSpaces();
-  const space = spaces.find((space) => abbreviateKey(space.key) === spaceKey);
+  const spaces = useSpaces({ all: true });
```
zhenyasav, line 22:
```suggestion
  const spaces = useSpaces({ includePartialSpaces: true });
```

zhenyasav: or something more intuitive than `all`

packages/core/mesh/messaging/src/messenger.test.ts
```diff
@@ -313,4 +317,39 @@ describe('Messenger', () => {
       expect(count).toEqual(1);
     });
   });
+
+  describe.skip('load', () => {
```
zhenyasav, line 321: skip?

packages/core/mesh/messaging/src/messenger.test.ts
```diff
@@ -313,4 +317,39 @@ describe('Messenger', () => {
...
+    test('many connections to KUBE', async () => {
```
zhenyasav, line 322: seems unrelated?

zhenyasav (changes requested):

packages/apps/composer-app/src/layouts/DocumentLayout/DocumentLayout.tsx
dmaretskyi: 👍

packages/core/mesh/messaging/src/messenger.test.ts
dmaretskyi: just wanted to keep the code here in case we need load-test signal

dmaretskyi: not really a test that should be running

PR #2879, opened 2023-03-30
refactor: Shell test manager

thure:


packages/apps/patterns/react-ui/src/testing/shell-manager.ts
```diff
@@ -0,0 +1,59 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import type { Locator, Page, FrameLocator } from 'playwright';
+
+type Scope = Locator | FrameLocator | Page;
+
+export class ShellManager {
+  page!: Page;
```
wittjosiah, line 10: I feel like there's probably a better way to include this to ensure that anything that extends it needs to provide the page...but I can't think of how that would look right now. Maybe leave a todo?

wittjosiah (approved):

PR #2898, opened 2023-03-31
fix: sync protos between dxos and kube repos

egorgripasov:
Before we could do a simple copy between repos, let's make sure all the protos are identical


packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -123,7 +130,8 @@ message Runtime {
     optional string public_url = 4 [(env) = "PUBLIC_URL"];
 
     optional BuildInfo build = 100;
-    optional google.protobuf.Struct env = 101;
+    // TODO: fix - is clashing with field option "env"
+    // optional google.protobuf.Struct env = 101;
```
egorgripasov, line 134: @dmaretskyi @wittjosiah I would need you help here - it seems like option `env` clashes with field `env`, e.g. go proto compiler is not happy with
```
Option field "(env)" is not a field or extension of message "FieldOptions".
```
Can we figure out naming there?

packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -6,9 +6,16 @@ syntax = "proto3";
...
 import "google/protobuf/any.proto";
 import "google/protobuf/struct.proto";
+import "google/protobuf/descriptor.proto";
 
 package dxos.config;
 
+option go_package = "github.com/dxos/kube/proto/def/dxos/config";
+
+extend google.protobuf.FieldOptions {
+    optional string env = 65000;
```
richburdon, line 16: indent seems off. 
start extensions at `1000`

packages/core/protocols/src/proto/dxos/config.proto
richburdon: for now i think you can rename the field extension to `env_var` and update its usages (i think only a few) since that would be non-destructive.

richburdon (approved):

packages/core/protocols/src/proto/dxos/config.proto
dmaretskyi: if rename fixes the error - lets go for it. the error message seems a bit wierd, it misparses the env field for an option?

egorgripasov: It seems so - renaming does fix it;
Updated option name;
@dmaretskyi do we actually use it somehow, or it's mainly a declarative thing? seems like throughout the code we are reading those values directly/explicitly from `process.env`

PR #2915, opened 2023-04-03
feat: Add web file storage

mykola-vrmchk:


packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -0,0 +1,173 @@
...
+  protected _root?: FileSystemDirectoryHandle;
+
+  constructor(private readonly path: string) {}
+
+  public get size() {
+    return this._files.size;
+  }
+
+  private _getFiles(path: string): Map<string, File> {
+    const fullPath = getFullPath(this.path, path);
```
dmaretskyi, line 31: is this for directory.list method? this doesn't seem like it would work corretly. this would only show previously opened files, not all of the files on disk. there's a browser api to enum a directory, lets use that instead

packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (offset + size > file.size) {
+      throw new Error('Read out of bounds');
+    }
+    return Buffer.from(new Uint8Array(await file.slice(offset, offset + size).arrayBuffer()));
+  }
+
+  async del(offset: number, size: number) {
+    const fileHandle: any = await this._fileHandle;
+    const writable = await fileHandle.createWritable({ keepExistingData: true });
+    await writable.write({ type: 'truncate', size: offset });
```
dmaretskyi, line 149: wont work correctly if offset + size < length

dmaretskyi (changes requested):
comments

packages/common/random-access-storage/src/browser/web-fs.ts
mykola-vrmchk: That is a bug (or expected behavior) in all our implementations for random-access-storage. I added a test that fails for all implementations and TODO in our issue, so it is weird to fix it only for WebFS, let's do it in separate PR for all implementations

PR #3036, opened 2023-04-14
fix: Fix webFS reset

mykola-vrmchk:
### Summary
🐛🚨🚧

Improved the browser-based `WebFS` random-access storage implementation and added a test case for the `reset` method. Fixed a bug in `WebFS.reset` and changed the error handling of `File.destroy`.

> _Sing, O Muse, of the web-born storage interface_
> _That lets the mortal browsers access files at random_
> _And of the skillful coder who fixed its errors and flaws_
> _And made it more consistent and robust for all._

### Walkthrough
*  Modify error handling of `destroy` method of `File` class to avoid rethrowing errors when deleting files ([link](https://github.com/dxos/dxos/pull/3036/files?diff=unified&w=0#diff-ccda6ccab8c2365217eb1ce5fb59fee83beca7a5b48b18eeec7d5f189b25abb4L82-R82)).

### DIFF
--- a/packages/common/random-access-storage/src/browser/web-fs.ts
+++ b/packages/common/random-access-storage/src/browser/web-fs.ts
@@ -79,14 +79,19 @@ export class WebFS implements Storage {
   private async _delete(path: string): Promise<void> {
     await Promise.all(
       Array.from(this._getFiles(path)).map(async ([path, file]) => {
-        await file.destroy().catch((err: any) => log.catch(err));
+        await file.destroy().catch((err: any) => log.warn(err));
         this._files.delete(path);
       })
     );
   }
 
   async reset() {
-    await this._delete('');
+    await this._initialize();
+    for await (const filename of await (this._root as any).keys()) {
+      assert(typeof filename === 'string');
+      this._files.delete(filename);
+      await this._root!.removeEntry(filename, { recursive: true }).catch((err: any) => log.warn(err));
+    }
     this._root = undefined;
   }
 
--- a/packages/common/random-access-storage/src/browser/web-fs.ts
+++ b/packages/common/random-access-storage/src/browser/web-fs.ts
@@ -88,7 +88,6 @@ export class WebFS implements Storage {
   async reset() {
     await this._initialize();
     for await (const filename of await (this._root as any).keys()) {
-      assert(typeof filename === 'string');
       this._files.delete(filename);
       await this._root!.removeEntry(filename, { recursive: true }).catch((err: any) => log.warn(err));
     }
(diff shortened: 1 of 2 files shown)


mykola-vrmchk:
* main
  * **PR #3036** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3036" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3036?utm_source=stack-comment).

packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -79,14 +79,19 @@ export class WebFS implements Storage {
...
+        await file.destroy().catch((err: any) => log.warn(err));
         this._files.delete(path);
       })
     );
   }
 
   async reset() {
-    await this._delete('');
+    await this._initialize();
+    for await (const filename of await (this._root as any).keys()) {
```
wittjosiah, line 90: types missing?

packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -79,14 +79,19 @@ export class WebFS implements Storage {
...
+      assert(typeof filename === 'string');
```
wittjosiah, line 91: in practice is this ever not a string? currently if this throws the reset will stop I think, may want to just do a conditional check or try/catch?

packages/common/random-access-storage/src/testing/storage.blueprint-test.ts
```diff
@@ -246,6 +247,47 @@ export const storageTests = (testGroupName: StorageType, createStorage: () => St
       }
     });
 
+    test('reset', async () => {
+      if (
+        testGroupName === StorageType.RAM || // RAM storage does not persist data.
+        testGroupName === StorageType.IDB // IDB storage is blocked by opened connection, and there is no handle to close it.
```
wittjosiah, line 253: it's not possible to reset idb without initializing?

packages/common/random-access-storage/src/browser/web-fs.ts
mykola-vrmchk: yep

packages/common/random-access-storage/src/browser/web-fs.ts
mykola-vrmchk: Make sense. It is always a string, but as you saw types are missing, so the `filename` type is `unknown`

packages/common/random-access-storage/src/testing/storage.blueprint-test.ts
mykola-vrmchk: It is only possible to reset IDB before initialization because the library we are using for IDBStorage doesn't properly dispose connection to IDB and that is blocking reset after initialization

PR #3056, opened 2023-04-18
feat: Add general TestBuilder pattern

mykola-vrmchk:
copilot:all


mykola-vrmchk:
* main
  * **PR #3056** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3056" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3056?utm_source=stack-comment).

packages/common/test-builder/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "types": "dist/types/src/index.d.ts",
+  "files": [
+    "dist",
+    "src"
+  ],
+  "dependencies": {
+    "@dxos/node-std": "workspace:*"
+  },
+  "publishConfig": {
+    "access": "public"
```
richburdon, line 22: always private by default so we don't pollute npm

packages/common/test-builder/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@dxos/test-builder",
+  "version": "0.1.40",
+  "description": "This package defines general purpose test builder ",
```
richburdon, line 4: "this package defines..." not necessary :) "Testing framework for ECHO"

packages/common/test-builder/src/builder.ts
```diff
@@ -0,0 +1,29 @@
...
+//
+
+import { Peer } from './peer';
+
+type CreatePeerOpts<T extends Peer> = {
+  factory: () => T;
+};
+
+export class Builder {
+  private readonly _peers = new Set<Peer>();
```
richburdon, line 12: map by key?

packages/common/test-builder/src/builder.ts
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright 2022 DXOS.org
+//
+
+import { Peer } from './peer';
+
+type CreatePeerOpts<T extends Peer> = {
```
richburdon, line 7: export since part of api

packages/common/test-builder/src/builder.ts
```diff
@@ -0,0 +1,29 @@
...
+  factory: () => T;
+};
+
+export class Builder {
+  private readonly _peers = new Set<Peer>();
+
+  createPeer<T extends Peer>(opts: CreatePeerOpts<T>): T {
```
richburdon, line 14: There isn't really enough context in this peer to assess if this is right -- there should always be minimally a test use case.

packages/common/test-builder/package.json
mykola-vrmchk: That is not just for ECHO, it is base TestBuilder and  everybody should use it as base class for their builders

packages/common/test-builder/src/builder.ts
mykola-vrmchk: Maybe

packages/common/test-builder/src/builder.ts
mykola-vrmchk: > Also how does this relate to the existing dxos/client/testing builders?

Those builders should `extend` this one

PR #3067, opened 2023-04-19
docs: composer spec

zhenyasav:
### Summary
📝🎼🌐

Added a new document `docs/docs/specs/v0/composer.md` that explains the design and architecture of Composer, a peer-to-peer knowledge management system for DXOS. The document aims to provide a clear and comprehensive overview of Composer's features, scenarios, and extensibility.

> _`Composer` design_
> _A peer-to-peer knowledge base_
> _Autumn of ideas_

### Walkthrough
*  Add a new document `composer.md` that describes the design and architecture of Composer, a peer-to-peer knowledge management system ([link](https://github.com/dxos/dxos/pull/3067/files?diff=unified&w=0#diff-c2e7a79caf635172b918b576878c6e1099e7f2ce5d02f89257ae4b504a799e13R1-R86))

### DIFF
--- a/docs/docs/specs/v0/composer-2.md
+++ b/docs/docs/specs/v0/composer-2.md
@@ -3,6 +3,7 @@
 A developer-first, peer-to-peer knowledge management system (KMS).
 
 ## Iteration history
+
 - `zhenya-0` proposed a static (v1) vs dynamic (v2) chrome paradigm.
 - `rich-0` proposed Surfaces, ECHO-bound Frames and embraced the dynamic chrome paradigm, getting ahead of `zhenya-0` in terms of flexibility.
 - `zhenya-1` (this version) embraces Surfaces, is ahead of `rich-0` in terms of flexibility, but behind `rich-0` in terms of dynamic discovery, loading, installation, and isolation of modular code. This version also decouples from ECHO and replaces `Frames` with regular `React.Component`.
@@ -81,7 +82,6 @@ Below is a diagram illustrating how:
 
 ![Composer Structural Diagram](./composer.drawio.svg)
 
-
 ## Stacks
 
 - Stacks are just components which rely on the same `<Surface />` concept to delegate the concrete rendering of stack elements to components provided by plugins.
@@ -120,16 +120,16 @@ Below is an example of how the App in the diagram above would be instantiated il
 <App
   plugins={[
     new SplitViewPlugin(), // provides SplitView with sidebar, main, tools surfaces
-    new TreePlugin(),      // provides Tree and Selection which visualize the graph
-    new SpacesPlugin(),    // provides graph nodes that represent ECHO spaces and their contents
-    new StackPlugin(),     // provides Stack
-    new ComposerPlugin(),  // provides Composer
-    new ImagePlugin(),     // provides Image
-    new HaloPlugin(),      // provides the HALO button and shell in the tools surface
-    new SearchPlugin()     // provides the Search element in the tools surface
+    new TreePlugin(), // provides Tree and Selection which visualize the graph
+    new SpacesPlugin(), // provides graph nodes that represent ECHO spaces and their contents
+    new StackPlugin(), // provides Stack
+    new ComposerPlugin(), // provides Composer
+    new ImagePlugin(), // provides Image
+    new HaloPlugin(), // provides the HALO button and shell in the tools surface
+    new SearchPlugin() // provides the Search element in the tools surface
   ]}
   state={{
-    // initial state with which the application state context will be created 
+    // initial state with which the application state context will be created
     // (or amended during a re-render)
     // e.g.: parsed route and params information from a parent route element goes here
     spaceKey: PublicKey.parse(routeArgs.spaceKey),
@@ -168,11 +168,59 @@ It's possible for the app to enter a mode where the surface-to-component binding
 
 `SearchPlugin` knows how to alter the `surfaces` section of the state such that a different control is presented instead of the `Tree` with flat search results whenever a search term is present in the input box. Alternatively, the `Tree` can equally respond to changes in the search term (global state) and filter itself down accordingly.
 
+## Example Plugins
+
+### Tree Plugin
+
+```tsx
+import { HaloButton } from '@dxos/react-shell';
+import { Tree } from './Tree';
+import { Selection } from './Selection';
+import { Plugin } from '@dxos/app-model';
+
+export const TreePlugin: Plugin = {
+  meta: {
+    id: 'TreePlugin'
+  },
+  provides: {
+    components: {
+      HaloButton,
+      Tree, // uses AppContext.plugins to grab graph nodes from all plugins
+      Selection // uses AppContext.tree.selection to grab ECHO objects in GraphNode.data and <Surface /> them
+    }
+  }
+};
+```
+
+### Spaces Plugin
+
+```tsx
+import { Plugin } from '@dxos/app-model';
+
+export const SpacesPlugin: Plugin = {
+  meta: {
+    id: 'SpacesPlugin'
+  },
+  provides: {
+    graph: {
+      getNodes(parent, context) { // return a GraphNode with label and lazy children
+        const { client } = context;
+        return client.getSpaces().transform((space) => ({
+          label: humanize(space.key),
+          children() { return space.query().transform((object) => ({ /* ... */ })) }
+        })); // transformed observable
+      }
+    }
+  }
+}
+```
+
 ## Separation from ECHO
 
 ### Non-ECHO graph nodes
 
 Graph nodes are not restricted to being ECHO objects. Examples:
+
 - nodes that represent the default surface-to-component bindings and allow users to configure or swap e.g.: the root `SplitView` for something else with the same (or different) surfaces.
 - nodes that represent the filesystem
 - nodes that represent entities in GitHub (which are not also represented in ECHO)
@@ -180,13 +228,14 @@ Graph nodes are not restricted to being ECHO objects. Examples:
 ### Non-Frame components
 
 Some of the components in the example above can be considered `Frames` because they bind directly to ECHO objects. Examples:
-  - Composer
-  - Image
-  - Stack
+
+- Composer
+- Image
+- Stack
 
 Others are not binding directly to ECHO objects, but observe other things, e.g.: only ephemeral elements of `AppState`:
-  - `Search` binds to `AppState.search.term` or similar
-  - `HaloButton` binds to nothing (no explicit data context) and invokes the shell when clicked. 
-  
-n.b.: One could say `HaloButton` binds to a HALO Identity, but that would be the case for an `Avatar` orb component or similar, the `HaloButton` component is parameterless and always calls `useIdentity` internally without requiring data context from the parent, making it trivial to use in any application.
 
+- `Search` binds to `AppState.search.term` or similar
+- `HaloButton` binds to nothing (no explicit data context) and invokes the shell when clicked.
+
+n.b.: One could say `HaloButton` binds to a HALO Identity, but that would be the case for an `Avatar` orb component or similar, the `HaloButton` component is parameterless and always calls `useIdentity` internally without requiring data context from the parent, making it trivial to use in any application.
--- a/docs/docs/specs/v0/composer-2.md
+++ b/docs/docs/specs/v0/composer-2.md
@@ -1,0 +1,180 @@
+# Composer
+
+A developer-first, peer-to-peer knowledge management system (KMS).
+
+## Scenarios
+
+#### Composer scenarios (production)
+
+1. Users can collaborate on a github issue together
+2. Users can work on folders of markdown files in a "stack-editor" paradigm
+3. Developers can extend Composer with custom data sources and visual surfaces
+4. Developers can visualize the structure of Composer's surfaces and examine metadata about fulfilling components.
+
+#### Additional Kai scenarios (labs)
+
+4. Users can work with email
+5. Users can see all notifications in a universal inbox (across email and other sources)
+6. Users can make presentations
+7. Users can draw together
+8. Users can collect lists of contacts
+9. Users can manage shared todo lists
+10. Users can launch and interact with bots in their spaces
+11. Users can play chess together
+12. Users can collaborate over kanban boards
+13. Users can collaborate over post-boards of notes
+14. Users can interact with LLM based agents
+15. Users can chat and message with other users
+
+## Goals
+
+1. To provide developers with full control over the interface of their KMS
+2. To make it possible for Kai and Composer to share a common application framework
+3. To make Composer a easier-to-ship subset of all of Kai's scenarios
+
+## Non goals
+
+1. Isolation of 3p code - all modular code is assumed to be trusted for now, with isolation to be tackled at a later point.
+2. Runtime extensibility - for now it's ok to assume all plugins are compiled-in, and later we'll tackle loading and installing them at runtime via ESM.
+3. Dynamic discovery and installation of modular UI components - although the design should not preclude this later once runtime extensibility is complete.
+4. Central registration in DMG - no dynamic discovery and installation means DMG is not required in this design.
+5. Coupling of any part of the composer model to ECHO - ECHO will be just one of the sources of state for the application, and therefore the application model does not need to and should not marry developers to ECHO as much as possible.
+
+## Definitions
+
+- **Application**: An application built with this composer framework which can use Surfaces to render components "dynamically" and provides a specific context to all nested components.
+- **Component**: A regular react component which can rely on a specific application context while inside the application. Components are free to nest as many Surfaces within them as necessary.
+- **Surface**: A way of delegating the concrete presentation of a piece of UI to external plugins. A surface accepts a data context, a set of plugins, and other props which are used to resolve and instantiate concrete components to 'fulfill' that surface.
+- **Plugin**: A unit of containment, modularization, and distribution of external/modular functionality that can be dynamically installed into an application (at build time, and later at runtime). Plugins provide things like components and other data to the application.
+- **Shell**: The HALO button and generic UI for join and space management flows that appear in popups overtop the main application and hosted in the vault iframe. A subset of the vault UI surface that is restricted to the calling application that saving developers the cost of implementing boilerplate UI flows for space and identity management.
+- **Chrome**: All the UI of an application that is not (but surrounds) the user generated content itself. (e.g.: buttons, toolbars, dividers, drag handles, ... etc.)
+- **Vault**: HALO universal PWA which stores identity and data
+- **Stack**: A user experience paradigm (component) which enables editing columns of mixed content with arbitrary (extensible) content types. A key feature is the draggability of paragraphs or content items to re-order them.
+
+## Non-definitions
+
+- **Frames**: are not defined because explicit coupling to ECHO is a non-goal, and without that Frames are just regular React Components which can assume certain contexts or props within this application framework. React Components or Components are used in place of Frames from prior art.
+
+## Applications are made of Surfaces and Components
+
+- The application is a component `<App />`.
+- A component can have any number of `<Surface />` as children inside it.
+- A Surface figures out what Component(s) to render given a set of `loaded plugins`, a `data context`, and `other control props`. It can be understood as the the developer intention to render/visualize the given `data context` (datum) using whatever appropriate component as provided by Plugins and configured by the the end user and/or the developer.
+- The main `<App />` component defines a single, full-screen `<Surface />` for rendering "anything". i.e.: does not opine about what it is, delegates that to plugins and initial configuration entirely.
+- The main `<App />` component also wraps the surface with a Context provider representing the total and entire state of the application which becomes available through a hook to any nested component.
+
+## Stacks
+
+- Stacks are just components which rely on the same `<Surface />` concept to delegate the concrete rendering of stack elements to components provided by plugins.
+
+## Routing
+
+- Routing (the main router and top level routes) are declared outside (around) the `<App />` component and allow the `App` to pass route information as state in the total `AppState` context for all descendants.
+- Nested components inside `<App />` are equally free to use `<Route />` elements if necessary as well.
+- Nested components are encouraged to perform navigation by submitting `Actions` as opposed to calling `navigate()` directly, although situations like `<Button to="/some/route"/>` should also be supported.
+
+## Application State and Actions
+
+- The App provides a global `AppState` (in the `AppContext`) to all nested components which is composed of some concrete and other dynamic sections to be filled by plugins
+- State is updated via a `dispatch(...actions)` API in the `AppContext` which enables Plugins to subscribe to each other's actions and update state accordingly
+- Plugins (and their provided Components) are free to extend/add more fields to the `AppState` as necessary, when it is desirable to share certain state application-wide.
+- State can be composed of `@preact/signals` (or similar reactive state solution) and can contain any other objects including `ECHO` objects.
+- When state is:
+  - local, ephemeral, per-session: it is managed by signals (and backed by local or session storage as appropriate). i.e. the selection state in a List or Tree is ephemeral and not shared between windows or devices, but may be shared among components and plugins in the application.
+  - peer-to-peer, permanent: it is managed by ECHO objects.
+- `AppState` includes current route and params in convenient form.
+- Actions are handled by functions that resemble `reducers` who should return new state reflecting the action committing, and possibly a list of side-effects which may result in more actions or state updates at a later time. (i.e. performing writes to echo objects in a batch is a side-effect, invoking a github API is a side-effect, etc. whereas the indicator that the app is loading something is a purely new state returned from the reducer)
+
+## Plugins, Components, Actions, and other `provides`
+
+- The application framework consumes a list of plugins (statically compiled-in for now).
+- Plugins can provide Components in response to a specific data context: i.e `plugin.provides.component(data: any): React.FC`
+- Compoents provided by plugins can assume they have access to the `AppContext` context and hooks, and may submit actions via `AppContext.dispatch(...actions)`.
+- Plugins can provide action handlers such as `plugin.handlers.[<action-type>](state, action): { state, effects }` and can react to actions coming from other plugins this way.
+- Plugins may provide any number of other data types and APIs which are specific to certain components which may be present in the Application tree at runtime. If a certain component does not exist, then it will not ask for the relevant `provides` from plugins, and they will therefore be inert. (i.e. Plugins provide a `graph` which can be consumed by any number of UI components including the `Tree` which itself is provided by the `TreePlugin` and may represent the entire and total graph built up by all the plugins loaded).
+
+## Structural Diagram
+
+Below is a diagram illustrating:
+
+- a full application is built up to handle a `Stack` of `Composer` and `Image`
+- `Surfaces` are the only element of abstraction necessary to both compose the entire application chrome and implement the abstract portion of `Stack`.
+
+![Composer Structural Diagram](./composer.drawio.svg)
+
+Below is an example of how this App would be instantiated illustrating how every surface and component comes from a plugin (none are statically defined by the App).
+
+```tsx
+<App
+  plugins={[
+    new SplitViewPlugin(), // provides SplitView with sidebar, main, tools surfaces
+    new TreePlugin(),      // provides Tree and Selection which visualize the graph
+    new SpacesPlugin(),    // provides graph nodes that represent ECHO spaces and their contents
+    new StackPlugin(),     // provides Stack
+    new ComposerPlugin(),  // provides Composer
+    new ImagePlugin(),     // provides Image
+    new HaloPlugin(),      // provides the HALO button and shell in the tools surface
+    new SearchPlugin()     // provides the Search element in the tools surface
+  ]}
+  state={{
+    // initial state with which the application state context will be created 
+    // (or amended during a re-render)
+    // e.g.: parsed route and params information from a parent route element goes here
+    spaceKey: PublicKey.parse(routeArgs.spaceKey),
+    entityId: routeArgs.id,
+    // a special area of state declaring the way surfaces are to be given props
+    surfaces: {
+      // the props to be passed to the main surface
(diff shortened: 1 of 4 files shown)


docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,91 @@
...
+
+### 2. Dynamic chrome (v2)
+
+Most of this will be discovered by experimentation with `static chrome` (v1) plugins to find the wishlists of plugin developers.
+
+The core assumptions of this model can be that the screen receives elements (views) horizontally in a "stack" from left to right. This way, an activity bar stacks before a sidebar, followed by the content area, and possibly other "areas" further to the right if that becomes relevant. This might be the way to enable deep drilldown into complex data.
+
+Views can be informed in general by the state of the views to the left. There is almost no statically defined chrome in Composer, except some paradigm for managing the views themselves (through the HALO button).
+
+The HALO button can become "the only piece of chrome the app needs", and encapsulate a pop-over shell with panels that contain a plugin store, space navigation and access control, profile and presence management, etc. Plugins are discovered, installed, and managed entirely in the shell.
```
wittjosiah, line 89: > a plugin store

The above plugin interface is specific to composer right? It's not necessarily a generic framework that would be applied to other apps in the way that frames, as discrete pieces of ui, might be. Is that accurate? If so, what do you imagine a shell plugin store looking like given that it would have to be generic enough to be applicable to all apps?

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Static chrome**, where we constrain the application chrome to be statically defined and not extensible by plugins. Elements of the chrome may each expose some API surface for plugins to extend, but the overall elements of the chrome are fixed and can be assumed by all plugins without explicit dependency expressions.
+
+**Dynamic chrome**, where the chrome and it's organization is also extensible and now plugins have to express explicitly the elements of chrome they depend on because their presence can no longer be assumed.
+
+These are in addition to another orthogonal dimension of complexity, the way we load plugins:
+
+**Static plugins**, where a `<Composer />` element is given a list of `Plugin[]` directly as props and has to be compiled into the app at build time.
+
+**Dynamic plugins**, where the app knows how to load plugins via ESM `import` calls in the browser at runtime.
```
wittjosiah, line 24: It makes sense to me that composer could provide a plugin interface more specific to it's chrome structure. We would then support those plugins being dynamically loaded/installed at some point. What I'm struggling with right now is reconciling this with the frames concept.

Plugins can include frames they require for rendering, this seems fine, but is still composer-specific. If frames are going to be published independent of composer plugins, then frames need their own system for being loaded/managed in some way that's not specific to composer. If that is true, then are we going to build the same thing again expect for plugins or would plugins somehow use that more generic loading system?

I feel like we want to work backwards from a high-level vision of the dynamic end-state because changes to the goal will likely influence where we start and make it easier/harder to get there.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,91 @@
...
+
+The sidebar presents a tree of nodes which can be populated by plugins.
+
+The content area can be filled by plugins, and can sense the state of the tree view (selection, nodes, etc).
+
+```ts
+
+type MaybePromise<T> = T | Promise<T>;
+
+interface Plugin {
```
wittjosiah, line 50: One surface that I think may be missing here is how are new nodes created? Composer needs to be able to create new objects in the space and plugins probably need to be able to teach it how to create new types of objects

zhenyasav: through actions (which are not yet defined in the spec) - presumably actions might submit new ECHO objects into the space, and presumably the things returned from `getTreeNodes` are somehow observable, so they would fire with new nodes when their internal queries get refreshed as a result of adding ECHO objects.

docs/docs/specs/v0/composer.md
zhenyasav: > Plugins can include frames they require for rendering, this seems fine, but is still composer-specific
yes composer specific is the design intention for the sake of reducing complexity

> If frames are going to be published independent of composer plugins
I don't think we should worry about that for now (seems like a story for sharing frames between more than one app)

zhenyasav: its good to work backwards from a very high level north star, but that doesn't mean we need to pre-engineer for all the complexity of that world today - the opposite. We need to work towards increasing complexity from a simpler place, accepting that we will go through major refactoring and major incompatibility boundaries on the way there. The problem with pre-engineering for more complexity too early is that it still doesn't eliminate these major refactors and compatibility boundaries, (because practice will teach us what's actually important and what's not) - but having more complexity from day 1 means more to worry about during these refactors, and more to throw away.

docs/docs/specs/v0/composer.md
dmaretskyi:
```typescript
type Plugin = {
  manifest: { // Manifest is serializable and could be stored in the space that makes use of the plugin for example.
      id: string
      name: string
      description: string
  },
  runtime: { // or name it`extension` ?
     sidebar: {  // Separating extensibility points into logical groups. Plugins can implement a subset of them.
           // plugins can provide TreeNodes to the Tree
          getTreeNodes(parent: TreeNode): MaybePromise<TreeNode[]>;
          // plugins can provide actions to TreeNodes
          getTreeNodeActions(parent: TreeNode): MaybePromise<Action[]>;
     },
     content: {
         // plugins can provide a UI for the content area
         // content area components access the rest of the UI state via context
         getComponent(selection: TreeNode[]): MaybePromise<React.FC>;
     }
  }
}
```

dmaretskyi: For future we might split `manifest` and `runtime` into separate JS bundles so that manifests could quickly be loaded by web apps without requiring loading code for compoenents

docs/docs/specs/v0/composer.md
wittjosiah: I think my concern was that this seemed to be imagining composer-specific stuff inside the shell

docs/docs/specs/v0/composer.md
zhenyasav: recommend `meta` instead of `manifest`

zhenyasav: recommend `extensions` instead of `runtime`

docs/docs/specs/v0/composer.md
wittjosiah: > yes composer specific is the design intention for the sake of reducing complexity

I didn't mean to imply it being composer-specific was bad, but just that it imposes some level of dependency restrictions (i.e., dxos deps don't know about this, it's a braneframe/composer-level thing)

wittjosiah: Fair enough about not pre-engineering things which might just end up in the trash. I think what makes me uneasy is having this frames concept that we've discussed quite a bit, have the expectation that it will end up under-girding plugins someway in the long run, but not have anything specifying that (even if it's fuzzy/uncertain/likely will change shape).

docs/docs/specs/v0/composer.md
wittjosiah: > recommend `extensions` instead of `runtime`

"modules"? I don't mind "extensions" either, it just feels a little weird to me to nest "extension" and "plugin" because they are synonyms

docs/docs/specs/v0/composer.md
zhenyasav: yes there is a hint about shell itself being extensible for apps to place custom panels in there - but I am not too worried about this yet, there are many ways we could do this

docs/docs/specs/v0/composer.md
zhenyasav: vscode uses something like `provides` here, but fair point

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+
+The HALO button (o) is in one of the corners and can be used to access the DXOS Shell.
+
+The sidebar presents a `tree` of nodes which can be populated by plugins.
+
+The `tree` is a logical first solution to the problem of organization and represents a model users and developers are very familiar with and have grown to expect, having found the same model in almost all nearby and competitive products: Notion, Quip, Dropbox Paper, Google Docs, VSCode, Obsidian, ... etc.
+
+Plugins provide components that can fill the content area fully, one at a time, and can sense the state of the tree view (selection, nodes, etc).
+
+Search happens via an overlay [Cmd+K experience](https://cmdk.paco.me/) where a box pops over all the surfaces and presents a fuzzy-search with autocomplete which can be used to locate objects to navigate to, or actions and commands to invoke.
```
richburdon, line 50: We should debate modal vs non-modal search. 

1). Many (most) frames will implement some form of search/filtering (inbox, calendar, tasks, etc. will filter individual objects; composer/stacks might filter/highlight matching elements within the AST, etc.) In these cases, search/filtering may be a "semi-sticky" state while work is continued. 

2). Having a consistent (and consistently located) search area (e.g., sidebar) obviates re-implementing (and cluttering) this functionality in all of these places.

3). When real estate exists (i.e., other than phone) it would be helpful for search not to occlude the main content.

4). Universal search across spaces should have a high priority.

5). Google-style text "search" will likely soon become obsolete. We should consider a hybrid semi-interactive search/suggest model that is "always on" and always visisble.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+type Effect = (state: ComposerState) => MaybePromise<ComposerState>;
+
+type Action = {
+  id: string;
+  label: string;
+  icon?: React.FC;
+  invoke(state: ComposerState): MaybePromise<Effect | Effect[]>;
+};
+
+type TreeNode<T = any> = {
```
richburdon, line 84: Helpful to think consider layers. E.g., "Action" seems out of context with "Tree" (i.e., horizontal concept vs. vertical concept).

Similarly I suggest removing "Composer" from framework type definitions.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+1. the markdown plugin - which provides a plain text editor for the content area and fills the Tree with document and folder nodes
+2. the filesystem plugin - which provides import / export to folders on disk
+3. the github plugin - which provides nodes representing github issues and assets
+4. the stacks plugin - which provides a stack editor for the content area and import / export actions to github
+
+If a stack with custom tiles (frames) is required, that is just an extension of the stacks plugin, where the `Stack` returned from `getComponent` is endowed with more kinds of frames statically.
+
+#### How Kai relates to this model
+
+In v1 (static chrome) Kai can take advantage of the composer chrome to get ahead on compatibility with mobile screens, take advantage of the magic search box, and reduce boilerplate. Kai can re-implement itself as a specific expression of the `<Composer />` element with custom values for the sidebar content and list of plugins.
```
richburdon, line 135: This is helpful re short-term convergence of Kai/Composer at the low-level component layer (e.g., common shell).
However, as discussed elsewhere, there's a large overlap (but not congruence) between the concepts of Frames explored and implemented in kai and Composer-specific plugins. Given the ambition of what's being laid out here, we should further tease out core/universal "Braneframe" framework concepts from v1/v2 Composer/Kai functionality (i.e., rather than Composer "the app" driving core framework decisions: e.g., trees, search, etc.)

That is to say, Composer can be the "driver" of focus of the framework, but should stand-alone from the framework conceptual organization. Otherwise we may overspecialize the framework.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+
+```tsx
+const Kai = () => (
+  <Composer
+    sidebar={<KaiSidebar />}
+    plugins={[
+      new FramesPlugin(),
```
richburdon, line 142: I can't cohere a mental model that reconciles plugins and frames.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+'------------'
+```
+
+Small screens turn the `sidebar` into a slide-over element that hides by default.
+
+The HALO button (o) is in one of the corners and can be used to access the DXOS Shell.
+
+The sidebar presents a `tree` of nodes which can be populated by plugins.
+
+The `tree` is a logical first solution to the problem of organization and represents a model users and developers are very familiar with and have grown to expect, having found the same model in almost all nearby and competitive products: Notion, Quip, Dropbox Paper, Google Docs, VSCode, Obsidian, ... etc.
```
richburdon, line 46: I agree as long as the framework doesn't force this assumption. I.e., I'd like to flesh out a higherlevel application ontology (partly explained in this PR already) that discusses aspects of organization, navigation (incl. URL routes, breadcrumbs, links), search, suggest, primary content, and content aggregation (frames/stacks).

E.g., for composer: top-level data organization is managed by a (n-level?) tree, where the root nodes are X, and leaves are Y (see other note asking questions about messages, tasks, etc. vs tables, task lists). In this case is composer an example of an app that happens to put the Tree in a sidebar. Or is the sidebar/tree a core concept of the framework. Again, trying to tease out application vs framework.

docs/docs/specs/v0/composer.md
```diff
@@ -0,0 +1,190 @@
...
+
+The `<KaiSidebar />` is equally free to avoid using a `Tree` entirely and can replace that content with any form of accordion or stacked views desired. This sidebar can use a context hook like `useComposerState` to get access to the current UI state which includes the list of loaded plugins and their APIs.
+
+Kai plugins are free to extend the base `Composer.Plugin` while at the same time expanding their behavior with things Composer can't do.
+
+In this example, Kai ignores the `tree` extensibility surface, and prescribes a different one for rendering a flat list of content items below the list of frames in the sidebar.
+
+```tsx
+interface KaiPlugin extends Plugin {
+  provides: Omit<Plugin['provides'], 'tree'> & {
```
zhenyasav, line 170:
```suggestion
  provides: Optional<Plugin['provides'], 'tree'> & {
```

docs/docs/specs/v0/composer.md
zhenyasav: addd a non goals section for debate

docs/docs/specs/v0/composer.md
zhenyasav: good discussion point.
in the meantime, I renamed `ComposerState` with `AppState`, illustrated Frames in a glossary, as well as outlined key differences between Kai frames and Composer extensibility at the bottom

docs/docs/specs/v0/composer.md
zhenyasav: let's discuss.
See description of how Kai implements frames just below https://github.com/dxos/dxos/pull/3067/files#diff-c2e7a79caf635172b918b576878c6e1099e7f2ce5d02f89257ae4b504a799e13R166

docs/docs/specs/v0/composer.md
zhenyasav: `ComposerState` renamed `AppState`

re: `Action` let's discuss, it's unclear the meaning of horizontal and vertical here.

docs/docs/specs/v0/composer.md
zhenyasav: replaced the cmd+k search with universal search description and mechanism in `AppState`

docs/docs/specs/v0/composer.md
zhenyasav: good discussion to have here.

I'm trying to make the position that the framework should not do certain things too abstractly too early, and should embrace a step function rewrite between v1 and v2. In v1 it's therefore correct to statically prescribe `tree` and `content`, and the big value step in v2 will be the unlocking of this to be more abstract. Lots of mileage is possible without the whole 9 yards here.

Added some more description about where content items go https://github.com/dxos/dxos/pull/3067/files#diff-c2e7a79caf635172b918b576878c6e1099e7f2ce5d02f89257ae4b504a799e13R168

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Stacks
+
+- Stacks are just components which rely on the same `<Surface />` concept to delegate the concrete rendering of stack elements to components provided by plugins.
+
+## Routing
+
+- Routing (the main router and top level routes) are declared outside (around) the `<App />` component and allow the `App` to pass route information as state in the total `AppState` context for all descendants.
+- Nested components inside `<App />` are equally free to use `<Route />` elements if necessary as well.
+- Nested components are encouraged to perform navigation by submitting `Actions` as opposed to calling `navigate()` directly, although situations like `<Button to="/some/route"/>` should also be supported.
```
thure, line 93: I don't think the interface for the app’s route should be distinct from any other part of the `AppState` – maybe that is already what this section intends to communicate, but I'd phrase some things this way:

> components are welcome to use subcomponents like `<Route … />` inasmuch as `Route` is like any other subcomponent that selects for specific `AppState` conditions, only focused on the `AppState`’s route substate.

> components can “navigate” by dispatching actions that affect `AppState`’s route substate.

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+- Plugins (and their provided Components) are free to extend/add more fields to the `AppState` as necessary, when it is desirable to share certain state application-wide.
+- State can be composed of `@preact/signals` (or similar reactive state solution) and can contain any other objects including `ECHO` objects.
+- When state is:
+  - local, ephemeral, per-session: it is managed by signals (and backed by local or session storage as appropriate). i.e. the selection state in a List or Tree is ephemeral and not shared between windows or devices, but may be shared among components and plugins in the application.
+  - peer-to-peer, permanent: it is managed by ECHO objects.
+- `AppState` includes current route and params in convenient form.
+- Actions are handled by functions that resemble `reducers` who should return new state reflecting the action committing, and possibly a list of side-effects which may result in more actions or state updates at a later time. (i.e. performing writes to echo objects in a batch is a side-effect, invoking a github API is a side-effect, etc. whereas the indicator that the app is loading something is a purely new state returned from the reducer)
```
thure, line 105: A component can have a reducer if it likes but I don’t think we should impose that pattern on strictly all components. If `AppState` is a set of signals, it’s sufficient for a component to subscribe to changes on a certain signal and subsequently dispatch changes to signals as it likes, i.e. all actions would just update signals.

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Plugins, Components, Actions, and other `provides`
+
+- The application framework consumes a list of plugins (statically compiled-in for now).
+- Plugins can provide Components in response to a specific data context: i.e `plugin.provides.component(data: any): React.FC`
+- Compoents provided by plugins can assume they have access to the `AppContext` context and hooks, and may submit actions via `AppContext.dispatch(...actions)`.
+- Plugins can provide action handlers such as `plugin.handlers.[<action-type>](state, action): { state, effects }` and can react to actions coming from other plugins this way.
```
thure, line 112: I don't think plugins would/should need to submit handlers to the engine if the state is already available as signals.

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+- When state is:
+  - local, ephemeral, per-session: it is managed by signals (and backed by local or session storage as appropriate). i.e. the selection state in a List or Tree is ephemeral and not shared between windows or devices, but may be shared among components and plugins in the application.
+  - peer-to-peer, permanent: it is managed by ECHO objects.
+- `AppState` includes current route and params in convenient form.
+- Actions are handled by functions that resemble `reducers` who should return new state reflecting the action committing, and possibly a list of side-effects which may result in more actions or state updates at a later time. (i.e. performing writes to echo objects in a batch is a side-effect, invoking a github API is a side-effect, etc. whereas the indicator that the app is loading something is a purely new state returned from the reducer)
+
+## Plugins, Components, Actions, and other `provides`
+
+- The application framework consumes a list of plugins (statically compiled-in for now).
+- Plugins can provide Components in response to a specific data context: i.e `plugin.provides.component(data: any): React.FC`
```
thure, line 110: I think this needs to be a predicate the engine can use to determine if a component can be rendered:
```tsx
plugin.provides.component(data: any): ReactNode | null
```

Similar to Fluent Blocks’ `renderIf…(data:any)`

thure: Also, the predicate signature probably needs at least two args, either nested in `data` or what have you: the entity the engine is looking for a plugin to render, and maybe the global state as well.

docs/docs/specs/v0/composer-2.md
thure: …unless by this you mean to say plugins can register middleware for `AppState`…?

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+## Application State and Actions
+
+- The App provides a global `AppState` (in the `AppContext`) to all nested components which is composed of some concrete and other dynamic sections to be filled by plugins
+- State is updated via a `dispatch(...actions)` API in the `AppContext` which enables Plugins to subscribe to each other's actions and update state accordingly
+- Plugins (and their provided Components) are free to extend/add more fields to the `AppState` as necessary, when it is desirable to share certain state application-wide.
+- State can be composed of `@preact/signals` (or similar reactive state solution) and can contain any other objects including `ECHO` objects.
+- When state is:
+  - local, ephemeral, per-session: it is managed by signals (and backed by local or session storage as appropriate). i.e. the selection state in a List or Tree is ephemeral and not shared between windows or devices, but may be shared among components and plugins in the application.
+  - peer-to-peer, permanent: it is managed by ECHO objects.
+- `AppState` includes current route and params in convenient form.
```
wittjosiah, line 104: presumably things like these would not be prescribed by are set by plugins? i.e. react-router aware plugin stores location in app state and ECHO aware plugins store ECHO objects in app state as needed?

docs/docs/specs/v0/composer-2.md
wittjosiah, line 105: For clarity, I think since ECHO writes are optimistic locally the side-effect would be handled internally by ECHO?

docs/docs/specs/v0/composer-2.md
wittjosiah, line 110: (perhaps out of scope) do you think it would be worthwhile for the type of `data` be `unknown` and have the component validate the data shape before consuming it?

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+- Actions are handled by functions that resemble `reducers` who should return new state reflecting the action committing, and possibly a list of side-effects which may result in more actions or state updates at a later time. (i.e. performing writes to echo objects in a batch is a side-effect, invoking a github API is a side-effect, etc. whereas the indicator that the app is loading something is a purely new state returned from the reducer)
+
+## Plugins, Components, Actions, and other `provides`
+
+- The application framework consumes a list of plugins (statically compiled-in for now).
+- Plugins can provide Components in response to a specific data context: i.e `plugin.provides.component(data: any): React.FC`
+- Compoents provided by plugins can assume they have access to the `AppContext` context and hooks, and may submit actions via `AppContext.dispatch(...actions)`.
```
wittjosiah, line 111:
```suggestion
- Components provided by plugins can assume they have access to the `AppContext` context and hooks, and may submit actions via `AppContext.dispatch(...actions)`.
```

docs/docs/specs/v0/composer-2.md
wittjosiah, line 112: The available plugins is also available via the `AppContext`?

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+        new SplitViewPlugin(), // provides SplitView with sidebar, main, tools surfaces
+        new TreePlugin(), // provides Tree and Selection which visualize the graph
+        new SpacesPlugin(), // provides graph nodes that represent ECHO spaces and their contents
+        new StackPlugin(), // provides Stack
+        new ComposerPlugin(), // provides Composer
+        new ImagePlugin(), // provides Image
+        new HaloPlugin(), // provides the HALO button and shell in the tools surface
+        new SearchPlugin() // provides the Search element in the tools surface
+      ]}
+      initialState={{
```
wittjosiah, line 139: What's the intention of making this a separate prop, to limit re-renders?

docs/docs/specs/v0/composer-2.md
```diff
@@ -0,0 +1,389 @@
...
+import { Selection } from './Selection';
+import { Plugin } from '@dxos/app-model';
+
+export const TreePlugin: Plugin = {
+  meta: {
+    id: 'TreePlugin'
+  },
+  provides: {
+    components: {
+      HaloButton,
```
wittjosiah, line 213: Shouldn't be here, or at least inconsistent with above `HaloPlugin.HaloButton`

docs/docs/specs/v0/composer-2.md
zhenyasav: consider for example a plugin which prefixes all navigate requests (or rewrites the urls) during action invocation (maybe a bad example) - but you could intercept requests for navigation from say buttons or hyperlinks in deep surfaces, and alter their destinations to be routed to a local version of halo (if the link is pointing to a halo url for example)

zhenyasav: yeah with ECHO that option is there, but this mechanism makes sense in general for all kinds of async processes like a fetch or a mutation in some slow store.

docs/docs/specs/v0/composer-2.md
zhenyasav: exactly, perhaps the interface needs more massage for sure, but if you squint at what you wrote above, that signature is remarkably like a component (FC) just that you called the props `data`. ;) Basically yes I agree, we just have to rationalize with the idea that FCs are just regular functions that can already conditionally return ReactNode, except that since they can also have side effects, perhaps we should wrap them in one more level of function, such as: 
```tsx
plugin.provides.component(data: any) => (props: any) => ReactNode
```

zhenyasav: perhaps all the components are expected to take a specific props type such as
```
type ComponentProps<TData> = {
  data: TData;
  surfaceContext: any
  ...
}
```
but also, such components are welcome to just use hooks like `useAppState` and/or `useSurfaceContext` to figure out where they are etc. all good options

zhenyasav: yes exactly I expect components are likely validating the shape and returning null or nothing when they feel the data isn't applicable, we'll likely employ some patterns for validation such as zod for example

docs/docs/specs/v0/composer-2.md
zhenyasav: thats right, having plugins affect state directly precludes any kind of interception or side-listening. The other big reason to think in terms of actions is the idea that LLMs will have an easier time controlling the UI if we go far enough to describe the effects of actions and try to keep them close enough to real user intents

zhenyasav: correct i think `useAppContext` is the way to get the list of all plugins

docs/docs/specs/v0/composer-2.md
zhenyasav: in this model the route information is assigned to the state every render via props, and to allow actions to alter the mapping of components-to-surfaces (Say a search term wats to swap tree view for a flat list result view) we can't be resetting/nuking the app state with all the `surfaces:` stuff every render. So thats why its split into `apply this to the state only once to init` and `apply this state to the state every time we render`

docs/docs/specs/v0/composer-2.md
zhenyasav: hm? unclear why. which above?

docs/docs/specs/v0/composer-2.md
zhenyasav: > i.e. all actions would just update signals.
some actions may not update signals and/or may encode other side effects

docs/docs/specs/v0/composer-2.md
wittjosiah: `HaloButton` comes from the `HaloPlugin` above https://github.com/dxos/dxos/pull/3067/files/fb24b17a05d91052343d480ed7aa6d3651b12a4f#diff-66de7abfdcafaeb017540570393b52e0f5dd07846750146869ab585ebd0fd2c2R148 whereas here it's being defined in the `TreePlugin`

docs/docs/specs/v0/composer-2.md
wittjosiah: So wouldn't `unknown` be more accurate and type-safe in this case then? Of course, wouldn't stop anyone else from using `any` or not even using TS but at least points to a suggested validation path

docs/docs/specs/v0/composer-2.md
thure: Mentioning here ftr that we had a conversation on Discord getting into greater detail on a model for how middleware would work.

docs/docs/specs/v0/composer-2.md
thure: > for example a plugin which prefixes all navigate requests

Navigation is mutating the route, a substate of `AppState`, so this is still a global state mutation

> may not update signals and/or may encode other side effects

I can see the need for that, but that may be an action that the extensibility engine would not know what to do with and expect a plugin to handle. I suppose what I meant by “all actions would update signals”, I meant “the extensibility engine only responds to actions which update signals”, since any other behavior is probably out of scope for the framework and within-scope for the plugins.

docs/docs/specs/v0/composer-2.md
thure: > components are welcome to just use hooks

A hook might not work well for a synchronous predicate, which I imagine the engine will need in order to quickly find the component of best fit for a surface-datum-appstate… the component should use the hooks to stay up-to-date, but otherwise I think it needs to resolve in something-or-null quickly for good UX.

thure: Not only that, the engine will probably need a way of re-evaluating which components get to occupy which surfaces… in React, this probably means memoizing on the set of dependencies for the range of possible components’ predicates… something to think about.

PR #3148, opened 2023-05-08
feat: surfaces experiment

zhenyasav:
### Summary
📝🆕🔧

This pull request refactors some types, adds some documentation, and introduces a new way of handling actions in the `Surface` framework. It improves the readability, consistency, and expressiveness of the code and the design of the framework. It also documents some of the developer feedback and questions that are relevant for the framework development.

> _We're working on the Surface framework, me hearties_
> _We're adding notes and types and actions, aye_
> _We're refactoring `T` to `TState`, me lads and lasses_
> _We're heaving on the code on the count of three, hooray_

### Walkthrough
*  Introduce a new way of defining and handling actions in the Surface framework using strongly typed actions and handlers ([link](https://github.com/dxos/dxos/pull/3148/files?diff=unified&w=0#diff-99d67a75f05fbf7ecc5922d2db58305b5bcd561040813b8b09f32834147a8102R1-R42))
* Rename the generic type parameter `T` to `TState` in the `App.tsx` file to make it more descriptive and consistent ([link](https://github.com/dxos/dxos/pull/3148/files?diff=unified&w=0#diff-29a0b81d5dc1cc789146e7854827a89a3f2a9f1d6706b38e741a96f04dfbb067L17-R38), [link](https://github.com/dxos/dxos/pull/3148/files?diff=unified&w=0#diff-29a0b81d5dc1cc789146e7854827a89a3f2a9f1d6706b38e741a96f04dfbb067L44-R45))
* Add some notes to the `README.md` file from Zhenya about the design choices and questions of the Surface framework ([link](https://github.com/dxos/dxos/pull/3148/files?diff=unified&w=0#diff-48c510d98912c40a55e794ac2389ed6c2b66f1e4356148bb5c624f8bd20c903aR2-R8))

PR #3228, opened 2023-05-23
feat: observable object

wittjosiah:
### Summary
📄🚀⚛️

This pull request adds a new package named `observable-object` to the `common` folder of the repository. The package provides a way to create and observe JavaScript objects that can be used for reactive programming. The package also includes a `react` subfolder that exports a higher-order component for integrating observable objects with React components. The pull request adds the source files, tests, license, readme, and configuration files for the package.

> _`observable-object`_
> _A new package for reactivity_
> _With license and docs_

### Walkthrough
*  Add a new package named `observable-object` that provides JavaScript objects that can be observed for reactivity ([link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-5e3f2a0e23249232ded9002ff6e558246daaa5f1cf8ba6dd124dbbe386c07199L1-R7), [link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-fd5dbe48a14c499a3936b9d4d99968aaa557b544ee4d4acb431637d1ec6b97d7R1-R25), [link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-91de0e7bd1c103b496bb9fafbba9c13ff37a640eee5862e4d46c081c64d5bbfeR1-R64), [link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-d8df36e9bfff5652a1d7d13600f72d67be3f3b5643f37a57b09921b0c935b220R1), [link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-518db121e84f3470c1b274a184be0c804d03120c49227d0ac3b400be5584670aR1-R6), [link](https://github.com/dxos/dxos/pull/3228/files?diff=unified&w=0#diff-a10bae7ef585dfe9b1a9e7691b172e430c237ce6aa5ec4e9e6d6e7dca94bf519R1-R28)).


wittjosiah:
* main
  * **PR #3148** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3148" target="_blank">Graphite</a> 
    * **PR #3228** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3228" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3228?utm_source=stack-comment).

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,127 @@
...
+    acc.add(obj[subscribe](onUpdate));
+
+    return acc;
+  }, new EventSubscriptions());
+
+  let firstUpdate = true;
+
+  const handle = {
+    update: (selection: Selection) => {
+      const newIds = new Set(getIdsFromSelection(selection));
```
dmaretskyi, line 75: I think we should stop relying on ids here, just push objects into the set and rely on referential equality

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Subscribe to database updates.
+ * Calls the callback when any object from the selection changes.
+ * Calls the callback when the selection changes.
+ * Always calls the callback on the first `selection.update` call.
+ */
+// TODO(burdon): Add filter?
+// TODO(burdon): Immediately trigger callback.
+export const createSubscription = (
+  observer: AccessObserver,
```
dmaretskyi, line 60: I don't think its valid here to pass in the observer. Lets have subscription and observer be decoupled. For example you could subscribe to an object manually, without utilizing the observer

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,127 @@
...
+  onUpdate: (info: UpdateInfo) => void,
+): SubscriptionHandle => {
+  let subscribed = true;
+
+  const subscriptions = Array.from(observer.accessed.values()).reduce((acc, obj) => {
```
dmaretskyi, line 65: Instead I would do subscriptions inside the update function. Compare the new selection and the old set. Subscribe to every new object, and unsubscribe from every old one.

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,127 @@
...
+    acc.add(obj[subscribe](onUpdate));
+
+    return acc;
+  }, new EventSubscriptions());
+
+  let firstUpdate = true;
+
+  const handle = {
```
dmaretskyi, line 73: might be a good time to turn this into a class

packages/common/observable-object/src/observable-object.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface ObservableObject {
+  [subscribe]: (callback: (value: any) => void) => UnsubscribeCallback;
+  _id: string;
+}
+
+class ObservableObjectImpl<T> implements ObservableObject {
+  /**
+   * @internal
+   */
+  _id = PublicKey.random().toHex();
```
dmaretskyi, line 21: If you get rid of ids in subscription, then you wouldn't need it here

packages/common/observable-object/src/observable-object.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  private _callbacks = new Set<(value: any) => void>();
```
dmaretskyi, line 23:
```suggestion
  private _subscribers = new Set<(value: any) => void>();
```

packages/core/echo/echo-schema/src/object.ts
```diff
@@ -51,6 +52,8 @@ export abstract class EchoObject<T extends Model = any> {
    */
   _modelConstructor: ModelConstructor<T>;
 
+  private _callbacks = new Set<(value: any) => void>();
```
dmaretskyi, line 55:
```suggestion
  private _subscribers = new Set<(value: any) => void>();
```

packages/common/observable-object/src/access-observer.ts
wittjosiah: that would then allow us to remove `_id` from the object interface probably

packages/common/observable-object/src/access-observer.ts
wittjosiah: where would the list of values come from in this case (if the observer isn't passed)? That was the only reason I added this, because I wasn't sure where to else to get that from

packages/common/observable-object/src/access-observer.ts
wittjosiah: what would be the goal of making it a class?

packages/common/observable-object/src/observable-object.ts
```diff
@@ -0,0 +1,66 @@
...
+  _emitUpdate() {
+    for (const callback of this._callbacks) {
+      callback(this);
+    }
+  }
+}
+
+export type TypedObservableObject<T extends Record<string, any> = Record<string, any>> = ObservableObjectImpl<T> & T;
+
+export const createStore = <T extends object | any[]>(data?: T): TypedObservableObject<T> => {
```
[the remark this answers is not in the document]
wittjosiah: fair, I can play with that

packages/common/observable-object/src/access-observer.ts
wittjosiah: that makes sense, and would solve the problem about where the list of values comes from, but...what about making sure its subscribed without updates?

packages/common/observable-object/src/access-observer.ts
dmaretskyi: its functionally already a class: it has fields and methods. Just uses wierd syntax right now

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Observes object access.
+ */
+export class AccessObserver {
+  accessed: Set<ObservableObject> = new Set();
+  constructor(public pop: () => void) {}
+}
+
+const getSetFromSelection = (selection: Selection): Set<any> => {
```
dmaretskyi, line 105: Reuse the same set (by recursively passing it into children). Much better for performance

```suggestion
const getSetFromSelection = (selection: Selection, result = new Set<any>): Set<any> => {
```

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        onUpdate({});
+      }
+
+      return handle;
+    },
+    subscribed,
+    selected: new Set<any>(),
+    unsubscribe: () => {
+      subscriptions.clear();
```
dmaretskyi, line 89: unsubscribe from each of them here before clearing

packages/common/observable-object/src/observable-object.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  _emitUpdate() {
+    for (const callback of this._subscriptions) {
+      callback(this);
+    }
+  }
+}
+
+export const createStore = <T extends object | any[]>(data?: T): T => {
```
dmaretskyi, line 50: nice, clean typing

packages/common/observable-object/src/access-observer.ts
```diff
@@ -0,0 +1,115 @@
...
+      if (removed.length > 0 || added.length > 0 || firstUpdate) {
+        firstUpdate = false;
+
+        removed.forEach((obj) => {
+          subscriptions.get(obj)?.();
+          subscriptions.delete(obj);
+        });
+
+        added.forEach((obj) => {
+          if (obj && subscribe in obj) {
```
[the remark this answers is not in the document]
wittjosiah: good call

PR #3271, opened 2023-05-29
feat: Add ability to open browser apps with CLI identity

mykola-vrmchk:
copilot:all

### DIFF
--- a/packages/devtools/cli/src/commands/app/open.ts
+++ b/packages/devtools/cli/src/commands/app/open.ts
@@ -45,7 +45,7 @@ export default class Open extends BaseCommand {
       const { args, flags } = await this.parse(Open);
 
       const pages = await Promise.all(
-        range(5).map(async () => {
+        range(flags.instances).map(async () => {
           const browser = await chromium.launch({ headless: false });
           return await browser.newPage();
         }),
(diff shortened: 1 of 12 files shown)


mykola-vrmchk:
* main
  * **PR #3271** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3271" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3271?utm_source=stack-comment).

packages/devtools/cli/src/commands/app/open.ts
```diff
@@ -0,0 +1,85 @@
...
+    await this.execWithClient(async (client: Client) => {
+      if (!client.halo.identity.get()) {
+        this.log(chalk`{red Profile not initialized.}`);
+        return {};
+      }
+      const { args, flags } = await this.parse(Open);
+
+      const pages = await Promise.all(
+        range(flags.instances).map(async () => {
+          const browser = await chromium.launch({ headless: false });
```
dmaretskyi, line 49: Try reusing the browser instance and instead creating separate `context`s and windows

packages/devtools/cli/src/commands/app/open.ts
```diff
@@ -0,0 +1,85 @@
+//
+// Copyright 2022 DXOS.org
+//
+
+import { Args, Flags, ux } from '@oclif/core';
+import chalk from 'chalk';
+import { chromium } from 'playwright';
```
dmaretskyi, line 7: seems like this could be a heavy dependency and it could slow down CLI startup time. I would do a dynamic import instead in the `run` function of the command

packages/devtools/cli/src/commands/app/open.ts
```diff
@@ -0,0 +1,85 @@
...
+          callbacks: {
+            onConnecting: async () => {
+              const invitationCode = InvitationEncoder.encode(observable.get());
+              pages.forEach(async (page) => {
+                const url = new URL(args.url);
+                url.searchParams.append('deviceInvitationCode', invitationCode);
+                await page.goto(url.href);
+              });
+
+              this.log(chalk`\n{blue Invitation}: ${invitationCode}`);
```
dmaretskyi, line 72: insert the code into the URL instead

PR #3276, opened 2023-05-29
feat(surface): GFM plugin stub

thure:


packages/experimental/surface/package.json
```diff
@@ -19,19 +19,26 @@
...
     "@dxos/aurora": "workspace:*",
     "@dxos/aurora-theme": "workspace:*",
-    "@phosphor-icons/react": "^2.0.5"
+    "react": "*",
+    "react-dom": "*"
+  },
+  "optionalDependencies": {
+    "@braneframe/types": "workspace:*",
+    "@phosphor-icons/react": "^2.0.5",
+    "octokit": "^2.0.14"
```
thure, line 42: I'm marking deps that only the main story needs as optional so they're more easily removed later.

packages/experimental/surface/src/framework/Plugin.ts
```diff
@@ -6,7 +6,10 @@ import { FC, PropsWithChildren } from 'react';
 
 export type PluginProvides<TProvides> = TProvides & {
   context?: FC<PropsWithChildren>;
-  component?: <P extends PropsWithChildren = PropsWithChildren>(datum: any, props?: Partial<P>) => FC<P>;
+  component?: <P extends PropsWithChildren = PropsWithChildren>(
+    datum: any,
+    props?: Partial<P>,
+  ) => FC<PropsWithChildren<{ data: any }>> | undefined | null;
```
thure, line 12: I might be missing some knowledge, I'm not sure how to declare a type for this generic signature in the call to `definePlugin` – how is that supposed to work?

Either way, `P` is not useful here since `Surface` does not render the `FC` using `P`, so I gave it the props it does render the `FC` with and also permit (some) falsy values since `Surface` also recognizes those.

packages/experimental/surface/src/plugins/GithubMarkdownPlugin/GithubMarkdownPlugin.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { definePlugin, PluginDefinition, Surface } from '../../framework';
+import { isSpace } from '../SpacePlugin';
+import { MainAll } from './MainAll';
+import { MainOne } from './MainOne';
+
+// import { OctokitProvider } from './OctokitProvider';
+
+export const isDocument = (datum: any): datum is Document => {
+  return 'type' in datum && 'name' in datum.type && Document.type.name === datum.type.name;
+};
```
thure, line 19: What’s the best way to build a guard like this for an ECHO object?

packages/experimental/surface/src/plugins/GithubMarkdownPlugin/GithubMarkdownPlugin.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const GithubMarkdownPlugin: PluginDefinition = definePlugin({
+  meta: {
+    id: 'dxos:GithubMarkdownPlugin',
+  },
+  provides: {
+    // context: ({ children }) => {
+    //   return <OctokitProvider>{children}</OctokitProvider>;
+    // },
```
thure, line 28: Something about Octokit wasn’t working, will return to this later.

packages/experimental/surface/src/plugins/GithubMarkdownPlugin/GithubMarkdownPlugin.tsx
```diff
@@ -0,0 +1,65 @@
...
+                main: { component: 'dxos:GithubMarkdownPlugin/MainOne' },
+              }}
+            />
+          ),
+        },
+      ],
+    },
+    component: (arg: any) => {
+      if (Array.isArray(arg)) {
+        const [datum, surfaceType] = arg;
```
thure, line 47: This is the pattern I wanted to check in about — maybe I'm misunderstanding the graph concept, but I wanted to explicitly flag somehow that a plugin-component was getting rendered into a surface that considers itself `main`.

Issue #3255, opened 2023-05-25
[composer-app] Stack


thure:
Subtasks:

- [x] Refactor `[data, role]` tuples to `{ data, role }` props on `Surface` to pass to the component it renders.
- [x] _(Wait for Josiah to finish ListView increment)_
- [x] Finish Composer parity (mostly by adjusting the sidebar)
- [x] Delete old Composer
- [ ] Stack support
  - [x] Refactor File I/O and Github I/O into distinct plugins
  - [ ] _(Wait for Josiah to finish final graph increment)_
  - [x] Add Stack plugin
  - [ ] Change `…` menu to a toolbar (?)
  - [ ] Extend File & Github plugins to support Stacks
- [ ] Design increment
  - [ ] Composer has an L0 menu?
  - [ ] DXOS button?

PR #3352, opened 2023-06-07
feat(plugin-stack): Stack plugin

thure:
This PR adds the first Surface plugin: Stack. Doesn’tresolve #3255 just yet.

https://github.com/dxos/dxos/assets/855039/0ba56f95-2772-4fd2-a040-53a767734511

### Summary
📦🔧📝

This pull request refactors the composer-app and extracts two common plugins: `@braneframe/plugin-client` and `@braneframe/plugin-markdown`. These plugins provide functionality for interacting with the DXOS client and markdown documents respectively, and can be reused across different applications. The pull request also updates the composer-app and its plugins to use the external packages and their exports, and adds the necessary configuration, documentation, and license files for the new packages.

> _`composer-app` splits_
> _plugins move to `braneframe` - kireji_
> _autumn of refactor kigo_

### Walkthrough
*  Extract common plugins from composer-app to separate packages: @braneframe/plugin-client and @braneframe/plugin-markdown ([link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-47b84ac128e4ba47b697389d563eb2770189956b67cea33857501b520432a1e1R12-R13), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4789b3ee5b3e63f0181c552c1e9723e927d3e9a1cdca65aa1109c542fc087d17R1-R37), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-32c679980743da55621aabb9ac50fdd63a18bff3f3b16287f227f6635c39f4c7R1-R47), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-2980bfaacf44efa379f06240cc85cd206bc5b09d919e1ecfb8b170c6c4d08cfdL16-R17), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-89029e959cef53069191ee40daffe60afa7921699cc4d0d4da6d17e2a2d466acR1-R5), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-e56910f2262fdeae3091d92ab8065942ae5b436d28156b7b554e0a37c9c88132R1-R33), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-e6d1d345821614148e5995e27dd43e3d23acd9c585a2ae3ef5911c496baf2ccdR1-R44), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-76fca4583f8bbbcd20d1a803a51cda6dcd8fcc919dfd4e00bcfa6680a01437fbR1-R58), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4d53bb710b39a0d80a584d6ab63b7b2fb89608698c12a8d5ac55795dad60d0fcL6-R11), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4d53bb710b39a0d80a584d6ab63b7b2fb89608698c12a8d5ac55795dad60d0fcL23-R23), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4d53bb710b39a0d80a584d6ab63b7b2fb89608698c12a8d5ac55795dad60d0fcR44-R47), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-546798560f91778ea66b6d40d0b725c90d91453b6d9482e567b8717a99a24e59), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-8fd190fe9e7c6e8fd04756668f2c352d146e498d8355fd1dc25b984d1cd19c39))
*  Add license, readme, and project files to the new packages ([link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-8e3e9c30780faf8995ca13a7cd6b0dcd169e28d0385643856991ea61a4104fb9L1-R7), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-cba5385818595541c0359dcdec9eeae5c7db312cc3f410f11206117fd78d1d0fR1-R15), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-682bc4e0ab5ffb6d4b2b4bd0f0fb868526982580a4d6090bc4df1ceb9430e1c8L1-R7), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-86f8480577b780b0f297133f1f8534ba867b5b9bbf3fe641b3b34cd18f6cf7dbR1-R15))
*  Update composer-app to import ClientPlugin and MarkdownPlugin from the new packages ([link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6L10-R14), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-aea39d106265440ec4ddf616421a86cb54f3f68c14c88956a8a5a63c5e4953eaL8-R10), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4c3644e83729f7f016855b3a60b8f477967bcabac0650be106ffd0e7657ec163R21))
*  Remove unnecessary imports and exports of the extracted plugins from composer-app ([link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6L25), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-4c3644e83729f7f016855b3a60b8f477967bcabac0650be106ffd0e7657ec163L39), [link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-7473e9c3e7b4963e6444155c0d34917b567d330e2054bb6dabac2f44ba152481L6))
*  Add references to the new packages in the tsconfig.json file of composer-app ([link](https://github.com/dxos/dxos/pull/3352/files?diff=unified&w=0#diff-aa01c591cecab0e1f8665cacbf254fc9a9418559a3649b0e930ec007c34e2d8cR75-R80))


packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const StackMain = ({
+  data: [stack, properties],
+}: {
+  data: [stack: StackModel, properties: StackProperties];
+}) => {
+  const { t } = useTranslation('dxos:stack');
+  return (
+    <Main.Content classNames='min-bs-[100vh] mli-auto max-is-[60rem] bg-white dark:bg-neutral-925'>
+      <Input.Root>
```
richburdon, line 63: this looks right, although we may want a lower-level reusable component for header (and other things that might live here) shared with the composer plugin. and similarly we may want an low-level abstraction of the stack itself that can be reused in the Presenter plugin which might not have a title/header. so just wanted to plant this thought before the concrete starts to dry.

packages/apps/types/src/proto/schema.proto
```diff
@@ -27,3 +27,28 @@ message Meta {
 
   repeated ForeignKey keys = 1;
 }
+
+// TODO(burdon): Rename.
+// TODO(burdon): System type for external GUIDs (identifies resolver).
+// TODO(burdon): Timestamp, temporal context, TTL.
+// TODO(burdon): Dependency graph (resolvers that preceded this).
+message Source {
```
richburdon, line 35: i think we need to bring in @wittjosiah and @dmaretskyi to discuss how we evolve the type system.
also, i don't think stack.sections should be externalizable objects.

packages/apps/plugins/plugin-stack/src/StackPlugin.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const StackPlugin: PluginDefinition = definePlugin({
+  meta: {
+    id: 'dxos:stack',
+  },
+  provides: {
+    translations,
+    component: (datum, role) => {
+      switch (role) {
+        case 'main':
+          if (Array.isArray(datum) && isStack(datum[0])) {
```
richburdon, line 20: why are we passed an array?

packages/apps/plugins/plugin-stack/src/StackPlugin.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { definePlugin, PluginDefinition } from '@dxos/react-surface';
+
+import { StackMain } from './components';
+import { isStack } from './props';
+import translations from './translations';
+
+export const StackPlugin: PluginDefinition = definePlugin({
+  meta: {
+    id: 'dxos:stack',
```
richburdon, line 13: need to start to think about GUIDs (which we've previously called DXNs). e.g., `dxos.org/plugin/stack`

packages/apps/plugins/plugin-stack/src/props.ts
```diff
@@ -0,0 +1,33 @@
...
+  source: { resolver: string; guid: string };
+  object: unknown;
+};
+
+export type StackModel = {
+  id: string;
+  sections: ObservableArray<StackSectionModel>;
+};
+
+export type StackProperties = {
```
richburdon, line 19: why have both properties and model? i.e., why isn't title just part of model?

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
thure: Sure, I agree. This logic is contained by a component called `StackMain`, ‘main’ indicating its role, so the idea is this is for a specific use-case and may have patterns specific to the role that can be factored out (like the input), or patterns all Stacks should use that can be factored out.

packages/apps/plugins/plugin-stack/src/props.ts
thure: I need to double-check on this – I was following a pattern-of-precedent in MarkdownPlugin, but need to sync with @wittjosiah on the reasoning for it.

richburdon: my suspicion was this is a leakage from the way ECHO objects are structured. and that in our striving to be pure, plugins should just define THEIR required data shape and something then does the mapping. i think this is important for us to be explicit about re proving the plugin model. otherwise we're just pretending to be decoupled from ECHO?

packages/apps/types/src/proto/schema.proto
thure: This model did look like a useful way for the extensibility framework to find a distinct plugin for resolving the `object`… I was struggling to imagine how StackPlugin would resolve these on its own.

packages/apps/plugins/plugin-stack/src/StackPlugin.tsx
wittjosiah: @richburdon I have a todo for this somewhere to revisit DXNs. The format of this is specifically the old DXN format but not well-defined. It's been on my mind :+1:

packages/apps/plugins/plugin-stack/src/props.ts
wittjosiah: @thure So for the markdown plugin the reason its separate is that I had a hard time convincing myself to pollute the `ComposerModel` contract with things its not directly concerned with. Example of this is title, there's no title in the composer model because its the contract for the editor itself. That meant that extra metadata for that model like title, etc. needed to be passed in as a separate parameter.

packages/apps/types/src/proto/schema.proto
wittjosiah: @richburdon My assumption was that by default we'd want stacks not to own the objects referenced in their sections (i.e. soft links). This would mean that the links are just item ids I think? Unless you have some way to indicate that in the protobuf already

packages/apps/plugins/plugin-stack/src/StackPlugin.tsx
zhenyasav: my instinct says to avoid tuples when an object with two keys could be used, so as not to step on the space occupied by arity. i.e. its nice to be able to not worry about special tuples when dealing with arrays, to be able to assume they're arrays of a uniform type - this enables techniques a la graphql, etc. Switching to an object with two keys makes it much easier to duck type with `in` operators.

PR #3513, opened 2023-06-21
feat(client-services): Add node lock file

mykola-vrmchk:
### Summary
📝🐛🛠️

Updated a comment in `run.ts` to clarify the behavior of the agent's forever daemon. This change is part of the effort to improve the agent's shutdown process.

> _`SIGINT` handler_
> _comment changed for clarity_
> _autumn leaves falling_

### Walkthrough
* Updated the comment to clarify the behavior of the agent's forever daemon and the `SIGINT` handler ([link](https://github.com/dxos/dxos/pull/3513/files?diff=unified&w=0#diff-a3925689cd956b853d6f2b1ef28dd26324cdc9ed4f1fd12b2279f7b77b16fbdbL55-R55))


mykola-vrmchk:
* main
  * **PR #3513** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3513" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3513?utm_source=stack-comment).

packages/core/agent/src/forever/forever-daemon.ts
```diff
@@ -57,6 +56,10 @@ export class ForeverDaemon implements Daemon {
 
   async start(profile: string): Promise<ProcessInfo> {
     if (!(await this.isRunning(profile))) {
+      if (isLocked({ lockKey: profile, root: path.join(this._rootDir, 'profile', profile) })) {
```
mykola-vrmchk, line 59: @dmaretskyi see `service-host.ts`

packages/sdk/client-services/src/packlets/services/node-resource-lock.ts
```diff
@@ -0,0 +1,59 @@
...
+    await this._onAcquire?.();
+    log('acquired lock');
+  }
+
+  async release() {
+    if (!lockfile.checkSync(getLockFilePath({ root: this._root, lockKey: this._lockKey }))) {
+      throw new Error(`Lock already acquired: ${this._lockKey}`);
+    }
+    lockfile.unlockSync(getLockFilePath({ root: this._root, lockKey: this._lockKey }));
+    await this._onRelease?.();
```
dmaretskyi, line 51: on release should happen **before** you release the lock: e.g. shut down all of the services that might use the files before releasing the lock

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -89,15 +92,26 @@ export class ClientServicesHost {
...
+            lockKey,
+            onAcquire: () => {
+              if (!this._opening) {
+                void this.open();
+              }
+            },
+            onRelease: () => this.close(),
+          })
+        : new NodeResourceLock({
+            root: path.join(DX_RUNTIME, 'agent', 'profile', lockKey),
```
dmaretskyi, line 106: why not combine root and lockKey into a single param?

browsers really dont care about the format of the string

packages/sdk/client-services/src/packlets/services/resource-lock.ts
```diff
@@ -0,0 +1,23 @@
...
+  lockKey: string;
+  onAcquire?: () => MaybePromise<void>;
+  onRelease?: () => MaybePromise<void>;
+};
+
+/**
+ * Resource lock. Used to coordinate access to shared resources.
+ * For example, start only one `ServicesHost` at a time.
+ */
+export interface ResourceLock {
```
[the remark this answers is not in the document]
dmaretskyi: perhaps in a separate PR?

packages/sdk/client-services/src/packlets/services/service-host.ts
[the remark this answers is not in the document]
mykola-vrmchk: We try to stick to XDG convention https://github.com/dxos/dxos/blob/main/packages/sdk/client-protocol/src/config.ts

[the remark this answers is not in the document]
mykola-vrmchk: So runtime files go into runtime dir and profile-specific files should be in the `./profile/${profile}` folder to avoid collisions

packages/sdk/client/src/packlets/client/utils.ts
```diff
@@ -43,10 +43,14 @@ export const fromIFrame = (
 /**
  * Creates stand-alone services without rpc.
  */
-export const fromHost = (config: Config = new Config()): ClientServicesProvider => {
+export const fromHost = ({
+  config = new Config(),
+  ...params
+}: ClientServicesHostParams = {}): ClientServicesProvider => {
```
dmaretskyi, line 49: is this just to forward the lock key?
unless there's a strong need to treat lock keys differently then I'd just derive them from the storage path and not do this change.

Also it makes sense for me to have the lock combined with the storage as that is what it is meant to protect

packages/core/agent/src/forever/forever-daemon.ts
dmaretskyi: how about `serviceHost.isLockTaken()` to encapsulate the lock

packages/sdk/client-services/src/packlets/services/service-host.ts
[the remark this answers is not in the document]
dmaretskyi: > We try to stick to XDG convention https://github.com/dxos/dxos/blob/main/packages/sdk/client-protocol/src/config.ts

IMHO, seems a bit arbitrary and complicates the code. What kinda benifit are we getting from it?

packages/devtools/cli/src/commands/agent/run.ts
```diff
@@ -52,7 +52,7 @@ export default class Run extends BaseCommand<typeof Run> {
       await agent.monitorEpochs({ limit });
     }
 
-    // NOTE: This is currently called by the agent's forever daemon.
+    // NOTE: This is not currently called by the agent's forever daemon.
```
richburdon, line 55: `forever.startDaemon` literally spawns a Node process that invokes the `dx agent run` handler (i.e., this method). 

Therefore `agent.start` is ALWAYS invoked via this handler. Sometimes by the user calling this directly and sometimes via the forever daemon. BUT only the daemon detects the lock. So running the agent manually and running it at the same time via the daemon will be in conflict.

So my expectation was that the locking should be at the `agent` not `daemon` level?

Also... the daemon doesn't know about agents that are running manually via `dx agent run`. Perhaps we should ONLY consider running agents via the daemon and have a `--foreground` flag that blocks the CLI? @dmaretskyi ?

packages/experimental/kai-framework/src/hooks/useClientProvider.ts
```diff
@@ -18,7 +18,7 @@ export const useClientProvider = (dev: boolean) => {
     const config = await configProvider();
     const client = new Client({
       config,
-      services: config.get('runtime.app.env.DX_VAULT') === 'false' ? fromHost(config) : fromIFrame(config),
+      services: config.get('runtime.app.env.DX_VAULT') === 'false' ? fromHost({ config }) : fromIFrame(config),
```
richburdon, line 21: Perhaps factor out (or look for) a parseBool that treats "0" as false also. I think we do this a lot.

packages/devtools/cli/src/commands/agent/run.ts
mykola-vrmchk: Agent level for sure, I want `daemon` to fail if you already running the agent yourself

packages/sdk/client-services/src/packlets/services/service-host.ts
mykola-vrmchk: I want to have the lock file and database separately because lock file is runtime data and will be cleared a lot and the database should be persistent

mykola-vrmchk: Also, I do not want to deal with the issues with browser lock that Josiah described

packages/sdk/client/src/packlets/client/utils.ts
mykola-vrmchk: I want to have the lock file and database separately because lock file is runtime data and will be cleared a lot and the database should be persistent

mykola-vrmchk: Also, I do not want to deal with the issues with browser lock that Josiah described

PR #3531, opened 2023-06-22
Invitation refactor spec

dmaretskyi:
### Summary
📡🤝🛠️

This pull request enhances the `services.proto` file to support offline and cross-network invitations, as well as different invitation types and authentication methods. It adds a `Beacon` message and an `IDENTITY` enum value, and simplifies the `InvitationProcess` message.

> _Sing, O Muse, of the cunning code that reshaped the `Invitation` message_
> _And added the `Beacon` type, a signal of friendship and courage_
> _Like the blazing star that guided the Argonauts to the golden fleece_
> _Or the fire that Prometheus stole from the gods for the sake of mortals._

### Walkthrough
*  Introduce new `Beacon` message type and new `IDENTITY` enum value for `Invitation` message to support offline and cross-network invitations and different authentication methods ([link](https://github.com/dxos/dxos/pull/3531/files?diff=unified&w=0#diff-1f246514dbba8145aea2c94eadcc3265b76da0f434106ef87e58db92bdaf5aa0L271-R279), [link](https://github.com/dxos/dxos/pull/3531/files?diff=unified&w=0#diff-1f246514dbba8145aea2c94eadcc3265b76da0f434106ef87e58db92bdaf5aa0L295-R332)).


packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -292,8 +297,39 @@ message Invitation {
...
+  /// Beacon allows two peers to rendezvous on a p2p network.
+  Beacon beacon = 5;
+
+  optional google.protobuf.Timestamp expiry = 6;
+
+  /// Space to join (only present if kind is SPACE).
+  optional dxos.keys.PublicKey space_key = 7;
+
+  /// Idenitity key of the invitation recipient (only present if auth_method is IDENTITY).
+  optional dxos.keys.PublicKey recipient = 8;
```
wittjosiah, line 326: I'd update the comments here to indicate that these are only filled out after the rendezvous, right? Currently they feel misleading that the space key would always be there if it's a space invitation, etc.

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -330,9 +356,6 @@ message Invitation {
   /// Guest's identity.
   optional dxos.keys.PublicKey identity_key = 8;
 
-  /// Space to join (only present if kind is SPACE).
-  optional dxos.keys.PublicKey space_key = 9;
-
   /// Authentication code created by host (only present if auth_method is SHARED_SECRET).
```
wittjosiah, line 359: This should be updated to say that it's only available for the host and is excluded from encoding when invitations are shared

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -292,8 +297,39 @@ message Invitation {
...
+  /// Kind of access the invitation will grant.
+  Kind kind = 3;
+
+  /// How the invitation is authenticated.
+  AuthMethod auth_method = 4;
+
+  /// Beacon allows two peers to rendezvous on a p2p network.
+  Beacon beacon = 5;
+
+  optional google.protobuf.Timestamp expiry = 6;
```
nferch, line 320: Should we include mention of the semantics around this? E.g. is this validated by the recipient, is it intended to be a security feature or just UI, etc.

nferch (commented):
Is there some reference to the associated user flow and behavior we could include here?

Is this protocol change expected to change that at all?

PR #3584, opened 2023-06-28
feat(composer-app): Stack integration

thure:
This PR adds `StackPlugin` to Composer and finishes its integration with `MarkdownPlugin`.

https://github.com/dxos/dxos/assets/855039/d7153423-a937-4bec-9149-b68d855bf649

### Summary
🧩📝🗃️

This pull request adds a new stack plugin to the composer app, and enhances the markdown plugin to support it. The stack plugin allows users to create and manipulate stacks of cards on the surface component, using different types of sections. The markdown plugin provides a section type for markdown documents, and a dialog for choosing them from the current space. The pull request also updates the proto schema, the package dependencies, the translations, and the TypeScript configurations for the affected packages.

> _The composer app got a new feature_
> _To create stacks of cards with a creature_
> _The markdown plugin_
> _Helps to fill them with grin_
> _And the stack plugin handles the structure_

### Walkthrough
*  Add and register the stack plugin to the composer app ([link](https://github.com/dxos/dxos/pull/3584/files?diff=unified&w=0#diff-47b84ac128e4ba47b697389d563eb2770189956b67cea33857501b520432a1e1R17), [link](https://github.com/dxos/dxos/pull/3584/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R15), [link](https://github.com/dxos/dxos/pull/3584/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R39), [link](https://github.com/dxos/dxos/pull/3584/files?diff=unified&w=0#diff-aa01c591cecab0e1f8665cacbf254fc9a9418559a3649b0e930ec007c34e2d8cR90-R92)).


packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -30,7 +30,11 @@ export const markdownPlugins = (plugins: Plugin[]): MarkdownPlugin[] => {
   return (plugins as MarkdownPlugin[]).filter((p) => Boolean(p.provides?.markdown));
 };
 
-type MarkdownPluginProvides = SpaceProvides & TranslationsProvides;
+type MarkdownPluginProvides = SpaceProvides &
+  TranslationsProvides & {
+    // todo(thure): Refactor this to be DRY, but avoid circular dependencies. Do we need a package like `plugin-types` 😬? Alternatively, StackPlugin stories could exit its package, but we have no such precedent.
```
wittjosiah, line 35: I think this type will be generalized and not specific to stack. `SpaceProvides` is effectively the same thing. Between interoperable graph and generic actions this should end up not depending on stack but just exposing something that stack can find and use.

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -85,7 +100,12 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
             }
             break;
           case 'section':
-            if (datum && typeof datum === 'object' && 'object' in datum && isMarkdown(datum.object)) {
+            if (
+              datum &&
+              typeof datum === 'object' &&
+              typeof (datum as { [key: string]: any }).object === 'object' &&
```
wittjosiah, line 106: if you want to avoid casting, the check typescript is looking for is `'object' in datum` before allowing you to do `typeof datum.object`

packages/apps/plugins/plugin-markdown/src/props.ts
```diff
@@ -15,8 +15,8 @@ export const isMarkdown = (datum: unknown): datum is ComposerModel =>
   datum && typeof datum === 'object'
     ? 'id' in datum &&
       typeof datum.id === 'string' &&
-      'content' in datum &&
-      (typeof datum.content === 'string' || datum.content instanceof YText)
+      (typeof (datum as { [key: string]: any }).content === 'string' ||
```
wittjosiah, line 18: Why change this to casting?

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
```diff
@@ -79,19 +80,20 @@ const StackMainImpl = ({ sections }: { sections: StackSections }) => {
   const [_, setIter] = useState([]);
   const { t } = useTranslation('dxos:stack');
 
-  useEffect(() => {
-    // todo(thure): TypeScript seems to get the wrong return value from `ObservableArray.subscribe`
-    return sections[subscribe](() => setIter([])) as () => void;
-  }, []);
+  // todo(thure): Is there a hook that is compatible with both `ObservedArray`s and `TypedObject`s?
```
wittjosiah, line 83: echo objects/arrays have `subscribe` key as well, I suspect what you may have run into though is that sections is the result of an echo query, which would just be a regular array I think @dmaretskyi perhaps something for us to think about with these apis

packages/apps/plugins/plugin-markdown/src/props.ts
thure: The `in` operator doesn’t work for the `Proxy`s of `TypedObject`s we may sometimes see here, so trying to `get` the property with dot notation seemed like a reasonable workaround, though I’d defer to any better pattern.

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
thure: I see — I’d definitely prefer to have consistent interface for subscribing for both cases.

PR #3588, opened 2023-06-28
fix: Logging service improvements

dmaretskyi:
copilot:all

### DIFF
--- a/packages/core/protocols/src/proto/dxos/client/services.proto
+++ b/packages/core/protocols/src/proto/dxos/client/services.proto
@@ -422,13 +422,21 @@ enum LogLevel {
 }
 
 message QueryLogsRequest {
+  enum MatchingOptions {
+    /// Filters match all log levels above. Default behaviour for the logger.
+    INCLUSIVE = 1;
+
+    /// Explicitly specify all desired log levels.
+    EXPLICIT = 2;
+  }
+
   message Filter {
     LogLevel level = 1;
     optional string pattern = 2;
   }
 
-  /// Explicitly specify all desired log levels.
   repeated Filter filters = 1;
+  optional MatchingOptions options = 10;
 }
 
 message LogEntry {
--- a/packages/common/codec-protobuf/src/substitutions/struct.ts
+++ b/packages/common/codec-protobuf/src/substitutions/struct.ts
@@ -19,7 +19,8 @@ const encodeStructValue = (structValue: any, visitedObjects: WeakSet<any>): any 
     case 'boolean': {
       return { boolValue: structValue };
     }
-    case 'object': { // null, Array, Object will have typeof 'object'
+    case 'object': {
+      // null, Array, Object will have typeof 'object'
       if (structValue === null || visitedObjects.has(structValue)) {
         return { nullValue: 0 };
       }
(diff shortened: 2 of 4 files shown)


packages/common/util/src/entry.ts
```diff
@@ -1,10 +1,10 @@
...
- * Inspired by https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry
- */
 //
 // Copyright 2023 DXOS.org
 //
 
+/**
+ * Inspired by https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry
+ */
 export const entry = <K, V>(map: Map<K, V>, key: K): MapEntry<K, V, undefined> => new MapEntry(map, key);
```
richburdon, line 8: rather than orInsert, to me it would be clearer to have something like https://docs.python.org/3/library/collections.html#collections.defaultdict

e.g., `const value = createOrInsert(map, key, () => 'default init value')`

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -426,13 +426,21 @@ enum LogLevel {
 }
 
 message QueryLogsRequest {
+  enum MatchingOptions {
+    /// Filters match all log levels above. Default behaviour for the logger.
+    INCLUSIVE = 1;
```
richburdon, line 431: AND/OR?

packages/sdk/client/src/packlets/diagnostics/diagnostics.ts
```diff
@@ -40,6 +40,7 @@ export type DiagnosticOptions = {
   humanize?: boolean;
 };
 
+// TODO(burdon): Move method to Monitor class.
```
richburdon, line 43: NOTE: intent to start to consolidate these -- since they are both new!

packages/sdk/client/src/packlets/diagnostics/monitor.ts
```diff
@@ -0,0 +1,100 @@
...
+          const { spaceKey } = event.context! as DataPipelineProcessed;
+          // TODO(burdon): Make default value a provider? entry(map, key, () => ({ value });
+          const state = entry(this._state, spaceKey).orInsert({ processed: 0, throughput: [] }).value;
+          state.processed++;
+
+          if (state.throughput.length === 0) {
+            state.throughput.push({ begin: Date.now(), end: 0, processed: 0 });
+          }
+
+          // TODO(burdon): Configure option to just store everything and reduce when requested?
```
richburdon, line 87: It would be good to have a config option so that we can maintain the full times series (need to add actual timestamp to log) and then map-reduce these as needed (e.g., specifying a bucket size/period and/or running different reducers/loggers -- incl. ability to define the reducer on the fly from the console).

packages/sdk/client/src/packlets/diagnostics/monitor.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { expect } from 'chai';
+
+import { describe, test } from '@dxos/test';
+import { range } from '@dxos/util';
+
+export type TimeRecord = { timestamp: number };
+export type TimeBucket = { start: number; period: number; count: number };
+
+export const createBucketReducer =
```
richburdon, line 13: E.g., here's a trivial example of a reducer we might want to call on-the-fly from console and/or devtools.

PR #3589, opened 2023-06-29
feat(echo-pipeline): Cancel Epoch download if there is new Epoch detected

mykola-vrmchk:
### Summary
🚀🛠️🔄

Improved snapshot loading in `echo-pipeline` by adding cancellation support. This allows `SnapshotManager` to abort loading when the context is cancelled or the host is closed.

> _The snapshot of doom is loading_
> _But we can cancel it with `Context`_
> _We won't let it crash our system_
> _We are the masters of `SnapshotManager`_

### Walkthrough
*  Add cancellation support for loading snapshots using `Context` and `cancelWithContext` ([link](https://github.com/dxos/dxos/pull/3589/files?diff=unified&w=0#diff-bcb3805db9fb8df2d1dd3cc8aba301356a57e030fc1e534cf6e9d7b18754b51aR7), [link](https://github.com/dxos/dxos/pull/3589/files?diff=unified&w=0#diff-bcb3805db9fb8df2d1dd3cc8aba301356a57e030fc1e534cf6e9d7b18754b51aL60-R67))
  - Import `Context` and `cancelWithContext` from `@dxos/context` in `snapshot-manager.ts` ([link](https://github.com/dxos/dxos/pull/3589/files?diff=unified&w=0#diff-bcb3805db9fb8df2d1dd3cc8aba301356a57e030fc1e534cf6e9d7b18754b51aR7))
  - Modify `load` method of `SnapshotManager` class to accept a `Context` parameter and use `cancelWithContext` to wrap snapshot loading and downloading operations ([link](https://github.com/dxos/dxos/pull/3589/files?diff=unified&w=0#diff-bcb3805db9fb8df2d1dd3cc8aba301356a57e030fc1e534cf6e9d7b18754b51aL60-R67))
  - Pass the `Context` parameter to the underlying `this._snapshotStore.loadSnapshot` and `this._objectSync.download` methods, which can use it for additional options or information ([link](https://github.com/dxos/dxos/pull/3589/files?diff=unified&w=0#diff-bcb3805db9fb8df2d1dd3cc8aba301356a57e030fc1e534cf6e9d7b18754b51aL60-R67))

### DIFF
--- a/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
+++ b/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
@@ -64,7 +64,7 @@ export class SnapshotManager {
       return local;
     }
 
-    const remote = await cancelWithContext(ctx, this._objectSync.download(id));
+    const remote = await cancelWithContext(ctx, this._objectSync.download({ id, ctx }));
     return schema.getCodecForType('dxos.echo.snapshot.SpaceSnapshot').decode((remote.payload as Any).value);
   }
 
--- a/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
+++ b/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
@@ -4,6 +4,7 @@
 
 import { trackLeaks } from '@dxos/async';
 import { Any } from '@dxos/codec-protobuf';
+import { Context, cancelWithContext } from '@dxos/context';
 import { timed } from '@dxos/debug';
 import { log } from '@dxos/log';
 import { schema } from '@dxos/protocols';
@@ -57,13 +58,13 @@ export class SnapshotManager {
   }
 
   @timed(10_000)
-  async load(id: string): Promise<SpaceSnapshot> {
-    const local = await this._snapshotStore.loadSnapshot(id);
+  async load(ctx: Context, id: string): Promise<SpaceSnapshot> {
+    const local = await cancelWithContext(ctx, this._snapshotStore.loadSnapshot(id));
     if (local) {
       return local;
     }
 
-    const remote = await this._objectSync.download(id);
+    const remote = await cancelWithContext(ctx, this._objectSync.download(id));
     return schema.getCodecForType('dxos.echo.snapshot.SpaceSnapshot').decode((remote.payload as Any).value);
   }
 
--- a/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
+++ b/packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
@@ -64,7 +64,7 @@ export class SnapshotManager {
       return local;
     }
 
-    const remote = await cancelWithContext(ctx, this._objectSync.download({ id, ctx }));
+    const remote = await cancelWithContext(ctx, this._objectSync.download(ctx, id));
     return schema.getCodecForType('dxos.echo.snapshot.SpaceSnapshot').decode((remote.payload as Any).value);
   }
 
--- a/packages/core/echo/echo-pipeline/src/space/data-pipeline.ts
+++ b/packages/core/echo/echo-pipeline/src/space/data-pipeline.ts
@@ -82,6 +82,7 @@ export class DataPipeline {
   public currentEpoch?: SpecificCredential<Epoch>;
   private _lastProcessedEpoch = -1;
   public onNewEpoch = new Event<Credential>();
+  private _epochCtx?: Context;
 
   get isOpen() {
     return this._isOpen;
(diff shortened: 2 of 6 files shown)


packages/core/echo/echo-pipeline/src/space/data-pipeline.ts
```diff
@@ -277,18 +273,33 @@ export class DataPipeline {
...
+  private async _processEpochInSeparateTask(epoch: SpecificCredential<Epoch>) {
+    if (epoch.subject.assertion.number <= this._lastProcessedEpoch) {
       return;
     }
-    this._lastProcessedEpoch = epoch.number;
-    assert(this._isOpen); // TODO: In the future we might process epochs before we are open so that data pipeline starts from the last one.
+    await this._epochCtx?.dispose();
+    const ctx = new Context();
+    this._epochCtx = ctx;
+    scheduleTask(ctx, async () => {
```
dmaretskyi, line 283: catch the CancelledError here

packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
```diff
@@ -57,13 +58,13 @@ export class SnapshotManager {
...
-  async load(id: string): Promise<SpaceSnapshot> {
-    const local = await this._snapshotStore.loadSnapshot(id);
+  async load(ctx: Context, id: string): Promise<SpaceSnapshot> {
+    const local = await cancelWithContext(ctx, this._snapshotStore.loadSnapshot(id));
     if (local) {
       return local;
     }
 
-    const remote = await this._objectSync.download(id);
+    const remote = await cancelWithContext(ctx, this._objectSync.download(id));
```
dmaretskyi, line 67: this is already good, but can you drill the cancellation deeper into objectSync (and add a unit test)

image


Download works by adding a CID to the "want list" and then waiting for other peers to send it to you. so the cancellation would remove this from the "want list". Becareful that there might be multiple concurent requests to `download` with the same CID so you need to count them and remove the CID from the want list only if there's 0 active downloads for it.

packages/core/echo/echo-pipeline/src/dbhost/snapshot-manager.ts
```diff
@@ -57,13 +58,13 @@ export class SnapshotManager {
...
-  async load(id: string): Promise<SpaceSnapshot> {
-    const local = await this._snapshotStore.loadSnapshot(id);
+  async load(ctx: Context, id: string): Promise<SpaceSnapshot> {
+    const local = await cancelWithContext(ctx, this._snapshotStore.loadSnapshot(id));
     if (local) {
       return local;
     }
 
-    const remote = await this._objectSync.download(id);
+    const remote = await cancelWithContext(ctx, this._objectSync.download({ id, ctx }));
```
dmaretskyi, line 67: prob no need to wrap if youre passing ctx furether down

dmaretskyi, line 67: keep the same api with ctx as first paramenter download(ctx, id)

packages/core/mesh/teleport-extension-object-sync/src/object-sync.ts
```diff
@@ -19,7 +19,7 @@ export type ObjectSyncParams = {
 export class ObjectSync {
   private readonly _ctx = new Context();
 
-  private readonly _downloadRequests = new Map<string, Trigger<DataObject>>();
+  private readonly _downloadRequests = new Map<string, { trigger: Trigger<DataObject>; counter: number }>();
```
dmaretskyi, line 22:
```
type DownloadRequest = {
 trigger: Trigger<DataObject>;
 counter: number
 }

```

packages/core/mesh/teleport-extension-object-sync/src/object-sync.ts
```diff
@@ -30,15 +30,39 @@ export class ObjectSync {
...
+    }).value;
 
     for (const extension of this._extensions) {
       extension.updateWantListInASeparateTask(new Set(this._downloadRequests.keys()));
     }
 
-    return trigger.wait();
+    ctx?.onDispose(() => {
+      // Remove request if context is disposed and nobody else requests it.
+      const request = entry(this._downloadRequests, id).value;
```
dmaretskyi, line 53: could be just this._downloadRequests.get(id) if youre not using other `entry` functions

packages/core/mesh/teleport-extension-object-sync/src/object-sync.ts
```diff
@@ -30,15 +30,39 @@ export class ObjectSync {
...
+      if (!request) {
+        return;
+      }
+      if (--request.counter === 0) {
+        this._downloadRequests.delete(id);
+      }
+      for (const extension of this._extensions) {
+        extension.updateWantListInASeparateTask(new Set(this._downloadRequests.keys()));
+      }
```
dmaretskyi, line 62: those line are repeated so extract them as a method

packages/core/mesh/teleport-extension-object-sync/src/object-sync.test.ts
```diff
@@ -78,4 +80,122 @@ describe('ObjectSync', () => {
...
+              return undefined;
+            },
+            setObject: async (data: DataObject) => {
+              // console.log('setObject', data);
+            },
+          }),
+        ),
+    });
+
+    await testBuilder.connzwect(peer1, peer2);
```
dmaretskyi, line 180: typo?

mykola-vrmchk: Yep

PR #3590, opened 2023-06-29
feat: kanban plugin

richburdon:
copilot:all


thure (commented):
I don't want to block `KanbanPlugin` from entering a fully-fledged app on `main`, but I’d also want an opportunity to more closely vet it and commit to its maintenance.

Could we set up an app separate from Composer that would be a good target for all the plugins coming from Kai frames, maybe `kai-next` or even `composer-next`?

PR #3601, opened 2023-07-04
feat: Chunked blob sync

dmaretskyi:
```[tasklist]
### Task list
- [x] Persistant blob store: data + meta (bitmap)
- [x] Blob sync protocol that syncs chunks
- [x] Compatibility between snapshot store and blob store: eventually we want to get rid of separate snapshot store, but for now we'd have duplicates to not break current spaces. Snapshot manager will fall back to old snapshot store on reads, and on writes write only to new blob store.
- [x] SnapshotManager takes ObjectSync as dependency
- [x] Rename ObjectSync to BlobSync: package, class, proto, extension name, etc.
- [x] Update want list as a sheduled task with throttling
- [x] throttle pushing chunks: max concurent requests per peer
- [x] stretch: devtools for blobs
```


mykola-vrmchk:
* main
  * **PR #3601** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3601" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3601?utm_source=stack-comment).

packages/core/mesh/teleport-extension-object-sync/src/blob-store.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    assert(meta.bitfield, 'Bitfield not present');
+    assert(chunk.chunkOffset !== undefined, 'chunkOffset is not present');
+
+    // Write chunk.
+    await this._getDataFile(chunk.id).write(chunk.chunkOffset, Buffer.from(chunk.payload));
+
+    // Update bitfield.
+    BitField.set(meta.bitfield, Math.floor(chunk.chunkOffset / meta.chunkSize), true);
```
dmaretskyi, line 131: assert chunk offset is a multiple of chunk size

same for length so it aligns to chunks

dmaretskyi: also if meta was previously set, assert that new chunkSize matches the old one

mykola-vrmchk: > also if meta was previously set, assert that new chunkSize matches the old one

see line 121

PR #3633, opened 2023-07-11
test: extended network testbench

egorgripasov:
Current test spec:

- 4 peers
- 5 swarms
- multiple test iterrations, within the single iterration:
  - each peer opens 15 bidirectional streams with data constantly flowing through
  - each peer makes separate rpc call within each swarm to each remotePeer
  - each peer joins 1 extra (delayed) swarm 

Changes PR introduce:

- new test extension to create streams between peers 
- muxer & framer now appreciate a backpressure from underlying transport (wrtc channel mainly)
- balancer that distributes data from extensions in a round robin fashion

### Summary
🔄🐝🔌

Refactored and updated the transport test plan for the gravity package. Moved the test logic from `transport-spec.ts` to `main.ts` and added new parameters to control the swarm formation and repetition. Improved the signal server script to support local testing and reconnection. Adjusted the test spec to use fewer agents and longer duration.

> _Oh we are the coders of the `main.ts` file_
> _We change the specs and we test the transport_
> _We heave and we ho and we wait for the swarm_
> _We repeat and we tweak till the code is performant_

### Walkthrough
*  Rewrite the `run` method of the `TransportTestPlan` class to implement the new logic for the transport test plan ([link](https://github.com/dxos/dxos/pull/3633/files?diff=unified&w=0#diff-13e9d8cb236a33b448cd0ee2aec268f56d3879148389d157f9a9b2a5a0a6aec2L78-R164)):


packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,233 @@
...
+      bytesSent: 0,
+      bytesReceived: 0,
+      sendErrors: 0,
+      receiveErrors: 0,
+      startTimestamp: Date.now(),
+    };
+
+    streamEntry.timer = setInterval(() => {
+      const chunk = randomBytes(chunkSize);
+      networkStream.write(chunk, 'binary', (err) => {
```
dmaretskyi, line 55: `write` method returns true/false based on it having free space in the buffer for more chunks https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback

you can add backpressure using that, though i'm not sure if it will work correctly in this test as it  needs to be correctly hooked up all the way to the transort

packages/core/mesh/teleport/src/muxing/rpc-port.ts
```diff
@@ -10,6 +10,6 @@ import { MaybePromise } from '@dxos/util';
  * NOTE: Copied from @dxos/rpc to avoid dependency. Structural typing should still work.
  */
 export interface RpcPort {
-  send: (msg: Uint8Array) => MaybePromise<void>;
+  send: (msg: Uint8Array, cb?: () => void) => MaybePromise<void>;
```
dmaretskyi, line 13: What's the difference between callback and the promise here?

packages/core/mesh/teleport/src/muxing/muxer.ts
```diff
@@ -115,14 +116,25 @@ export class Muxer {
     };
 
     // NOTE: Make sure channel.push is set before sending the command.
-    this._sendCommand({
-      openChannel: {
-        id: channel.id,
-        tag: channel.tag,
-        contentType: channel.contentType,
+    this._sendCommand(
```
dmaretskyi, line 119: why dont you just use promises?

egorgripasov: i've started with callbacks, since this is a default way of controlling streams, and kinda in process of converting to promises. i think i hit some issues and it just did not work quickly, so i left callback here for a moment

packages/core/mesh/teleport/src/muxing/rpc-port.ts
egorgripasov: i think `muxer` generally avoids `async` funcs, so it's simpler to make it work with the callbacks

packages/common/node-std/src/crypto.js
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+const notAvailable = () => {
```
richburdon, line 5: why factor this out as a separate function?

packages/core/mesh/network-manager/src/testing/test-wire-protocol.ts
```diff
@@ -47,4 +57,25 @@ export class TestWireProtocol {
     const connection = await this.waitForConnection(peerId);
     await connection.test(message);
   }
+
+  async startStream(
```
richburdon, line 61: rename open/close (standardized)

packages/core/mesh/teleport/src/index.ts
```diff
@@ -5,4 +5,5 @@
 export * from './muxing';
 export * from './teleport';
 export * from './testing/test-extension';
+export * from './testing/test-extension-with-streams';
```
richburdon, line 8: crate `index.ts` within folder.

packages/core/mesh/teleport/src/muxing/muxer.ts
```diff
@@ -175,12 +196,22 @@ export class Muxer {
     }
     this._destroying = true;
 
-    this._sendCommand({
-      destroy: {
-        error: err?.message,
-      },
-    });
-    this._dispose();
+    Promise.resolve(
```
richburdon, line 199: convert to try/catch/finally?

packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+export class TestExtensionWithStreams implements TeleportExtension {
+  public readonly open = new Trigger();
+  public readonly closed = new Trigger();
+
+  public extensionContext: ExtensionContext | undefined;
+  private _rpc!: ProtoRpcPeer<{ TestServiceWithStreams: TestServiceWithStreams }>;
+
+  private readonly _streams = new Map<string, TestStream>();
```
richburdon, line 29: gather readonly props up front together.

packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  constructor(public readonly callbacks: TestExtensionWithStreamsCallbacks = {}) {}
+
+  get remotePeerId() {
+    return this.extensionContext?.remotePeerId;
+  }
+
+  private async _loadStream(streamTag: string, interval = 5, chunkSize = 2048) {
```
richburdon, line 37: rename _open/_close stream (to match).

packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,244 @@
...
+          process.nextTick(pushChunk);
+        }
+      }, interval);
+    };
+
+    pushChunk();
+
+    this._streams.set(streamTag, streamEntry);
+
+    networkStream.on('data', (data) => {
```
richburdon, line 77: remove handlers on close.

packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,244 @@
...
+      streamEntry.bytesReceived += data.length;
+    });
+
+    networkStream.on('error', (err) => {
+      streamEntry.receiveErrors += 1;
+    });
+  }
+
+  private _closeStream(streamTag: string): Stats {
+    assert(this._streams.has(streamTag), `Stream ${streamTag} does not exist.`);
```
richburdon, line 87: For all errors, logs, try to make in the form of "Stream does not exist: xxx" (which helps keep things uniform and helps with filtering logs).

packages/core/mesh/teleport/src/testing/test-extension-with-streams.ts
```diff
@@ -0,0 +1,244 @@
...
+  bytesReceived: number;
+  sendErrors: number;
+  receiveErrors: number;
+  runningTime: number;
+};
+
+export type TestStreamStats = {
+  streamTag: string;
+  stats: {
+    localPeer: Stats;
```
richburdon, line 231: i think local/remote is perhaps just as clear?

packages/gravity/kube-testing/src/plan/transport-spec.ts
```diff
@@ -68,43 +72,248 @@ export class TransportTestPlan implements TestPlan<TransportTestSpec, TransportA
...
+            await swarm.protocol.startStream(
+              PublicKey.from(agentId),
+              streamTag,
+              spec.streamLoadInterval,
+              spec.streamLoadChunkSize,
+            );
+            actualStreams++;
+            log.info('test stream started', { agentIdx, swarmIdx });
+          } catch (error) {
+            log.info('test stream failed', { agentIdx, swarmIdx, error });
```
richburdon, line 219: log.error?

packages/gravity/kube-testing/src/plan/transport-spec.ts
```diff
@@ -2,7 +2,8 @@
 // Copyright 2023 DXOS.org
```
richburdon, line 2: why `-spec.ts` suffix? should this be `.test.ts`?

packages/gravity/kube-testing/src/plan/transport-spec.ts
```diff
@@ -68,43 +72,248 @@ export class TransportTestPlan implements TestPlan<TransportTestSpec, TransportA
...
+
+      await sleep(10_000);
 
-    log.info('sleeping', { agentIdx });
+      // Start streams on all swarms.
+      {
+        log.info('starting streams', { agentIdx });
+
+        // TODO(egorgripasov): Multiply by iterration number.
+        const desiredStreems = (numOfAgents - 1) * spec.swarmsPerAgent;
```
richburdon, line 203: actual/target?, numAgents?

packages/gravity/kube-testing/src/plan/transport-spec.ts
richburdon: Should we rename this pacakge?

packages/core/mesh/teleport/src/muxing/muxer.ts
egorgripasov: i'm having some troubles making it work with `@synchronized` tag on `teleport.destroy`, so i'll squeeze those change to the next iteration of network updates

packages/gravity/kube-testing/src/plan/transport-spec.ts
egorgripasov: it's not `.test.ts` because it's not a test runnable by `mocha`, but rather a spec for a test that is executed via our own testing framework (we use it within kube-testing to test signal, transport & echo)

PR #3641, opened 2023-07-12
feat: Agent plugins

richburdon:
copilot:all


packages/core/agent/src/defs.ts
mykola-vrmchk, line 1: @wittjosiah proposed to name such file `src/timeouts.ts` and put all timeouts for a package in it

mykola-vrmchk: see https://github.com/dxos/dxos/blob/main/packages/sdk/client-protocol/src/timeouts.ts

mykola-vrmchk (approved):

PR #3687, opened 2023-07-19
refactor(client): better support code-splitting

wittjosiah:
### Summary
🚚🛠️📦

This pull request refactors the `client-protocol` and `client-services` packages to use submodules and types more consistently, and updates the `todomvc` app to use the new structure of the `react-client` package. It also improves the vite config and the code style of some files.

> _`client` refactored_
> _submodules and types added_
> _vite config changed_

### Walkthrough
*  Update import paths for `@dxos/react-client` and `@dxos/client` packages, which have been restructured into submodules ([link](https://github.com/dxos/dxos/pull/3687/files?diff=unified&w=0#diff-d447a8d540fc418b81216f6b9c393ee43400c0bc67b007d71a9b2932f3051925L8-R10), [link](https://github.com/dxos/dxos/pull/3687/files?diff=unified&w=0#diff-18d15ae63cfd94c823d8a6accbab90f9ba6b33191480d79788fb0bd74ed44022L7-R7), [link](https://github.com/dxos/dxos/pull/3687/files?diff=unified&w=0#diff-7d0880dec3d6fb2faf12fec2810e4418ca91e2094cb49c6597f3ce5313c2c825L9-R10), [link](https://github.com/dxos/dxos/pull/3687/files?diff=unified&w=0#diff-31fb79b5fa724f1f5b78271c48be3075c5b4a5a4d2e8c88e9df5da45780fa90fL8-R8), [link](https://github.com/dxos/dxos/pull/3687/files?diff=unified&w=0#diff-f944b8f3441c07bd462c9f88616b3d94a9b7e0fbd20b80b6907d066552c3603fL12-R13)).


packages/sdk/client/src/client.ts
```diff
@@ -56,13 +54,14 @@ export class Client {
...
-  private readonly _modelFactory: ModelFactory;
-  private readonly _services: ClientServicesProvider;
-  private readonly _halo: HaloProxy;
-  private readonly _echo: EchoProxy;
-  private readonly _mesh: MeshProxy;
-  private readonly _monitor: Monitor;
+  private readonly _options: ClientOptions;
+  private _config!: Config;
+  private _modelFactory!: ModelFactory;
+  private _services!: ClientServicesProvider;
```
dmaretskyi, line 60: make the getters error with "Client not initialized"

packages/sdk/react-client/package.json
```diff
@@ -6,48 +6,96 @@
   "bugs": "https://github.com/dxos/dxos/issues",
   "license": "MIT",
   "author": "DXOS.org",
-  "main": "dist/lib/node/index.cjs",
-  "browser": {
-    "./dist/lib/node/index.cjs": "./dist/lib/browser/index.mjs",
-    "./testing.js": "./dist/lib/browser/testing/index.mjs"
+  "exports": {
```
dmaretskyi, line 9: did you have to do that by hand? would be cool for esbuild executor to set those

packages/sdk/client/src/index.ts
```diff
@@ -2,53 +2,57 @@
...
-  ShowDeletedOption,
-  type QueryOptions,
-  type SchemaDef,
-  type SchemaField,
-  type SchemaRef,
-} from '@dxos/echo-db';
+// export { generateSeedPhrase } from '@dxos/credentials';
+
+// export {
+//   TYPE_SCHEMA,
```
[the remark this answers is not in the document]
wittjosiah: oh whoops, I should remove these comments, they're exported from other entrypoints now

packages/sdk/react-client/package.json
wittjosiah: yeah, I copy-pasta, but executor doing it might make sense if it turns out to always be the same (which I think it is)

packages/sdk/client/src/client.ts
```diff
@@ -75,39 +74,23 @@ export class Client {
...
-  constructor({
-    config,
-    modelFactory,
-    services
-  }: ClientOptions = {}) {
+  constructor(options: ClientOptions = {}) {
     if (
       typeof window !== 'undefined' &&
       window.location.protocol !== 'https:' &&
       !window.location.origin.includes('localhost')
```
richburdon, line 81: window.locaiton.host === 'localhost'

packages/sdk/client/src/client.ts
```diff
@@ -226,11 +209,30 @@ export class Client {
...
+    const { fromHost, fromIFrame } = await import('./services');
+    const { Monitor } = await import('./diagnostics');
+    const { EchoProxy, createDefaultModelFactory } = await import('./echo');
+    const { HaloProxy } = await import('./halo');
+    const { MeshProxy } = await import('./mesh');
+
+    this._config = this._options.config ?? new Config();
+    // NOTE: Must currently match the host.
+    this._modelFactory = this._options.modelFactory ?? createDefaultModelFactory();
+    this._services = await (this._options.services ?? (isNode() ? fromHost(this._config) : fromIFrame(this._config)));
```
richburdon, line 221: templated to roll these all up into a well-formed (readonly properties set in constructor) ClientRuntime class that we create/destroy and test separately. So that Client.initialize just becomes an async factory.

```ts
async initialize() {
  const { HaloProxy } = await import('./halo');
  _runtime = new ClientRuntime({ HaloProxy, ... });
}
...
checkInitialized() {
  if (!this._runtime) {
    throw
  }
}
...
async destroy() {
  await this._runtime.destroy();
  this._runtime = undefined;
}
...
createSpace() {
  // Public accessors since well defined.
  return this._runtime.echo.createSpace();
}
```

packages/sdk/client/src/client.ts
wittjosiah: I think there was a specific reason for this...I think it was the kube localhost domains. Maybe this should be `endsWith`?

Issue #3797, opened 2023-08-02
[composer] Aug 02 QA bugs


zhenyasav:
bugs:
- [x] `current app name is ready to use offline` toast 
- [x] (staging) joining a space from a guest profile leaves the space permaloading
- [x] when opening the devce join dialog, the close button should not be automatically focused with tooltip showing
![Screenshot 2023-08-02 at 11 14 18 AM](https://github.com/dxos/dxos/assets/3398896/b27ed2a7-310b-4ade-9dbd-f0d791936c1f)
- [x] when opening a device join invitation url, shell panel opens with "an identity is required to continue" choice, it should proceed directly to the PIN instead
- [x] after completing the PIN when doing a device join, clicking next succeeds the host but does not dismiss the guest. guest app is broken after this event
![Screenshot 2023-08-02 at 11 17 03 AM](https://github.com/dxos/dxos/assets/3398896/1231213f-d353-4f53-bd9c-a1b66f6de8c7)

thure:
AFAIK all items are now either fixed or non-reproduceable in dev.

PR #3798, opened 2023-08-02
fix(react-shell): QA bugs

thure:
This PR addresses tasks from #3797:
- [x] “when opening a device join invitation url, shell panel opens with "an identity is required to continue" choice, it should proceed directly to the PIN instead”
- [x] Removes close button from non-alert OS dialogs

copilot:all


richburdon:
`join-machine` hyphenates since not capitalized class name.

PR #3805, opened 2023-08-03
feat: Make waitForSpace consistant across CLI; add user-facing errors; fix CLI reset

richburdon:
### Summary
🚀🗑️🗣️

Improved `reset` command in `devtools/cli` by adding verbosity option and removing duplicate paths.

> _We're sailing on the git-sea, me hearties, yo ho ho_
> _We need to reset our ship, but we don't want to go slow_
> _So we use the `--verbose` flag to see what we delete_
> _And we dedupe the paths to save some space and make it neat_

### Walkthrough
*  Remove duplicate paths from `reset` command by wrapping `paths` array in a `Set` constructor ([link](https://github.com/dxos/dxos/pull/3805/files?diff=unified&w=0#diff-5d00313c7beafef3c3b49e0ebe9be4c6cde9c9183b4232bd7f79e24f08e9fd8cL26-R36)).


packages/devtools/cli/src/commands/debug/generate.ts
```diff
@@ -0,0 +1,66 @@
...
+        // TODO(burdon): @type is undefined.
+        // TODO(burdon): @model is dxos:model/document.
+        space?.db.add(new Expando({ type: 'test', title: random.word() }));
+      }
+
+      const { objects } = space?.db.query({ type: 'test' });
+      if (objects.length) {
+        for (let i = 0; i < this.flags.mutations; i++) {
+          const object = random.element(objects);
+          object.title = random.word();
```
mykola-vrmchk, line 61: ECHO can batch mutations automatically, so use `await space.db.flush()` to make sure your mutation is committed

mykola-vrmchk (approved):

PR #3813, opened 2023-08-04
fix: chunk size limit

egorgripasov:
### Summary
🐛🚀🔧

Improved error handling and performance for the teleport muxing modules. Added chunking logic to the `framer` module to avoid sending large messages in one piece. Added logging and async fixes to the `balancer` module.

> _Balancer module_
> _logs errors, handles async_
> _better than `nextTick`_

### Walkthrough
* Import `log` module and use `log.catch` to handle errors in async functions ([link](https://github.com/dxos/dxos/pull/3813/files?diff=unified&w=0#diff-0e09ceb6178d7900eed47073b84ac4dca53d5a48851fba05440f6e8e08c681a7R6), [link](https://github.com/dxos/dxos/pull/3813/files?diff=unified&w=0#diff-f67b857af489bd83b17ccd9855ddde7a08c34df9a6485adb4cff8c5212cf5150L9-R19)).


packages/core/mesh/teleport/src/muxing/framer.ts
```diff
@@ -50,11 +62,19 @@ export class Framer {
     send: (message) => {
       // log('write', { len: message.length, frame: Buffer.from(message).toString('hex') })
       return new Promise<void>((resolve) => {
-        const canContinue = this._stream.push(encodeFrame(message));
-        if (!canContinue) {
-          this._responseQueue.push(resolve);
-        } else {
-          process.nextTick(resolve);
+        const isEmpty = this._chunkQueue.length === 0;
```
dmaretskyi, line 65: Is this happening after the balancer? I think it would be better to split up the chunks somewhere before or at the balancer level, so that one single big chunk doesnt block the entire stream

packages/core/mesh/teleport/src/muxing/framer.ts
```diff
@@ -82,6 +98,33 @@ export class Framer {
...
+    await this._processChunkQueue();
+  }
+
+  private async _sendData(data: Buffer) {
+    return new Promise<void>((resolve) => {
+      const canContinue = this._stream.push(data);
+      if (!canContinue) {
+        this._responseQueue.push(resolve);
+      } else {
+        process.nextTick(resolve);
```
dmaretskyi, line 119: is nextTick really needed here? I though resolving the promise would create a microtask anyway

packages/core/mesh/teleport/src/muxing/framer.ts
```diff
@@ -50,11 +62,19 @@ export class Framer {
...
+        const chunks = [];
+        for (let i = 0; i < data.length; i += MAX_CHUNK_SIZE) {
+          chunks.push(data.slice(i, i + MAX_CHUNK_SIZE));
+        }
+        chunks.forEach((chunk, index) => {
+          const callback = index === chunks.length - 1 ? resolve : undefined;
+          this._chunkQueue.push({ msg: chunk, callback });
+        });
+        if (isEmpty) {
+          this._processChunkQueue().catch(log.catch);
```
dmaretskyi, line 77: call the logger in a function call `log.catch(....)` this way the build metadata (filename, line number) would be automatically added during compilation

packages/core/mesh/teleport/src/muxing/framer.ts
```diff
@@ -18,6 +27,9 @@ export class Framer {
   private _messageCb?: (msg: Uint8Array) => void;
   private _subscribeCb?: () => void;
   private _buffer?: Buffer; // The rest of the bytes from the previous write call.
+
+  // TODO(egorgripasov): Will cause a memory leak if streams do not appreciate the backpressure.
+  private _chunkQueue: Chunk[] = [];
```
dmaretskyi, line 32: Add a warning for when the queued up chunks surpass a certain theshold (in bytes)

dmaretskyi:
Also could you ensure that the frames are reasseabled correctly with their bounds are preserved after they are split into chunks?

packages/core/mesh/teleport/src/muxing/framer.ts
egorgripasov: i was thinking about that as well; would it require adding some "envelope" to each chunk, and adding a way of bringing those chunks back together to a single message?
(It seems like currently framer would not work if chunks from different streams will be mixed together)
was wondering how much of overheat that could cause

PR #3846, opened 2023-08-10
docs(specs): `SessionGraph` proposal

thure:
This PR proposes `SessionGraph`

### DIFF
--- a/docs/docs/specs/v0.2/session-graph.md
+++ b/docs/docs/specs/v0.2/session-graph.md
@@ -9,7 +9,9 @@ We have identified a need for app plugins to be able to:
 2. present content items in relation to other content items,
 3. present actions related to a content item,
 4. persist mutations to a content item in the UI,
+
 in such a way that the solution is:
+
 1. independent of any method of persistence beyond the UA’s own JS runtime,
 2. itself able to be persisted or replicated (including sharing),
 3. reasonably extensible by plugins without also causing corruptions.
@@ -29,8 +31,8 @@ Nodes must support the following:
 1. A scalar identifier unique within the session
 2. A static or translatable label
 3. A serializable, but possibly arbitrary, record of properties that a plugin can recognize and use as a way to resolve an entity the Node represents
-	- This _may_ include details about provenance
-	- This _may_ include a graphical way to present the node (references to an icon or image)
+  - This _may_ include details about provenance
+  - This _may_ include a graphical way to present the node (references to an icon or image)
 
 ## Relations
 
@@ -82,7 +84,7 @@ type SessionGraph = {
         RelationKey,
         NodeKey | Set<NodeKey>
       >
-	>
+  >
   >;
 };
 ```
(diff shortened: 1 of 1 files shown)


docs/docs/specs/v0.2/session-graph.md
```diff
@@ -0,0 +1,292 @@
...
+
+#### Rendering lists by preferential relation
+
+```tsx
+
+const SeparatedLists = ({id, root}: {id: NodeKey, root: Node}) => {
+  const datum = useResolvedDatum(id);
+  const specialChildren = useRelatedNodes(id, 'navmenu-child');
+  const specialIds = new Set(specialChildren.keys());
+  const allChildren = useRelatedNodes(id, 'provenance-child');
```
wittjosiah, line 237: This is a little confusing to me, perhaps just slightly too abstract, I can't quite picture what each of these would contain to facilitate rendering the tree as expected.

docs/docs/specs/v0.2/session-graph.md
```diff
@@ -0,0 +1,292 @@
...
+> `MarkdownPlugin` can add nodes representing documents it finds and associate them with a node that represents itself as the documents’ provenance
+
+#### Adding content items
+
+```tsx
+const MARKDOWN_NS = 'dxos.org/markdown-plugin';
+
+const MARKDOWN_GROUP = `${MARKDOWN_PLUGIN}/documents`;
+
+const handleReady = async () => {
```
wittjosiah, line 110: Is this the general plugin ready callback or a graph-specific callback? Could this just be the graph `nodes` callback and return the set of nodes and relations?

docs/docs/specs/v0.2/session-graph.md
```diff
@@ -0,0 +1,292 @@
...
+    
+type Node = {
+  id: NodeKey;
+  label: string | [string, {ns: string}];
+  params?: Record<string, Params>;
+};
+
+type RelationType = string;
+
+type RelationKey = `${RelationType}-${ 'child' | 'parent' }`;
```
wittjosiah, line 74:
```suggestion
type RelationKey = string;
```

Probably no need to prescribe this, then it can be used to describe other things. (e.g. `create-action` relation to another node which represents the action)

docs/docs/specs/v0.2/session-graph.md
```diff
@@ -0,0 +1,292 @@
...
+    <ul>
+      {childNodes.entries().map((childId, node)=>(
+        <ListItem id={childId} node={node}/>
+      ))}
+    </ul>
+  </>
+}
+```
+
+The hooks `useResolvedDatum` and `useRelatedNodes` leverage signals to propagate updates to the substrate entity or the node’s relations respectively.
```
wittjosiah, line 182: FYI we deprecated the use of "datum" in the api because we didn't think it was useful to try to distinguish between singular and plural (e.g. if it's an array it could be either depending on how you look at it)

wittjosiah, line 182: How are you expecting the data to be resolved from a node if it's not directly referenced by the node? Will the plugin need to provide a resolver?

docs/docs/specs/v0.2/session-graph.md
thure: You mean the implementation of the hooks would contain, or what these hooks are returning?

If the former, I want to illustrate that. If the latter, they’re returning `Map<NodeKey, Node>`

docs/docs/specs/v0.2/session-graph.md
thure: I left that to the reader to decide since I wasn't sure myself 😅 — wherever a plugin should produce nodes and relations from scratch, this is that function.

I will illustrate what plugins should do to keep the nodes updated.

docs/docs/specs/v0.2/session-graph.md
thure: Sure, that's fair :)

docs/docs/specs/v0.2/session-graph.md
thure: Okay fair, I'll replace 'datum' with 'data'.

docs/docs/specs/v0.2/session-graph.md
wittjosiah: I mean more just a concrete example, like an example set of nodes and what would be returns if their ids were passed in.

docs/docs/specs/v0.2/session-graph.md
thure: Yes, was planning on illustrating that next.

docs/docs/specs/v0.2/session-graph.md
thure: `useRelatedNodes` takes as arguments the one _subject Node’s_ `NodeKey`, then a `RelationKey`. This use-case gets ids from `specialChildren.keys()` so it can do set math.

Is there something specific you have in mind you’re wondering about?

PR #3849, opened 2023-08-11
docs: graph API proposal

zhenyasav:
### Summary
📝📊🧩

Added a new document `graph.md` to propose a graph API for plugins. The document explains the rationale, design, and examples of the graph API.

> _`graph.md` proposes_
> _a new API for plugins_
> _cutting edge design_

### Walkthrough
*  Add a new document `graph.md` to propose a graph API for plugins ([link](https://github.com/dxos/dxos/pull/3849/files?diff=unified&w=0#diff-931b6fd1ec98e3975e577ea341c9aab6e7ca168c47595fd788cef1df94eff7f9L1-R83))

### DIFF
--- a/docs/docs/specs/v0.2/graph.md
+++ b/docs/docs/specs/v0.2/graph.md
@@ -1,0 +1,84 @@
+# Graph API
+a proposal for `plugin.provides.graph`
+
+```tsx
+export type Node<T = any> = {
+  id: string;
+  label: string;
+  data: T;
+};
+
+const toNode = (e: any) => ({});
+
+export const Plugin = () => {
+  return {
+    provides: {
+      graph: {
+        nodes: (parent: Node, publish: Function) => {
+          const space = parent.data;
+          if (!(space instanceof Object)) return;
+          const query = space.db.query();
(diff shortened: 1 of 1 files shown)


docs/docs/specs/v0.2/graph.md
```diff
@@ -0,0 +1,122 @@
+# Graph API
+a proposal for `plugin.provides.graph`
+
+```tsx
+// let's say nodes are like this
+export type Node<T = any> = {
+  id: string;
+  label: string;
+  data: T;
```
thure, line 9: Do we anticipate it could cause perf issues if we resolve _all the data_ the UI could possibly display on init?

docs/docs/specs/v0.2/graph.md
```diff
@@ -0,0 +1,122 @@
...
+  add(e: any): any;
+  remove(e: any): any;
+}
+
+export const PluginNesting2 = () => {
+  return {
+    provides: {
+      graph: {
+        nodes: (parent: Node2) => {
+          const space = parent.data;
```
thure, line 73: So, this is still hierarchical? Is there a way to get a node without referring to all of its parents, or traversing the tree to look for it?

PR #3850, opened 2023-08-11
fix: Fix diagnostics

mykola-vrmchk:
### Summary
:truck::alarm_clock::zap:

Improved the reliability and performance of the diagnostics module by adding timeouts and simplifying the platform detection logic. Moved the `getPlatform` function to the `diagnostics.ts` file to avoid unnecessary exports.

> _To improve the diagnostics module_
> _We moved `getPlatform` to its cubicle_
> _We added some timeouts_
> _To handle the dropouts_
> _And used `asyncTimeout` to be suitable_

### Walkthrough
* Import `asyncTimeout` function from `@dxos/async` to set timeouts for async operations in diagnostics module ([link](https://github.com/dxos/dxos/pull/3850/files?diff=unified&w=0#diff-75743e11fcfd6e8a341512a35696a42d7b1a91a6f660f803051a70a164c20655L7-R7)).

### DIFF
--- a/packages/sdk/client/src/diagnostics/diagnostics.ts
+++ b/packages/sdk/client/src/diagnostics/diagnostics.ts
@@ -4,7 +4,7 @@
 
 import invariant from 'tiny-invariant';
 
-import { Trigger } from '@dxos/async';
+import { asyncTimeout, Trigger } from '@dxos/async';
 import { Space } from '@dxos/client-protocol';
 import { ConfigProto } from '@dxos/config';
 import { PublicKey } from '@dxos/keys';
@@ -24,6 +24,8 @@ import { Timeframe } from '@dxos/timeframe';
 import { Client } from '../client';
 import { getPlatform, Platform } from './platform';
 import { jsonStringify, JsonStringifyOptions } from './util';
+
+const DEFAULT_DIAGNOSTICS_TIMEOUT = 1000;
 
 export type Diagnostics = {
   created: string;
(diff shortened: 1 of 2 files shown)


mykola-vrmchk:
* main
  * **PR #3850** <a href="https://app.graphite.dev/github/pr/dxos/dxos/3850" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/3850?utm_source=stack-comment).

packages/sdk/client/src/diagnostics/diagnostics.ts
```diff
@@ -91,7 +93,10 @@ export const createDiagnostics = async (client: Client, options: DiagnosticOptio
     const stream = client.services.services.LoggingService!.queryMetrics({});
     const trigger = new Trigger<Metrics>();
     stream?.subscribe(async (metrics) => trigger.wake(metrics!));
-    data.metrics = await trigger.wait();
+    data.metrics = await asyncTimeout(trigger.wait(), DEFAULT_DIAGNOSTICS_TIMEOUT).catch((err) => {
+      log.warn('Metics take to long to query.');
```
richburdon, line 97: log.catch(err) or log.warn(err.message).
i.e., no need to "reword" error messages like this (the original error message is good and tells the user about the timeout).
OR just let throw since oclifs error handler should catch it.

packages/sdk/client/src/diagnostics/diagnostics.ts
```diff
@@ -25,6 +25,8 @@ import { Client } from '../client';
 import { getPlatform, Platform } from './platform';
 import { jsonStringify, JsonStringifyOptions } from './util';
 
+const DEFAULT_DIAGNOSTICS_TIMEOUT = 1000;
```
richburdon, line 28: too short. and always try to declare consts in the function's options property as a default that the caller (e.g., CLI) can override.

PR #3871, opened 2023-08-14
refactor(plugin-graph): propose graph new interface

wittjosiah:
### Summary
🧪🌳📝

This pull request adds a new graph implementation called SessionGraph, which manages a hierarchical structure of nodes and actions. It also adds the type definitions for the Graph interface and its related types. The pull request includes a test file for the SessionGraph class, using the @dxos/test and chai libraries. The changes are part of the plugin-graph package, which provides graph functionality for the dxos apps.

> _`SessionGraph` tests_
> _Building a tree of nodes_
> _Autumn leaves falling_

### Walkthrough
*  Implement the SessionGraph class to manage a hierarchical structure of nodes and actions ([link](https://github.com/dxos/dxos/pull/3871/files?diff=unified&w=0#diff-95c66a49579b77fde184619153f6649453cf5cd813760c61050fb6cc75a638b1R1-R83))
* Add the type definitions for the Graph interface and its related types ([link](https://github.com/dxos/dxos/pull/3871/files?diff=unified&w=0#diff-c3627baa8917066852bedbbedcc8aa28433af0b953d1713a2042d1398ab1be0eR1-R37))
* Add a test file for the SessionGraph class using @dxos/test and chai ([link](https://github.com/dxos/dxos/pull/3871/files?diff=unified&w=0#diff-297703948494cd3f8703ab7810c73b83133a88ff3b723ec6a4efc8242782602fR1-R205))


packages/apps/plugins/plugin-graph/src/graph/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    removeProperty(key: string): void;
+  };
+
+  export type Action<TProperties extends { [key: string]: any } = { [key: string]: any }> = {
+    id: string;
+    label: string | [string, { ns: string; count?: number }];
+    icon?: FC<IconProps>;
+    intent?: Intent | Intent[];
+    properties: TProperties;
+    actionsMap: { [key: string]: Action };
```
richburdon, line 61: Subactions?

packages/apps/plugins/plugin-graph/src/graph/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    get actions(): Action[];
+    invoke: () => Promise<any>;
+    add<TActionProperties extends { [key: string]: any } = { [key: string]: any }>(
+      ...action: (Pick<Action, 'id' | 'label'> & Partial<Action<TActionProperties>>)[]
+    ): Action<TActionProperties>[];
+    remove(id: string): Action;
+    addProperty(key: string, value: any): void;
```
richburdon, line 68: Why do actions have properties?

packages/apps/plugins/plugin-graph/src/types.ts
```diff
@@ -2,60 +2,20 @@
...
-
-export type GraphNode<TData = any> = {
-  id: string;
-  index: Index;
-  label: string | [string, { ns: string; count?: number }];
-  description?: string;
-  icon?: FC;
-  data?: TData;
-  parent?: GraphNode;
-  onChildrenRearrange?: (child: GraphNode, nextIndex: Index) => void;
```
richburdon, line 25: does this become an action?

packages/apps/plugins/plugin-graph/src/graph/graph.ts
```diff
@@ -0,0 +1,189 @@
...
+
+import { RevertDeepSignal, deepSignal } from 'deepsignal/react';
+import get from 'lodash.get';
+
+import { SendIntent } from '@braneframe/plugin-intent';
+import { EventSubscriptions } from '@dxos/async';
+
+import { Graph } from './types';
+
+export class SessionGraph implements Graph {
```
richburdon, line 13: Why SessionGraph?

packages/apps/plugins/plugin-graph/src/graph/graph.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    if (direction === 'down') {
+      Object.values(from.children).forEach((child) => this.traverse({ from: child, predicate, onVisitNode }));
+    } else if (direction === 'up' && from.parent) {
+      this.traverse({ from: from.parent, direction, predicate, onVisitNode });
+    }
+  }
+
+  registerNodeBuilder(builder: Graph.NodeBuilder): void {
```
richburdon, line 47: I might try to keep the builders outside of the graph so that the graph itself is "inert" and keep it as simple as possible (because it is likely going to get more and more complex).

pseudocode:

```ts
const builder = new GraphBuilder().addNodeBuilder().addNodeBuilder();
const graph = builder.build();
```

packages/apps/plugins/plugin-graph/src/graph/graph.ts
```diff
@@ -0,0 +1,189 @@
...
+      actionsMap: {},
+      ...partial,
+      // TODO(wittjosiah): Default sort.
+      get children() {
+        return Object.values(node.childrenMap);
+      },
+      get actions() {
+        return Object.values(node.actionsMap);
+      },
+      add: (...partials) => {
```
richburdon, line 105: Since Node isn't pure (has functions) then make it an actual class rather than a struct with functions attached?
Is this a deepsignal concern?

packages/apps/plugins/plugin-graph/src/graph/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  export type NodeBuilder = (parent: Node) => UnsubscribeCallback | void;
+
+  export type TraverseOptions = {
+    from?: Node;
+    direction?: 'up' | 'down';
+    predicate?: (node: Node) => boolean;
+    onVisitNode?: (node: Node) => void;
+  };
+
+  export type Node<TData = any, TProperties extends { [key: string]: any } = { [key: string]: any }> = {
```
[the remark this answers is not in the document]
wittjosiah: for what purpose would we want the types to be separate?

packages/apps/plugins/plugin-graph/src/graph/types.ts
wittjosiah: yeah, makes sense `subactions` and `subactionsMap` :+1:

packages/apps/plugins/plugin-graph/src/graph/types.ts
wittjosiah: we thinned out the first class properties so things like `index` or `disposition` or `disabled` would all be captured as properties

packages/apps/plugins/plugin-graph/src/types.ts
wittjosiah: no because it's not a user-facing thing, it's internal implementation details, it fell into `properties` along with the other excess stuff

packages/apps/plugins/plugin-graph/src/graph/graph.ts
wittjosiah: So...I started with it being separate, but the builders and the graph are quite intertwined. I think it might be doable to pull them apart if not for the fact that we filter the builders as you build up the graph so nodes aren't passed to builders which contributed to their ancestors. Maybe it's still possible? I'm open to looking at it again, but definitely a difficult one.

packages/apps/plugins/plugin-graph/src/graph/graph.ts
wittjosiah: Yeah, I wanted to take advantage of deepsignal turns getters into `computed` calls in signal. I'm not sure if that works within classes, it might? I haven't tried, but also having deepsignal venture down into classes is often when I've had the most trouble with it, so I've just generally tried to avoid that (which is part of why it's used _inside_ this class rather than dropping this class into a deepsignal).

packages/apps/plugins/plugin-graph/src/graph/types.ts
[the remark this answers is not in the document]
zhenyasav: Rich and I discussed this, It boiled down to having the ability to write tight utilities against the GraphNode interface without "bringing along all of it". for example, utilities can be like `iCanReorderAThing(node: IOrderable) {}` where `GraphNode = IOrderable & IHaveChildren & ...`. I think two things should obviate the need to do anything about this now:
1. we haven't actually felt any pain from using just plain `GraphNode` interface in all such cases (i.e. let's just write such utilities against the full-sized GraphNode and see if that hurts anywhere to perform this surgery.
2. any such utility can just declare locally the signature subset of GraphNode that it cares about (in case we don't get around to redeclaring GraphNode as an algebraic sum before we need this). I.e. `INeedOrderable(node: { properties: { index: number })` and regular `GraphNodes` will be type-compatible with this either way.

It also helps to clarify any concrete scenario we have in front of us to motivate changes like this.

packages/apps/plugins/plugin-graph/src/graph/types.ts
zhenyasav: actually that would eliminate the class of tools that know how to work on `{ actions: [] }` and could use either a `Graph.Node` or an `Graph.Action` without caring about which one it is.

packages/apps/plugins/plugin-graph/src/graph/graph.ts
zhenyasav: `GraphStore`

packages/apps/plugins/plugin-graph/src/graph/graph.ts
zhenyasav: concur, let's not make a change like this until we understand clearly the imminent end user scenario that would require that change.

packages/apps/plugins/plugin-graph/src/graph/graph.ts
zhenyasav: Actually there are plenty of reasons not to apply classes without good rationale. They are now a rather weathered feature, and there's plenty of criticism of the `class` keyword and the notorious `this` keyword out there, and while I don't consider them to be serious drawbacks I think it's enough to warrant some thought about whether we're applying classes because inheritance or some other "valuable" property of classes is what we need here (and it isn't better solved with something like "composition" etc). fwiw if it aint broke we dont need to fix it, and writing expressions inline can be helpful to both read and write code and build static analysis tooling against (something we'll actually likely need to do soon in the type-safety of ECHO stream). It's also less verbose than writing classes. Either way the space for problems where OOP patterns are required is shrinking.

https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/this%20&%20object%20prototypes/README.md#you-dont-know-js-this--object-prototypes
https://en.wikipedia.org/wiki/Composition_over_inheritance

packages/apps/plugins/plugin-graph/src/graph/types.ts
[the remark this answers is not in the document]
wittjosiah: yeah I can see how that would be useful given a motivating use case :+1:

PR #3932, opened 2023-08-21
feat: replication spec

egorgripasov:
### Summary
🔄📝🚀

Added a new test case for feed replication in `kube-testing` using custom extensions and dependencies. The test case uses a `ReplicationTestPlan` class that creates and manages the signal server, the peers, the swarms, and the feeds using various packages from `@dxos` and local modules. The test case also uses the `ReplicatorExtension` to enable feed replication across peers using the `TeleportExtension` interface. The test case is exported from `spec/index.ts` and can be imported in other modules. Some minor changes were made to the `test-builder.ts`, `test-wire-protocol.ts`, `env.ts`, and `tsconfig.json` files to support the new test case.

> _Sing, O Muse, of the mighty deeds of the DXOS engineers_
> _Who, with skill and cunning, forged the `kube-testing` package_
> _To test the network of peers and feeds, replicating data across the swarm_
> _Using `TeleportExtension` and `ReplicationTestPlan`, tools of wondrous power_

### Walkthrough
*  Add `ReplicationTestPlan` class for testing feed replication across peers ([link](https://github.com/dxos/dxos/pull/3932/files?diff=unified&w=0#diff-d20cae8a33ccff28ed95d2bca98785a1228fc4ec840f6906beb90a0d6b9a872dR1-R303), [link](https://github.com/dxos/dxos/pull/3932/files?diff=unified&w=0#diff-d98500a34d78c457efc6082d18760f1fb306798537768fa493f87cd4b594257eR6), [link](https://github.com/dxos/dxos/pull/3932/files?diff=unified&w=0#diff-07d9a4fe62c5ce84347caf3b0b32a13540ad7cb7704251c7230663f3d92d4958L8-R8), [link](https://github.com/dxos/dxos/pull/3932/files?diff=unified&w=0#diff-07d9a4fe62c5ce84347caf3b0b32a13540ad7cb7704251c7230663f3d92d4958R87-R111)).


packages/gravity/kube-testing/src/util.ts
```diff
@@ -12,3 +20,27 @@ export const randomArraySlice = <T>(array: T[], size: number) => {
   }
   return result;
 };
+
+export const generateKeyPair = async (): Promise<TestKeyPair> => {
```
richburdon, line 24: Use Keyring directly?

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,351 @@
...
+
+export type ReplicationAgentConfig = {
+  agentIdx: number;
+  signalUrl: string;
+  swarmTopicIds: string[];
+  feeds: Record<string, FeedConfig[]>;
+  feedKeys: TestKeyPair[];
+};
+
+export class ReplicationTestPlan implements TestPlan<ReplicationTestSpec, ReplicationAgentConfig> {
```
richburdon, line 54: Comment:
The purpose of this test is to...
Mention teleport, networkmanager, feedstore, etc.
What does it measure?

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,351 @@
...
+      await sleep(5_000);
+
+      // Write to wratable feeds in all swarms.
+      {
+        log.info('writing to writable feeds', { agentIdx });
+        await forEachSwarmAndAgent(async (swarmIdx, swarm, agentId) => {
+          const feedsArr = feeds.get(swarm.topic.toString())!;
+          for (const feedObj of feedsArr) {
+            if (feedObj.writable) {
+              for (let i = 0; i < 10; i++) {
```
richburdon, line 294: configurable?

packages/core/halo/keyring/src/keyring.ts
```diff
@@ -135,6 +135,42 @@ export class Keyring implements Signer {
     }
     return keys;
   }
+
+  /**
+   * A way to import a key pair (for testing purposes only).
+   */
+  async _importKeyPair(privateKeyData: JsonWebKey, publicKeyData: JsonWebKey): Promise<PublicKey> {
```
dmaretskyi, line 142: change this to `async impotKey(keyPair: CryptoKeyPair)` this way it could be used for other key types, not just jwk

dmaretskyi: hold on, is this used anywhere?

packages/gravity/kube-testing/src/spec/transport.ts
```diff
@@ -192,7 +106,15 @@ export class TransportTestPlan implements TestPlan<TransportTestSpec, TransportA
 
         await Promise.all(
           swarms.map(async (swarm, swarmIdx) => {
```
dmaretskyi, line 108: not sure if having multiple swarms is useful in this test, how is it different from just adding more feeds?

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,412 @@
...
+type FeedStats = {
+  feedKey: string;
+  feedLength: number;
+  byteLength: number;
+  replicas: FeedReplicaStats[];
+};
+
+export type ReplicationTestSpec = {
+  agents: number;
+  swarmsPerAgent: number;
```
dmaretskyi, line 54: not sure if having multiple swarms is useful in this tests, how is it different from just adding more feeds?

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,412 @@
...
+  duration: number;
+
+  transport: 'webrtc' | 'webrtc-proxy';
+
+  targetSwarmTimeout: number;
+  fullSwarmTimeout: number;
+
+  feedsPerSwarm: number;
+  feedLoadInterval: number;
```
dmaretskyi, line 63: feedAppendInterval, feedMessageSize

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,412 @@
...
+                    byteLength: otherFeedStats.byteLength,
+                    replicationSpeed:
+                      (bytesInSecond / Math.pow(1024, e)).toFixed(2) + ' ' + ' kMGTP'.charAt(e) + 'B/sec',
+                  });
+                }
+              }
+            });
+            stats.push({
+              feedKey: feedStats.feedKey,
+              feedLength: feedStats.feedLength,
```
dmaretskyi, line 401: is feedLength the total length of the feed or the number of mutations replicated?

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,412 @@
...
+              byteLength: feedStats.byteLength,
+              replicas,
+            });
+          }
+        });
+      });
+    });
+
+    log.info('replication stats', { stats });
```
dmaretskyi, line 410: also output the cumulative stats for a give agent x swarm

packages/gravity/kube-testing/src/spec/replication.ts
```diff
@@ -0,0 +1,412 @@
...
+
+  transport: 'webrtc' | 'webrtc-proxy';
+
+  targetSwarmTimeout: number;
+  fullSwarmTimeout: number;
+
+  feedsPerSwarm: number;
+  feedLoadInterval: number;
+  feedLoadChunkSize: number;
+  feedLoadDuration: number;
```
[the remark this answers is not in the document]
egorgripasov: just to double-check: the idea was to make it different from the unit tests we have, and replicate for specific amount of time, which looks more like a realworld. Then, when writing stops evaluate what part was replicated and with what speed;
We could change it by message number, if you think it makes more sense

packages/gravity/kube-testing/src/spec/replication.ts
egorgripasov: that's the num of mutation, while `byteLength` is the total length in bytes

packages/gravity/kube-testing/src/spec/replication.ts
egorgripasov: can you elaborate more on that?
as of now the stats already grouped by feed, e.g. each writable feed + it's replicas, something like
```
{
      feedKey: '045b13556f0857555575a49dec7be1aa2878bffc4db028f4c76c1a168f83acd194a8ce58221d9b61e9348a5e2fd9d29583c2d952f8aec77f9a3c46b3ddcdd13fe3',
      feedLength: 1972,
      byteLength: 197200000,
      replicas: [
        {
          feedLength: 1542,
          byteLength: 154200000,
          replicationSpeed: '15.42 MB/s'
        },
        {
          feedLength: 1825,
          byteLength: 182500000,
          replicationSpeed: '18.25 MB/s'
        }
      ]
    },
```
where replicas a re replicated feeds on different agents within the same swarm

packages/gravity/kube-testing/src/spec/replication.ts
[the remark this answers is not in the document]
egorgripasov: ok, i've added both ways, so in spec we could choose load for time interval vs load by num of messages

packages/core/halo/keyring/src/keyring.ts
egorgripasov: ok, converted to common import func

PR #3933, opened 2023-08-21
feat(client): update halo profile

wittjosiah:
### Summary
🆕🛠️📝

This pull request adds the ability to update the profile of an identity in the client services layer. It implements a new RPC method in the `Halo` service, a corresponding method in the `Halo` interface and proxy, and a method in the `IdentityService` and `IdentityManager` classes. It also updates the protocol definition, the keyring usage, and the tests for the identity-related classes.

> _We are the halo, we control the flow_
> _We update the profile, we sign with the key_
> _We are the identity, we shape the reality_
> _We test the service, we emit the event_

### Walkthrough
*  Add a new RPC method to the Halo service for updating the profile document of an identity ([link](https://github.com/dxos/dxos/pull/3933/files?diff=unified&w=0#diff-1f246514dbba8145aea2c94eadcc3265b76da0f434106ef87e58db92bdaf5aa0R115)).


dmaretskyi (approved):

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -193,6 +193,23 @@ export class IdentityManager {
     return identity;
   }
 
+  /**
+   * Update the profile document of an existing identity.
+   */
+  async updateProfile(params: ProfileDocument) {
```
richburdon, line 199: `params` => `profile`

packages/sdk/client-services/src/packlets/identity/identity-service.test.ts
```diff
@@ -24,7 +24,11 @@ describe('IdentityService', () => {
   beforeEach(async () => {
     serviceContext = createServiceContext();
     await serviceContext.open(new Context());
-    identityService = new IdentityServiceImpl(serviceContext);
+    identityService = new IdentityServiceImpl(
+      (params) => serviceContext.createIdentity(params),
```
richburdon, line 28: `params` => `profile`

PR #4037, opened 2023-08-30
fix: don't leave any dangling connections

egorgripasov:
### Summary
🛠️🔄📝

This pull request fixes a bug in the `Peer` class that caused connection errors and inconsistencies when peers tried to connect to each other at the same time. It adds a new property and updates the `closeConnection` method to handle displaced connections. It also adjusts the call to the `closeConnection` method in the `Swarm` class to pass the connection argument.

> _We are the swarm, we share the torrent's flame_
> _We face the doom of simultaneous connections_
> _We close the old, we embrace the new_
> _We store the `_displacedConnection` in our veins_

### Walkthrough
*  Add a new property `_displacedConnection` to store the previous connection of a peer when it is displaced by a new incoming connection ([link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355R72))
*  Change the logic of handling a displaced connection to avoid closing it twice and to call the `closeConnection` method with the displaced connection as an argument ([link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355L117-R120), [link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355L140-R143))
*  Modify the `closeConnection` method in `peer.ts` to accept a connection parameter and to check if it is the current or the displaced connection. Update the peer state and availability accordingly for both cases ([link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355L243), [link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355L253-R279), [link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355L298-R308))
*  Pass the `peer.connection` as an argument to the `closeConnection` method in `swarm.ts` to comply with the new signature ([link](https://github.com/dxos/dxos/pull/4037/files?diff=unified&w=0#diff-719498fb2d70860e5ed9738f14b6fe4a8e85ef40011b0f8e96eb1dfba9005b1fL381-R381))


packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -287,19 +292,19 @@ export class Peer {
...
   }
 
-  async closeConnection(err?: Error) {
-    if (!this.connection) {
+  async closeConnection(connection?: Connection, err?: Error) {
+    if (!connection) {
       return;
     }
-    const connection = this.connection;
+
```
richburdon, line 307: at end of function set this._connection to undefined.

packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -114,7 +115,9 @@ export class Peer {
 
         if (this.connection) {
           // Close our connection and accept remote peer's connection.
-          await this.closeConnection(new Error('Connection displaced by remote initiator.'));
+          this.displacedConnection = this.connection;
+          this.connection = undefined;
+          await this.closeConnection(this.displacedConnection, new Error('Connection displaced by remote initiator.'));
```
richburdon, line 120: to retain class open/close semantics is it possible to keep closeConnection as is (i.e., operate on this._connection) and keep track of the displaced connection separately?

richburdon: I can't make out why you're separately holding on to the "displacedConnection" is that just for testing? If so, that might be better to expose as an `Event`.

```
private readonly _connection = new Event<{ state, Connection >});
```

egorgripasov: yes, it's used mainly for testing atm - i just was thinking we might want to use it somehow during the diagnostics through the devtools or smth

packages/core/mesh/network-manager/src/swarm/peer.ts
egorgripasov: in that case, we won't be able to know if this was the main connection in onClose handler, around https://github.com/dxos/dxos/blob/4fe7b20df48e87ba51808783abf2ce18b617dec4/packages/core/mesh/network-manager/src/swarm/peer.ts#L255

packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -250,27 +252,30 @@ export class Peer {
             initiator,
           });
 
-          if (this._lastConnectionTime && this._lastConnectionTime + CONNECTION_COUNTS_STABLE_AFTER < Date.now()) {
-            // If we're closing the connection, and it has been connected for a while, reset the backoff.
-            this._availableAfter = 0;
-          } else {
-            this.availableToConnect = false;
-            this._availableAfter = increaseInterval(this._availableAfter);
+          if (this.connection === connection) {
```
dmaretskyi, line 255: This looks more as an invariant rather then a condition: given the comment below is there a situation where a new connection could be openned before this callback finishes processing?

packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -16,6 +16,12 @@ import { WireProtocolProvider } from '../wire-protocol';
 import { Connection, ConnectionState } from './connection';
 import { ConnectionLimiter } from './connection-limiter';
 
+export class ConnectionDisplacedError extends Error {
```
dmaretskyi, line 19:
```suggestion
import { SystemError } from '@dxos/errors'
export class ConnectionDisplacedError extends SystemError {
```

dmaretskyi (approved):

PR #4088, opened 2023-09-01
feat(echo-db): Commit batch after if user idling or batch length reaches the threshold

mykola-vrmchk:
### Summary
🐛🚀🧪

This pull request enhances the database proxy module in the `echo-db` and `echo-pipeline` packages by adding automatic batching of mutations and testing it. It also fixes a bug in the `plugin-debug` app that affected the SavingIndicator component.

> _To debug a plugin with ease_
> _They fixed a bug in `SavingIndicator`_
> _And added some tests for the `database-proxy`_
> _That batches mutations with a policy_
> _And emits an event for the monitor_

### Walkthrough
* Fix a bug in the SavingIndicator component of the plugin-debug app that showed an error state even when there was no error ([link](https://github.com/dxos/dxos/pull/4088/files?diff=unified&w=0#diff-c60b39f11e625fee534469a79622fa62ea629736147b47789e4014de184806cbL146-R151)).

### DIFF
--- a/packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
+++ b/packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
@@ -274,6 +274,8 @@ export class DatabaseProxy {
     }
     this._currentBatch = new Batch();
     this._currentBatch.clientTag = `${this._clientTagPrefix}:${this._clientTagCounter++}`;
+    this.pendingBatch.emit({ size: this._pendingBatches.size + 1 });
+
     return true;
   }
 
--- a/packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
+++ b/packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
@@ -20,7 +20,7 @@ import { ItemManager } from './item-manager';
 
 const FLUSH_TIMEOUT = 5_000;
 
-const BATCH_COMMIT_AFTER = 200;
+export const BATCH_COMMIT_AFTER = 200;
 /**
  * Maximum number of mutations in a batch.
  * Note: It is used only in auto created batches, if user creates/commits batch outside it will be ignored.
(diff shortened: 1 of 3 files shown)


mykola-vrmchk:
* main
  * **PR #4088** <a href="https://app.graphite.dev/github/pr/dxos/dxos/4088" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/4088?utm_source=stack-comment).

packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
```diff
@@ -20,7 +20,12 @@ import { ItemManager } from './item-manager';
 
 const FLUSH_TIMEOUT = 5_000;
 
-const BATCH_INTERVAL = 200;
+export const BATCH_COMMIT_AFTER = 200;
+/**
+ * Maximum number of mutations in a batch.
+ * Note: It is used only in auto created batches, if user creates/commits batch outside it will be ignored.
+ */
+const BATCH_SIZE_LIMIT = 64;
```
richburdon, line 28: always move consts into an FooOptions type/object that is passed into the constructor -- this allows us to push the options up the tree. And to configure tests, etc.

packages/core/echo/echo-pipeline/src/tests/database-unit.test.ts
```diff
@@ -4,8 +4,9 @@
 
 import expect from 'expect';
 
+import { sleep } from '@dxos/async';
```
richburdon, line 7: bad name -unit.test; make functional

packages/core/echo/echo-db/src/packlets/database/database-proxy.ts
mykola-vrmchk: Makes sense, thanks

Issue #3975, opened 2023-08-24
WebrtcProxyTransport connections closed by closing tabs seem to be stuck in CLOSING state for couple of seconds


dmaretskyi:
Repro:

1. Create a space with 2 peers.
2. Open devtools for peer 1.
3. Observer swarm connections pannel.
4. You should have 2 browser tabs open for peer 1: orignal labs tab (holds webrtc connection), devtools tab.
5. Close the labs tab. Webrtc conneciton should get closed.
6. Observe connection being stuck in CLOSING state.
7. After a while new connection gets opened from the devtools tab.

dmaretskyi:
![Kapture 2023-08-24 at 14 47 05](https://github.com/dxos/dxos/assets/35851437/f7cf2445-2dbf-42fb-adea-51db22a518f4)

nferch:
@dmaretskyi can repro on main (8d1cf7074), although it looks like you were on something else given that I don't have the label or close reason column?

dmaretskyi:
@nferch Yeah, I've added them to @richburdon 's branch https://github.com/dxos/dxos/pull/3972 . But feel free to cherry-pick those commits if you need them sooner

nferch:
@dmaretskyi don't need them, just wanted to confirm that I shouldn't have been running some other branch that would have affected the reproduction of the issue.

nferch:
@dmaretskyi looks like the issue is due to each Extension close timing out, I guess we need some mechanism to propagate that the nature of the close so it doesn't wait for the bye response (why we do this at all?). when I comment out https://github.com/dxos/dxos/blob/1d5e9f952e75f2027e70eeef09cb29d963c7b160/packages/core/mesh/rpc/src/rpc.ts#L202 it will reconnect straight away.

nferch:
https://github.com/dxos/dxos/pull/4101 seems to fix this particular case, although I saw some other issues, like when one peer doesn't notice that the other has disconnected and still thinks it is CONNECTED (due to the underlying connection being UDP?)

PR #4101, opened 2023-09-02
fix: abort abruptly disconnected connections

nferch:
closes https://github.com/dxos/dxos/issues/3975

copilot:all

### DIFF
--- a/packages/core/echo/echo-pipeline/src/space/space-protocol.ts
+++ b/packages/core/echo/echo-pipeline/src/space/space-protocol.ts
@@ -250,7 +250,12 @@ export class SpaceProtocolSession implements WireProtocol {
     this._teleport.addExtension('dxos.mesh.teleport.blobsync', this._blobSync.createExtension());
   }
 
+  // TODO(nf): why destroy and not close?
   async destroy(): Promise<void> {
     await this._teleport.close();
+  }
+
+  async abort(): Promise<void> {
+    await this._teleport.abort();
   }
 }
--- a/packages/core/mesh/network-manager/src/swarm/connection.ts
+++ b/packages/core/mesh/network-manager/src/swarm/connection.ts
@@ -168,13 +168,21 @@ export class Connection {
 
     this._transport.closed.once(() => {
       this._transport = undefined;
-      this.close().catch((err) => this.errors.raise(err));
+      log('abort triggered by transport close');
+      this.abort().catch((err) => this.errors.raise(err));
     });
 
     this._transport.errors.handle((err) => {
+      log('transport error:', { err });
+      // TODO(nf): collect a list of errors?
       if (!this.closeReason) {
         this.closeReason = err?.message;
       }
+      if (err?.message.includes('OperationError')) {
+        log('aborting due to transport OperationError');
+        this.abort().catch((err) => this.errors.raise(err));
+      }
+
       if (this._state !== ConnectionState.CLOSED && this._state !== ConnectionState.CLOSING) {
         this.errors.raise(err);
       }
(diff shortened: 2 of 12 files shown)


packages/core/echo/echo-pipeline/src/space/space-protocol.ts
```diff
@@ -250,7 +250,12 @@ export class SpaceProtocolSession implements WireProtocol {
     this._teleport.addExtension('dxos.mesh.teleport.blobsync', this._blobSync.createExtension());
   }
 
+  // TODO(nf): why destroy and not close?
   async destroy(): Promise<void> {
```
dmaretskyi, line 254: yeah, rename to close

dmaretskyi (approved):

packages/core/echo/echo-pipeline/src/space/space-protocol.ts
richburdon: If so then open/close (rename initialize).
But see createTeleportProtocolFactory below (if changing then change across the package).

packages/core/mesh/network-manager/src/swarm/connection.ts
```diff
@@ -190,6 +198,41 @@ export class Connection {
     log.trace('dxos.mesh.connection.open-connection', trace.end({ id: this._instanceId }));
   }
 
+  @synchronized
+  // TODO(nf): make the caller responsible for recording the reason and determining flow control.
+  async abort(err?: Error) {
+    log('abort');
```
richburdon, line 204: we generally do
`log('aborting...', { err })`
for things that might take time or fail.
you also have `aborting...` below.

packages/core/mesh/network-manager/src/swarm/connection.ts
```diff
@@ -190,6 +198,41 @@ export class Connection {
...
+      return;
+    }
+    // TODO(nf): different state?
+    this._changeState(ConnectionState.CLOSING);
+
+    await this._ctx.dispose();
+
+    try {
+      // Forcefully close the stream flushing any unsent data packets.
+      log('aborting... protocol  ', { proto: this._protocol });
```
richburdon, line 219: Edit: `'aborting protocol...'`?

packages/core/mesh/network-manager/src/swarm/connection.ts
```diff
@@ -255,7 +298,7 @@ export class Connection {
       }
 
       // If signal fails treat connection as failed
-      log.warn('Signal failed', { err });
+      log.warn('Signal failed, closing', { err });
```
richburdon, line 301: Nits:
- our log message standard is generally terse and lower case `log.warn('signal failed', { err })`
- this.close already logs "closing..."

packages/core/mesh/teleport/src/teleport.ts
```diff
@@ -122,6 +122,23 @@ export class Teleport {
...
+
+    for (const extension of this._extensions.values()) {
+      try {
+        await extension.onAbort(err);
+      } catch (err: any) {
+        log.catch(err);
+      }
+    }
+
+    this._muxer.destroy(err);
```
richburdon, line 139: await (if muxer.destroy isn't async it should be -- like all iniitialize/destroy/open/close methods).

packages/core/mesh/network-manager/src/swarm/connection.ts
nferch: makes sense. was mostly just getting some early feedback on the overall approach here, and roughly had added the debugging traces that i thought might be useful later on. but can see how attention to detail in these complicated and progressive flows makes it easier to reason about.

Issue #4112, opened 2023-09-04
iOS client cannot reconnect after sleep


nferch:
TBD: see https://discord.com/channels/837138313172353095/1147164225555419156

PR #4139, opened 2023-09-06
fix: connectivity and error handling

nferch:
* check for reentry and race conditions
* add custom errors to improve disconnection behavior and error reporting/diagnostics

### Summary
🐛🚨📊

This pull request improves the error handling and logging of the network layer and related components. It introduces new custom error classes in `@dxos/protocols` and uses them to wrap and provide more context for signaling, connection, and transport errors. It also fixes some bugs and typos in the connection and teleport logic.

> _`@dxos/protocols`_
> _More errors wrapped and raised_
> _Winter of bugs ends_

### Walkthrough
*  Add new error classes to `@dxos/protocols` for more specific and consistent error handling across the network layer ([link](https://github.com/dxos/dxos/pull/4139/files?diff=unified&w=0#diff-03a27f40927e8abc2d959ae98af8a2940bbba33c950f0a803c86e1572b580aeaR81-R126)).


nferch:
address some failure modes discovered when troubleshooting https://github.com/dxos/dxos/issues/4112

packages/core/protocols/src/errors.ts
```diff
@@ -78,3 +78,49 @@ export class RpcNotOpenError extends Error {
     super('RPC has not been opened.');
   }
 }
+
+export class WrappableError extends Error {
```
wittjosiah, line 82: FYI that I moved all the `@dxos/errors` here https://github.com/dxos/dxos/pull/4138/files#diff-1f72a0497a21cbb885fdee03300127134c6b2002ea75430e83a15d79805e0ee4 as well in order to be able to have them bridge the RPC boundary. We'll likely want to reconcile the error classes when merging.

wittjosiah (commented):
Is this intended to help mitigate the mobile connectivity issues? Should I test this out locally with my reproduction steps?

packages/core/protocols/src/errors.ts
nferch: yeah, was uncertain how many of the protocol errors would be external-facing, but it's probably simpler to just have them all together.

btw i saw the `this.name` pattern suggested but it didn't seem to work with `ErrorStream` either, so back to string matching in https://github.com/dxos/dxos/pull/4139/commits/4edac01a97201893aa86bddbf97c3f7ac425a03e#diff-8fd5aaf7614ff1002b55a9d55308ab71bf42144d7246e3c8d68852c93cfc8745R185

packages/core/mesh/messaging/src/messenger.ts
```diff
@@ -133,6 +144,31 @@ export class Messenger {
...
+
+  private async rebootNetwork() {
+    if (this._rebooting) return;
+    this._rebooting = true;
+
+    log('rebooting Messenger/SignalManager');
+    this.close();
+    this._signalManager.close();
+    log('pausing');
+    await new Promise((f) => setTimeout(f, NETWORK_REBOOT_DELAY));
```
dmaretskyi, line 162: await sleep(NETWORK_REBOOT_DELAY)

packages/core/mesh/teleport-extension-gossip/src/gossip-extension.ts
```diff
@@ -69,7 +69,14 @@ export class GossipExtension implements TeleportExtension {
 
   async onAbort(err?: Error): Promise<void> {
     log('abort', { err });
-    await this._rpc?.abort();
+    try {
+      await this._rpc?.abort();
+      this._sendInterval && clearInterval(this._sendInterval);
+      this.onClose(err);
```
dmaretskyi, line 75: await this._callbacks.onClose?.(err);

dmaretskyi: not awaiting async function is an antipattern

packages/core/mesh/teleport-extension-gossip/src/gossip-extension.ts
```diff
@@ -78,7 +85,14 @@ export class GossipExtension implements TeleportExtension {
     }
     await this._opened.wait();
     invariant(this._rpc, 'RPC not initialized');
-    await this._rpc.rpc.GossipService.announce(message);
+    try {
+      await this._rpc.rpc.GossipService.announce(message);
+    } catch (err) {
+      // TODO(nf): always abort on RpcClosedError? gets stuck in retry loop otherwise
+      if (err instanceof RpcClosedError) {
```
dmaretskyi, line 92: check if this is still needed after the change above

packages/core/mesh/messaging/src/messenger.ts
```diff
@@ -133,6 +144,31 @@ export class Messenger {
...
   }
 
+  private async errorLimitCheck() {
+    log(`checking error limit ${this._errorCount}`);
+    if (this._errorCount++ > ERROR_LIMIT) {
+      this.rebootNetwork();
+    }
+  }
+
+  private async rebootNetwork() {
```
dmaretskyi, line 154: add a todo to move this into signal client and possible make it react to browser events (on sleep?) instead of errors

packages/core/protocols/src/errors.ts
nferch: @wittjosiah thanks for that heads up, moved my errors to `packages/sdk/errors/src/errors.ts` and use `BaseError`

packages/core/mesh/messaging/src/messenger.ts
nferch: where does sleep come from?

packages/core/protocols/src/errors.ts
nferch: after using this for a bit it feels a bit unnatural to have to import a separate package just for errors, i think it'd feel more natural as a consumer to import the errors from the same package of the code that produces them.

packages/core/mesh/teleport-extension-gossip/src/gossip-extension.ts
nferch: didn't, so i reverted this bit

packages/core/protocols/src/errors.ts
wittjosiah: @nferch on the PR above I moved all the errors to be defined in `@dxos/protocols` and they're all re-exported by `@dxos/client` (which is where I imagine most people would encounter them)

nferch: @wittjosiah makes sense, so just more of a personal/internal concern then.

PR #4151, opened 2023-09-07
refactor(plugin-space): rename on enter

wittjosiah:
### Summary
🔧🔄📝

Refactored and improved space renaming UI, space property subscription logic, and space proxy logging and reactivity. `PopoverRenameSpace` now uses hooks and local state, `SpacePlugin` now uses `subscribe` instead of `createSubscription`, and `SpaceProxy` now emits an event and logs a message for property document availability and caching.

> _To rename a space with a click_
> _`PopoverRenameSpace` got a fix_
> _With hooks and a state_
> _And keyboard and mouse traits_
> _It's now a much smoother UX_

### Walkthrough
*  Refactor `PopoverRenameSpace` component to use hooks and handle renaming action ([link](https://github.com/dxos/dxos/pull/4151/files?diff=unified&w=0#diff-9bcdd47f0bdbafe2a8cfa2ba2e682347c8a55bb8def0d0e3026996b62cf0bcb9L5-R5),[link](https://github.com/dxos/dxos/pull/4151/files?diff=unified&w=0#diff-9bcdd47f0bdbafe2a8cfa2ba2e682347c8a55bb8def0d0e3026996b62cf0bcb9R14-R20),[link](https://github.com/dxos/dxos/pull/4151/files?diff=unified&w=0#diff-9bcdd47f0bdbafe2a8cfa2ba2e682347c8a55bb8def0d0e3026996b62cf0bcb9L23-R38)).

### DIFF
--- a/packages/apps/plugins/plugin-space/src/components/PopoverRenameSpace.tsx
+++ b/packages/apps/plugins/plugin-space/src/components/PopoverRenameSpace.tsx
@@ -2,7 +2,7 @@
 // Copyright 2023 DXOS.org
 //
 
-import React from 'react';
+import React, { useCallback, useRef, useState } from 'react';
 
 import { Button, Input, Popover, useTranslation } from '@dxos/aurora';
 import { Space } from '@dxos/react-client/echo';
@@ -11,6 +11,13 @@ import { SPACE_PLUGIN } from '../types';
 
 export const PopoverRenameSpace = ({ data: [_, space] }: { data: [string, Space] }) => {
   const { t } = useTranslation(SPACE_PLUGIN);
+  const doneButton = useRef<HTMLButtonElement>(null);
+  const [name, setName] = useState(space.properties.name ?? '');
+
+  const handleDone = useCallback(() => {
+    space.properties.name = name;
+  }, [space, name]);
+
   // todo(thure): Why does the input value need to be uncontrolled to work?
   return (
     <div role='none' className='p-1 flex gap-2'>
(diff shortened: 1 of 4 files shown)


packages/sdk/client/src/echo/space-proxy.ts
```diff
@@ -131,6 +131,7 @@ export class SpaceProxy implements Space {
     if (this._properties) {
       return this._properties;
     } else {
+      log('Using cached properties');
```
richburdon, line 134: nit: i'm not strongly defending this, but the logging pattern that has emerged is all lower case with an optional object. vs. "Errors which are english like sentences with a period."

packages/apps/plugins/plugin-space/src/components/PopoverRenameSpace.tsx
```diff
@@ -20,12 +27,15 @@ export const PopoverRenameSpace = ({ data: [_, space] }: { data: [string, Space]
           <Input.TextInput
             defaultValue={space.properties.name ?? ''}
             placeholder={t('untitled space title')}
-            onChange={({ target: { value } }) => (space.properties.name = value)}
+            onChange={({ target: { value } }) => setName(value)}
+            onKeyDown={({ key }) => key === 'Enter' && doneButton.current?.click()}
```
richburdon, line 31: i think you have to put a TODO here saying this is a temporary work around.

richburdon: can we expose a context/hook for the popover to expose a close method?

PR #4167, opened 2023-09-08
feat: Aurora ComboBox

richburdon:
https://github.com/dxos/dxos/assets/3523355/eabf1928-114a-4cdc-9f6b-ebb7fb7cc3ee

### Summary
🎁🎨🐛

Added a new `Selector` component and fixed some linting and storybook issues. The `Selector` component is a typeahead selector that uses the `downshift` library and can be used for selecting items from a list. The component was added in the `packages/ui/aurora/src/components/Selector` folder and has storybook examples in the `Selector.stories.tsx` file. The pull request also fixed a linting error in the `List.stories.tsx` file and changed the order of some imports in various storybook files to follow a consistent style. The `downshift` dependency was added to the `package.json` file.

> _A new component called `Selector`_
> _Was added with `downshift` as a helper_
> _The imports were sorted_
> _The stories reported_
> _And a linting error was a wrecker_

### Walkthrough
*  Add a new `Selector` component that is a typeahead selector using the `downshift` library ([link](https://github.com/dxos/dxos/pull/4167/files?diff=unified&w=0#diff-c16bcfb697c04e756c25bf73637661ab89ab7039e60acc7d363818642aae41aeR1-R68), [link](https://github.com/dxos/dxos/pull/4167/files?diff=unified&w=0#diff-e3d133c40923437562294315f786d6e49986e62fd9fac4a85680d7d86be446b3R1-R92)).


packages/ui/aurora/package.json
```diff
@@ -40,6 +41,7 @@
     "@radix-ui/react-toolbar": "^1.0.3",
     "@radix-ui/react-tooltip": "^1.0.5",
     "@radix-ui/react-use-controllable-state": "^1.0.0",
+    "downshift": "^8.1.0",
```
thure, line 44: In principle we create a separate package when a component has special dependencies like this.

packages/ui/aurora-theme/src/styles/components/index.ts
```diff
@@ -15,6 +15,7 @@ export * from './message';
 export * from './popover';
 export * from './scroll-area';
 export * from './select';
+export * from './selector';
```
thure, line 18: Rename the component to `Combobox` (directory would ofc be `combobox`).

packages/ui/aurora/src/components/Selector/Selector.tsx
```diff
@@ -0,0 +1,205 @@
...
+                  item,
+                  className: tx('selector.item', 'selector__item', {}, classNames),
+                })}
+              >
+                {adapter(item).text}
+              </li>
+            ))}
+          </ul>
+        </PopperPrimitive.Content>
+      </div>
```
thure, line 191: This will need to be refactored to export subcomponents to match ontologies with the rest of aurora.

thure (commented):
Made some non-blocking recommendations, but we'll need to follow up on them.

packages/ui/aurora/package.json
thure: Or, this should be a primitive first, then themed in Aurora.

packages/ui/aurora/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,200 @@
...
+      onInputChange?: (text?: string) => void;
+    }>
+  >
+>;
+
+/**
+ * Type-ahead combobox.
+ * https://www.downshift-js.com
+ */
+const ComboBoxRoot = <T,>({
```
thure, line 49: I haven’t seen generics used this way in the Radix ecosystem. I'm not sure we can expect developers will make use of this, and it could cause trouble if it has no defaults.

thure: Maybe it becomes irrelevant with subcomponents, since this appears to govern the schema of `items`, which should instead be the props of some `ComboboxItem` component.

packages/ui/aurora/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+const ComboBoxAnchor = forwardRef<HTMLDivElement, ComboBoxAnchorProps>(
+  ({ __scopeComboBox, classNames, placeholder }: ComboBoxAnchorProps, forwardedRef) => {
+    const { tx } = useThemeContext();
+    const popperScope = usePopperScope(__scopeComboBox);
+    const { getInputProps, getToggleButtonProps, isOpen } = useComboBoxContext(ANCHOR_NAME, __scopeComboBox);
+
+    return (
+      <PopperPrimitive.Anchor asChild {...popperScope} ref={forwardedRef}>
+        <div className='flex items-center gap-1'>
```
thure, line 106: `flex` should almost never be default, and classnames mustn’t be baked-in like this, this needs to be refactored to use `tx`.

packages/ui/aurora/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,200 @@
...
+      CONTENT_NAME,
+      __scopeComboBox,
+    );
+
+    if (!isOpen) {
+      return null;
+    }
+
+    return (
+      <div>
```
thure, line 152: What purpose does this extra `div` serve? Can it be eliminated? If not, it needs to support `asChild` and needs a role.

thure (changes requested):
There are just a couple issues we'd need to follow-up on ASAP.

packages/ui/aurora/src/components/ComboBox/ComboBox.tsx
richburdon: It's actually needed for the ValueAdapter.

packages/ui/aurora/src/components/ComboBox/ComboBox.tsx
richburdon: scrolling didn't work without it.

PR #4233, opened 2023-09-13
fix: ignore offer if connection not initial

nferch:
This should avoid the misleading "Connection displaced by remote initiator." errors from closed/aborted connections.

### Summary
🚫📞🔄

Prevent multiple connection offers from interfering with peer state. Add a condition to `handleOffer` in `peer.ts` to reject offers if the peer is already busy.

> _`handleOffer` checks_
> _if peer is busy or not_
> _rejects in autumn_

### Walkthrough
*  Add condition to reject connection offers if peer is busy ([link](https://github.com/dxos/dxos/pull/4233/files?diff=unified&w=0#diff-47bc81126b7b4f9732f7ae901368cd7eeada3d95eb37e7e693e1a1b3388be355R111-R114))

### DIFF
--- a/packages/core/mesh/network-manager/src/swarm/peer.ts
+++ b/packages/core/mesh/network-manager/src/swarm/peer.ts
@@ -108,6 +108,10 @@ export class Peer {
   async onOffer(message: OfferMessage): Promise<Answer> {
     const remoteId = message.author;
 
+    if (this.connection && this.connection.state !== ConnectionState.INITIAL) {
+      log.info(`received offer when connection already in ${this.connection.state} state`);
+      return { accept: false };
+    }
     // Check if we are already trying to connect to that peer.
     if (this.connection || this.initiating) {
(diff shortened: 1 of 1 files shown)


packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -108,6 +108,10 @@ export class Peer {
   async onOffer(message: OfferMessage): Promise<Answer> {
     const remoteId = message.author;
 
+    if (this.connection && this.connection.state !== ConnectionState.INITIAL) {
```
egorgripasov, line 111: if we ever "bring back" connection limiter, then connection's first state is actually `CREATED`, which gets switched to `INITIAL` after it went through the limiter, so maybe it should be something like
```
if (this.connection && ![ConnectionState.CREATED, ConnectionState.INITIAL].includes(this.connection.state)) {
```

Issue #4226, opened 2023-09-13
[agent] Running agent for sufficient amount of time cause Gossip error


mykola-vrmchk:
## Setup
* Agent running in the foreground
* Labs joins space that agent created
* Labs debug plugin creates mutation each 500 ms
* Run it for ~1 hour

## Failure
Agent (read Client) errors with 
![Image](https://github.com/dxos/dxos/assets/36420699/1c179f97-a888-46ea-bab2-007dd1dc22f2)
after that error replication with the agent stops

mykola-vrmchk:
cc @nferch

PR #4234, opened 2023-09-13
fix: downgrade RpcClosedError severity for gossip

nferch:
should help with https://github.com/dxos/dxos/issues/4226

### Summary
📦🛠️🙌

Improved error handling for `Gossip` class in `teleport-extension-gossip` package. Used `RpcClosedError` to detect and handle RPC channel closures in `sendAnnounce` method.

> _`RpcClosedError`_
> _Gives `Gossip` more resilience_
> _Autumn leaves falling_

### Walkthrough
* Import `RpcClosedError` class from `@dxos/protocols` package to handle RPC channel closure errors ([link](https://github.com/dxos/dxos/pull/4234/files?diff=unified&w=0#diff-8c0d22e91ef78cf597ee52fe4f8f36f1608d2baad932a27bf9d7f097e805101aR13)).

### DIFF
--- a/packages/core/mesh/teleport-extension-gossip/src/gossip.ts
+++ b/packages/core/mesh/teleport-extension-gossip/src/gossip.ts
@@ -111,7 +111,7 @@ export class Gossip {
         .catch(async (err) => {
           // TODO(nf): handle here or in the Extension?
           if (err instanceof RpcClosedError) {
-            log.warn('sendAnnounce failed because of RpcClosedError', { err });
+            log('sendAnnounce failed because of RpcClosedError', { err });
           } else {
             log.catch(err);
           }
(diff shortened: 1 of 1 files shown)


egorgripasov (approved):

packages/core/mesh/teleport-extension-gossip/src/gossip.ts
```diff
@@ -108,7 +109,12 @@ export class Gossip {
         })
 
         .catch(async (err) => {
-          log.catch(err);
+          // TODO(nf): handle here or in the Extension?
+          if (err instanceof RpcClosedError) {
+            log.warn('sendAnnounce failed because of RpcClosedError', { err });
```
dmaretskyi, line 114: I'd say we can even ignore this entirely. Gossip messaging does not guarntee reliability. This error just means that the connection was closed while we were waiting on an ACK for this message

nferch: that's kind of what i was thinking as well, was just reluctant to totally swallow it as it's useful (at least now) for identifying other connection issues.

is DEBUG a worthwhile compromise?

PR #4287, opened 2023-09-19
feat: Update expando to set meta/schema via options.

richburdon:
### Summary
🔧📝🔐

The pull request refactors the `TypedObject` class and its subclasses to use a consistent and simplified interface for creating and manipulating typed objects with schemas and meta fields. It also renames some properties and fixes some bugs in the code and tests that use the `TypedObject` class. The affected files include `schema.ts` snippets in the docs, the `plugin-debug` and `plugin-markdown` packages, the `echo-schema`, `echo-typegen`, and `kai-functions` packages, and the `serializer.test.ts` file.

> _We are the `TypedObject`s, we have the power of schemas_
> _We use the `__meta` field to store our secrets and keys_
> _We refactor and simplify, we fix the bugs and errors_
> _We are the `TypedObject`s, we are the masters of the code_

### Walkthrough
*  Refactor the way of passing schema type information to the `TypedObject` base class using a single `opts` parameter with a `schema` property ([link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-b9b9ca0d1db5b722f779b39557a7788e55054c621cb0fe5a2e6d579be6af390eL22-R23), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-da78c3cd254a2c97bb910dd1daca19edb9be7e5d5d609fedf0468d8692f183ebL22-R23), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-fb4678018340b303901f83cebbdcb241f8b6b85798bdff391c01dda993aae24eL65-R107), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-fb4678018340b303901f83cebbdcb241f8b6b85798bdff391c01dda993aae24eL518-R536), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-fb4678018340b303901f83cebbdcb241f8b6b85798bdff391c01dda993aae24eL538-R547), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-7e79d5937010f09f3619776bd506b065a31aa0f03d3dc498d103cf1bc1a2bd6bL192-R194), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-03df0599a16900681fb98f0f9bf2e2a55df29d940b8edf133101ec5a35d58cceL42-R56), [link](https://github.com/dxos/dxos/pull/4287/files?diff=unified&w=0#diff-81f957adb0f79cf5d3660100adec34b4b31e90f342e85c85a5d0570cb5d81871L88-R98)).

### DIFF
--- a/packages/apps/plugins/plugin-markdown/src/types.ts
+++ b/packages/apps/plugins/plugin-markdown/src/types.ts
@@ -7,6 +7,7 @@ import { IntentProvides } from '@braneframe/plugin-intent';
 import { TranslationsProvides } from '@braneframe/plugin-theme';
 import { Document } from '@braneframe/types';
 import { EditorMode, MarkdownComposerProps } from '@dxos/aurora-composer';
+import { ObjectMeta } from '@dxos/client/echo';
 
 export const MARKDOWN_PLUGIN = 'dxos.org/plugin/markdown';
 
@@ -16,9 +17,11 @@ export enum MarkdownAction {
   CREATE = `${MARKDOWN_ACTION}/create`,
 }
 
-// TODO(burdon): This is just pretending not to be an ECHO object.
 export type MarkdownProperties = {
   title: string;
+
+  // TODO(burdon): Since this is always very precisely an ECHO object why obfuscate it?
+  __meta: ObjectMeta;
   readOnly?: boolean;
 };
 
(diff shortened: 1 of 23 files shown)


packages/apps/plugins/plugin-markdown/src/types.ts
```diff
@@ -17,17 +16,16 @@ export enum MarkdownAction {
   CREATE = `${MARKDOWN_ACTION}/create`,
 }
 
+// TODO(burdon): This is just pretending not to be an ECHO object.
```
wittjosiah, line 19: It's not an echo object because local files also implements it and it's not using echo. If we were using local-only spaces instead of deepsignal then maybe it actually would be an echo object.

wittjosiah (approved):

packages/apps/plugins/plugin-markdown/src/types.ts
richburdon: in that case it should not rely on ECHO type definitions!

dmaretskyi (approved):

PR #4444, opened 2023-10-17
fix: presenter plugin tweaks

jessmartin:
### Summary
📝🐛🎨

Updated the documentation and metadata of the `plugin-presenter` package to match its actual name and functionality. Fixed the incorrect references to the `plugin-draw` package in the `package.json` and `README.md` files.

> _`package.json` changed_
> _Description matches README_
> _Autumn of cleanup_

### Walkthrough
* Renamed and updated the plugin-presenter package to avoid confusion with the plugin-draw package ([link](https://github.com/dxos/dxos/pull/4444/files?diff=unified&w=0#diff-6895c9a0f18716eacd2841f62531c164fdb434e7e06e1c5a46e23422c295cedeL1-R3), [link](https://github.com/dxos/dxos/pull/4444/files?diff=unified&w=0#diff-1439ec5375def5962e6283257461232595b00e5d3b92a25227f0bbde4ac0ca73L4-R4))
* Changed the package.json description field to match the README.md file ([link](https://github.com/dxos/dxos/pull/4444/files?diff=unified&w=0#diff-1439ec5375def5962e6283257461232595b00e5d3b92a25227f0bbde4ac0ca73L4-R4))


packages/apps/plugins/plugin-ipfs/src/components/FileMain.tsx
```diff
@@ -59,8 +59,8 @@ export const FileSlide: FC<{ file: TypedObject }> = ({ file }) => {
 
   // TODO(burdon): Config object-container/-fit.
   return (
-    <div className='h-full'>
-      <FilePreview type={file.type} url={url} className='object-cover' />
+    <div className='h-full flex justify-center align-center'>
```
richburdon, line 62: perhaps set `cover` as an option so that both are supported in principle and you can decide on the default.
then later we can figure out how the user can configure this (e.g., markdown frontmatter, or some UX chrome in the slide section).

richburdon (changes requested):

packages/apps/plugins/plugin-ipfs/src/components/FileMain.tsx
jessmartin: were you thinking passing a `cover` parameter to the `FileSlide` component?

Something like this:

```tsx
export const FileSlide: FC<{ file: TypedObject; cover?: boolean }> = ({ file, cover }) => {
  const config = useConfig();
  if (!file.cid) {
    return null;
  }

  const url = getIpfsUrl(config, file.cid);

  // TODO(burdon): Config object-container/-fit.
  return (
    <div className='h-full flex justify-center align-center'>
      <FilePreview type={file.type} url={url} className={`object-${cover ? 'cover' : 'contain'}`} />
    </div>
  );
};
```

PR #4460, opened 2023-10-19
fix: balancer should wait on framer to send chunk

egorgripasov:


packages/core/mesh/teleport/src/muxing/balancer.test.ts
```diff
@@ -47,4 +93,28 @@ describe('Balancer', () => {
...
     expect(decoded.chunk).to.deep.equal(chunk);
   });
+
+  test('should buffer chunks on the balancer for separate channels', async () => {
+    const channels = 3;
+    const { balancer, stream } = setupBalancer(channels, true);
+
+    await sleep(20);
+
+    expect(balancer.buffersCount).to.equal(channels);
```
dmaretskyi, line 103: can you assert that stream._write wasn't called more then once?

dmaretskyi: just checking the number of buffers doesn't really tell much about backpressure

packages/core/mesh/teleport/src/muxing/framer.ts
```diff
@@ -24,6 +25,10 @@ export class Framer {
   private _bytesSent = 0;
   private _bytesReceived = 0;
 
+  private _writable = true;
+
+  readonly drain = new Event();
```
dmaretskyi, line 30: it would be better to turn this class into two transform streams - they'll handle backpressure automatically

egorgripasov: I'm not a pro in Nodejs streaming capabilities, but isn't `Duplex` also supposed to handle backpressure automatically?
I _think_ the way it happens in the stream is by checking if the writable buffer is full and emitting the `drain` event when it's ready to accept more data, so one way or another we need to create listeners;
I'm considering merging the fix in that form, since we want to eliminate the OOM caused by queueing callbacks, and do experimentation on Transform streams as a second phase

PR #4552, opened 2023-10-31
fix: balancer race condition

nferch:
### Summary
🔄🚀🐛

Refactor `Balancer` class to optimize data transfer over multiple channels. Use a new `_sending` property and a `_enqueueChunk` method to manage the sending queue and avoid duplication. Add error checking and logging with `invariant` and console.

> _Sing, O Muse, of the cunning Balancer, who skillfully_
> _Refined the art of sending and receiving data chunks_
> _Over many channels, swift as the winged messenger Hermes._
> _He added a new property, `_sending`, and a new method, `_enqueueChunk`,_

### Walkthrough
*  Import `invariant` function to assert send buffer conditions ([link](https://github.com/dxos/dxos/pull/4552/files?diff=unified&w=0#diff-0e09ceb6178d7900eed47073b84ac4dca53d5a48851fba05440f6e8e08c681a7R8)).


packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -140,42 +115,80 @@ export class Balancer {
...
       }
+      return chunk;
     }
-    return chunk;
+    return null;
   }
 
-  private async _sendChunk() {
-    if (this._sendBuffers.size === 0) {
+  private async _sendChunk(channelId: number) {
```
dmaretskyi, line 167: channelId not used

dmaretskyi: rename to _sendChunks?

packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -69,36 +71,9 @@ export class Balancer {
...
-      const msg = encodeChunk({
-        chunk,
-        channelId,
-        dataLength: index === 0 ? data.length : undefined,
-      });
-      sendBuffer.push({ msg, trigger: index === chunks.length - 1 ? trigger : undefined });
-    });
-
+    this._enqueueChunk(data, trigger, channelId);
     // Start processing calls if this is the first call.
```
dmaretskyi, line 75: remove comment?

packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -140,42 +115,80 @@ export class Balancer {
...
-      await this._framer.port.send(chunk.msg);
-      chunk.trigger?.wake();
-    } catch (err: any) {
-      chunk.trigger?.throw(err);
+    this._sending = true;
+    let chunk: ChunkEnvelope | null;
+    chunk = this._getNextChunk();
+    while (chunk) {
+      // this is not needed since we await the chunk send
+      if (!this._framer.writable) {
```
dmaretskyi, line 176: 👍

dmaretskyi: does this mean we could remove `framer.drain` and `framer.writable` if they are unused?

packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -140,42 +115,80 @@ export class Balancer {
     return this._channels[index];
   }
 
-  private _getNextChunk(): ChunkEnvelope {
+  private _enqueueChunk(data: Uint8Array, trigger: Trigger, channelId: number) {
+    if (!this._channels.includes(channelId)) {
```
dmaretskyi, line 119: I'm a bit confused about what `_channels` is meant to do. The `.includes` call does a linear O(n) search on the array. Can we merge _sendBuffers and channels into one map instead?

packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -140,42 +115,80 @@ export class Balancer {
...
+      throw new Error(`Unknown channel ${channelId}`);
+    }
+
+    if (!this._sendBuffers.has(channelId)) {
+      this._sendBuffers.set(channelId, []);
```
dmaretskyi, line 124: move this to `addChannel` to avoid having to perform this check

packages/core/mesh/teleport/src/muxing/balancer.ts
nferch: _channels is used for round-robin'ing the channel chunks, which didn't seem great but didn't seem like it could have been the source of the bug.

I also wanted to use a single datastructure here but I wanted to be sure I could implement the round-robin algo properly.

packages/core/mesh/teleport/src/muxing/balancer.ts
nferch: whoops, that one eluded my `git add -p`. more cleanup in general to do if this works.

PR #4609, opened 2023-11-07
feat(functions): Add GitHub resolver

mykola-vrmchk:
copilot:all


mykola-vrmchk:
* main
  * **PR #4609** <a href="https://app.graphite.dev/github/pr/dxos/dxos/4609" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/4609?utm_source=stack-comment).

packages/core/functions/src/runtime/trigger-manager.ts
```diff
@@ -92,7 +92,7 @@ export class TriggerManager {
         count++;
       });
       // TODO(burdon): DSL for query (replace props).
-      const query = space.db.query(Filter.typename(trigger.subscription.type, trigger.subscription.props));
+      const query = space.db.query(Filter._typename(trigger.subscription.type));
```
richburdon, line 95: Filter.typename should work now?

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+import { PublicKey } from '@dxos/keys';
+import { log } from '@dxos/log';
+
+type GithubContributors = RestEndpointMethodTypes['repos']['listContributors']['response']['data'];
+type GithubUser = RestEndpointMethodTypes['users']['getByUsername']['response']['data'];
+
+export const handler: FunctionHandler<FunctionSubscriptionEvent> = async ({ event, context }) => {
+  const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
+
+  scheduleTask(new Context(), async () => {
```
richburdon, line 21: i think this should await (block) since the handle is async.

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+    for (const objectId of event.objects) {
+      const project = context.client.spaces.query({ id: objectId }).objects[0];
```
richburdon, line 23: I think this is clearer:
```
const { objects: projects } = client.query();
const [ project ] = projects[0];
```
But perhaps we need a `getObject(id: string)` API since this comes up a lot.

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+
+      if (project && project.repo && project.repo.includes('github.com') && project.__meta.keys.length === 0) {
```
richburdon, line 25: `if (project?.repo?.includes())`

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+            if (!user.name) {
+              log.warn('Missing user name.');
+              return;
+            }
+            const existing = space.db.query({ name: user.name }).objects;
+            if (existing.length !== 0) {
+              log.info('User already exists', { name: user.name });
+            }
+            space.db.add(
+              new TypedObject(
```
richburdon, line 54: I think this should be `Expando`, but it raises the question about whether we should expose the TypedObject constructor or -- more likely -- make it protected.

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+                {
+                  name: user.name,
+                  email: user.email,
+                  org: user.company,
```
richburdon, line 58: This should be a reference?

packages/experimental/labs-functions/src/functions/github/handler.ts
```diff
@@ -2,6 +2,71 @@
...
+                },
+                { schema: personSchema },
+              ),
+            );
+          }),
+        );
+        await space.db.flush();
+        project.__meta.keys.push({ source: 'github' });
```
richburdon, line 66: - `github.com` (FQ: fully qualified).
- set meta in options of constructor (and note, meta should be immutable so we should really ONLY let you set this in the constructor).

packages/experimental/labs-functions/src/functions/github/obtokit.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const username = 'mykola-vrmchk';
+    const response = await octokit.users.getByUsername({
+      username,
+    });
+
+    const userData = response.data;
+    log.info('userData', { userData });
+  });
+
+  test('Parse project', async () => {
```
richburdon, line 22: CI unit tests should not call the network.
we should consider using a fake here -- and consider different test suite classifications to run in a different CI cycle.
however, I think it's useful to have a "playpen" for testing octokit -- but actually testing it in CI at all doesn't make sense (it's not our code).

richburdon (changes requested):

packages/experimental/labs-functions/src/functions/github/obtokit.test.ts
mykola-vrmchk: Yes

packages/experimental/labs-functions/src/functions/github/handler.ts
mykola-vrmchk: It processes only github links

PR #4686, opened 2023-11-10
feat: Replace `ComboBox` with ⌘K-style component `SearchList`

thure:
This PR adds `SearchList`, a ⌘K-style component implementing similar functionality to the `downshift`-based `ComboBox` component which it replaces.

https://github.com/dxos/dxos/assets/855039/672529d2-ea98-42a5-92d5-6cf2e991ca54

### Etymology

Note that the term ‘combobox’ refers to many patterns. I’d prefer to avoid using the term directly for a component that doesn’t facilitate strictly all of those patterns.

One such pattern is facilitated by this component. The trigger must be a button, not a text input, but it opens a popover or dialog that contains a text input for querying.

This component can also be used on its own to implement a modal ⌘K. In that configuration, it is just a `dialog`, not a `combobox`. It suits the `combobox` role only when the trigger and popover/dialog are used together.

Since it is both narrower in scope and can be used in ways that don’t suit the `combobox` role, it merits a name indicating as much. I chose the term `SearchList` to reflect that this is specifically a ⌘K-style dialog content which is almost always a `listbox`.

#### Why not `Command`

This component inherits from a component called `Command`, however I believe this term should be avoided because it is already defined by ARIA: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/command_role

> ARIA command role
> Best practices
> Do not use.

Furthermore, this component can be used for more than just commands (and the term ‘command’ does not superficially convey it is a queryable list), so as a term it seems to violate the [maxim of manner](https://en.wikipedia.org/wiki/Cooperative_principle#Maxim_of_manner_(clarity)).


thure:
This works as expected in Composer with `TablePlugin` enabled. Will set to merge.

PR #4757, opened 2023-11-22
test(echo-schema): add failing query test

wittjosiah:
### Summary
🧪🔗📁

Added a new test case for `query` module of `echo-schema` package. The test case verifies that query results can handle refs to other objects.

> _`echo-schema` tests_
> _querying refs with flatMap_
> _autumn leaves scattered_

### Walkthrough
* Add a test case for mapping over refs in query results ([link](https://github.com/dxos/dxos/pull/4757/files?diff=unified&w=0#diff-718d737fa6b5a5f359faaa930b144180f7fd8dc9e3592b1ea2cc696aa2028a0cR234-R255)). The test case uses the `query` and `flatMap` methods of the `QueryResult` class to get the task objects from the folder object's ref array. The test case asserts that the result matches the expected task objects. This improves the test coverage and functionality of the `query` module of the `echo-schema` package, which provides a schema system for the `echo` database.

### DIFF
--- a/packages/core/echo/echo-schema/src/object/array.test.ts
+++ b/packages/core/echo/echo-schema/src/object/array.test.ts
@@ -11,6 +11,7 @@ import { Expando, TypedObject } from './typed-object';
 import { AutomergeArray } from '../automerge';
 import { getGlobalAutomergePreference } from '../automerge-preference';
 import { createDatabase, testWithAutomerge } from '../testing';
+import { log } from '@dxos/log';
 
 describe('Arrays', () => {
   testWithAutomerge(() => {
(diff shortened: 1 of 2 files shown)


dmaretskyi:
👍

yarolegovich:
Maybe this was discussed already.
This is happening due to [documented flatMap behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#calling_flatmap_on_non-array_objects).

> // Array-like objects returned from the callback won't be flattened

An option I'd suggest is to modify flatMap of EchoArray to ensure array gets flattened:
```patch
@@ -283,7 +283,12 @@ export class EchoArray<T> implements Array<T> {
     callback: (this: This, value: T, index: number, array: T[]) => U | readonly U[],
     thisArg?: This | undefined,
   ): U[] {
-    return Array.from(this.values()).flatMap(callback, thisArg);
+    const array = Array.from(this.values());
+    const elements: U[] = array.flatMap(function (this: This, value, index, arr) {
+      const result = callback.call(thisArg as This, value, index, arr);
+      return result instanceof EchoArray ? [...(result as EchoArray<U>)] : result;
+    }, thisArg);
+    return new EchoArray(elements);
   }

```
As well as other methods, like:
```patch
   map<U>(callbackfn: (value: T, index: number, array: T[]) => U, thisArg?: any): U[] {
-    return Array.from(this.values()).map(callbackfn, thisArg);
+    return new EchoArray(Array.from(this.values()).map(callbackfn, thisArg));
   }
 
   filter<S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[];
   filter(predicate: (value: T, index: number, array: T[]) => unknown, thisArg?: any): T[];
   filter<S extends T>(predicate: any, thisArg?: unknown): T[] | S[] {
-    return Array.from(this.values()).filter(predicate, thisArg);
+    return new EchoArray(Array.from(this.values()).filter(predicate, thisArg));
   }
```
So that `array.filter(...).flatMap(...)` and `array.flatMap(...)` behave the same.

And I'd also suggest to explicitly specify which methods return `EchoArray` to avoid confusion:
```ts
//case 1 - all good
const results = query.objects.flatMap(...); 
//case 2 - doesn't work, but I can see that query.objects was an EchoArray
const results = [...query.objects].flatMap(...) 
```
We can also add `toArray()` method to the class, so that `[...echoArray]` is not the only way for casting `EchoArray` to `Array`.
Wdyt?

dmaretskyi:
Do we want `EchoArray`s to preserve them being `EchoArray`s after transformations like `map` and `flatMap`?

One aspect of `EchoArray` is its attachment to the object: mutating the array mutates the echo object that stores the data.
But there's another aspect which is reactivity: changes to the array trigger signal notifications.

I think we might want keep #2 but not #1.

dmaretskyi:
@yarolegovich casting EchoArray to array does seem to resolve the issue. 
For example `query.objects.flatMap(({ objects }) => Array.from(objects))` works for me.

It would be nice to have flatMap accept the echo array natively

mykola-vrmchk:
* `main`
  * **PR #4757** <a href="https://app.graphite.dev/github/pr/dxos/dxos/4757?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/4757?utm_source=stack-comment).

PR #4777, opened 2023-11-24
feat: Tree plugin to test hierarchical ECHO objects (move, copy, delete, etc.)

richburdon:
- [x] Create, move, delete.
- [ ] Cascading deletion/reference counting.
- [ ] Import/export subgraphs.

https://github.com/dxos/dxos/assets/3523355/50e7375c-d776-4821-9f4f-63c0ff275ce1

Screenshot 2023-11-25 at 1 34 15 AM

Screenshot 2023-11-25 at 1 55 48 AM

https://github.com/dxos/dxos/assets/3523355/ce677a46-e581-407a-88db-278a3ea29f4e

### Summary
📝🚀🎨

This pull request implements a new plugin for managing tasks in the composer app. It adds the `plugin-tasks` package, which defines the plugin component, metadata, types, and UI components. It also updates the `composer-app` package to register and import the plugin. It also improves the code style of the `StackPlugin` component and the layout of the `StandaloneLayout` component in other plugins.

> _We are the masters of the `TaskList`_
> _We create and edit with the `MarkdownEditor`_
> _We navigate and indent with the arrow keys_
> _We are the `@braneframe/plugin-tasks`_

### Walkthrough
*  Add a new plugin for managing tasks in the composer app ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-47b84ac128e4ba47b697389d563eb2770189956b67cea33857501b520432a1e1R38), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R36), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6L111-R117), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R169), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-aa01c591cecab0e1f8665cacbf254fc9a9418559a3649b0e930ec007c34e2d8cR140-R142), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-e784dcce1e33090afb7f843df9b299bee4335eb64caf37d93c23ab88e7e7ec15R1-R9), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-4135a6bb1289e0b04d8d17036dc793d44f6981971d93b213afd284c5c5ce3bf4R1-R35), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-0384648ed4d5233bc5509dc2542657c7025be9ae679c69d5acb3e07b6c7b1bd6R1-R9), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-250cf438acdee607b897fe66b13d96a6c0d60ede3a46df346438e47fcc9d45a7R1-R51), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-e1d5e01e641c0d08d946297d4b3da102b933ba33bf8ae5e08a07c202e1d6f1b9L1-R7), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-6980eecfd56804ca5da1d1ad2a682a95845d334f67f0b0c6245621449615b54fR1-R15), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-b5bcacb17d6f57e6541b0087d84fb187a40bb52ec3a38a05644edc5adeb308bcR1-R59), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-4401b1e65d9fd00a1111df8c4a0d5a3ece5944de0c1360659d2e0a403cebb914R1-R33), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-bc231459c03d52d86cc5037ee21573951b0ca20c45fbd1bed2b82f43da837e22R1-R111), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-98049cc218f84a167e56e220ff481a77b08234a704294ff90ba3daf5797f4eedR1-R92), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-5ccd50ae6726f3bb73f8863136a35f314bccda3adfb7486db1ac90da2e977d7bR1-R304), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-c41988688437d93b55b66b81a2acf02403e152f7ae45d990ffc370e12d00330dR1-R5), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-3b24e277b1c95f0866d0e56b78c2661ca4c195b6ef46c79205f3d2abdf7f2ab7R1-R28), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-624d318a321756f17354a08743119bfc2e9dbd75d9c31e9a842257a0809e1c88R1-R28), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-a7a29513c1295f8e94e1e2d9277ed9ebdd26f59c457b958f5ccf1efc3e57fee7R1-R31))
  * Define the plugin metadata, interface, and types in the `TasksPlugin.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-bc231459c03d52d86cc5037ee21573951b0ca20c45fbd1bed2b82f43da837e22R1-R111))
  * Define the plugin component and its provides object in the `TasksPlugin.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-bc231459c03d52d86cc5037ee21573951b0ca20c45fbd1bed2b82f43da837e22R1-R111))
  * Define the plugin logic for creating, deleting, and indenting tasks, as well as resolving intents and rendering surface components in the `TasksPlugin.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-bc231459c03d52d86cc5037ee21573951b0ca20c45fbd1bed2b82f43da837e22R1-R111))
  * Define the `Task` type and the `TaskList` component for rendering a nested list of tasks with keyboard navigation and editing features in the `TaskList.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-5ccd50ae6726f3bb73f8863136a35f314bccda3adfb7486db1ac90da2e977d7bR1-R304))
  * Define the `TaskBlock` component for rendering a task object as a markdown editor block in the `TaskBlock.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-624d318a321756f17354a08743119bfc2e9dbd75d9c31e9a842257a0809e1c88R1-R28))
  * Define the `TaskMain` and `TaskSection` components for rendering the subtasks of a task object as the main or section surface of the plugin in the `TaskMain.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-a7a29513c1295f8e94e1e2d9277ed9ebdd26f59c457b958f5ccf1efc3e57fee7R1-R31))
  * Add Storybook stories for the `TaskList` and `TaskBlock` components in the `TaskList.stories.tsx` and `TaskBlock.stories.tsx` files ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-98049cc218f84a167e56e220ff481a77b08234a704294ff90ba3daf5797f4eedR1-R92), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-3b24e277b1c95f0866d0e56b78c2661ca4c195b6ef46c79205f3d2abdf7f2ab7R1-R28))
  * Add a module file for re-exporting the `TaskList` component in the `index.ts` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-c41988688437d93b55b66b81a2acf02403e152f7ae45d990ffc370e12d00330dR1-R5))
  * Add a `package.json` file for defining the package metadata and exports in the `package.json` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-b5bcacb17d6f57e6541b0087d84fb187a40bb52ec3a38a05644edc5adeb308bcR1-R59))
  * Add a `project.json` file for configuring the NX project, the ESBuild compiler, and the Storybook executor in the `project.json` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-4401b1e65d9fd00a1111df8c4a0d5a3ece5944de0c1360659d2e0a403cebb914R1-R33))
  * Add a `LICENSE` file for specifying the MIT license in the `LICENSE` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-e1d5e01e641c0d08d946297d4b3da102b933ba33bf8ae5e08a07c202e1d6f1b9L1-R7))
  * Add a `README.md` file for providing some basic information about the plugin in the `README.md` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-6980eecfd56804ca5da1d1ad2a682a95845d334f67f0b0c6245621449615b54fR1-R15))
  * Add a `.eslintrc.js` file for configuring ESLint for the plugin code in the `.eslintrc.js` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-e784dcce1e33090afb7f843df9b299bee4335eb64caf37d93c23ab88e7e7ec15R1-R9))
  * Add a `.storybook` folder for configuring Storybook for the plugin components in the `.storybook/main.cjs`, `.storybook/preview-head.html`, and `.storybook/preview.cjs` files ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-4135a6bb1289e0b04d8d17036dc793d44f6981971d93b213afd284c5c5ce3bf4R1-R35), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-0384648ed4d5233bc5509dc2542657c7025be9ae679c69d5acb3e07b6c7b1bd6R1-R9), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-250cf438acdee607b897fe66b13d96a6c0d60ede3a46df346438e47fcc9d45a7R1-R51))
  * Add a dependency on the `@braneframe/plugin-tasks` package to the `composer-app` package in the `packages/apps/composer-app/package.json` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-47b84ac128e4ba47b697389d563eb2770189956b67cea33857501b520432a1e1R38))
  * Import the plugin metadata from the `@braneframe/plugin-tasks` package in the `packages/apps/composer-app/src/main.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R36))
  * Add the plugin metadata to the array of plugin metadata in the `packages/apps/composer-app/src/main.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6L111-R117))
  * Add a lazy import of the plugin component from the `@braneframe/plugin-tasks` package in the `packages/apps/composer-app/src/main.tsx` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-9f2aee287f92edd662c13bed460a28a1590c9186b62c98ccc2e00f6877c3d7e6R169))
  * Add a path alias for the `@braneframe/plugin-tasks` package in the `packages/apps/composer-app/tsconfig.json` file ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-aa01c591cecab0e1f8665cacbf254fc9a9418559a3649b0e930ec007c34e2d8cR140-R142))
* Modify the padding of the `StandaloneLayout` component in the `plugin-markdown` package for layout consistency across plugins ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-2dad9280901d107c07117e32c1eaa07331e303c02d11699a00ec95431dd64eacL24-R24))
* Remove unnecessary braces and whitespace in the `parse` function of the `StackPlugin` component in the `plugin-stack` package for code style improvement ([link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-73f12af5e023ef40bc928d988f0d7626f1db6bf7d07a1d91a2dfd4553a50d9e3L55-R60), [link](https://github.com/dxos/dxos/pull/4777/files?diff=unified&w=0#diff-73f12af5e023ef40bc928d988f0d7626f1db6bf7d07a1d91a2dfd4553a50d9e3L116))


packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export type LayoutType = 'tidy' | 'radial' | 'edge';
+
+// TODO(burdon): Normalize API and styling.
+
+export type TreeLayoutSlots = {
+  node?: string;
+  path?: string;
+  text?: string;
+};
```
thure, line 28: IMO we should deprecate the “slots” pattern in favor of either:
- supplying the downstream consumer with subcomponents that support `classNames`, or
- if `classNames` is the only supported prop on these subcomponents, just offering that as a prop name without calling anything “slots” (the latter implies the slot should/will implement a spread of all element props, but only `classNames` is supported making this a misnomer).

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+// https://observablehq.com/@d3/hierarchical-edge-bundling?intent=fork
+
+type Renderer = (svg: SVGSVGElement, data: any, options: any) => void;
+
+export type LayoutType = 'tidy' | 'radial' | 'edge';
+
+// TODO(burdon): Normalize API and styling.
+
+export type TreeLayoutSlots = {
+  node?: string;
```
thure, line 25: If this value is passed to `mx`, the proper type for this is `ClassNameValue`, which can be an array, which is also the reason the prop name should be `classNames` with an ‘s’ (or we should discuss renaming all such props).

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+  slots?: TreeLayoutSlots;
+  radius?: number;
+
+  width: number;
+  height: number;
+  margin?: number;
+
+  padding?: number;
+  // Radius of nodes.
+  r?: number;
```
thure, line 42: These feel mutually redundant (at least `radius` and `r`, also `slots` and all the other values), and also seem like something of a smell… generally top-level plugin components shouldn’t offer such fine-grained control.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+};
+
+export const defaultTreeLayoutSlots: TreeLayoutSlots = {
+  node: 'fill-blue-600',
+  path: 'fill-none stroke-blue-400 stroke-[0.5px]',
+  text: 'stroke-[0.5px] stroke-neutral-700 text-xs', // TODO(burdon): Create box instead of halo.
+};
```
thure, line 49: These should be organized the same way as all other styles are organized. They should go in a `theme.ts` or similar, and export strings/`ComponentFragment`s/`ComponentFunction`s the same as other theme files.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+  ['tidy', TidyTree],
+  ['radial', RadialTree],
+  ['edge', HierarchicalEdgeBundling],
+]);
+
+export type TreeComponentProps<N = unknown> = {
+  space: Space;
+  selected?: string;
+  type?: LayoutType;
+  onClick?: (node?: N) => void;
```
thure, line 61: I don't think we should conflate `onClick`, which is a native handler having the signature `(event: MouseEvent) => void`, with component-specific convenience handlers like this.

For Table, I made the distinction by adding ‘data’ or ‘datum’ (depending on cardinality; ‘datum’ for one-and-only-one of the dataset is supplied) to the prop name. Here, I’d recommend calling the prop `onDatumClick` or `onNodeClick`.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+};
+
+const renderers = new Map<LayoutType, Renderer>([
+  ['tidy', TidyTree],
+  ['radial', RadialTree],
+  ['edge', HierarchicalEdgeBundling],
+]);
+
+export type TreeComponentProps<N = unknown> = {
+  space: Space;
```
thure, line 58: This could be better supplied by a context rather than drilling it down through props.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
thure: For an example in an intermediate component like this, see https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-table/src/theme.ts

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+        marginRight: (width - radius * 2) / 2,
+        marginTop: (height - radius * 2) / 2,
+        marginBottom: (height - radius * 2) / 2,
+        slots: defaultTreeLayoutSlots,
+      };
+
+      const renderer = renderers.get(type);
+      renderer?.(context.ref.current!, data, options);
+    }
+  }, [data, width, height]);
```
thure, line 98: This looks like a smell, is this really the best way to update a d3 graph? Maybe it is…

packages/apps/plugins/plugin-markdown/src/components/StandaloneLayout.tsx
```diff
@@ -21,7 +21,7 @@ export const StandaloneLayout = ({
   return (
     <Main.Content bounce classNames={[baseSurface, topbarBlockPaddingStart]}>
       <div role='none' className={mx(textBlockWidth, 'pli-2')}>
-        <div role='none' className={mx('pbs-4 pbe-6', 'min-bs-[calc(100dvh-var(--topbar-size))] flex flex-col')}>
+        <div role='none' className={mx('pbs-6 pbe-6', 'min-bs-[calc(100dvh-var(--topbar-size))] flex flex-col')}>
```
thure, line 24:
```suggestion
        <div role='none' className={mx('plb-6', 'min-bs-[calc(100dvh-var(--topbar-size))] flex flex-col')}>
```

packages/apps/plugins/plugin-script/src/components/FrameContainer/frame.tsx
```diff
@@ -38,9 +38,9 @@ const services = new ClientServicesProxy(
 
 // TODO(burdon): Error handling.
 createRoot(document.getElementById('root')!).render(
-  <ClientProvider services={() => services}>
-    <div className='flex fixed inset-0'>
+  <div className='flex fixed inset-0'>
```
thure, line 41: Please avoid using existing fragments for this — `fixedInsetFlexLayout` applies these styles (and some more). If you need a different fragment, we should discuss.

Generally we should avoid proliferating `fixed inset-0` elements, this approach has introduced significant tech debt before. What problem to those styles solve in this case?

thure: I see this isn’t actually new for this plugin, so my comment isn’t blocking, just food for thought for now.

packages/apps/plugins/plugin-template/src/translations.ts
```diff
@@ -9,7 +9,7 @@ export default [
     'en-US': {
       [TEMPLATE_PLUGIN]: {
         'plugin name': 'Template',
-        'object title placeholder': 'New object',
+        'object placeholder': 'New object',
```
thure, line 12: 👍✨

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+import React, { useState } from 'react';
+
+import { PublicKey } from '@dxos/keys';
+import { DensityProvider } from '@dxos/react-ui';
+
+import { Tree } from './Tree';
+
+const Story = () => {
+  const [root] = useState(
+    deepSignal({
```
thure, line 18: `deepSignal` is already recursive, all of the calls to `deepSignal` inside this top one are redundant.

thure: IIUC, it’s an antipattern to use `deepSignal` and `useState` together, since they achieve the same goal slightly differently. Consider either instantiating a `deepSignal` and passing it as a prop to a component or using `useState` on its own.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const handleCreate = () => {
+    return deepSignal({
+      id: PublicKey.random().toHex(),
+    });
+  };
+
+  return (
+    <DensityProvider density='fine'>
+      <div className='flex w-full h-screen justify-center bg-neutral-100 dark:bg-neutral-900'>
+        <div className='flex h-full w-[600px] p-2 bg-white dark:bg-black'>
```
thure, line 84: Worth confirming that all this layout business is really necessary. Generally we should avoid intervening on a story’s natural layout so visual regression testing more accurately represents the component’s true layout.

Please use surface fragments for background colors.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+import { getNext, getParent, getPrevious, getItems, type Item, getLastDescendent } from './types';
+
+// TODO(burdon): Break/join lines.
+// TODO(burdon): TextObject/MarkdownEditor
+// TODO(burdon): Cut-and-Paste.
+// TODO(burdon): Drag-and-drop.
+// TODO(burdon): Better key nav.
+
+type TreeOptions = Pick<HTMLAttributes<HTMLInputElement>, 'placeholder' | 'spellCheck'> & {
+  checkbox?: boolean;
```
thure, line 20: The term ‘checkbox’ is a role, but this is extending a collection of input element attributes, so it probably shouldn’t be called this.

There isn’t a good html attribute that encapsulates this, but elsewhere in the repo we use the flag `selectable`.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+};
+
+//
+// Item
+//
+
+type ItemProps = {
+  item: Item;
+  active?: boolean;
```
thure, line 29: `active` is generally reserved for DnD – could this be called something else?

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+  checkbox,
+  placeholder,
+  spellCheck = false,
+  onFocus,
+  onEnter,
+  onDelete,
+  onIndent,
+  onNav,
+}: ItemProps) => {
+  const [focused, setFocused] = useState(false);
```
thure, line 49: It’s best practices to control focus _outside_ of React, using element `ref`s and `id`s. I’d at least put a to-do here that this should be refactored.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+        }
+        break;
+      case 'Escape':
+        // TODO(burdon): Toggle checkbox?
+        break;
+    }
+  };
+
+  return (
+    <div className='flex items-center px-2 gap-3'>
```
thure, line 89: Did `TreeItem.Root` not work for this?
If not, this should have `role='treeitem'`

thure: Is `flex` necessary? Does the layout work as expected when the content is long?

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+        <Input.Root>
+          <Input.Checkbox
+            checked={item.done}
+            onCheckedChange={(checked) => {
+              item.done = !!checked;
+            }}
+          />
+        </Input.Root>
+      )) || (
+        <div className='shrink-0'>
```
thure, line 100: This div looks unnecessary, you can pass this classname to `Square` directly.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+              <DotsThreeVertical />
+            </Button>
+          </DropdownMenu.Trigger>
+          <DropdownMenu.Portal>
+            <DropdownMenu.Content>
+              <DropdownMenu.Viewport>
+                {onDelete && (
+                  <DropdownMenu.Item onClick={onDelete}>
+                    <X className={getSize(5)} />
+                    <p>Delete item</p>
```
thure, line 138: This should come from a `t` call.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+};
+
+//
+// Branch
+//
+
+type BranchProps = TreeOptions & {
+  root: Item;
+  active?: string;
+  onFocus?: (item: Item) => void;
```
thure, line 157: `onFocus` is a native DOM handler name, recommend qualifying this e.g. `onItemFocus`

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+          if (next) {
+            setActive(next.id);
+          }
+        }
+        break;
+    }
+  };
+
+  return (
+    <div className='w-full'>
```
thure, line 302: Is this extra div necessary? `is-full` could simply be passed to `Branch`.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+            onFocus={() => onFocus?.(item)}
+            onEnter={(before) => onCreate?.(root, item, before)}
+            onDelete={() => onDelete?.(root, item)}
+            onIndent={(left) => onIndent?.(root, item, left)}
+            onNav={onNav}
+            {...props}
+          />
+          {(item.items?.length ?? 0) > 0 && (
+            // TODO(burdon): Indent based on density.
+            <div className='pl-4'>
```
thure, line 181: Is this extra div necessary? This `pl-4` could be passed to `Branch`.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+  onCreate?: (parent: Item, item: Item, after?: boolean) => Item;
+  onDelete?: (parent: Item, item: Item) => void;
+  onIndent?: (parent: Item, item: Item, left?: boolean) => void;
+} & Pick<ItemProps, 'onNav'>;
+
+const Branch = ({ root, active, onFocus, onCreate, onDelete, onIndent, onNav, ...props }: BranchProps) => {
+  return (
+    <div className='flex flex-col w-full'>
+      {root.items?.map((item) => (
+        <div key={item.id} className='w-full'>
```
thure, line 167: Is this extra div necessary? This `is-full` could be passed to `Item`.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,321 @@
...
+  active?: string;
+  onFocus?: (item: Item) => void;
+  onCreate?: (parent: Item, item: Item, after?: boolean) => Item;
+  onDelete?: (parent: Item, item: Item) => void;
+  onIndent?: (parent: Item, item: Item, left?: boolean) => void;
+} & Pick<ItemProps, 'onNav'>;
+
+const Branch = ({ root, active, onFocus, onCreate, onDelete, onIndent, onNav, ...props }: BranchProps) => {
+  return (
+    <div className='flex flex-col w-full'>
```
thure, line 165: Using `flex flex-col` here is probably why you have to specify `is-full` on its children — I think it’s quite likely a `block` layout would require less intervention further down.

packages/apps/plugins/plugin-tree/src/components/TreeMain.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { baseSurface, topbarBlockPaddingStart, textBlockWidth, mx, inputSurface } from '@dxos/react-ui-theme';
+
+import { Tree } from './Tree';
+
+export const TreeMain: FC<{ tree: TreeType }> = ({ tree }) => {
+  return (
+    // TODO(burdon): Style like document.
+    <Main.Content classNames={[baseSurface, topbarBlockPaddingStart]}>
+      <div role='none' className={mx(textBlockWidth, inputSurface, 'pli-2')}>
+        <div role='none' className={mx('pbs-4 pbe-4', 'min-bs-[calc(100dvh-var(--topbar-size))] flex flex-col')}>
```
thure, line 18: We should probably make this a fragment…
Also, should this have the `plb-6` you switched to in StandaloneLayout?

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -75,52 +77,57 @@ export const Section = forwardRef<HTMLLIElement, SectionProps>(
...
-                    hoverableControlItem,
-                    hoverableFocusedControls,
-                    hoverableOpenControlItem,
-                    active === 'overlay' && 'invisible',
-                  ]}
-                  // data-testid={testId}
+
+            {/* Menu */}
+            <div role='none' className='flex shink-0'>
+              <div className='p-1'>
```
thure, line 87: These extra divs are unnecessary — just give `m-1 shrink-0` to the `Button` of `DropdownMenu.Trigger`, since the rest of the children are out-of-flow.

packages/ui/react-ui-theme/src/styles/components/input.ts
```diff
@@ -112,7 +112,7 @@ export const inputCheckbox: ComponentFunction<InputStyleProps> = ({ size = 5 },
   mx(
     getSize(size),
     'shrink-0 flex items-center justify-center rounded text-white',
-    'bg-neutral-200 dark:bg-neutral-700 aria-checked:bg-primary-600 dark:aria-checked:bg-primary-600 aria-[checked=mixed]:bg-primary-600 dark:aria-[checked=mixed]:bg-primary-600',
+    'bg-neutral-100 dark:bg-neutral-800 aria-checked:bg-primary-600 dark:aria-checked:bg-primary-600 aria-[checked=mixed]:bg-primary-600 dark:aria-[checked=mixed]:bg-primary-600',
```
thure, line 115: This makes checkboxes invisible in dark mode within chrome surfaces.

thure:
```suggestion
    'bg-neutral-100 dark:bg-neutral-700 aria-checked:bg-primary-600 dark:aria-checked:bg-primary-600 aria-[checked=mixed]:bg-primary-600 dark:aria-[checked=mixed]:bg-primary-600',
```
Would avoid these disappearing completely in certain situations.

packages/ui/react-ui-theme/src/styles/fragments/focus.ts
```diff
@@ -5,6 +5,7 @@
 export const focusRing =
   'focus:outline-none focus-visible:z-[1] focus-visible:hover:outline-none dark:focus-visible:hover:outline-none focus-visible:ring-2 focus-visible:ring-offset-0 focus-visible:ring-primary-350 focus-visible:ring-offset-white dark:focus-visible:ring-primary-450 dark:focus-visible:ring-offset-black';
 
+// TODO(burdon): Requires container to have m-1.
```
thure, line 8: All rings render outside of the [border-box](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model#parts_of_a_box) of the element they’re applied to, unless `ring-inset` is given. I think this comment is less of an actionable to-do and more of a reminder of how rings work.

packages/ui/react-ui-theme/src/styles/components/input.ts
thure: Maybe unselected checkboxes should consistently lighten/darken their surface, in which case we could look into using partly-transparent color tokens instead of fully opaque ones.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
thure: Whatever is at the top of the tree should get `role='tree'`.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
richburdon: I was struggling with this. I think there are Radix style components where the developer builds up the component where className is sufficient, but in this case where there's a more complex (hierarchical) layout I don't know how we'd do that. At the same time classNames seems to collide with the react-ui notion of providing a list (all for the given component). I also consider `classes` like material (i.e., for a map of classNames).

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
richburdon: Would slots become `classNameses`? (Only joking -- yes, point taken).

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
thure: > the react-ui notion of providing a list (all for the given component)

There is no such pattern that I can think of… certainly Table behaves differently, but both Tree and List export subcomponents. What are you referring to?

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.stories.tsx
richburdon: Ah nice catch/lesson. Thanks.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Oh I thought you were using active to represent the current position in the table where the cursor was.

packages/ui/react-ui-theme/src/styles/fragments/focus.ts
richburdon: Did you notice I added m-1 to stacks since the ring was being cut-off.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
richburdon: At some point I'll want to redo this component to have drag-and-drop, and we might face some of the same challenges as the sidebar Tree (perhaps they will become more unified?)

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
thure: No, that’s focus, and because it’s focus in particular that has to be handled outside of React. Some related concepts handled within React include `selected` and `current`, are those relevant here?

packages/ui/react-ui-theme/src/styles/fragments/focus.ts
thure: Yes, but that won’t strictly always be necessary except in that particular situation: where the element’s border box is the same size as its parent’s content box _and_ the parent does not have visible overflow.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
richburdon: Thanks for the reference -- that's helpful.
I had meant to bring up for a while though a slight concern I have about "polluting" (too strong a term) the global namespace with style fragments (esp. given that there are so many of them). Would like to discuss.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.stories.tsx
richburdon: Removed.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Re `selectable`: in this case though it's really meant to enable something like a task list.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Here I do mean focus so will change accordingly.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: I don't understand this point and since I did struggle a bit with this (and also tried to use tabster), perhaps we can discuss offline and then I'll update.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Perhaps I could have used TreeItem more extensively throughout?
This whole thing started as a devtools quick experiment. Let's discuss.

packages/apps/plugins/plugin-explorer/src/components/Tree/Tree.tsx
thure: Sure — currently we have two precedents:
- low-level components must set styles in theme objects (which happen to resemble the old slot objects) in `react-ui-theme`
- intermediate components are allowed to set styles directly in any `className` or `classNames` props, and some, like Table, provide theme objects at the package root, as if they might migrate directly to `react-ui-theme`.

I could see cause for defining theme objects elsewhere, especially for plugin components like this – location seems less important than that style defaults & overrides are defined and applied consistently.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: In that case I've onItemXXXd all callbacks.

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Here I think it is since the Item needs a container if there's also a branch.
And `<></>` can't take a key. (Tell me if there's another way).

packages/apps/plugins/plugin-tree/src/components/Tree/Tree.tsx
richburdon: Great. Deleted both.

PR #4881, opened 2023-12-06
feat(automerge): Add client <-> services synchronization for automerge ECHO implementation

mykola-vrmchk:
copilot:all


mykola-vrmchk:
https://github.com/dxos/dxos/assets/36420699/c58e5945-9ad5-439d-8105-9b02950cab01

PR #4945, opened 2023-12-15
fix: Editor styles

richburdon:
copilot:all


packages/ui/react-ui-editor/src/styles/markdown.ts
```diff
@@ -6,13 +6,14 @@ import { mx } from '@dxos/react-ui-theme';
...
-  3: 'mbs-4 mbe-2 text-2xl font-bold text-inherit no-underline',
-  4: 'mbs-4 mbe-2 text-xl font-extrabold text-inherit no-underline',
-  5: 'mbs-4 mbe-2 text-lg font-extrabold text-inherit no-underline',
-  6: 'mbs-4 mbe-2 font-black text-inherit no-underline',
+  1: 'mbs-4 mbe-2 text-4xl font-body font-medium text-inherit no-underline',
+  2: 'mbs-4 mbe-2 text-3xl font-body font-medium text-inherit no-underline',
+  3: 'mbs-4 mbe-2 text-2xl font-body font-medium text-inherit no-underline',
+  4: 'mbs-4 mbe-2 text-xl font-body font-medium text-inherit no-underline',
+  5: 'mbs-4 mbe-2 text-lg font-body font-medium text-inherit no-underline',
+  6: 'mbs-4 mbe-2 font-body font-medium text-inherit no-underline',
```
thure, line 16: `font-body` should be inherited from the markdown language style, is that not the case? Otherwise this sets the precedent that _all highlightings_ for markdown should set the font family, which should be unnecessary.

packages/ui/react-ui-navtree/src/components/NavTreeItemAction.stories.tsx
```diff
@@ -0,0 +1,73 @@
+//
```
thure, line 1: Move all of the NavTreeItemAction business to a separate PR, I think.

packages/ui/react-ui-searchlist/src/components/SearchList.tsx
```diff
@@ -139,7 +140,7 @@ const SearchListItem = forwardRef<HTMLDivElement, SearchListItemProps>(
         {...props}
         onSelect={handleSelect}
         className={mx(
-          'p-1 rounded select-none cursor-pointer data-[selected]:bg-neutral-450/10 data-[selected]:hover:bg-25/10',
+          'p-1 select-none cursor-pointer data-[selected]:bg-neutral-450/10 data-[selected]:hover:bg-25/10',
```
thure, line 143: Please don’t remove `rounded`, this is a highlightable element, and all such elements in our design system are rounded, even currentable table rows.

thure: If the idea is to remove rounding on highlightable elements, we should/must apply the same treatment to the items in Select, DropdownMenu, Tree, Table, and possibly others that I’m not recalling.

richburdon: is that true also for Select options?

richburdon: anyway, moving to a different PR...

PR #4967, opened 2023-12-19
test(react-ui-stack): setup playwright

wittjosiah:
- test(react-ui-stack): setup playwright testing
- test: fill out tests

copilot:all


packages/ui/react-ui-mosaic/src/mosaic/Container.tsx
```diff
@@ -49,9 +49,9 @@ export type MosaicDropEvent<TPosition = unknown> = MosaicMoveEvent<TPosition> &
   operation: MosaicOperation;
 };
 
-export type MosaicContainerProps<TData extends MosaicDataItem = MosaicDataItem, TPosition = unknown> = Pick<
-  HTMLAttributes<HTMLDivElement>,
-  'className'
+export type MosaicContainerProps<TData extends MosaicDataItem = MosaicDataItem, TPosition = unknown> = Omit<
+  HTMLAttributes<HTMLElement>,
+  'onDrop'
```
thure, line 54: Does MosaicContainer actually pass these props to an element? Seems like we’re not spreading props to one, so we should just pick the ones we’re actually using.

thure (approved):
Excellent

packages/ui/react-ui-mosaic/src/mosaic/Container.tsx
wittjosiah: The change here was to allow any html attribute to be passed through to the container. This is being using currently to pass `data-testid` to `Stack`. `onDrop` is omitted because we override it. If there's a better way to allow html attributes on the container types I'm all ears.

PR #4972, opened 2023-12-20
feat: Effect Schema + Proxy reactivity prototype

dmaretskyi:
copilot:all


packages/core/echo/echo-schema/src/effect/spec.md
```diff
@@ -0,0 +1,20 @@
+# ECHO Schema
+
+- ECHO Schema is a lightweight runtime type system to define objects independently of their storage (e.g., native file system, browser storage, ECHO)
+- All objects are associated with a schema definition.
+- Schemas are defined by EffectTS.
+- Schemas may be statically defined by code, or be runtime mutable.
+- Each space maintains a schema registry.
+- Schema definitions are versioned with semvar semantics.
```
yarolegovich, line 8: not sure how we benefit from semver with schemas
we can consider adding/removing an optional field a patch, but what do minor and major version increases mean?

richburdon: minor = breaking change (field removed or type changed); possibly addressed by auto migration
major = migration not possible via tools

richburdon: plus consider that schema may have validation code associated with it and it might be helpful to have the schema version and ts lib version be in the same version space.

wittjosiah: afaik in semver breaking change is always a major version number

wittjosiah: I think we might want to consider a versioning scheme like https://gitlab.com/staltz/comver for schemas because all that really matters for them is "is this compatible with previous version"?

With comver, I'd propose something like:
MINOR: 100% backwards compatible (adding/removing optional field)
MAJOR: requires a lens/migration if you want compatibility

yarolegovich: > With comver, I'd propose something like:
MINOR: 100% backwards compatible (adding/removing optional field)
MAJOR: requires a lens/migration if you want compatibility

this sounds more useful imo 👍

PR #4975, opened 2023-12-21
fix(plugin-stack,react-ui-stack): Stack DnD & overreactivity

thure:
This PR addresses a number of issues related to Stack & StackPlugin:
- [x] overreactivity, which was the root cause of:
  - keyboard access issues
  - incorrect order on drop (or that resolved on its own at some point)
  - the usual consequences of overreactivity when the component is more expensive to mount (PDF iframes, CodeMirror instances, etc)
- [x] dragging new item to end of stack is/was not ultimately placing it there
- [x] Overlay is not getting a truthy `StackSection`, so copying is no longer to-spec

This PR also:
- adjusts some styles so DnD interactions with Stack are aligned with the same interactions elsewhere.
- refactors Mosaic to use the correct `classNames` prop name when its value is eventually passed to `mx`
- fixes `all-stories`’s glob patterns

https://github.com/dxos/dxos/assets/855039/ddf3937e-0670-4c25-aa9b-783d0acad285


packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
```diff
@@ -77,6 +79,7 @@ export const StackMain: FC<{ stack: StackType }> = ({ stack }) => {
     const object = parseData?.(active.item, 'object');
     // TODO(wittjosiah): Stop creating new section objects for each drop.
     if (object && over.path === Path.create(id, over.item.id)) {
+      console.log('[splice]', 'todo: find whether over is last');
```
wittjosiah, line 82: 👀

packages/ui/react-ui-mosaic/src/mosaic/Tile.tsx
```diff
@@ -104,37 +104,34 @@ export const DraggableTile = ({
...
-      path={path}
-      type={type}
-      position={position}
-      operation={operation}
-      isOver={isOver}
-      {...props}
-    />
-  );
-};
+export const DroppableTile = forwardRef<HTMLDivElement, MosaicTileProps<any>>(
```
wittjosiah, line 107: Elsewhere in mosaic I think I used `HTMLElement` for the generic ref type at the api level to allow it to be cast to any element inside the component

packages/ui/react-ui-stack/src/components/props.ts
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import type { FC, PropsWithChildren } from 'react';
+
+import type { MosaicActiveType, MosaicContainerProps, MosaicDataItem, MosaicTileProps } from '@dxos/react-ui-mosaic';
+
+export type StackSectionContent = MosaicDataItem & { title?: string };
```
wittjosiah, line 9: skeptical about factoring out all props to a single file away from actual source/usage, maybe an antipattern? what's the motivation here?

packages/ui/react-ui-stack/src/components/Stack.tsx
```diff
@@ -59,52 +38,37 @@ export const Stack = ({
...
-        return active === 'overlay' ? <List>{section}</List> : section;
-      }),
-    [id, SectionContent, transform, activeItem],
-  );
-
   // TODO(burdon): Create context provider to relay inner section width.
-  const getOverlayStyle = useCallback(() => ({ width: Math.min(width, 59 * 16) }), [width]);
+  const getOverlayStyle = useCallback(() => {
+    return { width: Math.min(width, 59 * 16) };
+  }, [width]);
```
wittjosiah, line 48: revert to one-liner

packages/ui/react-ui-stack/src/components/Stack.tsx
```diff
@@ -114,19 +78,30 @@ export const Stack = ({
...
-                item={item}
+                item={{
+                  // Spreading doesn’t always get the id if it is a Proxy(TypedObjectImpl)
+                  id: item.id,
+                  object: item.object,
+                  transform,
+                  onRemoveSection,
+                  onNavigateToSection,
+                  SectionContent,
+                }}
```
wittjosiah, line 104: this feels like the tile needs some other mechanism to pass forward props, because these aren't really the "item" per say

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -18,120 +18,148 @@ import {
...
+      </Section>
     );
+
+    return active === 'overlay' ? <List>{section}</List> : section;
   },
 );
+
+SectionTile.displayName = SECTION_TILE_NAME;
+
+export { Section, SectionTile };
```
wittjosiah, line 165: is there a reason for this vs `export const`? generally I'd favour consistency within the codebase

thure: This is how all exports are done in `react-ui` core, which is to align with Radix exports which look like this. I believe it’s so the call to set `.displayName` looks sure.

packages/ui/react-ui-stack/src/components/Stack.tsx
thure: I was able to revert this later.

packages/ui/react-ui-stack/src/components/props.ts
thure: This was to prevent circular dependencies between Stack and its Sections, since they’re in different files. I only move types to a common file when genuinely necessary, will reassess here since things have changed since your comment.

packages/ui/react-ui-mosaic/src/mosaic/Tile.tsx
thure: Right, I was experimenting with ref passing and intended to polish this up before merging. It’s no longer necessary, so I'll revert this.

packages/ui/react-ui-stack/src/components/Section.tsx
thure: Essentially I use this export style in any module that sets `.displayName`, which is done to improve ergonomics if anything gets logged about `useContext`.

Using a StackContext didn’t help with this solution, though, so I’ll see about reverting.

packages/ui/react-ui-mosaic/src/mosaic/Tile.tsx
thure: Reverted 👍

packages/ui/react-ui-stack/src/components/Section.tsx
thure: Was able to revert 👍

packages/ui/react-ui-stack/src/components/props.ts
thure: Was able to revert 👍

PR #4984, opened 2023-12-22
feat: agent provider UI and API

nferch:
copilot:all


packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+      onAgentDestroy={handleAgentDestroy}
+      validationMessage={validationMessage}
+      agentStatus={agentStatus}
+      agentActive={agentActive}
+    />
+  );
+};
+
+// TODO(zhenyasav): impl shouldn't need send()
+export const AgentFormImpl = (props: AgentFormImplProps) => {
```
richburdon, line 129: Impl is a bit strange; why do you need another function?
Also, generally destructure inside function: 
e.g., `const AgentForm = ({ ... }: Props) => `

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+        setAgentStatus(agentStatus);
+      } else {
+        setAgentActive(false);
+        setValidationMessage('No agent deployed via provider.');
+      }
+    };
+    void fetchAgentStatus();
+  }, []);
+
+  const onAgentCreateEvent = useCallback(async (invitation: Invitation) => {
```
richburdon, line 55:
```
const handleX = () => {};
return <Foo onX={handleX} />;
```

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { useClient, useAgentHostingProviderClient } from '@dxos/react-client';
+import { type Identity, useHaloInvitations } from '@dxos/react-client/halo';
+import { Invitation, InvitationEncoder } from '@dxos/react-client/invitations';
+import { useTranslation } from '@dxos/react-ui';
+
+import { Action, Actions, StepHeading, Input } from '../../../components';
+import { type IdentityPanelStepProps } from '../IdentityPanelProps';
+import { type IdentityEvent } from '../identityMachine';
+
+export interface AgentFormProps extends Omit<IdentityPanelStepProps, 'send'> {
```
richburdon, line 17: The general pattern is to use `type` defs with the Props suffix associated with a component. I don't see the need for the two defs?

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+    invitations.forEach((invitation) => invitation.cancel());
+
+    // TODO(nf): do this work in the hosting provider client?
+    const invitation = client.halo.share({ type: Invitation.Type.MULTIUSE, authMethod: Invitation.AuthMethod.NONE });
+
+    invitation.subscribe(onAgentCreateEvent);
+  };
+
+  const handleAgentDestroy = async () => {
+    const identity = client.halo.identity.get();
```
richburdon, line 94: I think this should use `invariant` to throw rather than being defensive.

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+// TODO(zhenyasav): impl shouldn't need send()
+export const AgentFormImpl = (props: AgentFormImplProps) => {
+  const { active, onAgentCreate, onAgentDestroy, send, validationMessage, agentStatus, agentActive } = props;
+  const disabled = !active;
+  const { t } = useTranslation('os');
+  return (
+    <>
+      <div role='none' className='grow flex flex-col justify-center'>
+        <Input
```
richburdon, line 136: why use an input rather than a div?

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export type AgentHostingProvider = {
+  name: string;
+  BaseURL: string;
+};
+
+// TODO: Load from config or dynamically discover
+const DEFAULT_AGENT_HOSTING_PROVIDER: AgentHostingProvider = {
+  name: 'default',
+  BaseURL: 'http://localhost:8082/v1alpha1/',
```
richburdon, line 15: baseUrl
also i think we have a doc/note somewhere about port numbers in an effort to coordinate across all of our servers; i'll try to find it and link it back.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { log } from '@dxos/log';
+
+export type AgentHostingProvider = {
+  name: string;
+  BaseURL: string;
+};
+
+// TODO: Load from config or dynamically discover
+const DEFAULT_AGENT_HOSTING_PROVIDER: AgentHostingProvider = {
```
richburdon, line 13: extend config.proto?

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+      log.warn('bad response from agent create', { res });
+    }
+  }
+
+  public async getAgent(agentID: string) {
+    const res = await fetch(new URL('agent/' + agentID, this.agentHostingProvider.BaseURL));
+    if (res.status !== 200) {
+      log.warn('request to agent get failed', { res });
+      return null;
```
richburdon, line 54: as above, consider throwing and defining an ApiError class.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    log.info('getAgent', { res });
+
+    try {
+      const agent = await res.json();
+      return agent.metadata.uid;
+    } catch (e) {
+      if (e instanceof TypeError) {
+        log.warn('failed to parse response from agent create', { res });
```
richburdon, line 64: log the exception: `log.error('failed...', err);`

packages/sdk/react-client/src/client/useAgentHostingProvider.ts
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { AgentHostingProviderClient } from '@dxos/client/services';
+
+export const useAgentHostingProviderClient = (): AgentHostingProviderClient => {
+  return new AgentHostingProviderClient();
+};
```
richburdon, line 9: Assume this is a placeholder; you might do something like this
```
const useAgentManager = () => {
  const config = useConfig();
  return useMemo(() => new AgentManager(config), [config]);
}
```

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const invitationCode = InvitationEncoder.encode(invitation);
+    if (invitation.state === Invitation.State.CONNECTING) {
+      console.log(JSON.stringify({ invitationCode, authCode: invitation.authCode }));
+      if (!identity) {
+        setValidationMessage('Identity not found?');
+        console.error('Identity not found');
+        return;
+      }
+      agentHostingProviderClient
+        ?.createAgent(invitationCode, identity.identityKey.truncate())
```
richburdon, line 65: as above, i think you should pass in the PublicKey.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// Interface to REST API to manage agent deployments
+// TODO(nf): for now API just simply returns created k8s CRD objects, define backend-agnostic API
+export class AgentHostingProviderClient {
+  private agentHostingProvider: AgentHostingProvider;
+  constructor() {
+    this.agentHostingProvider = DEFAULT_AGENT_HOSTING_PROVIDER;
+  }
+
+  public async createAgent(invitationCode: string, identityKey: string) {
```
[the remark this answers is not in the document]
nferch: I considered that, but I didn't want to prematurely couple the agent identifier with the public key of the identity, e.g. if the same identity wanted to create multiple agents. That it was derived from the public key was incidental and out of convenience.

Somewhat related, I think it's still an open question whether this API would drive product subscription or if it just handles the deployment, taking some sort of acceptance token generated out of band through an IFRAME to a separate app or federation via OAuth.

packages/sdk/client/src/services/agent-hosting-provider.ts
nferch: my assumption was that `TypeError` would only result from decoding the JSON, so only the result would be relevant context.

but I agree regarding the need for error classes so the caller can distinguish transport, encoding, and transaction errors

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
```diff
@@ -0,0 +1,161 @@
...
+        .then((res) => {
```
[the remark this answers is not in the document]
nferch: me too!

packages/sdk/react-shell/src/panels/IdentityPanel/steps/AgentPanel.tsx
nferch: I'm not sure if I follow, possibly due to my poor function naming. `onAgentCreateEvent` is passed to `invitation.subscribe()` within `handleAgentCreate` which is referred to in `<AgentFormImpl... onAgentCreate={handleAgentCreate}`

PR #4985, opened 2023-12-22
feat: add Docker publish

nferch:
copilot:all


nferch:
a couple things noteworthy:

* didn't notice any breaking changes with the `vite-plugin` upgrade, but obviously a big blast radius. cc @wittjosiah 
* rename of the cli binary within the repo, which I don't think is directly referenced anywhere else. cc @mykola-vrmchk 
* still puzzled by the tsc/protobuf issue, but this workaround seems stable.
* should be trivial to add Docker hub as a target if we want, which seems a bit more visible than ghcr.io. cc @zhenyasav

wittjosiah:
@nferch iirc the vite plugin was primarily doing source map uploads during productions builds and doesn't have much if any affect on runtime, so should be alright

PR #5000, opened 2023-12-27
feat: Paste into outliner from LogSeq

dmaretskyi:
copilot:all


packages/apps/plugins/plugin-outliner/src/utils/parser.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  test('does not parse plain text', () => {
+    const outline = tryParseOutline('plain text\n- bullet');
+
+    expect(outline).to.be.undefined;
+  });
+});
+
+// NOTE: Keep tabs intact.
+const TEXT = `- Contacts
```
richburdon, line 30: more repeatable:

```
const tests = [{
  text: '',
  expected: []
}]
```

packages/apps/plugins/plugin-outliner/src/utils/parser.ts
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { Tree } from '@braneframe/types';
+import { TextObject } from '@dxos/client/echo';
+
+export const tryParseOutline = (outline: string): Tree.Item[] | undefined => {
```
richburdon, line 8: just parseOutline.
Is this just markdown? If not provide a reference for the format?
E.g., is this Loqseq specific?

packages/apps/plugins/plugin-outliner/src/components/Outliner/Outliner.tsx
```diff
@@ -173,6 +178,24 @@ const OutlinerItem = ({
         <TextEditor
           ref={editorRef}
           model={model}
+          extensions={[
+            EditorView.domEventHandlers({
```
richburdon, line 182: pls factor out the extension to separate file.

richburdon (changes requested):

PR #5001, opened 2023-12-27
fix: Remove margin/paddings from stack sections

richburdon:
- Experimental contiguous layout for stacks.
- Need to figure out dragging outlines.

Screenshot 2024-01-04 at 11 45 53 PM

copilot:all


packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -28,6 +28,8 @@ export type SectionProps = PropsWithChildren<{
   // Data props.
   id: string;
   title: string;
+  index: number;
+  count: number;
```
thure, line 32: Prefer using tailwind’s `first` and `last` selectors over passing props purely for stylistic reasons: https://tailwindcss.com/docs/hover-focus-and-other-states#first

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -38,20 +40,25 @@ export type SectionProps = PropsWithChildren<{
...
           <div
             role='none'
             className={mx(
               surfaceElevation({ elevation: 'group' }),
               inputSurface,
               hoverableControls,
-              'flex rounded',
+              'flex',
+              index === 0 && 'rounded-t',
+              index === count - 1 && 'rounded-b',
```
thure, line 61: Prefer logical properties — instead of physical `t` use logical `bs`, instead of physical `b` use logical `be`.

thure (commented):
Recommending a few changes that aren't blocking but worth trying before merging.

packages/ui/react-ui-stack/src/components/Section.tsx
richburdon: nice. i was able to use group and group-first:border-be, etc.

PR #5018, opened 2024-01-03
fix: broken npm create initializers

zhenyasav:
copilot:all

### DIFF
--- a/packages/devtools/initializers/create-bare/project.json
+++ b/packages/devtools/initializers/create-bare/project.json
@@ -2,6 +2,14 @@
   "sourceRoot": "packages/devtools/initializers/create-bare/src",
   "projectType": "library",
   "targets": {
+    "build": {
+      "executor": "@nrwl/js:tsc",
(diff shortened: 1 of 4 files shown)


packages/devtools/initializers/create-bare/project.json
```diff
@@ -2,6 +2,14 @@
   "sourceRoot": "packages/devtools/initializers/create-bare/src",
   "projectType": "library",
   "targets": {
+    "build": {
+      "executor": "@nrwl/js:tsc",
```
wittjosiah, line 6: I think this isn't needed because it'll now inherit from nx.json where this is specified

wittjosiah (approved):

packages/devtools/initializers/create-bare/project.json
wittjosiah: actually a lot of other packages just have `"build": {},` that might even be enough with inheritance if it follows the patterns

PR #5022, opened 2024-01-03
feat(composer-app): add basic native URL support

mjamesderocher:
This adds support for handling URLs using an app-specific URL scheme (composer://) in the native Mac app.

To test, you will need the id of a document. You can get this by opening the dev tools while the native app is running and entering `window.location.pathname` into the console. Combine everything after the forward slash with with `composer://` and enter it into the URL bar of Safari.

copilot:all


packages/apps/composer-app/src/main.tsx
```diff
@@ -87,6 +90,7 @@ const main = async () => {
       ThemeMeta,
       // Outside of error boundary so that updates are not blocked by errors.
       PwaMeta,
+      NativeMeta,
```
wittjosiah, line 93: you also need to add it to the `plugins` map, e.g. `[NativeMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-native')`

packages/apps/composer-app/src/main.tsx
```diff
@@ -25,6 +25,9 @@ import MapMeta from '@braneframe/plugin-map/meta';
 import MarkdownMeta from '@braneframe/plugin-markdown/meta';
 import MermaidMeta from '@braneframe/plugin-mermaid/meta';
 import MetadataMeta from '@braneframe/plugin-metadata/meta';
+
+import NativeMeta from '../../plugins/plugin-native/src/meta';
```
wittjosiah, line 29: make sure to add `@braneframe/plugin-native` as dep to composer and import from that as well

packages/apps/composer-app/src/main.tsx
mjamesderocher: Thanks!

packages/apps/composer-app/README.md
```diff
@@ -3,7 +3,7 @@
 To run the app with HALO.
 
 ```bash
-pnpm -w nx run composer-app:serve-with-vault
+pnpm -w nx run composer-app:serve
```
wittjosiah, line 6: 🙏

packages/apps/composer-app/src/main.tsx
```diff
@@ -164,6 +161,8 @@ const main = async () => {
       [MarkdownMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-markdown')),
       [MermaidMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-mermaid')),
       [MetadataMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-metadata')),
+      ...(isSocket ? { [PwaMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-pwa')) } : {}),
+      [NativeMeta.id]: Plugin.lazy(() => import('@braneframe/plugin-native')),
```
wittjosiah, line 165: can you `isSocket` this one too? I think if it's in list of available plugin it will show up in the settings as available to enable

packages/apps/plugins/plugin-native/README.md
```diff
@@ -0,0 +1,15 @@
+# @braneframe/plugin-native
+
+Composer plugin that runs when on a native device.
```
wittjosiah, line 3: maybe note that this is specifically for Socket Supply?

packages/apps/plugins/plugin-native/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@types/react-dom": "^18.0.6",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "vite": "^5.0.11"
+  },
+  "peerDependencies": {
+    "@dxos/app-framework": "workspace:*",
+    "@phosphor-icons/react": "^2.0.5",
+    "react": "^18.2.0"
+  },
```
wittjosiah, line 45: can you review these and remove deps that aren't being used in this plugin?

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -94,3 +115,10 @@ export const safeParseJson = <T extends object>(data: string | undefined | null,
   }
   return defaultValue;
 };
+
+export const NativePlugin = (): PluginDefinition => ({
+  meta,
+  ready: async (plugins) => {
+    void initializeNativeApp(plugins);
```
wittjosiah, line 122: I'm wondering if it actually makes sense to await this? Presumably there's functionality in here that would make the app behave weird if it wasn't available?

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -12,12 +16,19 @@ const KEY_WINDOW_SIZE = 'dxos.org/composer/settings/window/size';
  * https://www.npmjs.com/package/@socketsupply/socket
  * https://github.com/socketsupply/socket-examples
  */
-export const initializeNativeApp = async () => {
+export const initializeNativeApp = async (plugins: Plugin[]) => {
```
wittjosiah, line 19:
```suggestion
const initializeNativeApp = async (plugins: Plugin[]) => {
```

packages/apps/composer-app/src/main.tsx
```diff
@@ -53,19 +54,14 @@ import { setupConfig } from './config';
...
   const config = await setupConfig();
   const services = await createClientServices(config);
-
   const isSocket = !!(globalThis as any).__args;
-  if (isSocket) {
-    void initializeNativeApp();
-  }
-
   const App = createApp({
+
```
wittjosiah, line 64:
```suggestion

  const App = createApp({
```

small nit about spacing

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -94,3 +115,10 @@ export const safeParseJson = <T extends object>(data: string | undefined | null,
   }
   return defaultValue;
 };
+
+export const NativePlugin = (): PluginDefinition => ({
+  meta,
+  ready: async (plugins) => {
+    void (await initializeNativeApp(plugins));
```
wittjosiah, line 122:
```suggestion
    await initializeNativeApp(plugins);
```
`void` is the opposite of `await` so I'm not sure this will actually await it or not. The linter requires `void` in front of unawaited promises just so it's clear that they're promises when reading the code.

wittjosiah (commented):
Looking good, nice work!

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -81,6 +92,16 @@ export const initializeNativeApp = async () => {
     }
   });
 
+  window.addEventListener('applicationurl', (event: any) => {
+    const slug = event.url.host;
```
richburdon, line 96: - perhaps more compact and localizes logic below:
- also, please add a comment/link reference for applicationurl
- does it make sense to validate the url here or let the ACTIVATE function do that?

```tsx
window.addEventListener('applicationurl', ({ url }: any) => {
  if (!url.host.match(/../)) {
    // TODO(xxx): Better way to handle error than alert?
    alert();
    return;
  }

  void intentPlugin?.provides.intent.dispatch({
    action: LayoutAction.ACTIVATE,
    data: { id },
  });
});
```

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -2,8 +2,12 @@
...
 //
 
+import type { Plugin, PluginDefinition } from '@dxos/app-framework';
+import { LayoutAction, parseIntentPlugin, resolvePlugin } from '@dxos/app-framework';
 import { log } from '@dxos/log';
 
+import meta from './meta';
+
 // TODO(burdon): Reconcile with other properties.
 const KEY_WINDOW_SIZE = 'dxos.org/composer/settings/window/size';
```
richburdon, line 12: now that this is a plugin:

`dxos.org/plugin/native/window-size`

richburdon (changes requested):

richburdon:
For PRs, please add a short description of how the feature works/can be tested.

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
mjamesderocher: Because there is already an error message displayed by the app when the URL is not supported, I am going to remove the validity check here and always fire LayoutAction.ACTIVATE

wittjosiah: @mjamesderocher is this resolved? I still see the check below?

mjamesderocher: @wittjosiah Yes, I had changed the code to be simply:

```// applicationurl is a custom event fired by the Socket Supply Runtime:
  // https://github.com/socketsupply/socket/blob/ef7fb5559876e41062d5896aafb7b79989fc96e5/api/internal/events.js#L6
  window.addEventListener('applicationurl', ({ url }: any) => {
    void intentPlugin?.provides.intent.dispatch({
      action: LayoutAction.ACTIVATE,
      data: { id: url.host },
    });
  });```
  
Am I missing something else that needs to be changed?

wittjosiah: I just thought you were saying we didn't need to validate here because ACTIVATE would do that, but seems like the code here is still checking `match` on the `slug` before navigating?

mjamesderocher: @wittjosiah can you show me the lines where you are seeing a check for match? https://github.com/dxos/dxos/blob/7607cbcf2e8002d666fd3142669de5222630afd6/packages/apps/plugins/plugin-native/src/NativePlugin.tsx

wittjosiah: Ah my apologies I think I must've been looking at an outdate diff LGTM 👍

PR #5024, opened 2024-01-03
refactor: cleanup settings, translations, commands

wittjosiah:
copilot:all


packages/apps/composer-app/vite.config.ts
```diff
@@ -99,7 +99,6 @@ export default defineConfig({
       content: [
         resolve(__dirname, './index.html'),
         resolve(__dirname, './src/**/*.{js,ts,jsx,tsx}'),
-        resolve(__dirname, './node_modules/@dxos/app-framework/src/**/*.{js,ts,jsx,tsx}'),
```
richburdon, line 102: i somehow needed to do this -- if you've tested all the dialogs then perhaps this is now resolved.

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -2,27 +2,48 @@
...
+  type GraphBuilderProvides,
+  resolvePlugin,
   type MetadataRecordsProvides,
   type PluginDefinition,
   type SurfaceProvides,
   type TranslationsProvides,
+  type Plugin,
+  parseIntentPlugin,
+  LayoutAction,
+  type GraphProvides,
```
richburdon, line 18: i wish lint would consistently sort this. types first? consts, function, classes, etc. ;)

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -87,6 +114,26 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
...
+            return;
+          }
+
+          const intentPlugin = resolvePlugin(plugins, parseIntentPlugin);
+          parent.addAction({
+            id: 'dxos.org/plugin/navtree/open-commands',
+            label: ['open commands label', { ns: NAVTREE_PLUGIN }],
+            icon: (props) => <MagnifyingGlass {...props} />,
+            keyBinding: 'meta+k',
+            invoke: () =>
```
richburdon, line 129: thanks

packages/apps/plugins/plugin-navtree/src/translations.ts
```diff
@@ -9,7 +9,12 @@ import { NAVTREE_PLUGIN } from './meta';
 export default [
   {
     'en-US': {
-      [NAVTREE_PLUGIN]: {},
+      [NAVTREE_PLUGIN]: {
+        'open settings label': 'Settings',
```
richburdon, line 13: i don't think these are intrinsically part of the navtree plugin though.
other apps might not use the navtree, which is why i moved them out (since i started with them here)... 
perhaps we should do what you've done with settings and create another plugin?
TODO?

packages/apps/plugins/plugin-settings/src/components/SettingsDialog.tsx
```diff
@@ -34,7 +34,7 @@ export const SettingsDialogContent = () => {
   // TODO(burdon): Common treatment for dialogs (e.g., shrink height to fit, mobile, etc.)
   return (
     <Dialog.Content classNames={['h-full md:max-is-[40rem] overflow-hidden']}>
-      <Dialog.Title>{t('settings dialog title', { ns: 'os' })}</Dialog.Title>
+      <Dialog.Title>{t('settings dialog title')}</Dialog.Title>
```
richburdon, line 37: Note, PluginList below is the kind of thing we should have as a higher level component (react-ui-list) so that we have keynav/selection/highlight.

packages/apps/plugins/plugin-settings/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { SettingsPlugin } from './SettingsPlugin';
+
+export default SettingsPlugin;
+
+export * from './SettingsPlugin';
+
+// TODO(wittjosiah): Remove.
+//  Settings should be exposed from plugins as state and intents rather than components.
+export * from './components';
```
richburdon, line 13: ok to remove this now and expose the dialog via the plugin's surface?

packages/apps/plugins/plugin-help/src/components/Shortcuts/ShortcutsDialog.tsx
```diff
@@ -26,7 +27,6 @@ export const ShortcutsDialogContent = () => {
   const { t } = useTranslation('os');
```
richburdon, line 27: ok, so this is good enough to handle all of the plugin-specific translations?

packages/apps/composer-app/vite.config.ts
wittjosiah: this was because you added ui components to app-framework

packages/apps/plugins/plugin-navtree/src/translations.ts
wittjosiah: yeah this will likely go away, probably deserves a TODO. The sidebar will likely introduce surfaces or something like that. I think it's better here than in the `os` translations though, this at least couples the translation with the plugin that's using it.

packages/apps/plugins/plugin-settings/src/index.ts
wittjosiah: That might work, we haven't used surfaces with children a ton but theoretically should work. Not super motivated to make that change since it'd go away once we generate settings UI from data but not opposed to it either.

packages/apps/plugins/plugin-help/src/components/Shortcuts/ShortcutsDialog.tsx
wittjosiah: ah missed this one, this should be `HELP_PLUGIN`

PR #5051, opened 2024-01-06
fix(react-ui-editor): Don't store positions in widgets + other fixes

marijnh:
This makes it so that widgets don't need to be re-rendered when their position in the document changes.


richburdon (approved):

richburdon (approved):

marijnh:
@richburdon I've added a bunch of patches here that would otherwise conflict with the initial one. So the original title of the PR doesn't cover all of these patches anymore. They should be self-contained and probably best reviewed by commit by commit.

packages/ui/react-ui-editor/src/themes/default.ts
```diff
@@ -49,12 +49,12 @@ export const defaultTheme: ThemeStyles = {
   '&.cm-focused': {
     outline: 'none',
   },
-  '& .cm-scroller': {
+  '.cm-scroller': {
```
richburdon, line 52: oh i thought this was a requirement to anchor a specific instance (e.g., if there are multiple instances on the page).

packages/ui/react-ui-editor/src/extensions/link.ts
```diff
@@ -92,16 +89,12 @@ export const link = (options: LinkOptions = {}): Extension => {
...
-    private readonly _url: string,
-    private readonly _onAttach: LinkOptions['onRender'],
-  ) {
+  constructor(private readonly _url: string, private readonly _onAttach: LinkOptions['onRender']) {
     super();
   }
 
   override eq(other: LinkButton) {
-    return this._pos === other._pos && this._url === other._url;
+    return this._url === other._url;
```
richburdon, line 97: if there were multiple links on the same page with the same URL, this would still be fine?

packages/ui/react-ui-editor/src/extensions/image.ts
```diff
@@ -64,6 +80,8 @@ class ImageWidget extends WidgetType {
     const img = document.createElement('img');
     img.setAttribute('src', this._url);
     img.setAttribute('class', 'cm-image');
+    // Images are hidden until successfully loaded to avoid flickering effects.
+    img.onload = () => img.classList.add('cm-loaded-image');
```
richburdon, line 84: ah nice.

richburdon (approved):

packages/ui/react-ui-editor/src/extensions/link.ts
marijnh: Yes. The widgets just display the URL. They'd be interchangeable.

PR #5073, opened 2024-01-08
fix: Harmonize document menus

thure:
This PR harmonizes document menus by reusing node actions to produce a three-dots menu after the breadcrumbs on all documents.

To do
- [x] Use node actions to render a menu next to the breadcrumbs
- [x] Make it so popovers invoked by node actions are anchored by their invoker
- [x] Restore popovers invoked by keyboard shortcuts
- [x] Do something about Github actions for Markdown documents, which were being rendered into the Surface that this menu replaces

https://github.com/dxos/dxos/assets/855039/68ab852f-d5c2-4766-9536-8ab34db60071


packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
```diff
@@ -82,7 +82,7 @@ export const MainLayout = ({ fullscreen, showComplementarySidebar = true }: Main
                   <MenuIcon weight='light' className={getSize(4)} />
                 </Button>
 
-                <Surface role='heading' limit={2} />
+                <Surface role='navbar-start' limit={1} />
```
wittjosiah, line 85: why limit 1? if it's limit one and used by a core plugin it's essentially not extensible

packages/sdk/app-graph/src/action.ts
```diff
@@ -10,6 +10,10 @@ import type { MaybePromise } from '@dxos/util';
 // TODO(thure): `Parameters<TFunction>` causes typechecking issues because `TFunction` has so many signatures.
 export type Label = string | [string, { ns: string; count?: number }];
 
+export type InvokeParams = Partial<{
+  caller: string;
```
wittjosiah, line 14: Interesting, do you think this is generic enough that the typedef is a framework-level thing? I was envisioning that the params would be type unknown and you'd have to validate them in the invoke (sort of like surface component data). That way you could pass in anything, but if you want to use it you have to validate it is what you think it is.

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -563,9 +564,10 @@ export const SpacePlugin = ({
               return intentPlugin?.provides.intent.dispatch({
                 action: LayoutAction.OPEN_POPOVER,
                 data: {
-                  anchorId: `dxos.org/ui/navtree/${intent.data.object.id}`,
+                  anchorId: `dxos.org/ui/${intent.data.caller}/${intent.data.object.id}`,
```
wittjosiah, line 567: should the caller just pass the whole anchor id that it wants it anchored to?

packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
thure: It’s a horizontal layout, which I don’t think plugins can coordinate. I’d argue it’s not extensible by the design provided by LayoutPlugin, since it does not accommodate overflow.

If it should be unlimited, I think we basically need to redesign this surface.

packages/sdk/app-graph/src/action.ts
thure: Yeah I wasn’t entirely sure about this approach. We do a lot of this validation sniffing which is logically expensive and may be unnecessary here, so I figured this should probably assume less at the outset, hence the framework-level typedef.

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
thure: It’s a good question… I’m actually not sure. That seems like it would have a different set of consequences & use-cases.

PR #5077, opened 2024-01-09
fix(react-ui-editor): Make it easier to move the cursor into tables

marijnh:
With keyboard, by not making this a block widget (which is skipped in vertical arrow motion), and with mouse, by letting the editor handle mouse clicks on the table.


marijnh:
The CI failure is `fatal: Could not read from remote repository.`, which doesn't sound like an issue introduced by the patch. I don't seem to have the power to restart the test run, I think.

mykola-vrmchk:
> The CI failure is `fatal: Could not read from remote repository.`, which doesn't sound like an issue introduced by the patch. I don't have the power to restart the test run, I think.

@marijnh , I restarted it for you

marijnh:
Thanks. Not sure what "Validate PR title" is waiting for now, but the CI seems to have run.

@richburdon This is ready for review too.

thure (approved):
lgtm

PR #5099, opened 2024-01-10
feat(composer): Cmd-shift-delete to delete objects

jessmartin:
### Motivation

The keyboard shortcut of command-delete to delete an item has no confirmation or recovery. Very easy to accidentally delete an item and have no way of getting it back. command-delete is also the shortcut for "delete this line to the beginning of the line" in macOS, so I use it often.

### Details

- [x] changed shortcut to command-shift-delete
- [x] added a confirmation dialog box copying from the rename object pattern

### Visuals

https://github.com/dxos/dxos/assets/27258/1cd35766-8b26-487a-974b-ffb7974cb718

### DIFF
--- a/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
+++ b/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
@@ -270,9 +270,16 @@ export const SpacePlugin = ({
                 return <PopoverRenameObject object={data.subject} />;
               } else if (
                 data.component === 'dxos.org/plugin/space/RemoveObjectPopover' &&
-                isTypedObject(data.subject.object)
+                data.subject &&
+                typeof data.subject === 'object' &&
+                isTypedObject((data.subject as Record<string, any>)?.object)
               ) {
-                return <PopoverRemoveObject object={data.subject.object} folder={data.subject.folder} />;
+                return (
+                  <PopoverRemoveObject
+                    object={(data.subject as Record<string, any>)?.object}
+                    folder={(data.subject as Record<string, any>)?.folder}
+                  />
+                );
               } else {
                 return null;
               }
--- a/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
+++ b/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
@@ -28,7 +28,7 @@ import { LocalStorageStore } from '@dxos/local-storage';
 import { log } from '@dxos/log';
 import { Migrations } from '@dxos/migrations';
 import { type Client, PublicKey } from '@dxos/react-client';
-import { type Space, SpaceProxy, getSpaceForObject } from '@dxos/react-client/echo';
+import { type Space, SpaceProxy } from '@dxos/react-client/echo';
 import { useFileDownload } from '@dxos/react-ui';
 import { inferRecordOrder } from '@dxos/util';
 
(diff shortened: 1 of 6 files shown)


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -562,37 +575,20 @@ export const SpacePlugin = ({
...
+                  anchorId: `dxos.org/ui/${intent.data.caller}/${intent.data.object.id}`,
+                  component: 'dxos.org/plugin/space/RemoveObjectPopover',
+                  subject: {
+                    object: intent.data.object,
+                    folder: intent.data.folder,
+                  },
+                  caller: intent.data.caller,
+                },
+              });
+              // break;
```
wittjosiah, line 591: 👀

packages/apps/plugins/plugin-space/src/components/PopoverRemoveObject.tsx
```diff
@@ -0,0 +1,66 @@
...
+    // remove the object
+    if (space) {
+      space.db.remove(object);
+      return true;
+    }
+  }, [object]);
+
+  return (
+    <div role='none' className='p-1'>
+      <p className='mlb-1 mli-2'>Delete this item?</p>
```
wittjosiah, line 58: should add this to space plugin translations

packages/apps/plugins/plugin-space/src/components/PopoverRemoveObject.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    // If the item is active, navigate to "nowhere" to avoid navigating to a removed item
+    if (layoutPlugin?.provides.layout.active === object.id) {
+      await intentPlugin?.provides.intent.dispatch({
+        action: LayoutAction.ACTIVATE,
+        data: { id: undefined },
+      });
+    }
+
+    // remove a folder
```
wittjosiah, line 34: Remove object from the folder it's in

packages/apps/plugins/plugin-space/src/components/PopoverRemoveObject.tsx
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    const space = getSpaceForObject(object);
+
+    // remove the folder from the graph?
+    const folder = space?.properties[Folder.schema.typename];
+    if (folder instanceof Folder) {
+      const index = folder.objects.indexOf(object);
+      index !== -1 && folder.objects.splice(index, 1);
+    }
```
wittjosiah, line 47: This is probably covered by the previous condition now. This removes it from the root folder specifically.

packages/apps/plugins/plugin-space/src/components/PopoverRemoveObject.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    // remove the object
```
wittjosiah, line 49: Remove the object from the space

wittjosiah (approved):
Thanks for adding those comments 👍

packages/apps/plugins/plugin-space/src/components/PopoverRemoveObject.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { SPACE_PLUGIN } from '../meta';
+
+export const PopoverRemoveObject = ({ object, folder: propsFolder }: { object: TypedObject; folder: Folder }) => {
+  const { t } = useTranslation(SPACE_PLUGIN);
+  const deleteButton = useRef<HTMLButtonElement>(null);
+
+  const layoutPlugin = useResolvePlugin(parseLayoutPlugin);
+  const intentPlugin = useResolvePlugin(parseIntentPlugin);
+
+  const handleDelete = useCallback(async () => {
```
jessmartin, line 21: @wittjosiah shouldn't all of this delete behavior, taking into account folders and such, be moved somewhere else? feels weird for it to just be here. feels like this component should just call it.

wittjosiah: good question, this is sort of a consequence of the confirm popover, otherwise the logic would be in the intent. Perhaps we should consider a way to inject a confirm step into the intent process

PR #5100, opened 2024-01-10
feat(plugin-space): better navbar info

wittjosiah:
### Motivation / Background

This Pull Request improves the information we make visible about the current space in the navbar.

### Details

- adds a data persistence indicator
- updates the presence indicator to include all online peers

### Additional information

- viewers in presence is still broken, but that data will be used to sort the online peers once it is available

https://github.com/dxos/dxos/assets/4529818/5e4d6079-0f97-4c89-b42a-e3781edc20bb

image

image


packages/apps/composer-app/src/stories/NavtreePresence.stories.tsx
```diff
@@ -99,7 +99,7 @@ export const Demo = {
                 // @ts-ignore
                 presence: ({ data: { object } }: { data: { object: any } }) => {
                   return (
-                    <ObjectPresence size={2} viewers={object?.viewers ?? []} classNames={[fineBlockSize, 'is-6']} />
+                    <SmallPresence size={2} members={object?.viewers ?? []} classNames={[fineBlockSize, 'is-6']} />
```
thure, line 102: Generally I’d prefer a concept like `small` to be applied as an enum value of a prop like `variant` or `size` and be wrapped by a component that defers to internal components as needed, but this is fine for now.

packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
```diff
@@ -82,11 +82,9 @@ export const MainLayout = ({ fullscreen, showComplementarySidebar = true }: Main
                   <MenuIcon weight='light' className={getSize(4)} />
                 </Button>
 
-                <Surface role='navbar-start' limit={1} />
+                <Surface role='navbar-start' />
                 <div role='none' className='grow' />
-
-                {/* TODO(burdon): Too specific? status? contentinfo? */}
-                <Surface role='presence' limit={1} />
+                <Surface role='navbar-end' direction='inline-reverse' />
```
thure, line 87: Love it! Thank you ✨

packages/apps/plugins/plugin-space/src/components/PersistenceStatus.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <ArrowsCounterClockwise className={mx(getSize(4), 'me-1')} />
+          <span className={mx('text-sm', staticPlaceholderText)}>{t('persistence pending label')}</span>
+        </div>
+      );
+    case Status.PERSISTED_LOCALLY:
+    default:
+      return (
+        <Tooltip.Root delayDuration={400}>
+          <Tooltip.Trigger>
+            <div className='flex items-center'>
```
thure, line 71: Generic elements should generally have a role — I find this helps reduce div-itis, encourage thinking about semantic values & connecting to stuff, and it tidies-up the accessibility tree.

I think this is a real `role='status'` actually, so that applies here. From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role):

> A status is a type of [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) providing advisory information that is not important enough to justify an alert, which would immediately interrupt the announcement of a user's current activity.

thure: Also, `Tooltip.Trigger` will render a `<button>` — if this div is just for styling purposes, just remove the div and add props to the trigger:
```tsx
<Tooltip.Trigger role='status' classNames='flex items-center'>
```

packages/apps/plugins/plugin-space/src/components/ShareSpaceButton.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { withTheme } from '@dxos/storybook-utils';
+
+import { ShareSpaceButtonImpl } from './ShareSpaceButton';
+import translations from '../translations';
+
+export default {
+  title: 'plugin-space/ShareSpaceButton',
+  component: ShareSpaceButtonImpl,
+  decorators: [withTheme],
+  parameters: { translations },
```
thure, line 16: Love this pattern!

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -2,118 +2,176 @@
...
+    <AvatarGroup.Root size={size} classNames='mbs-2 mie-4'>
+      {members.slice(0, 3).map((member, i) => {
+        const memberHex = member.identity.identityKey.toHex();
+        const status = member.match ? 'current' : 'active';
+        const name = member.identity.profile?.displayName ?? generateName(member.identity.identityKey.toHex());
+        return (
+          <Tooltip.Root key={memberHex}>
+            <Tooltip.Trigger>
+              <AvatarGroupItem.Root color={getColorForValue({ value: memberHex, type: 'color' })} status={status}>
+                <Avatar.Frame style={{ zIndex: members.length - i }} onClick={() => onMemberClick?.(member)}>
```
thure, line 111: There’s a way to handle this without messing with `zIndex`, but I forget… if this didn’t work out of the box without doing this, this is fine for now, it just may cause problems because there are other elements on the page with low z-indexes.

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -2,118 +2,176 @@
...
+                  <Avatar.Fallback text={name[0]} />
+                </Avatar.Frame>
+              </AvatarGroupItem.Root>
+            </Tooltip.Trigger>
+            <Tooltip.Portal>
+              <Tooltip.Content side='bottom' classNames='z-[70]'>
```
thure, line 117: I think `z-[70]` isn’t necessary if it’s portaled, but I don’t rightly recall… either way this doesn’t add much tech debt, there are plenty of other instances of this we should decruft.

packages/apps/plugins/plugin-space/src/translations.ts
```diff
@@ -29,8 +29,12 @@ export default [
...
         'confirm restore body': 'Restoring from a backup will overwrite the contents of any documents that match.',
         'upload file message': 'Drag file here or click to browse',
-        'presence label_one': '{{count}} other viewing this item',
-        'presence label_other': '{{count}} others viewing this item',
+        'viewers label_zero': '',
+        'viewers label_one': '{{count}} peer viewing this item & ',
+        'viewers label_other': '{{count}} peers viewing this item & ',
+        'members online label_zero': 'no other members online',
+        'members online label_one': '{{count}} other member online',
+        'members online label_other': '{{count}} other members online',
```
thure, line 37: Very nice! Well done.

thure (approved):
Lgtm! Just some markup & z-indexes worth thinking about, but not blocking.

packages/apps/composer-app/src/stories/NavtreePresence.stories.tsx
wittjosiah: Yeah, tbh I think these might even be two separate components or the small one might not even be long for this world so I left this close-ish to what it was previously. Makes sense though 👍

packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
wittjosiah: FYI I didn't end up hooking this up inside `Surface` yet but you get the idea. If there comes a time that you actually need this working then let me know and we can find where it should inside `Surface`.

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
wittjosiah: This was how this control was done previously, I didn't add this so I'm not exactly sure of it's intended purpose.

PR #5103, opened 2024-01-11
feat: support for cross-platform keybindings

wittjosiah:
The keybindings within Composer are currently very focused towards macOS. The web is cross-platform so we want to be able to support keybindings across platforms. This PR introduces the utils we need to be able to create cross-platform keybindings.


packages/common/keyboard/src/util.ts
```diff
@@ -2,28 +2,68 @@
...
-const symbols: Record<string, string> = {
-  // Mods.
-  ctrl: '⌃',
-  shift: '⇧',
-  alt: '⌥',
-  meta: '⌘',
+const alt: Record<string, string> = {
+  macos: '⌥',
+  ios: '⌥',
+  windows: 'Alt',
```
richburdon, line 15: the shortcut dialog current attempts to render these in a 24x24px container, which i guess wasn't going to work for everything anyway...

packages/common/keyboard/src/util.ts
```diff
@@ -2,28 +2,68 @@
...
+  linux: 'Alt',
+  unknown: 'Alt',
```
richburdon, line 17: just designate linux as the default when looking up?

richburdon (approved):

packages/common/keyboard/src/util.ts
wittjosiah: Platform lookup is based on this an includes unknown, but I've added fallbacks such that not all platforms need to be provided to support all platforms.

PR #5105, opened 2024-01-11
feat: send a target with space invitations

### LINKED ISSUE #3968: [shell] Space invites should append to the end of the current app URL
## Motivation

When sending someone a space invite, the user's desired action is often that the person they are sending the space invitation to would be redirected to the space of object that is being shared or currently viewed. The "web friendly" way to do this is to use a URL that can be routed to.

## Finish Line

Currently, when a Space Invitation is created by Shell, it truncates the URL and doesn't preserve the context of where the user was when they extended the invite.

- Shell should merely append the `?spaceInvitationCode=xxxx` to the end of the URL
- When we clear the `spaceInvitationCode` once the invite flow is complete, we should not clear the entire URL, but merely remove that query param

wittjosiah:
### Motivation

When sharing a space there's often a specific object within the space that is of focus immediately. This PR adds a field to invitations which allows a target within the space to be specified for focus after authentication. Resolves #3968 

### Details

- adds `target` to invitation protobuf
- exposes `target` from `Shell` apis
- uses `target` within Composer to automatically navigate to the object that was open when the invitation was sent


packages/apps/plugins/plugin-client/src/ClientPlugin.tsx
```diff
@@ -257,19 +247,6 @@ export const ClientPlugin = ({
 
             case ClientAction.SHARE_IDENTITY:
               return client.shell.shareIdentity();
-
```
richburdon, line 260: +1

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -407,6 +407,9 @@ message Invitation {
 
   /// Authentication code created by host (only present if auth_method is SHARED_SECRET).
   optional string auth_code = 10;
+
+  /// Path or identifier to navigate to after successful authentication.
+  optional string target = 11;
```
richburdon, line 412: ok, so purposefully leaving this ambiguous (rather than path or object) so we can change/interpret it's meaning?

richburdon (approved):

packages/core/protocols/src/proto/dxos/client/services.proto
wittjosiah: Yeah, exactly

PR #5202, opened 2024-01-14
feat: Comments demo

richburdon:


packages/apps/plugins/plugin-client/src/ClientPlugin.tsx
```diff
@@ -81,7 +81,8 @@ export const ClientPlugin = ({
   // TODO(burdon): Document.
   registerSignalFactory();
 
-  const settings = new LocalStorageStore<ClientSettingsProps>('dxos.org/settings');
+  const settings = new LocalStorageStore<ClientSettingsProps>('dxos.org/settings', { automerge: true });
```
jessmartin, line 84: Does this now make Automerge the default backend? I know this was discussed on Friday during a call, but probably worth calling out. Maybe in PR description?

PR #5212, opened 2024-01-15
fix: Repo guide re vault/storybooks

richburdon:


REPOSITORY_GUIDE.md
```diff
@@ -68,6 +68,14 @@ Examples of ways to run different test workloads:
 | `pnpm nx e2e halo-app` | Runs the playwright tests for `halo-app` |
 | `pnpm nx e2e halo-app --inspect` | Runs tests with playwright inspector |
 
+## Storybooks
+
+All Storybooks can be run with the following command:
```
wittjosiah, line 73:
```suggestion
Most stories are available in the storybook run with the following command:
```

There's a few other storybooks we have as well (e.g., examples and e2e)

jessmartin (approved):
There are a few mentions of vault in the last two sections. I couldn't determine whether they are relevant any longer, but they are worth looking at.

Other than that, looks good!

richburdon:
> There are a few mentions of vault in the last two sections. I couldn't determine whether they are relevant any longer, but they are worth looking at.
> 
> Other than that, looks good!

I'll defer to @wittjosiah on that perhaps for another pass.

PR #5218, opened 2024-01-15
fix(plugin-presenter): fix for render documents

### LINKED ISSUE #5194: [presenter] present mode is broken, documents render as blank slides

jessmartin:
The type check for Markdown content changed, breaking Documents in presenter plugin. I added an alternative type check, but I'm sure it can be improved.

fixes #5194

### DIFF
--- a/packages/apps/plugins/plugin-presenter/src/PresenterPlugin.tsx
+++ b/packages/apps/plugins/plugin-presenter/src/PresenterPlugin.tsx
@@ -6,7 +6,7 @@ import { Presentation } from '@phosphor-icons/react';
 import { deepSignal } from 'deepsignal';
 import React from 'react';
 
-import { isMarkdown, isMarkdownContent } from '@braneframe/plugin-markdown';
+import { isMarkdownContent } from '@braneframe/plugin-markdown';
 import { isStack } from '@braneframe/plugin-stack';
 import { resolvePlugin, type PluginDefinition, parseIntentPlugin, LayoutAction } from '@dxos/app-framework';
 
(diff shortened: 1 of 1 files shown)


packages/apps/plugins/plugin-presenter/src/PresenterPlugin.tsx
```diff
@@ -22,6 +23,11 @@ type PresenterState = {
   presenting: boolean;
 };
 
+const isLocalMarkdownContent = (data: unknown): data is { content: EditorModel } =>
```
jessmartin, line 26: @wittjosiah this is most certainly incorrect. could probably be more terse. move back into `util.tsx`? although having it in `util.tsx` is probably what broke this in the first place - wasn't clear all of the places that it's being used.

dmaretskyi, line 26: Not sure, but I think EditorModel shouldn't appear in ECHO objects. cc @wittjosiah 

```suggestion
const isLocalMarkdownContent = (data: unknown): data is { content: TextObject } =>
```

packages/apps/plugins/plugin-presenter/src/PresenterPlugin.tsx
```diff
@@ -14,6 +14,7 @@ import { PresenterMain, MarkdownSlideMain } from './components';
 import meta, { PRESENTER_PLUGIN } from './meta';
 import translations from './translations';
 import { PresenterContext, type PresenterPluginProvides } from './types';
+import { type EditorModel } from '../../../../ui/react-ui-editor/src';
```
dmaretskyi, line 17:
```suggestion
import { type EditorModel } from '@dxos/react-ui-editor';
```

packages/apps/plugins/plugin-presenter/src/PresenterPlugin.tsx
dmaretskyi, line 26: I usually don't bother with exact types and do `data: any` in the parameters so that the code below is easier to read

jessmartin:
closing in favor of #5248

PR #5220, opened 2024-01-16
feat: Enable automerge by default

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-schema/src/automerge-preference.ts
```diff
@@ -0,0 +1,39 @@
+//
```
richburdon, line 1: lint

Issue #5184, opened 2024-01-13
[authoring] CodeMirror tasklist


richburdon:
Illustrated in the video below:

- [x] Bug: Cursor jumps when moving down into block (if in first column)
- [x] Bug: Formatting breaks when pressing Enter in front of a bullet that has indented children
- [x] Convert indents to natural spaces
- [x] Remove checkbox widget (convert back to text) when cursor moves into it (popular request; and similar to Obsidian)
- [x] Pressing Enter on an empty last line of the list should end the list (currently have to press Enter three times)
- [x] Restore previous position of cursor after clicking on checkbox

https://github.com/dxos/dxos/assets/3523355/e76f6b22-958d-408f-866b-2ed22a5dd290

richburdon:
I get the point, but I do think the behavior implemented __here__ in this GitHub textbox is more intuitive.

```
- [ ] 1
- [ ] 2
- [ ] 3
- [ ]               << pressing enter here will end the list.
```

marijnh:
One thing I've been considering is to make that command a bit smarter so that, for the first list item you get the current behavior (single-item lists not super common), but then for others it only goes through the spacing of the rest of the items (no blank line if the list is tight, blank line if not), so you can exit it with two enter presses. Does that sound like it's work for you here?

marijnh:
Well, [here](https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-editor/src/themes/default.ts#L166) is the reason your widgets misbehave and break the cursor. Patch 560f82b5d9 , for some reason, intentionally breaks CodeMirror's hack that prevents precisely this.

marijnh:
Upgrading to @codemirror/lang-markdown 6.2.4 should help with `Enter` key behavior. Is bumping a dependency just a matter of updating the relevant package's package.json, reinstalling with pnpm to update the package lock, and committing the result?

marijnh:
@richburdon Please see https://github.com/dxos/dxos/pull/5221, https://github.com/dxos/dxos/pull/5222, and https://github.com/dxos/dxos/pull/5223

jessmartin:
@richburdon the entire checklist is checked off. can this issue be closed now?

PR #5221, opened 2024-01-16
fix(react-ui-editor): Make task list checkboxes less invasive

marijnh:
- Only replaces the actual checkbox
 - Don't mess with focus or cursor position when toggled
 - Expand to text when the cursor moves into them
 - Use the syntax tree, not a regexp, to find the checkboxes

Issue https://github.com/dxos/dxos/issues/5184


packages/ui/react-ui-editor/src/extensions/tasklist.ts
```diff
@@ -2,118 +2,94 @@
...
-    if (!this._onCheck) {
+    if (view.state.readOnly) {
       input.setAttribute('disabled', 'true');
-    }
-    if (this._onCheck) {
-      input.onchange = (event: Event) => {
-        this._onCheck?.((event.target as any).checked);
-        return true;
+    } else {
+      input.onmousedown = (event: Event) => {
```
richburdon, line 25: curious why you chose this over onclick?
my understanding is onclick would also catch the case where the input is focused via the keyboard and then pressing enter?

richburdon (approved):

packages/ui/react-ui-editor/src/extensions/tasklist.ts
marijnh: Once click (or change, or mouseup) fires, the focus has already been moved to the element, and lost the editor. Since I guess our keyboard store for these is going to be regular text editing, now that they expand when you move the cursor to them, I didn't really account for keyboard interaction.

PR #5222, opened 2024-01-16
fix(react-ui-editor): Don't hide widget buffers

marijnh:
Widget buffers are a thing CodeMirror inserts into the editable DOM to prevent a bunch of browser-contenteditable misbehaviors when the cursor is next to uneditable content, such as invisible cursors, weird cursor motion, and IME input weirdly interacting with the widgets. They should not be set to `display: none`.

Issue https://github.com/dxos/dxos/issues/5184

### DIFF
--- a/packages/ui/react-ui-editor/src/themes/default.ts
+++ b/packages/ui/react-ui-editor/src/themes/default.ts
@@ -161,14 +161,6 @@ export const defaultTheme: ThemeStyles = {
   },
 
   //
-  // widgets
-  //
-  '.cm-widgetBuffer': {
-    display: 'none',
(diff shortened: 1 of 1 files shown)


richburdon:
> Widget buffers are a thing CodeMirror inserts into the editable DOM to prevent a bunch of browser-contenteditable misbehaviors when the cursor is next to uneditable content, such as invisible cursors, weird cursor motion, and IME input weirdly interacting with the widgets. They should not be set to `display: none`.

got it. i think it was seeing some unwanted padding, but i can't remember where this was now.

richburdon (approved):

PR #5248, opened 2024-01-16
fix(plugin-presenter): render documents

jessmartin:
The type check for Markdown content changed, breaking Documents in presenter plugin. I added an alternative type check, but I'm sure it can be improved.

fixes https://github.com/dxos/dxos/issues/5194

### DIFF
--- a/packages/apps/plugins/plugin-presenter/src/components/MarkdownSlideMain.tsx
+++ b/packages/apps/plugins/plugin-presenter/src/components/MarkdownSlideMain.tsx
@@ -11,7 +11,8 @@ import { Container, Slide } from './Markdown';
 export const MarkdownSlideMain: FC<{ document: DocumentType }> = ({ document }) => {
   return (
     <Container>
-      <Slide content={document.content.text} />;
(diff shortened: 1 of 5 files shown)


wittjosiah (approved):

packages/apps/plugins/plugin-presenter/src/components/MarkdownSlideMain.tsx
```diff
@@ -4,12 +4,15 @@
...
+
 import { Container, Slide } from './Markdown';
 
-export const MarkdownSlideMain: FC<{ slide: any }> = ({ slide }) => {
+export const MarkdownSlideMain: FC<{ document: DocumentType }> = ({ document }) => {
   return (
     <Container>
-      <Slide content={String(slide.content)} />;
+      {/* TODO: content is a YText object, but should be a string. Update after automerge migration. */}
+      <Slide content={document.content.content as any} />;
```
jessmartin, line 15: @dmaretskyi ended up having to typecast this YText object to any in order to match types. I assume this will all change very shortly here. I've tested this and it works with Automerge documents.

PR #5281, opened 2024-01-17
test(composer-app): signout

### LINKED ISSUE #4402: [composer] device invite flow succeeds, but requires app refresh to load
Seems like some errors are getting thrown after the Device Invite flow that crashes Composer. All seems to work after cmd-R.

Repros:
- Invite another device
- Watch it hang when it succeeds

image

### LINKED ISSUE #4847: [halo] device join flows
## Motivation

_Everything_ is a device join now! Because browsers. We introduced progressive multiplayer which sets up an identity automatically, but we never introduced the ability to merge an existing profile into a device join.

## Finish Line

- When you're receiving a device invitation, check if identity exists, and if it does, prompt with the device reset flow, reset the identity, join the invited identity

wittjosiah:
Introduces playwright tests for the recently added signout capabilities in shell.

Resolves #4402 
Resolves #4847


packages/apps/composer-app/src/playwright/app-manager.ts
```diff
@@ -38,6 +40,14 @@ export class AppManager {
     this._initialized = true;
   }
 
+  async openIdentityManager() {
+    await this.page.keyboard.press('Meta+Shift+.');
+  }
+
+  async openSpaceManager() {
+    await this.page.keyboard.press('Meta+.');
```
thure, line 48: Oh this is neat, I didn’t think the API would work this way. Very cool.

thure (approved):
Beautiful ✨

PR #5288, opened 2024-01-17
feat(app-framework): integrate error boundary w/ plugin host

### LINKED ISSUE #4977: [composer] handle plugin errors more gracefully
While working on native apps for Composer, we ran into an issue with the IPFS plugin. The error causes the whole app to not work. A plugin failure should not cause the whole app to stop working.

[Discord conversation for context](https://discord.com/channels/837138313172353095/1180035274345025617/1187031550232113215)

### Steps to reproduce

* Use the code from [this commit](https://github.com/dxos/dxos/pull/4960/commits/5c44b9ce167e723c905d79434907aac3d0fc8feb)
* Build Composer and SSC native app
* Enable the IPFS plugin then quit and reopen the app

wittjosiah:
### Motivation / Background

This PR improves the error handling of @dxos/app-framework such that it should catch more app startup errors.

### Details

- moves the error boundary to the root of the plugin host
- introduces a default plain error boundary fallback that can be customized via the plugin host
- removes the error plugin and refactors the reset dialog to composer
- reconnects reset button in dialog
- client plugin now defers to built in error bubbling by throwing in `ready` if client fails to initialize
  - it still catches errors from initialization because plugins which fail to initialize are simply filtered out

### Additional information

- Resolves #4977

Default fallback

Screenshot 2024-01-17 at 18 44 44

Composer fallback

Screenshot 2024-01-17 at 18 45 14


packages/sdk/app-framework/src/plugins/PluginHost/PluginHost.tsx
```diff
@@ -63,65 +66,100 @@ export const PluginHost = ({
...
+
+  if (error) {
+    throw error;
+  }
+
+  if (!state.ready) {
+    return <>{placeholder}</>;
+  }
+
+  const ComposedContext = composeContext(state.plugins);
```
thure, line 158: This is an antipattern — we should avoid as much as possible defining components within the render function of a component, as it causes reactivity issues.

I can see this was present in the previous version of the code, but worth keeping an eye on this instance.

thure (approved):
Keep an eye on the antipattern of defining components to be rendered in a parent component’s render function, otherwise lgtm 👍

packages/sdk/app-framework/src/plugins/PluginHost/PluginHost.tsx
```diff
@@ -63,65 +66,100 @@ export const PluginHost = ({
...
 };
 
+const DefaultFallback = ({ error }: { error: Error }) => {
+  return (
+    <div style={{ padding: '1rem' }}>
+      {/* TODO(wittjosiah): Link to docs. */}
+      <p style={{ fontSize: '0.8rem', fontStyle: 'italic' }}>
+        This is the default ErrorBoundary fallback. Consider providing your own!
+      </p>
+      <h1 style={{ fontSize: '1.2rem', fontWeight: 700, margin: '0.5rem 0' }}>{error.message}</h1>
```
richburdon, line 86: Is this prior to tailwind being available? Otherwise use classes?

packages/sdk/app-framework/src/plugins/PluginHost/PluginHost.tsx
```diff
@@ -63,65 +66,100 @@ export const PluginHost = ({
...
     },
   };
 };
 
+const DefaultFallback = ({ error }: { error: Error }) => {
+  return (
+    <div style={{ padding: '1rem' }}>
+      {/* TODO(wittjosiah): Link to docs. */}
+      <p style={{ fontSize: '0.8rem', fontStyle: 'italic' }}>
+        This is the default ErrorBoundary fallback. Consider providing your own!
```
richburdon, line 84: Not a fan of conversational defaults. Instead provide an informative and sensible default and document how to change it in the docs.

packages/sdk/app-framework/src/plugins/PluginHost/PluginHost.tsx
```diff
@@ -63,65 +66,100 @@ export const PluginHost = ({
...
+      <h1 style={{ fontSize: '1.2rem', fontWeight: 700, margin: '0.5rem 0' }}>{error.message}</h1>
+      <pre>{error.stack}</pre>
+    </div>
+  );
+};
+
+type RootProps = {
+  order: PluginDefinition['meta'][];
+  state: PluginContext;
+  definitions: Record<string, () => Promise<PluginDefinition>>;
```
richburdon, line 95: definitions => plugins
core => corePlugins
?

Issue #5040, opened 2024-01-05
[authoring] comments on documents


richburdon:
## Product features

- [ ] Show username in comment ( #5323 )
- [ ] Editable comments ( #5161 )
- [ ] Resolve comments (with undo) ( #5162 )
- [ ] Show comments next to stack ( #5324 )

## Low-level

- [x] Storybook: https://github.com/dxos/dxos/pull/5199
- [ ] Preserve comment range/thread when cut/paste.
  - [ ] Detect thread deleted (mark for garbage collection).
- [ ] Focus thread on creation (currently required setTimeout hack.)
- [ ] When selecting a thread, highlight the associated text (scroll to, etc.)
- [x] When selecting a comment range, scroll to the associated thread.
- [x] Delete/resolve thread.
- [x] Button to resolve thread.
- [x] RB: Fix theme/slots for TextEditor (currently applied to container div not cm-scroller).
  - [ ] Would be useful to be able to apply className to `dom` and `scrollerDom` divs.
  - [ ] Scrolling is broken (i.e., doesn't scroll properly in Composer when navigating with the cursor).
  - [ ] Fix gaps between lines with comment bg class (`span` is not as high as `div`)
- [x] Consider how to not require comments to be passed to TextEditor (instead update extension?)
  - i.e., https://github.com/dxos/dxos/pull/5199/files#diff-b00930ab8b3935dc62edbd2f9df9136934286d2c57d50d8c45f260a6fb3a84d5L99

## Story

```bash
pnpm -w nx storybook stories
```

```
http://localhost:9009/iframe.html?args=&id=react-ui-editor-comments--default&viewMode=story
```

https://github.com/dxos/dxos/assets/3523355/695434eb-434b-40d2-9289-7c3b1059bc28

marijnh:
I'm going to try and use this issue to get an understanding of the set of requirements for the comment system. @richburdon, maybe you could answer the questions below and point out where I'm off with my assumptions.

I guess the data model for comments is a set of commented ranges, each with a resolved flag and an array of comment messages.
 - Where are these maintained? I would guess in the CRDT. How do they flow from there into the editor (including updates after the editor has been initialized). Should the editor track them separately as edits happen, or re-sync with the CRDT on every update?
 - Are messages plain text? Or should they also support Markdown?
 - Should deleting the last remaining message in a thread remove the entire comment? Or do we provide some alternative control for that?
[...]

PR #5294, opened 2024-01-18
feat: Move comments when cut and pasted

marijnh:
Issue https://github.com/dxos/dxos/issues/5040

This implements the feature where cutting and pasting comments will move them in the text. Requires consumers to provide an onMove callback Wires up the comments story to use this, but does not do this for the threads plugin yet.

### DIFF
--- a/packages/ui/react-ui-editor/src/extensions/comments.ts
+++ b/packages/ui/react-ui-editor/src/extensions/comments.ts
@@ -2,7 +2,7 @@
 // Copyright 2023 DXOS.org
 //
 
-import { type Extension, StateEffect, StateField } from '@codemirror/state';
+import { type Extension, StateEffect, StateField, type Text } from '@codemirror/state';
 import { hoverTooltip, keymap, type Command, Decoration, EditorView, type Rect } from '@codemirror/view';
 import sortBy from 'lodash.sortby';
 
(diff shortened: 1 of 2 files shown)


packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -151,6 +155,65 @@ export type CommentsOptions = {
   onHover?: (el: Element, shortcut: string) => void;
 };
 
+const trackPastedComments = (onMove: (threadID: string, cursor: string) => void) => {
```
richburdon, line 158: exposing onCut would be useful to allow the app to mark comments that are to be potentially deleted.

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -151,6 +155,65 @@ export type CommentsOptions = {
...
+      tracked = {
+        text: view.state.doc.slice(main.from, main.to),
+        comments: inSel.map((range) => ({ from: range.from - main.from, to: range.to - main.from, id: range.id })),
+      };
+    }
+  };
+
+  return [
+    EditorView.domEventHandlers({
+      cut: registerCopy,
```
richburdon, line 179: How can we handle undo?
E.g., cut paragraph containing selection, then ctrl-z.
Potentially with history (i.e., multiple times).

richburdon (approved):

richburdon:
We use https://github.com/marketplace/actions/semantic-pull-request for the PR titles.
I'll change to `feat`.

packages/ui/react-ui-editor/src/extensions/comments.ts
marijnh: Aren't those just the comments that cover an empty range?

PR #5298, opened 2024-01-18
fix(plugin-space): hide share button in personal space

wittjosiah:
Addresses regression where the share button was being rendered for the personal space in composer.


thure (approved):
\*nodding\*

PR #5300, opened 2024-01-18
chore: dedupe faker

wittjosiah:
The debug plugin was pulling in multiple copies of faker which was causing the memory usage of bundling composer get even more ridiculous. This dedupes the faker dependency to reign it back in a bit.


thure (approved):
lgtm

PR #5316, opened 2024-01-19
feat(client-services): add already joined error to invitations

### LINKED ISSUE #1922: [sdk] SDK errors for joining own space
need a sensible error from client when trying to join own space from same profile. This is to help applications deal with situations when an invite code gets pasted into the same browser profile from where the invite came from orignally.

### LINKED ISSUE #3568: [composer] joining a space you are already a member of shouldn't be a failure
i joined my mobile device to a random space from composer and it turned out my mobile device had already been a member. I was taken through the pin code process only to find out "failed" and upon dismissing I saw the space was already there. 

This should have "succeeded" sooner by navigating to the first document in that space and popping a non-modal (toast) that the "you are already a member of this space".

IMG_0561

wittjosiah:
Adds an appropriate ALREADY_JOINED error when attempting to accept an invitation to a resource that a client already has access to.

Resolves #1922
Resolves #3568


thure (approved):
Top notch ✨

packages/sdk/client-services/src/packlets/invitations/device-invitation-protocol.ts
```diff
@@ -46,6 +47,17 @@ export class DeviceInvitationProtocol implements InvitationProtocol {
     };
   }
 
+  checkInvitation(invitation: Partial<Invitation>) {
+    try {
+      const identity = this._getIdentity();
+      if (identity) {
+        return new AlreadyJoinedError('Currently only one identity per client is supported.');
```
dmaretskyi, line 54: returning errors feels kinda odd, why not throw?

wittjosiah: I started by throwing and was having issues w/ it. In the end I don't think that was the issue but I kept it because throwing seemed to generate a console message even if it was caught? Would you prefer if this was expected to throw if there's a problem and have a return type of `void`?

richburdon: IMO @dmaretskyi 
- throw is something in our code went wrong.
- this is a protocol level expected result so we should return an error.

packages/sdk/client-services/src/packlets/invitations/invitation-protocol.ts
```diff
@@ -2,24 +2,62 @@
...
+
+  /**
+   * Get profile information to send to the host to identify the guest.
+   */
   createIntroduction(): IntroductionRequest;
+
+  /**
+   * Get key information to send to the host in order to create an admission credential for the guest.
+   */
   createAdmissionRequest(): Promise<AdmissionRequest>;
```
richburdon, line 57: Why add Request suffix -- others do not.

richburdon (commented):
Glad to have this.
Please test with the CLI/agent joining a space; it should not throw and exception, but instead log the request failed.

PR #5327, opened 2024-01-20
chore: Codemirror comments/state

richburdon:
## Important
- [x] Factor out automerge sync dependency via syncFacet.
- [x] Remove automerge view plugin; instead trigger sync in dispatch.
- [x] Simplified state management. 

## Misc
- [x] Autofocus thread when first creating a comment.
- [x] If creating comment at the bottom of the document, append a newline so that the comment range isn't sticky to the end.
- [x] Don't allow selection of deleted ranges.

### DIFF
--- a/packages/ui/react-ui-editor/src/extensions/automerge/semaphore.ts
+++ b/packages/ui/react-ui-editor/src/extensions/automerge/semaphore.ts
@@ -20,11 +20,12 @@ import {
 } from './defs';
 import { updateAutomerge } from './update-automerge';
 import { updateCodeMirror } from './update-codemirror';
+import { type Syncer } from '../sync';
 
 /**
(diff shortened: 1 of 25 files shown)


packages/ui/react-ui-editor/src/extensions/automerge/semaphore.ts
```diff
@@ -58,7 +65,7 @@ export class PatchSemaphore {
 
     // NOTE: null and undefined each come from automerge and repo respectively.
     if (newHeads === null || newHeads === undefined) {
-      // TODO: @alexjg this is the call that's resetting the editor state on click
+      // TODO(alexjg): this is the call that's resetting the editor state on click.
```
wittjosiah, line 68: noting that I'd likely interpret this as a todo _from_ alex rather than as a note to pass on to him

PR #5343, opened 2024-01-22
fix(keyboard): make shortcut parsing case insenstiive

wittjosiah:
This PR makes keybindings provided to graph actions be case-insensitive.


packages/common/keyboard/src/util.ts
```diff
@@ -39,7 +39,7 @@ const meta: Record<string, string> = {
 
 const getSymbol = (part: string) => {
   const platform = getHostPlatform();
-  switch (part) {
+  switch (part.toLowerCase()) {
```
richburdon, line 42: the special keys are standard capitalized names so I would tend to keep them as such?
https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

wittjosiah: The _output_ (line 64) is uppercase, this is just lowercasing in order to make the _comparison_ case insensitive.

This ensures that `meta+backspace` works the same as `Meta+Backspace` and `ctrl+k` works the same as `Ctrl+K`, because we're not enforcing that and those are frustrating things to track down.

PR #5352, opened 2024-01-23
chore: remove YJS deps from text model.

richburdon:


packages/ui/react-ui-editor/src/components/TextEditor/codemirror.stories.ts
wittjosiah, line 1: This story used to be separate and had it's own storybook because it's html-only and allowed us to diagnose and distinguish between echo integration issues & and react issues. If you think we're beyond that then I'm fine with this going away.

packages/ui/react-ui-editor/src/testing/replicator.ts
wittjosiah, line 1: This was used for yjs-specific stories w/o ECHO. Do we have automerge-specific stories w/o ECHO as well now?

wittjosiah (approved):

packages/ui/react-ui-editor/src/testing/replicator.ts
richburdon: @dmaretskyi what do you think?

PR #5371, opened 2024-01-24
feat(app-framework): optional suspense boundary for surfaces

### LINKED ISSUE #5313: [composer] Wrap plugin components in React.lazy and add Suspense boundaries
- target `main` surface to start, that is where all the heavy components will render

wittjosiah:
### Motivation / Background

The goal is to allow plugins to lazily load heavy components. Currently plugins are all loaded at app startup (if they are enabled) but this will allow components to be loaded lazily and lighten the initial load for apps.

Resolves #5313

### Details

- add optional suspense boundary to `Surface`
- uses suspense boundary for content areas (main, section, card, slide)
- updates all content plugin components to be lazy


dmaretskyi (commented):
Nice, did you notice changes in the output chunks/load times?

wittjosiah:
Screenshot 2024-01-24 at 20 57 32
Screenshot 2024-01-24 at 21 05 17

@dmaretskyi seems to have helped to split the largest chunk from debug plugin, yes

packages/apps/plugins/plugin-markdown/src/components/Layout.tsx
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import React, { type PropsWithChildren } from 'react';
+
+import { Main } from '@dxos/react-ui';
+import { baseSurface, mx, textBlockWidth, topbarBlockPaddingStart } from '@dxos/react-ui-theme';
+
+export const MainLayout = ({ children }: PropsWithChildren) => {
```
richburdon, line 10: note to think about factoring out common containers across plugins for uniformity.

richburdon:
Do scripts still work with this change?

packages/apps/plugins/plugin-markdown/src/components/Layout.tsx
wittjosiah: I've wondered about this. @thure do you think that the layout plugin should just include these styles around the main surface?

wittjosiah:
> Do scripts still work with this change?

I've not tested scripts specifically, but all this is doing is lazy-loading the react components, so I don't think it should effect it.

packages/apps/plugins/plugin-markdown/src/components/Layout.tsx
thure: I’m not entirely sure I’m on the same page as either of you… I think we have a meaningful separation of concerns between `Main.Content` and however an app plugin might want to lay out content in there.

1) `flex flex-col` can almost always be removed (block layout is usually fine and introduces less risk of introducing subtle bugs)
  - in this case it’s used so the viewport height is filled, but that is redundant with the `bounce` flag on `Main.Content`
2) `textBlockWidth` is a meaningful extrinsic size that doesn’t need to be a component because it is more useful as a fragment
3) Everything else here is spacing, which is highly sensitive to context.
4) AFACIT, at least one of these `<div>`s can be removed.

thure: `Main.Content` probably doesn’t need `baseSurface` either, since it is never superimposed onto anything.

wittjosiah: > Everything else here is spacing, which is highly sensitive to context

I guess this is why I was wondering if the layout should wrap the main surface w/ a container that can help. Otherwise the plugins need to know about the height of the navbar, whether the sidebar is open or not, how big is the sidebar, etc.

PR #5373, opened 2024-01-24
feat(composer-app): introduce separate devtools entrypoint

### LINKED ISSUE #5225: Add devtools app as a separate HTML entrypoint to composer

wittjosiah:
### Motivation / Background

This PR adds the devtools app as a standalone entrypoint to Composer. This is primarily intended for use in situations when Composer doesn't startup, since devtools is also available within Composer itself.

Previously this wasn't useful because of the identity being stored on a separate origin but since the in-browser vault deprecation this is useful for accessing the identity on Composer's origin. While there is also the devtools browser extension, that does not yet have the ability to inspect if the app fails to load.

### Details

Resolves #5225


jessmartin:
Hurrah for this!

Issue #4900, opened 2023-12-12
[sdk] bug in iterating over arrays can lead to internal reference objects corrupting user data


dmaretskyi:
https://github.com/dxos/dxos/pull/4757

jessmartin:
this is blocking @zhenyasav's ability to recover his broken folders

PR #5377, opened 2024-01-24
fix: bug in iterating over arrays can lead to internal reference objects corrupting user data

yarolegovich:
### Motivation / Background

Fixes #4757 #4900

### Details

The [documented flatMap behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#calling_flatmap_on_non-array_objects) is as follows:
> // Array-like objects returned from the callback won't be flattened

It seems impractical and unrealistic to expect users to know and remember this, so instead of asking them to cast EchoArray to a regular array whenever its returned as flatMap callback result the suggestion is to use to at least cover the (seemingly) most common case: `query.objects.flatMap(({ v }) => v)` where v is an EchoArray.

To avoid confusion made `objects` and `results` type explicit. By confusion I mean understanding why `query.objects.flatMap(...)` works, but `[...query.objects].flatMap(...)` doesn't.

### Additional information

In the discussion of the referenced issue it was mentioned that Array.isArray check doesn't work for EchoArray. I'm not sure there's a practical benefit of making it return true so there's no code for that in the PR. I think it only increases complexity and introduces unnecessary hacks. 

An alternative suggested was to use `new Proxy([], echoArray)`. But for this to work handler should not intercept mutations to update its data which `EchoArray` does. So unless we keep all the items in the array as well flatMap won't work. The approach sounds very hackish.

Not sure if this logic conceptually belongs to EchoArray and wrapping QueryResult with it is appropriate. Won't object to keeping this logic scoped to query and QueryResult.


packages/core/echo/echo-schema/src/automerge/automerge-array.ts
```diff
@@ -277,7 +277,11 @@ export class AutomergeArray<T> implements Array<T> {
     callback: (this: This, value: T, index: number, array: T[]) => U | readonly U[],
     thisArg?: This | undefined,
   ): U[] {
-    return Array.from(this.values()).flatMap(callback, thisArg);
+    const array = Array.from(this.values());
+    return array.flatMap(function (this: This, value, index, arr) {
```
dmaretskyi, line 281: 👍

packages/core/echo/echo-schema/src/object/array.ts
```diff
@@ -284,7 +284,11 @@ export class EchoArray<T> implements Array<T> {
     callback: (this: This, value: T, index: number, array: T[]) => U | readonly U[],
     thisArg?: This | undefined,
   ): U[] {
-    return Array.from(this.values()).flatMap(callback, thisArg);
+    const array = Array.from(this.values());
+    return array.flatMap(function (this: This, value, index, arr) {
```
dmaretskyi, line 288: 👍

packages/core/echo/echo-schema/src/query/query.test.ts
```diff
@@ -254,4 +254,26 @@ testWithAutomerge(() => {
...
+
+    const folder = peer.db.add(new TypedObject({ name: 'folder', objects: [] }));
+    const objects = [
+      new TypedObject({ idx: 0, title: 'Task 0' }),
+      new TypedObject({ idx: 1, title: 'Task 1' }),
+      new TypedObject({ idx: 2, title: 'Task 2' }),
+    ];
+    for (const object of objects) {
+      folder.objects.push(object);
+    }
```
dmaretskyi, line 270: inline syntax should work as well
```suggestion
    const folder = peer.db.add(new TypedObject({ name: 'folder', objects: [
      new TypedObject({ idx: 0, title: 'Task 0' }),
      new TypedObject({ idx: 1, title: 'Task 1' }),
      new TypedObject({ idx: 2, title: 'Task 2' }),
    ]
    }));
```

packages/core/echo/echo-schema/src/query/query.ts
```diff
@@ -149,8 +149,12 @@ export class Query<T extends TypedObject = TypedObject> {
 
   private _ensureCachePresent() {
     if (!this._resultCache) {
-      this._resultCache = Array.from(this._sources).flatMap((source) => source.getResults()) as QueryResult<T>[];
-      this._objectCache = this._resultCache.map((result) => result.object!).filter((object): object is T => !!object);
+      this._resultCache = new EchoArray(
```
dmaretskyi, line 152: This doesnt look right. The idea behind EchoArray is to be a data-access object: the changes made to EchoArray are automatically serialized as mutations and persisted on disk. 

Is there a way to instead make `flatMap` work with `EchoArray` natively?
e.g. `[new EchoArray([1, 2]), new EchoArray([3])].flatMap(x => x) == [1, 2, 3]`

dmaretskyi:
I'm suspecting `flatMap` calls `Array.isArray` on the every result to see if it should be flatened. Here it works with proxies but doesnt work with Sets:
image

dmaretskyi:
Hmm, I've tried redirecting the proxy to a different object but it seems like flat map doesnt care
image

dmaretskyi:
nvm, I just had to implement `has`
image

dmaretskyi:
Notice the proxy target doesnt hold any data - it's just a dummy array to pass `Array.isArray` check. The data access is handled by proxy traps

dmaretskyi:
anyway, I'd suggest changing the proxy target in this PR as well, as that would solve this issue in a more general way allowing for things like `[new EchoArray([1, 2, 3])].flatMap(x => x)]` to work

packages/core/echo/echo-schema/src/query/query.test.ts
yarolegovich: It references objects array later in expect

packages/core/echo/echo-schema/src/query/query.ts
yarolegovich: yeah, this didn't feel right, considered creating a flatMap decorator instead of it

but it's nice that Proxying to an empty array instance worked. tests seem to pass, but I'm still a bit scared that this might not be the most robust construct and smth can break along the way. but I guess we'll fix it when it comes to it

dmaretskyi (approved):

mykola-vrmchk (approved):

PR #5383, opened 2024-01-24
feat: clear CircleCI pnpm job

nferch:
* e2e workflow doesn't appear to be used, but updated anyway
* also: excluded deploy from all scheduled pipelines as it isn't needed for the other "Test Report Daily" schedule


wittjosiah:
@nferch e2e workflow is used to manually run `e2e` on any branch by setting `e2e` trigger parameter to `true` in the UI

nferch:
Need to update branch in CircleCI to main once this is merged.

.circleci/config.yml
```diff
@@ -218,13 +233,20 @@ workflows:
             - Github
 
   e2e:
-    when: << pipeline.parameters.e2e >>
+    when:
+      and:
+        - << pipeline.parameters.e2e >>
+        - not:
+          equal: [ clear_pnpm_cache, << pipeline.schedule.name >> ]
```
wittjosiah, line 240: does this need another intent?

wittjosiah (approved):

.circleci/config.yml
nferch: I don't think so, it passes the schema validation and didn't trigger on the daily run.

wittjosiah: interesting, yaml is confusing sometimes :)

PR #5387, opened 2024-01-24
feat(app-framework): introduce intent result

wittjosiah:
### Motivation / Background

This PR allows intents to send other intents by returning them.
This is much simpler than having to find the intent plugin and calling dispatch.

Thanks to @ZaymonFC for the inspiration ✨

### Details

- updates intent types
- updates how data from results is forwarded in chains
- updates all plugins to respect new types

### Additional information

I also found some plugins were handling intents without returning a truthy value.
Currently considering this an anti-pattern as only one plugin should resolve each intent, as such I've tried to update these to return a result now.
Potentially in the future we may want to consider a way to have an intent be "broadcast" such that it allows multiple plugins to resolve it.


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -548,118 +567,152 @@ export const SpacePlugin = ({
                 element.setAttribute('download', `${filename}.zip`);
                 element.setAttribute('target', 'download');
                 element.click();
-                return true;
+                return { data: true };
```
richburdon, line 570: return url if successful?

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -190,14 +190,14 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
...
-              return { object: new ThreadType() };
+              return { data: new ThreadType() };
             }
 
             case ThreadAction.SELECT: {
               state.threads = intent.data?.threads;
               state.active = intent.data?.active;
               state.focus = intent.data?.focus;
-              break;
+              return { data: true };
```
richburdon, line 200: pre previous. seems a bit odd to return true.
if it's just success then the absence of an error would be ok?
and we can standardize a way to return errors? or just `return { err };`

richburdon (approved):

thure (approved):
lo me gusta ✨

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
wittjosiah: data is optional so strictly this could also just be `{}`, it just needs to return something truthy

wittjosiah: currently the type is `error` vs `err` but yes, that's the expected way to return errors (and the chain will terminate and return if an error is returned in the middle)

PR #5388, opened 2024-01-24
feat: Ability to upload diagnostics report and repair client.

richburdon:
https://github.com/dxos/dxos/assets/3523355/b26a9dd8-cc8f-43e7-a617-5f4c475e7431


packages/apps/plugins/plugin-debug/src/components/DebugSettings.tsx
```diff
@@ -2,25 +2,91 @@
...
+          <DownloadSimple className={getSize(5)} />
+        </Button>
+      </SettingsValue>
+      <SettingsValue label={t('settings repair')}>
+        <Button onClick={handleRepair}>
+          <FirstAidKit className={getSize(5)} />
+        </Button>
+      </SettingsValue>
+
+      {toast && (
```
wittjosiah, line 79: perhaps TODO to add `toast` to layout? Having it here seems less ideal since it would presumably unmount if the dialog closes

richburdon: Done, but it still seems a bit odd to me to centralize popovers, dialogs and toasts in a single plugin. Still thinking...

Issue #5402, opened 2024-01-25
[composer] telemetry dashboard


zhenyasav:
see spec here https://github.com/dxos/product/blob/main/2024-Q1/specs/composer-telemetry.md

wittjosiah:
Initial step here is helping to land #4454

wittjosiah:
Making this issue dashboard-specific. Instrumentation done with #5641 via #5414, feel free to re-open #5641 or open a new issue if we find missing data from the spec.

jessmartin:
@zhenyasav what say ye

PR #5414, opened 2024-01-26
feat(plugin-observability): migrate from telemetry plugin & add settings

wittjosiah:
- refactor telemetry plugin to observability plugin
- add settings to disable observability
- also factors out headless observability helpers into @dxos/observability
- adds entrypoint to composer for tagging internal users
- adds internal marker to halo button

https://github.com/dxos/dxos/issues/5402

image


packages/apps/composer-app/vite.config.ts
```diff
@@ -39,6 +39,7 @@ export default defineConfig({
     minify: process.env.DX_ENVIRONMENT === 'development' ? false : undefined,
     rollupOptions: {
       input: {
+        internal: resolve(__dirname, './internal.html'),
```
nferch, line 42: could this be the default when DX_ENVIRONMENT === 'development'? is that what we'd want?

packages/apps/composer-app/src/internal.ts
```diff
@@ -0,0 +1,14 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { storeObservabilityGroup } from '@dxos/observability';
+
+import { appKey } from './constants';
+
+const run = async () => {
+  await storeObservabilityGroup(appKey, 'dxos');
```
nferch, line 10: could this also set and store the mode to full? or should we just do away with the concept of that and check for the group?

wittjosiah: At present, I've not seen any difference between basic/full in the browser, the differences seemed more specific to node right now.

packages/apps/composer-app/vite.config.ts
wittjosiah: right now we're not even sending any data from the dev environment, the api keys are only embedded into the production builds

packages/apps/composer-app/src/internal.ts
nferch: per the latter comment, i've found it useful to have them separate for testing, but it might depend on how much we want to distinguish our "prod" internal usage with the "development". or how much appetite we have for cleaning data post-ingest

nferch: There's just a small amount of changes on the metrics side, I didn't understand the intent of the semantics. But adding the device and identity tags is what I had considered gating behind mode === 'full', and could consider what other stuff we have hardcoded to check for the group?

packages/apps/plugins/plugin-navtree/src/components/HaloButton.tsx
```diff
@@ -6,24 +6,27 @@ import React from 'react';
 
 import { Avatar, type Size, useJdenticonHref } from '@dxos/react-ui';
 
+const INTERNAL = 'https://avatars.githubusercontent.com/u/57182821';
```
jessmartin, line 9: probably should pull from a locally-bundled image file?

packages/apps/plugins/plugin-observability/src/translations.ts
```diff
@@ -0,0 +1,18 @@
...
+import { OBSERVABILITY_PLUGIN } from './meta';
+
+export default [
+  {
+    'en-US': {
+      [OBSERVABILITY_PLUGIN]: {
+        'plugin name': 'Observability',
+        'observability enabled label': 'Observability Enabled',
+        'observability description':
+          'When enabled, basic observability data will be sent to the DXOS team in order to improve the product. This includes performance metrics, error logs, and usage data. No personally identifiable information, other than your public key, is included with this data and no private data ever leaves your devices. TODO: Add link about telemetry privacy.',
```
jessmartin, line 14: :+1:

jessmartin (approved):
Looks good to me! Small nits

packages/apps/composer-app/src/internal.ts
wittjosiah: At this stage we're going to be relying to the truncated identity key for user identification, so I think that falls into `basic` for now. Certainly once we're at this stage https://github.com/orgs/dxos/discussions/5438 then it would be nice to have a way to still include it optionally for debugging purposes and that sounds like `full` mode?

packages/apps/composer-app/vite.config.ts
nferch: oh yeah, I had forgotten about that. Or to configure the Datadog key. Should we reconsider that now that we're using dev for our internal use?

packages/apps/composer-app/src/internal.ts
nferch: ok, that makes sense. will remove the check and start documenting the behavior in the README.

PR #5416, opened 2024-01-26
fix(client): change worker services log bridge prefix

wittjosiah:
This will disambiguate logs coming from same domain worker vs vault worker.


jessmartin (commented):
LGTM!

PR #5435, opened 2024-01-29
fix: Fix string access for automerge/YJS objects.

richburdon:
### Motivation / Background

Improve consistency and testing of accessing string properties during migration.


packages/core/echo/echo-schema/src/object/text-object.ts
```diff
@@ -151,16 +151,23 @@ export const setTextContent = (object: TextObject, text: string) => {
...
+  (object: TextObject | undefined): string | undefined;
+  (object: TextObject | undefined, defaultValue: string): string;
+} = (object: TextObject | undefined, defaultValue?: string) => {
+  if (!object) {
+    return defaultValue;
+  }
+
   if (isAutomergeObject(object)) {
-    return (object as any).content;
+    return (object as any)?.content ?? defaultValue;
```
mykola-vrmchk, line 163:
```suggestion
    return (object as any)?.[(object as any)?.field] ?? defaultValue;
```

PR #5439, opened 2024-01-29
feat: Include @dxos/tracing data into diagnostics

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/common/tracing/src/trace-processor.ts
```diff
@@ -194,6 +194,21 @@ export class TraceProcessor {
     return res;
   }
 
+  getDiagnostics() {
```
richburdon, line 197: define type that can be referenced in diagnostics below?

PR #5443, opened 2024-01-30
fix: observability

nferch:
* create Node helper
* move/align common telemetry tags to Observability
* rejigger observability tag handling


packages/sdk/observability/src/helpers/common.ts
```diff
@@ -0,0 +1,25 @@
...
+//  - Identifier can be synced via HALO to allow for correlation of events bewteen devices.
+//  - Identifier should also be stored outside of HALO such that it is available immediately on startup.
+export const getTelemetryIdentifier = (client: Client) => {
+  if (!client?.initialized) {
+    return undefined;
+  }
+
+  const identity = client.halo.identity.get();
+  if (identity) {
+    return identity.identityKey.toHex();
```
wittjosiah, line 19:
```suggestion
    return identity.identityKey.truncate();
```
to align with other places we're sending this?

packages/sdk/observability/src/helpers/index.ts
```diff
@@ -2,6 +2,8 @@
 // Copyright 2022 DXOS.org
 //
 
-export * from './app-observability';
+export * from './browser-observability';
+export * from './node-observability';
+export * from './common';
 export * from './map-spaces';
 export * from './setup-telemetry-listeners';
```
wittjosiah, line 9: sort? 🙂

packages/sdk/observability/src/helpers/node-observability.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  if (existsSync(configDir)) {
+    if (!statSync(configDir).isDirectory()) {
+      throw new Error(`Config directory ${configDir} exists but is not a directory`);
+    }
+  } else {
+    await mkdir(configDir, { recursive: true });
+  }
+
+  // named 'telemetry.yml' for backwards compatibility
```
wittjosiah, line 45: fwiw probably will hurt less to change this sooner rather than later

wittjosiah (approved):

packages/sdk/observability/src/helpers/common.ts
nferch: whoops, must have missed this in the merge

packages/sdk/observability/src/helpers/node-observability.ts
nferch: will change the localforage key then too?

PR #5446, opened 2024-01-30
feat: initial vitest random-access-storage setup

yarolegovich:
### Motivation / Background

Initial setup for running vitest on random-access-storage in headless browser mode. 

### Details

* Ported random-access-storage tests to vitest.
* `nx run random-access-storage:test` runs only `node` environment tests.
* `pnpm test` runs browser tests in headless browser mode. 4 tests currently fail. Haven't dug much into this, but 2 fail because chrome-specific `.cswrap` files are returned from `directory.list()` and the actual entry count doesn't match expected count. 2 other seem to fail on attempt to delete a read-only object.
* `VITEST_ENV=browser nx run random-access-storage:test` for some reason fails with:
```
Cannot find module 'vite-plugin-node-polyfills/shims/buffer'
```
Downgrading to 0.17.0 as suggested [here](https://github.com/davidmyersdev/vite-plugin-node-polyfills/issues/74) helps but buffer/ alias gets broken somehow.

### DIFF
--- a/packages/common/random-access-storage/package.json
+++ b/packages/common/random-access-storage/package.json
@@ -16,6 +16,9 @@
     "dist",
     "src"
   ],
+  "scripts": {
+    "test": "VITEST_ENV=browser vitest --config vite.config.ts --run"
+  },
   "dependencies": {
     "@dxos/async": "workspace:*",
     "@dxos/context": "workspace:*",
@@ -34,10 +37,17 @@
     "random-access-web": "^2.0.3"
   },
   "devDependencies": {
+    "@dxos/esbuild-plugins": "workspace:*",
     "@types/pify": "^3.0.2",
     "@types/randombytes": "^2.0.0",
+    "@vitest/browser": "^1.2.2",
+    "@vitest/utils": "^1.2.2",
     "randombytes": "^2.1.0",
-    "source-map-support": "^0.5.12"
+    "source-map-support": "^0.5.12",
+    "vite": "^5.0.12",
+    "vite-plugin-node-polyfills": "0.19.0",
+    "vitest": "^1.2.1",
+    "webdriverio": "^8.29.1"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/common/random-access-storage/project.json
+++ b/packages/common/random-access-storage/project.json
@@ -12,7 +12,17 @@
       }
     },
     "lint": {},
-    "test": {}
+    "test": {
+      "executor": "@nx/vite:test",
+      "inputs": [
+        {
+          "env": "VITEST_ENV"
+        }
+      ],
+      "options": {
+        "config": "{projectRoot}/vite.config.ts"
+      }
+    }
   },
   "$schema": "../../../node_modules/nx/schemas/project-schema.json",
   "implicitDependencies": [
--- a/packages/common/random-access-storage/src/browser/storage.test.ts
+++ b/packages/common/random-access-storage/src/browser/storage.test.ts
@@ -2,12 +2,10 @@
 // Copyright 2021 DXOS.org
 //
 
-// @dxos/test platform=browser
-
-import expect from 'expect';
 import 'source-map-support/register';
 
-import { describe, test } from '@dxos/test';
+import { expect } from 'chai';
+import { describe, test } from 'vitest';
 
 import { createStorage } from './storage';
 import { StorageType } from '../common';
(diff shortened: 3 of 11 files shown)


packages/common/random-access-storage/package.json
```diff
@@ -16,6 +16,9 @@
     "dist",
     "src"
   ],
+  "scripts": {
+    "test": "VITEST_ENV=browser vitest --config vite.config.ts --run"
```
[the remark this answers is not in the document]
dmaretskyi: actually, i've noticed you've added it as a job to project.json - thats's good. In that case you can remove this script

packages/common/random-access-storage/project.json
```diff
@@ -12,7 +12,17 @@
       }
     },
     "lint": {},
-    "test": {}
+    "test": {
```
dmaretskyi, line 15: add as a separate vitest target

packages/common/random-access-storage/project.json
```diff
@@ -12,7 +12,17 @@
...
+      "executor": "@nx/vite:test",
+      "inputs": [
```
dmaretskyi, line 17: you need to add a glob for source files here as well - see the definition for `test` in root `nx.json`

wittjosiah: yeah, beware that it will inherit all options from nx.json so may want to explicitly define a lot of them for now until we move everything over to vitest in order to avoid conflicts between options of the executors

packages/common/random-access-storage/src/browser/storage.test.ts
```diff
@@ -2,12 +2,10 @@
 // Copyright 2021 DXOS.org
 //
 
-// @dxos/test platform=browser
-
-import expect from 'expect';
 import 'source-map-support/register';
 
-import { describe, test } from '@dxos/test';
+import { expect } from 'chai';
```
wittjosiah, line 7: I think vitest provides expect as well https://vitest.dev/guide/features.html#chai-and-jest-expect-compatibility probably better to start migrating to that to simplify dependencies

vite.config.ts
```diff
@@ -0,0 +1,80 @@
...
+  const environment = (process.env.VITEST_ENV ?? "").toLowerCase();
+  switch (environment) {
+    case 'chrome':
+      return createBrowserConfig(environment);
+    case 'node':
+    default:
+      if (environment.length > 0 && environment !== 'node') {
+        console.log("Unrecognized VITEST_ENV value, falling back to 'node': " + environment);
+      }
+      return createNodeConfig();
```
dmaretskyi, line 76:
```suggestion
  const environment = (process.env.VITEST_ENV ?? "node").toLowerCase();
  switch (environment) {
    case 'chrome':
      return createBrowserConfig(environment);
    default:
      console.log("Unrecognized VITEST_ENV value, falling back to 'node': " + environment);
    case 'node':
      return createNodeConfig();
```

yarolegovich: I wanted to writer it that way but then there's some ts lint check about pass-through statements and I didn't want to add ts-ignore directive on top of switch cause they are ugly

packages/common/random-access-storage/src/browser/storage.test.ts
yarolegovich: makes sense, I left it cause I though maybe the team preferred chai syntax, changed to expect from vitest

vite.config.ts
mykola-vrmchk: > some ts lint check about pass-through statements
Strange, I used `switch` previously in our project, and never encountered this, maybe someone added a new lint rule

.circleci/config.yml
```diff
@@ -70,7 +74,22 @@ aliases:
...
+      mode:
+        description: "Test execution environment, incompatible tests will be ignored"
+        type: enum
+        enum: [ "node", "chrome" ]
+        default: "node"
+    steps:
+      - run:
+          command: pnpm nx affected --base=$NX_BASE --head=$NX_HEAD --target=vitest
+          environment:
+            VITEST_XML_REPORT: true
```
mykola-vrmchk, line 91: image

Do we need an XML report?
If we do, could we also have inline results in the circle CI console?

yarolegovich: I think as the number of tests grow and console log ouput becomes bigger having an exported xml report is very useful. Added the second reporter in vitest.shared.ts, now it should print to console in addition to exporting the report

mykola-vrmchk (approved):

PR #5447, opened 2024-01-30
feat(react-ui-editor): Start wiring up toolbar state to editor state

marijnh:
Defines a data type `Formatting` that describes the formatting around the editor selection. Provides a function that retrieves this, and uses that to put the editor toolbar in the proper state.

This isn't finished yet. The PR is to ask for some feedback and show progress.


packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.stories.tsx
```diff
@@ -50,11 +61,12 @@ const Story: FC<{ content: string }> = ({ content }) => {
     return null;
   }
 
+  // TODO Wire up the toolbar state to the editor properly
```
marijnh, line 64: Since the toolbar supposedly will become part of the markdown extension, is there a reason to have a separate Toolbar story that needs to wire everything up separately?

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -24,15 +24,18 @@ import React, { type PropsWithChildren } from 'react';
 import { Button, type ButtonProps, DensityProvider, Select } from '@dxos/react-ui';
 import { getSize } from '@dxos/react-ui-theme';
 
+import { type Formatting } from '../../extensions/markdown';
+
 // TODO(burdon): Revert to string to make extensible?
 export type ActionType =
   | 'blockquote'
-  | 'bold'
+  | 'strong'
```
marijnh, line 32: I went with the strong/emphasis terminology from the CommonMark spec instead of bold/italic.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -68,19 +72,28 @@ const ToolbarRoot = ({ children, onAction }: ToolbarProps) => {
...
+  const { onAction, state } = useToolbarContext('ToolbarButton');
   const handleClick = () => {
     const action = onClick();
     if (action) {
       onAction?.(action);
     }
   };
+  const active = getState ? getState(state) : false;
 
+  // TODO: use the correct way to style these as depressed
```
marijnh, line 88: Do we have some CSS class I can add to make the buttons look active?

packages/ui/react-ui-editor/src/extensions/markdown/toolbar.tsx
```diff
@@ -0,0 +1,51 @@
...
+      pos: main.from,
+      end: main.to,
+      create: createToolbarTooltip,
+      above: true,
+    };
+  });
+
+const createToolbarTooltip = (view: EditorView): TooltipView => {
+  const wrapper = document.createElement('div');
+  // TODO: Use a portal
```
marijnh, line 30: I've been told, in other contexts, that one should use [`createPortal`](https://react.dev/reference/react-dom/createPortal) for this, rather than `createRoot`, since multiple React roots are expensive or otherwise problematic. But that requires the portal to somehow connect to some higher-up React tree, and I'm not entirely sure how to set that up.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.stories.tsx
richburdon: The toolbar should be optional -- i.e., the extension should not depend on it, so it felt right to keep the story separate as with comments.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
richburdon: i've asked our ux designer to look into this and provide something. in the interim a noisy hack would be to add `ring` which would be good enough for testing/development.

packages/ui/react-ui-editor/src/extensions/markdown/toolbar.tsx
richburdon: I'll confer with the the team's React expert, but note that we do this elsewhere in the codebase by just supplying `document.body` as the parent div.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -49,13 +52,14 @@ export type Action = {
 
 export type ToolbarProps = PropsWithChildren<{
   onAction?: (action: Action) => void;
+  state: Formatting;
```
richburdon, line 55: small point; pls sort onX handlers at the end of type def.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -89,9 +102,14 @@ const ToolbarButton = ({ Icon, onClick, title }: ToolbarButtonProps) => {
 const ToolbarSeparator = () => <div className='grow' />;
 
 const MarkdownHeading = () => {
-  const { onAction } = useToolbarContext('MarkdownFormatting');
+  const { onAction, state } = useToolbarContext('MarkdownFormatting');
+  const header = state.blockType && /heading(\d)/.exec(state.blockType);
+  const value = header ? header[1] : state.blockType === 'paragraph' || !state.blockType ? '0' : null;
+  if (value === null) {
+    return <></>;
```
richburdon, line 109: return null to render nothing. is that the intention though? shouldn't the select be always visible?

packages/ui/react-ui-editor/src/extensions/markdown/formatting.test.ts
```diff
@@ -0,0 +1,129 @@
...
+describe('getFormatting', () => {
+  const t = (name: string, doc: string, result: Partial<Formatting>) => {
+    test(name, () => {
+      const formatting = getFormatting(createState(doc));
+      const expected = Object.assign({}, emptyFormatting, result);
+      expect(formatting).to.deep.equal(expected);
+    });
+  };
+
+  t('returns nothing special for regular content', 'hello {world}', {});
```
richburdon, line 78: nicely comprehensive test suite!

are you handling cases where the formatting isn't valid? e.g., (extreme) attempting to bold span that includes a link or codeblock.

packages/ui/react-ui-editor/src/extensions/markdown/toolbar.tsx
```diff
@@ -0,0 +1,51 @@
...
+// Copyright 2024 DXOS.org
+//
+
+import { type Extension } from '@codemirror/state';
+import { showTooltip, type TooltipView, type EditorView } from '@codemirror/view';
+import React from 'react';
+import { createRoot } from 'react-dom/client';
+
+import { getFormatting } from './formatting';
+import { Toolbar } from '../../components/Toolbar/Toolbar';
```
richburdon, line 11: we might be out of sync here.

first, I don't think extensions should import React code; instead if needed pass in a renderer to the constructor (as with `link`) since the app will bring in it's own rendering opinions, classes, etc. and how it wants to configure the Toolbar -- e.g., extended buttons, menu, etc.

second, for now i think we'll opt to have a fixed toolbar set-up by the app (as with the storybook) rather then a tooltip.

but for completeness, it would be something like (sketch):

```tsx
const CustomToolbar = () => {
  const handler = useActionHandler(view);
  return (
    <Toolbar.Root onAction={handler} state={getFormatting(view.state)}>
      <Toolbar.Markdown />
      <Toolbar.Separator />
      <Toolbar.Extended />
    </Toolbar.Root>
  );
}

const renderToolbar = (el, view) => {
  createPortal(<CustomToolbar view={view} />, el);
}

const Story = () => {
  return <Editor extensions={[toolbar({ onRender: renderToolbar })]} />
}
```

packages/ui/react-ui-editor/src/components/TextEditor/hooks.stories.tsx
```diff
@@ -74,11 +76,13 @@ const Story = ({ autoFocus, placeholder, doc, readonly }: StoryProps) => {
...
   const handleAction = useActionHandler(view);
 
+  // FIXME This doesn't update the state on view changes. Also not
+  // sure if view is even guaranteed to exist at this point.
   return (
     <div className={mx(fixedInsetFlexLayout, groupSurface)}>
       <div className='flex h-full justify-center'>
         <div className='flex flex-col h-full w-[800px]'>
-          <Toolbar.Root onAction={handleAction}>
+          <Toolbar.Root onAction={handleAction} state={view ? getFormatting(view.state) : emptyFormatting}>
```
richburdon, line 85: `view && getFormatting(view.state)`
undefined should be handled by caller which then decides on the appropriate no-op.
e.g.,

```
getState={(s) => s?.strikethrough ?? false}
```

i think the modern `?` and `??` operators make it clearer how to handle undefined as an expected state.

packages/ui/react-ui-editor/src/extensions/markdown/formatting.ts
```diff
@@ -4,74 +4,142 @@
...
+  // Whether there are links in the selected text.
+  link: boolean;
+  // If all selected blocks have the same (innermost) list style, that
+  // is indicated here.
+  listStyle: null | 'ordered' | 'bullet' | 'task';
+  // Whether all selected text is wrapped in a blockquote.
+  blockquote: boolean;
+};
+
+export const emptyFormatting: Formatting = {
```
richburdon, line 52: (remove: see comment above).

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.stories.tsx
```diff
@@ -50,11 +61,12 @@ const Story: FC<{ content: string }> = ({ content }) => {
...
   return (
     <div className={mx(fixedInsetFlexLayout, groupSurface)}>
       <div className='flex h-full justify-center'>
         <div className='flex flex-col h-full w-[800px]'>
-          <Toolbar.Root onAction={handleAction}>
+          <Toolbar.Root onAction={handleAction} state={editorView ? getFormatting(editorView.state) : emptyFormatting}>
```
richburdon, line 69: i don't see how/when this is updated since we don't want to re-render Toolbar on each document update, so getFormatting will be frozen.

i think i would create a hook to decouple the dependencies (invert the dependency) to be used by the external component that renders the toolbar.

```tsx
const useToolbarState = () => {
  const [state, setState] = useState({});
  const observer = useMemo(() => EditorView.updateListener.of((update) => {
    setState(getFormatting(update.view));
  }, []);
  return [
    state,
    observer
  ];
}

const Story = () => {
  const [state, observer] = useToolbarState();

  return (
    <Toolbar state={state} />
    <TextEditor ref={ref} extensions={[observer]} />
  );
}
```

marijnh: Thanks, that's what I was looking for.

packages/ui/react-ui-editor/src/extensions/markdown/toolbar.tsx
marijnh: I'll drop this extension, and just make the Toolbar story work with the hook you wrote in the other comment.

packages/ui/react-ui-editor/src/extensions/markdown/formatting.test.ts
marijnh: Yes, aiming to have these behave in a somewhat well-defined way on any input.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
marijnh: If you're in a table cell or code block, this control can't really do a lot. I could disable it instead, if you prefer.

packages/ui/react-ui-editor/src/extensions/markdown/toolbar.tsx
marijnh: Moot here anyway if we're not doing this as a tooltip over the editor.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
marijnh: I've used a background style for the moment, which makes this look pretty much right, but doesn't seem to be how you do styling in general.

marijnh:
Integrated the comments in 4a67a1d67

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
richburdon: yes i think better than the visible editor changing shape; for tootips it should be removed as you've done.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.stories.tsx
richburdon: NOTE: you'd want to not create a new state object if the state hasn't changed since i know updateListener is called very frequently.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
richburdon: right, we use tailwind classes, so would prefer:

`classNames={mx('p-2', active && 'ring')}`

richburdon (approved):

marijnh:
This is now mostly finished. I expect more corner cases will turn up, since automated Markdown modification is messy. But there's a reasonably broad test suite checking the behavior of the commands.

Due to things like [flanking](https://spec.commonmark.org/0.31.2/#left-flanking-delimiter-run) rules, it's not even possible to add styles to some text ranges all, so these commands are best-effort and should be understood to require further user intervention to actually finish/check the result in some cases.

Also note that adding a style to a cursor, in the full editor, will put the cursor after the style markers. This isn't something the commands do (see test suite) but is caused by what I assume to be a problem in your automerge integration. The original transaction puts the cursor in the right place, but the one that gets applied to the editor moves it to a different spot.

PR #5477, opened 2024-01-31
fix(plugin-thread): fix flashing when creating a new thread

### LINKED ISSUE #5464: [comments] all threads flicker when creating a new thread

wittjosiah:
### Motivation / Background

Resolves #5464

### Details

- inverts the dependency between thread and markdown plugins, markdown plugin is more core/essential so i
t should not require threads plugin to work properly
- cleans up markdown extension provides, removes the vestigial `onChange`
- stops thread selection from clearing existing thread positions state

### Additional Info

- [ ] TODO: https://github.com/dxos/dxos/issues/5478

### DIFF
--- a/packages/apps/plugins/plugin-files/package.json
+++ b/packages/apps/plugins/plugin-files/package.json
@@ -24,6 +24,7 @@
   ],
   "dependencies": {
     "@dxos/async": "workspace:*",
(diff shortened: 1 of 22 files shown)


thure (approved):
lgtm, thanks for fixing this ✨

PR #5482, opened 2024-02-01
feat: set and synchronize device profile

nferch:
* add type flag to Device profile
* implement update of device profile
* add unit test for propagation of identity profile update

* hackish way to specify a device profile when joining identity through `IdentityServce`

TODO:
* automatically update empty device profile for existing devices
* fix identity-manager test?


packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -224,6 +236,27 @@ export class IdentityManager {
...
+      assertion: {
+        '@type': 'dxos.halo.credentials.DeviceProfile',
+        profile,
+      },
+    });
+
+    const receipt = await this._identity.controlPipeline.writer.write({ credential: { credential } });
+    await this._identity.controlPipeline.state.waitUntilTimeframe(new Timeframe([[receipt.feedKey, receipt.seq]]));
+    this.stateUpdate.emit();
+    return { deviceKey: this._identity.deviceKey, kind: 0, profile };
```
mykola-vrmchk, line 257:
```suggest
return { deviceKey: this._identity.deviceKey, kind: DeviceKind.CURRENT, profile };
```

mykola-vrmchk, line 257:
```suggestion
    return { deviceKey: this._identity.deviceKey, kind: DeviceKind.CURRENT, profile };
```

packages/devtools/cli/src/commands/device/update.ts
```diff
@@ -0,0 +1,57 @@
...
+import { asyncTimeout } from '@dxos/async';
+import { type Client } from '@dxos/client';
+import { invariant } from '@dxos/invariant';
+import { DeviceProfileDocument } from '@dxos/protocols/proto/dxos/halo/credentials';
+
+import { BaseCommand } from '../../base-command';
+import { IdentityWaitTimeoutError } from '../../errors';
+import { IDENTITY_WAIT_TIMEOUT } from '../../timeouts';
+
+export default class Info extends BaseCommand<typeof Info> {
```
mykola-vrmchk, line 17:
```suggestion
export default class Update extends BaseCommand<typeof Info> {
```

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -137,15 +157,7 @@ export class IdentityManager {
...
-      credentials.push(
-        await generator.createDeviceProfile({
-          platform: platform.name,
-          platformVersion: platform.version,
-          architecture: typeof platform.os?.architecture === 'number' ? String(platform.os.architecture) : undefined,
-          os: platform.os?.family,
-          osVersion: platform.os?.version,
-        }),
-      );
+      credentials.push(await generator.createDeviceProfile(deviceProfile));
```
mykola-vrmchk, line 160: We need to add similar credentials in the method `IdentityManager.acceptIdentity`

packages/devtools/cli/src/commands/halo/join.ts
```diff
@@ -50,6 +72,8 @@ export default class Join extends BaseCommand<typeof Join> {
           onConnecting: async () => ux.action.stop(),
           onReadyForAuth: async () => secret ?? ux.prompt(chalk`\n{red Secret}`),
           onSuccess: async () => {
+            ux.log('setting device profile...');
+            await updateDeviceProfile(client, label);
```
mykola-vrmchk, line 76: Maybe make `client.halo.join` accept deviceProfile same as `client.halo.create`?

nferch: yeah, I had considered that but was trying to avoid having to make such a big API change. but you are probably right, it'll ensure consistency in the browser and make it easier on the caller.

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -146,6 +147,7 @@ service IdentityService {
   rpc QueryIdentity(google.protobuf.Empty) returns (stream QueryIdentityResponse);
   rpc UpdateProfile(halo.credentials.ProfileDocument) returns (Identity);
   rpc SignPresentation(SignPresentationRequest) returns (dxos.halo.credentials.Presentation);
+  rpc SetCurrentDeviceProfile(halo.credentials.DeviceProfileDocument) returns (google.protobuf.Empty);
```
dmaretskyi, line 150: UpdateDeviceProfile - to be consistent with other names

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -204,6 +243,13 @@ export class IdentityManager {
       displayName: this._identity.profileDocument?.displayName,
     });
 
+    if (!this.currentDeviceProfile) {
```
dmaretskyi, line 246: can we pass this as a parameter to `acceptIdentity` (and transitively other flows that trigger this)? This should avoid this extra state in the service

nferch: yeah, I had failed implementing it in that manner by modifying the `DeviceInvitationProtocol` but now see the problem.

PR #5491, opened 2024-02-01
fix(plugin-thread): include title for comment threads

### LINKED ISSUE #5490: [comments] include highlighted text as thread title
Use `Comment` `cursor` to grab text from document

wittjosiah:
Resolves #5490

image


jessmartin:
image


I like how @thure's design added some touches to indicate it's a quote:
- surrounding it with quotes
- changing the saturation from the main comment text (I think you've already done that, but there are no comments in the screenshot to compare)
- the little down arrow
- bolding the comment (I'm not sure about that one...)

PR #5499, opened 2024-02-01
feat: Disable table if not blank line.

richburdon:


marijnh:
I don't think `blockType` being non-null is a great proxy for being on a blank line. We may want to add a separate field to that to the formatting data structure.

The changes to the list command won't renumber lists after the cursor, or follow the indentation and numbering of any list above the line, but otherwise looks reasonable.

marijnh:
I implemented all that logic to wrap/unwrap arbitrary blocks as codeblocks, but I guess now the whole control is disabled when the cursor is not on a blank line, for some reason?

PR #5501, opened 2024-02-01
fix(plugin-thread): prevent empty messages from being sent

### LINKED ISSUE #5493: [comments] long delay after pressing enter on comment causes multiple comments to be entered
- in composer create a comment.
- press enter twice.
- you'll see two comments.

wittjosiah:
Resolves #5493


thure (approved):
Thanks for this 👍

PR #5520, opened 2024-02-03
fix: Automerge extension to fix history

richburdon:
- Resolve history bug.
- Resolve link/parens cursor positioning bug.


packages/ui/react-ui-editor/src/extensions/automerge/sync.ts
```diff
@@ -12,67 +12,59 @@ import { next as A } from '@dxos/automerge/automerge';
...
       view.dispatch({
-        changes: inverted,
-        annotations: reconcileAnnotation.of(true),
+        effects: updateHeads(newHeads),
+        annotations: reconcileAnnotation.of(false),
       });
     }
+  }
 
-    // Apply the unreconciled transactions to the document.
```
wittjosiah, line 63: Having the equivalent of some of these comments on the new code would be helpful for parsing what each of these methods is doing

wittjosiah (commented):
Which of the many issues that we think are related to this, does this resolve?

dmaretskyi (approved):

PR #5525, opened 2024-02-04
chore(react-ui-theme): Use fork of `colors`

### LINKED ISSUE #5487: [react-ui] fork @fluent-blocks/colors and remove old engine requirements
Currently the following warning is omitted when installing, including to downstream consumers of the app templates

Image

thure:
This PR changes the procedural color dependency of `react-ui-theme` to use a package we control. Resolves #5487.


wittjosiah (approved):
@thure One note is that `pnpm` is still listed in the `engines`, I have a feeling that this will still omit a warning when installed w/ `npm` even though the version range is now more permissive

PR #5541, opened 2024-02-05
feat(app-framework): support undo for intents

### LINKED ISSUE #5041: [composer] undo plugin
- [ ] Undo plugin
- [ ] Toast appears for 5s to enable undo

wittjosiah:
### Motivation / Background

This PR adds history to the intent plugin in order to be able to support undo.

Resolves #5041

### Details

- update intent types to handle the possibility of inverse intents
- add intent for when undo is available
- layout plugin resolves available undos to a toast message
- test undo by making thread deletion undoable

### Additional information

- At present undo only supports one action but history is total, so in the future you should be able to browse the history as well as undo several actions.
- Initially implemented via a toast, but expectation is that this will have custom UI designed for it in the future as well.
- [ ] remove old undo plugin

Screenshot 2024-02-05 at 14 09 11


packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -221,6 +223,25 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
...
+              const { document: doc, thread, cursor } = intent.data ?? {};
+              if (!isDocument(doc) || !isThread(thread)) {
+                return;
+              }
+
+              if (intent.inverse && typeof cursor === 'string') {
+                doc.comments.push({ thread, cursor });
+                return { data: true };
+              } else {
+                const index = doc.comments.findIndex((comment) => comment.thread?.id === thread.id);
```
richburdon, line 237: @dmaretskyi should we actively delete the object and not just splice it (and restore above)?

packages/sdk/app-framework/src/plugins/IntentPlugin/intent.ts
```diff
@@ -30,6 +30,11 @@ export type Intent = {
    */
   action: string;
 
+  /**
+   * Whether or not the intent is being undone.
+   */
+  inverse?: boolean;
```
richburdon, line 36: i think i prefer `undo`

packages/sdk/app-framework/src/plugins/IntentPlugin/provides.ts
```diff
@@ -21,3 +21,15 @@ export const parseIntentPlugin = (plugin: Plugin) =>
...
+//
+// Intents
+//
+
+const INTENT_ACTION = 'dxos.org/plugin/intent';
+export enum IntentAction {
+  /**
+   * Fired after an intent is dispatched if the intent is undoable.
+   */
+  UNDO_AVAILABLE = `${INTENT_ACTION}/undo-available`,
```
richburdon, line 34: I think `SHOW_UNDO` is more consistent. unless there's an even broader intent such as HISTORY_UPDATED where the layout would get the current state of the history (incl. undo being available).

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -221,6 +223,25 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
...
               return { data: true };
             }
+
+            case ThreadAction.DELETE: {
+              const { document: doc, thread, cursor } = intent.data ?? {};
+              if (!isDocument(doc) || !isThread(thread)) {
+                return;
+              }
+
+              if (intent.inverse && typeof cursor === 'string') {
```
richburdon, line 233: it feels a bit error prone to `else` here if EITHER of these are false. feels like instead cursor === string should be an invariant inside the inverse (undo) case?

richburdon (commented):
nicely done

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
wittjosiah: @richburdon note that comments are not standalone objects at the moment but apart of the the `Document` schema. Only `Thread`s are separate objects.

packages/apps/plugins/plugin-layout/src/translations.ts
```diff
@@ -20,6 +20,10 @@ export default [
         'toggle fullscreen label': 'Toggle fullscreen',
         'settings enable complementary sidebar label': 'Show context sidebar.',
         'settings show footer label': 'Show footer.',
+        'undo available label': 'Click to undo previous action.',
```
thure, line 23: I don’t suppose we might be able to provide the user with more information than this? I think ideally it’d say something like:

> You deleted a thread.
> Undo | Dismiss

thure: Can sort it out in a follow-up if we get feedback indicating folks would want this, I know it’s potentially a lot of added complexity.

wittjosiah: Good point, let me see if I can find a way that we could have a custom label here specified by the action.

thure (approved):
Meant to approve, especially so if we can tell the user what triggered the toast in the toast, but that can come later.

PR #5542, opened 2024-02-05
fix(plugin-markdown): fix infinite loop when selecting a comment thread

### LINKED ISSUE #5537: [authoring] clicking the heading for a comment on a doc crashes browser
### Repro

- create a document
- select some text and create a comment on it
- in the right sidebar, click on the heading, the part that previews the text that was highlighted on
- browser hangs forever ♾️

wittjosiah:
### Motivation / Background

Resolves #5537

### Details

- return truthy intent result from markdown intent resolver so that intent plugin knows intent was
handled
- remove old unused intent resolver


packages/sdk/observability/src/helpers/setup-telemetry-listeners.ts
```diff
@@ -13,8 +14,7 @@ let totalTime = 0;
 export const setupTelemetryListeners = (namespace: string, client: Client, observability: Observability) => {
   const clickCallback = (event: any) => {
     if (observability.group === 'dxos' && event.target && !event.target.id) {
-      // TODO(wittjosiah): Use @dxos/log so these can be filtered.
-      console.warn('Click event on element without id:', event.target);
+      log.info('Click event on element without id:', event.target);
```
richburdon, line 17:
```
log.warn('click on element without id', { target: event.target });
```

PR #5543, opened 2024-02-05
fix: Polish spacing & other styles, refactor toolbar implementation

### LINKED ISSUE #5239: [authoring] editor padding is excessive on mobile devices
Image

thure:
This PR resolves #5239.

This PR also refactors Editor’s Toolbar to use the Toolbar from `react-ui`.

https://github.com/dxos/dxos/assets/855039/1c4d6da8-8639-4bed-a30c-87b7ab05ede8

#### Caveats

- Select issues, which should probably be fixed in a follow-up since they may be upstream:
  - keyboard access appears to be broken by `Select.Trigger`
  - `Select.Content` is not getting positioned correctly (not sure if regression)


wittjosiah (approved):
Unsure of how I feel about the border around the text editor, but the rest LGTM

PR #5546, opened 2024-02-05
feat(composer): Add native URL redirection

mjamesderocher:
This PR adds redirection to the native app from the browser if the native app is installed.

To test, the native Mac app must be built and opened (The next version of Socket Supply will work with this even if the app is not opened).

Then open a URL in the browser and confirm the dialog that asks if you want to allow the native app to open the link.


packages/apps/composer-app/src/main.tsx
```diff
@@ -74,6 +74,42 @@ const main = async () => {
   );
   const isSocket = !!(globalThis as any).__args;
```
jwerle, line 75: soon you'll be able to use `globalThis.isSocketRuntime`

zhenyasav (approved):

PR #5547, opened 2024-02-05
feat: Factor out vim mode.

richburdon:
- [x] Simplify TextEditor by factoring out vim extension.
- [x] Fix Esc/Enter handling for TextEditor tab/focus management.
- [x] Restores tabstar handling.


packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -184,29 +163,30 @@ export const BaseTextEditor = forwardRef<EditorView, TextEditorProps>(
...
           case 'Escape': {
-            if (editorMode === 'vim' && (altKey || shiftKey || metaKey || ctrlKey)) {
+            if (altKey || shiftKey || metaKey || ctrlKey) {
+              rootRef.current?.focus();
+              break;
+            } else {
               view?.focus();
             }
             break;
           }
```
wittjosiah, line 186: I only added this for the vim extension so it should probably be factored out as well. Maybe pairing the vim extension w/ a keymap extension that has this functionality?

richburdon: the CM keymap maps onto keypress not up/down so it can't handle ESCAPE, nor does it have access to the outer div. So my approach here was to treat the outer div as a container that takes the tabster attributes and handles external navigation.

richburdon: i think i've properly factored this out now.

PR #5550, opened 2024-02-06
fix: restore visibility of chat thread

richburdon:
- Restores visibility of chat thread.

From @thure:

try grouping message blocks with closely-grouped timestamps into the same Message before adjusting how the username is styled, I anticipate that will resolve the issue, and I expect diverging from how Discord/Slack/Teams style usernames will have an alienating effect
check for unmounts using useEffect(()=>(()=>{console.log('[unmount]')})), I suspect since arrays are involved, something might be referentially changing that shouldn’t
Adding sticky block-end-0 to the MessageTextbox of ChatContainer is one way of doing this – use fixed instead of sticky if you prefer the textbox always stay at the bottom
To keep messages at the bottom as they appear, use the AnchoredOverflow component
Note that if there are reactivity issues, AnchoredOverflow won’t work as expected — messages need to be added to the end without other messages or the thread getting unmounted
This should all be achievable without intervening on overflow, programmatically setting scrollTop, or using reverse flex flow direction, please let me know if you have any issues and I can investigate.


packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -138,13 +138,13 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
               const location = navigationPlugin?.provides.location;
               const activeNode = location?.active ? graph?.findNode(location.active) : undefined;
               const active = activeNode?.data;
-              const space = isDocument(active) && getSpaceForObject(active);
+              const space = active && getSpaceForObject(active);
```
wittjosiah, line 141:
```suggestion
              const space = isTypedObject(active) && getSpaceForObject(active);
```
what happens if you pass a non-echo object to `getSpaceForObject`?

richburdon: i think it just returns undefined. adding a test.

PR #5560, opened 2024-02-06
test: minor random access storage changes + very basic extra tests

yarolegovich:
### Details

* getDiskInfo test and fixes
* operation on a destroyed file test
* double flush fix
* changed vitest reporter to verbose to have more output in console


packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -128,22 +128,28 @@ export class WebFS implements Storage {
...
+      );
       files.delete(filename);
-      await this._root!.removeEntry(filename, { recursive: true }).catch((err: any) => log.warn(err));
     }
     this._root = undefined;
   }
 
   async close() {
-    await Promise.all(Array.from(files.values()).map((file) => file.close()));
+    await Promise.all(
```
yarolegovich, line 140: added catch to be consistent with reset

packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -158,7 +164,7 @@ export class WebFS implements Storage {
           (async () => {
             switch (entry.kind) {
               case 'file':
-                used += await (entry as FileSystemFileHandle).getFile().then((f) => (used += f.size));
+                used += await (entry as FileSystemFileHandle).getFile().then((f) => f.size);
```
yarolegovich, line 167: double-counting

packages/common/random-access-storage/src/browser/web-fs.ts
```diff
@@ -323,51 +329,50 @@ export class WebFile extends EventEmitter implements File {
...
 
   private async _flushNow() {
-    this._flushPromise = (this._flushPromise ?? Promise.resolve())
-      .then(() => this._flushCache())
-      .catch((err) => log.warn(err));
+    this._flushPromise = this._flushPromise ?? Promise.resolve();
+
+    try {
+      await this._flushCache();
+      this._flushPromise = null;
```
yarolegovich, line 336: the logic in scheduled flush was broken:
1. _flushPromise was initialized.
2. callback captures and awaits it.
3. then compares whether promise changed to avoid double-flush, but it was the same

packages/common/random-access-storage/src/node/node-storage.ts
```diff
@@ -69,11 +69,11 @@ export class NodeStorage extends AbstractStorage implements Storage {
     const recurse = async (path: string) => {
       const pathStats = await stat(path);
 
-      used += pathStats.size;
```
yarolegovich, line 72: exclude directory entry sizes

packages/common/random-access-storage/src/browser/web-fs.ts
dmaretskyi: This doesnt look right: `_flushNow` must also guard against concurrent invocations of `_flushCache`. I think here two concurrent calls to `_flushNow` will also result in `_flushCache` running concurrently too

yarolegovich: oh true, the flow is super weird after the change, but as for the comment:

>  I think here two concurrent calls to _flushNow will also result in _flushCache running concurrently too

I think that's the desired behavior in the realm of weird api usages.
imagine I write and flush while the ongoing flush went past its write call. In this case there'll be unflushed data in our buffer.
it's probably best to introduce a `dirty` flag here and decide whether to start a new flush based on it.

dmaretskyi (approved):

PR #5592, opened 2024-02-07
fix: Table rendering and resizing bug

### LINKED ISSUE #5585: [tables] lazy loading error
- seems to be specific to tables
- only seems to happen with vite dev server

Image

richburdon:
Resolves #5585


wittjosiah:
I'm not too familiar w/ table, @richburdon can you explain a bit more what the solution here was?

richburdon:
> I'm not too familiar w/ table, @richburdon can you explain a bit more what the solution here was?

The react state bug I think is some kind of race condition.
Separating the outer component that instantiates the table's context provider (principally with the output of the main useTable hook) from a new inner TableImpl component that consumes the context helped -- and made debugging much simpler. I think also there may be a bug in the tanstack code where setting a row-grouping callback handler when no row-group is specified causes the bug.

The other bug was due to the column resizing information not being handled properly.

Issue #5168, opened 2024-01-12
[organization] remove the upload image button at the bottom of stacks, move to regular action


zhenyasav:

wittjosiah:
This is higher priority because it leads to broken functionality is the default application setup.

richburdon:
how so? this should only be visible when the appropriate plugins are enabled.

wittjosiah:
@richburdon the problem is that currently that's not true, it's visible when stacks plugin is enabled even if ipfs plugin is disabled. This functionality should only be available when IPFS plugin is also enabled.

richburdon:
ok then i'll fix that.

PR #5598, opened 2024-02-07
feat: Show stack file upload if IPFS is configured

richburdon:
Resolves [#5168](https://github.com/dxos/dxos/issues/5168)

Screenshot 2024-02-07 at 4 59 39 PM

Screenshot 2024-02-07 at 5 00 13 PM


packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
```diff
@@ -44,6 +44,7 @@ export const DebugPlugin = (): PluginDefinition<DebugPluginProvides> => {
         .prop(settings.values.$debug!, 'debug', LocalStorageStore.bool)
         .prop(settings.values.$devtools!, 'devtools', LocalStorageStore.bool);
 
+      // TODO(burdon): Remove hacky dependency on global variable?
```
wittjosiah, line 47: Do you want to expose this in the UI? Otherwise, not sure how else to get playwright access.

packages/sdk/app-framework/src/plugins/common/file.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { type Plugin } from '../PluginHost';
+
+export const defaultFileTypes = {
+  images: ['png', 'jpg', 'jpeg'],
+  media: ['mp3', 'mp4', 'mov', 'avi'],
+  text: ['pdf', 'txt', 'md'],
+};
+
+export type FileInfo = {
```
wittjosiah, line 13: this seems rather ipfs specific rather than a generic file info type? not that that is necessarily a bad thing, but makes me question it being in `common`. Perhaps warrants a TODO alluding to how this will attempt to provide a "common" files interface

packages/sdk/app-framework/src/plugins/common/metadata.ts
```diff
@@ -2,7 +2,7 @@
 // Copyright 2023 DXOS.org
 //
 
-import type { Plugin } from '../PluginHost';
+import { type Plugin } from '../PluginHost';
```
wittjosiah, line 5: interesting, you prefer the `type` keyword inside the braces?

wittjosiah (approved):

packages/sdk/app-framework/src/plugins/common/metadata.ts
richburdon: i just find it more consistent over time.

packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
richburdon: I didn't realize this was for playwright.
What do you think about passing some option in the the client API, where the client calls you back with an "root access" API or additional control plane. This is fine if it's a one-off, but it tends not to be -- e.g., Dima has at least 2 different ways of poking the client from the console.

wittjosiah: Yeah, that sounds reasonable. FWIW there's other things within the app that use this pattern as well of exposing things on window for playwright. Not that I think that makes it better, but just that it's not just client that needs a solution. I'd been thinking that I'd like a way to expose them only when running for playwright.

PR #5602, opened 2024-02-08
feat: Interactive page for Spaces

### LINKED ISSUE #5232: [organization] basic page for spaces
- stop redirecting to first object
- at minimum should show all the available actions for the space
- perhaps list the members as well?
- maybe list objects in the space from graph?
- later #4365

thure:
Resolves #5232.

Remaining to do:
- [x] Check diff, remove any unneeded changes
- [x] Update Figma

Screenshot 2024-02-12 at 16 35 24
Screenshot 2024-02-12 at 16 31 43
Screenshot 2024-02-12 at 16 31 19

https://github.com/dxos/dxos/assets/855039/4dba5c2b-a9e2-42c1-89ca-085cef35ebdb


packages/apps/plugins/plugin-space/src/components/SpaceMain/SpaceMembersSection.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+export const SpaceMembersSection = ({ space }: { space: Space }) => {
+  const { t } = useTranslation(SPACE_PLUGIN);
+  const invitations = useSpaceInvitations(space.key);
+  const { dispatch } = useIntent();
+
+  const handleCloseDialog = () =>
+    dispatch({ action: LayoutAction.SET_LAYOUT, data: { element: 'dialog', state: false } });
+
+  const handleInvitationSelect = ({
```
wittjosiah, line 48: I think we should move forward with this for now, but something to consider is, while there may be different types of member lists with various information, perhaps shell owns invitations? Instead of managing invitations on this page there could just be an invite/share button which opens the shell.

wittjosiah (approved):

packages/apps/plugins/plugin-space/src/components/SpaceMain/SpaceMembersSection.tsx
thure: There’s some real utility in not depending on dialogs for these tasks — I’m hoping this experiment will show that if we want tidier code, we may need shell to be able to own these operations in a way that doesn’t also restrict its UI to a dialog.

Issue #5492, opened 2024-02-01
[authoring] can't highlight parts of code w/ mouse inside code block


wittjosiah:
- cursor is pointer
- double clicking does not highlight word
- can still key into words and edit them

https://github.com/dxos/dxos/assets/4529818/1b1ed09d-9c9e-4a42-9e63-1ea91f530dcc

Related #5604

PR #5603, opened 2024-02-08
fix: Fix markdown codeblock formatting and selection

richburdon:
Resolves: #5492

- [x] Fix codeblock formatting (previously, could not click on words in code since position relative).
- [x] Fix focus margins for document (main and stack section).
- [x] Min height for document sections (at least as tall as the min height of 4-rem defined by the section).


packages/apps/plugins/plugin-markdown/src/components/DocumentSection.tsx
```diff
@@ -33,7 +32,7 @@ const DocumentSection: FC<{
       placeholder={t('editor placeholder')}
       slots={{
         root: {
-          className: mx('flex flex-col grow m-0.5', attentionSurface, focusRing),
+          className: mx('flex flex-col grow m-0.5 min-bs-[8rem]', attentionSurface, focusRing),
```
thure, line 35: 8rem = 32 tailwind units, so `min-bs-32`

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -82,6 +82,7 @@ const EditorMain = ({ comments, toolbar, extensions: _extensions, ...props }: Ed
           root: {
             className: mx(
               focusRing,
+              'm-0.5 mx-1', // TODO(burdon): Standardize for focusRing. Just focus "paper" of document?
```
thure, line 85: Right, I think the solution for this problem belongs elsewhere. Glad to take it on if you like, would like to incorporate the feedback here as well: https://discord.com/channels/837138313172353095/837688768872972358/1204725830148685865

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -251,7 +251,7 @@ const MarkdownBlocks = () => {
   const { t } = useTranslation(translationKey);
   const value = markdownBlocks.find(({ getState }) => state && getState(state));
   return (
-    <NaturalToolbar.ToggleGroup type='single' value={value?.type}>
+    <NaturalToolbar.ToggleGroup type='multiple' value={value?.type ? [value.type] : []}>
```
thure, line 254: This looks redundant — `single` should be sufficient here and `type=multiple` affects how this behaves down to the HTML.

If value might be falsy, I believe you can use an empty string: `value?.type ?? ''`.

packages/ui/react-ui-editor/src/extensions/markdown/code.ts
```diff
@@ -23,42 +23,53 @@ const styles = EditorView.baseTheme({
...
+    '&::after': {
+      background: getToken('extend.semanticColors.input.light'),
+    },
+  },
+  '&dark .cm-codeblock-first, &dark .cm-codeblock-last': {
+    mixBlendMode: 'lighten',
+    '&::after': {
+      background: getToken('extend.semanticColors.input.dark'),
+    },
+  },
```
thure, line 72: Thank you for using the new semantic colors ✨

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
thure: Come to think of it, if we follow with our principle that the focus indicator is optional when the element will have a blinking cursor inside it, then we should actually look into preventing direct focus of the text area and instead always pass focus within.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
richburdon: i would have assumed single could take undefined; to me `''` IS a value.

thure: Since this is a controllable state prop, I think Radix does distinguish between `undefined` and `''` as signaling whether you intend to control `value` or not. Using `''` this group behaves as expected:

https://github.com/dxos/dxos/assets/855039/3e639444-8376-4d1c-a5f8-199f159d22c8

thure: `''` is okay as a value because it doesn’t match the id of any actual option.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
thure: I wasn’t able to find a comfortable solution for skipping direct focus of the text area, but I think this may be a reasonable compromise for now since EditorMain no longer really has a top- or bottom-edge (that it is also responsible for rendering):

https://github.com/dxos/dxos/assets/855039/1d885fe3-af70-4d24-af6e-d154b995e97d

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
richburdon: yes that's a good point re undefined and controlled values. i think `null` would be a more appropriate value though.

richburdon: an issue in your video is that the tasks button should be disabled while you're in a bullet list so the state doesn't seem to be respected (or is incorrect)?

thure: Yes I wasn’t quite sure what was going on there… it seems like CM is interpreting the list type as being the latest one applied, even if it might seem as if multiple are applied.

PR #5624, opened 2024-02-09
fix(plugin-thread): Chat layout

### LINKED ISSUE #5608: [composer] Chat thread should be pinned from bottom of sidebar
The stand-alone chat thread should be pinned to the bottom of the sidebar.

Image

thure:
Resolves #5608.

This PR devolves `ThreadContainer` into `ChatContainer` and `CommentContainer` (likewise renaming `ThreadsContainer` to `CommentsContainer`) since the design diverges.

The new `ChatContainer` gets the expected layout: messages scroll backwards from a compose box pinned to the bottom.

Getting scroll to reverse is still a nuanced affair, so please let me know if you encounter issues.

https://github.com/dxos/dxos/assets/855039/38c6ceb9-ad1a-4370-9653-9aed3c7a7e42

https://github.com/dxos/dxos/assets/855039/26688a30-3cfe-443e-947e-8e73765aa533

https://github.com/dxos/dxos/assets/855039/67481b46-d069-482e-b7b7-963e3b823a19

### DIFF
--- a/packages/apps/plugins/plugin-thread/src/components/ChatContainer.tsx
+++ b/packages/apps/plugins/plugin-thread/src/components/ChatContainer.tsx
@@ -9,6 +9,7 @@ import { TextObject, getTextContent, useMembers } from '@dxos/react-client/echo'
 import { useIdentity } from '@dxos/react-client/halo';
 import { ScrollArea, useTranslation } from '@dxos/react-ui';
 import { useTextModel } from '@dxos/react-ui-editor';
+import { mx } from '@dxos/react-ui-theme';
 import { MessageTextbox, type MessageTextboxProps, Thread, ThreadFooter, threadLayout } from '@dxos/react-ui-thread';
 
 import { MessageContainer } from './MessageContainer';
@@ -93,8 +94,8 @@ export const ChatContainer = ({ space, thread, currentRelatedId, current, autoFo
       classNames='bs-full grid-rows-[1fr_min-content_min-content] overflow-hidden'
     >
       <ScrollArea.Root classNames='col-span-2'>
-        <ScrollArea.Viewport classNames='overflow-anchored after:overflow-anchor after:block after:bs-px after:-mbs-px'>
-          <div role='none' className={threadLayout}>
+        <ScrollArea.Viewport classNames='overflow-anchored after:overflow-anchor after:block after:bs-px after:-mbs-px [&>div]:min-bs-full [&>div]:!grid [&>div]:grid-rows-[1fr_0]'>
+          <div role='none' className={mx(threadLayout, 'place-self-end')}>
             {thread.messages.map((message) => (
               <MessageContainer key={message.id} message={message} members={members} onDelete={handleDelete} />
             ))}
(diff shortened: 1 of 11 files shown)


packages/apps/plugins/plugin-thread/src/components/ChatContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+        blocks: [block],
+      }),
+    );
+
+    setNextMessage(() => {
+      autoFocusAfterSend.current = true;
+      return { text: new TextObject() };
+    });
+
+    setTimeout(() => threadScrollRef.current?.scrollIntoView({ behavior: 'smooth', block: 'end' }), 10);
```
richburdon, line 73: factor out function (and call same above)?

packages/apps/plugins/plugin-thread/src/components/types.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import type { Thread as ThreadType } from '@braneframe/types';
+import { type Space } from '@dxos/react-client/echo';
+import type { ThreadProps } from '@dxos/react-ui-thread';
+
+export type ThreadContainerProps = {
+  space: Space;
+  thread: ThreadType;
+  currentRelatedId?: string;
+  autoFocusTextBox?: boolean;
```
richburdon, line 13: just autoFocus

packages/apps/plugins/plugin-thread/src/components/ChatContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { useTextModel } from '@dxos/react-ui-editor';
+import { MessageTextbox, type MessageTextboxProps, Thread, ThreadFooter, threadLayout } from '@dxos/react-ui-thread';
+
+import { MessageContainer } from './MessageContainer';
+import { command } from './command-extension';
+import { type ThreadContainerProps } from './types';
+import { useStatus, useMessageMetadata } from '../hooks';
+import { THREAD_PLUGIN } from '../meta';
+
+/**
```
richburdon, line 20: nit: better to just document the TS def as a pattern.

packages/apps/plugins/plugin-thread/src/components/ChatContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const block = {
+      timestamp: new Date().toISOString(),
+      content,
+    };
+
+    thread.messages.push(
+      new MessageType({
+        from: { identityKey: identity.identityKey.toHex() },
+        context: { object: currentRelatedId },
```
richburdon, line 63: pass in entire context?

thure: This was inherited from the previous implementation — I’m not sure what a MessageType’s `context` is for, maybe you could advise?

packages/apps/plugins/plugin-thread/src/components/types.ts
thure: I disagree, where a property is a React- or HTML-native attribute, that is understood to apply to the outer element, which in this case would be `Thread`, not `MessageTextbox` which is a child. Any message could be autofocused, and it may eventually somehow be possible to autofocus the thread itself, so I don’t think the ambiguity is worthwhile.

We already qualify props when the name alone is not sufficient, e.g. `onCreateInvitationClick` instead of just `onClick` (since multiple children could be clicked).

PR #5625, opened 2024-02-09
chore: Load composer plugins from source in dev mode

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/apps/composer-app/tsconfig.json
```diff
@@ -1,202 +1,191 @@
...
     {
-      "path": "../plugins/plugin-thread"
+      "path": "../plugins/plugin-thread",
     },
     {
-      "path": "../plugins/plugin-wildcard"
+      "path": "../plugins/plugin-wildcard",
     },
     {
-      "path": "../types"
```
wittjosiah, line 199: trailing commas valid json?

toolbox.json
```diff
@@ -21,6 +21,11 @@
     "fixedKeys": [
       "extends",
       "compilerOptions"
-    ]
+    ],
+    "pathMapping": {
+      "include": [
+        "packages/apps/plugins/*"
```
wittjosiah, line 27: if I understand correctly, this means only the plugins will hot reload? lower level libraries will still require recompilation unless they're added to this list?

dmaretskyi: yea, we can add them progressivelly as long as they are compiled "normally" (without any special plugins)

wittjosiah: makes sense 👍 just wanted to test my understanding. In the long run lower level libs (e.g. protocols, dxos-ui, etc.) will be coming from another repo so this seems like the right level to set this at. Lower levels we should be test with unit tests/storybooks vs hotreloading composer

PR #5647, opened 2024-02-12
feat(plugin-space): navigate to peer on click

### LINKED ISSUE #5231: [composer] stop using letters, always use jdenticons
jdenticons are unique but not clearly identifiable enough to be useful, especially at the sizes that we usually render them at

### LINKED ISSUE #5244: [collab] go to peer
this jumps to wherever the peer is currently present, but does not follow them. Later we can do #5245 

Depends on #5174

wittjosiah:
### Motivation / Background

This PR allows someone to jump to where a peer is in the space by clicking on their avatar in the space presence.

Resolves #5244
Resolves #5231 

### Details

- hooks up onMemberClick
- updates tooltip to list out other members

### Additional information

- expecting this to be rewritten based on the redesign, but the data is all there and working now so seemed worthwhile to hookup

Screenshot 2024-02-12 at 14 30 19


packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -90,59 +108,62 @@ export type MemberPresenceProps = ThemedClassName<{
...
             </AvatarGroupItem.Root>
           </Tooltip.Trigger>
           <Tooltip.Portal>
             <Tooltip.Content side='bottom'>
-              <span>
-                {t('viewers label', { count: members.filter((member) => member.match).length })}
-                {t('members online label', { count: members.length - members.filter((member) => member.match).length })}
-              </span>
               <Tooltip.Arrow />
+              <List classNames='max-h-56 overflow-scroll flex flex-col gap-2'>
```
thure, line 155: `overflow-y-auto` is preferred, the `scroll` value forces a scrollbar to appear.

thure: `flex flex-col` is almost always unnecessary — I believe `List` has that by default anyway, but even that shouldn’t apply that since `block` (the default) is better to work with wherever possible.

thure (approved):
my suggestions are non-blocking, otherwise this looks good ✨

PR #5653, opened 2024-02-12
fix: remove custom font from theme for logo demo

richburdon:


thure (approved):
This is great, thank you ✨

PR #5654, opened 2024-02-13
fix(tools): force ligth mode in composer devtools

yarolegovich:
### Motivation / Background

Composer devtools aren't usable with dark mode system preference and implementing proper dark mode support doesn't seem like an easy task, so a quickfix suggestion here.

Here's how dark mode looks:

image

### DIFF
--- a/packages/devtools/devtools/src/app/Devtools.tsx
+++ b/packages/devtools/devtools/src/app/Devtools.tsx
@@ -70,6 +70,17 @@ export const Devtools: FC<{
     },
   });
 
+  // TODO: remove hardcoded themeMode when Devtools dark mode support is ready
+  useEffect(() => {
+    const isDarkMode = document.documentElement.classList.contains('dark');
+    if (isDarkMode) {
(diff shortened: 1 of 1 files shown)


yarolegovich:
> These changes remove dark mode for the whole app while devtools is open, right? I wonder if there's a way to just scope them to devtools component. Perhaps @thure has an idea there. Would nesting a ThemeProvider with hardcoded light mode help?

Yeah, correct.
Nesting ThemeProvider was my first try, but it didn't work and I decided not to spend too much time trying to understand why. If I just missed smth then yeah, that'd be a better solution.

thure:
I think a solution like this is risky and likely to encourage further color token confusion in future feature work.

Instead, devtools should be refactored to support dark mode. It’s much easier now to choose color tokens using the new semantic tokens, and finding tokens that need replacing in devtools is also not difficult.

If this has become important, I can try to squeeze it in, cc @zhenyasav.

yarolegovich:
ok, closing the PR, will just keep it as a local change

Issue #5557, opened 2024-02-06
Space Fragmentation


dmaretskyi:
- [x] New doc structure
- [x] Migrate space key to "access" section in DocStructure
- [x] Object re-binding
- [x] Make sure doc load timeouts dont deadlock the database + check if they are needed after AM bump
- [x] Adapt automerge-db to handle a set of doc handles
- [x] Automerge DB scans the doc handle set and re-binds objects based on inline and linked entries in the DocStructure
  - Prioritize inline over linked
  - New objects are created as linked
  - Load linked docs lazily using `resolveLink` as trigger

PR #5656, opened 2024-02-13
feat(echo): new automerge space document structure WIP

yarolegovich:
### Motivation / Background

Partially addresses #5557

The goal of these changes is to resolve performance issues occurring from all objects in a space being store in one Automerge document.

### Details

* Introduce new `SpaceDoc` interface that supports linked and embedded objects.
* When creating new objects create `SpaceDoc` instances for them and links to them from space root document.
* In AutomergeDb keep track of events emitted by different DocHandle objects.
* Init EchoObjects when their corresponding documents are resolved.


packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -144,7 +144,7 @@ export class DataSpaceManager {
 
     const automergeRoot = this._automergeHost.repo.create();
     automergeRoot.change((doc: any) => {
-      doc.experimental_spaceKey = spaceKey.toHex();
+      doc.access = { spaceKey: spaceKey.toHex() };
```
dmaretskyi, line 147: `delete doc.experimental_spaceKey`

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -218,36 +231,125 @@ export class AutomergeDb {
...
-    .filter(Boolean);
+  private _createObjectOnDocumentLoad(objectId: string, handle: DocHandle<SpaceDoc>) {
+    handle
+      .doc(['ready'])
+      .then(() => {
+        if (this._ctx?.disposed ?? true) {
+          log.warn('document loaded after database was closed');
+          return;
+        }
+        handle.on('change', this._onDocumentUpdate.bind(this));
```
dmaretskyi, line 301: clear those subscriptions when db closes: `this._ctx.onDispose(() => ....)`

yarolegovich: this is handled in `_onDispose`

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -218,36 +231,125 @@ export class AutomergeDb {
...
+      event.handle,
+      updatedObjects.filter((id) => !this._objects.has(id)),
+    );
+    this._emitUpdateEvent(updatedObjects);
+  }
+
+  private _onDispose() {
+    this._spaceRootDocHandle?.off('change');
+    for (const docHandle of this._objectDocumentHandles.values()) {
+      docHandle.off('change');
```
dmaretskyi, line 279: doesn't this remove all listeners? This sounds unsound since AutomergeRepo maintains a map of docHandles to be shared with all consumers

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -218,36 +231,125 @@ export class AutomergeDb {
...
+  }
+
+  private _linkObjectDocument(object: AutomergeObject, handle: DocHandle<SpaceDoc>) {
+    this._objectDocumentHandles.set(object.id, handle);
+    this._spaceRootDocHandle.change((newDoc: SpaceDoc) => {
+      if (newDoc.links) {
+        newDoc.links[object.id] = handle.url;
+      } else {
+        newDoc.links = { [object.id]: handle.url };
+      }
```
dmaretskyi, line 247:
```suggestion
      newDoc.links ?? = {};
      newDoc.links[object.id] = handle.url;
```

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -218,36 +231,125 @@ export class AutomergeDb {
     }
   }
 
+  private _initDocAccess(handle: DocHandle<SpaceDoc>) {
+    handle.change((newDoc: SpaceDoc) => {
+      newDoc.access = { spaceKey: this.spaceKey.toHex() };
```
dmaretskyi, line 236: Even though this only runs when `acesss === null`, doing it this way still feels a bit safer:

```suggestion
      newDoc.access ??= {};
      newDoc.access.spaceKey = this.spaceKey.toHex();
```

what do you think?

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -182,10 +185,17 @@ export class AutomergeDb {
 
     invariant(isAutomergeObject(obj));
     invariant(!this._objects.has(obj.id));
+
+    const spaceDocHandle = this.automerge.repo.create<SpaceDoc>();
+    this._initDocAccess(spaceDocHandle);
+
     this._objects.set(obj.id, obj);
+    spaceDocHandle.on('change', this._onDocumentUpdate.bind(this));
+    this._linkObjectDocument(obj, spaceDocHandle);
```
dmaretskyi, line 194: Right now, because we don't have an object index, this will break queries (they rely on all objects being in memory). The interim solution is to have a hardcoded list of object types that go into linked storage. For now this is only TextObject that gets put into linked storage, other types go into inline map.

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
yarolegovich: I don't mind applying the suggestion, just curious why does it feel safer?

packages/core/echo/echo-schema/src/automerge/automerge-doc-loader.ts
```diff
@@ -0,0 +1,286 @@
...
+        retryLoading: shouldRetryLoading,
+        err,
+      });
+      if (shouldRetryLoading) {
+        await this._createObjectOnDocumentLoad(handle, objectId);
+      }
+    }
+  }
+
+  private _getInlineAndLinkChanges(event: DocHandleChangePayload<SpaceDoc>) {
```
dmaretskyi, line 252: if this method isn't using any class fields/methods it could be extracted as a helper function in this file

packages/core/echo/echo-schema/src/automerge/automerge-doc-loader.ts
```diff
@@ -0,0 +1,286 @@
...
+  public onObjectCreatedInDocument(handle: DocHandle<SpaceDoc>, objectId: string) {
+    this._objectDocumentHandles.set(objectId, handle);
+    this._createdObjectIds.add(objectId);
+  }
+
+  public onObjectRebound(handle: DocHandle<SpaceDoc>, objectId: string) {
+    this._objectDocumentHandles.set(objectId, handle);
+  }
+
+  public setDocumentLoadingListener(listener: DocumentLoadingListener | null) {
```
dmaretskyi, line 118: use Event from @dxos/async

dmaretskyi (approved):

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
dmaretskyi: I imagine future versions might store more info inside `access`. This code would delete every field inside `access`

dmaretskyi: My conceptual model was that everything in the SpaceDoc up until user data is treated as a key-value store. So intermediate objects like `access` are just lazily initialized on write but never overriden

yarolegovich: > I imagine future versions might store more info inside access. This code would delete every field inside access

that's a sound argument 👍

PR #5657, opened 2024-02-13
fix: Fix warning on Storage config loading

mykola-vrmchk:


mykola-vrmchk:
* `main`
  * **PR #5657** <a href="https://app.graphite.dev/github/pr/dxos/dxos/5657?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dxos/dxos/5657?utm_source=stack-comment).

packages/sdk/config/src/loaders/browser.js
```diff
@@ -38,18 +38,21 @@ export const Defaults = () => {
...
+    if (!localforage) {
+      localforage = await import('localforage');
+    }
     const config = await localforage.getItem('dxos.org/settings/config');
     if (config) {
       return config;
     }
   } catch (err) {
-    log.warn('Failed to load config', { err });
+    log('Failed to load config', { err });
```
wittjosiah, line 55: Why did you make this not a warning? Seems like a valid warning, without it we wouldn't have known this was failing. Really it should be `log.catch` since it's an error that happened, was handled, but we'd still want to know.

packages/sdk/config/src/loaders/browser.js
```diff
@@ -38,18 +38,21 @@ export const Defaults = () => {
...
 
+let localforage = null;
 /**
  * Settings config from browser storage.
  */
 export const Storage = async () => {
   try {
-    const localforage = await import('localforage');
+    if (!localforage) {
+      localforage = await import('localforage');
```
wittjosiah, line 48: My assumption was that once this code path was run once this wouldn't re-load the code but run access the already loaded code. Is that not the case?

packages/apps/plugins/plugin-debug/src/components/DebugSettings.tsx
```diff
@@ -34,12 +33,7 @@ export const DebugSettings = ({ settings }: { settings: DebugSettingsProps }) =>
   const client = useClient();
   const download = useFileDownload();
   // TODO(mykola): Get updates from other places that change Config.
-  const [storageConfig, setStorageConfig] = React.useState<ConfigProto>();
-
-  useAsyncEffect(async () => {
-    const config = await Storage();
-    config && setStorageConfig(config);
-  }, []);
+  const [storageConfig, setStorageConfig] = React.useState<ConfigProto>(client.config.values);
```
wittjosiah, line 36: nit: prefer importing `useState` to `React.useState`

packages/sdk/config/src/loaders/browser.js
mykola-vrmchk: That is valid remark

Issue #5473, opened 2024-01-31
incorrect peer dependencies in Docker build


nferch:
pnpm is reporting incorrect peer dependencies when doing the `pnpm deploy` step of the Docker build:

```
#14 28.48  ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies
#14 28.48 
#14 28.48 packages/devtools/cli
#14 28.48 └─┬ @dxos/observability 0.3.11-main.c865ffb
#14 28.48   └─┬ @dxos/react-client 0.3.11-main.c865ffb
#14 28.48     └─┬ @dxos/react-ui 0.3.11-main.c865ffb
#14 28.48       ├── ✕ missing peer @phosphor-icons/react@^2.0.5
#14 28.48       ├─┬ @dxos/react-input 0.3.11-main.c865ffb
#14 28.48       │ └─┬ @radix-ui/react-primitive 1.0.3
#14 28.48       │   └── ✕ missing peer @types/react-dom@"*"
#14 28.48       ├─┬ @fluentui/react-tabster 9.17.4
#14 28.48       │ └── ✕ missing peer @types/react-dom@">=16.9.0 <19.0.0"
#14 28.48       ├─┬ @radix-ui/react-context-menu 2.1.5
#14 28.48       │ ├── ✕ missing peer @types/react-dom@"*"
#14 28.48       │ └─┬ @radix-ui/react-menu 2.0.6
#14 28.48       │   ├── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   ├─┬ @radix-ui/react-collection 1.0.3
#14 28.48       │   │ └── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   ├─┬ @radix-ui/react-dismissable-layer 1.0.5
#14 28.48       │   │ └── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   ├─┬ @radix-ui/react-focus-scope 1.0.4
#14 28.48       │   │ └── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   ├─┬ @radix-ui/react-popper 1.1.3
#14 28.48       │   │ ├── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   │ └─┬ @radix-ui/react-arrow 1.0.3
#14 28.48       │   │   └── ✕ missing peer @types/react-dom@"*"
#14 28.48       │   ├─┬ @radix-ui/react-portal 1.0.4
[...]

PR #5663, opened 2024-02-13
chore: check for missing peer dependencies on PR

nferch:
### Motivation / Background
https://github.com/dxos/dxos/issues/5473

`pnpm` only warns on missing peer dependencies, which can cause problems in downstream builds.

### Details
`pnpm` lacks functionality to check for missing peer dependencies (https://github.com/pnpm/pnpm/issues/6893), but we can workaround by using `install --no-lockfile` which will fail if there are missing peer dependencies.


wittjosiah (approved):

nferch:
Does not currently work, see https://github.com/dxos/dxos/pull/5685/files#diff-d457dd0159c665b772f2e7a6a8e0697e212cb9f4c0f8f15ab3ff80f3bb4abc59R112

PR #5671, opened 2024-02-14
fix(plugin-markdown): enable vim mode doesn't get stuck

### LINKED ISSUE #5642: [composer] changing the editor mode doesn't take effect immediately
### Repro

- create a document and navigate to it
- open settings and change the editor mode (either Default or VIM mode, doesn't matter)
- hit "Done" on settings
- the editor mode has not changed
- navigate to a different item and navigate back to the document
- the editor mode has now changed

wittjosiah:
### Motivation / Background

Resolves #5642

### Details

- extensions changes did not cause the editor view to be recreated
- this PR makes sure they do but this means that extensions arrays need to be memoized always

### Additional information

- attempted to use `useTextEditor` to accomplish this, but it's not ready yet
  - it's very easy to introduce bugs with it where the editor is rendered multiple times
  - it does not handle strict mode so isn't concurrency-safe


packages/ui/react-ui-editor/src/hooks/useTextEditor.ts
```diff
@@ -45,46 +46,52 @@ export const useTextEditor = ({
...
+        }
+        view.update(trs);
+      },
+    });
+
+    view?.destroy();
+    setView(newView);
+
     return () => {
       view?.destroy();
```
richburdon, line 90: so this should be newView.destroy?
(why i preferred having a single variable -- so that only the current scope is considered).

PR #5685, opened 2024-02-15
fix: cli observability regressions

nferch:


nferch:
cc @wittjosiah re: missing tags restored at https://github.com/dxos/dxos/pull/5685/files#diff-d457dd0159c665b772f2e7a6a8e0697e212cb9f4c0f8f15ab3ff80f3bb4abc59R112

PR #5692, opened 2024-02-15
feat(native): Added support for command palette global hotkey

mjamesderocher:
This PR adds support for the global hotkey (cmd + k) on the Mac native app bundled with Socket Supply Co. 

It also will bring the window into focus when a URL is opened from the browser.

NOTE: This is currently only working if Socket is built from their Master branch on Github, since this PR relies on features not in the current package released on NPM.

https://github.com/dxos/dxos/assets/6023999/9a7ce20e-99d4-4cd1-8e1d-2f148b4a6add


packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -117,12 +123,23 @@ const initializeNativeApp = async (plugins: Plugin[]) => {
...
+    appWindow.restore();
     void intentPlugin?.provides.intent.dispatch({
       action: NavigationAction.ACTIVATE,
       data: { id: url.host },
     });
   });
 
+  // Global hotkey listener
+  binding.addEventListener('hotkey', () => {
+    console.log('Key pressed');
```
zhenyasav, line 135:
```suggestion

```

packages/apps/plugins/plugin-native/src/NativePlugin.tsx
```diff
@@ -23,9 +23,15 @@ const initializeNativeApp = async (plugins: Plugin[]) => {
...
   const module = 'socket:application';
   const app = await import(/* @vite-ignore */ module);
+  const appWindow = await app.getCurrentWindow();
+  const windowModule = 'socket:window';
+  const socketWindow = await import(/* @vite-ignore */ windowModule);
   const { meta_title: appName } = app.config;
   const intentPlugin = resolvePlugin(plugins, parseIntentPlugin);
 
+  const binding = await socketWindow.hotkey.bind('cmd + k');
+  console.log(binding);
```
zhenyasav, line 33:
```suggestion

```

Issue #5683, opened 2024-02-14
[shell] hide agent mgmt behind isInternal


zhenyasav:
same isInternal from the telemetry spec

wittjosiah:
FYI @zhenyasav this is implemented in our observability as the `group` being `dxos`

PR #5727, opened 2024-02-19
feat(shell): hide agent hosting for non-internal

nferch:
closes https://github.com/dxos/dxos/issues/5683


nferch:
probably not doing anything more fancy until we have more need for persistent runtime config, external service discovery, agent hosting versioning, etc? WDYT @wittjosiah

packages/sdk/react-client/src/client/useAgentHostingProvider.ts
```diff
@@ -11,7 +11,14 @@ import { type Config } from '@dxos/config';
 import { log } from '@dxos/log';
 
 export const useAgentHostingProviderClient = (config: Config): AgentHostingProviderClient | null => {
```
wittjosiah, line 13: Since this is being exported from a core sdk package, can we add a jsdoc here which notes this as experimental in some way? one way we've done this is just by marking it as deprecated

wittjosiah (approved):
What I had in mind was using the observability group directly rather than storing another field for it. I can see how that might be non-trivial to wire all the way through from observability init in the shell entry point to the identity panel though, given that we don't have any sort of hooks or provider for accessing the observability class.

Given that, and the todo about configuring, I think this seems like a reasonable interim step.

packages/sdk/react-client/src/client/useAgentHostingProvider.ts
nferch: yeah, wasn't quite sure how to capture that 1) It's experimental and the internal and external API is unstable 2) It's a feature we're developing internally and not making the backend or infrastructure open source, 3) it's coupled to a product we're developing, which is why the client is kind of a road to nowhere, but doesn't preclude you from running your own agent, or even building a hosting platform of your own that matches the API

nferch: ahh, I had forgotten about that aspect

PR #5729, opened 2024-02-20
chore(plugin-observability): tag errors with plugins enabled

wittjosiah:
Help with debugging errors by tagging them with plugins that were enabled at the time.


packages/apps/plugins/plugin-observability/src/ObservabilityPlugin.tsx
```diff
@@ -110,6 +112,10 @@ export const ObservabilityPlugin = (options: {
       void options.observability().then(async (obs) => {
         observability = obs;
 
+        // Ensure errors are tagged with enabled plugins to help with reproductions.
+        const pluginsEnabled = pluginHost?.provides?.plugins?.enabled?.join(',');
```
nferch, line 116: it'd be better for cardinality to set individual tags for each plugin enabled. and also enable searching and filtering by individual plugin, i'm not sure if you can search by substring.

wittjosiah: @nferch you mean using the plugin id as the key and just `true` as the value? I was less concerned about querying and more concerned with just having the information available when looking at an instance of the error. That being said, being able to query/filter by it would probably be helpful.

nferch: yeah, something like 
```
(pluginHost?.provides?.plugins?.enabled ?? []).map((p) => observability.setTag('pluginEnabled-'+p, true, 'errors'))
```

unless there's some other way to include metadata about the error? might align better with more comprehensive structured logging which we may want at some point.

PR #5746, opened 2024-02-22
chore: speed up codecov upload

nferch:


wittjosiah (approved):
It supports directories natively now? Excellent.

PR #5765, opened 2024-02-23
chore(docs): fix telemetry config

wittjosiah:


nferch (approved):
Should we make mode a required option? 'disabled` being the default feels a bit non-obvious

PR #5766, opened 2024-02-23
fix(plugin-thread): deleting threads works again

wittjosiah:
### Motivation / Background

This Pull Request fixes deleting threads.

### Details

- the document wasn't being passed to the intent properly

### Additional information

- first, we should create a pattern which ensures intent data is type-safe (arena app is an example)
- second, we should consider stopping using `document` as a variable name for document objects


thure (approved):
Aha

Issue #4884, opened 2023-12-07
[sdk] Delegated device invitations


nferch:
# Use cases

## "Offline" agent provisioning

It should be possible to give an agent permission to join as a device of an identity without a peer being online at the time of activation. This would happen through a managed agent provisioning flow, or when manually provisioning an agent started through the CLI.

In a managed provisioning flow, once the user has granted the permission for a managed hosting provider to create an agent on their behalf, they should not need to remain online for the provider to complete the provisioning flow.

## Managed agent migrations

A managed hosting provider should be able to create an agent using some stored credential, which would enable the ability for the hosting provider to destroy and create an agent without maintaining the local storage of an agent.

This could be used when upgrading the software of the agent, migrating the agent's runtime between runtime nodes, or recovering from a corrupted or crashed agent.

richburdon:
The agent can only be admitted by peers that have a device (signing) key that belongs to the same identity (user).

Proposal:
1. Shell panel triggers Agent deployment and shows "provisioning..."
2. Shell can be closed.
3. When Shell is reopened it polls the Agent API to get a status.
4. If the status is ready, it requests the public key and the user does a direct admission (writes a credential to the HALO).

PR #5773, opened 2024-02-24
feat: persistent invitations

nferch:
### Motivation / Background

* Attempts to partially solve some of the needs for offline/async invitations (see https://github.com/dxos/dxos/issues/5129 and https://github.com/dxos/dxos/issues/4884)
* Address some of the existing invitation behavior (e.g. invitations never expire as long as the client stays running)

### Details
* Invitations are now persistent by default, which means the client will save them to local storage and restart them if the client
* Invitations have a default lifetime of one day, at which point they will be automatically canceled.
* Open invitations can now be listed in the CLI using `agent invitations`. Space invitations in the shell show the expiry by mouseover'ing the QR code button.

### Additional information
* As before, the peer that initiated that the invitation must be online in order for the guest to accept the invitation. Thinking about an API to remotely request an agent to generate an invitation, which would alleviate more of the needs for offline invitations.


packages/core/echo/echo-pipeline/src/metadata/metadata-store.ts
```diff
@@ -247,6 +248,29 @@ export class MetadataStore {
     await this.flush();
   }
 
+  getPersistentInvitations(): Invitation[] {
+    return this._metadata.persistentInvitations ?? [];
+  }
+
+  async addPersistentInvitation(invitation: Invitation) {
+    // TODO(nf): lock?
```
dmaretskyi, line 256: _save is already synchronized so it should be fine without a lock

packages/sdk/client/src/invitations/invitations-proxy.ts
```diff
@@ -156,6 +156,24 @@ export class InvitationsProxy implements Invitations {
     return observable;
   }
 
+  async resumePersistentInvitations() {
```
dmaretskyi, line 159: What happens here if two separate clients pointing to the same service call this? Are the invitaitons gonna be doubled?

I think should should be a part of the service-host startup sequence instead.

packages/sdk/client-services/src/packlets/invitations/invitations-service.ts
```diff
@@ -31,6 +34,8 @@ export class InvitationsServiceImpl implements InvitationsService {
   constructor(
     private readonly _invitationsHandler: InvitationsHandler,
     private readonly _getHandler: (invitation: Invitation) => InvitationProtocol,
+    // TODO(nf): avoid making InvitationManager?
```
dmaretskyi, line 37: what do you mean?

packages/sdk/client-services/src/packlets/invitations/invitations-service.ts
```diff
@@ -43,6 +48,7 @@ export class InvitationsServiceImpl implements InvitationsService {
   createInvitation(options: Invitation): Stream<Invitation> {
     let invitation: CancellableInvitation;
 
+    const savePersistentInvitationCtx = new Context();
```
dmaretskyi, line 51: unhandled error here (from `addPersistentInvitaion`).

I'd move the whole `scheduleTaskCall` into `return new Stream<Invitation>(({ next, close }) => {` so that you can bubble errors up

packages/sdk/client/src/invitations/invitations-proxy.ts
nferch: hmm, good question. I don't think the invitation will be duplicated because createInvitation checks for existing invitations.

not sure about what the implications of multiple observables are, it might actually be beneficial so both clients get notified when an invitation reaches a terminal state. how would that work if this moves to the service-host startup?

packages/core/echo/echo-pipeline/src/metadata/metadata-store.ts
```diff
@@ -247,6 +248,29 @@ export class MetadataStore {
     await this.flush();
   }
 
+  getPersistentInvitations(): Invitation[] {
```
richburdon, line 251: since the context does the "persistent" prefix add anything? vs. getInvitations.
i'm not sure. i.e., i would not expect to get non-persistent invitations from a store. same comment throughout about decoupling the name.

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -442,6 +442,12 @@ message Invitation {
 
   /// Path or identifier to navigate to after successful authentication.
   optional string target = 11;
+
+  /// Host should resume invitation on startup until timeout.
+  optional bool persistent = 12;
+
+  /// Host should resume invitation on startup until expiry.
+  optional google.protobuf.Timestamp persistence_expiry = 13;
```
richburdon, line 450: 1. consider adding a timestamp field for the creation and a separate expiration or TTL field. also the field could just be `expiration`. e.g., we might want a policy to set a max TTL, so we'd need the creation timestamp for that.

2. is it required to have a field to designated "persistent" i.e., is that usefully part of the data structure. from the usage below it could be an option on the method. IOW, what is the use case: does the user know ahead of time that they are creating a persistent invitation?

3. how is that different from TYPE = OFFLINE? in the OFFLINE docs above we required a public key -- is there a different auth method we can implement? e.g., requiring a PIN. @dmaretskyi?

packages/sdk/client-services/src/packlets/invitations/invitations-handler.ts
```diff
@@ -73,6 +73,8 @@ export class InvitationsHandler {
       state = Invitation.State.INIT,
       timeout = INVITATION_TIMEOUT,
       swarmKey = PublicKey.random(),
+      persistent = false,
```
richburdon, line 76: skip since false is the default. also skip undefined.

packages/sdk/client-services/src/packlets/invitations/invitations-handler.ts
```diff
@@ -181,6 +185,13 @@ export class InvitationsHandler {
     };
 
     let swarmConnection: SwarmConnection;
+
+    // TODO(nf): cancel invitations when the persistence deadline is reached.
+    // TODO(nf): honor some deadline for non-persistent invitations as well?
```
richburdon, line 190: right -- re my comment to just have a field named "expiration" (and add a "created" TS).

packages/sdk/client-services/src/packlets/invitations/invitations-service.ts
```diff
@@ -72,6 +90,22 @@ export class InvitationsServiceImpl implements InvitationsService {
     });
   }
 
+  async getPersistentInvitations(): Promise<GetPersistentInvitationsResponse> {
+    const persistentInvitations = this._metadataStore.getPersistentInvitations();
+
+    // get saved persistent invitations, filter and remove from storage those that have expired.
+    const freshInvitations = persistentInvitations.filter(async (invitation) => {
```
richburdon, line 97: Meta: Given the scale, this is of course, but in general I might have a separate function to flush expired invitations (something that can be polled periodically).

And in general possibly `removeInvitations(ids)` so that you first get a filtered set (synchronously) then atomically delete them all in one call.

packages/sdk/shell/src/panels/SpacePanel/steps/SpaceManager.tsx
```diff
@@ -80,6 +80,22 @@ export const SpaceManager = (props: SpaceManagerProps) => {
         }
       }, [space]),
     },
+    invitePersistent: {
```
richburdon, line 83: ok re question above i'll ask the product team what they have in mind about the need for the user to make this decision.

packages/sdk/client-services/src/packlets/invitations/invitations-service.ts
nferch: that the service is just a passthrough for the storage and doesn't have any state or logic of its own.

packages/sdk/shell/src/panels/SpacePanel/steps/SpaceManager.tsx
nferch: i think there's also some decisionmaking/intent on the app developer as well. and as you noted, some interaction between the invitation parameters, e.g. is an invitation requiring authentication useful to be persistent?

i think it may be simpler to make all invitations persistent, but then i worry about them accumulating, and how to make it clear to the user that there's outstanding invitations, as much as those are separate concerns.

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -450,6 +451,17 @@ message Invitation {
...
+  // TODO(nf): should these exist elsewhere since they only have meaning to the host?
+  /// Host should resume invitation on startup until timeout.
+  optional bool persistent = 12;
+
+  optional google.protobuf.Timestamp created = 13;
+
+  /// Number of seconds the host will honor the invitation. 0 means the invitation persists forever.
+  // TODO(nf): some feedback mechanism or GC for immortal invitations?
+  // TODO(nf): should the creator have a way to make an invitation last as long as the client? this was the previous behavior.
+  optional int32 lifetime = 14;
```
dmaretskyi, line 464: change to expiration ts?

nferch: lol I had that originally, but I think Rich's comment swayed me to the benefits of seeing the TTL as a developer without having to calculate if we want creation as well anyway.

i don't have strong opinions either way, but I'm curious on your rationale?

PR #5776, opened 2024-02-26
fix: Editor scrolling hack

richburdon:
- [x] Remove hack to position initial scroll. (Still requires scrollTo in TextEditor.)
- [x] Support scrollTo, overscroll (via CM extension), and positioning cursor at end of line as text editor options.

Issues
- tooltips now always show when using the toolbar (e.g., select heading); title worked better.
- react-ui-editor style re grid should possibly be removed (see TODO and storybook issue below)

Screenshot 2024-02-26 at 1 12 19 PM


packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -75,23 +69,21 @@ const EditorMain = ({ model, comments, toolbar, extensions: _extensions, ...prop
...
         >
           <Toolbar.Markdown />
           <Toolbar.Separator />
           <Toolbar.Extended />
         </Toolbar.Root>
       )}
-      <div
-        role='none'
-        data-toolbar={toolbar ? 'enabled' : 'disabled'}
-        className='is-full bs-full overflow-hidden data-[toolbar=disabled]:pbs-8'
```
thure, line 90: Why did you remove `data-[toolbar=disabled]:pbs-8`? This was added to address a complaint that the editor was too close to the top of the viewport when the toolbar is disabled.

If it is more valuable not to do that (I would encourage checking what it looks like with the toolbar disabled), then please also remove the `data-toolbar` attribute which is no longer necessary.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -103,15 +95,13 @@ const EditorMain = ({ model, comments, toolbar, extensions: _extensions, ...prop
...
               'data-testid': 'composer.markdownRoot',
             } as HTMLAttributes<HTMLDivElement>,
-            editor: {
-              className: mx(editorFillLayoutEditor, !toolbar && 'border-bs'),
-            },
             content: {
-              // after:block after:is-px after:bs-px after:overflow-anchor after:-mbs-px
-              className: mx(editorHalfViewportOverscrollContent, '!p-2 sm:!p-6 md:!p-8'),
+              // TODO(burdon): Override (!) required since base theme sets padding and scrollPastEnd sets bottom.
+              className: mx('!px-2 sm:!px-6 md:!px-8 !pt-2 sm:!pt-6 md:!pt-8'),
```
thure, line 104:
```suggestion
              className: mx('!pli-2 sm:!pli-6 md:!pli-8 !pt-2 sm:!pbs-6 md:!pbs-8'),
```
Please always use logical properties

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -83,12 +97,7 @@ export const BaseTextEditor = forwardRef<EditorView | null, TextEditorProps>(
 
     // Set focus.
     useEffect(() => {
-      if (autoFocus && !view?.hasFocus) {
-        if (view?.state.selection.main?.from === 0) {
-          // Start at end of line.
-          const { to } = view.state.doc.lineAt(0);
-          view?.dispatch({ selection: { anchor: to } });
-        }
+      if (autoFocus) {
```
thure, line 100: Ensure this change of predicate does not impact threads.

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -254,7 +272,7 @@ export const defaultSlots: TextEditorSlots = {
     className: focusRing,
   },
   editor: {
-    className: 'min-bs-full',
+    className: 'h-full overflow-scroll',
```
thure, line 275:
```suggestion
    className: 'bs-full overflow-scroll',
```

packages/ui/react-ui-editor/src/styles/layout.ts
```diff
@@ -2,11 +2,8 @@
 // Copyright 2024 DXOS.org
 //
 
+// TODO(burdon): Remove these styles which are specific to plugin usage, not component?
```
thure, line 5: These styles are here mainly because Toolbar is also defined here, and it was useful to offer fragments for laying the toolbar and editor in an expected configuration. They are also used in the stories.

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
thure: It’s vitally important to never use `overflow-scroll`. We should find a way to remove it from the design system entirely.

Furthermore, here it is applied to an element that does not overflow (your approach maintains that `cm-scroller` has the scrollbar), so it forces it to apply padding for a scrollbar that it will never use. See screenshots:

Screenshot 2024-02-26 at 09 54 15
Screenshot 2024-02-26 at 09 51 23

thure:
Regarding:
> tooltips now always show when using the toolbar (e.g., select heading); title worked better.
We can’t rely on `title`, it doesn’t show a tooltip on mobile devices. The issue you refer to can be fixed with our `suppressNextTooltip` pattern. I can commit that to this branch if you like.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
richburdon: my guess is the feedback was since previously the text area padding wasn't working since  the `!md:px-8` was `md:!px-8` and not working?

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
richburdon: do you have a cheatsheet since the TW docs don't explain these as far as i could find:
https://tailwindcss.com

thure:
Regarding:

> react-ui-editor style re grid should possibly be removed (see TODO and storybook issue below)

The fix is simple, apologies for my omission when making the change: `<MainLayout toolbar={toolbar}>`

Screenshot 2024-02-26 at 10 26 23

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
thure: Yes for sure: https://stevecochrane.github.io/tailwindcss-logical/

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
thure: No, I can find the issue if you like, it was because without the toolbar the top edge of the “paper” effect of the document was flush with the bottom edge of the topbar.

If this wasn’t removed to fix something else, I think it’s safest to restore it.

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
richburdon: it's optional.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
richburdon: ah ok

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -196,7 +196,7 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
               if (isDocument(data.active)) {
                 const { readonly } = settings.values.state[data.active.id] ?? {};
                 return (
-                  <MainLayout toolbar={settings.values.toolbar}>
```
thure, line 199: Please restore this — `MainLayout` needs the `toolbar` flag to apply the correct layout when the toolbar is present.

thure (approved):
now lgtm!

Issue #5701, opened 2024-02-16
[authoring] code block rendering


wittjosiah:
- shouldn't render until the block is closed
- but also large code blocks un-render when closing is hidden?
- why is there a line at the bottom of the code block?

https://github.com/dxos/dxos/assets/4529818/c1eb9874-72c1-474b-b49c-be9b1d9356bf

marijnh:
> shouldn't render until the block is closed

We could turn that off, but if there's any other code block below the unclosed one, the backticks for that one will be taken to close it, so that solution probably doesn't do what you are hoping for.

> but also large code blocks un-render when closing is hidden?

That should be fixed as of f10c1808dc1ff5b55fe02592a8b1de7b63260ef2 already.

> but also large code blocks un-render when closing is hidden?

Looks like a rounding error is causing the browser to put space between these elements. I cannot reproduce it on my platform but I expect a tiny negative margin on these blocks will fix it.

wittjosiah:
> We could turn that off, but if there's any other code block below the unclosed one, the backticks for that one will be taken to close it, so that solution probably doesn't do what you are hoping for.

I think you're right. Testing this again now and it feels reasonable. For some reason I remembered the backticks being hidden after they were typed and so we couldn't tell if it had been closed or not. This doesn't seem to be happening now.

> That should be fixed as of https://github.com/dxos/dxos/commit/f10c1808dc1ff5b55fe02592a8b1de7b63260ef2 already.

yes, looks to be resolved from I can tell

> Submitted a PR that I expect will avoid the gap/line issue

The other thing I noticed was that I couldn't click into the backtick lines in the editor when the cursor was outside the code block. Would that be resolved by the PR as well or is that a separate issue?

dmaretskyi:
Noticed this issue with hidden chracters not showing up

https://github.com/dxos/dxos/assets/35851437/77e6c1e3-9791-47d2-855b-9131fa58249c

wittjosiah:
Yeah, I think the main issue in the video is that the control characters aren't shown even though the cursor is in the code block (as we established further up the thread why the code block needs to be rendered with only opening brackets).

That does make me wonder though, could it automatically add the closing backticks when you open a code block?

marijnh:
The code is assuming there's an end marker on the last line of a fenced code block, and incorrectly hiding that line. This patch should help with that https://github.com/dxos/dxos/pull/5835

marijnh:
[...]

PR #5783, opened 2024-02-26
fix(react-ui-editor): Simplify styling of first/last codeblock lines

marijnh:
This should fix the gap that sometimes shows up between the lines.

Issue https://github.com/dxos/dxos/issues/5701


thure (approved):
Thank you for resolving this! ✨

PR #5788, opened 2024-02-26
chore: update storybook config

richburdon:
- fix dxos/brand storybook names
- attempt to standardize storybook configs 
- so far, failed attempt to configure sorting books and configure manager


wittjosiah:
can you make this a `chore` rather than a `fix` so it doesn't show up in the release notes?

PR #5800, opened 2024-02-28
feat: Support storing and recalling EchoReactiveObject from the database

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -446,21 +461,49 @@ const throwIfCustomClass = (prop: string | symbol, value: any) => {
...
-    const idProperty = AST.getPropertySignatures(schema.ast).find((prop) => prop.name === 'id');
-    if (idProperty != null) {
-      throw new Error('"id" property name is reserved');
+    slot.handler = echoHandler;
+    const target = slot.target as ProxyTarget;
+    target[symbolPath] = [];
+    slot.handler._proxyMap.set(target, proxy);
+    slot.handler._init(target);
+
+    if (schema != null) {
```
yarolegovich, line 483: nit: extract the duplicated block into a function

yarolegovich (approved):

PR #5804, opened 2024-02-28
chore(composer-app): make devtools easier to access

wittjosiah:
### Motivation / Background

This Pull Request ensures that devtools is accessible even if it wasn't turned on beforehand.

### Details

- sets the default values for the debug plugin settings to true
- adds a confirm dialog to enter devtools if you haven't turned it on before


jessmartin (approved):
LGTM!

PR #5805, opened 2024-02-28
feat: don't expand markdown tags if moving up/down through document.

richburdon:
Motivation
When moving up/down through docs with images, links, etc. currently the markdown decorations are removed. this PR only removes the decoration when moving horizontally through the document.
This feature might be only be set in a particular mode.


marijnh:
Is the problem you are trying to address with this the horizontal shifting of the cursor position due to markup appearing when the cursor lands in it?

It looks like this will cause the markup characters to not appear at if you land inside of some construct with vertical motion. That seems inconsistent and confusing—sometimes you see the markup, sometimes you don't, depending on how you arrived at your cursor position.

It feels like we need to do some design work before starting to throw code at this.

marijnh:
Since a similar jarring effect exists for mouse-drag-selection (and probably also for horizontal selection in some situations), I don't think special-casing vertical motion is what we want here.

One possible approach would be to delay updates to the hiding/unhiding of markup when cursor motion happens. And then after a given period of inactivity (half a second, maybe), update it for the current cursor position.

If you want to go really fancy, we could make the extension compute precisely which characters it is hiding/unhiding at that point, and try to show an animation where these smoothly appear/disappear. That'd be more work though.

richburdon:
> Since a similar jarring effect exists for mouse-drag-selection (and probably also for horizontal selection in some situations), I don't think special-casing vertical motion is what we want here.
> 
> One possible approach would be to delay updates to the hiding/unhiding of markup when cursor motion happens. And then after a given period of inactivity (half a second, maybe), update it for the current cursor position.

Agreed. I wasn't sure how to do timed effects but I think this would be effective.

> If you want to go really fancy, we could make the extension compute precisely which characters it is hiding/unhiding at that point, and try to show an animation where these smoothly appear/disappear. That'd be more work though.

Great to explore this at a later stage I think.

marijnh:
See https://github.com/dxos/dxos/pull/5821

PR #5811, opened 2024-02-29
feat: Emoji fallbacks for Avatars

### LINKED ISSUE #5743: [composer] interim custom avatars
- long term we want custom avatars ( #5233 )
- short term we don't have integrated blob storage
- currently we're using jdenticons for avatars, but they are too abstract/not identifiable + can't be customized
- proposed interim solution is to use emoji + background colour:
  - default would be computed based on your identity key similar to jdenticon
  - shell edit profile would allow the user to customize their emoji/colour pair

thure:
Resolves #5743.

This PR removes `jdenticon` entirely, replacing it with an emoji and a hue which the user can configure in `ProfileForm`.

This PR also uses the hue to determine cursor & highlight color in documents, refactoring all places that use similar features to instead rely on sources of truth provided by fragments defined in `hue.ts` in `react-ui-theme`.

https://github.com/dxos/dxos/assets/855039/59a57477-35ee-404d-86da-5d271428290f

#### Caveats
- ⚠️ Keyboard access is broken in the emoji picker
- ⚠️ Scrolling with the mousewheel or trackpad is broken in the emoji picker
- ⚠️ `hue` and `emoji` in the `ProfileDocument` remain unset outside of shell


packages/sdk/shell/src/components/EmojiPicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const [emojiValue, setEmojiValue] = useControllableState<string>({
+    prop: emoji,
+    onChange: onChangeEmoji,
+    defaultProp: defaultEmoji,
+  });
+
+  const [emojiPickerOpen, setEmojiPickerOpen] = useState<boolean>(false);
+
+  // @ts-ignore
```
wittjosiah, line 32: needed?

packages/sdk/shell/src/components/HuePicker.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const HuePicker = ({ disabled, hue, onChangeHue, defaultHue, onClickClear }: HuePickerProps) => {
+  const { t } = useTranslation('os');
+
+  const [hueValue, setHueValue] = useControllableState<string>({
+    prop: hue,
+    onChange: onChangeHue,
+    defaultProp: defaultHue,
+  });
+
+  // @ts-ignore
```
wittjosiah, line 37: needed?

wittjosiah (approved):
We'll track down that persistence issue w/ the profile document but otherwise LGTM

Issue #5755, opened 2024-02-22
[composer] Misc editor issues.


richburdon:
## Smaller

- [x] Decorator bug with large code blocks.
  - https://github.com/dxos/dxos/issues/5701 
- [x] Do no remove decorations when traversing via up/down key until after delay.
- [x] Render images when scrolling through document.
- [ ] @Mention (selective autocomplete)
  - Auto-complete on typing if `@` otherwise require ctrl-space (e.g., to complete page link dialog)
  - Show "/" tooltip on blank line with popup menu.
- [ ] Smooth scroll into view; ideally able to position in the middle of the viewport.

## Larger

- [ ] Language extension for schema.
  -  Provide example for simple extension to JS language (e.g., https://effect-ts.github.io/schema/modules/Schema.ts.html)
- [ ] Hierarchical editor (e.g., similar to logseq to replace current OutlinerPlugin):
  - drag-and-drop rows (current `dxos/mosaic` lib allows DND across components -- e.g., to/from sidebar).

richburdon:
Scrolling issues:
- don't expand links until delay
- images don't render until select
- jitters when scrolling

Uploading Screen Recording 2024-02-27 at 10.53.56 PM.mov…

marijnh:
PR https://github.com/dxos/dxos/pull/5799 seems to help with the images loading late and flickering when scrolled into view.

PR #5821, opened 2024-03-01
fix(react-ui-editor): Delay hiding/unhiding of markdown markup during cursor motion

marijnh:
To avoid disruptive popping of the content around the cursor.

Issue https://github.com/dxos/dxos/issues/5755

### DIFF
--- a/packages/ui/react-ui-editor/src/extensions/markdown/decorate.ts
+++ b/packages/ui/react-ui-editor/src/extensions/markdown/decorate.ts
@@ -3,7 +3,7 @@
 //
 
 import { syntaxTree } from '@codemirror/language';
-import { RangeSetBuilder, type EditorState } from '@codemirror/state';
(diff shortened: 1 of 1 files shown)


richburdon:
I think this is useful, but it seems to introduce a problem where you can't click or cursor into the link label.

marijnh:
Good point. I made all of the decorations atomic (so that you don't get your cursor stuck in invisible code when arrowing through markers), but only the ones that hide stuff should be. Updated the PR to fix that.

richburdon:
Seeing another issue where the cursor loses its position (floats in whitespace) after the decoration is restored. It only seems to happen on short lines.

marijnh:
That turned out to be an issue in CodeMirror itself. https://github.com/dxos/dxos/pull/5833 pulls in an upgrade that fixes it.

PR #5839, opened 2024-03-04
chore: wrap graph updates in batch

wittjosiah:
This pull request batches graph updates in plugin graph builders in order to improve performance by avoiding recomputing the navtree more than is necessary.


dmaretskyi:
did it have the desired effect?

wittjosiah:
I'm not sure if it resolves all the problems but it certainly helps

PR #5845, opened 2024-03-05
feat(echo): in-memory effect schema registration for lookups on restore

yarolegovich:
### Motivation / Background

When we're restoring saved objects we need to be able to attach effect Schema objects to them.
For this we maintain an in-memory map of typename -> schema, forcing users to register their types through Hypergraph types object. 

### Details

* Expose an API for registering effect schema in `Hypergraph`
* Store schema in automerge doc as a legacy encoded `Reference` under `[objects, id, system, type]` for compatibility with old objects
* Decode the `Reference` and use it for `Schema` lookup in `Hypergraph` when we're constructing an object on startup
* throw if a `Schema` is not annotated with `EchoObjectAnnotation`


packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -404,7 +428,7 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
     inspectFn: (value: any, options?: InspectOptionsStylized) => string,
   ) {
     const handler = this[symbolHandler] as EchoReactiveHandler;
-    const isTyped = Boolean((this as any)[symbolSchema]);
+    const isTyped = Boolean(handler._objectCore.getType());
```
dmaretskyi, line 431:
```suggestion
    const isTyped = !!handler._objectCore.getType();
```

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -58,15 +58,13 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
     invariant(!(target as any)[symbolIsProxy]);
     invariant(Array.isArray(target[symbolPath]));
 
-    if (target[symbolSchema]) {
-      SchemaValidator.initTypedTarget(target);
-    } else {
-      this.makeUntypedArraysReactive(target);
-    }
+    this.makeUntypedArraysReactive(target);
```
dmaretskyi, line 61: target will be empty...

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -58,15 +58,13 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
...
 
     if (target[symbolPath].length === 0) {
       this.validateInitialProps(target);
-      this._objectCore.initNewObject(target);
+      if (this._objectCore.mountPath.length === 0) {
```
dmaretskyi, line 65: extract as a separate method?

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -258,11 +256,33 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
 
   private validateValue(target: ProxyTarget, prop: string | symbol, value: any): any {
     throwIfCustomClass(prop, value);
-    return (target as any)[symbolSchema] ? SchemaValidator.validateValue(target, prop, value) : value;
+    const rootObjectSchema = this.getSchema();
+    if (rootObjectSchema == null) {
+      return value;
```
dmaretskyi, line 261: always returns value

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -258,11 +256,33 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
...
+    return value;
+  }
+
+  getSchema(): S.Schema<any> | undefined {
+    invariant(this._objectCore.database, 'EchoHandler used without database');
+    const typeReference = this._objectCore.getType();
+    if (typeReference == null) {
+      return undefined;
+    }
+    const effectSchema = this._objectCore.database.graph.types.getEffectSchema(typeReference.itemId);
```
dmaretskyi, line 277: TODO: Reactivity

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -396,6 +416,10 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
     this._signal.notifyWrite();
   }
 
+  private getPropertyMountPath(path: PropPath): string[] {
```
dmaretskyi, line 419:
```suggestion
  private _getPropertyMountPath(path: PropPath): string[] {
```

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -481,33 +507,50 @@ export const createEchoReactiveObject = <T extends {}>(init: T): EchoReactiveObj
...
 
+export const initEchoReactiveObjectRootProxy = (core: AutomergeObjectCore) => {
+  const target = { [symbolPath]: [] };
+  const handler = new EchoReactiveHandler();
+  handler._objectCore = core;
+  handler._objectCore.rootProxy = createReactiveProxy<ProxyTarget>(target, handler) as any;
+};
+
+const validateSchema = (schema: S.Schema<any>) => {
+  invariant(isTypeLiteral(schema.ast));
```
dmaretskyi, line 530: also check that there's an `echoObject` annotation

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -481,33 +507,50 @@ export const createEchoReactiveObject = <T extends {}>(init: T): EchoReactiveObj
...
+    const encodedRef = handler._objectCore.encode(getSchemaTypeRefOrThrow(schema));
+    handler._objectCore.set([SYSTEM_NAMESPACE, 'type'], encodedRef);
+  }
+};
+
+export const getSchemaTypeRefOrThrow = (schema: S.Schema<any>): Reference => {
+  const typeReference = getTypeReference(schema);
+  if (typeReference == null) {
+    throw new Error(
+      'EchoObject schema must have a valid annotation: MyTypeSchema.pipe(R.echoObject("MyType", "1.0.0")',
```
dmaretskyi, line 548:
```suggestion
      'EchoObject schema must have a valid annotation: MyTypeSchema.pipe(R.echoObject("MyType", "1.0.0"))',
```

packages/core/echo/echo-schema/src/effect/schema-validator.ts
```diff
@@ -50,7 +50,7 @@ export class SchemaValidator {
     return value;
   }
 
-  private static getPropertySchema(target: any, prop: string | symbol): S.Schema<any> {
+  private static getTargetPropertySchema(target: any, prop: string | symbol): S.Schema<any> {
```
dmaretskyi, line 53:
```suggestion
  private static _getTargetPropertySchema(target: any, prop: string | symbol): S.Schema<any> {
```

packages/core/echo/echo-schema/src/type-collection.ts
```diff
@@ -42,6 +48,28 @@ export class TypeCollection {
...
+    });
+  }
+
+  isEffectSchemaRegistered(schema: S.Schema<any>): boolean {
+    const typename = getTypenameOrThrow(schema);
+    return this._effectSchemaDefs.has(typename);
+  }
+
+  getEffectSchema(typename: string): S.Schema<any> | undefined {
+    if (typename === LEGACY_TEXT_TYPE) {
```
dmaretskyi, line 67: just add to map

packages/core/echo/echo-schema/src/effect/echo-proxy.test.ts
```diff
@@ -113,8 +129,49 @@ describe('Reactive Object with ECHO database', () => {
       expect(obj.id).to.eq(id);
       expect(obj.string).to.eq('foo');
 
-      // TODO(dmaretskyi): Schema lookup?
-      // expect(R.getSchema(obj)).to.eq(TestSchema);
+      expect(R.getSchema(obj)).to.eq(EchoObjectSchema);
+    }
+  });
+
+  test('effect-protobuf schema interop', async () => {
```
dmaretskyi, line 136: 👍

PR #5855, opened 2024-03-05
fix(client-services): always pass through whole profile

wittjosiah:
Followup to #5811 which fixes a couple small issues:
- always pass through whole profile from the space service
- don't hardcode emoji/hue in profile proto


dmaretskyi (approved):

thure (approved):
lgtm 👌 ✨

PR #5857, opened 2024-03-05
feat: Migrate to E.object API

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/apps/plugins/plugin-markdown/src/types.ts
```diff
@@ -69,3 +71,17 @@ export type MarkdownPluginProvides = SurfaceProvides &
...
+  title: S.optional(S.string),
+  content: S.optional(S.string),
+  comments: S.optional(
+    S.struct({
+      // TODO(wittjosiah): Add thread schema.
+      // thread: S.optional(ThreadSchema),
+      cursor: S.optional(S.string),
+    }),
+  ),
+}).pipe(E.echoObject('dxos.sdk.client.Properties', '0.1.0'));
```
dmaretskyi, line 85: as long as the typename exactly matches the ones generated from protobuf, the old data should be usable as well

dmaretskyi: we can think about migrating those names later

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -105,15 +105,20 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
 
     this._signal.notifyRead();
 
+    const isRootObject = target[symbolPath].length === 0;
+
     if (typeof prop === 'symbol') {
+      if (isRootObject && prop === data) {
```
dmaretskyi, line 111: also

```ts
if(isRootObject && prop === 'toJSON') {
  return () => this._toJSON(target)
  }
  ```

File: packages/apps/plugins/plugin-stack/src/components/StackMain.tsx [hunk 3]
```diff
@@ -48,11 +48,11 @@ const StackMain: FC<{ stack: StackType; separation?: boolean }> = ({ stack, sepa
     // TODO(wittjosiah): Render placeholders for missing objects so they can be removed from the stack?
     .filter(({ object }) => Boolean(object))
     .map(({ id, object }) => {
-      const rest = metadataPlugin?.provides.metadata.resolver(object.__typename ?? 'never');
+      const rest = metadataPlugin?.provides.metadata.resolver(E.typeOf(object)?.itemId ?? 'never');
```
dmaretskyi, line 51: `E.schemaOf(object).typename`

PR #5865, opened 2024-03-06
feat: Typescript service for script plugin

richburdon:
- Provides dynamic TS environment in the browser for the script editor.
- Dynamically import typedefs.
- Update to Typescript 5.3.3

https://github.com/dxos/dxos/assets/3523355/a4303d2c-6853-49fb-a07f-c7e4ca84caac


packages/apps/plugins/plugin-script/src/components/ScriptEditor/ScriptEditor.tsx
```diff
@@ -2,56 +2,137 @@
...
-  const { parentRef } = useTextEditor(
+  const { parentRef, view } = useTextEditor(
     {
       extensions,
       doc: DocAccessor.getValue(source),
     },
     [extensions],
   );
 
+  useEffect(() => {
```
dmaretskyi, line 133: wouldn't you wanna run this on every edit (with thottling) instead of only on load?

dmaretskyi (approved):

packages/apps/plugins/plugin-script/src/components/ScriptEditor/ScriptEditor.tsx
richburdon: the custom extension listener above does this. this effect is just for on start.

Issue #5852, opened 2024-03-05
[observability] attach breadcrumbs to errors from preceding warnings


wittjosiah:
- attach warnings that occur before an error or any other useful breadcrumbs for tracking down the cause of errors in sentry

Issue #5851, opened 2024-03-05
[observability] replace the capture console sentry integration w/ a log processor


wittjosiah:
- capture console can be noisy
- console log formatting is not the same formatting that we want for sentry necessarily 
- a sentry log processor already exists in the code but needs to be re-evaluated and then hooked up while the capture console integrations need to be removed
- once this is done we need to verify that we're getting cleaner/more useful errors from within the platform code in sentry

wittjosiah:
one other thing to consider with this one is to ensure we're not double reporting logs from the worker because of this logging service used to surface worker error logs on the main thread console https://github.com/dxos/dxos/blob/main/packages/sdk/client-services/src/packlets/logging/logging-service.ts#L29

Issue #5850, opened 2024-03-05
[observability] filter out noisy sentry breadcrumbs


wittjosiah:
- of particular issue is the segment api requests
- consider filtering out anything else that is being automatically tracked that isn't useful

Source: https://composer.dxos.org/844c136e2345eb291e05100b311dfcacbc9e74301f933229fe3b733491bff96e

PR #5872, opened 2024-03-07
feat(sentry): crash reporting improvements

yarolegovich:
### Motivation / Background

[Original doc](https://composer.dxos.org/844c136e2345eb291e05100b311dfcacbc9e74301f933229fe3b733491bff96e).
This PR addresses #5850 #5851 #5852

### Details

* Disabled fetch integration, because we only have segment API call which isn't useful as a breadcrumb.
* Pass error as `LogEntry.error` when `log.catch` is called.
* Pass meta to browser log processors for log entries received from a shared worker.
* Get rid of CSS formatting in Sentry log messages by introducing a custom SentryLogProcessor and disabling console instrumentation.
* Search for an `instanceof Error` object in `LogEntry.context` for errors to pass it to `Sentry.captureException` method.
* For issues reported from shared worker set `service_host_issue: true` and `service_host_session: randomId` tags passed as `LogEntry.meta` set in `LoggingService`, as well as `uptime_seconds` extra field.
* Collect log entries as breadcrumbs and attach them to events before sending. Currently these are attached to every event, not sure if we should be doing this warnings, but let's how useful & spammy it gets.
* Format sentry message as `([worker] )?${scope} ${message}`, with only message being used for fingerprinting. Ensures scope instance ids don't influence grouping. For exceptions only stacktrace is used for fingerprinting (log.catch or log.error with error in context). 
* Added some simple console breadcrumb merging by message. Multiple sequential breadcrumbs with the same message get merged into a single breadcrumb with `data: {count, firstBreadcrumbData, lastBreadcrumbData}`. [An example](https://dxos.sentry.io/issues/5042444138/?environment=development&project=4504340179779584&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=10) where log gets polluted by such message.

### Examples

* [A warning from shared worker](https://dxos.sentry.io/issues/5044624558/?environment=development&project=4504340179779584&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0)
* [An exception from tab](https://dxos.sentry.io/issues/5044624666/?environment=development&project=4504340179779584&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=2). Here message extra is the same as stacktrace, but I fixed that in my last commit.
* [An exception from shared worker](https://dxos.sentry.io/issues/5044757085/?environment=development&project=4504340179779584&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0).


packages/sdk/client/src/services/worker-client-services.ts
```diff
@@ -106,16 +106,16 @@ export class WorkerClientServices implements ClientServicesProvider {
     this._loggingStream.subscribe((entry) => {
       switch (entry.level) {
         case LogLevel.DEBUG:
-          log.debug(`[worker] ${entry.message}`, entry.context);
+          log.debug(`[worker] ${entry.message}`, entry.context, mapLogMeta(entry.meta));
```
dmaretskyi, line 109: IDEA: add `processId` to the log entry to put `[worker]` in there

PR #5877, opened 2024-03-07
chore: Change text editor components to hooks and remove EditorModel

richburdon:
The next installment. Last stop before removing TextEditor.

## Next
- Remove yjs deps, remove TextModel, TextObject.
- Fix surface/model deps (e.g., local files?)
- move DocAccessor deps into new package dxos/types.


packages/ui/react-ui-editor/src/extensions/automerge/defs.ts
```diff
@@ -40,7 +40,7 @@ export interface DocAccessor<T = any> {
 
 // TODO(burdon): Reconcile with echo.
 export const DocAccessor = {
-  getValue: (accessor: DocAccessor) => get(accessor.handle, accessor.path),
+  getValue: (accessor: DocAccessor): string => get(accessor.handle, accessor.path),
```
dmaretskyi, line 43: It's not always a string. This would return any value stored at the given path. I'd type this as `never` and then assert that this is a string at the place where this is called.

```suggestion
  getValue: (accessor: DocAccessor): never => get(accessor.handle, accessor.path),
```

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -39,61 +29,43 @@ export type CursorInfo = {
...
-  };
-  content?: {
-    className?: string;
-  };
-};
-
 export type TextEditorProps = {
-  model: EditorModel; // TODO(burdon): Optional (e.g., just provide content if readonly).
-  readonly?: boolean; // TODO(burdon): Move into model.
+  dataTestId?: string; // TODO(burdon): Generalize?
```
wittjosiah, line 33: doesn't seem like this is used currently? what about having these props extend `div` props and forward any unknown props into the parent div? could omit `className` below then as well

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -39,61 +29,43 @@ export type CursorInfo = {
...
+  doc?: string; // TODO(burdon): Rename text/value/content?
```
wittjosiah, line 34: maybe `value`

packages/apps/plugins/plugin-thread/src/components/MessageContainer.tsx
```diff
@@ -68,27 +69,26 @@ const TextboxBlock = ({
...
   const identity = useIdentity();
-  const model = useTextModel({ text });
-  const extensions = useMemo(() => [command], []);
-  const textboxWidth = onBlockDelete ? 'col-span-2' : 'col-span-3';
   const readonly = identity?.identityKey.toHex() !== authorId;
+  const textboxWidth = onBlockDelete ? 'col-span-2' : 'col-span-3';
+  const { parentRef } = useTextEditor({
+    doc: getTextContent(text),
+    extensions: [
+      //
```
wittjosiah, line 79: ?

packages/apps/plugins/plugin-thread/src/components/MessageContainer.tsx
```diff
@@ -68,27 +69,26 @@ const TextboxBlock = ({
...
+  const { themeMode } = useThemeContext();
   const identity = useIdentity();
-  const model = useTextModel({ text });
-  const extensions = useMemo(() => [command], []);
-  const textboxWidth = onBlockDelete ? 'col-span-2' : 'col-span-3';
   const readonly = identity?.identityKey.toHex() !== authorId;
+  const textboxWidth = onBlockDelete ? 'col-span-2' : 'col-span-3';
+  const { parentRef } = useTextEditor({
+    doc: getTextContent(text),
+    extensions: [
```
wittjosiah, line 78: extensions need to be memoized

packages/apps/plugins/plugin-pwa/src/PwaPlugin.tsx
```diff
@@ -26,7 +26,7 @@ export const PwaPlugin = (): PluginDefinition => ({
               id: `${PWA_PLUGIN}/need-refresh`,
               title: translations[0]['en-US'][PWA_PLUGIN]['need refresh label'],
               description: translations[0]['en-US'][PWA_PLUGIN]['need refresh description'],
-              duration: 240e3, // 4 minutes
+              duration: 4 * 60 * 1000, // 4m
```
wittjosiah, line 29: not sure how this is clearer?

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -2,48 +2,68 @@
...
+  comments?: Comment[];
+} & Pick<TextEditorProps, 'doc' | 'extensions'>;
 
-export const EditorMain = ({ model, comments, toolbar, extensions: _extensions, ...props }: EditorMainProps) => {
+// TODO(burdon): Invalid cursor error when switching between editors.
+export const EditorMain = ({ id, readonly, toolbar, comments, doc, extensions: _extensions }: EditorMainProps) => {
   const { t } = useTranslation(MARKDOWN_PLUGIN);
+  const { themeMode } = useThemeContext();
+
+  // TODO(burdon): Remove useEditorView. Check interaction with useComments with @wittjosiah.
```
wittjosiah, line 58: `viewInvalidated` is how we resolved the problem of attempting to render comment highlights from a new doc onto the previous doc. The ref doesn't change until after render has run but we can tie the ref to the id coming in such that we immediately know that a view is invalid and stop using it.

packages/apps/plugins/plugin-markdown/src/components/DocumentSection.tsx
```diff
@@ -20,25 +28,32 @@ const DocumentSection: FC<{
...
-      slots={{
-        root: {
-          className: mx('flex flex-col grow m-0.5 min-bs-[8rem]', attentionSurface, focusRing),
-          'data-testid': 'composer.markdownRoot',
-        } as HTMLAttributes<HTMLDivElement>,
-        editor: {
-          className: 'h-full py-4',
+  const { themeMode } = useThemeContext();
+  const { doc, accessor } = useDocAccessor(document.content);
+  const { parentRef } = useTextEditor({
```
wittjosiah, line 34: usage of hook here vs `TextEditor` component? is the assumption that tabster and other stuff in `TextEditor` is only needed for the main editor?

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -39,61 +29,43 @@ export type CursorInfo = {
...
-      theme,
-      slots = defaultSlots,
       extensions = [],
-      scrollPastEnd: _scrollPastEnd,
       debug,
     },
     forwardedRef,
   ) => {
+    // TODO(burdon): Make tabster optional.
     const tabsterDOMAttribute = useFocusableGroup({ tabBehavior: 'limited' });
```
wittjosiah, line 63: tabster could be moved inside `useTextEditor` and it could return dom attributes to provide to the parent alongside the ref

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
```diff
@@ -39,61 +29,43 @@ export type CursorInfo = {
...
   extensions?: Extension[];
   debug?: boolean;
 };
 
 /**
- * Base text editor.
+ * Thin wrapper for text editor.
+ * Handles tabster and focus management.
  */
-// TODO(burdon): Replace with useTextEditor.
```
wittjosiah, line 71: not expecting to replace this anymore? is there something here that necessarily differentiates it from `useTextEditor`?

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -242,17 +240,19 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
                 isMarkdownProperties(data.properties)
               ) {
-                const main = <EditorMain model={data.model} extensions={extensions} />;
-                if ('view' in data && data.view === 'embedded') {
-                  return <EmbeddedLayout>{main}</EmbeddedLayout>;
-                } else {
-                  return <MainLayout>{main}</MainLayout>;
-                }
+                return null;
+                // const main = <EditorMain extensions={extensions} />;
```
wittjosiah, line 250: this is why we need stories for the editor w/o automerge, so that we can ensure that works in non-automerge scenarios in composer (e.g. local files)

packages/apps/plugins/plugin-kanban/src/components/KanbanCard.tsx
```diff
@@ -30,13 +36,24 @@ export const KanbanCardComponent: FC<{
...
     id: item.id,
     data: { type: 'item', column },
   });
   const tx = transform ? Object.assign(transform, { scaleY: 1 }) : null;
-  const model = useTextModel({ text: item.title });
+
+  const { doc, accessor } = useDocAccessor(item.title);
+  const extensions = useMemo(
+    () => [
+      //
```
wittjosiah, line 50: ?

packages/apps/plugins/plugin-kanban/src/components/KanbanCard.tsx
```diff
@@ -50,15 +67,7 @@ export const KanbanCardComponent: FC<{
...
-          {model && (
-            <MarkdownEditor
-              model={model}
-              placeholder={t('item title placeholder')}
-              slots={{
-                root: { className: mx(focusRing, 'p-1') },
-              }}
-            />
-          )}
+          <TextEditor doc={doc} extensions={extensions} className={mx(focusRing, 'p-1')} />
```
wittjosiah, line 70: why does kanban card use `TextEditor` vs editor card uses the `useTextEditor` hook?

wittjosiah (commented):
One other note is that I checked this out and ran it and it seemed to roughly be working, which is a good sign, but most of the e2e tests that touch the editor or comments were failing. This may be related to some sets of extensions not being memoized or something else. We should address the issues causing those failures before merging this in case they are indicative of larger problems.

packages/ui/react-ui-editor/src/extensions/cursor-line-margin.ts
```diff
@@ -5,34 +5,33 @@
...
 
-// Seems to the best approximation of CM5's `cursorScrollMargin`
-// https://discuss.codemirror.net/t/cursorscrollmargin-for-v6/7448
-export const cursorLineMargin: Extension = EditorView.updateListener.of(
-  ({ transactions, state, selectionSet, startState, view }) => {
+type Options = {
+  lines?: number;
+};
+
+// TODO(burdon): Remove. Buggy and creates scrolling problems. Find better (more general) way to deal with HALO notch.
```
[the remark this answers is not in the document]
richburdon: gets stuck or just disappears?

to me the app felt really broken. clicking would scroll the window. worse, cursoring up forces the window to scroll. i think it's rare that the notch will cover the text (small screen, no sidebar), and it's uneven to have this only work for the text editor.

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.stories.tsx
```diff
@@ -4,45 +4,282 @@
...
+  const { accessor } = useDocAccessor(new TextObject(text));
+
+  const viewRef = useRef<EditorView>(null);
+  useComments(viewRef.current, id, comments);
+
+  const { themeMode } = useThemeContext();
+  const extensions = useMemo(
+    () => [
+      createBasicExtensions({ readonly, placeholder }),
+      createMarkdownExtensions({ themeMode }),
```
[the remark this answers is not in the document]
richburdon: Added two basic stories to the hook storybook.

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -508,7 +508,8 @@ export const focusComment = (view: EditorView, id: string, center = true) => {
 /**
  * Update comments state field.
  */
-export const useComments = (view: EditorView | null, id: string, comments: Comment[] = []) => {
+// TODO(burdon): Replace with extension that has callback or deepsignal.
+export const useComments = (view: EditorView | null | undefined, id: string, comments: Comment[] = []) => {
```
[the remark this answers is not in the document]
richburdon: I get your point, but the editor package does have a comments button in the toolbar, and these are comment-specific decorations, say vs: https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-editor/src/extensions/annotations.ts (which is an experimental decoration for LLM detected objects).

packages/ui/react-ui-editor/src/extensions/automerge/automerge.stories.tsx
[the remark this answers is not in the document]
richburdon: yes i see an error also:
`chunk-6NX3RPDS.mjs:2021 ControlPipeline#5 Error: invariant violation: Feed already added`

@dmaretskyi does this look familiar?

packages/apps/plugins/plugin-chain/src/components/PromptTemplate/PromptTemplate.tsx
```diff
@@ -98,37 +104,30 @@ type PromptTemplateProps = {
...
+      }),
+      createThemeExtensions({
+        themeMode,
+        slots: {
+          content: { className: '!p-3' },
+        },
+      }),
+      promptExtension,
+    ],
+    [themeMode, accessor],
```
thure, line 125: Should calls to `create…Extensions` always be memoized? To encourage memoization, should we provide a `useExtensions` hook that can call known create functions?

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
richburdon: Added TODO. i do see this as important once we've landed the AM branches.

packages/apps/plugins/plugin-kanban/src/components/KanbanCard.tsx
richburdon: removed. but my way of defeating prettier.

packages/apps/plugins/plugin-kanban/src/components/KanbanCard.tsx
richburdon: Changed.

packages/apps/plugins/plugin-chain/src/components/PromptTemplate/PromptTemplate.tsx
richburdon: Good catch. I'm considering making the useTextEditor take a callback fn.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -2,69 +2,94 @@
...
 
-export const EditorMain = ({ model, comments, toolbar, extensions: _extensions, ...props }: EditorMainProps) => {
+export const EditorMain = ({ id, readonly, toolbar, comments, doc, extensions: _extensions }: EditorMainProps) => {
   const { t } = useTranslation(MARKDOWN_PLUGIN);
+  const { themeMode } = useThemeContext();
 
-  const [editorRef, viewInvalidated] = useEditorView(model.id);
-  useComments(viewInvalidated ? null : editorRef.current, model.id, comments);
-  const handleAction = useActionHandler(editorRef.current);
+  const { ref: editorRef, value: view } = useRefCallback<EditorView>();
```
thure, line 53: If this is the best or only way to get `view`, I think it’d be easier to proliferate this pattern if `react-ui-editor` exported its own `useEditorRef` hook.

packages/apps/plugins/plugin-pwa/src/PwaPlugin.tsx
thure: (personally [I’ve often used this syntax](https://gitlab.com/accord.coop/accord/-/blob/staging/api/lib/accord_slack/views/new_decision_step_2.ex#L91-126), though it’s more valuable when multiple durations are defined together.

packages/ui/react-ui-editor/src/components/TextEditor/TextEditor.tsx
thure: +1

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.stories.tsx
```diff
@@ -4,61 +4,73 @@
...
+            setComments((comments) => [...comments, { id, cursor }]);
+            return id;
+          },
+        }),
+        decorateMarkdown(),
+        formattingKeymap(),
+        image(),
+        table(),
+        formattingObserver,
+      ],
```
thure, line 65: Memoize extensions? It’s just a story, but folks will copypaste from stories, so probably worth setting an example for best practices.

packages/ui/react-ui-editor/src/extensions/cursor-line-margin.ts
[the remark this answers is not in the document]
thure: > clicking would scroll the window
If by ‘window’ you mean CM’s overflow parent, that is a good thing if moving the cursor to the beginning of the line would otherwise get it occluded by the notch

> cursoring up forces the window to scroll
This is definitely a bug that should be addressed, but IMO not an indication that the solution has no value. At least the cursor remains visible.

> it's uneven to have this only work for the text editor
I’m not sure yet, but because this uses `scrollIntoView` I think this can work, or already does work, for text editors in stacks and threads.

> i think it's rare that the notch will cover the text (small screen, no sidebar)
This is a fair point, perhaps this behavior only activates when the notch might possibly occlude the cursor, otherwise it is disabled.

packages/ui/react-ui-editor/src/themes/default.ts
```diff
@@ -262,29 +262,29 @@ export const defaultTheme: ThemeStyles = {
...
+// };
+//
+// export const codeTheme: ThemeStyles = {
+//   '.cm-scroller': {
+//     fontFamily: get(tokens, 'fontFamily.mono', []).join(','),
+//   },
+//   '.cm-placeholder': {
+//     fontFamily: get(tokens, 'fontFamily.mono', []).join(','),
+//   },
+// };
```
thure, line 290: I think it’s probably best to just remove these.

packages/ui/react-ui-thread/src/Message/Message.tsx
```diff
@@ -41,14 +41,14 @@ export const MessageMeta = forwardRef<HTMLDivElement, MessageMetaProps>(
           className={mx('grid grid-cols-subgrid col-span-2', classNames)}
           ref={forwardedRef}
         >
-          <div role='none' className={mx('flex flex-col items-center gap-2', messageCell)}>
+          <div role='none' className={mx('flex flex-col items-center gap-2 plb-2 min-is-0')}>
```
thure, line 44:
```suggestion
          <div role='none' className={mx('flex flex-col items-center gap-2 plb-2')}>
```
`min-is-0` should not be needed here, the `Avatar.Frame` already has an extrinsic size and should not need to shrink.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
richburdon: i debated that but given that it's only used once and i want to try to remove the component based approach altogether decided against. for now.

thure (approved):
Definitely looking forward to this ✨.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -84,33 +109,22 @@ export const EditorMain = ({ model, comments, toolbar, extensions: _extensions,
         data-toolbar={toolbar ? 'enabled' : 'disabled'}
         className='is-full bs-full overflow-hidden data-[toolbar=disabled]:pbs-2'
       >
-        <MarkdownEditor
-          ref={editorRef}
+        <TextEditor
```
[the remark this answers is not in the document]
richburdon: ok. i may try to remove in the next iteration. this was a gnarly PR, so getting it down now to just one component and the hooks used in most places was a win.

packages/ui/react-ui-thread/src/Message/Message.tsx
richburdon: ok thanks. note i bumped the padding on the icon so that the text more closely centers with it.

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
thure: Makes sense 👌

packages/apps/plugins/plugin-markdown/src/components/DocumentSection.tsx
```diff
@@ -20,25 +28,32 @@ const DocumentSection: FC<{
...
+    doc,
+    extensions: [
```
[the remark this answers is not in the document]
richburdon: Changed useTextEditor to take a callback like useMemo.

PR #5873, opened 2024-03-07
chore: enable no-console eslint rule

yarolegovich:
### Motivation / Background

Related to #5872
Using a custom Sentry log processor instead of console instrumentation means we need to ensure all messages go through `dxos/log` interface so that problems don't go unnoticed. + unification is good

### Details

* Set `no-console: error` as the project default excluding test and story files.
* Set `no-console: off` override for some of the projects, mostly under tools.
* Fix existing console usages to make the rule pass.

PR #5881, opened 2024-03-07
fix(tables): 💪 Restore table virtualisation

ZaymonFC:
At some point we stopped supplying props to the `<TableImpl>` component. This caused virtualisation to never be used since `getScrollElement`—a function that provides a ref to the parent container—is always `undefined`.

I've also restored a few other props that were being swallowed and look like they shouldn't have been.


packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -181,6 +184,10 @@ const TableImpl = <TData extends RowData>(props: TableProps<TData>) => {
     );
   }
 
+  if (process.env.NODE_ENV === 'development' && !getScrollElement) {
```
wittjosiah, line 187: I think we've found that this wasn't necessarily set in vite bundled apps

packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -181,6 +184,10 @@ const TableImpl = <TData extends RowData>(props: TableProps<TData>) => {
...
+    console.warn('Table: getScrollElement is not set. This is required for virtualized tables.');
```
wittjosiah, line 188: can you use @dxos/log package? there's a pending PR which would make this a lint error https://github.com/dxos/dxos/pull/5873

ZaymonFC: Sure thing! I was wondering if there were a better way.

PR #5883, opened 2024-03-07
feat: new composer plugin template app

zhenyasav:
this ensures new plugin template is configured for public publishing


packages/sdk/examples/src/template/README.md
wittjosiah, line 1: this probably isn't needed? (also the license file above in the same dir)

wittjosiah (approved):

Issue #5376, opened 2024-01-24
[sdk] platform telemetry instrumentation


zhenyasav:
see spec here https://github.com/dxos/product/blob/main/2024-Q1/specs/platform-telemetry.md

Need signalling protocol update to include keys in messages

PR #5892, opened 2024-03-08
feat: add telemetry metadata to signaling

nferch:
Naive client implementation for https://github.com/dxos/dxos/issues/5376


packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -19,26 +20,39 @@ import { trace } from '@dxos/tracing';
...
+  signalTelemetryEnabled?: boolean,
 ): Promise<ClientServicesProvider> => {
-  return new LocalClientServices({
+  const getSignalMetadata = () => {
+    return {
+      ...lcs.signalMetadataTags,
+      ...(observabilityGroup ? { group: observabilityGroup } : {}),
+    };
+  };
+  const lcs = new LocalClientServices({
```
wittjosiah, line 32:
```suggestion
  const services = new LocalClientServices({
```
uncommon abbreviations invite too many questions 😅

packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -73,12 +87,28 @@ export class LocalClientServices implements ClientServicesProvider {
...
   constructor(params: ClientServicesHostParams) {
     this._params = params;
+    // TODO(nf): extract
+    if (typeof window === 'undefined') {
+      // TODO(nf): collect ClientServices metadata as param?
+      this.signalMetadataTags.origin = 'undefined';
+    } else {
+      // SocketSupply native app
+      if ((globalThis as any).__args) {
+        this.signalMetadataTags.origin = 'native';
```
wittjosiah, line 106: fwiw I think window.location.origin will still work here and will give an indication of _which_ app it is

packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -112,6 +142,17 @@ export class LocalClientServices implements ClientServicesProvider {
     });
     await this._host.open(this._ctx);
     this._isOpen = true;
+    // don't set identity tags for non-browser usage, e.g. agents
+    // TODO(nf): fix
```
wittjosiah, line 146: what needs to be fixed?

packages/sdk/client/src/services/client-services-factory.tsx
```diff
@@ -41,6 +41,8 @@ export const Remote = (target: string | undefined): Partial<ConfigProto> => {
 export const createClientServices = (
   config: Config,
   createWorker?: WorkerClientServicesParams['createWorker'],
+  observabilityGroup?: string,
+  signalTelemetryEnabled?: boolean,
```
wittjosiah, line 45: maybe worth a jsdoc somewhere describing what this does

packages/sdk/client/src/services/worker-client-services.ts
```diff
@@ -47,11 +49,21 @@ export class WorkerClientServices implements ClientServicesProvider {
...
+    createWorker,
+    logFilter = 'error,warn',
+    observabilityGroup,
+    signalTelemetryEnabled,
+  }: WorkerClientServicesParams) {
     this._config = config;
     this._createWorker = createWorker;
     this._logFilter = parseFilter(logFilter);
+    this._observabilityGroup = observabilityGroup;
+    this._signalTelemetryEnabled = signalTelemetryEnabled ?? false;
```
wittjosiah, line 66: having the default for this be false everywhere seems like it defeats the goal of introducing it?

that being said, if we want the default to be true, should probably consider how to make it clear to developers that its happening so that they can choose to opt out. cc @jessmartin @zhenyasav

packages/sdk/client/src/services/local-client-services.ts
nferch: ok, i still can't get the app to run so I'll just have to take your word for it :)

packages/sdk/client/src/services/local-client-services.ts
nferch: now I remember...

I wasn't sure what the behavior should be when the app isn't known, which is currently the agent, but might be other usages in the future.

I don't think including these sources and not including the origin field will significantly affect the usefulness of the data, but maybe @zhenyasav / @jessmartin have other ideas.

packages/sdk/client/src/services/worker-client-services.ts
nferch: I intended to use the same observability opt-out decision from the app, but I forgot to reverse the condition.

So the default should be on unless the user opts out.

PR #5900, opened 2024-03-09
enhancement(tables) - Quality of life improvements

ZaymonFC:
### 🐛 Bug Fixes:
- Ensure all dependencies are correctly specified in useEffect and useCallback hooks.
- Fixed inconsistency in handleEdit function behavior when onUpdate is undefined.
- Addressed a visual bug to maintain the focus ring visibility on focused inputs.

### 💅 UX Improvements:
- Implemented focus restoration for string input cells on Enter key press.

### ✨ Performance Enhancements:
- Significantly improved column resizing performance by switching to a memoized table during resizing.
  - https://tanstack.com/table/latest/docs/framework/react/examples/column-resizing-performant


thure (approved):
lgtm! ✨

packages/ui/primitives/react-hooks/src/useDefaultValue.ts
```diff
@@ -0,0 +1,28 @@
+//
```
richburdon, line 1: could you take a look at common/react-async which has a bunch of similar hooks (e.g., useStateRef). "async" isn't a good place for these since they are general purpose, so perhaps they need to be transplanted here. Would be good for you to review these.

ZaymonFC: Great idea.

PR #5909, opened 2024-03-11
fix(echo): removed unnecessary cancelWithContext

yarolegovich:
### Motivation / Background

* Remove unnecessary `cancelWithContext` that gets lots of listeners assigned (on every update we add a listener. No point in cancelling on context dispose as we're not performing operations that can hang indefinitely. And IO will anyway complete once it's started.
image

* Refactored `markDirty` into a batch API. Easier to change the internal implementation when we need to optimize io + more convenient usages on call sites.


packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
```diff
@@ -178,28 +178,18 @@ export class AutomergeHost {
       return;
     }
 
-    const markingDirtyPromise = Promise.all(
-      objectIds.map(async (objectId) => {
-        const spaceKey = getSpaceKeyFromDoc(event.doc);
-        if (!spaceKey) {
-          return;
-        }
-        await cancelWithContext(
```
mykola-vrmchk, line 187: why do you think this is unnecessary?

Issue #4860, opened 2023-12-05
[stack] resize stack section affordance


zhenyasav:
this helps images in stacks

jessmartin:
would personally prefer to be able to full-height a stack section rather than arbitrarily resize

zhenyasav:
collections hath landed!

PR #5917, opened 2024-03-12
feat(react-ui-stack,plugin-stack): Stack section design increment

### LINKED ISSUE #4865: [stack] insert content between sections
# Motivation

The only way to insert new content between sections is to append and drag, it's impossible to directly insert something between sections.

# Finish line

there is an affordance which can be used to add sections between stack sections (see Figma "Composer Explorations" as a input to the design process)

### LINKED ISSUE #5733: [stack] section chrome iteration
## Questions

- are we happy with using the whole left gutter of the section just for a centered drag handle?
- how could we support collapsing a section?
- should the section have a canonical way to expose a toolbar?
  - sections need to be able to expose a toolbar to action the content of the section
    - editor has a toolbar
    - table will likely have a toolbar
    - sketch has tools as well
  - toolbar could be rendered within the content area of the section
  - having a place for toolbars in the section however will likely make it feel more consistent
- how do the toolbar and section actions related?

thure:
Resolves #5733.

This PR will:
- [x] Combine stack section dropdown menu triggers and drag handles into one button, à la Notion
- [x] Implement expand/collapse on stack sections

https://github.com/dxos/dxos/assets/855039/50d5cb06-eca1-4f08-b0d8-7ea1dd62ba78

The next PR will:
- [ ] Add actions to StackPlugins’ sections:
  - [ ] Add item above
  - [ ] Add item below

These PRs stop short of:
- #4860 Implementing resize handles


packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
```diff
@@ -48,10 +48,16 @@ const StackMain: FC<{ stack: StackType; separation?: boolean }> = ({ stack, sepa
   const items = stack.sections
     // TODO(wittjosiah): Should the database handle this differently?
     // TODO(wittjosiah): Render placeholders for missing objects so they can be removed from the stack?
-    .filter(({ object }) => Boolean(object));
+    .filter(({ object }) => Boolean(object))
+    .map(({ id, object }) => {
+      const rest = metadataPlugin?.provides.metadata.resolver(object.__typename ?? 'never');
```
wittjosiah, line 53: is this to get the placeholder for the collapsed state?

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -2,48 +2,48 @@
...
 export type StackSectionContent = MosaicDataItem & { title?: string };
 
+export type CollapsedSections = Record<string, boolean>;
+
 export type StackContextValue<TData extends StackSectionContent = StackSectionContent> = {
   SectionContent: FC<{ data: TData }>;
   transform?: (item: MosaicDataItem, type?: string) => StackSectionItem;
   onDeleteSection?: (path: string) => void;
   onNavigateToSection?: (id: string) => void;
+  collapsedSections?: CollapsedSections;
```
wittjosiah, line 45: can we leave a todo to merge this into the sections prop?

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -2,48 +2,48 @@
...
+  setCollapsedSections?: Dispatch<SetStateAction<CollapsedSections>>;
```
wittjosiah, line 46: `onCollapseSection?: (id: string) => void`?

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -53,128 +53,167 @@ export type StackItem = MosaicDataItem &
...
+
+            <ListItem.Heading
+              classNames={
+                collapsed
+                  ? ['grid grid-rows-subgrid grid-cols-subgrid items-center rounded-sm mlb-1 mie-1', focusRing]
+                  : 'sr-only'
+              }
+              {...(collapsed && { ...sectionContentGroup, tabIndex: 0 })}
+            >
+              {/* TODO(thure): This needs to be made extensible; Markdown document titles especially are difficult */}
```
wittjosiah, line 186: what do you think about just using a separate `Surface` here? Rather than trying to recompute what the title should be just defer to a plugin that knows how to render it. Stack could provide this title span as a fallback?

wittjosiah (commented):
Sorry for the delay!

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
thure: Yes, this was an attempt to get the fallback title, but the placeholder still seemed like progress.

packages/ui/react-ui-stack/src/components/Section.tsx
thure: I want to make a distinction between two patterns:

- `{prop, onChangeProp, defaultProp}` when controllable
- `{prop, setProp}` when drilling `useState` down

For me, the name `setCollapsedSections` signals that this will directly affect a state, as much as I’d like to use something that sounds more natural like you suggest.

Do you think we should make exceptions for contexts, or is your suggestion applying a principle we can apply more broadly?

packages/ui/react-ui-stack/src/components/Section.tsx
thure: I was wondering if this might be the right approach, I can give it a try!

packages/ui/react-ui-stack/src/components/Section.tsx
wittjosiah: I'm curious why you want to make that distinction? My sense would be that the component shouldn't care what the callback is being mapped back to.

All that aside, I think it mapping directly back to `useState` is going to be a temporary state based on this TODO https://github.com/dxos/dxos/pull/5917#discussion_r1530526158. Which another reason why letting how it's being used influence the api seems like a bad idea to me.

packages/ui/react-ui-stack/src/components/Section.tsx
wittjosiah: another option might be to just find the associated graph node and use it's label...both seems to have tradeoffs to me.

if you use the surface you may end up with some duplicate logic within plugins, but it nicely decouples things and also would allow you to render more than just text as a heading potentially

if you use the graph, you're able to just use existing labels, but it couples stack to knowing about the graph and means that all sections need to be in the graph (however, stack knowing about the graph doesn't seem like the end of the world as far as coupling goes and sections will likely always be in the graph anyways)

packages/ui/react-ui-stack/src/components/Section.tsx
thure: I see, so in other words the principle to apply is “name it something intentionally different because its implementation should not make any guarantees”? I think I’m amenable to that.

thure: It would be similar to `handle…` vs `on…`, where the former is only used locally. Essentially we argue `set…` should be local-only as well.

packages/ui/react-ui-stack/src/components/Section.tsx
wittjosiah: fwiw everything that gets rendered in the main area is the datum of a graph node, its just that the graph node isn't what is passed to the surface...

wittjosiah: I'd say just do what makes most sense for now. We will revisit soon because after the schema work lands stacks & folders are going to devolve into collections.

thure: Ah, I remember why I didn’t use `Surface`, it’s part of `app-framework` which no other UI package uses. This would set a new precedent, but I think it doesn’t need to and shouldn’t.

packages/ui/react-ui-stack/src/components/Section.tsx
thure: Ah, however, I notice your suggestion reframes this as a singular. Because the implementation of this state should change, I intentionally didn’t try to split the collective state setter into single state setters for each section, to avoid complexity we’re soon to remove.

For now I think it’s best to just leave a TODO here.

wittjosiah: `onCollapseSection?: (sections: CollapsedSections) => void` with a todo explaining why it's passing all of them then?

thure: I suppose either leaving the code alone and adding a comment about it, or making changes to the code in addition to adding a comment about it, are roughly equivalent approaches to me. I’m also concerned about minting `onCollapseSection` when that might come off as a design assertion when we might refactor this to use a pattern e.g. signals which doesn’t have those callbacks.

wittjosiah: Yeah, I think I'm trying to make a design assertion on the api. @zhenyasav and I discussed the reactive object pattern and the conclusion was basically, it's an optimization and you still need callbacks. The reason being that you need somewhere to inject business logic, so the component can't be directly mutating the props. There may be exceptions and we could consider defaults for the callbacks which are simple mutation pass-throughs but we still need the callbacks in most scenarios.

thure: I see, got it, will make the change in that case.

PR #5929, opened 2024-03-13
feat: ECHO Reactive API updates, better TS types, bugfixes

dmaretskyi:
### Motivation / Background

This Pull Request ...

### Details

- ID field is present on objects with ECHO object schema.
- Better typescript types.
- New `ObjectType` helper to extract ECHO object type from schema.
- New `AnyEchoObject` export to use for references that don't have a concrete target type.
- Nested ECHO objects are now automatically added to the DB.
- Various refactorings.


packages/common/util/src/assign.ts
```diff
@@ -8,7 +8,7 @@ import { invariant } from '@dxos/invariant';
  * Initialize a deeply nested object.
  * @returns The value of the prop after assignment.
  */
-export const assignDeep = <T>(doc: any, path: (string | number)[], value: T): T => {
+export const assignDeep = <T>(doc: any, path: readonly (string | number)[], value: T): T => {
```
wittjosiah, line 11: doesn't this mean that you can _only_ pass readonly paths? won't this make it generally hard to use? (i.e. require casting a lot of things when passing in)

packages/core/echo/echo-schema/src/automerge/automerge-types.ts
```diff
@@ -25,14 +27,17 @@ export interface IDocHandle<T = any> {
...
   getValue: <T>(accessor: DocAccessor): T => get(accessor.handle.docSync(), accessor.path) as T,
 };
 
-export const createDocAccessor = <T = any>(text: TextObject): DocAccessor<T> => {
+export const createDocAccessor = <T = any>(
+  text: TextObject | EchoReactiveObject<{ content: string }>,
+): DocAccessor<T> => {
   const obj = text as any as AutomergeTextCompat;
-  return getRawDoc(obj, [obj.field]);
+  // TODO(dmaretskyi): I don't think `obj.field` is a thing anymore, can we remove it?
```
wittjosiah, line 41: the outcome of this todo would probably just be calling `getRawDoc` directly

packages/core/echo/echo-schema/src/database.ts
```diff
@@ -116,7 +118,7 @@ export class EchoDatabaseImpl implements EchoDatabase {
     return this._automerge.getObjectById(id) as T | undefined;
   }
 
-  add<T extends OpaqueEchoObject>(obj: T): T extends EchoObject ? T : EchoReactiveObject<T> {
+  add<T extends OpaqueEchoObject>(obj: T): T extends EchoObject ? T : EchoReactiveObject<{ [K in keyof T]: T[K] }> {
```
wittjosiah, line 121: why doesn't this always return an `EchoReactiveObject`?

packages/core/echo/echo-schema/src/database.ts
```diff
@@ -45,7 +41,13 @@ export interface EchoDatabase {
...
    * Query objects.
    */
-  query<T extends TypedObject>(filter?: FilterSource<T>, options?: QueryOptions): Query<T>;
+  query(): Query<TypedObject>;
+  query<T extends OpaqueEchoObject = TypedObject>(
+    filter?: Filter<T> | undefined,
+    options?: QueryOptions | undefined,
+  ): Query<T>;
+  query(filter: OperatorFilter<TypedObject>, options?: QueryOptions): Query<TypedObject>;
+  query<T extends {}>(filter?: T | undefined, options?: QueryOptions | undefined): Query<TypedObject>;
```
wittjosiah, line 50: tbh this is technically a breaking change because it forced you to change existing queries for build to pass (e.g. https://github.com/dxos/dxos/pull/5929/files#diff-77617c3cbe23613edc912f0782264e08b4ccde482e5e801cbfe0ab8f4b570f4bR65) . We need to be careful with this sort of thing. It's likely fine but should be communicated before being landed so that people are not surprised when it ships. cc @zhenyasav

packages/common/util/src/assign.ts
dmaretskyi: It goes the other way around: `string[]` is assignable to `readonly string[]` but not vice-versa

packages/core/echo/echo-schema/src/automerge/automerge-types.ts
dmaretskyi: Right, lets resolve it at the same time as we remove TextObject

packages/core/echo/echo-schema/src/database.ts
dmaretskyi: Right, we're also gonna end up removing untyped filters anyway I think

PR #5935, opened 2024-03-14
feat: allow localStorage.dxlog to override config

nferch:
Rework of https://github.com/dxos/dxos/pull/5931

### Motivation / Background

e.g. so you can `localStorage.dxlog='{ "filter": "messaging:debug,info"}'` in the browser console.

### DIFF
--- a/packages/common/log/src/platform/browser/index.ts
+++ b/packages/common/log/src/platform/browser/index.ts
@@ -6,9 +6,21 @@ import { type LogOptions } from '../../config';
 
 // NOTE: Implementation for the browser. See `package.json`.
 export const loadOptions = (filepath?: string): LogOptions | undefined => {
+  let dxlog: string | undefined;
+  if (typeof localStorage === 'undefined') {
+    if ((globalThis as any).localStorage_dxlog) {
+      dxlog = (globalThis as any).localStorage_dxlog;
+    }
+  } else {
+    dxlog = localStorage.getItem('dxlog') ?? undefined;
+  }
+  if (!dxlog) {
+    return undefined;
+  }
   try {
-    return JSON.parse(localStorage.getItem('dxlog')!);
+    return JSON.parse(dxlog);
   } catch (err) {
+    console.info("can't parse dxlog config", err);
     return undefined;
   }
 };
--- a/packages/common/log/src/platform/browser/index.ts
+++ b/packages/common/log/src/platform/browser/index.ts
@@ -7,6 +7,8 @@ import { type LogOptions } from '../../config';
 // NOTE: Implementation for the browser. See `package.json`.
 export const loadOptions = (filepath?: string): LogOptions | undefined => {
   let dxlog: string | undefined;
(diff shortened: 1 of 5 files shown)


packages/sdk/client/src/worker/onconnect.ts
```diff
@@ -57,6 +57,11 @@ export const onconnect = async (event: MessageEvent<any>) => {
   const systemChannel = new MessageChannel();
   const appChannel = new MessageChannel();
 
+  // set locally set log configuration forwarded from localStorage
+  // TODO(nf): block worker initialization until this is set? we usually win the race.
+  port.onmessage = (event) => {
+    (globalThis as any).localStorage_dxlog = event.data.dxlog;
```
wittjosiah, line 63: maybe worth wrapping this in some condition to check that this is actually a dxlog message?

packages/sdk/client/src/worker/onconnect.ts
```diff
@@ -57,6 +57,11 @@ export const onconnect = async (event: MessageEvent<any>) => {
   const systemChannel = new MessageChannel();
   const appChannel = new MessageChannel();
 
+  // set locally set log configuration forwarded from localStorage
```
wittjosiah, line 60: "set locally set log" 🤔

packages/common/log/src/platform/browser/index.ts
```diff
@@ -6,9 +6,21 @@ import { type LogOptions } from '../../config';
 
 // NOTE: Implementation for the browser. See `package.json`.
 export const loadOptions = (filepath?: string): LogOptions | undefined => {
+  let dxlog: string | undefined;
+  if (typeof localStorage === 'undefined') {
+    if ((globalThis as any).localStorage_dxlog) {
```
wittjosiah, line 11: a comment on when this gets used would be helpful. I was a little confused until I scrolled further down in the diff

wittjosiah (approved):

packages/sdk/client/src/worker/onconnect.ts
nferch: Considered that, but I wasn't sure what that would look like, it later gets parsed as JSON and now at least prints a log if that fails, but I think generally it'll only be used if you know what you're doing.

Had also considered mirroring all of localStorage, or how else this initial message could be used, but couldn't justify making it more complicated or what would stand in the way of changing it if it turns out to be too naive.

packages/sdk/client/src/worker/onconnect.ts
nferch: "set log configuration forwarded from localStorage setting"?

nferch:
@wittjosiah thanks for the 👀 ! Seems like there's an increasing amount of (internal?) developer documentation like the usage of things like this, log filter format in general, debugging, etc we should try and maintain. Do you think `REPOSITORY_GUIDE.md` is a good place for that?

wittjosiah:
@nferch repo guide seems like as good a place as any 👍

PR #5940, opened 2024-03-14
fix: Presence reactivity

thure:
This PR fixes issues with Presence reactivity.

This PR also fixes a missing translation, and issues encountered when `1rem` > `16px`.

https://github.com/dxos/dxos/assets/855039/1fc6e0f6-52c3-4fc0-892b-d0d2b1671d00


packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -59,8 +69,9 @@ export const SpacePresence = ({ object, spaceKey }: { object: TypedObject; space
     .map((member) => ({
       ...member,
       match: currentObjectViewers.has(member.identity.identityKey),
-      lastSeen: currentObjectViewers.get(member.identity.identityKey)?.lastSeen ?? Infinity,
+      lastSeen: currentObjectViewers.get(member.identity.identityKey)?.lastSeen ?? -Infinity,
```
wittjosiah, line 72: aha, yes 👍

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -44,6 +47,13 @@ export const SpacePresence = ({ object, spaceKey }: { object: TypedObject; space
   const space = spaceKey ? client.spaces.get(spaceKey) : getSpaceForObject(object);
   const spaceMembers = useMembers(space?.key);
 
+  const [moment, setMoment] = useState(Date.now());
+
+  useEffect(() => {
```
wittjosiah, line 52: maybe just add a short note on why the intervals are needed. It makes a lot of sense (we need to update the UI w/o the data necessarily changing) but at first I was confused and I suspect I would be again when reading this code without the context of the issue

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
```diff
@@ -28,6 +28,9 @@ import { ComplexMap, keyToFallback } from '@dxos/util';
 import { SPACE_PLUGIN } from '../meta';
 import type { ObjectViewerProps, SpacePluginProvides } from '../types';
 
+const REFRESH_INTERVAL = 10_000;
+const ACTIVITY_DURATION = 30_000;
```
wittjosiah, line 32: ideally we could align this with the constant that the plugin uses to set the interval for broadcasting updates

wittjosiah (approved):

packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
thure: Is one available?

PR #5956, opened 2024-03-15
chore: simplify destructuring reference types from `props` with `useDefaultValue`

ZaymonFC:
If you assign default values when destructuring props, any reference type will get a new instance every render which may result in re-rendering bugs or unexpected behaviour.

Used the following regex to search for instances in the codebase -- thanks ChatGPT.

```regex
const\s+\{(?:\s|\n)*[^\}]*=[\s]*(?:\[\]|\{\})(?:\s|\n)*[^\}]*\}(?:\s|\n)*=\s*props;
```

I wonder if there's a way to turn the above into an `ESLint` rule?

This example caused an infinite re-rendering bug on empty tables.

image

### DIFF
--- a/packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
+++ b/packages/apps/plugins/plugin-space/src/components/SpacePresence.tsx
@@ -21,6 +21,7 @@ import {
   useTranslation,
   List,
   ListItem,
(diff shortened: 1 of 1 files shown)


wittjosiah (approved):
nice!

PR #5963, opened 2024-03-17
feat(functions): lookup triggers

yarolegovich:
* Extended /lookup preset with two triggers.
* Document changes trigger sends a signal to `extract-terms` function.
* `extract-terms` function highlights terms provided by LLM in a document.
* Timer changes trigger `lookup-next-highlight` function.
* `lookup-next-highlight` chooses a thread where there was no message from LLM and triggers /lookup prompt on it.


yarolegovich:
closing as way too experimental

PR #5964, opened 2024-03-17
fix(table): Infinite re-render on column resize

ZaymonFC:
This PR fixes an infinite re-render issue on column resize.

Initially, when the resize handler was activated, it updated the parent state, which in turn caused the child table component to re-render. Upon this re-render, the resize condition was still true, triggering the resize handler again. 🔁

Introduce `useDidTransition` and `useOnTransition` hooks to fire the `columnResized` handler. If we approve of these hooks, we should move them into `packages/ui/primitives/react-hooks` with our other hooks before this merges.

- `useOnTransition` is useful because it only fires once per valid transition even when the component re-renders with the same data. Currently it achieves this through `const dirty = useRef(false) ...` but I would prefer to use `rxjs` or another stream abstraction.

### DIFF
--- a/packages/ui/react-ui-table/src/components/Table/Table.tsx
+++ b/packages/ui/react-ui-table/src/components/Table/Table.tsx
@@ -18,7 +18,6 @@ import {
 import { useVirtualizer, type VirtualizerOptions } from '@tanstack/react-virtual';
 import React, { Fragment, useCallback, useEffect, useState } from 'react';
 
-import { debounce } from '@dxos/async';
 import { log } from '@dxos/log';
 import { useDefaultValue } from '@dxos/react-ui';
 
@@ -61,6 +60,14 @@ export const Table = <TData extends RowData>(props: TableProps<TData>) => {
   }, [columns, setColumnSizing]);
 
   const [columnSizingInfo, setColumnSizingInfo] = useState<ColumnSizingInfoState>({} as ColumnSizingInfoState);
+
+  useEffect(() => {
+    if (columnSizingInfo.columnSizingStart?.length !== 0) {
+      return;
+    }
+
+    onColumnResize?.(columnSizing);
+  }, [columnSizingInfo, onColumnResize]);
 
   const [rowSelection = {}, setRowSelection] = useControllableState({
     prop: props.rowSelection,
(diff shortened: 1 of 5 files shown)


packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -58,10 +64,17 @@ export const Table = <TData extends RowData>(props: TableProps<TData>) => {
           return state;
         }, {}),
     );
+
+    setColumnsInitialised(true);
   }, [columns, setColumnSizing]);
 
   const [columnSizingInfo, setColumnSizingInfo] = useState<ColumnSizingInfoState>({} as ColumnSizingInfoState);
 
+  // Notify on column resize
```
richburdon, line 73: a nit here re our code style. Comments are full sentences terminated with period (prevent accidental cut-off).

packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -58,10 +64,17 @@ export const Table = <TData extends RowData>(props: TableProps<TData>) => {
...
+  const notifyColumnResize = useCallback(() => onColumnResize?.(columnSizing), [onColumnResize, columnSizing]);
+  useOnTransition(columnSizingInfo.isResizingColumn, (v) => typeof v === 'string', false, notifyColumnResize);
+
+  // --- Row selection
```
richburdon, line 77: `// Row selection.`

packages/ui/react-ui-table/src/hooks/useTransitions.ts
```diff
@@ -0,0 +1,75 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { useRef, useEffect, useState } from 'react';
+
+// Type guard to check if toValue is a function
```
richburdon, line 7: period

packages/ui/react-ui-table/src/hooks/useTransitions.ts
```diff
@@ -0,0 +1,75 @@
...
+ * This is an internal custom hook that checks if a value has transitioned from a specified 'from' value to a 'to' value.
+ *
+ * @param currentValue - The value that is being monitored for transitions.
+ * @param fromValue - The initial value or a predicate function that determines the start of the transition.
+ * @param toValue - The final value or a predicate function that determines the end of the transition.
+ * @returns A boolean indicating whether the transition from 'fromValue' to 'toValue' has occurred.
+ *
+ * @internal Consider using `useOnTransition` for handling transitions instead of this hook.
+ */
+export const useDidTransition = <T>(
```
richburdon, line 22: since not table specific, consider moving to common/*

richburdon (approved):

PR #6007, opened 2024-03-18
feat: Incremental improvements to the reactive ECHO API

dmaretskyi:
### Motivation / Background

This Pull Request ...

### Details

- Bumps effect deps
- Handling of `id`, `toJSON` and `[data]` in echo proxy
- Better TS types around mutability
- Support second argument in `Filter.schema`
- Effect schema for Properties
- Adds feature flag to use the new API to config


packages/sdk/client/src/echo/space-list.ts
```diff
@@ -254,7 +257,12 @@ export class SpaceList extends MulticastObservable<Space[]> implements Echo {
     const spaceProxy = (this.get().find(({ key }) => key.equals(space.spaceKey)) as SpaceProxy) ?? failUndefined();
 
     await spaceProxy._databaseInitialized.wait({ timeout: CREATE_SPACE_TIMEOUT });
-    spaceProxy.db.add(new Properties(meta));
+    if (this._config?.values?.runtime?.client?.useReactiveObjectApi ?? false) {
+      // TODO(wittjosiah): Remove cast.
+      spaceProxy.db.add(E.object(PropertiesSchema, (meta ?? {}) as any));
```
wittjosiah, line 262: how are we could to track these api changes that need to be made?

wittjosiah: sure that's fine, I just want to make sure it's being tracked _somewhere_ because it'll be hard to find them all later once they're merged into main. The feature branch gives us somewhere to review but once you move to main it gets lost in the sea of other changes.

PR #6026, opened 2024-03-19
fix(table): Each table should get own component instance

ZaymonFC:
Experienced some errors switching between tables because we're not re-building the tree / context when a new table is presented. This makes switching a little slower but more deterministic.

At this stage in the plugin development I think having a deterministic starting point is valuable for iteration.

My thinking is that we'll build a more sophisticated state / store mechanism for tables than context. In the meantime this will suffice.


packages/apps/plugins/plugin-table/src/components/TableMain.tsx
```diff
@@ -11,12 +11,19 @@ import { ObjectTable, type ObjectTableProps } from './ObjectTable';
...
+
   return (
     <Main.Content
       classNames={[baseSurface, 'fixed inset-inline-0 block-start-[--topbar-size] block-end-0 overflow-auto']}
       ref={containerRef}
     >
-      <ObjectTable table={table} stickyHeader role='grid' getScrollElement={() => containerRef.current} />
+      <ObjectTable
+        table={table}
+        key={table.id.toString()}
```
wittjosiah, line 22: I think this will always be a string so there shouldn't be a need for `.toString()`

wittjosiah (commented):
I think it would be worthwhile to include a comment about why the key is there for future consideration if/when someone wants to remove it later.

packages/apps/plugins/plugin-table/src/components/TableMain.tsx
ZaymonFC: Typed as `Symbol` at the moment. Maybe I should just do `as string`?

ZaymonFC: Ah, for some reason the type was coming through as `Symbol` but now it's string.

PR #6027, opened 2024-03-19
feat: More options for adding new sections to Stack

### LINKED ISSUE #4865: [stack] insert content between sections

thure:
This PR resolves #4865.

This PR adds a new SearchList-oriented Dialog for adding sections to a Stack, triggering it from the “Add to Stack” action at the end as well as from new menu items in each section: “Add section before” and “Add section after”.

This PR changes the `+` button at the end of StackMain to invoke the same Dialog.

https://github.com/dxos/dxos/assets/855039/8d554538-f2fc-450c-b985-841f168e033a

https://github.com/dxos/dxos/assets/855039/d7f8ff41-eaf5-48c4-91c5-7db0e395f6fb

#### Caveats

Scroll position can end up in an unexpected place in some situations. This may have something to do with Stack’s reactivity. I did’t attempt a fix in this PR because the matter is probably P2, but I’m open to input if it should be investigated.


packages/ui/react-ui-mosaic/src/mosaic/Tile.tsx
```diff
@@ -45,6 +45,8 @@ export type MosaicTileProps<TData extends MosaicDataItem = MosaicDataItem, TPosi
     onSelect?: () => void;
     onDelete?: (force?: boolean) => void;
     onNavigate?: () => void;
+    onAddBefore?: () => void;
+    onAddAfter?: () => void;
```
wittjosiah, line 49: 🫣

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
```diff
@@ -146,41 +155,29 @@ const StackMain: FC<{ stack: StackType; separation?: boolean }> = ({ stack, sepa
...
-                    >
-                      <Icon className={getSize(6)} />
-                      <span>{typeof label === 'string' ? label : t(...(label as [string, { ns: string }]))}</span>
-                    </DropdownMenu.Item>
-                  );
-                })}
-              </DropdownMenu.Viewport>
-            </DropdownMenu.Content>
-          </DropdownMenu.Root>
+          <Button
```
wittjosiah, line 165: do we think this is still needed since you can add before/after any section?

packages/apps/plugins/plugin-stack/src/components/AddSectionDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { SearchList } from '@dxos/react-ui-searchlist';
+import { type AddSectionPosition } from '@dxos/react-ui-stack';
+import { getSize, mx } from '@dxos/react-ui-theme';
+
+import { STACK_PLUGIN } from '../meta';
+import { type StackPluginProvides } from '../types';
+
+type AddSectionDialogProps = { path?: string; position: AddSectionPosition; stack: StackType };
+
+export const dataHasAddSectionDialogProps = (data: any): data is { subject: AddSectionDialogProps } => {
```
wittjosiah, line 20: is this used?

wittjosiah (approved):

packages/apps/plugins/plugin-stack/src/components/StackMain.tsx
thure: It’s a good question… I mostly kept it as company for the empty state (consistency helps here I think) and for the file upload button (which seems exotic and weird on its own). I think we can consider adjusting the design all-up in a separate increment.

packages/apps/plugins/plugin-stack/src/components/AddSectionDialog.tsx
thure: Yes, it is the guard used by the component to give a type to `data`.

packages/ui/react-ui-mosaic/src/mosaic/Tile.tsx
thure: I know… it’s another part of this interface we may like to factor out. Mosaic and `app-framework` are increasingly seeming intertwined…

packages/apps/plugins/plugin-stack/src/components/AddSectionDialog.tsx
wittjosiah: oh I see it now 👍

PR #6029, opened 2024-03-19
fix(tables): adding rows, deletions and cell edits should be durable

ZaymonFC:
I'm actually not sure why this fixes the durability / edit issues in table. But it does.

Everything works without these force-update lines now too since we're now tracking deps properly.

`¯\_(ツ)_/¯`


wittjosiah (approved):
if it works, it works :shipit:

PR #6030, opened 2024-03-19
fix(tables): table edits should be durable

ZaymonFC:
Looks like it was actually the way we were specifying deps for the column memoization. Suspect there was a race-condition and the component was rebuilding the columns and reinitialising the table before the update callbacks had a chance to be invoked. This one makes more sense @wittjosiah.


packages/apps/plugins/plugin-table/src/components/ObjectTable/ObjectTable.tsx
```diff
@@ -115,7 +110,7 @@ export const ObjectTable: FC<ObjectTableProps> = ({ table, role, stickyHeader, g
     });
 
     return builder.createColumns();
-  }, [space, tables, reactDeps(table, table.schema), newObject]);
+  }, [space, tables, table, table.schema, newObject]);
```
wittjosiah, line 113: I'm not super familiar with this plugin, but this looks like a step in the right direction to me 👍

wittjosiah (approved):

PR #6032, opened 2024-03-20
feat: Drag-and-drop images into editor.

richburdon:


packages/ui/react-ui-editor/src/themes/default.ts
```diff
@@ -90,6 +90,10 @@ export const defaultTheme: ThemeStyles = {
     color: get(tokens, 'extend.semanticColors.description.dark', 'rgba(255,255,255,.2)'),
   },
 
+  '.cm-dropCursor': {
+    border: '1px solid red',
```
wittjosiah, line 94: red?

packages/ui/react-ui-editor/src/extensions/dnd.ts
```diff
@@ -3,15 +3,23 @@
...
+export type DNDOptions = { onDrop?: (view: EditorView, event: { files: FileList }) => void };
 
-// TODO(burdon): DND experiment.
-export const dnd = (): Extension => {
-  return EditorView.domEventHandlers({
-    drop: (event, view) => {
-      log.info('domEventHandlersDrop', { event });
-    },
-  });
+export const dnd = (options: DNDOptions = {}): Extension => {
```
wittjosiah, line 10: seems like this is specifically for file dropping?

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -72,10 +72,21 @@ export const EditorMain = ({ id, readonly, toolbar, comments, extensions: _exten
...
+    const info = await fileManagerPlugin?.provides.file.upload?.(files[0]);
+    if (info) {
+      processAction(view, { type: 'image', data: info.url });
+    }
+  };
+
   const extensions = useMemo(() => {
     return [
       _extensions,
+      fileManagerPlugin?.provides.file.upload && dnd({ onDrop: handleDrop }),
```
wittjosiah, line 89:
```suggestion
      fileManagerPlugin && dnd({ onDrop: handleDrop }),
```
this isn't being used as a type guard so the rest I don't think is necessary

packages/ui/react-ui-editor/src/themes/default.ts
richburdon: suggest

packages/ui/react-ui-editor/src/extensions/dnd.ts
richburdon: ok. renamed.

packages/ui/react-ui-editor/src/themes/default.ts
wittjosiah: just thinking we should use the theme colours probably. I'd defer to @thure on colour choices here

packages/ui/react-ui-editor/src/extensions/markdown/action.ts
```diff
@@ -0,0 +1,114 @@
...
+      createComment(view);
+      break;
+  }
+
+  // TODO(burdon): Hack otherwise remains on heading selector.
+  setTimeout(() => {
+    if (!view.hasFocus) {
+      view.focus();
+    }
+  });
```
ZaymonFC, line 113: nit: It might be better to use `requestAnimationFrame` so that this gets scheduled next tick.

richburdon: nice. good to know.

PR #6037, opened 2024-03-20
feat(echo): effect schema class dsl

yarolegovich:
### Details

Addresses the problem of boilerplate required for effect schema type declarations.
Instead of:
```ts
const _TextV0Schema = S.struct({
  content: S.string,
}).pipe(E.echoObject('dxos.Text.v0', '0.1.0'));
export interface TextV0Type extends E.ObjectType<typeof _TextV0Schema> {}
export const TextV0Schema: S.Schema<TextV0Type> = _TextV0Schema;

export const isTextV0 = (data: unknown): data is E.EchoReactiveObject<TextV0Type> =>
  !!data && E.getSchema<any>(data) === TextV0Schema;
```

you can now do:
```ts
export class TextV0Type extends EchoObjectSchema({ typename: 'dxos.Text.v0', version: '0.1.0' })({
  content: S.string,
}) {}

// TextV0Type.isInstance(obj) for free
// TextV0Type.typename() vs E.getEchoObjectAnnotation(TextV0Schema)!.typename
```


packages/core/echo/echo-schema/src/effect/echo-object-class.ts
```diff
@@ -0,0 +1,84 @@
...
+    const klass: any = class {
+      constructor() {
+        throw new Error('use E.object(MyClass, fields) to instantiate an object');
+      }
+
+      static typename() {
+        return args.typename;
+      }
+
+      static isInstance(obj: unknown): obj is Klass {
```
dmaretskyi, line 64: do we want to consider `Symbol.hasInstance`?

packages/core/echo/echo-schema/src/effect/echo-object-class.ts
```diff
@@ -0,0 +1,84 @@
...
+          version: args.version,
+        },
+      }),
+    );
+    const klass: any = class {
+      constructor() {
+        throw new Error('use E.object(MyClass, fields) to instantiate an object');
+      }
+
+      static typename() {
```
dmaretskyi, line 60: what do you think about converting this to a property instead so its just `Schema.typename` to get the typename?

PR #6040, opened 2024-03-20
feat(tables): Mobile support

ZaymonFC:


packages/apps/plugins/plugin-table/src/schema/builder.tsx
```diff
@@ -128,24 +128,15 @@ export const createActionColumn = (
...
-      },
-    },
+    size: 48,
     // TODO(burdon): Translation.
     header: onColumnUpdate
       ? () => (
-          <Button variant='ghost' onClick={handleAddColumn} title='New column' classNames='m-0 p-0'>
-            <Plus className={getSize(4)} />
-          </Button>
+          <div className='flex justify-center items-center'>
```
thure, line 135: I’m beginning to prefer `grid place-items-center` for this. Grid overflow behavior in this case make a bit more sense IMO, and it makes it easier to link-up with a parent grid as a subgrid if that later becomes relevant.
```suggestion
          <div className='grid place-items-center'>
```

thure (approved):
lgtm! just a few minor items.

PR #6046, opened 2024-03-21
fix: Fix 'Space is not initialized' error in plugins

mykola-vrmchk:


mykola-vrmchk:
* **#6046** <a href="https://app.graphite.dev/github/pr/dxos/dxos/6046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @mykola-vrmchk and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -625,6 +626,7 @@ export const SpacePlugin = ({
 
               if (intent.data?.target instanceof SpaceProxy) {
                 const space = intent.data.target;
+                await space.waitUntilReady();
```
wittjosiah, line 629: Under what circumstances would we be adding an object to a space that's not open yet?

packages/apps/plugins/plugin-space/src/util.tsx
```diff
@@ -41,7 +41,7 @@ export const isSpace = (data: unknown): data is Space =>
     : false;
 
 export const getSpaceDisplayName = (space: Space): string | [string, { ns: string }] => {
-  return (space.properties.name?.length ?? 0) > 0
+  return space.state.get() === SpaceState.READY && (space.properties.name?.length ?? 0) > 0
```
wittjosiah, line 44: is the space state not handled by the second part of this ternary? my understanding is that the properties should be cached so that they're accessible even if the space is closed

mykola-vrmchk: I deleted cached properties, they were broken for 2 months now, and we do not plan to fix them

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
mykola-vrmchk: Valid point, I can delete this, but it is just a sanity check

packages/apps/plugins/plugin-space/src/util.tsx
mykola-vrmchk: Now properties is available only if state is `READY`

wittjosiah: we need to discuss how we're going to support an equivalent UX to the one cached properties enabled then

PR #6057, opened 2024-03-21
fix(tables): don't use memoizing context in table

ZaymonFC:
### Summary
This PR refactors our table component's state management from `@radix-ui/react-context` to standard React context. 

There is a significant reduction in lag post making table edits after this change.

### Motivation
The memoization benefits of `@radix-ui/react-context` were found to be less effective for the dynamic and large datasets managed by the table component.

### Future Considerations
Ideally, we'd use an indexed store mechanism or something like `use-context-selector` to improve performance even more and minimize re-rendering across the tree. Context is a bad fit for this kind of state.


thure (approved):
lgtm 👌

PR #6074, opened 2024-03-26
feat(echo): dynamic schema

yarolegovich:
### Details

`cherry-picked` changes from [this branch](https://github.com/dxos/dxos/tree/yarolegovich/dynamic-schema-plugin), which is forked from `feature/echo-api-migration` and contains some working basic `plugin-table` functionality.

* `EchoDatabase` exposes a new getter `get schemaRegistry(): DynamicSchemaRegistry;`
* `schemaRegistry` is used for adding schemas that can be changed at runtime and querying stored schemas
* `StoredEchoSchema` is the representation of a schema saved as an ECHO object
* `DynamicEchoSchema` is a wrapper around `StoredEchoSchema` that exposes utility methods and allows working with effect schema instead of JSON schema
* `FieldMetaAnnotation` is a way to store custom field properties [(like digits in plugin-table)](https://github.com/dxos/dxos/commit/7fe4b8135e4b0b96b618c1f16e5afcb8b025c6b8)


packages/core/echo/echo-schema/src/effect/dynamic/dynamic-schema.ts
```diff
@@ -0,0 +1,145 @@
...
+      if (index >= 0) {
+        updatedProperties.splice(index, 1, property);
+      } else {
+        updatedProperties.push(property);
+      }
+    }
+    const newAst: any = { ...oldAst, propertySignatures: updatedProperties };
+    const schemaWithUpdatedColumns = S.make(newAst);
+    this.serializedSchema.jsonSchema = effectToJsonSchema(schemaWithUpdatedColumns);
+  }
```
ZaymonFC, line 74: What happens when a column update causes existing entities to become invalid based on the new schema? Just curious if there are plans for how we're going to handle schema / data migrations.

ZaymonFC:
Nice work on this! The APIs for the `dynamic-schema` and `dynamic-schema-registry` look good to me for a first cut. I'm sure we can refine as needed based on usage.

What happens if a user provides `fieldMeta` on a schema that has `annotations` already? Do they get merged? Overwritten?

packages/core/echo/echo-schema/src/effect/dynamic/dynamic-schema.ts
yarolegovich: yeah, that's a tough question.
validations on write will be performed against the new schema, there are however currently no validations on read, so old values not matching the new type will be returned. for dynamic schema it would be more correct to return undefined in such case, since all columns are optional undefined will always match the schema. 
performing a validation on every read does sound like an undesirable overhead. but I think we'll need to get back to this topic at some point, in the context of lenses and different schema version interoperability

yarolegovich:
> What happens if a user provides FieldMetaAnnotations on a schema that has annotations already? Do they get merged? Overwritten?

Yeah, these will get merged.

PR #6079, opened 2024-03-27
feat(tables): Make pinned row sticky

ZaymonFC:
This PR keeps the pinned row at the bottom of the container when the table grows larger than the container.

We also scroll to the bottom when a new row is added to the bottom using the 'add row' row.

Demo
---
https://github.com/dxos/dxos/assets/12402727/e0095ff6-e9b5-4317-b6a3-f8474054b508

Dev notes
---
I don't love the `pinLastRow` option in the table API, but I'm happy with it for now until we need something more generic and expressive wrt row pinning. Overall still feels a bit janky because there's still some perf hitch in the virualiser but that's existing behaviour.

I'm thinking a bit more about state management and how there are a few things to sequence here with some race-conditions being expressed currently.

When 'add row' row is modified, we add a new empty row below, this triggers a useEffect to switch the pinned row to the new last row, then we scroll (but sometimes the container isn't full height yet).

There seems to be a few renders in between a few steps of the above leading to non-deterministic results.


thure (approved):
lgtm!

PR #6136, opened 2024-03-27
feat(tables): Column menu dropdown

ZaymonFC:
- [x] Switch to dropdown menu
- [x] Popover column menu should replace menu when clicked
- [x] Sort direction indicator
- [x] Sort direction indicator should reverse the sort then toggle off (no matter the starting direction)
- [x] Fix dropdown menu bug when opening a dropdown in scroll overflow
- [ ] Fix failing tests


packages/ui/react-ui-table/src/components/ColumnMenu.tsx
```diff
@@ -25,224 +25,157 @@ export type ColumnMenuProps<TData extends RowData, TValue> = {
...
-      </div>
-      <div className='flex gap-0'>
-        {canSort && toggleSort && <SortButton sortDirection={sortDirection} onClick={toggleSort} />}
-        <ColumnPanel {...props} column={column} />
+    <div className='flex items-center justify-center justify-between'>
+      <div className='flex items-center gap-1 truncate' title={title}>
+        <SortIndicator direction={sortDirection} onClick={onToggleSort} />
+        <div className='truncate'>{title}</div>
       </div>
+      {isMounted && (
```
thure, line 65: This doesn’t feel quite right — what’s the issue exactly? Is it because you need to pass an anchor ref?

ZaymonFC: Exactly. The `ref.current` is `undefined` until first render which causes an exception.

Feels gross though!

thure (commented):
All of this looks great, would just want/hope to find a better solution for the anchor ref conundrum. I have alternatives in mind, happy to pair on it tomorrow if you like!

packages/ui/react-ui-table/src/components/ColumnMenu.tsx
```diff
@@ -25,224 +25,157 @@ export type ColumnMenuProps<TData extends RowData, TValue> = {
...
+          {...props}
+          column={column}
+          anchorNode={columnSettingsAnchorRef.current}
+          open={isColumnSettingsOpen}
+          setOpen={setIsColumnSettingsOpen}
+        />
+      )}
+
+      <DropdownMenu.Root modal={false}>
+        <DropdownMenu.Trigger ref={columnSettingsAnchorRef}>
```
thure, line 76: Ah, I see the issue — this needs `asChild` since it has a styled button within, and its parent should be `<Popover.Anchor asChild>`, extracted from `ColumnSettingsPanel`. Both DropdownMenu and Popover roots should be ancestors of all of this.

thure: Note that `asChild` uses Radix’s `Slot` to (mostly) cleanly compose refs and handlers.

PR #6140, opened 2024-03-28
feat(cli): JSONL logging

nferch:


packages/common/log/src/processors/file-processor.ts
```diff
@@ -7,21 +7,47 @@ import { dirname } from 'node:path';
...
+    }
+    if (!shouldLog(entry, filters)) {
       return;
     }
     if (!fd) {
-      try {
-        mkdirSync(dirname(path));
-      } catch {}
-      fd = openSync(path, 'w');
+      switch (path) {
```
dmaretskyi, line 38: overloading path with special strings seems a bit flaky. Technically `stdout` is a valid filename.

Perhaps change it to `pathOrFd: string | number`?

nferch: yeah, `-` is idiomatic for stdout, but nothing exists for stderr afaik.

how would you determine from a CLI argument whether something is a string or a number?

PR #6142, opened 2024-03-29
refactor: Make proxy handlers into singletons

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -405,44 +496,48 @@ export class EchoReactiveHandlerImpl extends EchoReactiveHandler implements Reac
...
-    return target;
+    return target as EchoArrayTwoPointO<any>;
   }
 
-  getMeta(): ObjectMeta {
-    const target: any = { [symbolPath]: [], [symbolNamespace]: META_NAMESPACE };
-    return createReactiveProxy(target, this) as ObjectMeta;
+  getMeta(target: ProxyTarget): ObjectMeta {
+    const metaTarget: ProxyTarget = {
+      [symbolInternals]: target[symbolInternals],
```
yarolegovich, line 527: dataPath doesn't include namespace, so should be a separate map for meta proxy or full path should be used when doing lookup

packages/core/echo/echo-schema/src/effect/echo-handler.ts
```diff
@@ -573,42 +630,60 @@ export const createEchoReactiveObject = <T extends {}>(init: T): EchoReactiveObj
...
-    const echoHandler = new EchoReactiveHandlerImpl();
-    echoHandler._objectCore.rootProxy = proxy;
+    const core = new AutomergeObjectCore();
+    core.rootProxy = proxy;
 
-    slot.handler = echoHandler;
+    slot.handler = EchoReactiveHandlerImpl.instance;
     const target = slot.target as ProxyTarget;
+
+    target[symbolInternals] = {
```
yarolegovich, line 639: nit: extract into createObjectInternals function

packages/core/echo/echo-schema/src/effect/echo-handler.ts
dmaretskyi: for targets map?

yeah, its also broken for expandos - see my comment on where it is declared

yarolegovich: I mean imagine you have `struct({ keys: S.array(S.any) })`
here you will create an object and cache it using `['keys']` key
https://github.com/dxos/dxos/blob/2ec80d874299475b4718b958e7447f503c9ec3d2/packages/core/echo/echo-schema/src/effect/echo-handler.ts#L249
but then you request a meta object and when someone accesses keys they get the cache target from `data` namespace

dmaretskyi: done

PR #6149, opened 2024-04-01
fix: Throttle document load updates

dmaretskyi:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/core/echo/echo-schema/src/automerge/automerge-db.ts
+++ b/packages/core/echo/echo-schema/src/automerge/automerge-db.ts
@@ -2,7 +2,7 @@
 // Copyright 2023 DXOS.org
 //
 
-import { Event, asyncTimeout, synchronized } from '@dxos/async';
+import { Event, UpdateScheduler, asyncTimeout, synchronized } from '@dxos/async';
 import { type DocHandle, type DocHandleChangePayload, type DocumentId } from '@dxos/automerge/automerge-repo';
 import { Context, ContextDisposedError } from '@dxos/context';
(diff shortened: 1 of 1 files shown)


packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -388,6 +391,27 @@ export class AutomergeDb {
...
+  private _objectsForNextUpdate = new Set<string>();
+  private readonly _updateScheduler = new UpdateScheduler(
+    this._ctx,
+    async () => {
+      const ids = [...this._objectsForNextUpdate];
+      this._objectsForNextUpdate.clear();
+      this._emitUpdateEvent(ids);
+    },
+    {
+      maxFrequency: 10,
```
yarolegovich, line 404: isn't it too aggressive to potentially delay an update by 100 ms? I'd not go higher than 16ms (60hz)?

dmaretskyi: 👍

packages/core/echo/echo-schema/src/automerge/automerge-db.ts
```diff
@@ -39,7 +39,9 @@ export class AutomergeDb {
 
   readonly _updateEvent = new Event<ItemsUpdatedEvent>();
 
-  private _ctx?: Context = undefined;
+  private _isOpen = false;
```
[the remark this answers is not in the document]
dmaretskyi: I have a Resource pattern on a different branch that will solve this

PR #6170, opened 2024-04-02
feat(tables): Generate columns from effect-schema

ZaymonFC:
Currently looking for early feedback. I'm a big fan of doing the simple, naive working thing, and then evolving the implementation from there based on our needs.

Open questions and notes:
- Do we need to provide a mechanism for the caller to supply extra styles / props / meta for each type in the builder?
- We should have a generic mechanism for issuing updates
  - Table should be able to invoke some sort of validation pass first so that we can present issues to the user if the update doesn't conform to the schema.


wittjosiah (commented):
This is looking good to me so far, nice work!

> doing the simple, naive working thing, and then evolving the implementation from there

💯 

> Do we need to provide a mechanism for the caller to supply extra styles / props / meta for each type in the builder?

What's the use case you're thinking of for this?

> We should have a generic mechanism for issuing updates
> Table should be able to invoke some sort of validation pass first so that we can present issues to the user if the update doesn't conform to the schema.

This makes sense to me, but seems separable from the changes here.

Another thing that I think would be worth doing is handling the nested types better (like be discussed yesterday). This probably would mean flattening the nested fields into separate column defs (open question on how the column is labelled, perhaps default concatenates the field names with `.`?)

ZaymonFC:
Yeah I agree with you there that data {mutation|validation} should be handled at a higher layer and is orthogonal to this work.

> Another thing that I think would be worth doing is handling the nested types better (like be discussed yesterday). This probably would mean flattening the nested fields into separate column defs (open question on how the column is labelled, perhaps default concatenates the field names with `.`?)

Good point! I've updated the implementation to flatten out all nested properties.

There's going to be some interesting 'uneditable' cases here if there's an optional nested struct with all required fields. But we'll cross that bridge when we come to it. Required data is a quagmire for tables in general I think.

Issue #6099, opened 2024-03-27
[sketch] migrate tldraw files from other tldraw versions


zhenyasav:
## Motivation

A Composer user wants to import the tldraw files they've created using tldraw.com but they are failing to import.

## Finish Line

- The Sketch plugin should support importing tldraw files, regardless of the schema with which they were created.
- Update tldraw to the latest version, currently `2.0.2`
- Ensure we can migrate existing sketch objects created with the previous version of tldraw to the latest version.
- RELATED: to upgrade past `2.0.0.alpha-19`, we need to talk to the tldraw team about a [commercial license](https://tldraw.substack.com/i/139921257/the-new-license) (pricing is unclear)

jessmartin:
Did some exploration on this today and here's some notes:
- Any sketch objects created with `2.0.0.alpha-19` will fail to load once we upgrade unless we manually migrate them.
- TLDraw provides migrations across versions, and it should be possible to run the migration code to get from `2.0.0.alpha-19` to `2.0.2`
   - Here's [an example of the migration with the tldraw-yjs integration](https://github.com/tldraw/tldraw-yjs-example/commit/cc348898065d7f532618f8a888d26a77bd89e429)
   - The migration works by calling [`migratePersistedRecord()`](https://tldraw.dev/reference/store/StoreSchema#migratePersistedRecord) and passing in the serialized schema.
- When a client running an older version of the sketch plugin tries to read a newer file, it's actually possible to do `down` migrations to migrate backwards.

Some open questions:
- How do we get the serialized schemas for `2.0.0.alpha-19` objects when we haven't been persisting it up to this point? Maybe we can deploy the serialized schema in code along with the upgrade and load it somehow?
- How does this interact, if at all, with the work we're doing on echo-schema?

zhenyasav:
blocked on export issue #6490

PR #6172, opened 2024-04-02
chore: Update to TLDraw 2.0

richburdon:
- [x] Upgrade to 2.1.4
- [x] Auto-migrate data on-the-fly

- ISSUE: Need to go read-only if detect other peers have migrated the document and we are behind.

https://github.com/dxos/dxos/assets/3523355/b059fe32-d9f9-4d27-9c9c-d4d053b04004


jessmartin:
@richburdon I started on this as well, just note that we need to not land this until we handle migration, as described in #6099

packages/apps/plugins/plugin-sketch/src/hooks/adapter.ts
```diff
@@ -259,5 +280,58 @@ const decode = (value: any): any => {
...
+          store.put([record]);
+        } catch (err) {
+          log.catch(err, { record });
+        }
+      }
+    }
+  }
+};
+
+// TODO(burdon): Need unit test and check that records are duplicated.
```
wittjosiah, line 324: I don't know enough about the tldraw impl to assess this too much, but I'd say based on our other experiences with migrating data that grabbing some real sketches from existing prod spaces and testing this on them manually would be a good sanity check before shipping this.'

jessmartin (approved):

packages/apps/plugins/plugin-sketch/project.json
```diff
@@ -19,9 +19,12 @@
         ]
       }
     },
-    "lint": {}
+    "lint": {},
+    "test": {},
```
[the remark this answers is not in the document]
richburdon: i couldn't get vitest to work.

packages/apps/plugins/plugin-sketch/src/components/Sketch.stories.tsx
```diff
@@ -13,13 +13,15 @@ import { Button, Toolbar } from '@dxos/react-ui';
 
 import SketchComponent from './Sketch';
 
+const createSketch = () => create(SketchType, { data: create(Expando, {}) });
```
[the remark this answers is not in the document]
richburdon: yes

packages/apps/plugins/plugin-sketch/src/hooks/adapter.ts
richburdon: grabbing = copy and paste right? that works.
i just noticed though that copying from tldraw.com (which i assume is 2.0) doesn't work.

jessmartin:
Is the grid and snap-to-grid a setting that can be toggled?

dmaretskyi (approved):

jessmartin:
@richburdon Testing Notes:

OP = Composer peer with old TLDraw (main)
NP = Composer peer with new TLDraw (running `burdon/tldraw-2` branch) 

## Test 1: Old sketch file gets automatically updated

- Created a new space with OP
- Created a sketch file with OP and put some edits in it
- Invited NP into the same space
- Opened the sketch file with NP
- File is read-only for NP and missing a few elements (arrows and lines - see image)

Screenshot 2024-05-21 at 2 50 47 PM

My expectation is that opening the sketch file with NP would automatically migrate the file and OP would be in read-only mode.

## Test 2: New Sketch files in existing spaces are created on the new version

- Created a new space with OP
- Invited NP into the same space
- Created a new sketch file with NP
- File is read-only for NP and editable by OP

My expectation is that the sketch file would be editable by NP and read-only for OP.

## Test 3: New Sketch files in new spaces are created on the new version

- Created a new space with NP
- Created a new sketch file with NP
- File is editable by NP
- Invited OP into the same space
- File remains editable by NP and crashes OP ("Content failed to render.")
 
image

## Test 4: Old Sketch files get automatically updated to the new version (single reader/writer)

- Created a new space with OP
- Created a sketch file with OP and put some edits in it
- Updated the OP to the new code (by switching branch) and restarted Composer
- Opened the sketch file with peer running code on branch (technically, a NP, but same client)
- File is read-only, uneditable, and has some missing content (arrows, lines)

My expectation is that the file should have been automatically migrated and should now be editable.

## Other Problems

- [ ] controls and grid disappear jarringly whenever the mouse leaves the immediate canvas area

    https://github.com/dxos/dxos/assets/27258/8a7d87e3-0df9-4503-91e8-d382341d462e

- [ ] "Move to page" is available in the context menu when selecting some options, but there's no way to navigate between pages

    image

packages/apps/plugins/plugin-sketch/src/components/Sketch.tsx
```diff
@@ -30,83 +25,116 @@ export type SketchComponentProps = {
...
+      <Tldraw
+        // Setting the key forces re-rendering when the content changes.
+        key={sketch.id}
+        store={adapter.store}
+        hideUi={!active}
+        // TODO(burdon): Customize assets: https://tldraw.dev/docs/assets
+        maxAssetSize={1024 * 1024}
+        components={{
+          DebugPanel: null,
+          Grid: CustomGrid,
```
jessmartin, line 101: What if we don't like this new grid and prefer the TLDraw-standard grid? Can this be a setting?

packages/apps/plugins/plugin-sketch/src/components/Sketch.tsx
```diff
@@ -30,83 +25,116 @@ export type SketchComponentProps = {
...
-        styles.background,
-        className,
-      )}
+      className={mx('is-full bs-full', className)}
+      onPointerEnter={() => {
+        setActive(!readonly && !adapter.readonly);
+      }}
+      onPointerLeave={() => {
+        setActive(false);
+      }}
```
jessmartin, line 89: This looks like maybe it was added to make Sketch work better in Stacks/Collections, but it's really awkward when Sketch is presented in the `main` surface.

PR #6171, opened 2024-04-02
feat(shell): Reset increment

thure:
Resolves #6158. Resolves #6163.

This PR adjusts how and when storage is reset, and makes adjustments to labels & messages about same.

https://github.com/dxos/dxos/assets/855039/3dc6dfe3-1c39-4e5c-a03c-1ab5b7da3d05

https://github.com/dxos/dxos/assets/855039/d62480c9-50e1-4e86-8a8b-a951a2fe0b91

PR #6192, opened 2024-04-03
fix: Restore shell-manager

thure:
This PR restores e2e tests which rely on changes to the reset flow introduced by #6171.


thure:
Merging this optimistically, since I am able to demonstrate reset tests are only failing due to space object counts having changed, which appears to be unrelated.

PR #6197, opened 2024-04-04
chore(echo): remove typedObject and expando usages WIP

yarolegovich:
### Details

Removed `TypedObject` and `Expando` from exports and migrate remaining usages.

Work in progress.


packages/apps/tasks/project.json
```diff
@@ -31,13 +31,7 @@
...
       "outputs": ["{options.outputFile}"]
     },
-    "prebuild": {
-      "executor": "nx:run-script",
-      "options": {
-        "script": "prebuild"
-      },
-      "outputs": ["{projectRoot}/src/proto/gen"]
-    },
+    "prebuild": {},
```
wittjosiah, line 34: I think this is probably not needed anymore

packages/apps/todomvc/src/components/Todos.tsx
wittjosiah, line 1: we'll need to run the e2e tests for this to make sure it still works, have you tried it out at all yet?

PR #6212, opened 2024-04-06
feat(shell): IdentityPanel increment

thure:
This PR will implement an increment of the design of IdentityPanel:
- in-situ list of devices and agents
- in-situ profile editing, public key copying, connecting/disconnecting
- clearer join flows

https://github.com/dxos/dxos/assets/855039/02f20e39-6e6f-4a35-9018-f918cf4dedac


packages/core/protocols/src/proto/dxos/halo/credentials.proto
```diff
@@ -173,6 +173,9 @@ message DeviceProfileDocument {
   optional string os = 5;
   optional string os_version = 6;
   optional DeviceType type = 7;
+  optional string hue = 8;
+  optional string emoji = 9;
+  optional string displayName = 10;
```
thure, line 178: This isn’t used anywhere yet, though if this is functionality we might like to have it’s worth keeping it.

packages/sdk/shell/src/panels/IdentityPanel/steps/ProfileForm.tsx
```diff
@@ -10,11 +10,19 @@ import { type Identity } from '@dxos/react-client/halo';
...
+  Input,
+  useClipboardContext,
+  EmojiPickerBlock,
+  HuePickerBlock,
+} from '../../../components';
 import { type IdentityPanelStepProps } from '../IdentityPanelProps';
 import { type IdentityEvent } from '../identityMachine';
 
-export interface ProfileFormProps extends Omit<IdentityPanelStepProps, 'send'> {
+export interface ProfileFormProps extends Omit<IdentityPanelStepProps, 'send' | 'devices'> {
```
wittjosiah, line 25: is this form still used somewhere?

packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
```diff
@@ -44,83 +38,28 @@ export const IdentityActionChooser = (props: IdentityPanelStepProps) => {
...
-          >
-            <Power className={getSize(6)} />
-            <span className='grow mli-3'>{t('choose sign out label')}</span>
-            <CaretRight weight='bold' className={getSize(4)} />
-          </Action>
-        </div>
-      </DensityProvider>
-      {/* <PanelActions>{doneActionParent ? cloneElement(doneActionParent, {}, doneAction) : doneAction}</PanelActions> */}
-    </div>
+    <DeviceList
```
wittjosiah, line 58: its a little weird to me that the `IdentityActionChooser` is actually just the device list, could we avoid the wrapping and just directly render the device list? if not maybe rename the wrapper?

packages/sdk/shell/src/components/HuePicker.tsx
```diff
@@ -25,7 +34,86 @@ export type HuePickerProps = {
   onClickClear?: ButtonProps['onClick'];
 };
 
-export const HuePicker = ({ disabled, hue, onChangeHue, defaultHue, onClickClear }: HuePickerProps) => {
+export const HuePickerToolbarButton = ({
```
wittjosiah, line 37: could do with jsdoc helping to understand the difference between the two pickers and when each should be used (for emoji too)

packages/sdk/shell/src/components/DeviceList/DeviceList.tsx
```diff
@@ -0,0 +1,111 @@
...
+        data-testid='devices-panel.create-invitation'
+        onClick={onClickAdd}
+      >
+        <div role='img' className={mx(getSize(8), 'm-1 rounded-sm surface-input grid place-items-center')}>
+          <Plus weight='light' className={getSize(6)} />
+        </div>
+        <span className='grow font-medium text-start'>{t('choose add device label')}</span>
+        <CaretRight weight='bold' className={getSize(4)} />
+      </Button>
+      {agentHostingEnabled && (
```
wittjosiah, line 58: I'd tend to think that this should be its own component and not buried inside the `DeviceList` even if they do end up rendered together

packages/sdk/shell/src/panels/IdentityPanel/steps/ProfileForm.tsx
thure: It could be, if hue/icon can be adjusted for devices. Could maybe just remove for now, I’m not sure we’d want to keep much of its current implementation.

packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
thure: Based on your comment about the agent, this will host that as well. Maybe ‘action chooser’ has become a misnomer, since this is really more like “identity landing page”.

PR #6234, opened 2024-04-09
fix(tables): Myriad fixes after schema migration

ZaymonFC:
Various fixes, refactors and changes to bring Object Table / schema editing to a working state.

Resolves: #6268, #6293, #5965, #6273, #6271

Fixes:
- Can now change schema property name
- Can now change column label
- Various row pinning issues (no scroll on edit, no scroll on delete)


packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useObjects.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// TODO(Zan): Consolidate.
+const Stable = {
+  empty: {
+    object: Object.freeze({}),
+    array: Object.freeze([] as any[]),
+  },
+};
+
+export const useObjects = (space?: Space, schema?: S.Schema<any>) => {
```
wittjosiah, line 17: could you explain why this is needed?

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useObjects.ts
```diff
@@ -0,0 +1,27 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { useFilteredObjects } from '@braneframe/plugin-search';
```
wittjosiah, line 5: was table already depending on search plugin or is this new?

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useTables.ts
```diff
@@ -0,0 +1,14 @@
...
+//
+
+import { useMemo } from 'react';
+
+import { TableType } from '@braneframe/types';
+import { Filter } from '@dxos/echo-schema';
+import { type Space, useQuery } from '@dxos/react-client/echo';
+
+export const useTables = (space?: Space) => {
+  const tableFilter = useMemo(() => Filter.schema(TableType), []);
```
wittjosiah, line 12: is this still needed? does not memoizing this result in a superfluous updates? if so, we should address this at the react client level because everyone would end up needing to do this which isn't ideal

packages/apps/plugins/plugin-table/src/components/TableSettings/TableSettings.stories.tsx
```diff
@@ -8,6 +8,8 @@ import React, { useEffect, useState } from 'react';
 
 import { TableType } from '@braneframe/types';
 import { createSpaceObjectGenerator } from '@dxos/echo-generator';
+import * as E from '@dxos/echo-schema';
```
wittjosiah, line 11:
```suggestion
import { create, type DynamicEchoSchema } from '@dxos/echo-schema';
```

we're trying to move away from the `import * as E` syntax

packages/apps/plugins/plugin-table/src/schema/builder.tsx
```diff
@@ -21,49 +21,37 @@ import { getSize } from '@dxos/react-ui-theme';
...
   onColumnUpdate?: (id: string, column: ColumnProps) => void;
   onColumnDelete?: (id: string) => void;
   onRowUpdate?: (object: any, key: string, value: any) => void;
   onRowDelete?: (object: any) => void;
 };
 
-/**
- *
- */
-export class TableColumnBuilder {
```
wittjosiah, line 34: These builder classes is a pattern we've used a bunch of other packages and might be good to keep for consistency unless there's a reason it needs to go away

ZaymonFC: In this case the class implementation really obfuscates what is essentially just a call to a pure function.

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useObjects.ts
ZaymonFC: It was always there.

packages/apps/plugins/plugin-table/src/components/ObjectTable/ObjectTable.tsx
```diff
@@ -23,125 +22,117 @@ export type ObjectTableProps = Pick<TableProps<any>, 'stickyHeader' | 'role' | '
...
+    return <TableSettings open={showSettings} table={table} schemas={schemas} onClose={handleClose} />;
+  } else {
+    return (
+      <ObjectTableTable table={table} role={role} stickyHeader={stickyHeader} getScrollElement={getScrollElement} />
+    );
+  }
+};
 
-  const rows = useMemo(() => [...filteredObjects, newObject], [filteredObjects, newObject]);
+const ObjectTableTable: FC<ObjectTableProps> = ({ table, role, stickyHeader, getScrollElement }) => {
```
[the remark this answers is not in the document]
ZaymonFC: I considered that. I don't really mind. I might leave it until we have the general purpose `on create configure` pattern.

packages/apps/plugins/plugin-table/src/components/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -28,6 +28,15 @@ const Story = () => {
...
     generator.createObjects({ [TestSchemaType.project]: 6 });
 
+    // TODO(zan): Currently we need to cast as any since `_graph` is marked @internal.
+    if (!(client as any)._graph.types.isEffectSchemaRegistered(TableType)) {
+      (client as any)._graph.types.registerEffectSchema(TableType);
+    }
+
+    // We need a table to reference
+    // TODO(zan): Workout how to get this to not double add in debug.
+    space.db.add(create(TableType, { title: 'Other table', props: [], schema: generator.schemas[3] }));
```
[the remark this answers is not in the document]
ZaymonFC: Nice, thanks for that.

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useTables.ts
ZaymonFC: You're right. I removed this from the objects query but forgot this one 👍.

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useObjects.ts
ZaymonFC: It's just a little noisy in the code since it composes the query and filtering together.

ZaymonFC: The object table shouldn't care about filtering, and it felt weird seeing 'filtered objects' in the rest of the component implementation.

packages/apps/plugins/plugin-table/src/components/ObjectTable/ObjectTable.stories.tsx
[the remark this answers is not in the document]
ZaymonFC: That doesn't work out because we need to run the following first with access to client.

```
 // TODO(zan): Currently we need to cast as any since `_graph` is marked @internal.
    if (!(client as any)._graph.types.isEffectSchemaRegistered(TableType)) {
      (client as any)._graph.types.registerEffectSchema(TableType);
    }
```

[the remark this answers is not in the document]
wittjosiah: ah, we can update the callback so it has access to client, but fine to leave as a todo for now

packages/apps/plugins/plugin-table/src/components/ObjectTable/hooks/useObjects.ts
wittjosiah: Could you maybe add a jsdoc for these hooks to explain intended usage?

Issue #6227, opened 2024-04-09
[agent] file logging fails with EAGAIN


nferch:
stdout/stderr are sometimes non-blocking: https://github.com/nodejs/node/issues/42826

```Error: EAGAIN: resource temporarily unavailable, write
    at Object.writeSync (node:fs:929:3)
    at Object.writeFileSync (node:fs:2328:26)
    at appendFileSync (node:fs:2384:6)
    at /prod/cli/node_modules/.pnpm/file+packages+common+log/node_modules/@dxos/log/dist/lib/node/index.cjs:425:40
    at /prod/cli/node_modules/.pnpm/file+packages+common+log/node_modules/@dxos/log/dist/lib/node/index.cjs:516:52
    at Array.forEach (<anonymous>)
    at processLog (/prod/cli/node_modules/.pnpm/file+packages+common+log/node_modules/@dxos/log/dist/lib/node/index.cjs:516:29)
    at log2.info (/prod/cli/node_modules/.pnpm/file+packages+common+log/node_modules/@dxos/log/dist/lib/node/index.cjs:507:30)
    at ProtocolStream.<anonymous> (/prod/cli/node_modules/.pnpm/file+packages+core+mesh+teleport-extension-replicator/node_modules/@
dxos/teleport-extension-replicator/dist/lib/node/index.cjs:389:22)
    at ProtocolStream.emit (node:events:529:35) {
  errno: -11,
  syscall: 'write',
  code: 'EAGAIN'
}```

**Comment by nferch** [comment]
to repro:

```
package main

import (
        "log"
        "os"
        "os/exec"
)

func main() {

        cmd := exec.Command("node", "bin/dev", "agent", "start", "-f", "--json-log", "--verbose")
        cmd.Env = append(os.Environ(), "LOG_FILTER=debug", "JSON_LOG_FILTER=debug", "DX_PROFILE=agent-device-test4")

        cmd.Stdout = os.Stdout
        //cmd.Stderr = os.Stderr

        _, err := cmd.StderrPipe()
        if err != nil {
                panic(err)
        }

        err = cmd.Start()
        if err != nil {
                panic(err)
        }

        err = cmd.Wait()
        if err != nil {
                log.Printf("Command finished with error: %v", err)
        }
}```

may need to run a few commands like `halo identity` to generate enough log output

PR #6235, opened 2024-04-09
fix: file log processor

nferch:
closes https://github.com/dxos/dxos/issues/6227


nferch:
Also considered including a synchronous wait (perhaps https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait as https://github.com/sindresorhus/sleep-synchronously uses, but hopefully the write is not blocked for too long.

packages/common/log/src/processors/file-processor.ts
```diff
@@ -52,7 +54,31 @@ export const createFileProcessor = ({
...
+
+    // Retry writing if EAGAIN is encountered.
+    //
+    // Node may set stdout and stderr to non-blocking. https://github.com/nodejs/node/issues/42826
+    // This can cause EAGAIN errors when writing to them.
+    // In order to not drop logs, make log methods asynchronous, or deal with buffering/delayed writes, spin until write succeeds.
+
+    while (true) {
+      try {
+        return appendFileSync(fd, JSON.stringify(record) + '\n');
```
dmaretskyi, line 67: I think a better solution would be to use an async version of this function. The logger doesn't need to wait for the log to be written to disk. The only caveat here is ensuring that the logs are written in the correct oreder

packages/common/log/src/processors/file-processor.ts
```diff
@@ -52,7 +54,31 @@ export const createFileProcessor = ({
...
+      } catch (err: any) {
+        if (err.code !== 'EAGAIN') {
+          throw err;
+        }
+        if (retryTS === 0) {
+          retryTS = performance.now();
+        } else {
+          if (performance.now() - retryTS > EAGAIN_MAX_DURATION) {
```
dmaretskyi, line 75: It's concerning that this can block the thread for up to 1 second. Not sure if it ever is gonna happen in practice

nferch: yeah, it's definitely a corner case, a very large log message that exceeds the buffer size of the stderr pipe (which I'm checking whether we have any control over increasing), but it's also something this should be able to handle safely and predictably.

1s was a super high defensive estimate, I don't think it should ever be more than a ms or two in practice.

PR #6242, opened 2024-04-10
feat: Detect thread saturation

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/common/async/src/task-scheduling.ts
```diff
@@ -141,3 +142,60 @@ export const scheduleExponentialBackoffTaskInterval = (
...
+
+// If the timer is delayed, we know that the thread is saturated.
+const scheduleIdleTask = () => {
+  idleTimeout = setTimeout(() => {
+    lastIdleTime = Date.now();
+
+    // Run the tasks that were scheduled to run on idle.
+    const tasksToRun = Array.from(tasksOnIdle);
+    tasksOnIdle.length = 0;
+    for (const task of tasksToRun) {
```
yarolegovich, line 163: let's also check whether we're not taking too long after each task and potentially yield and continue on the next idle period

packages/common/async/src/task-scheduling.test.ts
```diff
@@ -44,4 +44,11 @@ describe('task-scheduling', () => {
       expect(called).to.be.false;
     });
   });
+
+  test('isThreadSaturated', async () => {
+    // just a sanity check
+    if (isThreadSaturated()) {
```
yarolegovich, line 50: I'd add another method that'll perform both operations, like if `yieldIfThreadSaturated`, I think that's going to be the most common usecase

dmaretskyi:
Needs testing

PR #6266, opened 2024-04-11
chore(shell): add missing dep

wittjosiah:
todomvc app bundling mysteriously started failing with the following error:

> failed to resolve import "@dxos/protocols" from "/home/circleci/project/packages/sdk/shell/dist/bundle/shell.mjs"

this seems to resolve it


thure (approved):
Yes but why 😅

Issue #4745, opened 2023-11-21
API Review


richburdon:
Triage major decisions for 0.4 release.
Consider future proofing requirements.

- [ ] `obj.id` vs. `obj[base]` vs. `obj.__meta` vs. `obj@foo` vs. `getSpaceForObject(obj)`
- [ ] Query/Filter.from
- [ ] links/references (strong/weak; RI)
- [ ] Expando/TypedObject/EchoObject
- [ ] Schema (zod/effect)
- [ ] Don't export SpaceProxy
- [x] https://github.com/dxos/dxos/pull/4757
- [ ] export schema with data snapshot
- [ ] deprecate protobuf
- [ ] FQ type names
- [ ] https://github.com/dxos/dxos/issues/5407
- [ ] #5326

wittjosiah:
echo api notes I have which we should review:
- dangling references are not typesafe
	- e.g., create a folder, add an object, delete that object without removing from the folder, now mapping over that folders items one will be undefined even though the types say otherwise
- signals vs multicast observable
- signals when deleted
- query update fires on object changes
- need to make it clearer for people how to subscribe to object changes in non-react model
- symbol vs __ vs @ vs fn
- include id, __typename, __schema when spreading object?
- `batch` api?
- array mutation helpers
```
const objectIndex = grid.collection.objects.findIndex((object) => object.id === id);
if (objectIndex !== -1) {
  grid.collection.objects.splice(objectIndex, 1);
}
```
- map/record data type
- factor out `TypedObject` to be able to use elsewhere
	- maybe for providing "typename" to graph datums
	- makes it simpler for dragging graph datums into ECHO, all serializable datums will "just work" as expandos
	- potentially replace the thin `deepsignal` layer with our own proxy layer that we have more control over
- creating local-only spaces
- creating spaces without identity
- looking up spaces by name
- integrate ephemeral data feed

wittjosiah:
We met today and reviewed most of the items listed above. We grouped the issues into two categories:
- now, things that should be worked on leading to a 0.5.0 release
- later, things that can be added incrementally or are for future consideration

These are listed below and need to be broken out into individual issues.

### Now

- [x] obj.id (only prop)
- [ ] reactive object does not depend on echo [Dima/Yaroslav]
- [ ] REVIEW: functions: e.g., `getSpaceForObject(obj)`, `metaOf(obj)` [Rich/Josiah]
  - [ ] `getSpace()`
  - [ ] `getMeta()` (keys only?)
[...]

PR #6277, opened 2024-04-11
refactor(echo-schema): make text helpers generic

wittjosiah:
Stop hardcoding the 'content' path in all of the text helpers and reconcile with editor cursor converter.

Part of #4745

### DIFF
--- a/packages/core/echo/echo-schema/src/text.ts
+++ b/packages/core/echo/echo-schema/src/text.ts
@@ -6,7 +6,7 @@ import get from 'lodash.get';
 
 import { next as A } from '@dxos/automerge/automerge';
 
-import { createDocAccessor } from './automerge';
+import { type DocAccessor, createDocAccessor, type KeyPath } from './automerge';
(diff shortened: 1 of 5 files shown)


packages/core/echo/echo-schema/src/text.ts
```diff
@@ -6,23 +6,28 @@ import get from 'lodash.get';
...
-// TODO(burdon): Reconcile with cursorConverter.
-
-// TODO(wittjosiah): Path shouldn't be hardcoded.
-const path = ['content'];
-
-export const toCursor = <T>(object: EchoReactiveObject<T>, pos: number) => {
-  const accessor = createDocAccessor(object, path);
-  const doc = accessor.handle.docSync();
-  if (!doc) {
+export const toCursor = <T>({
```
dmaretskyi, line 17: hmm, why not `toCursor(createDocAccessor(obj, path), index)` to make the API leaner?

dmaretskyi: e.g. having only one way of doing something...

PR #6281, opened 2024-04-11
feat(react-ui-navtree): Right-click on tree item heading opens actions menu

### LINKED ISSUE #5816: [composer] right-click to open context menu in sidebar
I'd like to be able to right-click on items in the sidebar to open the context menu. Much larger click target then having to hit the little triple-dot menu.

thure:
Resolves #5816.

This PR uses a `contextmenu` specifically to open a tree item’s action menu (if rendered). Note that this does not facilitate long-tap, however the three-dots menu trigger will always be visible on mobile.

https://github.com/dxos/dxos/assets/855039/0616f551-da38-4aaa-837d-f00a34de3427


ZaymonFC (approved):
LGTM ⚡️

PR #6284, opened 2024-04-11
feat(client)!: stop exporting proxy classes

wittjosiah:
Avoid leaking some internal implementation details as a part of the api.

Part of #4745


packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
```diff
@@ -215,12 +215,11 @@ export const DebugPlugin = (): PluginDefinition<DebugPluginProvides> => {
...
               return active === 'devtools' ? (
                 <DevtoolsMain />
-              ) : !active || typeof active !== 'object' ? null : 'space' in active &&
-                active.space instanceof SpaceProxy ? (
+              ) : !active || typeof active !== 'object' ? null : 'space' in active && isSpace(active.space) ? (
                 <DebugSpace
                   space={active.space}
                   onAddObjects={(objects) => {
-                    if (!(active.space instanceof SpaceProxy)) {
+                    if (!isSpace(active.space)) {
```
richburdon, line 222: do we need this? `instance of Space` would be fine?

packages/sdk/client-protocol/src/echo.ts
```diff
@@ -45,7 +47,7 @@ export interface Echo extends MulticastObservable<Space[]> {
   /**
    * Creates a new space.
    */
-  create(): Promise<Space>;
+  create(meta?: PropertiesProps): Promise<Space>;
```
richburdon, line 50: PropertiesProps or SpaceProps?

packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
wittjosiah: `Space` is not a class it's an interface so that won't work

packages/sdk/client-protocol/src/echo.ts
wittjosiah: I believe they're PropertiesProps because they're used as initial values for the space properties

packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
wittjosiah: If we can find a way to avoid needing this helper that'd be great though

wittjosiah: Maybe if it was an abstract class? Not sure if that works with instanceof or not

PR #6299, opened 2024-04-15
chore: Rationalize different transport implementations.

richburdon:
- [x] Implement open method to remove complex side effects from constructors.
- [x] Normalize options.
- [x] Prep for experimental signaling service to replace current KUBE solution.

### DIFF
--- a/.idea/dictionaries/default_user.xml
+++ b/.idea/dictionaries/default_user.xml
@@ -1,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="default.user">
+    <words>
+      <w>libdatachannel</w>
+    </words>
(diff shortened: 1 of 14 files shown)


richburdon:
> OTOH the async gymnastics in LibDataChannelTransport are cumbersome and confusing too

yes. i found that really difficult to reason about and debug. in general i like for classes to be well formed by their constructor and not involve any async side effects. constructor + open leaves open the possibility for an async factory that returns a completely well formed object, but i didn't want to go that far here. 

another goal was to normalize the signaling so that i can complete the worker experiment.

if you're confident we have sufficient tests in place (they look reasonable) then i'm inclined to merge this PR pending other feedback.

Issue #6301, opened 2024-04-15
[halo] delegated invitations 


yarolegovich:
- [x] Add support for `AuthMethod.KNOWN_PUBLIC_KEY`.
 - [x] Add the code for creating a `dxos.halo.credentials.SpaceInvitation` credential.
 - [x] Implement listening on the topic asserted in `SpaceInvitation` credentials (handling possibly set expiration time).
 - [x] Implement peer admission based on the successful challenge (including revocation if not multiUse).
 - [x] Handle peer selection when multiple peers are listening on the topic.

PR #6302, opened 2024-04-15
refactor(halo): make multiuse into a separate flag instead of invitation type

yarolegovich:
### Details

A small refactoring: make multiuse into a separate flag instead of invitation type.
Both offline and interactive invitations can be multiuse.
Tested on a local composer build.

re #6301

### DIFF
--- a/packages/core/protocols/src/proto/dxos/client/services.proto
+++ b/packages/core/protocols/src/proto/dxos/client/services.proto
@@ -405,10 +405,8 @@ message Invitation {
   enum Type {
     /// Requires both to be online to complete key exchange.
     INTERACTIVE = 0;
-    /// Guest's identity key is known; invitation can be accepted by any valid peer.
+    /// Invitation can be accepted by any valid peer.
     OFFLINE = 1;
-    /// Multiple-use interactive invitation.
-    MULTIUSE = 2;
   }
 
   enum Kind {
@@ -422,6 +420,9 @@ message Invitation {
 
     /// Guest should call `Authenticate` with the shared secret.
     SHARED_SECRET = 1;
+
+    /// Guest should prove they possess a private key corresponding to the known public key recorded in an invitation.
+    KNOWN_PUBLIC_KEY = 2;
   }
 
   enum State {
@@ -480,6 +481,9 @@ message Invitation {
   // TODO(nf): some feedback mechanism or GC for immortal invitations?
   // TODO(nf): should the creator have a way to make an invitation last as long as the client? this was the previous behavior.
   optional int32 lifetime = 14;
+
+  /// Whether an invitation can be used multiple times.
+  optional bool multi_use = 15;
 }
 
 message AcceptInvitationRequest {
--- a/packages/core/protocols/src/proto/dxos/client/services.proto
+++ b/packages/core/protocols/src/proto/dxos/client/services.proto
@@ -407,6 +407,11 @@ message Invitation {
     INTERACTIVE = 0;
     /// Invitation can be accepted by any valid peer.
     OFFLINE = 1;
+    /**
+     * Multi-use interactive invitations.
+     * @deprecated use multiUse flag with type=interactive instead.
+     */
+    MULTIUSE = 2;
   }
 
   enum Kind {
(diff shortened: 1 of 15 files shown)


packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -405,10 +405,8 @@ message Invitation {
   enum Type {
     /// Requires both to be online to complete key exchange.
     INTERACTIVE = 0;
-    /// Guest's identity key is known; invitation can be accepted by any valid peer.
+    /// Invitation can be accepted by any valid peer.
     OFFLINE = 1;
-    /// Multiple-use interactive invitation.
-    MULTIUSE = 2;
```
wittjosiah, line 411: if removing, perhaps this should be deprecated first so that we retain compatibility for a few versions?

packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -480,6 +481,9 @@ message Invitation {
   // TODO(nf): some feedback mechanism or GC for immortal invitations?
   // TODO(nf): should the creator have a way to make an invitation last as long as the client? this was the previous behavior.
   optional int32 lifetime = 14;
+
+  /// Whether an invitation can be used multiple times.
+  optional bool multi_use = 15;
```
wittjosiah, line 486: Generally the question that we've asked whenever adding a boolean to the api is what other values might it have later? This is why you'll tend to see more enums in the protos than bools.

Questions I'd have about this are:
- is multiuse meaningfully distinct from timeout? (probably? you might want to have an indefinite timeout but only allow it to be redeemed once, or vice versa)
- is the generic case of this just a integer for the number of uses rather than a boolean?

packages/core/protocols/src/proto/dxos/client/services.proto
wittjosiah: though...would need to consider whether the implementation supported that as well. I'm just wary of this breaking invitations between versions of shell

yarolegovich: after creation we seem to rely on authCode presence, not type

wittjosiah: but doesn't this all get encoded into the invitation code? would old implementations decode it properly?

packages/core/protocols/src/proto/dxos/client/services.proto
yarolegovich: > is the generic case of this just a integer for the number of uses rather than a boolean?

that's a good question and there can probably be legit cases for "n-use invitations" but sounds like an overcomplication since we'll need to also implement "used counter" of a kind to support this

wittjosiah: actually I'm realizing, what is the use case for multi-use + offline? if it's offline it needs a known public key and if it's got a known public key it can only admit that key so how could it be used multiple times?

packages/core/protocols/src/proto/dxos/client/services.proto
yarolegovich: right. I think protobuf maps uknown enum to 0 so we'd just use multiUse flag. 
but yeah, I'll bring it back with a deprecation comment

packages/core/protocols/src/proto/dxos/client/services.proto
wittjosiah: > that's a good question and there can probably be legit cases for "n-use invitations" but sounds like an overcomplication since we'll need to also implement "used counter" of a kind to support this

even if we don't fully implement the protocol we might consider making it an integer in the protobuf to avoid needing to change the protocol later, while our implementation could omit use-tracking in the mean time

yarolegovich: known public key is going to belong to an ephemeral keypair. you'll use it to sign a challenge and if signature passes verification your identity will be admitted

yarolegovich: so different identities can reuse it

wittjosiah: > known public key is going to belong to an ephemeral keypair

ah I now see why you don't like the term "contact" invitation 😆 

maybe "deferred" or "delegated" would be better than "offline"? I think "offline" doesn't make sense because if you're actually offline/disconnected from peers no one will see the invitation to be able to admit the person you invite even if you are offline when they show up.

yarolegovich: I like delegated

PR #6303, opened 2024-04-15
fix: Fix IDB random-access-storage driver reset

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/common/random-access-storage/src/common/storage.ts
```diff
@@ -9,6 +9,9 @@ export enum StorageType {
   RAM = 'ram',
   IDB = 'idb',
   CHROME = 'chrome',
```
wittjosiah, line 11: what is "chrome" storage?

packages/common/random-access-storage/src/common/storage.ts
```diff
@@ -9,6 +9,9 @@ export enum StorageType {
...
+  /**
+   * @deprecated
+   */
   FIREFOX = 'firefox',
   NODE = 'node',
   WEBFS = 'webfs',
```
wittjosiah, line 17: this is going to be deprecated too, right?

dmaretskyi: yes, eventually

packages/common/random-access-storage/src/common/storage.ts
dmaretskyi: Doesn't seem to be used, we can deprecate this too

PR #6312, opened 2024-04-16
fix: Defer indexing updates

dmaretskyi:
### Motivation / Background

This Pull Request ...


mykola-vrmchk (approved):

packages/core/echo/indexing/src/index-service.ts
```diff
@@ -96,15 +97,11 @@ export class IndexServiceImpl implements IndexService {
         } catch (error) {
           log.catch(error);
         }
-      };
+      });
 
-      const unsub = this._params.indexer.indexed.on(update);
+      this._params.indexer.indexed.on(ctx, () => updateTask.schedule());
```
mykola-vrmchk, line 102: It will spam warnings `too many callbacks on a context`

PR #6342, opened 2024-04-18
fix(test-bench): Table virtualiser height estimate

ZaymonFC:
Provide an option to set the estimated row height for the table virtualiser.

Fixes an issue in the testbench where the rows start to drift and offset.

We should probably sample rows and measure them in JS but this is a quick fix for now.

### DIFF
--- a/packages/apps/testbench-app/src/components/ItemTable.tsx
+++ b/packages/apps/testbench-app/src/components/ItemTable.tsx
@@ -42,6 +42,7 @@ export const ItemTable = <T extends EchoReactiveObject<any>>({ schema, objects =
         stickyHeader
         border
         getScrollElement={() => containerRef.current}
+        estimatedRowHeight={29}
       />
     </div>
   );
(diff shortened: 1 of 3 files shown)


packages/apps/testbench-app/src/components/ItemTable.tsx
```diff
@@ -42,6 +42,7 @@ export const ItemTable = <T extends EchoReactiveObject<any>>({ schema, objects =
         stickyHeader
         border
         getScrollElement={() => containerRef.current}
+        estimatedRowHeight={29}
```
richburdon, line 45: I think our standards are 32 (fine) and 40 (course) re density.
These values are defined in the theme @thure ?

ZaymonFC: Ah, if these values are coming from the DensityProvider then I should be able to use that directly instead of supply a value as a prop.

thure: Yes, though I can’t speak to whether they apply here specifically.

thure: @ZaymonFC if the row height is being used here, you may need to make a ternary operation on the provider’s value — we don’t define the size in pixels, (rather it’s `rem`), so it will differ between mobile and desktop as well as between the different density values.

ZaymonFC:
A more robust solution has been implemented in #6408

PR #6345, opened 2024-04-18
feat: agent HTTP server and WS authentication

nferch:
Implement agent HTTP endpoint with static token authentication.

### Motivation / Background
* Healthcheck appropriate for Docker runners
* WS API for external usage

### DIFF
--- a/packages/core/protocols/src/proto/dxos/config.proto
+++ b/packages/core/protocols/src/proto/dxos/config.proto
@@ -131,7 +131,10 @@ message Runtime {
     /// Location of the remote client host.
     // TODO(burdon): Rename vault_endpoint (generalize http/ws; drop vault.html).
     optional string remote_source = 6;
-    
+
+    // Authenticate via `Sec-WebSocket-Protocol` header hack.
+    optional string remote_source_authentication_token = 8;
+
     /// Connect to and serve client services to a remote proxy.
     optional string devtools_proxy = 7;
 
--- a/packages/core/protocols/src/proto/dxos/config.proto
+++ b/packages/core/protocols/src/proto/dxos/config.proto
@@ -135,7 +135,6 @@ message Runtime {
     /// Connect to and serve client services to a remote proxy.
     optional string devtools_proxy = 7;
 
-    optional string remote_source_authentication_token = 8;
   }
 
   message App {
--- a/packages/core/protocols/src/proto/dxos/config.proto
+++ b/packages/core/protocols/src/proto/dxos/config.proto
@@ -134,6 +134,8 @@ message Runtime {
     
     /// Connect to and serve client services to a remote proxy.
     optional string devtools_proxy = 7;
+
+    optional string remote_source_authentication_token = 8;
   }
 
   message App {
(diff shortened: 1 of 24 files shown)


packages/core/protocols/src/proto/dxos/config.proto
```diff
@@ -134,6 +134,8 @@ message Runtime {
     
     /// Connect to and serve client services to a remote proxy.
     optional string devtools_proxy = 7;
+
+    optional string remote_source_authentication_token = 8;
```
wittjosiah, line 138: maybe put next to remote source for clarity?

packages/devtools/cli/src/base-command.ts
```diff
@@ -206,8 +210,16 @@ export abstract class BaseCommand<T extends typeof Command = any> extends Comman
       log.config({ filter: !process.env.LOG_FILTER && !process.env.LOG_CONFIG ? LogLevel.ERROR : undefined });
     }
 
-    // Load user config file.
-    await this._loadConfig();
+    if (this.flags.target) {
+      this.log('Using remote target:', this.flags.target);
+      this._clientConfig = new Config(
```
wittjosiah, line 215: would you still have the file config here but just to override it with the target?

packages/sdk/client/src/services/socket.ts
```diff
@@ -31,13 +36,25 @@ export const fromSocket = async (url: string): Promise<ClientServicesProvider> =
...
         handlers: {},
       });
+
+      dxRpcClient.error.on(async (error) => {
+        log.warn('websocket rpc client error', { error });
+        // Browsers do not include the error message in the event object, so we cannot discern 401 errors from other errors.
+        if (error.message.includes('401')) {
+          log.warn('websocket authentication failed');
+        }
+        fatal.emit(error);
```
wittjosiah, line 51: I think that this `fatal` event might be leaking internal implementation details of the socket services provider into client. The way I would expect this to work is that if there's an error initially that opening would throw [this trigger](https://github.com/dxos/dxos/pull/6345/files#diff-785917a9d319e96d99ca9d7921879288a3b0921055c1cb6e0db85a7cc63d2d1bR312) wouldn't be needed.

If this is more for if the websocket fails after initialization, then I think we can just extend the `closed` event with a `fatal` flag that tells client not to try to restart

packages/devtools/cli/src/base-command.ts
nferch: hmm, good question. the only thing I've set in my two days of using it in this manner is the log filter since by design most of the action is happening on the remote side, but I suppose there could be things that could be applicable.

Some of this feels astonishing to me given that we use a single config file/protobuf for everything, but this isn't the only instance of that pattern and there's plenty of other things about config that I find confusing. But probably better to be consistent.

packages/sdk/client/src/services/socket.ts
nferch: yeah, I suspected that might be the case, and your solution is probably better than making initializing the ClientServicesProvider synchronous.

PR #6358, opened 2024-04-19
refactor(halo): extracted invitation management into serviceContext from serviceHost

yarolegovich:
### Motivation

Services use Managers to manage data (Service -> Manager). 
All spaces are created through DataSpaceManager. On space creation, DataSpaceManager will attach a listener of spaceState to handle delegated invitations.
DataSpaceManager shouldn't have access to InvitationService (Service <- Manager). 
Create a class that'll be used by both DataSpaceManager and InvitationService, and that'll manage invitations preventing duplicates.

### Details

* Created `InvitationsManager` and pulled the majority of code from `InvitationsService` there.
* Fixed duplicated subscriptions on `existingInvitation` return (added early return for existing invitations).
* Fixed missing subscription for invitations created in `loadPersistentInvitations` (for `_createInvitations` updates) and `.filter(async (invitation)` always returning true.
* Reordered operations in `cancelInvitation` (remove persistent invitation from metadata store first).


nferch:
LGTM

Issue #4886, opened 2023-12-07
[stack] cover/fill option in images in stacks


zhenyasav:
- Chrome for stack sections

thure:
Type-specific surfaces such as this are currently blocked by ongoing Collections work.

wittjosiah:
@thure what makes you say this is blocked by the collections work? I think this could likely be accomplished ahead of that using just the app framework plugin provides:
```ts
export type StackProvides = {
  stack: {
    creators?: StackSectionAction[];
    actions?: (object: Stack.Section['object']) => StackSectionAction[];
  };
};
```

thure:
That is a good way plugins can define actions, but  then the way actions are finally rendered into the stack section’s rail or options menu is to be determined.

I think the way this is done depends on whether titles and actions are brought with, or queried by, a section’s object (so the UI component has enough to work with from the Tile), or if the UI components do somehow need to expose more subcomponents so plugins can fill things in with greater granularity. The latter approach might work today, but we might not want to keep that implementation long enough to merit its cost.

wittjosiah:
Ah I see, so the reason you're suggesting that this is related to collections is along the same lines of what we were talking about with the collapsed state? Basically, the way to pass this stuff through to the section tile means consolidating the things into a single schema for a section.

thure:
Exactly, I think something we’d want to look at is whether/how we can enable the UI to know where to get what it needs via schemas, or if it should remain agnostic. I’m hoping we can find a solution for the former.

PR #6359, opened 2024-04-19
feat: collections

### LINKED ISSUE #5166: [organization] stacks and folders are just collections
- underlying both stacks & folders are collections of objects
- stack is just one way to render a collection of objects (another would be grid)
- we cannot simply render a folder as a stack/grid though as they have extra view metadata
  - this metadata could have reasonable defaults, but if edited it needs to be persisted somewhere
- the expectation is to resolve this with a "data layer" above the ECHO db where different objects can be joined together into a single view
- with resolvers in a data layer a "stack" can be a joined view of a collection and stack view metadata (e.g. section heights)
- currently blocked by #5124 as that is what will form the data layer (which should probably be another issue as well)

### LINKED ISSUE #6104: [organization] make spaces look like collections

### LINKED ISSUE #6166: [stack] show document title in collapsed section
image

image

wittjosiah:
### Motivation / Background

This Pull Request introduces the concept of collections, replacing folders and stacks at the data level.

Resolves #5166
Resolves #6166
Resolves #6104 
Unblocks #4886

### Details

- removes the old folder & stack types from `@braneframe/types`
- adds a `Collection` type
- adds a `StackView` type
- removes graph builder & intent resolver from stack plugin, updates it to learn how to render collections
- consolidates extra view data regarding sections (e.g. collapsed) into the section type to be wired up to mosaic tiles as items
- renders spaces as the root collection

### Additional information

- the design of the data structure ensures that the state of a view always converges, however given that views are referenced objects there could end up being orphaned objects. Consider whether we should clean these up or perhaps if views shouldn't be separate objects.
- notably this does not yet allow objects to exist in multiple collections at once (at the app level)
- allowing objects to exist in multiple collections will follow but requires some graph work as well as UX work
- it would be good to wire up a file system directory handle to map to the collection schema and render as a stack too but that can also be done as a followup

TODO
- [x] migrations
- [x] tests
- [x] verify presentations still work
- [ ] verify labs functions still work
- [x] review remaining "folder" references

image


packages/apps/types/src/schema/collection.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Expando, ref, S, TypedObject } from '@dxos/echo-schema';
+
+export class Collection extends TypedObject({ typename: 'dxos.Collection', version: '0.1.0' })({
+  name: S.optional(S.string),
+  objects: S.mutable(S.array(ref(Expando))),
+  // Key is schema typename and value is reference to a view object of the associated schema.
+  // Having collection reference the views rather than vice versa ensures that the state converges to a single view per key (i.e. type).
+  // This also leaves open a future where this key could be changed to allow for multiple stack views per section.
+  // TODO(wittjosiah): Any way to make this more type safe?
+  views: S.mutable(S.record(S.string, ref(Expando))),
```
thure, line 14: What kinds of objects can `views` reference?

thure: At first glance, this seems like it should be much less permissive than `objects`.

packages/ui/react-ui-stack/src/components/Section.tsx
```diff
@@ -299,70 +306,53 @@ export const SectionToolbar = ({ children, classNames }: SectionToolbarProps) =>
...
-      </Section>
-    );
+  const section = (
+    <Section
+      ref={forwardedRef}
+      title={title}
+      id={transformedItem.id}
+      size={transformedItem.view?.size}
+      icon={transformedItem.metadata?.icon}
+      collapsed={transformedItem.view?.collapsed}
```
thure, line 341: I see, this is making sense and looking intuitive 👍

packages/apps/types/src/schema/collection.ts
wittjosiah: The object for the view should be of the schema associated with the `typename` of the record key. Would be nice to find a way to have that be more strongly typed but I didn't want to block on that. Open to ideas here though

Issue #6370, opened 2024-04-22
[docs] "Edit this page" button has a broken link throughout the site


seagreen:
For instance, the link on [Guide/Introduction](https://docs.dxos.org/guide/) 404s, in it the second "docs" should be swapped for "content".

Before: https://github.com/dxos/dxos/edit/main/docs/docs/guide/README.md
After: https://github.com/dxos/dxos/edit/main/docs/content/guide/README.md

seagreen:
I'll take this one so I can get more familiar with how docs gen works

PR #6371, opened 2024-04-22
Fix(docs): "Edit this page" button broken link #6370

seagreen:
One-word change so that "Edit this page" goes to the right place, tested locally on a few pages.


seagreen:
Oooh, I wasn't familiar with the semantic-pull-request format. I'll change the title to lowercase `fix` and see if that helps.

wittjosiah (approved):

PR #6378, opened 2024-04-22
fix(tables): Table reactivity

### LINKED ISSUE #5722: [tables] tables don't update reactively when data changes
I know that table is not high priority right now, but I'm using this as part of the Effect Days demo.

### Reproduction

1. Open Composer and enable the Tables plugin
2. Create a table and enter some data
3. Open Composer in a second window and navigate to the same table
4. Make changes to the table in either window and notice that it does not update in the other window
5. If you navigate away from the table, and back again, the table will be updated appropriately

### LINKED ISSUE #6376: [tables] reactivity workaround
TanStack Table doesn't seem to be able to have cell-level reactivity ( #5722 ):
https://github.com/TanStack/table/issues/754#issuecomment-359112470
https://github.com/TanStack/table/issues/2824#issuecomment-1770149034

As a workaround we'd like to hook up the whole table to re-render when data changes so it's at least reactive still if not performant or particularly friendly.

ZaymonFC:
- Resolves #6376
- Resolves #5722

### Reactive updates in UI Table
https://github.com/dxos/dxos/assets/12402727/59a03595-35e8-4e9c-8d42-4c3ec1ee9c79

### Reactive updates across plugin-table instances
https://github.com/dxos/dxos/assets/12402727/7d06fcb2-7f0f-4791-abb3-19ec77779178


packages/ui/react-ui-table/src/components/Table/Table.stories.tsx
```diff
@@ -393,3 +394,48 @@ export const TenThousandRows = {
     );
   },
 };
+
+const state = deepSignal({ items: createItems(10) });
```
ZaymonFC, line 398: What's the in-house way to do this?

packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -44,57 +39,21 @@ export const Table = <TData extends RowData>(props: TableProps<TData>) => {
...
-        }, {}),
-    );
-
-    setColumnsInitialised(true);
-  }, [columns, setColumnSizing]);
+  const [data, setData] = useState([...incomingData]);
 
-  const [columnSizingInfo, setColumnSizingInfo] = useState<ColumnSizingInfoState>({} as ColumnSizingInfoState);
+  const forceUpdate = useCallback(() => setData([...incomingData]), [incomingData]);
+  useEffect(() => forceUpdate(), [JSON.stringify(incomingData), forceUpdate]);
```
ZaymonFC, line 47: Here's where the forced update change is happening

packages/ui/react-ui-table/src/helpers.tsx
```diff
@@ -102,14 +102,14 @@ const StringBuilderCell = <TData extends RowData>(cellContext: CellContext<TData
...
-    if ((initialValue as any)?.content) {
-      return (initialValue as any)?.content;
-    }
+    return (initialValue as any)?.content ?? initialValue;
+  }, [initialValue]);
 
-    return initialValue;
-  });
+  const [value, setValue] = useState(getInitialValue);
+  useEffect(() => setValue(getInitialValue()), [getInitialValue]);
```
ZaymonFC, line 112: String cells were only initialising once, caching the initial value.

packages/ui/react-ui-table/src/components/Table/Table.stories.tsx
wittjosiah:
```
import { create } from '@dxos/echo-schema';

const state = create({ items: createItems(10) });
```

(optionally schema as the first arg)

packages/ui/react-ui-table/src/components/Table/Table.tsx
wittjosiah: probably worth a comment & todo with a link to the cell reactivity issue you make

wittjosiah:
maybe rename PR "table reactivity" since "cell reactivity" is still TODO?

packages/ui/react-ui-table/src/components/Table/Table.stories.tsx
ZaymonFC: Excellent

PR #6385, opened 2024-04-23
refactor: Refactor IndexSchema to use map of sets

mykola-vrmchk:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/core/echo/indexing/src/index-schema.ts
+++ b/packages/core/echo/indexing/src/index-schema.ts
@@ -23,6 +23,7 @@ export class IndexSchema extends Resource implements Index {
 
   /**
    * Map `typename` -> Set `index id`.
+   * @see https://v8.dev/blog/hash-code for performance estimations.
    */
   private readonly _index = new Map<string | null, Set<ObjectPointerEncoded>>();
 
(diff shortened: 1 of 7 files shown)


packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -1,113 +1,78 @@
...
   static async load({ serialized, identifier }: LoadParams): Promise<IndexSchema> {
-    const deserialized = JSON.parse(serialized);
-
     const index = new IndexSchema();
-    await index.open();
-    invariant(index._orama, 'Index is not initialized');
+    const serializedIndex: Array<{ type: string | null; ids: string[] }> = JSON.parse(serialized).index;
     index._identifier = identifier;
-    await orama.load(index._orama, deserialized);
+    serializedIndex.forEach(({ type, ids }) => {
```
dmaretskyi, line 71: replace with a for loop

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -1,113 +1,78 @@
...
   }
 
   @trace.span({ showInBrowserTimeline: true })
   static async load({ serialized, identifier }: LoadParams): Promise<IndexSchema> {
-    const deserialized = JSON.parse(serialized);
-
     const index = new IndexSchema();
-    await index.open();
-    invariant(index._orama, 'Index is not initialized');
+    const serializedIndex: Array<{ type: string | null; ids: string[] }> = JSON.parse(serialized).index;
```
dmaretskyi, line 69: stick with `T[]` syntax for arrays 

```suggestion
    const serializedIndex: { type: string | null; ids: string[] }[] = JSON.parse(serialized).index;
```

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -1,113 +1,78 @@
...
     index._identifier = identifier;
-    await orama.load(index._orama, deserialized);
+    serializedIndex.forEach(({ type, ids }) => {
+      index._index.set(type, new Set(ids));
+    });
     return index;
   }
 }
+
+const getTypeFromObject = (object: Partial<ObjectStructure>) => object.system?.type?.itemId ?? null;
```
dmaretskyi, line 78:
```suggestion
const getTypeFromObject = (object: Partial<ObjectStructure>): string | null => object.system?.type?.itemId ?? null;
```

PR #6397, opened 2024-04-24
feat(plugin-deck): Deck plugin

### LINKED ISSUE #6352: [deck] integrate deck layout with composer
#### Parity
- [x] Plank heading has same title & actions as tree item

#### New
- [x] Pinning to sidebars
- [ ] Opening comments associated with documents
- [x] Closing planks
- [x] Moving planks
- [ ] URLs definitely work
- [ ] Attention _stub_, might have unpolished edges

##### Not in-scope:
- Inter-stack DnD
- Search
- Any other plugin-specific behavior (besides MarkdownPlugin and ThreadsPlugin) that relied on a monolithic `locaion.active` which will now need to use `attention.attended`.

## Original scope

- [x] fork the layout plugin and update to a new deck layout
- [x] allow the deck layout to be swapped for the regular layout via `DX_DECK` env var
- [x] the first column of the deck with be specified to be non-removable, it will be a navigation surface which the navtree fills and will be pinned by default
- [x] clicking on new items in the navtree (or any other column) will add a new column to the deck
- [x] the deck is not persisted anywhere but the url
  - ~minimally this will be the base object of the stack and the other columns will be ephemeral~
  - [x] stretch goal: include all deck column object ids in the url so the whole deck survives reloads

thure:
Resolves #6352.

https://github.com/dxos/dxos/assets/855039/9b472cff-28c7-41d7-8c27-04c6e8f1d6f1

https://github.com/dxos/dxos/assets/855039/56413cda-562f-4fe3-84e1-c1df5d5f02d0

https://github.com/dxos/dxos/assets/855039/7e572489-3bac-4e44-94fc-9d4165171eac

https://github.com/dxos/dxos/assets/855039/8ba32f78-e006-4085-9d79-d9984a32927e

This PR integrates Deck into Composer via `DeckPlugin`, a new plugin which can be used instead of `LayoutPlugin`.

`DeckPlugin` is integrated as an experimental layout enabled only with the env variable `DX_DECK`. This PR does not do any of the integration work necessary for plugins to participate with the Deck layout, though it provides some integration in MarkdownPlugin and StackPlugin.

This PR also:
- refactors app-framework’s navigation actions to better support both LayoutPlugin and DeckPlugin

Remaining:
- [ ] ThreadsPlugin can participate in both DeckPlugin and LayoutPlugin (maybe?)
- [x] Ensure LayoutPlugin is still production-ready


packages/common/util/src/navigation.ts
wittjosiah, line 1: heads up that checkAppScheme is super composer-specific

packages/apps/plugins/plugin-deck/src/components/MainLayout.tsx
wittjosiah, line 1: this isn't actually using deck yet, right?

packages/common/util/src/navigation.ts
thure: If `LayoutPlugin` and `DeckPlugin` should both share this, where should it go?

packages/apps/plugins/plugin-deck/src/components/MainLayout.tsx
thure: No, I’m settling the ontology & contexts first, then will sort out markup. You came early to the party 😅

packages/common/util/src/navigation.ts
wittjosiah: I'd just duplicate the code between the two for now. There being two is temporary while we test out deck.

PR #6428, opened 2024-04-26
refactor: Introduce EchoClient

dmaretskyi:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/core/echo/echo-db/src/client/echo-client.ts
+++ b/packages/core/echo/echo-db/src/client/echo-client.ts
@@ -1,11 +1,16 @@
-import { Context, LifecycleState, Resource } from '@dxos/context';
+//
+// Copyright 2024 DXOS.org
+//
+
+import { type Context, LifecycleState, Resource } from '@dxos/context';
 import { invariant } from '@dxos/invariant';
 import { PublicKey } from '@dxos/keys';
-import { DataService } from '@dxos/protocols/proto/dxos/echo/service';
+import { type DataService } from '@dxos/protocols/proto/dxos/echo/service';
 import { ComplexMap } from '@dxos/util';
-import { Hypergraph } from '../hypergraph';
+
 import { AutomergeContext } from '../automerge';
 import { EchoDatabaseImpl } from '../database';
+import { Hypergraph } from '../hypergraph';
 
 export type EchoClientParams = {};
 
@@ -32,7 +37,7 @@ export class EchoClient extends Resource {
   private _automergeContext: AutomergeContext | undefined = undefined;
   private _databases = new ComplexMap<PublicKey, EchoDatabaseImpl>(PublicKey.hash);
 
-  constructor({}: EchoClientParams) {
+  constructor(params: EchoClientParams) {
     super();
     this._graph = new Hypergraph();
   }
--- a/packages/core/echo/echo-db/src/client/echo-client.ts
+++ b/packages/core/echo/echo-db/src/client/echo-client.ts
@@ -29,7 +29,6 @@ export type CreateDatabaseParams = {
   owningObject?: unknown;
 };
 
-// TODO(dmaretskyi): Move to echo-db package.
 export class EchoClient extends Resource {
   private readonly _graph: Hypergraph;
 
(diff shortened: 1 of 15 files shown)


packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -0,0 +1,87 @@
...
+  spaceKey: PublicKey;
+
+  /**
+   * Space proxy reference for SDK compatibility.
+   */
+  // TODO(dmaretskyi): Remove.
+  owningObject?: unknown;
+};
+
+// TODO(dmaretskyi): Move to echo-db package.
```
wittjosiah, line 32: done?

packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -0,0 +1,87 @@
...
+import { Hypergraph } from '../hypergraph';
+
+export type EchoClientParams = {};
+
+export type ConnectToServiceParams = {
+  dataService: DataService;
+};
+
+export type CreateDatabaseParams = {
+  // TODO(dmaretskyi): Change to string id.
```
wittjosiah, line 22: why string?

packages/core/echo/echo-db/src/database.ts
```diff
@@ -98,6 +98,10 @@ export class EchoDatabaseImpl implements EchoDatabase {
     return this._automerge.spaceKey;
   }
 
+  async open() {}
```
wittjosiah, line 101: why have these if they aren't implemented? comment to explain?

packages/sdk/client/src/client/client.ts
```diff
@@ -217,8 +214,20 @@ export class Client {
     };
   }
 
+  /**
+   * @internal
+   * @deprecated
+   */
+  get _graph() {
```
wittjosiah, line 221: is this being used places right now that make it not removeable?

packages/core/echo/echo-db/src/client/echo-client.ts
dmaretskyi: done.

packages/core/echo/echo-db/src/database.ts
dmaretskyi: Stubs for the future use, just so we can structure our code around them being there. --- added a comment

packages/core/echo/echo-db/src/client/echo-client.ts
dmaretskyi: 1. space id being a public key is more of a HALO concept
2. I'm not sure if it will stay a PublicKey in HALO - we might wanna move to DIDs

String just seems more flexible, since for ECHO its just an opaque identifier

yarolegovich (approved):

packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  get graph(): Hypergraph {
+    return this._graph;
+  }
+
+  createDatabase({ spaceKey, owningObject }: CreateDatabaseParams) {
+    invariant(this._lifecycleState === LifecycleState.OPEN);
+    invariant(!this._databases.has(spaceKey), 'Database already exists.');
+    const db = new EchoDatabaseImpl({
+      spaceKey,
```
richburdon, line 78: order props (even for a map) by order of invariance: things that don't change first, to things that are being instantiated: (context, graph, key).
the first props are being "pushed down" the abstraction chain.

packages/core/echo/echo-db/src/database.ts
richburdon: see comment about open/close context/resources/interface aspect.

packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -0,0 +1,86 @@
...
+  spaceKey: PublicKey;
+
+  /**
+   * Space proxy reference for SDK compatibility.
+   */
+  // TODO(dmaretskyi): Remove.
+  owningObject?: unknown;
+};
+
+export class EchoClient extends Resource {
```
richburdon, line 32: the API for all classes should be driven from the outset from a test!

richburdon:
awesome

PR #6439, opened 2024-04-26
docs: remove apidoc's clean repo check

seagreen:
It was useful when typedoc and apidoc were run separately. Now that they're combined it prevents running apidoc.

### DIFF
--- a/tools/apidoc/src/log.ts
+++ b/tools/apidoc/src/log.ts
@@ -1,9 +1,0 @@
-//
-// Copyright 2024 DXOS.org
-//
-
-import chalk from 'chalk';
-
-export const info = (...args: any[]) => console.warn(chalk.cyan('[info]'), ...args);
-export const warn = (...args: any[]) => console.warn(chalk.yellow('[warn]'), ...args);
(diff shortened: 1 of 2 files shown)


tools/apidoc/src/log.ts
```diff
@@ -1,9 +0,0 @@
-//
-// Copyright 2024 DXOS.org
-//
-
-import chalk from 'chalk';
-
-export const info = (...args: any[]) => console.warn(chalk.cyan('[info]'), ...args);
-export const warn = (...args: any[]) => console.warn(chalk.yellow('[warn]'), ...args);
-export const err = (...args: any[]) => console.warn(chalk.red('[err]'), ...args);
```
seagreen, line 9: This is perfectly correct generally useful code. I chose to remove it since it's now unused, but would be happy to leave it as well.

seagreen:
@zhenyasav might be a good reviewer for this one, since he has the most context.

seagreen:
(I'm going to try again and get the branch name and the "Resolves" syntax right)

Issue #6437, opened 2024-04-26
[docs] `pnpm nx apidoc docs` clean repo check issues


seagreen:
Currently `pnpm nx apidoc docs` gives this error partway through, even if run from a clean repo:
```
[err] 
git repository not clean prior to regenerating docs, bailing.
```

The state of the repo at this point will have a modified `docs/content/api.json` file and a bunch of deleted `docs/content/api/` files.

This can be avoided with `DIRTY_OK=true pnpm nx apidoc docs`. The check is also turned off in CI.

None of this is necessarily bad, but we should give a couple minutes thought to whether there's a smoother way to do this.

seagreen:
(From discord): We can remove this check, it was meant for an earlier version of the system where `apidoc` didn't run `typedoc`.

PR #6440, opened 2024-04-26
docs: remove apidoc's clean repo check

seagreen:
Resolves: #6437

The check was useful when typedoc and apidoc were run separately. Now that they're combined it prevents running apidoc.


tools/apidoc/src/log.ts
```diff
@@ -1,9 +0,0 @@
-//
-// Copyright 2024 DXOS.org
-//
-
-import chalk from 'chalk';
-
-export const info = (...args: any[]) => console.warn(chalk.cyan('[info]'), ...args);
-export const warn = (...args: any[]) => console.warn(chalk.yellow('[warn]'), ...args);
-export const err = (...args: any[]) => console.warn(chalk.red('[err]'), ...args);
```
seagreen, line 9: This is perfectly correct generally useful code. I chose to remove it since it's now unused, but would be happy to leave it as well.

seagreen:
@zhenyasav might be a good reviewer for this one, since he has the most context.

(This is the 2nd PR for this, switched to the `Resolves #ISSUE_NUMBER` syntax and prefixed the branch name with my username, though the latter might be overkill since the branch is from my own repo)

tools/apidoc/src/log.ts
seagreen: I'm wrong, it's being used elsewhere. I'll add it back.

zhenyasav (approved):

PR #6443, opened 2024-04-26
fix(shell): Remove IdentityPanel effect

thure:
This PR resolves a code quality issue introduced by using an effect instead of calling the async updater as a secondary action of setting state.


nferch:
thanks @thure for looking into this. I'm also wondering if we should create a separate issue for the API to reject ineffectual Identity and Device profile changes or if it's solely the responsibility of the caller to ensure they're not duplicate? What expectation would you have as a caller?

PR #6447, opened 2024-04-28
refactor: Introduce EchoHost

dmaretskyi:
#6451


packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -50,8 +59,13 @@ export class EchoClient extends Resource {
     this._dataService = dataService;
   }
 
+  disconnectFromService() {
```
mykola-vrmchk, line 62: Is this used anywhere?

mykola-vrmchk: maybe delete if unused

packages/core/echo/echo-db/src/query/query-state.ts
```diff
@@ -24,7 +24,7 @@ type QueryRunResult = {
   changed: boolean;
 };
 
-/**
+/** s
```
mykola-vrmchk, line 27:
```suggestion
/**
```

packages/core/echo/indexing/src/types.ts
```diff
@@ -3,10 +3,20 @@
 //
 
 import { type Event } from '@dxos/async';
-import { type Filter } from '@dxos/echo-db';
 import { type ObjectStructure } from '@dxos/echo-pipeline';
 import { type IndexKind } from '@dxos/protocols/proto/dxos/echo/indexing';
 
+/**
+ * @deprecated
```
mykola-vrmchk, line 10: Add JSdoc why new code is already deprecated

packages/sdk/client/src/tests/indexing.test.ts
```diff
@@ -71,50 +65,44 @@ describe('Index queries', () => {
...
-      service,
-      echo: client.spaces,
-    });
-    client.experimental.graph.registerQuerySourceProvider(indexQuerySourceProvider);
-    const space = await client.spaces.create();
-    await addContact(space, john);
-
-    await queryIndexedContact(space, john);
+  // TODO(dmaretskyi): This test shouldn't create its own indexer.
+  test.skip('indexing stack', async () => {
```
mykola-vrmchk, line 69: Delete this test, or transfer to indexing package

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -50,14 +49,19 @@ export class IndexSchema extends Resource implements Index {
   }
 
   @trace.span({ showInBrowserTimeline: true })
-  async find(filter: Filter) {
-    if (filter.type?.itemId === undefined) {
+  async find(filter: IndexQuery) {
```
mykola-vrmchk, line 52: The `update` method still requires reference to an object, so it is strange that in one place in the same class we pass Reference, and in another we use IndexQuery

PR #6452, opened 2024-04-29
chore: Auto-generate NX tags

dmaretskyi:
```
px run-many --target=test --projects=tag:scope:echo 
```


packages/apps/composer-app/project.json
```diff
@@ -65,18 +78,6 @@
         "{options.outputPath}"
       ]
     },
-    "bundle": {
-      // Don't cache because this is writing into the output of the app bundle
```
wittjosiah, line 69: 😭 I always forget the tooling erases comments in these files

dmaretskyi: I didn't even know nx supported comments in those files

dmaretskyi: Why do people still stick with JSON for configuration instead of much better options like YAML or TOML?

wittjosiah: ubiquity presumably... we could fix this by using [json5](https://json5.org/) in our tooling though because it won't strip out comments

wittjosiah (approved):

PR #6471, opened 2024-04-29
feat: collections (again)

### LINKED ISSUE #6464: [composer] add interstitial dialog about asking for migrations on downlevel spaces
when reaching an entity in an unmigrated space, we need to bring it up more prominently to the user that the migration needs to happen.

This should trigger on the following events:
- a navigation to the space or an entity in it
- before any action at the space level that would modify the space data
- before the graph node expands in the sidebar when clicking the chevron

```
This space needs to be upgraded to continue.
(Continue, Cancel)
```

If cheap, we should consider adding a third button to this dialog for "save to disk" instantly.

wittjosiah:
Previously: #6359 #6465 

Issues:
- [x] comment thread detachment on migration
~~- [ ] rendering space while loading~~ -> can't reproduce issue
- [x] Resolves #6464
- [x] [devtool] raw json export of a space times out for large spaces
- [x] testing migration of exports of real spaces
- [x] add a step migration to delete all existing collections from the space to handle to botched partially migrated spaces


packages/core/echo/echo-db/src/core-db/core-database.ts
```diff
@@ -49,6 +49,7 @@ export class CoreDatabase {
   readonly _updateEvent = new Event<ItemsUpdatedEvent>();
 
   private _isOpen = false;
```
dmaretskyi, line 51: Is it critical that we differentiate openning, vs open states?

If so create an enum with CLOSED, OPENING, OPEN, CLOSING, variants and use it instead of the booleans

packages/sdk/client/src/echo/space-proxy.ts
```diff
@@ -412,8 +414,16 @@ export class SpaceProxy implements Space {
...
-  private async _createEpoch({ migration }: { migration?: CreateEpochRequest.Migration } = {}) {
-    await this._clientServices.services.SpacesService!.createEpoch({ spaceKey: this.key, migration });
+  private async _createEpoch({
+    migration,
+    automergeRootUrl,
+  }: { migration?: CreateEpochRequest.Migration; automergeRootUrl?: string } = {}) {
+    log('create epoch', { migration, automergeRootUrl });
+    await this._clientServices.services.SpacesService!.createEpoch({ spaceKey: this.key, migration, automergeRootUrl });
+    if (this._db.rootUrl !== automergeRootUrl) {
+      this._databaseInitialized.reset();
```
dmaretskyi, line 424: I dont think this is a good choice since database does not become un-initialized while the epoch is chainging.

A better solution would be an event (maybe on the DB iteself?) to watch for root changes

packages/sdk/migrations/src/migration-builder.ts
```diff
@@ -0,0 +1,154 @@
...
+      access: {
+        spaceKey: this._space.key.toHex(),
+      },
+      objects: this._rootDoc.objects,
+      links: {
+        ...previousLinks,
+        ...this._newLinks,
+      },
+    });
+    await this._newRoot.whenReady();
```
dmaretskyi, line 122: no need to call this

packages/sdk/migrations/src/migration-builder.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+
+    // Create new epoch.
+    await this._space.internal.createEpoch({
+      migration: CreateEpochRequest.Migration.REPLACE_AUTOMERGE_ROOT,
+      automergeRootUrl: this._newRoot.url,
+    });
+  }
+
+  private async _buildNewRoot() {
```
dmaretskyi, line 106: shouldn't be async

packages/sdk/migrations/src/migration-builder.ts
```diff
@@ -0,0 +1,154 @@
...
+  async addObject(schema: S.Schema<any>, props: any) {
+    const core = await this._createObject({ schema, props });
+    return core.id;
+  }
+
+  deleteObject(id: string) {
+    this._deleteObjects.push(id);
+  }
+
+  async changeProperties(changeFn: (properties: ObjectStructure) => void) {
```
dmaretskyi, line 70: shouldnt be async

packages/sdk/migrations/src/migration-builder.ts
```diff
@@ -0,0 +1,154 @@
...
+    const newHandle = this._repo.create<SpaceDoc>({
+      access: {
+        spaceKey: this._space.key.toHex(),
+      },
+      objects: {
+        [core.id]: core.getDoc(),
+      },
+    });
+    this._newLinks[core.id] = newHandle.url;
+    await newHandle.whenReady();
```
dmaretskyi, line 146: no need to call tihs

packages/sdk/migrations/src/migrations.ts
```diff
@@ -47,21 +52,20 @@ export class Migrations {
...
-        await migration.down({ space });
-        const index = this.migrations.indexOf(migration);
-        space.properties[this.versionProperty] = this.migrations[index - 1]?.version;
+        const builder = new MigrationBuilder(space);
+        await migration.next({ space, builder });
+        await builder.changeProperties((propertiesStructure) => {
+          invariant(this.versionProperty, 'Migrations namespace not set');
+          propertiesStructure.data[this.versionProperty] = migration.version;
+        });
+        await builder._commit();
```
dmaretskyi, line 65: rename to `commit`

wittjosiah: any reason why? generally we've prefixed internal/private apis with `_`

packages/sdk/client-protocol/src/space.ts
```diff
@@ -22,7 +23,7 @@ export interface SpaceInternal {
   get data(): SpaceData;
 
   // TODO(dmaretskyi): Return epoch info.
-  createEpoch(): Promise<void>;
+  createEpoch(options?: { migration?: CreateEpochRequest.Migration; automergeRootUrl?: string }): Promise<void>;
```
dmaretskyi, line 26: extract to CreateEpochOptions

dmaretskyi (commented):
.

dmaretskyi (changes requested):
.

Issue #6369, opened 2024-04-22
[docs] detect dead links in CI


seagreen:
Types of dead links we're interested in:

### 1. Internal links in markdown files in general

Eg the top level `README.md`.

We don't have to have a dead link detector for these, but it might be nice.

### 2. Internal links within the documentation

We definitely want a dead link detector for these, perhaps the official [links-check](https://ecosystem.vuejs.press/plugins/links-check.html) plugin.

Alternately, perhaps a solution to (1) the general dead links problem would take care of this too, but it might not work well enough since [VuePress's routing](https://vuepress.vuejs.org/guide/directory-structure.html#default-page-routing) works a little differently than a general tool would expect.

### 3. External links

It would be nice to have a checker for dead external links too, though we might want to leave it out of CI since it would be non-deterministic.

seagreen:
I see now that we already get warnings for dead links, so the issue is just adding them to CI.

seagreen:
One last title change, sorry!

seagreen:
Using VuePress's [links-check](https://ecosystem.vuejs.press/plugins/links-check.html) would certainly be easy though, if we're not sure we could always just start with that.

seagreen:
Reading through the VuePress docs [here](https://vuepress.github.io/guide/markdown.html#links), it looks like they do see being able to browse the links in your editor/GitHub as being an advantage (in the context of relative paths, but it would also apply to file extensions):

> Relative paths are a valid links to the target files, and they can navigate correctly when browsing the source files in your editor or repository.

VSCode can still identify and follow links fine even without `.md` the issue is GitHub and the dead link detectors.

seagreen:
Turns out the [links-check](https://ecosystem.vuejs.press/plugins/links-check.html) plugin option may not be the easy route after all. It requires bumping `vuepress` and some other deps in `./docs/package.json`.

### Details

Trying to install the latest version of `@vuepress/plugin-links-check`:
```
 WARN  Issues with peer dependencies found
docs
└─┬ @vuepress/plugin-links-check 2.0.0-rc.26
  ├── ✕ unmet peer vuepress@2.0.0-rc.9: found 2.0.0-rc.0
  └─┬ @vuepress/helper 2.0.0-rc.26
    └── ✕ unmet peer vuepress@2.0.0-rc.9: found 2.0.0-rc.0
```

And with the oldest one on npm:
```
 ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies

docs
[...]

PR #6483, opened 2024-04-30
Check for dead links w/ lychee, starting with a subset of the repo

seagreen:
Related PR: https://github.com/dxos/dxos/issues/6369

This PR just checks that dead links aren't present in `./docs/content/composer` for now. We can expand that to the rest of the docs and repo (converting links to end in `.md` along the way) once we're sure that lychee works well for us in CI.

### DIFF
--- a/docs/content/composer/plugins/core.md
+++ b/docs/content/composer/plugins/core.md
@@ -4,15 +4,15 @@ order: 4
 
 # Core Plugins
 
-## [Intent](intents)
+## [Intent](intent.md)
 
 Intents are a way for plugins to communicate with each other. They represent user actions and enable plugins to respond to changes in state initiated by the user or any other plugin. Similar to redux actions.
 
-## [Surface](surface)
+## [Surface](surface.md)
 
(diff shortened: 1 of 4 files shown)


seagreen:
Still need to actually enable this in CircleCI, looking into that now.

.circleci/config.yml
```diff
@@ -231,6 +231,16 @@ jobs:
       - run: date +%s > "pnpm-lock.yaml"
       - *cache_save
 
+  #
+  # Check for dead local links
+  #
+  dead_links_check:
+    docker:
+      - image: 'lycheeverse/lychee:0.14.3'
```
seagreen, line 239: Pinned, we could also use the latest but then we risk surprise failures to pass on PRs that don't touch links.

seagreen:
Ready for review.

I also considered converting the links in `./docs/content/guide` to end in `.md` as well and adding them to this PR.

I left that out for two reasons:

1. I wanted to make sure I'm on the right track before doing that work
2. There are a good number of actual broken links in there, and I didn't want to mix rewriting docs themselves with changing the docs machinery in the same PR

Issue #6468, opened 2024-04-29
[table] keep "add row" focused after new rows are created


zhenyasav:
this allows rapid multi-row entry

PR #6484, opened 2024-04-30
feat(sdk): Allow consumers of `create` to supply `id`

ZaymonFC:
### Proposed change
- If `id` is an instance of `PublicKey` or a `PublicKeyLike` that can be parsed we should allow the user to supply it.

### Rationale

When mixing rendering of in-memory / db objects it's useful to have consistent id's to key into.

Being able to assign an `id` will work towards solving #6468.


dmaretskyi:
It quite common to prohibit user-generated IDs in the API (e.g. automerge-repo). I'm not against this change, but let's discuss pros and cons before commiting.

ZaymonFC:
Found another solution. (Using a create call to make the in-memory object).

PR #6509, opened 2024-05-01
fix(shell): ensure mesh is not bundled into the shell

### LINKED ISSUE #6508: [sdk] identity panel crashes in prebuilt shell

wittjosiah:
Since mesh entrypoint wasn't marked as external it was being bundled into shell.
This was causing the identity panel to fail because the bundled hook had its own context.

Resolves #6508


ZaymonFC (approved):
LGTM.

> Thou shalt not bundle mesh into shell.

PR #6514, opened 2024-05-01
feat: agentmanager API authentication

nferch:


packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -44,7 +76,44 @@ export class DXOSAgentHostingProviderClient implements AgentHostingProviderClien
...
     };
-    log.info('AgentHostingProviderClient initialized', { config: this._config });
+  }
+
+  init(authToken?: any) {
+    if (!this._checkAuthorization(authToken)) {
+      return false;
+    }
+
+    if (!this._config.baseURL.endsWith('/')) {
```
richburdon, line 86: if needed move this check to the constructor, but it's odd to muatate the config that's passed in. Anyway URL constructor below should normalize the trailing slash.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -44,7 +76,44 @@ export class DXOSAgentHostingProviderClient implements AgentHostingProviderClien
...
+    this._rpc.disconnected.on(() => {
+      this._rpcState = 'disconnected';
+    });
+    return true;
+  }
+
+  /**
+   * Check auth token from CF worker whether identity is allowed to create agent.
+   *
+   * Note: This will prevent the client from making unnecessary requests to the AgentHostingProvider API. The
```
richburdon, line 110: Prefer to put each sentence on a separate line. Makes reading and editing simpler.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -57,10 +126,114 @@ export class DXOSAgentHostingProviderClient implements AgentHostingProviderClien
...
+    const { deviceKey } = this._halo.device!;
+
+    invariant(deviceKey, 'deviceKey not found');
+    const authDeviceCreds = await this._queryCredentials('dxos.halo.credentials.AuthorizedDevice', (cred) =>
+      PublicKey.equals(cred.subject.id, deviceKey),
+    );
+
+    invariant(authDeviceCreds.length === 1, 'Improper number of authorized devices');
+    invariant(this._rpc, 'RPC not initialized');
+    // TODO: factor out RPC operations and state management
```
richburdon, line 158: TODO(nf): Please try to make comments capitalized and end with a period.

packages/sdk/client/src/services/agent-hosting-provider.ts
```diff
@@ -57,10 +126,114 @@ export class DXOSAgentHostingProviderClient implements AgentHostingProviderClien
...
+    if (this._rpcState === 'disconnected') {
+      await this._rpc.open();
+    }
+    await this._agentManagerAuth(authDeviceCreds[0]);
+  }
+
+  async _agentManagerAuth(authDeviceCreds: Credential) {
```
richburdon, line 165: Good to have a 1 sentence comment for this critical fn. (_authAgentManager?)

E.g., Initiates the auth sequence to retrieve the public key, then presents the credential...

packages/core/protocols/src/proto/dxos/service/agentmanager.proto
```diff
@@ -0,0 +1,53 @@
...
+
+message InitAuthSequenceRequest {
+  // Auth token for the agentmanager to validate authorization.
+  // TODO: any?
+  optional string auth_token = 1;
+}
+message InitAuthSequenceResponse {
+  enum InitAuthSequenceResult {
+    UNKNOWN = 0;
+      SUCCESS = 1;
```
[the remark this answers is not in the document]
nferch: oof, thought my IDE would autoformat these.

packages/core/protocols/src/proto/dxos/service/agentmanager.proto
```diff
@@ -0,0 +1,53 @@
...
+import "dxos/halo/credentials.proto";
+import "dxos/keys.proto";
+
+package dxos.service.agentmanager;
+
+option go_package = "github.com/dxos/kube/proto/def/dxos/service/agentmanager";
+
+message InitAuthSequenceRequest {
+  // Auth token for the agentmanager to validate authorization.
+  // TODO: any?
```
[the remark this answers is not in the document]
nferch: considered using an `Any` but didn't know how to do that using dxrpc. string is probably fine as it's opaque anyway.

packages/sdk/client/src/services/agent-hosting-provider.ts
nferch: I'll move it, but the check is necessary as it won't append the path unless there's a trailing slash and we don't want to assume the caller will provide one. the [WhatWG](https://url.spec.whatwg.org/) API is pretty astonishing to me in this regard.

packages/sdk/react-client/src/client/useAgentHostingProvider.ts
```diff
@@ -2,36 +2,43 @@
...
+      log.info('using FakeAgentHostingProvider');
+      return new FakeAgentHostingProvider(config, halo);
 
-  return new DXOSAgentHostingProviderClient(config);
+    case 'AGENTHOSTING_API':
+      if (localStorage.getItem('dxos.org/shell/features/agentHosting') !== 'true') {
+        log.info('dxos.org/shell/features/agentHosting not enabled');
+        return null;
+      }
+      agentHostingProviderClient = new AgentManagerClient(config, halo);
```
thure, line 33: Yeah, as a rule of thumb hooks should not contain the `new` keyword outside of a `useRef` or `useState` call’s arguments.

Instead, seems like we need an `AgentHostingContext` which creates the client using the same props as this hook’s current arguments, then this hook simply gets that instance.

thure (approved):
lgtm! make sure to push the context provider we made while pairing

PR #6536, opened 2024-05-02
fix: refactor Halo.queryCredentials

nferch:
* create `credentials` getter that uses `MulticastObservable`
* change `queryCredentials` into a convenience function that returns currently known credentials. very little actual usage of this method currently.


packages/sdk/client-protocol/src/halo.ts
```diff
@@ -31,14 +32,7 @@ export interface Halo {
...
 
-  // TODO(wittjosiah): Migrate to multicast observable.
-  queryCredentials(options?: {
-    ids?: PublicKey[];
-    type?: string;
-  }): ObservableProvider<
-    { onUpdate: (credentials: Credential[]) => void; onError: (error?: Error) => void },
-    Credential[]
-  >;
+  queryCredentials(options: { ids?: PublicKey[]; type?: string }): Credential[];
```
wittjosiah, line 35: something about this feels off, but it's certainly not worse than what was already here so ✅ 

I think what feels off is that it now feels like these credentials methods should be off of `credentials` instead sort of like what we did with `spaces`. Probably just a todo for the next api review though.

realizing this is _technically_ an breaking api change but it's such an obscure api at this point that it's probably fine

wittjosiah (approved):

packages/sdk/client-protocol/src/halo.ts
nferch: thanks for the 👀 !

yeah, I considered removing it completely but see some utility, but it has some non-obvious implications. should probably mark it `@experimental` at least?

I could see how making the filter/predicate off of `credentials` would provide better ergonomics, but I think the more dicey thing is finding a way as a caller to express when the query should be complete, given that the credentials may be in the process of being loaded from disk or replicated from a peer.

maybe this is better attacked once we have a clearer sense of what the future of auth/hypercore is?

Issue #6538, opened 2024-05-03
[shell] shell is not coming up on composer prod


zhenyasav:
clicking my halo button or clicking on share a space action causes composer to freeze and doesn't bring up the shell (production)

wittjosiah:
I noticed this a few days ago but at the time it seemed to only be happening to my profile. Good to know you're seeing it too, maybe it's related to profile size? I looked into it a bit today and why it "freezes" is because the shell iframe is unhidden but never renders anything so there's an invisible thing in front of the whole window. I've not yet been able to track down _why_ nothing is rendering though.

PR #6549, opened 2024-05-03
fix(shell): shell should always be closeable

wittjosiah:
### Motivation / Background

This Pull Request helps to mitigate the issue seen in #6538

### Details

- updates shell manager to not make the shell iframe visible if the rpc connection fails
- updates shell runner to render a fallback dialog which allows the shell to be closed if the shell client fails to initialize

### Additional information

- This likely does not resolve the issue, but it seems to be caused by one of the runtime or client timing out so this will help mitigate the issue now, while we continue to investigate, and in the future


thure (approved):
👌 lgtm

PR #6550, opened 2024-05-03
docs: change the background color to near white

seagreen:
(This is for the light theme)

I found the gray background to be too low contrast.

### DIFF
--- a/docs/content/.vuepress/styles/palette.scss
+++ b/docs/content/.vuepress/styles/palette.scss
@@ -8,7 +8,7 @@ $font-family: 'DM Sans';
 $font-family-heading: 'Sharp Sans, sans-serif';
 
 $bg-color: (
-  light: $zinc-100,
+  light: #fcfdfc,
   dark: $zinc-800,
 );
 


seagreen:
Side-by-side:
image

docs/content/.vuepress/styles/palette.scss
```diff
@@ -8,7 +8,7 @@ $font-family: 'DM Sans';
 $font-family-heading: 'Sharp Sans, sans-serif';
 
 $bg-color: (
-  light: $zinc-100,
+  light: #fcfdfc,
```
thure, line 11: Is there a `$zinc-50`? Light themes tend to need these fractional shades, but sometimes they’re not available…

thure (approved):
Fine by me, though I might prefer to split the difference or make all the surfaces the same background, since the difference between #ffffff and #fcfdfc might not even be perceivable.

thure (changes requested):
Actually, if we’re going to bother making changes to something like this, we should probably try to align this with Composer instead of doing something totally new.

Screenshot 2024-05-03 at 15 27 11

thure:
These are Composer’s design tokens:

## Semantic
Screenshot 2024-05-03 at 15 28 51

## Physical
Screenshot 2024-05-03 at 15 29 48

docs/content/.vuepress/styles/palette.scss
seagreen: There sure is. `zinc-50` on left, current site on right:
image

thure: Are all the other surfaces also using zinc-50? They look indiscernable, and now the message box in the main area looks out-of-flow.

I think elevation effects (i.e. shadows) are way overapplied in this theme, def think we should flatten things a bit.

seagreen: > Are all the other surfaces also using zinc-50?

If you mean the sidebar and nav bar then yes, they use the background too. The message box is plain `#ffffff`, but the way it now blends into the background is a good example of how changing the background will have lots of other effects, so we should do it once and get it right.

> I think elevation effects (i.e. shadows) are way overapplied in this theme, def think we should flatten things a bit.

Definitely makes sense to me.

seagreen:
Made an issue: https://github.com/dxos/dxos/issues/6551

PR #6552, opened 2024-05-04
docs: remove audio from autoplaying video

seagreen:
It was distracting when you navigate to the page and didn't serve an important purpose since it was only mouse click sounds, not a voiceover.

The result matches what we do for the other autoplaying video on http://localhost:8080/guide/getting-started.html, where the "volume" button is grayed out and can't be clicked.


seagreen:
Put the wrong link in here, hold up.

seagreen:
New PR: https://github.com/dxos/dxos/pull/6553

PR #6553, opened 2024-05-04
docs: remove audio from autoplaying video

seagreen:
It was distracting when you navigate to the page and didn't serve an important purpose since it was only mouse click sounds, not a voiceover.

The result matches what we do for the other autoplaying video on https://docs.dxos.org/guide/getting-started.html, where the "volume" button is grayed out and can't be clicked.


jessmartin:
ugh yes, I noticed this as well. thanks for this PR.

jessmartin (approved):

PR #6554, opened 2024-05-04
docs: de-emphasize Firefox issue since it only affect private mode

seagreen:
Also: adjust headers so they display correctly in the VuePress sidebar. It expects only one top-level header at the top of the page, the others need to be increased a level (`## Foo` instead of `# Foo`).


seagreen:
Before:
image
After:
image

Assuming I understand things right, this is annoying because we'll need to avoid `# Foo` level headers throughout the docs or the sidebar won't show correctly (except the first header of a page which needs to be `# Foo`).

jessmartin (approved):
LGTM! thanks for picking up on that comment in the Discord.

zhenyasav (approved):

PR #6556, opened 2024-05-05
chore: remove unneeded @cloudflare/ai dependency

Cherry:
### Details

Cloudflare ships the `Ai` type directly in `workers-types` now so there's no need to add `@cloudflare/ai` as a separate dependency.

This bumps `workers-types`, and then removes the no longer needed `@cloudflare/ai` package.


packages/experimental/labs-workers/package.json
```diff
@@ -30,7 +29,7 @@
     "modelfusion": "^0.137.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240423.0",
+    "@cloudflare/workers-types": "^4.20240502.0",
```
richburdon, line 32: we'd need to bump packages/apps/composer/app also to be in sync.

richburdon (approved):
thanks for spotting this.

packages/experimental/labs-workers/package.json
Cherry: good catch, updated this now!

Cherry:
No longer needed after https://github.com/dxos/dxos/pull/6448 was merged

PR #6557, opened 2024-05-06
feat(halo): delegated invitations

yarolegovich:
### Details

* Cancel delegated invitation in HALO chain when an invitation is cancelled.
* Fixed connection initiation with a peer which left a swarm while we were sleeping.
* Expose `allPeers` to topology as part of `SwarmState` (needed for invitation topology to cleanup peer ids).
* Allow overriding control heartbeat interval on `Teleport` creation for faster failure detection in tests.
* Fixed rpc closed while opening early return.
* Close control connection as soon as `RpcClosedError` is received to speed up connection termination.
* Split InvitationExtension into separate host&guest files.
* Enforce flow state invariants in host invitation flow.
* Use mutex to serialize invitation flow for many-guests and/or many-hosts scenarios.
* Created `InvitationTopology` to try properly connecting hosts with guests.
* Reviewed error handling during invitation flow and added tests for invitations handler.

re #6301

### DIFF
--- a/packages/common/async/src/mutex.ts
+++ b/packages/common/async/src/mutex.ts
@@ -4,7 +4,6 @@
 
 // Import explicit resource management polyfill.
 import '@dxos/util';
-import { cancelWithContext, type Context } from '@dxos/context';
 import { warnAfterTimeout } from '@dxos/debug';
 
 /**
@@ -139,32 +138,4 @@ export const synchronized = (
     }
   };
   Object.defineProperty(descriptor.value, 'name', { value: propertyName + '$synchronized' });
-};
-
-export const acquireInContext = async (
-  ctx: Context,
-  mutex: Mutex,
-  options: { releaseOnDispose: boolean } = { releaseOnDispose: true },
-): Promise<MutexGuard> => {
-  let guard: MutexGuard | undefined;
-  try {
-    return await cancelWithContext(
-      ctx,
-      (async () => {
-        guard = await mutex.acquire();
-        if (ctx.disposed) {
-          guard.release();
-          guard = undefined;
-          throw new Error('Guard released because acquired after context was disposed.');
-        }
-        if (options.releaseOnDispose) {
-          ctx.onDispose(() => guard?.release());
-        }
-        return guard;
-      })(),
-    );
-  } catch (e) {
-    guard?.release();
-    throw e;
-  }
 };
--- a/packages/common/async/src/mutex.ts
+++ b/packages/common/async/src/mutex.ts
@@ -2,11 +2,10 @@
 // Copyright 2020 DXOS.org
 //
 
-import { warnAfterTimeout } from '@dxos/debug';
-
 // Import explicit resource management polyfill.
 import '@dxos/util';
 import { cancelWithContext, type Context } from '@dxos/context';
+import { warnAfterTimeout } from '@dxos/debug';
 
 /**
  * A locking mechanism to ensure that a given section of the code is executed by only one single "thread" at a time.
--- a/packages/common/async/src/mutex.ts
+++ b/packages/common/async/src/mutex.ts
@@ -6,6 +6,7 @@ import { warnAfterTimeout } from '@dxos/debug';
 
 // Import explicit resource management polyfill.
 import '@dxos/util';
+import { cancelWithContext, type Context } from '@dxos/context';
 
 /**
  * A locking mechanism to ensure that a given section of the code is executed by only one single "thread" at a time.
@@ -25,11 +26,16 @@ import '@dxos/util';
  */
 export class Mutex {
   private _queue = Promise.resolve();
+  private _queueLength = 0;
 
   private _tag: string | null = null;
 
   get tag() {
     return this._tag;
+  }
+
+  isLocked(): boolean {
+    return this._queueLength > 0;
   }
 
   /**
@@ -43,8 +49,10 @@ export class Mutex {
 
     // Immediately update the promise before invoking any async actions so that next invocation waits for our task to complete.
     let guard!: MutexGuard;
+    this._queueLength++;
     this._queue = new Promise((resolve) => {
       guard = new MutexGuard(() => {
+        this._queueLength--;
         this._tag = null;
         resolve();
       });
@@ -132,4 +140,32 @@ export const synchronized = (
     }
   };
   Object.defineProperty(descriptor.value, 'name', { value: propertyName + '$synchronized' });
+};
+
+export const acquireInContext = async (
+  ctx: Context,
+  mutex: Mutex,
+  options: { releaseOnDispose: boolean } = { releaseOnDispose: true },
+): Promise<MutexGuard> => {
+  let guard: MutexGuard | undefined;
+  try {
+    return await cancelWithContext(
+      ctx,
+      (async () => {
+        guard = await mutex.acquire();
+        if (ctx.disposed) {
+          guard.release();
+          guard = undefined;
+          throw new Error('Guard released because acquired after context was disposed.');
+        }
+        if (options.releaseOnDispose) {
+          ctx.onDispose(() => guard?.release());
+        }
+        return guard;
+      })(),
+    );
+  } catch (e) {
+    guard?.release();
+    throw e;
+  }
 };
--- a/packages/core/halo/credentials/src/credentials/credential-generator.ts
+++ b/packages/core/halo/credentials/src/credentials/credential-generator.ts
@@ -249,3 +249,18 @@ export const createDelegatedSpaceInvitationCredential = async (
   });
   return { credential: { credential } };
 };
+
+export const createCancelDelegatedSpaceInvitationCredential = async (
+  signer: CredentialSigner,
+  subject: PublicKey,
+  invitationCredentialId: PublicKey,
+): Promise<FeedMessage.Payload> => {
+  const credential = await signer.createCredential({
+    subject,
+    assertion: {
+      '@type': 'dxos.halo.invitations.CancelDelegatedInvitation',
+      credentialId: invitationCredentialId,
+    },
+  });
+  return { credential: { credential } };
+};
--- a/packages/core/halo/credentials/src/credentials/credential-generator.ts
+++ b/packages/core/halo/credentials/src/credentials/credential-generator.ts
@@ -250,6 +250,11 @@ export const createDelegatedSpaceInvitationCredential = async (
   return { credential: { credential } };
 };
 
+/**
+ * @param signer - credential issuer.
+ * @param subject - key of the space the invitation was for.
+ * @param invitationCredentialId id of a dxos.halo.invitations.DelegateSpaceInvitation credential.
+ */
 export const createCancelDelegatedSpaceInvitationCredential = async (
   signer: CredentialSigner,
   subject: PublicKey,
--- a/packages/core/mesh/network-manager/src/swarm/connection.ts
+++ b/packages/core/mesh/network-manager/src/swarm/connection.ts
@@ -2,7 +2,7 @@
 // Copyright 2021 DXOS.org
 //
 
-import { DeferredTask, Event, sleep, scheduleTask, scheduleTaskInterval, synchronized } from '@dxos/async';
+import { DeferredTask, Event, sleep, scheduleTask, scheduleTaskInterval, synchronized, Trigger } from '@dxos/async';
 import { Context, cancelWithContext } from '@dxos/context';
 import { ErrorStream } from '@dxos/debug';
 import { invariant } from '@dxos/invariant';
@@ -100,6 +100,9 @@ export class Connection {
   private readonly _ctx = new Context();
   private connectedTimeoutContext = new Context();
 
+  private _protocolClosed = new Trigger();
+  private _transportClosed = new Trigger();
+
   private _state: ConnectionState = ConnectionState.CREATED;
   private _transport: Transport | undefined;
   closeReason?: string;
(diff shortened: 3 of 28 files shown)


packages/core/mesh/network-manager/src/swarm/swarm.ts
```diff
@@ -357,6 +358,7 @@ export class Swarm {
 
     // It is likely that the other peer will also try to connect to us at the same time.
     // If our peerId is higher, we will wait for a bit so that other peer has a chance to connect first.
+    const peer = this._getOrCreatePeer(remoteId);
```
yarolegovich, line 361: Create peer before going to sleep so that peer can be removed if we receive peerLeft event. In this case we won't try to establish a connection after we wake up.

packages/core/mesh/rpc/src/rpc.ts
```diff
@@ -170,16 +170,15 @@ export class RpcPeer {
     await Promise.race([this._remoteOpenTrigger.wait(), this._closingTrigger.wait()]);
 
     this._clearOpenInterval?.();
-    this._state = RpcState.OPENED;
```
yarolegovich, line 173: the next if-check could never trigger (`this._state !== RpcState.OPENED`). instead assign before waking up `_remoteOpenTrigger`

packages/core/mesh/rpc/src/rpc.ts
```diff
@@ -539,6 +539,14 @@ export class RpcPeer {
   }
 }
 
+const DEBUG_LOG_ALL_CALLS = true;
```
yarolegovich, line 542: quick way to disable very noisy logs

packages/core/mesh/teleport/src/control-extension.ts
```diff
@@ -106,17 +109,20 @@ export class ControlExtension implements TeleportExtension {
...
+                  remotePeerId: this.remotePeerId.truncate(),
+                });
+              }
             }
           }
         } catch (err: any) {
           const now = Date.now();
           if (err instanceof RpcClosedError) {
             log('ignoring RpcClosedError in heartbeat');
+            this._extensionContext.close(err);
```
yarolegovich, line 125: speed up rpc failure detection

packages/core/mesh/network-manager/src/swarm/connection.ts
```diff
@@ -348,6 +352,18 @@ export class Connection {
     this._callbacks?.onClosed?.(err);
   }
 
+  private async _closeProtocol() {
```
yarolegovich, line 355: this addresses the non-reentrancy of synchronized destroy. 
if `this._protocol.stream.on('closed'` is triggered when `protocol.close()` is called we'll start waiting for the lock for `destroy()` which we actually hold.
use a trigger to signal the code in `destroy()` to continue

packages/sdk/client-services/src/packlets/invitations/invitations-manager.ts
```diff
@@ -239,6 +244,40 @@ export class InvitationsManager {
     return { ctx, stream, observableInvitation };
   }
 
+  private _createObservableAcceptingInvitation(handler: InvitationProtocol, initialState: Invitation) {
```
yarolegovich, line 247: extracted it from invitations-handler for API consistency with host-side invitation flow

packages/core/mesh/teleport/src/rpc-extension.ts
```diff
@@ -17,15 +17,15 @@ export abstract class RpcExtension<Client, Server> implements TeleportExtension
   constructor(private readonly _rpcParams: Omit<ProtoRpcPeerOptions<Client, Server>, 'port' | 'handlers'>) {}
 
   get initiator() {
-    return this._extensionContext.initiator;
+    return this._extensionContext?.initiator;
```
yarolegovich, line 20: not feeling confident about this change, but made it just to reduce noise in logs. fixes property access on undefined for when a teleport connection was aborted during opening.
connection registration cannot be easily synchronized with teleport destroy and abort because it's not done inside teleport instance
```ts
export const createTeleportProtocolFactory = (
  onConnection: (teleport: Teleport) => Promise<void>,
  defaultParams?: Partial<TeleportParams>,
): WireProtocolProvider => {
  return (params) => {
    const teleport = new Teleport({ ...defaultParams, ...params });
    return {
      stream: teleport.stream,
      open: async (sessionId?: PublicKey) => {
        await teleport.open(sessionId);
        await onConnection(teleport); // here
      }
      ...
};
```

packages/core/halo/credentials/src/credentials/credential-generator.ts
```diff
@@ -249,3 +249,18 @@ export const createDelegatedSpaceInvitationCredential = async (
   });
   return { credential: { credential } };
 };
+
+export const createCancelDelegatedSpaceInvitationCredential = async (
+  signer: CredentialSigner,
+  subject: PublicKey,
```
dmaretskyi, line 255: not immediatelly obvious who would be the subject here -- add jsdoc pls

packages/core/mesh/network-manager/src/topology/topology.ts
```diff
@@ -36,6 +36,11 @@ export interface SwarmState {
    * Candidates for connection. Does not intersect with a set of already connected peers.
    */
   candidates: PublicKey[];
+
+  /**
+   * Candidates for connection. Does not intersect with a set of already connected peers.
```
dmaretskyi, line 41: wrong jsdoc?

packages/core/mesh/rpc/src/rpc.ts
```diff
@@ -170,16 +170,15 @@ export class RpcPeer {
     await Promise.race([this._remoteOpenTrigger.wait(), this._closingTrigger.wait()]);
 
     this._clearOpenInterval?.();
-    this._state = RpcState.OPENED;
 
-    if (this._state !== RpcState.OPENED) {
+    if ((this._state as RpcState) !== RpcState.OPENED) {
```
dmaretskyi, line 174: is this a bug in tsc type reduction?

packages/core/mesh/rpc/src/rpc.ts
dmaretskyi: lol

packages/core/mesh/rpc/src/rpc.ts
dmaretskyi, line 174: since you've removed the assignment above, is the cast still needed?

packages/core/mesh/rpc/src/rpc.ts
dmaretskyi: this seems like an anti pattern:

first, this will loose any logging metadata (line numbers, instance ids, etc.)
second, there's a way to filter logs by filename, perhaps a better way would be to improve those filters instead (e.g, by adding negative filers: `-debug:rpc.ts`)

packages/core/mesh/teleport/src/control-extension.ts
dmaretskyi: so this covers the case where a remote peer closes the RPC first.
Perhaps a better way would be to add a `close` event to RpcPeer and propagate the closure to the extension context

packages/common/async/src/mutex.ts
```diff
@@ -133,3 +140,31 @@ export const synchronized = (
...
+      ctx,
+      (async () => {
+        guard = await mutex.acquire();
+        if (ctx.disposed) {
+          guard.release();
+          guard = undefined;
+          throw new Error('Guard released because acquired after context was disposed.');
+        }
+        if (options.releaseOnDispose) {
+          ctx.onDispose(() => guard?.release());
```
dmaretskyi, line 161: onDispose returns a function to clear this callback. Ideally we'd call it when the guard gets released. Otherwise the callbacks can pile-up in a long-running context

packages/common/async/src/mutex.ts
```diff
@@ -133,3 +140,31 @@ export const synchronized = (
   };
   Object.defineProperty(descriptor.value, 'name', { value: propertyName + '$synchronized' });
 };
+
+export const acquireInContext = async (
```
dmaretskyi, line 144: Can you pls add a short jsdoc describing the usecase.

packages/sdk/client-services/src/packlets/invitations/invitation-guest-extenstion.ts
```diff
@@ -0,0 +1,129 @@
...
+    await this._destroy();
+  }
+
+  override async onAbort(): Promise<void> {
+    await this._destroy();
+  }
+
+  private async _destroy() {
+    try {
+      await this._ctx.dispose();
```
dmaretskyi, line 120: context.dispose never throws

packages/sdk/client-services/src/packlets/invitations/invitation-protocol.ts
```diff
@@ -32,10 +32,15 @@ export interface InvitationProtocol {
   getInvitationContext(): Partial<Invitation> & Pick<Invitation, 'kind'>;
 
   /**
-   * Once authentication is successful, the host can admit the guest to the requested resource.
+   * Allow authorized peers to handle this invitation behalf of invitation creator.
    */
   delegate(invitation: Invitation): Promise<PublicKey>;
```
dmaretskyi, line 37: What is the PublicKey being returned?

packages/sdk/client-services/src/packlets/invitations/invitation-topology.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    // don't start a connection while we have an active invitation flow
+    if (connected.length > 0) {
+      // update seenPeers here as well in case another host initiated a connection with us
+      connected.forEach((c) => this._seenPeers.add(c));
+      return;
+    }
+
+    const firstUnknownPeer = candidates.find((peerId) => !this._seenPeers.has(peerId));
```
dmaretskyi, line 54: this seems to exclude any possibility for retries (due to a network failure for example)

packages/sdk/client-services/src/packlets/invitations/invitations-handler.ts
```diff
@@ -162,111 +173,112 @@ export class InvitationsHandler {
...
               log('connected', { ...protocol.toJSON() });
-              setState({ state: Invitation.State.CONNECTED });
+              guardedState.set(extension, Invitation.State.CONNECTED);
 
               // 1. Introduce guest to host.
               log('introduce', { ...protocol.toJSON() });
-              const introductionResponse = await extension.rpc.InvitationHostService.introduce(
-                protocol.createIntroduction(),
-              );
+              const introductionResponse = await extension.rpc.InvitationHostService.introduce({
```
dmaretskyi, line 257: for the future: perhaps this should be an xstate state-machiene

packages/common/async/src/mutex.ts
```diff
@@ -133,3 +140,31 @@ export const synchronized = (
...
+  let guard: MutexGuard | undefined;
+  try {
+    return await cancelWithContext(
+      ctx,
+      (async () => {
+        guard = await mutex.acquire();
+        if (ctx.disposed) {
+          guard.release();
+          guard = undefined;
+          throw new Error('Guard released because acquired after context was disposed.');
```
dmaretskyi, line 158: ContextDisposedError?

packages/common/async/src/mutex.ts
```diff
@@ -133,3 +140,31 @@ export const synchronized = (
...
+        }
+        if (options.releaseOnDispose) {
+          ctx.onDispose(() => guard?.release());
+        }
+        return guard;
+      })(),
+    );
+  } catch (e) {
```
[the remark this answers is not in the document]
yarolegovich: `await cancelWithContext` throws if context is disposed, but yeah, the guard-acquiring promise should be enough:
```ts
        guard = await mutex.acquire();
        // if we're the second to settle we'll go inside this block and release
        // if we're the first to settle cancelWithContext won't throw 
        if (ctx.disposed) {
          guard.release();
          guard = undefined;
          throw new Error('Guard released because acquired after context was disposed.');
        }
        if (options.releaseOnDispose) {
          ctx.onDispose(() => guard?.release());
        }
        return guard;
```

packages/sdk/client-services/src/packlets/invitations/invitation-topology.ts
yarolegovich: with the same host - yes, unless peer rejoins the swarm (random peer ids are generated for invitations)

packages/sdk/client-services/src/packlets/invitations/invitation-guest-extenstion.ts
```diff
@@ -0,0 +1,129 @@
...
+  override async onOpen(context: ExtensionContext) {
+    await super.onOpen(context);
+
+    try {
+      log('guest acquire lock');
+      this._invitationFlowLock = await acquireInContext(this._ctx, this._invitationFlowMutex, {
+        releaseOnDispose: false,
+      });
+      log('guest lock acquired');
+      await cancelWithContext(this._ctx, this.rpc.InvitationHostService.options({ role: Options.Role.GUEST }));
```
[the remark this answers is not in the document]
yarolegovich: I think we should because otherwise guest won't know we're in a flow with another guest and won't service them, so they'll similarly stall, just on a later stage.
the problem is how the code is structured: part of the flow is in extension, but part is in handler. I also agree about xstate, but I didn't want to start a major refactoring cause these changes are quite big and complicated already and the vision for halo2.0 isn't clear yet (local-first-auth?), so this code might soon change again

dmaretskyi:
Lets go through the core invitation code together when you have time

packages/core/mesh/rpc/src/rpc.ts
yarolegovich: some typescript weirdness as if it doesn't recognize await 
image

packages/core/mesh/network-manager/src/swarm/connection.ts
yarolegovich: `await protocol.close() -> teleport.close() -> extension.close() -> context.dispose() -> leaveSwarm() -> peer.destroy() -> connection.close()` 
I could fix it for this particular case, but it's very error-prone 🤷‍♂️

PR #6615, opened 2024-05-08
feat: joyride improvements

zhenyasav:
- [x] gives our joyride helper the ability to process async "before" steps between joyride steps
- [x] shows elements that would otherwise be hidden when joyriding on them
- [x] tweaks concrete joyride steps
- [x] fix joyride replay-ability


packages/ui/react-ui-theme/src/styles/fragments/surface.ts
```diff
@@ -25,3 +25,5 @@ export const modalSurface = 'modal-surface';
 // Elements that actively have the user’s “attention”; prefer to select with CSS-native selectors like focus-within,
 // but may be applied by app state. ⚠️ Do not apply statically.
 export const attentionSurface = 'attention-surface';
+
+export const accentSurface = 'surface-accent fg-inverse';
```
thure, line 29: I meant to mark these fragments as deprecated, though I suppose elements with `surface-accent` are far more consistent about the foreground than any other.

Still, if this is applied to a Button, it needs `hover:surface-accentHover`, and the better scope of concern is the component style rather than a fragment.

thure (approved):
lgtm, not super confident about promoting accent surfaces so far upstream, but it’s fine for now.

PR #6617, opened 2024-05-09
fix(react-ui-stack): Label getter call

thure:
This PR fixes cases where a Stack section’s `transformedItem.label` may be defined but uncallable.


ZaymonFC (approved):
LGTM

PR #6636, opened 2024-05-10
feat: Add WNFS plugin

icidasset:
### Motivation / Background

Adds a WNFS plugin that allows for private image uploads in Composer which are synced to other devices using the Blob (edge) service.

### Details

Initial working version of the idea.
Connects to the `http://localhost:8787` blob service, is there a way to configure the URL for this?
Depends on https://github.com/dxos/edge/pull/91

Demo:
https://discord.com/channels/837138313172353095/1142114470479536368/1284153524103549009

Needs some fine tuning, notes/todos:
- Codemirror plugin needs improving (I don't have any experience with this, just threw something together)
- A WNFS is built up out of several merkle DAGs, it makes an HTTP request for each one of them (only 1st time, after = cache). Both for retrieval and storage. So this might turn out slow for big files, could write a batching mechanism.
- Related to the above. What if the person closes the browser during the remote blob storage call? The blob service won't have the data. I guess we need some guard rails here. Like tracking if the blob upload was successful, if not, retry on next app launch. Or maybe, prevent the browser from closing during upload.
- Still needs to check if the browser is online or not. If not, don't try to push to the remote blob store and track which ones need to uploaded later on when the connection comes back.
- Loading animations for images.

### DIFF
--- a/packages/plugins/experimental/plugin-wnfs/src/WnfsPlugin.tsx
+++ b/packages/plugins/experimental/plugin-wnfs/src/WnfsPlugin.tsx
@@ -1,0 +1,102 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { FileCloud, type IconProps } from '@phosphor-icons/react';
+import React, { type Ref } from 'react';
+
+import { type PluginDefinition, isObject, resolvePlugin } from '@dxos/app-framework';
+import { invariant } from '@dxos/invariant';
+import { parseClientPlugin } from '@dxos/plugin-client';
+import { type DocumentType } from '@dxos/plugin-markdown/types';
+import { type Space, getSpace } from '@dxos/react-client/echo';
+import { isTileComponentProps } from '@dxos/react-ui-mosaic';
+
+import * as Blockstore from './blockstore';
+import { FileCard, FileMain, FileSection, FileSlide } from './components';
+import { image as imageExtension } from './extensions';
+import meta, { WNFS_PLUGIN } from './meta';
+import translations from './translations';
+import { FileType } from './types';
+import { type WnfsPluginProvides } from './types';
+import { upload } from './upload';
+
+export const WnfsPlugin = (): PluginDefinition<WnfsPluginProvides> => {
+  let blockstore: Blockstore.MixedBlockstore | undefined;
+
+  return {
+    meta,
+    ready: async (plugins) => {
+      const client = resolvePlugin(plugins, parseClientPlugin)?.provides.client;
+      invariant(client);
+      const apiHost = client.config.values.runtime?.services?.edge?.url || 'http://localhost:8787';
+      blockstore = Blockstore.create(apiHost);
+      await blockstore.open();
+    },
+    provides: {
+      translations,
+      metadata: {
+        records: {
+          [FileType.typename]: {
+            placeholder: ['file title placeholder', { ns: WNFS_PLUGIN }],
+            icon: (props: IconProps) => <FileCloud {...props} />,
+          },
+        },
+      },
+      echo: {
+        schema: [FileType],
+      },
+      // TODO(burdon): Add intent to upload file.
+      file: {
+        upload: async (file: File, space: Space) => {
+          if (!blockstore) {
+            throw new Error('Blockstore is not ready yet');
+          }
+
+          return await upload({ blockstore, file, space });
+        },
+      },
+      markdown: {
+        extensions: ({ document }: { document?: DocumentType }) => {
+          if (!blockstore) {
+            throw new Error('Blockstore is not ready yet');
+          }
+
+          if (document) {
+            const space = getSpace(document);
+            return space ? [imageExtension({ blockstore, space })] : [];
+          }
+
+          return [];
+        },
+      },
+      surface: {
+        component: ({ data, role, ...props }, forwardedRef) => {
+          switch (role) {
+            case 'main':
+              return data.active instanceof FileType ? <FileMain file={data.active} /> : null;
+            case 'slide':
+              return data.slide instanceof FileType ? <FileSlide file={data.slide} cover={false} /> : null;
+            case 'section':
+              return data.object instanceof FileType ? <FileSection file={data.object} /> : null;
+            case 'card': {
+              if (
+                isObject(data.content) &&
+                typeof data.content.id === 'string' &&
+                data.content.object instanceof FileType
+              ) {
+                const cardProps = { ...props, item: { id: data.content.id, object: data.content.object } };
+                return isTileComponentProps(cardProps) ? (
+                  <FileCard {...cardProps} ref={forwardedRef as Ref<HTMLDivElement>} />
+                ) : null;
+              }
+              break;
+            }
+          }
+
+          return null;
+        },
+      },
+    },
+  };
+};
--- a/packages/plugins/experimental/plugin-wnfs/src/blockstore.ts
+++ b/packages/plugins/experimental/plugin-wnfs/src/blockstore.ts
@@ -1,0 +1,96 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { BaseBlockstore } from 'blockstore-core';
+import { IDBBlockstore } from 'blockstore-idb';
+import type { CID } from 'multiformats';
+
+import { storeName } from './common';
+
+export const create = (apiHost?: string) => {
+  return new MixedBlockstore(apiHost);
+};
+
+/**
+ * A blockstore that communicates with the DXOS blob service,
+ * and which uses an indexedDB blockstore as a client cache.
+ */
+export class MixedBlockstore extends BaseBlockstore {
+  readonly #apiHost: string | undefined;
+  readonly #idbStore: IDBBlockstore;
+
+  constructor(apiHost?: string) {
+    super();
+
+    this.#apiHost = apiHost;
+    this.#idbStore = new IDBBlockstore(storeName());
+  }
+
+  async open() {
+    await this.#idbStore.open();
+  }
+
+  url(apiHost: string, cid?: CID) {
+    const path = cid ? cid.toString() : '';
+    return `${apiHost}/api/file/${path}`;
+  }
+
+  // Blockstore implementation
+  override async delete(key: CID): Promise<void> {
+    await this.#idbStore.delete(key);
+
+    if (this.#apiHost) {
+      await fetch(this.url(this.#apiHost, key), {
+        method: 'DELETE',
+      });
+    }
+  }
+
+  override async get(key: CID): Promise<Uint8Array> {
+    if (await this.#idbStore.has(key)) {
+      return await this.#idbStore.get(key);
+    }
+
+    if (this.#apiHost) {
+      return await fetch(this.url(this.#apiHost, key), {
+        method: 'GET',
+      })
+        .then((r) => r.arrayBuffer())
+        .then((r) => new Uint8Array(r));
+    }
+
+    return await this.#idbStore.get(key);
+  }
+
+  override async has(key: CID): Promise<boolean> {
+    if (await this.#idbStore.has(key)) {
+      return true;
+    }
+
+    if (this.#apiHost) {
+      return await fetch(this.url(this.#apiHost, key), {
+        method: 'HEAD',
+      }).then((r) => r.ok);
+    }
+
+    return false;
+  }
+
+  override async put(key: CID, val: Uint8Array): Promise<CID> {
+    await this.#idbStore.put(key, val);
+
+    if (this.#apiHost) {
+      await fetch(this.url(this.#apiHost, key), {
+        method: 'POST',
+        body: val,
+      });
+    }
+
+    return key;
+  }
+
+  async destroy(): Promise<void> {
+    await this.#idbStore.destroy();
+  }
+}
--- a/packages/plugins/experimental/plugin-wnfs/src/common.ts
+++ b/packages/plugins/experimental/plugin-wnfs/src/common.ts
@@ -1,0 +1,84 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import type { Blockstore } from 'interface-blockstore';
+import { CID } from 'multiformats/cid';
+import { sha256 } from 'multiformats/hashes/sha2';
+import { type BlockStore as WnfsBlockStore } from 'wnfs';
+
+import * as crypto from '@dxos/crypto';
+import { type Space } from '@dxos/react-client/echo';
+
+//
+// PATH
+//
+
+export const filePath = (fileName: string, space: Space) => {
+  return ['spaces', space.properties.id, 'files', fileName];
+};
+
+//
+// RNG
+//
+
+export class Rng {
+  randomBytes(count: number): Uint8Array {
+    return new Uint8Array(crypto.randomBytes(count));
+  }
+}
+
+//
+// STORE
+//
+
+export const store = (blockstore: Blockstore): WnfsBlockStore => {
+  const getBlock = async (cid: Uint8Array): Promise<Uint8Array | undefined> => {
+    const decodedCid = CID.decode(cid);
+    return await blockstore.get(decodedCid);
+  };
+
+  const hasBlock = async (cid: Uint8Array): Promise<boolean> => {
+    const decodedCid = CID.decode(cid);
+    return await blockstore.has(decodedCid);
+  };
+
+  const putBlockKeyed = async (cid: Uint8Array, bytes: Uint8Array): Promise<void> => {
+    const decodedCid = CID.decode(cid);
+    await blockstore.put(decodedCid, bytes);
+  };
+
+  // Don't hash blocks with the rs-wnfs default Blake 3, sha256 has better support.
+  const putBlock = async (bytes: Uint8Array, codec: number): Promise<Uint8Array> => {
+    const hash = await sha256.digest(bytes);
+    const cid = CID.create(1, codec, hash);
+    await blockstore.put(cid, bytes);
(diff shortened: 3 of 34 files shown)


packages/sdk/config/src/plugin/definitions.ts
```diff
@@ -25,8 +25,8 @@ export const definitions = ({
   env,
 }: ConfigPluginOpts) => {
   const KEYS_TO_FILE = {
-    __CONFIG_DEFAULTS__: configPath ?? resolve(CWD, 'dx.yml'),
-    __CONFIG_ENVS__: envPath ?? resolve(CWD, 'dx-env.yml'),
+    __CONFIG_DEFAULTS__: configPath && configPath.length ? configPath : resolve(CWD, 'dx.yml'),
+    __CONFIG_ENVS__: envPath && envPath.length ? envPath : resolve(CWD, 'dx-env.yml'),
```
icidasset, line 29: Environment wouldn't load for me without this.

packages/ui/react-ui-editor/src/extensions/markdown/image.ts
```diff
@@ -63,6 +63,9 @@ const buildDecorations = (from: number, to: number, state: EditorState) => {
         if (urlNode) {
           const hide = state.readOnly || cursor < node.from || cursor > node.to;
           const url = state.sliceDoc(urlNode.from, urlNode.to);
+          if (url.match(/^https?:\/\//) === null) {
```
icidasset, line 66: Avoid rendering non-HTTP urls to images.

packages/plugins/experimental/plugin-wnfs/src/load.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { Blockstore } from 'interface-blockstore';
+import { CID } from 'multiformats';
+import * as Uint8Arrays from 'uint8arrays';
+import { AccessKey, PrivateForest, PrivateDirectory, PrivateNode } from 'wnfs';
+
+import { type Space } from '@dxos/react-client/echo';
+
+import { Rng, store } from './common';
+
+// 🚀
```
wittjosiah, line 14: prefer descriptive comments, I'm not sure this communicates intent clearly enough

packages/plugins/experimental/plugin-wnfs/src/meta.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { pluginMeta } from '@dxos/app-framework';
+
+export const WNFS_PLUGIN = 'dxos.org/plugin/WNFS';
+
+export default pluginMeta({
+  id: WNFS_PLUGIN,
+  name: 'WNFS',
+  description: 'Upload & view files with WNFS.',
+  tags: ['experimental'],
+  iconComponent: (props: IconProps) => <FileCloud {...props} />,
```
wittjosiah, line 17: if you pull from `main` these were recently updated to be icon strings which index a sprite sheet. This will need to align with that for the icon to work now

packages/sdk/config/src/plugin/definitions.ts
wittjosiah: interesting... wonder why that's not come up before. I think it could be shortened to `configPath?.length` but seems reasonable to me

packages/ui/react-ui-editor/src/extensions/markdown/image.ts
wittjosiah: Could you leave a comment here to explain why this is needed?

wittjosiah (commented):
looking good, nice work!

one thing I couldn't find was where the `FileType` objects were being created. I was looking because I was thinking it would probably be good to include some sort of meta key on them to indicate that they were associated with the blob store service

packages/plugins/experimental/plugin-wnfs/src/components/index.ts
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import React from 'react';
+
+// Lazily load components for content surfaces.
```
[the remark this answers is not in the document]
icidasset: Yeah, that's where I started months ago.
I have no idea how those components work 😅 
Let me know what to do there and I'll make another PR for that.

packages/sdk/config/src/plugin/definitions.ts
icidasset: Hmm.. I'm guessing it's because I'm using this shell that almost no one else uses (elvish).

icidasset:
Thanks @wittjosiah! I've applied your feedback. Also not sure where the actual `FileType` objects are created, I vaguely remember something about the uploader interface of composer/markdown-plugin? 🤷 Definitely not in this new plugin though.

packages/plugins/experimental/plugin-wnfs/src/blockstore.ts
```diff
@@ -0,0 +1,82 @@
...
+ * and which uses an indexedDB blockstore as a client cache.
+ */
+class MixedBlockstore extends BaseBlockstore {
+  readonly #apiHost: string;
+  readonly #idbStore: IDBBlockstore;
+
+  constructor() {
+    super();
+
+    this.#apiHost = 'http://localhost:8787';
```
icidasset, line 26: @wittjosiah This is supposed to target the Blob service, is there a production instance yet?

icidasset: @richburdon Any thoughts?

packages/plugins/experimental/plugin-wnfs/src/WnfsPlugin.tsx
```diff
@@ -0,0 +1,118 @@
...
+            true,
+            new Uint8Array(await file.arrayBuffer()),
+            new Date(),
+            forest,
+            wnfsStore,
+            new Rng(),
+          );
+
+          const [_, updatedForest] = await result.rootDir.store(result.forest, wnfsStore, new Rng());
+
```
richburdon, line 65: i think group these lines (no blank lines) since they are closely related. perhaps adding a comment to orient people unfamiliar with wnfs.

packages/plugins/experimental/plugin-wnfs/src/blockstore.ts
richburdon: obtain from the config object (useConfig).
add `runtime.services.blobstore.endpoint`

packages/plugins/experimental/plugin-wnfs/src/common.ts
```diff
@@ -0,0 +1,81 @@
...
+    const decodedCid = CID.decode(cid);
+    return await blockstore.has(decodedCid);
+  };
+
+  const putBlockKeyed = async (cid: Uint8Array, bytes: Uint8Array): Promise<void> => {
+    const decodedCid = CID.decode(cid);
+    await blockstore.put(decodedCid, bytes);
+  };
+
+  // Don't hash blocks with the rs-wnfs default Blake 3, sha256 has better support
```
richburdon, line 48: our comments standard is to put a period at the end of the sentence (to make sure it isn't truncated).

packages/plugins/experimental/plugin-wnfs/src/common.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import type { Blockstore } from 'interface-blockstore';
+import { CID } from 'multiformats/cid';
+import { sha256 } from 'multiformats/hashes/sha2';
+import { type BlockStore as WnfsBlockStore } from 'wnfs';
+
+import * as crypto from '@dxos/crypto';
+import { type Space } from '@dxos/react-client/echo';
+
+// PATH
```
richburdon, line 13: nit

```
//
// Path
//
```

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
```diff
@@ -0,0 +1,23 @@
...
+import React, { type FC } from 'react';
+
+import { Main } from '@dxos/react-ui';
+import { baseSurface, topbarBlockPaddingStart, fixedInsetFlexLayout } from '@dxos/react-ui-theme';
+
+import { FilePreview } from './FilePreview';
+import { type FileType } from '../types';
+
+const FileMain: FC<{ file: FileType }> = ({ file }) => {
+  const url = 'TODO';
```
richburdon, line 14: ?

packages/plugins/experimental/plugin-wnfs/src/components/FilePreview.tsx
```diff
@@ -0,0 +1,24 @@
...
+export type FilePreviewProps = {
+  type: string;
+  url: string;
+  className?: string;
+};
+
+/**
+ * File/content preview iframe.
+ */
+export const FilePreview: FC<FilePreviewProps> = ({ type, url, className }) => {
```
richburdon, line 18: you might be copying one of our antipatterns, since we've done this in the past, but now prefer

FilePreview = ({ ... }: FilePreviewProps) =>

packages/plugins/experimental/plugin-wnfs/src/components/FileSlide.tsx
```diff
@@ -0,0 +1,20 @@
...
+// Copyright 2023 DXOS.org
+//
+
+import React, { type FC } from 'react';
+
+import { FilePreview } from './FilePreview';
+import { type FileType } from '../types';
+
+const FileSlide: FC<{ file: FileType; cover?: boolean }> = ({ file, cover }) => {
+  const url = 'TODO';
```
richburdon, line 11: ?

packages/plugins/experimental/plugin-wnfs/src/components/index.ts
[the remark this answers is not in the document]
richburdon: this is fine.

packages/plugins/experimental/plugin-wnfs/src/extensions/image.ts
```diff
@@ -0,0 +1,247 @@
...
+      let eff;
+      for (const effect of tr.effects) {
+        if (effect.is(stateEffect)) {
+          eff = effect.value;
+        }
+      }
+      if (eff === undefined) {
+        return value;
+      }
+      return value.map(tr.changes).update({
```
richburdon, line 53: i think this would be clearly (and more compact) if you just put this inline inside the for/if loop.

packages/plugins/experimental/plugin-wnfs/src/extensions/image.ts
```diff
@@ -0,0 +1,247 @@
...
+        throw new Error(err);
+      });
+
+    const update = (update: ViewUpdate) => {
+      if (!update.docChanged) {
+        return;
+      }
+
+      const { from, to } = changedRange(update);
+
```
richburdon, line 93: remove blank line (group statements by context)

packages/plugins/experimental/plugin-wnfs/src/extensions/image.ts
```diff
@@ -0,0 +1,247 @@
...
+          view.dispatch({
+            effects: stateEffect.of({
+              decorations,
+              from,
+              to,
+            }),
+          });
+        })
+        .catch((err) => {
+          throw new Error(err);
```
richburdon, line 110: log.catch(err)?

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
```diff
@@ -0,0 +1,23 @@
+//
```
richburdon, line 1: is it possible to create a unit test or a react story to demo upload and preview of images? e.g., look at TextEdtitor.stories.tsx (low-level) or mermaid.stories.tsx (which is another plugin).

richburdon (commented):
Really nice!

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
icidasset: I haven't worked on these components, was planning to do that in a separate PR. I'd also like some input on how to proceed with these components.

packages/plugins/experimental/plugin-wnfs/src/components/FileSlide.tsx
icidasset: Same comment as https://github.com/dxos/dxos/pull/6636#discussion_r1796784021

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
icidasset: Sure yeah, I'll add that to this PR. I assume without involving the blob service right?

icidasset: @richburdon Added one.
Screenshot 2024-10-24 at 19 03 01

Renders the uploaded images.

icidasset:
Rebased, hopefully interpreted the changes on `main` correctly.
Other changes:
- Added story component.
- Used a CID as the file name of the uploaded file, in case multiple images with the same name are uploaded (but with different contents).

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
icidasset: @richburdon See the comment above.

wittjosiah:
> I got an error of how I used useConfig in the plugin, so I used parseClientPlugin instead. Is that the right approach @wittjosiah ?

Yeah that looks right. `useConfig` is a react hook and `ready` runs outside of the realm of react so that would be why it wouldn't work.

packages/plugins/experimental/plugin-wnfs/src/common.ts
```diff
@@ -0,0 +1,85 @@
...
+  randomBytes(count: number): Uint8Array {
+    return new Uint8Array(crypto.randomBytes(count));
+  }
+}
+
+//
+// STORE
+//
+
+export const store = (blockstore: Blockstore): WnfsBlockStore => {
```
richburdon, line 35: why not make this a class?

packages/plugins/experimental/plugin-wnfs/src/components/FileCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+            </Card.Menu>
+          </Card.Header>
+          <Card.Media src={url} />
+        </Card.Root>
+      </div>
+    );
+  },
+);
+
+export default FileCard;
```
richburdon, line 45: in general we prefer explicit `export const`

packages/plugins/experimental/plugin-wnfs/src/components/FileMain.tsx
richburdon: ok we're doing a rewrite of mosaic, so i think fine to leave this.

packages/plugins/experimental/plugin-wnfs/src/components/FilePreview.tsx
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import React from 'react';
+
+import { mx } from '@dxos/react-ui-theme';
+
+export type FilePreviewProps = {
+  type: string;
+  url: string;
+  className?: string;
```
richburdon, line 12: `type FilePreviewProps = ThemedClassName<{ type, url }>`

packages/plugins/experimental/plugin-wnfs/src/extensions/image.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <strong className='text-green-600'>Select an image file to add to WNFS and render it.</strong>
+          </p>
+        )}
+      </div>
+    </>
+  );
+};
+
+export default {
+  title: 'plugin-wnfs/image',
```
richburdon, line 133: this has changed recently: 'pluings/plugin-wnfs/image'

packages/plugins/experimental/plugin-wnfs/src/load.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as Uint8Arrays from 'uint8arrays';
+import { AccessKey, PrivateForest, PrivateDirectory, PrivateNode } from 'wnfs';
+
+import { type Space } from '@dxos/react-client/echo';
+
+import { Rng, store } from './common';
+
+// LOAD
+// ----
+
```
richburdon, line 16:
```
//
// Load
//
```

(as above)

packages/plugins/experimental/plugin-wnfs/src/load.ts
```diff
@@ -0,0 +1,63 @@
...
+// PRIVATE
+// -------
+
+const createWnfsDir = async (blockstore: Blockstore, space: Space) => {
+  const wnfsStore = store(blockstore);
+
+  const rng = new Rng();
+  const forest = new PrivateForest(rng);
+
```
richburdon, line 32: remove black links to logically group related statements?

richburdon (approved):
Great work steven!

packages/plugins/experimental/plugin-wnfs/src/common.ts
icidasset: Could do that, but thought a function was more appropriate seeing how I use the resulting `WnfsBlockStore`. Pattern goes like so:
```ts
const blockstore: Blockstore = makeIndexedDbBlockstore()

await wnfsInstance.write(
  ...,
  common.store(blockstore)
  // Feels a bit nicer to write, as opposed to:
  new common.Store(blockstore)
  // And using a class might give people more the idea you're creating a new store (which isn't the case)
)
```

PR #6638, opened 2024-05-10
feat(composer): Status bar

### LINKED ISSUE #5384: [composer] feedback button for users to send feedback to the team
### Motivation

Make it easy for end-users to send us feedback.

### Finish Line

- button in the UI that brings up a form so that users can send feedback
- could just be a `mailto:` for now
- or send an email or post a message somewhere
- give them the option of including their public key so that we can look for their telemetry (errors, etc)

### LINKED ISSUE #6519: [composer] join discord button
let's display a subtle join discord button next to the submit feedback / bug button

ZaymonFC:
Implements a new plugin and UI to show a status bar in composer.

- Resolves #6519 
- Resolves #5384


packages/apps/plugins/plugin-status-bar/src/translations.ts
```diff
@@ -0,0 +1,42 @@
...
+    'feedback panel title': 'Feedback',
+    'name label': 'Name',
+    'name placeholder': 'Full name',
+    'email input label': 'Email',
+    'email input placeholder': 'your-email@example.com',
+    'feedback text area label': 'Feedback',
+    'feedback text area placeholder': 'tell us what you think',
+    'send feedback label': 'Send Feedback',
+  }),
+};
```
ZaymonFC, line 40: TODO: Workshop the copy.

packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
```diff
@@ -177,6 +177,10 @@ export const MainLayout = ({
           </div>
         )}
 
+        <div className='fixed bottom-0 left-0 right-0'>
```
thure, line 180:
```suggestion
        <div className='fixed block-end-0 inset-inline-0'>
```

packages/apps/plugins/plugin-status-bar/src/components/FeedbackForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const { errors, handleSubmit, canSubmit, touched, getInputProps } = useForm<FeedbackFormState>({
+    initialValues,
+    validate: (values) => validate(FeedbackFormSchema, values),
+    onSubmit: (values) => onSubmit(values),
+  });
+
+  return (
+    <div role='form' className='p-3 flex flex-col gap-2'>
+      <Section className='space-b-1'>
+        <Input.Root>
```
thure, line 67: The `Input` set of components handles associating accessibility id matching… I think these overrides will still work, but you can just give `id` to `Input.Root` and it will generally handle everything within its subtree.

thure: 'everthing' meaning it will set the appropriate `id`, `htmlFor`, `aria-invalid`, `aria-errormessage`, `aria-labelledby`, and `aria-describedby` attributes on all the `Input` subcomponents within its subtree.

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+    >
+      {children}
+    </button>
+  ),
+);
+
+type StatusBarItemProps = ThemedClassName<{ children: ReactNode }>;
+
+const StatusBarItem = forwardRef<HTMLDivElement, StatusBarItemProps>(({ classNames, children }, forwardedRef) => (
+  <div className={mx('flex items-center gap-2 p-1 px-2 rounded-sm select-none', classNames)} ref={forwardedRef}>
```
thure, line 38: Lots of `div`s here that have no `role`, this will be fairly inaccessible as it is currently.

I suspect the status bar’s content is probably a [`menubar`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menubar_role) with items (some clickable, some not) within.

packages/apps/plugins/plugin-status-bar/src/components/StatusBarImpl.tsx
```diff
@@ -0,0 +1,45 @@
...
+              <StatusBar.Text classNames='hidden sm:block'>Feedback</StatusBar.Text>
+            </StatusBar.Button>
+          </Popover.Trigger>
+          {/* TODO(Zan): Configure this? */}
+          <a href='https://discord.gg/zsxWrKjteV' target='_blank' rel='noopener noreferrer'>
+            <StatusBar.Button>
+              <DiscordLogo />
+              <StatusBar.Text classNames='hidden sm:block'>Discord</StatusBar.Text>
+            </StatusBar.Button>
+          </a>
```
thure, line 32: `Statusbar.Button` should support `asChild` so these don’t need to nest like this.

packages/apps/plugins/plugin-status-bar/src/translations.ts
```diff
@@ -0,0 +1,42 @@
...
+// -- helpers.
+const entry = (translation: RequiredTranslation): TranslationEntry => ({ [STATUS_BAR_PLUGIN]: translation });
+export const mkTranslation = (t: ReturnType<typeof useTranslation>['t']) => (key: StatusBarTranslationKeys) => t(key);
+
+const translations: Record<Language, TranslationEntry> = {
+  'en-US': entry({
+    'plugin name': 'Status Bar',
+    'feedback panel title': 'Feedback',
+    'name label': 'Name',
+    'name placeholder': 'Full name',
```
thure, line 33: Less scary to ask for “preferred name”

packages/apps/plugins/plugin-status-bar/src/translations.ts
```diff
@@ -0,0 +1,42 @@
...
+    'email input label': 'Email',
+    'email input placeholder': 'your-email@example.com',
```
thure, line 35: placeholders should generally align in purpose — this is an example, whereas the name field is an expanded label… fwiw I’m not sure what the best approach is, really… maybe can just leave this alone.

packages/apps/plugins/plugin-status-bar/src/translations.ts
```diff
@@ -0,0 +1,42 @@
...
+    'feedback text area label': 'Feedback',
+    'feedback text area placeholder': 'tell us what you think',
```
thure, line 37: Capitalization should align, generally sentence case is best.

thure (changes requested):
I think the `Input` id linking business should probably be fixed before merging this (with my apologies this wasn’t better documented), otherwise it’s looking good ✨

packages/apps/plugins/plugin-observability/src/ObservabilityPlugin.tsx
```diff
@@ -193,6 +195,12 @@ export const ObservabilityPlugin = (options: {
               observability.event(event);
               return { data: event };
             }
+
+            case ObservabilityAction.CAPTURE_USER_FEEDBACK: {
+              const feedback = intent.data as UserFeedback;
+
+              observability.captureUserFeedback(feedback.email, feedback.name, feedback.message);
```
wittjosiah, line 202: make sure to return something non-null here otherwise the intent plugin will interpret this as unhandled

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.stories.tsx
wittjosiah, line 1: yay stories :)

packages/apps/plugins/plugin-status-bar/src/hooks/useForm.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import { type ValidationError } from '../util';
+
+interface FormOptions<T> {
+  initialValues: T;
+  validate: (values: any) => ValidationError[] | undefined;
+  onSubmit: (values: T) => void;
+}
+
+export const useForm = <T extends object>({ initialValues, validate, onSubmit }: FormOptions<T>) => {
```
wittjosiah, line 14: I think think is probably generally useful, I'd throw a `TODO: Factor out.` on here.

Might also be useful to have a jsdoc explaining intended usage.

packages/apps/plugins/plugin-status-bar/src/meta.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { pluginMeta } from '@dxos/app-framework';
+
+export const STATUS_BAR_PLUGIN = 'dxos.org/plugin/status-bar';
+
+export default pluginMeta({
+  id: STATUS_BAR_PLUGIN,
+  name: 'Status Bar',
+  description: 'Display a bar with status and actions.',
+  tags: ['experimental'],
```
wittjosiah, line 16: if we're turning this on always I wouldn't bother with this

packages/apps/plugins/plugin-status-bar/src/components/StatusBarImpl.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { DiscordLogo, Lightning, PaperPlaneTilt } from '@phosphor-icons/react';
+import React from 'react';
+
+import { Popover } from '@dxos/react-ui';
+
+import { FeedbackForm } from './FeedbackForm';
+import { StatusBar } from './StatusBar';
+
+export const StatusBarImpl = () => {
```
wittjosiah, line 13: Interestingly I feel like elsewhere I've inverted these name (impl for the internal "pure" version and then plain name for the exported general usage component)

packages/apps/plugins/plugin-observability/src/ObservabilityPlugin.tsx
ZaymonFC: Do you have a go to return value to just indicate 'handled'?

wittjosiah: The default I've done elsewhere is just `return { data: true };`

packages/apps/plugins/plugin-status-bar/src/components/StatusBarImpl.tsx
ZaymonFC: Good point. It's tricky when you have components / plugins sharing names. I'm vacillating between both options.

packages/apps/plugins/plugin-status-bar/src/components/FeedbackForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+type FeedbackFormState = S.Schema.Type<typeof FeedbackFormSchema>;
+
+const initialValues: FeedbackFormState = { name: '', email: '', message: '' };
+
+const Section: FC<PropsWithChildren & { className?: string }> = ({ children, className }) => (
+  <div role='none' className={mx(className)}>
+    {children}
+  </div>
+);
```
thure, line 43: Essentially equivalent to a literal `<div role='none'>` except it prevents refs or other attributes even though the type accepts them… I think it could be removed and consumers could just use the literal, or this could render a [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) or provide `role='group'`…

thure: Otherwise, either its type should be narrowed to `PropsWithChildren<ThemedClassName<{}>>`, or it should spread its props (and possibly forward refs).

thure: As a general rule, the prop that forwards to `mx` should be called `classNames` (with an ‘s’), since in theory `className` should override anything here.

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+    role='menuitem'
+    className={mx('flex items-center gap-2 p-1 px-2 rounded-sm select-none', classNames)}
+    ref={forwardedRef}
+  >
+    {children}
+  </div>
+));
+type StatusBarContainerProps = ThemedClassName<{ children: ReactNode }>;
+
+// TODO(zan): tabable group with tabster
```
thure, line 56: `const groupAttrs = useArrowNavigationGroup({ axis: 'horizontal' })`

thure (approved):
one minor nit and a suggestion, but otherwise lgtm!

packages/apps/plugins/plugin-status-bar/src/components/FeedbackForm.tsx
ZaymonFC: You're right!

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
ZaymonFC: Nice. Done

Issue #6145, opened 2024-03-29
[Composer] `cmd-k` menu should have the top edge anchored


ZaymonFC:
I really struggle to use `cmd-k` because of the height of the command panel changing and content moving up and down on the screen.

My eyes are naturally anchored to the text input and the text input shouldn't move as you type.

I propose we anchor the menu to the top edge of the screen with an offset.

Github's implementation of this works well.
<img width="1274" alt="image" src="https://github.com/dxos/dxos/assets/12402727/30e97058-9a5a-456a-852e-93770745ec38">

PR #6640, opened 2024-05-10
fix(composer): Anchor settings dialog to prevent content jumps

### LINKED ISSUE #6488: [composer] settings should anchor to viewport to prevent bouncing when switching between panels

ZaymonFC:
Pretty simple fix since we were able to leverage the work done for #6145

- Resolves #6488

https://github.com/dxos/dxos/assets/12402727/6c501fc4-60b6-4237-8383-0e9d8454aff7

### DIFF
--- a/packages/apps/plugins/plugin-settings/src/components/SettingsDialog.tsx
+++ b/packages/apps/plugins/plugin-settings/src/components/SettingsDialog.tsx
@@ -35,7 +35,7 @@ export const SettingsDialog = ({
     .sort(({ name: a }, { name: b }) => a?.localeCompare(b ?? '') ?? 0);
 
   return (
-    <Dialog.Content classNames={['h-auto h-max-full md:max-is-[40rem] overflow-hidden']}>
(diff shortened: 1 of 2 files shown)


packages/apps/plugins/plugin-settings/src/components/SettingsDialog.tsx
```diff
@@ -35,7 +35,7 @@ export const SettingsDialog = ({
     .sort(({ name: a }, { name: b }) => a?.localeCompare(b ?? '') ?? 0);
 
   return (
-    <Dialog.Content classNames={['h-full md:max-is-[40rem] overflow-hidden']}>
+    <Dialog.Content classNames={['h-auto h-max-full md:max-is-[40rem] overflow-hidden']}>
```
thure, line 38:
```suggestion
    <Dialog.Content classNames={['bs-content max-bs-full md:max-is-[40rem] overflow-hidden']}>
```
Whether this will limit its size depends on the layout (I’m forgetting), a `max-bs` rule might be unnecessary or have no effect.

thure (approved):
lgtm! just a couple nits to consider

PR #6643, opened 2024-05-12
feat: Performance monitor

richburdon:
## Diagnostics Panel

- [x] Prototype in testbench
- [ ] Factor out and include in Composer (via plugin for status bar)

Screenshot 2024-05-13 at 9 37 08 PM


packages/sdk/shell/src/testing/shell-manager.ts
```diff
@@ -59,22 +61,22 @@ export class ShellManager extends ScopedShellManager {
 
   async resetDevice() {
     const handleDialog = async (dialog: Dialog) => {
-      if (dialog.type() === 'confirm') {
+      if (dialog.type() === RESET) {
```
dmaretskyi, line 64: im not sure those two constants are related: one is dialog type and the other one is what the user is expected to type into the input

PR #6662, opened 2024-05-14
feat: Attention & relatedness

thure:
This PR implements the attention and relatedness concepts as designed.

https://github.com/dxos/dxos/assets/855039/d9deaf93-45f3-4b76-af53-864f81f9f42f

Remaining:
- [x] Ensure this still works with non-Deck layout
- [x] DocumentMain should render attention like in stacks instead of always


packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -182,7 +183,12 @@ export const DeckPlugin = ({
       },
       context: (props: PropsWithChildren) => {
         return (
-          <AttentionProvider attended={attention.attended}>
+          <AttentionProvider
+            attended={attention.attended}
+            onChangeAttend={(nextAttended) => {
+              attention.attended = nextAttended;
+            }}
```
ZaymonFC, line 190: Consider updating this through an intent?

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -304,10 +298,11 @@ export const DeckLayout = ({
               .map((id, index, main) => {
                 const node = resolveNodeFromSlug(graph, id);
                 const part = ['main', index, main.length] satisfies PartIdentifier;
+                const attendableAttrs = useAttendable(id);
                 return (
-                  <Deck.Plank key={id}>
+                  <Deck.Plank key={id} {...attendableAttrs}>
```
ZaymonFC, line 303: Nice. I'm liking this data annotation approach.

packages/ui/react-ui-deck/src/components/Attention/AttentionProvider.tsx
```diff
@@ -6,22 +6,73 @@
...
+const getAttendables = (cursor: FocusEvent['target'], acc: string[] = []): string[] => {
+  const closestAttendable = cursor.closest('[data-attendable-id]');
+  if (!closestAttendable) {
+    return acc;
+  } else {
+    acc.push(closestAttendable.getAttribute('data-attendable-id')!);
+    // (attempt tail recursion)
+    return !closestAttendable.parentElement ? acc : getAttendables(closestAttendable.parentElement, acc);
+  }
+};
```
ZaymonFC, line 45: Can you add a doc comment explaining the semantics of this? Especially the behaviour of `cursor.closest(['data-attendable-id'])`.

Just so I get this clear in my head, we're searching up the DOM and collecting all of the attendable id's along the way. How can you supply an `Element` in place of a `FocusEvent['target']`?

packages/apps/plugins/plugin-layout/src/LayoutPlugin.tsx
```diff
@@ -279,7 +279,10 @@ export const LayoutPlugin = ({
                 },
               };
 
-        const attended = useMemo(() => new Set(location.active ? [location.active] : []), [location.active]);
+        const attended = useMemo(
+          () => new Set(location.active ? [firstMainId(location.active)] : []),
+          [location.active],
+        );
```
ZaymonFC, line 285: Would attention / attended state be better suited to being declared as a signal alongside `settings`, `layout` and `location`?

ZaymonFC (approved):
I'm liking the shape of this. A few questions and suggestions.

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
thure: I’m not sure it’s quite the same kind of event — this is connecting a UI component’s controllable state to a plugin’s `provides` state, which other plugins are free to access and subscribe to as a signal.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
thure: Yeah, I’m enjoying taking tabster’s lead on this, seems productive.

packages/apps/plugins/plugin-layout/src/LayoutPlugin.tsx
thure: Yes, I’m just now getting to LayoutPlugin’s implementation, this was just a formatting change.

packages/ui/react-ui-deck/src/components/Attention/AttentionProvider.tsx
thure: > How can you supply an Element in place of a FocusEvent['target']?

This is demonstrated in line 43, `closestAttendable.parentElement` satisfies the type. Will add a JSDoc to explain the rest, certainly.

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
wittjosiah: I'll say that I've trended away from writing directly to signals even if they're exposed by provides. I've sometimes even made them readonly in the provides type to discourage that. It helps when changing any sort of logic around how the state is updated, there's just one place to do it.

For example, we used to update the `active` state like this, but that became problematic when I wanted to add `previous`.

thure: I suppose for this particular feature I’m of the (non-)opinion that this should change when we can demonstrate significant value can be delivered by changing it.

Currently, `attention` shares the same implementation as `location` in both `LayoutPlugin` and `DeckPlugin`, so when we decide these are better implemented another way, we can change both (or indeed all `provides`-related calls to `echo-schema`’s `create`) in one go.

thure: fwiw this is the only place `attended` should be updated. Maybe it should be read-only to consumers since the way that they “write” to this state is to `const attrs = useAttendable(id)` and `<div {...atts}>`. Do either of you anticipate any specific harm from leaving the current implementation as-is?

wittjosiah: I'm fine with leaving it as is, I'd maybe suggest adding a TODO as a reminder that if this in undergoing changes later that it probably warrants an intent.

PR #6707, opened 2024-05-16
fix: devtools perf panels

richburdon:


packages/core/echo/echo-protocol/src/reference.ts
```diff
@@ -21,6 +21,7 @@ export class Reference {
 
   // prettier-ignore
   constructor(
```
dmaretskyi, line 23: I think we might wanna just encode refrences as URIs, what do you think?

PR #6715, opened 2024-05-17
fix: query objects with static and persisted schemas

jessmartin:
### Background

I ran into an issue querying for objects with a statically-defined schema and a persisted schema in [Contacts App](https://github.com/dxos/contacts-app/pull/3). 

### Expected Behavior

I believe that all objects that match a `typename` should be returned by the query, regardless of whether the object was created with a statically-defined schema or a persisted schema.

The failing test demonstrates the expected behavior.

As an aside, it seems that we have entangled the ideas of "persisted" and "dynamic" schemas. A statically-defined schema should be able to be persisted without becoming dynamic. Persistence just implies I want to share objects and schemas with other peers without depending on them to have a statically-defined type definition in their codebase. It does not imply that I want to modify that schema at runtime (make it dynamic).

### DIFF
--- a/packages/core/echo/echo-db/src/query/filter.test.ts
+++ b/packages/core/echo/echo-db/src/query/filter.test.ts
@@ -2,11 +2,10 @@
 // Copyright 2023 DXOS.org
 //
 
-import * as S from '@effect/schema/Schema';
 import { expect } from 'chai';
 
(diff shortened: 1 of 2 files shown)


jessmartin:
@ZaymonFC here's a failing unit test that demonstrates the problem we ran into between Contacts app and Table.

PR #6727, opened 2024-05-19
feat: Email sync demo

richburdon:
Email end-to-end.

https://github.com/dxos/dxos/assets/3523355/e9be952c-5152-41fe-af40-301eb04ca6b1


packages/core/echo/echo-schema/src/getter.ts
```diff
@@ -48,17 +50,21 @@ export const getMeta = <T extends {}>(obj: T): ObjectMeta => {
...
 export const isDeleted = <T extends {}>(obj: T): boolean => {
   const proxyHandlerSlot = getProxyHandlerSlot(obj);
-  return proxyHandlerSlot.handler?.isObjectDeleted(obj) ?? false;
+  return proxyHandlerSlot.handler?.isDeleted(obj) ?? false;
 };
 
+// TODO(burdon): Replace most uses with getTypename (and rename itemId property).
 export const getType = <T extends {}>(obj: T | undefined): Reference | undefined => getTypeReference(getSchema(obj));
 
+export const getTypename = <T extends {}>(obj: T | undefined): string | undefined =>
```
dmaretskyi, line 59: not against this, but why not stick with `getType(obj).name`?

packages/sdk/client-protocol/src/service-definitions.ts
```diff
@@ -52,16 +51,16 @@ export type ClientServices = {
...
+   * The connection to the services provider was terminated.
    * This should fire if the services disconnect unexpectedly or during a client reset.
    */
   closed: Event<Error | undefined>;
   descriptors: ServiceBundle<ClientServices>;
   services: Partial<ClientServices>;
 
-  open(ctx: Context): Promise<void>;
-  close(ctx: Context): Promise<void>;
+  // TODO(burdon): Should take context from parent?
```
dmaretskyi, line 61: 👍

packages/apps/plugins/plugin-debug/src/components/DebugSpace.tsx
```diff
@@ -46,18 +53,21 @@ const DebugSpace: FC<{ space: Space; onAddObjects?: (objects: ReactiveObject<any
...
-        return space.key.toHex().startsWith(key);
-      }),
-    );
-  };
 
-  useEffect(() => {
-    void handleRefresh();
-  }, [space]);
+  const [update, handleUpdate] = useRefresh();
+  useAsyncEffect(
```
wittjosiah, line 58: we've been avoiding this for a while, I'm surprised it's not marked as deprecated

wittjosiah: generally we've been using

```
useEffect(() => {
  const timeout = setTimeout(async () => {
    // your async code here
  });

  return () => clearTimeout(timeout);
}, []);
```

packages/apps/plugins/plugin-inbox/src/components/Email/Mailbox.tsx
```diff
@@ -49,31 +49,34 @@ const Mailbox = ({ mailbox, options = {} }: MailboxProps) => {
     };
   }, [selected]);
 
-  const messages = [...mailbox.messages]
-    .filter(nonNullable)
+  const messages = [...(mailbox.messages ?? [])]
+    .filter(nonNullable) // TODO(burdon): Why is this necessary?
```
wittjosiah, line 53: iirc @yarolegovich said it shouldn't be but I've found in the past with stacks that sometimes list of references sometimes come back without all of the resolved. I'm not sure if this is a bug or if it's related to deleted objects but we decided to leave the new type this way to force the undefined to be handled because then you won't get unexpected results.

packages/common/async/src/events.ts
```diff
@@ -71,6 +71,7 @@ export type ListenerOptions = {
  * 6. Removes the cases where event names intersect when used in cases with inheritance.
  * 7. Remove the need to namespace events when developing a class with events that will be used as a base-class.
  */
+// TODO(burdon): Rename EventSink? (Event is a built-in type).
```
wittjosiah, line 74: it's more of an event source?

packages/sdk/client/src/echo/agent-query-source-provider.ts
```diff
@@ -16,6 +16,8 @@ import { type EchoObject as EchoObjectProto } from '@dxos/protocols/proto/dxos/e
 import { type QueryRequest, type QueryResponse } from '@dxos/protocols/proto/dxos/echo/query';
 import { type GossipMessage } from '@dxos/protocols/proto/dxos/mesh/teleport/gossip';
 
+const ERR_CLOSING = new Error();
```
wittjosiah, line 19: no message?

packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -25,19 +29,14 @@ export const fromHost = async (
   config = new Config(),
   params?: ClientServicesHostParams,
   observabilityGroup?: string,
-  signalTelemetryEnabled?: boolean,
+  signalTelemetryEnabled?: boolean, // TODO(burdon): Not used?
```
wittjosiah, line 32: cc @nferch, potentially should be forwarded into `setupNetworking`

packages/core/echo/echo-schema/src/getter.ts
richburdon: i didn't want to rename `itemId => name` since i don't know what that would break.
also i think having an accessor for typename isn't a bad thing.
can you take another pass anyway (soon) to cleanup all of this.

packages/apps/plugins/plugin-debug/src/components/DebugSpace.tsx
richburdon: how do you check if it's mounted?

packages/apps/plugins/plugin-inbox/src/components/Email/Mailbox.tsx
richburdon: sure. it's a strange DX though. @dmaretskyi something to think about in the next API pass. e.g., query option to filter null refs; or make null by default.

packages/sdk/client/src/echo/agent-query-source-provider.ts
richburdon: it's just a "marker" exception to check by value if we threw this.
@mykola-vrmchk i think this is better than checking the "text" value.

packages/apps/plugins/plugin-debug/src/components/DebugSpace.tsx
wittjosiah: the timeout is cleared if its unmounted so it isn't run on unmounted component, is that the problem you're talking about?

PR #6732, opened 2024-05-20
feat: Support indexed queries with `Filter.or`

mykola-vrmchk:
resolves https://github.com/dxos/dxos/issues/6725

### DIFF
--- a/packages/core/echo/echo-db/src/query/query-state.ts
+++ b/packages/core/echo/echo-db/src/query/query-state.ts
@@ -5,7 +5,7 @@
 import { type DocumentId } from '@dxos/automerge/automerge-repo';
 import { LifecycleState, Resource } from '@dxos/context';
 import { type AutomergeHost, getSpaceKeyFromDoc } from '@dxos/echo-pipeline';
-import { LogicalModifiers, type Indexer, type IndexQuery } from '@dxos/indexing';
+import { type Indexer, type IndexQuery } from '@dxos/indexing';
 import { PublicKey } from '@dxos/keys';
 import { idCodec } from '@dxos/protocols';
 import { type Filter as FilterProto } from '@dxos/protocols/proto/dxos/echo/filter';
@@ -163,8 +163,7 @@ export class QueryState extends Resource {
 // TODO(burdon): Process Filter DSL.
 const filterToIndexQuery = (filter: Filter): IndexQuery => {
   return {
-    typenames: filter.or.length >= 1 ? filter.or.map((f) => f.type?.itemId) : [filter.type?.itemId],
-    modifier: filter.or.length > 0 ? LogicalModifiers.OR : undefined,
+    typenames: filter.type?.itemId ? [filter.type.itemId] : filter.or.map((f) => f.type?.itemId).filter(nonNullable),
     inverted: filter.not,
   };
 };
--- a/packages/core/echo/echo-db/src/query/query-state.ts
+++ b/packages/core/echo/echo-db/src/query/query-state.ts
@@ -6,6 +6,7 @@ import { type DocumentId } from '@dxos/automerge/automerge-repo';
 import { LifecycleState, Resource } from '@dxos/context';
 import { type AutomergeHost, getSpaceKeyFromDoc } from '@dxos/echo-pipeline';
 import { type Indexer, type IndexQuery } from '@dxos/indexing';
+import { invariant } from '@dxos/invariant';
 import { PublicKey } from '@dxos/keys';
 import { idCodec } from '@dxos/protocols';
 import { type Filter as FilterProto } from '@dxos/protocols/proto/dxos/echo/filter';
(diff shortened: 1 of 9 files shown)


mykola-vrmchk:
* **#6732** <a href="https://app.graphite.dev/github/pr/dxos/dxos/6732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @mykola-vrmchk and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -49,7 +56,11 @@ export class IndexSchema extends Resource implements Index {
   }
 
   @trace.span({ showInBrowserTimeline: true })
-  async find(filter: IndexQuery) {
+  async find(filter: IndexQuery): Promise<FindResult[]> {
+    if (filter.or && filter.or.length > 0) {
```
dmaretskyi, line 60: can you pls add a test that `Filter.not(Filter.or(...))` also works?

packages/core/echo/indexing/src/types.ts
```diff
@@ -19,6 +19,8 @@ export type IndexQuery = {
    */
   typename?: string | null;
 
+  // TODO(mykola): Extract to QueryPlanner in future
+  or?: IndexQuery[];
```
dmaretskyi, line 23: I think it would be better if this type wasn't hierarchical -- consider merging this clause with the `typename` fieldsto to represent a request for any typeneme from the list.

mykola-vrmchk: Then it would not support `Filter.or(Filter.not(<something>), <something>)`

packages/core/echo/indexing/src/types.ts
```diff
@@ -17,10 +17,13 @@ export type IndexQuery = {
    * null means all Expando objects.
    * undefined means all objects (no filter).
    */
-  typename?: string | null;
+  typenames: (string | null | undefined)[];
+
+  /**
+   * Concatenate all results for each typename if `true`.
+   */
+  or?: boolean;
```
dmaretskyi, line 25: what's the behavior with false?

packages/core/echo/indexing/src/types.ts
```diff
@@ -17,10 +17,13 @@ export type IndexQuery = {
    * null means all Expando objects.
    * undefined means all objects (no filter).
    */
-  typename?: string | null;
+  typenames: (string | null | undefined)[];
```
dmaretskyi, line 20: does it make sense to include `undefiend` as one possible array value

packages/core/echo/indexing/src/types.ts
mykola-vrmchk: Same as `undefined`

packages/core/echo/indexing/src/types.ts
mykola-vrmchk: To support `db.query()` (query all objects)

dmaretskyi: what would a filter of `["example/Type", undefined]` do? query everying?

packages/core/echo/echo-db/src/query/query-state.ts
```diff
@@ -163,7 +163,7 @@ export class QueryState extends Resource {
 // TODO(burdon): Process Filter DSL.
 const filterToIndexQuery = (filter: Filter): IndexQuery => {
   return {
-    typename: filter.type?.itemId,
+    typenames: filter.type?.itemId ? [filter.type.itemId] : filter.or.map((f) => f.type?.itemId).filter(nonNullable),
```
dmaretskyi, line 166: what if some of the nested filters have a .not flag?

dmaretskyi: I guess something (this function? query state?) should bail and not do an indexed search if the index doesn't support this type of query

dmaretskyi (changes requested):

packages/core/echo/echo-db/src/query/query-state.ts
```diff
@@ -162,8 +162,13 @@ export class QueryState extends Resource {
 
 // TODO(burdon): Process Filter DSL.
 const filterToIndexQuery = (filter: Filter): IndexQuery => {
-  return {
-    typename: filter.type?.itemId,
-    inverted: filter.not,
-  };
+  if (filter.type || (filter.or.length > 0 && filter.or.every((subFilter) => !subFilter.not && subFilter.type))) {
```
dmaretskyi, line 165: is it guaranteed that filter.or and filter.type are exclusive?

mykola-vrmchk: Good question, but how would you resolve filter with `type` and `or` section

PR #6736, opened 2024-05-20
feat: Add ability to query Expando

mykola-vrmchk:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/core/echo/echo-db/src/query/filter.ts
+++ b/packages/core/echo/echo-db/src/query/filter.ts
@@ -230,12 +230,10 @@ const filterMatchInner = (filter: Filter, core: AutomergeObjectCore): boolean =>
         return false;
       }
     } else {
-      if (!type) {
+      if (!type && filter.type.itemId !== 'Expando') {
         return false;
-      }
-
-      // TODO(burdon): Comment.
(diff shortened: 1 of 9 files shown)


packages/core/echo/echo-db/src/query/filter.ts
```diff
@@ -230,12 +230,10 @@ const filterMatchInner = (filter: Filter, core: AutomergeObjectCore): boolean =>
         return false;
       }
     } else {
-      if (!type) {
+      if (!type && filter.type.itemId !== 'Expando') {
```
dmaretskyi, line 233: extract 'Expando' to a constant.
is there any way for this and Expando deifnition in echo-schema to use the same constant?

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -60,7 +60,7 @@ export class IndexSchema extends Resource implements Index {
     // TODO(burdon): Handle inversion.
     if (filter.inverted) {
       return Array.from(this._index.entries())
-        .filter(([key]) => !filter.typenames.includes(key) === false)
+        .filter(([key]) => !filter.typenames.includes(key ?? 'Expando') === false)
```
dmaretskyi, line 63: Extract constant

PR #6751, opened 2024-05-21
feat(functions): functions and triggers are ECHO objects

yarolegovich:
TBD

### DIFF
--- a/packages/core/functions/src/trigger/trigger-registry.ts
+++ b/packages/core/functions/src/trigger/trigger-registry.ts
@@ -61,9 +61,12 @@ export class TriggerRegistry extends Resource {
     super();
   }
 
+  public getActiveTriggers(space: Space): FunctionTrigger[] {
+    return this._getTriggers(space, (t) => t.activationCtx != null);
+  }
+
   public getInactiveTriggers(space: Space): FunctionTrigger[] {
-    const allSpaceTriggers = this._triggersBySpaceKey.get(space.key) ?? [];
-    return allSpaceTriggers.filter((t) => t.activationCtx == null).map((t) => t.trigger);
+    return this._getTriggers(space, (t) => t.activationCtx == null);
   }
 
   async activate(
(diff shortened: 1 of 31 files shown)


packages/core/functions/src/types.ts
```diff
@@ -2,26 +2,46 @@
...
+ * Every spec has a type field of type FunctionTriggerType that we can use to understand which
+ * type we're working with.
+ * https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions
+ */
+export enum FunctionTriggerType {
+  TIMER = 'timer',
+  WEBHOOK = 'webhook',
+  WEBSOCKET = 'websocket',
+  ECHO = 'echo',
+}
```
ZaymonFC, line 18: Is there a reason this can't be a normal string union?

```ts
type FunctionTriggerType = 'timer' | 'webhook' | 'websocket' | 'echo';
```

packages/core/functions/src/trigger/trigger-registry.ts
```diff
@@ -0,0 +1,158 @@
...
+export type OnTriggerCallback = (args: object) => Promise<ResponseCode>;
+
+export type FunctionTriggerContext = { space: Space };
+
+export type TriggerFactory<Spec extends TriggerSpec, Options = any> = (
+  ctx: Context,
+  triggerContext: FunctionTriggerContext,
+  spec: Spec,
+  callback: OnTriggerCallback,
+  options?: Options,
```
ZaymonFC, line 31: Consider making these args an object.

ZaymonFC:
Love the amount of tests 🧪

richburdon (commented):
I've pushed a bunch of updates to this PR.

General nits:
- avoid single letter v names
- catch e => err
- break long functions into paragraphs (with comments for any clever/non-trivial things)
- ECHO typenames in the FQ (fully qualified) URI form of: dxos.org/type/Foo (we need to migrate the existing braneframe types).
- Avoid "foo" in tests (if you can come up with a more realistic example)
- consider creating builders for repetitive or complex test infrastructure (e.g., TestBuilder)
- add TODOs to factor out useful/repetitive things (e.g., omitEchoId)
- `log('initializing: ${this.id}');` => `log('initializing', { id: this.id });`
- (in general don't put "data" in the text of a message/error/log).

PR #6752, opened 2024-05-21
fix(composer): Status bar dark mode colors

ZaymonFC:


packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
```diff
@@ -27,7 +27,10 @@ const StatusBarButton = forwardRef<HTMLButtonElement, StatusBarButtonProps>(
       <Component
         role='button'
         className={mx(
-          'flex items-center gap-2 p-1 px-2 rounded-sm select-none hover:bg-gray-200 cursor-pointer active:bg-gray-300',
+          'flex items-center gap-2 p-1 px-2 rounded-sm',
+          'select-none cursor-pointer',
+          'hover:bg-neutral-75 active:bg-neutral-150',
+          'dark:hover:bg-neutral-750 dark:active:bg-neutral-700',
```
thure, line 33: We have the `surface-hover` token for this, see if that works for this case. Generally we haven’t bothered with CSS active states, not that it needs to be removed here necessarily – if you feel principled about it, it should be added to the system.

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
```diff
@@ -66,7 +69,9 @@ const StatusBarContainer = forwardRef<HTMLDivElement, StatusBarContainerProps>(
         {...groupAttrs}
         className={mx(
           'bs-[--statusbar-size]',
-          'flex justify-end items-center gap-2 border-solid border-t border-gray-200 bg-gray-50 text-gray-500',
+          'flex justify-end items-center gap-2',
+          'surface-base fg-description',
+          'border-solid border-t border-bs separator-separator',
```
thure, line 74:
```suggestion
          'border-bs separator-separator',
```
`border-solid` is inferred by `border-bs`, and `border-t` is overridden by `border-bs`

packages/ui/react-ui-theme/src/config/colors.ts
```diff
@@ -218,8 +218,8 @@ export const semanticColors = {
     dark: physicalColors.neutral['100'],
   },
   description: {
-    light: physicalColors.neutral['300'],
-    dark: physicalColors.neutral['500'],
+    light: physicalColors.neutral['500'],
+    dark: physicalColors.neutral['400'],
```
thure, line 222: Thank you!

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
```diff
@@ -27,7 +27,10 @@ const StatusBarButton = forwardRef<HTMLButtonElement, StatusBarButtonProps>(
       <Component
```
thure, line 27: I know this isn’t part of the PR, but generally this should be achieved using Radix-style `asChild` and `Slot`, just noting for future reference.

ZaymonFC: Hmm. I'm using the radix Slot component. I still haven't internalised the pattern yet.

thure: Ah, nevermind, yes you are! Generally the variable is named `Root` in this case, nbd.

ZaymonFC:
```ts
const StatusBarButton = forwardRef<HTMLButtonElement, StatusBarButtonProps>(
  ({ classNames, children, asChild, ...props }, forwardedRef) => {
    const classes = mx(
      classNames,
      'flex items-center gap-2 p-1 px-2 rounded-sm',
      'select-none cursor-pointer',
      'hover:bg-neutral-75 active:bg-neutral-150',
      'dark:hover:bg-neutral-750 dark:active:bg-neutral-700',
    );

    return asChild ? (
      <Slot role='button' ref={forwardedRef} className={classes}>
        {children}
      </Slot>
    ) : (
      <button className={classes} ref={forwardedRef} {...props}>
        {children}
      </button>
    );
  },
);
```

Like this?

thure: No, just rename `Component` to `Root` and it’ll be consistent with Radix and our other components that do this.

thure: I was just confused because we sometimes pass a component to use for rendering mosaic tiles as a prop named `{…}Component` and assumed without looking at the context that that’s what was going on, welcome to ignore this whole thread 😅.

ZaymonFC: Now that I'm here I like the ternary better. ¯\_(ツ)_/¯.

Noted on the naming convention though!

packages/apps/plugins/plugin-status-bar/src/components/StatusBar.tsx
ZaymonFC: It didn't look right in this case. I quite like active states, we should consider adding it to the system as you say.

thure: I’m wary, then, of introducing another semantic token for backgrounds applied only on hover purely for aesthetic reasons, and the exception begs questions… should the hover semantic token be adjusted? Do these physical colors work here and not elsewhere, and if so why?

PR #6754, opened 2024-05-22
feat(plugin-threads): Open comments in c11y sidebar when attention changes

### LINKED ISSUE #6571: [stack] comments on stacks by showing comments for currently attended section
the ask was to create a comments thread view for collections (stacks), with a concatenation of comment threads from sections.

The current proposal: comments sidebar fills with threads for the currently attended section in a stack. 

We found it hard to imagine the UX being good around all those threads in a big list together, several gaps occurred such as:
- where do the comments on the collections go? do they go at the top?
- how does scrolling work in both content and threads views?
- is there a way to submit a new comment without selecting something? where would that affordance be?

thure:
Resolves #6571.

https://github.com/dxos/dxos/assets/855039/a4311949-cad9-43bb-9f22-17c625c88e8f

This PR adds a subscription to ThreadsPlugin’s `ready` which opens threads in the complementary sidebar if any are found related to the attended items.


ZaymonFC (approved):
Well that was simple. Nice one!

PR #6755, opened 2024-05-22
feat: Composer CLI extension.

richburdon:
- [x] Starting point for Composer CLI extension to import/export data.


packages/apps/types/src/schema/thread.ts
```diff
@@ -67,7 +71,7 @@ export class MessageType extends TypedObject({ typename: 'braneframe.Message', v
 //   Interoperability should be handled by lenses or some other transformation mechanism.
 //   Requiring the same schema for all types of messages will not scale -
 //   It also makes the simple cases much more complex than they need to be.
-export class ThreadType extends TypedObject({ typename: 'braneframe.Thread', version: '0.1.0' })({
+export class ThreadType extends TypedObject({ typename: 'dxos.org/type/Thread', version: '0.1.0' })({
```
yarolegovich, line 74: are we ok with breaking all the existing threads? 
the app will be querying for threads using `dxos.org/type/Thread` typename filter, but the data is actually stored with `braneframe.Thread` typename. that's why we kept all names during effect schema migration

PR #6785, opened 2024-05-22
chore: Clean-up reactive object creation.

richburdon:


packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -0,0 +1,88 @@
...
+    return _create<T>({ ...obj } as T, meta, objOrSchema as S.Schema<T>);
+  } else if (obj && (objOrSchema as any) === Expando) {
+    return _create<T>({ ...obj } as T, meta, undefined, true);
+  } else {
+    // TODO(burdon): Breaks if cloned?
+    return _create<T>(objOrSchema as T, meta);
+  }
+};
+
+const _create = <T extends {}>(obj: T, meta?: ObjectMeta, schema?: S.Schema<T>, expando = false): ReactiveObject<T> => {
```
dmaretskyi, line 38: Add options object for the expando flag

packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (!isValidProxyTarget(obj)) {
+    throw new Error('Value cannot be made into a reactive object.');
+  }
+
+  if (schema) {
+    if (getEchoObjectAnnotation(schema)) {
```
dmaretskyi, line 44: const needsId = expando || getEchoObjectAnnotation(schema)

packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+};
+
+// TODO(dmaretskyi): UUID v8.
+const _generateId = () => PublicKey.random().toHex();
+
+/**
+ * Set ID on ECHO objects (Schema and Expando).
+ */
+const _setId = <T extends {}>(obj: ExcludeId<T>) => {
```
dmaretskyi, line 65: Remove underscore -- I think we generally only use them on class members or when there a collision with an existing name (e.g. create)

packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -0,0 +1,88 @@
...
+  invariant(!('id' in (obj as any)), 'Object already has an `id` field, which is reserved.');
+  (obj as any).id = _generateId();
+};
+
+const symbolMeta = Symbol.for('@dxos/meta');
+
+/**
+ * Set metadata on object.
+ */
+const _initMeta = <T>(obj: T, meta: ObjectMeta = { keys: [] }) => {
```
dmaretskyi, line 75: Remove underscore -- I think we generally only use them on class members or when there a collision with an existing name (e.g. create)

packages/core/echo/echo-schema/src/types.ts
```diff
@@ -2,20 +2,47 @@
...
 export const TYPE_PROPERTIES = 'dxos.sdk.client.Properties';
 
+// TODO(burdon): Use consistently (with serialization utils).
+export const ECHO_ATTR_ID = '@id';
+export const ECHO_ATTR_TYPE = '@type';
+export const ECHO_ATTR_META = '@meta';
+
+export type ExcludeId<T> = Simplify<Omit<T, 'id'>>;
+
+export const ForeignKeySchema = S.struct({
```
dmaretskyi, line 20: I'd suggest using the aliasing pattern to link the const with the typedef

packages/core/echo/echo-schema/src/types.ts
```diff
@@ -2,20 +2,47 @@
...
-// This type doesn't change the shape of the object, it is rather used as an indicator that the object is reactive.
-export type ReactiveObject<T> = { [K in keyof T]: T[K] };
+export const RawObject = <T>(schema: S.Schema<T>): S.Schema<ExcludeId<T> & WithMeta> => {
+  return S.make(AST.omit(schema.ast, ['id']));
+};
 
 /**
  * Has `id`.
  */
+// TODO(burdon): Rename BaseObject?
```
dmaretskyi, line 45: I think aspects are better here since not all proxies have id

packages/core/echo/echo-schema/src/types.ts
```diff
@@ -25,33 +52,22 @@ export interface Identifiable {
...
-  source?: string;
-
-  /**
-   * Id within the foreign database.
-   */
-  id?: string;
-};
+export type ReactiveObject<T> = { [K in keyof T]: T[K] };
+
+export type EchoReactiveObject<T> = ReactiveObject<T> & Identifiable;
```
dmaretskyi, line 61: Rename to just echo object?

packages/core/echo/echo-schema/src/types.ts
```diff
@@ -25,33 +52,22 @@ export interface Identifiable {
...
 /**
- * Echo object metadata.
+ * Utility to split meta property from raw object.
  */
-export type ObjectMeta = {
-  /**
-   * Foreign keys.
-   */
-  keys: ForeignKey[];
+export const splitMeta = <T>(object: T & WithMeta): { object: T; meta?: ObjectMeta } => {
```
dmaretskyi, line 69: Perhaps we just have 'create' accept @meta ?

dmaretskyi (approved):

packages/core/echo/echo-schema/src/types.ts
yarolegovich: In a separate PR maybe 
image

packages/core/echo/echo-schema/src/types.ts
yarolegovich: wdym, meta was already added to create

richburdon: Hmmm. I think I prefer it separate.

PR #6786, opened 2024-05-23
feat: Add space key to index metadata

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/protocols/src/indexing.ts
```diff
@@ -3,19 +3,58 @@
...
+  /**
+   * `#01|${spaceKey}|${documentId}|${objectId}`
+   */
+  V1 = 1,
+}
+
+export const objectPointerCodec = {
+  encode: ({ spaceKey, documentId, objectId }: ObjectPointerDecoded): ObjectPointerEncoded => {
+    if (spaceKey === undefined) {
+      return `${documentId}|${objectId}`;
```
mykola-vrmchk, line 35: Do we need this path?

mykola-vrmchk (approved):

packages/common/util/src/join-tables.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Joins two untyped tables on the given columns.
+ * @param leftColumn The column to join on in the first table.
+ * @param rightColumn The column to join on in the second table.
+ */
+//
```
richburdon, line 6: move?

richburdon (commented):
Would it be straightforward to add the meta.keys to the index?

PR #6789, opened 2024-05-23
feat: Pass space ref to file.upload provider

icidasset:
### Motivation / Background

This PR aims to complete the first two TODOs from #6717 

### Details

- [x] Passes the relevant space to `provides.file.upload`
- [x] Remove the need for `useResolvePlugin(parseFileManagerPlugin)` in the `EditorMain` component of the markdown plugin

### DIFF
--- a/packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
+++ b/packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
@@ -4,7 +4,7 @@
 
 import { type IconProps, TextAa } from '@phosphor-icons/react';
 import { batch, effect } from '@preact/signals-core';
-import React, { useCallback, useMemo, type Ref } from 'react';
+import React, { useMemo, type Ref } from 'react';
 
 import { parseClientPlugin } from '@braneframe/plugin-client';
 import { parseSpacePlugin, updateGraphWithAddObjectAction } from '@braneframe/plugin-space';
--- a/packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
+++ b/packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
@@ -247,20 +247,19 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
           }, [doc, space, settings.values.editorMode]);
 
           const fileManagerPlugin = useResolvePlugin(parseFileManagerPlugin);
-          const onFileUpload = useCallback(
-            async (file: File) => {
-              if (space === undefined) {
-                return undefined;
-              }
+          const onFileUpload = useMemo(() => {
+            if (space === undefined) {
+              return undefined;
+            }
 
-              if (fileManagerPlugin?.provides.file.upload === undefined) {
-                return undefined;
-              }
+            if (fileManagerPlugin?.provides.file.upload === undefined) {
+              return undefined;
+            }
 
-              return await fileManagerPlugin.provides.file.upload(file, space);
-            },
-            [fileManagerPlugin, space],
-          );
+            return async (file: File) => {
+              return await fileManagerPlugin?.provides?.file?.upload?.(file, space);
+            };
+          }, [fileManagerPlugin, space]);
 
           switch (role) {
             // TODO(burdon): Normalize layout (reduce variants).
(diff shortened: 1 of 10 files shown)


packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -243,6 +245,19 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
+            if (space === undefined) {
+              throw new Error('Unable to determine space used');
+            }
+
+            if (fileManagerPlugin?.provides.file.upload === undefined) {
+              throw new Error('No upload method was provided');
+            }
+
+            return await fileManagerPlugin.provides.file.upload(file, space);
+          };
```
ZaymonFC, line 259: Can you convert this to a `useCallback` with `space` and `fileManagerPlugin` as deps?

packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -49,15 +49,23 @@ const useTest = (view: EditorView | null) => {
 
 export type EditorMainProps = {
   id: string;
+  onFileUpload?: (file: File) => Promise<FileInfo | undefined>;
   readonly?: boolean;
   toolbar?: boolean;
   comments?: Comment[];
 } & Pick<TextEditorProps, 'doc' | 'selection' | 'scrollTo' | 'extensions'>;
```
ZaymonFC, line 56: Nit: Can you put the `onFileUpload` last please?

packages/sdk/app-framework/src/plugins/common/file.ts
```diff
@@ -17,7 +19,7 @@ export type FileInfo = {
   cid?: string; // TODO(burdon): Meta key? Or other common properties with other file management system? (e.g., WNFS).
 };
 
-export type FileUploader = (file: File) => Promise<FileInfo | undefined>;
+export type FileUploader = (file: File, space: Space) => Promise<FileInfo | undefined>;
```
ZaymonFC, line 23: It seems weird to me that this type lives in the `AppFramework`. CC @wittjosiah.

It should be fine for now.

ZaymonFC (approved):
Some minor issues to address before merging. Nice work. Thanks for contributing! ✨

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
wittjosiah: I also think that it would be good to omit `onFileUpload` if the plugin doesn't exist rather than erroring when its called. This would give the editor a hint to disable/hide the functionality if it's not available.

packages/sdk/app-framework/src/plugins/common/file.ts
wittjosiah: Yeah, its fine for now. We've defined provides types for "common" plugins that we expect there are likely to be multiple implementations of here for now as an easy place to share types without creating new packages.

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
icidasset: Done 👍  I removed the errors @wittjosiah, now just returns `undefined` if the space and file manager plugin aren't present. Unless you want the `onFileUpload` variable itself to be undefined if those two dependencies aren't present?

icidasset:
Feedback applied & rebased

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
ZaymonFC: Yeah the `onFileUpload` prop should be undefined if the file manager plugin isn't present.

wittjosiah: Yeah, what Zaymon said. Because the editor handles the case where `onFileUpload` isn't present by hiding those controls, if the capabilities are not available the callback itself should just be `undefined`. Once that change is made, I think we should be able to merge this.

icidasset: Updated ✅

wittjosiah:
@icidasset it looks like a lint check is failing, could you take a look at that?

icidasset:
@wittjosiah Should be good now, remaining failures are unrelated to the changes made here.

PR #6806, opened 2024-05-24
feat(composer): Trigger editor

ZaymonFC:


ZaymonFC:
- Changes integrated into #6835

PR #6808, opened 2024-05-24
fix: Fragment imported space; Improve import speed

mykola-vrmchk:
- Implement test changes
- Delete only
- change variant
- Change check
- Fix serializer


packages/core/echo/echo-db/src/automerge/automerge-db.ts
```diff
@@ -545,16 +544,12 @@ export interface ItemsUpdatedEvent {
...
-    // NOTE: We need to store properties in the root document since space-list initialization
-    //  expects it to be loaded as space become available.
-    if (core.getType()?.itemId === TYPE_PROPERTIES) {
-      return false;
-    }
-    return true;
-  } else {
+  // NOTE: We need to store properties in the root document since space-list initialization
+  //  expects it to be loaded as space become available.
+  if (core.getType()?.itemId === TYPE_PROPERTIES) {
```
dmaretskyi, line 549:
```ts
return core.getType()?.itemId !== TYPE_PROPERTIES
```

PR #6838, opened 2024-05-25
feat(functions): Wire in schema

ZaymonFC:


richburdon (approved):

packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
```diff
@@ -19,6 +19,16 @@ import {
...
+import { distinctBy } from '@dxos/util';
+
+type TriggerId = string;
+
+const stateInitialValues = {
+  schemas: [StackType, SketchType, DocumentType, FileType, MessageType] as any[],
+  selectedSchema: {} as Record<TriggerId, any>,
+};
+
+const state = create<typeof stateInitialValues>(stateInitialValues);
```
ZaymonFC, line 31: Lightweight alternative to context. I'm quite fond of this pattern and it's much better for performance to use signals like this.

packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
```diff
@@ -29,15 +39,40 @@ export const TriggerEditor = ({ space, trigger }: { space: Space; trigger: Funct
...
+    if (spec.type === 'subscription') {
+      if (spec.filter && spec.filter.length > 0) {
+        const type = spec.filter[0].type;
+        const foundSchema = state.schemas.find((schema) => schema.typename === type);
+        if (foundSchema) {
+          state.selectedSchema[trigger.id] = foundSchema;
+        }
+      }
+    }
+  }, [JSON.stringify(trigger.spec), state.schemas]);
```
ZaymonFC, line 68: Would love a signal utility that deeply watches a collection.

richburdon: yes -- we must do this -- there's a couple of hacks around the place to do this kind of thing.

PR #6851, opened 2024-05-27
chore(cli-composer): remove missing script

wittjosiah:


wittjosiah:
https://github.com/dxos/dxos/pull/6852

PR #6860, opened 2024-05-27
chore: revert node version upgrade

wittjosiah:
may be causing issues with publishing https://github.com/pnpm/pnpm/issues/4348


wittjosiah:
#6863 ended up solving the issue

PR #6862, opened 2024-05-27
chore(deck): Get devtools working with Deck

ZaymonFC:


packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
```diff
@@ -199,48 +202,62 @@ export const DebugPlugin = (): PluginDefinition<DebugPluginProvides> => {
...
+          const { active, object } = data;
+
           switch (role) {
             case 'settings':
               return data.plugin === meta.id ? <DebugSettings settings={settings.values} /> : null;
             case 'status':
               return <DebugStatus />;
           }
 
-          if (!settings.values.debug) {
```
ZaymonFC, line 210: `settings.values.debug` and `settings.values.devtools` can be enabled separately.

packages/sdk/app-framework/src/plugins/common/navigation.ts
```diff
@@ -8,7 +8,7 @@ import type { IntentData } from '../IntentPlugin';
 import type { Plugin } from '../PluginHost';
 
 // NOTE(thure): These are chosen from RFC 1738’s `safe` characters: http://www.faqs.org/rfcs/rfc1738.html
-export const SLUG_LIST_SEPARATOR = '.';
+export const SLUG_LIST_SEPARATOR = '+';
```
ZaymonFC, line 11: + gives nicer URL semantics and allows plank id's to contain `.` like in the case of `dxos.org.plugin.debug.devtools`.

wittjosiah: @thure

wittjosiah (approved):

packages/sdk/app-framework/src/plugins/common/navigation.ts
thure: I'd be fine with this change, sure 👍

PR #6877, opened 2024-05-28
fix: Complementary sidebar opening behavior

### LINKED ISSUE #6875: [deck] clicking on a node already open in a plank will scroll to it but not attend it
is this by design? seems like it should be ok to just focus and attend it, that seems to be the most probable intent behind "navigate to"

thure:
This PR
- resolves #6875


ZaymonFC (approved):
LGTM :>

PR #6890, opened 2024-05-29
fix: Scroll plank into view when navigating from a Stack section

### LINKED ISSUE #6885: [deck] navigating to an item from a stack does not scroll to it
if the content is already in the deck somewhere, it will correctly attend the item but will not scroll it into view

thure:
This PR:
- resolves #6885
- adjusts stack sections to use border instead of outline, aligning appearance with documents

https://github.com/dxos/dxos/assets/855039/2b9615f7-0f9e-4db0-a337-31a745824f44


ZaymonFC (approved):
LGTM

Issue #4348, opened 2023-09-29
[docs] add subscribe snippets to ts docs


zhenyasav:
yes the `subscribe` symbol is how you would do that, not sure why it's missing from the current snippets, that's an oversight, but it would look something like this:

```ts
import { Client } from '@dxos/client';
import { Expando, subscribe } from '@dxos/client/echo';

const client = new Client();

(async () => {
  await client.initialize();
  if (!client.halo.identity.get()) await client.halo.createIdentity();

  const space = client.spaces.get()[0];

  const object = new Expando({ type: 'task', title: 'buy milk' });

  await space.db.add(object);

  object[subscribe]((object) => {
    // object.title === 'buy chocolate'
  });

  object.title = 'buy chocolate';
})();
```

 we should make sure some sort of basic `subscribe` example gets included in the snippets

PR #6892, opened 2024-05-29
docs: add page on subscriptions in TypeScript

seagreen:
Fix: https://github.com/dxos/dxos/issues/4348


docs/content/guide/echo/typescript/snippets/subscription.ts
```diff
@@ -0,0 +1,37 @@
...
+    objects.forEach(object => {
+      if (object.type === 'task') {
+        console.log('Do something with this task');
+      } else {
+        throw new Error('Non-task object returned (this will never happen)');
+      }
+    });
+  });
+
+  try {
```
seagreen, line 29: @yarolegovich: I used try/catch here instead of a [Context](https://github.com/dxos/dxos/blob/a7f191de8b3bd78fac9aa8c993c4080902a9eb2b/packages/common/context/src/context.ts#L40) because I wasn't confident enough in my understanding of Context to use it or document it.

Do you think this is good enough for now, or should we switch to `Context` in the code and explain how it works?

docs/content/guide/echo/typescript/subscriptions.md
```diff
@@ -0,0 +1,44 @@
+---
+order: 3
+---
+
+# Subscriptions
+
+Use subscriptions to listen for changes within a [space](./README.md).
```
seagreen, line 7: More prose wouldn't hurt here if people have suggestions on the details of how they work.

seagreen:
This code could use careful review for sure.

docs/content/guide/echo/typescript/subscriptions.md
seagreen: ...in particular, now that this page covers two different types of subscriptions, some writing on how they differ could be useful.

docs/content/guide/echo/typescript/subscriptions.md
```diff
@@ -0,0 +1,85 @@
...
+    space.db.add(eventObject);
+  } finally {
+    unsubscribeFn();
+  }
+};
+```
+
+## Do something when an object changes
+
+Objects returned by ECHO queries are [ReactiveObject](../../../api/@dxos/client/types/ReactiveObject.md)s, which are a kind of `ProxyObject` that will trigger reactivity and subscriptions like a Preact signal. Unlike signals, the values are read and modified directly rather than through `.value`.
```
seagreen, line 49: @thure I hyperlinked `ReactiveObject`, is there also a URL `ProxyObject` should point to?

docs/content/guide/echo/typescript/snippets/on-object-change.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  object.title = "buy cookies"
+
+  if (titles.join() === ["buy milk", "buy cookies"].join()) {
+    console.log("Success.");
+  } else {
+    console.log("This will never happen.");
+  }
+}
```
thure, line 37: Maybe call the `unsubscribeFn` at the end?

thure (approved):
This looks fine to me! If the code runs, it otherwise passes my scrutiny.

docs/content/guide/echo/typescript/subscriptions.md
thure: Sure, that one is native, so: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

thure: Speaking of which, now that I’ve looked it up, I suppose mentions of `ProxyObject` should change to ‘`Proxy` object’.

thure: (since `Proxy` alone is maybe ambiguous with ‘proxy server’ etc…)

docs/content/guide/echo/typescript/snippets/on-object-change.ts
seagreen: Done!

docs/content/guide/echo/typescript/subscriptions.md
seagreen: That makes more sense. Changed '`ProxyObject`' to '[Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object'.

seagreen:
> This looks fine to me! If the code runs, it otherwise passes my scrutiny.

Always good thinking to double check. In this case it paid off since I somehow typed a rogue word in [here](https://github.com/dxos/dxos/pull/6892/commits/9a2a5dc4ecf1fcc5d922f4cdf68672be220ddef2).

Fixed now and all running✅

jessmartin:
@ZaymonFC I believe you are working on a typescript app and going through the subscriptions flow. Just wanted to make sure you knew this PR was here.

zhenyasav (approved):

Issue #6749, opened 2024-05-21
[composer] drag-and-drop between spaces is broken for sketches


seagreen:
Steps to reproduce:

1. Create two spaces
2. Create a Sketch in the first
3. drag-and-drop the sketch into the second space

https://github.com/dxos/dxos/assets/1310143/38416b19-e531-4700-b098-37556ca89d4a

Console error:
```
react-D-u6Bfzb.js:32 Error: invariant violation [isReactiveObject(obj)] at packages/core/echo/echo-db/src/automerge/automerge-object.ts:16
```

PR #6905, opened 2024-05-30
docs: fill out the user guide for Spaces somewhat

seagreen:


docs/content/composer/user-guide/spaces.md
```diff
@@ -8,10 +8,26 @@ headerDepth: 0
...
 
-### Shared Spaces
+### Shared spaces
 
 Users can create shared spaces and invite others to join them. The data in a shared space is replicated across all of the devices of the users who have access to the space.
+
+## Drag-and-drop
+
+::: warning Technology Preview
+Moving objects between spaces is a work in progress.
```
seagreen, line 30: This is a somewhat strong statement, but I figured due to https://github.com/dxos/dxos/issues/6749 it was better to set expectations low. We can remove this statement whenever that issue's fixed.

(As far as I can tell drag-and-drop of folders isn't implemented yet either, which is another thing users are going to want immediately. If that's wrong let me know.)

PR #6908, opened 2024-06-01
chore: Update effect schema

richburdon:


jessmartin:
A lot of changes in this PR unrelated to the Effect Schema upgrade. 

@richburdon Also, this seems to merge in the tldraw upgrade. Were the problems with the tldraw migration fixed and tested?

PR #6941, opened 2024-06-05
feat: Random composer startup perf optimizations

dmaretskyi:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/apps/plugins/plugin-debug/src/components/Tree.tsx
+++ b/packages/apps/plugins/plugin-debug/src/components/Tree.tsx
@@ -3,7 +3,11 @@
 //
 
 import React, { type FC, type HTMLAttributes, useState } from 'react';
-import SyntaxHighlighter from 'react-syntax-highlighter';
+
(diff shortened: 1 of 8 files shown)


packages/apps/plugins/plugin-debug/src/components/Tree.tsx
```diff
@@ -3,7 +3,10 @@
 //
 
 import React, { type FC, type HTMLAttributes, useState } from 'react';
-import SyntaxHighlighter from 'react-syntax-highlighter';
+// Importing light & async version directly from dist to avoid any chance of the heavy one being loaded.
```
wittjosiah, line 6: clarify that "light" means "lightweight" in this context

packages/devtools/devtools/src/components/JsonView.tsx
```diff
@@ -3,8 +3,10 @@
 //
 
 import React, { type FC } from 'react';
+// Importing light & async version directly from dist to avoid any chance of the heavy one being loaded.
```
wittjosiah, line 6: ditto above

PR #6956, opened 2024-06-10
chore(mesh): observability metrics

yarolegovich:
### Details

* Added `invariant_violation` tag for simple sentry filtering. Invariant failures should be fixed ASAP because it's supposed to be an unexpected situation without graceful recovery path.
* Removed `log.catch` on gossip abort. We're just logging control extension heartbeat timeouts there.
* Refactored and stabilized (hopefully, no failures in 100 consecutive suite runs) signal manager tests.
* Extracted state reconciliation logic from signal-manager into signal-local-state.
* Use `failedServers` bitfield to avoid spamming "restarting server" warnings during downtimes. The flag is cleared on first successful message send. Same for consecutive reconciliation failures. Was thinking about just filtering timeout errors here.
* Added metrics for system-level behavior like long gc counter, reliable message sending split by retry attempts etc. These should help better understand how often certain code-paths are taken.
* Added peer-level metrics for signal message sending like bytes-in, bytes-out, message sending status, rtt. Cardinality is potentially unbounded, but as we discussed with Josiah we'll get rid of those when the amount of data becomes problematic. Can get rid of split by sender and still have a useful distribution.
* Tried to make signal client classes more robust. SignalClient is a Resource now to avoid open&close call overlaps, socket callbacks check `closed` status, rpc-client callbacks check that the corresponding client is still the active client, stream message handlers check we are still subscribed to messages / topic before delivering.


wittjosiah (approved):

packages/core/mesh/messaging/src/signal-manager/websocket-signal-manager.ts
```diff
@@ -142,26 +145,35 @@ export class WebsocketSignalManager implements SignalManager {
...
+      void server
+        .sendMessage({ author, recipient, payload })
+        .then(() => this._clearServerFailedFlag(serverName, index))
+        .catch((err) => {
+          if (err instanceof RateLimitExceededError) {
+            log.info('WSS rate limit exceeded', { err });
+            this._monitor.recordRateLimitExceeded();
+          } else if (err instanceof TimeoutError || err.constructor.name === 'TimeoutError') {
+            log.info('WSS sendMessage timeout', { err });
+            void this.checkServerFailure(serverName, index);
```
nferch, line 158: Should this try and distinguish application or RPC-level failures (e.g. software failure on the server side) with network-level errors (e.g. loss of Internet connectivity)?

packages/core/mesh/messaging/src/signal-manager/websocket-signal-manager.ts
```diff
@@ -24,13 +26,18 @@ const WSS_SIGNAL_SERVER_REBOOT_DELAY = 3_000;
  */
 export class WebsocketSignalManager implements SignalManager {
   private readonly _servers = new Map<string, SignalClientMethods>();
+  private readonly _monitor = new WebsocketSignalManagerMonitor();
+
+  /**
+   * Used to avoid logging failed server restarts more than once until the server actually recovers.
+   */
+  private readonly _failedServersBitfield: Uint8Array;
```
nferch, line 34: Does the client do any backoff on failure or are we just suppressing log messages here?

packages/core/mesh/messaging/src/signal-client/signal-client-monitor.ts
```diff
@@ -0,0 +1,107 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { trace } from '@dxos/tracing';
+
+import type { Message } from '../signal-methods';
+
+export class SignalClientMonitor {
```
nferch, line 9: I (think) I understand the difference between the messenger and the signal-client, but I wonder for the sake of observability whether it'd be beneficial to hide that complexity. With a lot of these metrics it'll be difficult to discern what is due to network connectivity/system load vs system or software issues, I think the highest fidelity might be in in:
P0
* number of sent messages (overall amount of traffic)
* number of rate limited messages

P1
* number of retransmissions 
* number of connection aborted due to failure
* number of failed messages

in general, less is better

packages/core/mesh/messaging/src/messenger-monitor.ts
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { trace } from '@dxos/tracing';
+
+export class MessengerMonitor {
+  public recordLongGc() {
```
nferch, line 8: what is this metric indicative of? how would we discern a software issue vs system load?

nferch (approved):
LGTM!

I think my only reservations are around adding additional complexity/state tracking (I wonder how this limits our scalability on both the client and server side) and the operational usefulness of the metrics (it's easy to get lost in a sea of data), but this is more open questions than anything.

I'm also missing some context on how the metrics are used on the Sentry side, would love to learn more about how that compares with the other solutions.

packages/core/mesh/messaging/src/signal-manager/websocket-signal-manager.ts
yarolegovich: client has a static 3s wait delay, but the field is used only for log message suppression

packages/core/mesh/messaging/src/signal-client/signal-client-monitor.ts
yarolegovich: >  but I wonder for the sake of observability whether it'd be beneficial to hide that complexity

you mean changing metric names (`mesh.signal.signal-client.received-total` -> `mesh.signal.received-total`)?
I'm ok with having a small number of core class-agnostic metrics like sent message count, call duration distribution etc. 
but I'd also keep class level metrics. the idea here is to use counters for understanding how often (if at all) certain code paths are taken. revisiting the PR after two weeks I think "recordLongGc" might be a bad example. if this starts happening, this will likely be a rare event (easy to miss with time series counter) and I'll probably want to examine at least some of the sessions in detail.
but there might be some code branches expected to be taken, but only rarely. logging warning (medium-priority sentry issues) is an overkill, but I want to have a cheap way for getting a signal when the branch starts being taken often (with case-based definition of often). I used to do this with logs, but a counter sounds like a better way in our env

Issue #6955, opened 2024-06-10
Use ULID for echo ID generation


dmaretskyi:
https://github.com/ulid/javascript/tree/master

PR #6957, opened 2024-06-10
refactor: Change ECHO ids to use Ulid

mykola-vrmchk:
closes https://github.com/dxos/dxos/issues/6955

### DIFF
--- a/packages/core/echo/echo-db/src/database.test.ts
+++ b/packages/core/echo/echo-db/src/database.test.ts
@@ -5,7 +5,16 @@
 import { expect } from 'chai';
 import { inspect } from 'node:util';
 
-import { create, Expando, getMeta, getSchema, getType, type ReactiveObject } from '@dxos/echo-schema';
+import {
+  create,
+  Expando,
+  getMeta,
+  getProxyHandlerSlot,
+  getSchema,
+  getType,
+  type ReactiveObject,
+} from '@dxos/echo-schema';
+import { PublicKey } from '@dxos/keys';
 import { describe, test } from '@dxos/test';
 import { range } from '@dxos/util';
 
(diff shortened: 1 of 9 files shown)


packages/core/echo/echo-db/src/database.test.ts
```diff
@@ -206,6 +215,21 @@ describe('Database', () => {
...
   });
 
+  test('Database works with old PublicKey IDs and new Ulid IDs', async () => {
+    const { db } = await builder.createDatabase();
+    const obj = db.add(create(Expando, { string: 'foo' })); // Ulid by default
+
+    // Old format
+    const oldId = PublicKey.random().toHex();
+    const reactiveObjWithOldId = create(Expando, { string: 'foo' });
+    (getProxyHandlerSlot(reactiveObjWithOldId).target as any).id = oldId;
```
dmaretskyi, line 225: extract to `dangerouslyAssignProxyId` function

packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -62,8 +62,7 @@ const _create = <T extends {}>(
   }
 };
 
-// TODO(dmaretskyi): UUID v8.
-const _generateId = () => PublicKey.random().toHex();
+const _generateId = () => ulid();
```
dmaretskyi, line 65: can we remove the underscore?

packages/core/echo/echo-schema/src/handler/object.ts
```diff
@@ -62,15 +62,14 @@ const _create = <T extends {}>(
...
 };
 
-// TODO(dmaretskyi): UUID v8.
-const _generateId = () => PublicKey.random().toHex();
+const generateId = () => ulid();
 
 /**
  * Set ID on ECHO objects (Schema and Expando).
  */
 const setId = <T extends {}>(obj: ExcludeId<T>) => {
```
dmaretskyi, line 70:
```suggestion
/**
 * Set ID on ECHO object targets during creation.
 * Used for objects with schema and the ones explicitly marked as Expando.
 */
const setIdOnTarget = (target: any) => {
```

dmaretskyi (approved):

PR #6964, opened 2024-06-11
refactor: Remove reference to EchoDatabase from lower level database

mykola-vrmchk:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/core/echo/echo-db/src/database.ts
+++ b/packages/core/echo/echo-db/src/database.ts
@@ -229,8 +229,8 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
     const target = getProxyHandlerSlot(echoObject).target as ProxyTarget;
     EchoReactiveHandler.instance.setDatabase(target, this);
     EchoReactiveHandler.instance.saveLinkedObjects(target);
-
     this._automerge.add(echoObject);
+
     return echoObject as any;
   }
 
--- a/packages/core/echo/echo-db/src/database.ts
+++ b/packages/core/echo/echo-db/src/database.ts
@@ -228,9 +228,9 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
 
     const target = getProxyHandlerSlot(echoObject).target as ProxyTarget;
     EchoReactiveHandler.instance.setDatabase(target, this);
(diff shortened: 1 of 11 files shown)


packages/core/echo/echo-db/src/query/filter.ts
```diff
@@ -188,16 +188,30 @@ export class Filter<T extends {} = any> {
...
+/**
+ * Query logic that checks if object complaint with a filter.
+ * @param echoObject used for predicate filters only.
+ * @returns
+ */
+export const filterMatch = (
+  filter: Filter,
+  core: AutomergeObjectCore | undefined,
+  // TODO(mykola): Remove predicate filters from this level query. Move it to higher proxy level.
+  echoObject?: EchoReactiveObject<any> | undefined,
```
dmaretskyi, line 200:
```suggestion
  echoObject: EchoReactiveObject<any> | undefined,
```

packages/core/echo/echo-db/src/database.ts
```diff
@@ -253,6 +264,10 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
    * @deprecated
    */
   get objects(): EchoReactiveObject<any>[] {
+    // Initialize all proxies.
+    for (const core of this._automerge.allObjectCores()) {
```
dmaretskyi, line 268: this is way to expensive to be done on this getter

packages/core/echo/echo-db/src/database.ts
```diff
@@ -179,7 +186,7 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
       return undefined;
     }
 
-    const object = defaultMap(this._rootProxies, core, () => initEchoReactiveObjectRootProxy(core));
+    const object = defaultMap(this._rootProxies, core, () => initEchoReactiveObjectRootProxy(core, this));
```
dmaretskyi, line 189: extract `defaultMap(this._rootProxies, core, () => initEchoReactiveObjectRootProxy(core, this));`

packages/core/echo/echo-db/src/database.ts
mykola-vrmchk: This getter should not be used at all

PR #6967, opened 2024-06-11
refactor: Refactor @dxos/echo-db folder structure

mykola-vrmchk:


packages/core/echo/echo-db/src/proxy-db/database.ts
```diff
@@ -106,7 +106,7 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
   /**
    * @internal
    */
-  _automerge: AutomergeDb;
+  _automerge: CoreDatabase;
```
dmaretskyi, line 109: rename field

packages/core/echo/echo-db/src/proxy-db/database.ts
```diff
@@ -278,7 +278,7 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
   /**
    * @deprecated
    */
-  get automerge(): AutomergeDb {
+  get automerge(): CoreDatabase {
```
dmaretskyi, line 281: rename field

packages/core/echo/echo-db/src/proxy-db/database.ts
dmaretskyi: also this could be private since there's a getter for it

packages/core/echo/echo-db/src/core-db/types.ts
```diff
@@ -5,14 +5,41 @@
...
+  | { [key: string]: DecodedAutomergeValue }
+  | Reference
+  | EchoReactiveObject<any>;
 
 //
 // Automerge types.
 // TODO(burdon): Factor out to new low-level type package: @dxos/types or to @dxos/automerge?
 //
 
+export type KeyPath = readonly (string | number)[];
```
dmaretskyi, line 38: it was better when all of those were in their own files

dmaretskyi (approved):

PR #6968, opened 2024-06-11
feat(cli): list schema

yarolegovich:
### Details

Schema-related API changes:
* `db.schemaRegistry` -> `db.schema`
* `schemaRegistry.getAll()` -> `schemaRegistry.listDynamic()`
* Introduced `StaticSchema` type for immutable schema (both runtime and stored)
* Introduced `db.schema.list()` method for retrieving all static schemas

CLI-related changes:
* `dx space schema list` shows space picker if no space provided and lists all schemas.
* `dx space schema list --typename prop` lists schemas with "prop" in typename (case-insensitive).
* `dx space schema list --extended` prints schema json representation.

image

### DIFF
--- a/packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
+++ b/packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
@@ -46,8 +46,8 @@ export const TriggerEditor = ({ space, trigger }: { space: Space; trigger: Funct
   const fn = useMemo(() => query.find((fn) => fn.uri === trigger.function), [trigger.function, query]);
 
   useEffect(() => {
-    void space.db.schemaRegistry
-      .getAll()
+    void space.db.schema
+      .listDynamic()
       .then((schemas) => {
         // TODO(Zan): We should solve double adding of stored schemas in the schema registry.
         state.schemas = distinctBy([...state.schemas, ...schemas], (schema) => schema.typename).sort((a, b) =>
--- a/packages/core/echo/echo-db/src/dynamic-schema-registry.ts
+++ b/packages/core/echo/echo-db/src/dynamic-schema-registry.ts
@@ -72,7 +72,7 @@ export class DynamicSchemaRegistry {
     const storedSnapshots = storedSchemas.map((storedSchema) => {
       const schema = new DynamicEchoSchema(storedSchema);
       return {
-        storedSchemaId: storedSchema.id,
+        id: storedSchema.id,
         typename: schema.typename,
         version: storedSchema.version,
         schema: schema.schema,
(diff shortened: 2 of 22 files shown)


packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
```diff
@@ -46,8 +46,8 @@ export const TriggerEditor = ({ space, trigger }: { space: Space; trigger: Funct
   const fn = useMemo(() => query.find((fn) => fn.uri === trigger.function), [trigger.function, query]);
 
   useEffect(() => {
-    void space.db.schemaRegistry
-      .getAll()
+    void space.db.schema
+      .listDynamic()
```
richburdon, line 50: i think just `get()` (and if necessary pass in options) to match the other APIs.

packages/core/echo/echo-db/src/dynamic-schema-registry.ts
```diff
@@ -61,7 +67,22 @@ export class DynamicSchemaRegistry {
     return this._schemaByType.get(typename);
   }
 
-  public async getAll(): Promise<DynamicEchoSchema[]> {
+  public async list(): Promise<StaticSchema[]> {
+    const { objects: storedSchemas } = await this.db.query(Filter.schema(StoredEchoSchema)).run();
+    const storedSnapshots = storedSchemas.map((storedSchema) => {
+      const schema = new DynamicEchoSchema(storedSchema);
+      return {
+        storedSchemaId: storedSchema.id,
```
richburdon, line 75: just `id`?

richburdon (approved):

jessmartin:
> Introduced `StaticSchema` type for immutable schema (both runtime and stored)

I assume `StaticSchema` can be changed, but only by bumping the version number and optionally writing migration scripts?

> Introduced `db.schema.list()` method for retrieving all static schemas

So in the case of my Contacts app, I should be able to persist the Contacts schema to the database and then query it in Composer?

What work needs to be done to Composer's tables to support these new APIs? I assume that by default Table creates a dynamic schema, but we probably want table to support static schemas as well. In that case, you wouldn't be able to add/remove columns.

yarolegovich:
> I assume StaticSchema can be changed, but only by bumping the version number and optionally writing migration scripts?

Static here means it's a snapshot of what was registered at the moment of query, those are not "live" objects you can mutate for changes to be reflected in any other place.

> So in the case of my Contacts app, I should be able to persist the Contacts schema to the database and then query it in Composer?

Unfortunately, not yet. That's just some quick changes for space inspection tooling.

packages/apps/plugins/plugin-function/src/components/TriggerEditor.tsx
yarolegovich: return types are different, agreed to leave as is

Issue #6980, opened 2024-06-12
[deck] make new plank position configurable and default to "before"


zhenyasav:

ZaymonFC:
The domain terms I think we should use are `start` and `end`.

PR #6985, opened 2024-06-12
feat(deck): Various improvements

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/6981
- Resolves https://github.com/dxos/dxos/issues/6980


packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -323,27 +325,38 @@ export const DeckPlugin = ({
...
+                                  prev.forEach((id) => partMembers.add(id));
+                                  break;
+
+                                case 'end':
+                                  prev.forEach((id) => partMembers.add(id));
+                                  newIds.forEach((id) => partMembers.add(id));
+                                  break;
+
+                                default:
+                                  throw new Error(`Unhandled Case of NewPlankPosition: ${newPlankPositioning}`);
```
wittjosiah, line 355: does this need to throw or could it just pass through? if it can pass through you could just `log.error` instead and if it started happening regularly we'd see it in sentry

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -197,31 +197,7 @@ const NodePlankHeading = ({
           )
         }
         close={part[0] === 'complementary' ? 'minify-end' : true}
-      >
-        {/* TODO(thure): This, and all other hardcoded `comments` references, needs to be refactored. */}
-        {node && !!node.data?.comments && !slug?.endsWith('comments') && (
```
wittjosiah, line 202: Were you able to verify that this button doesn't serve a purpose anymore (i.e. there's another button to open the complementary sidebar & it's content changes based on attention)?

wittjosiah (approved):

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
ZaymonFC: Good point.

PR #6987, opened 2024-06-13
refactor: Change SpaceKey -> SpaceId inside database

mykola-vrmchk:


packages/core/echo/echo-db/src/hypergraph.ts
```diff
@@ -33,14 +33,18 @@ import { RuntimeSchemaRegistry } from './runtime-schema-registry';
  * Manages cross-space database interactions.
  */
 export class Hypergraph {
-  private readonly _databases = new ComplexMap<PublicKey, EchoDatabaseImpl>(PublicKey.hash);
+  /**
+   * Used for References resolution.
+   * @deprecated use SpaceId.
+   * TODO(mykola): Delete on References migration.
+   */
+  private readonly _spaceKeyVsId = new ComplexMap<PublicKey, SpaceId>(PublicKey.hash);
```
dmaretskyi, line 41:
```suggestion
  private readonly _spaceKeyToId = new ComplexMap<PublicKey, SpaceId>(PublicKey.hash);
```

dmaretskyi (approved):

packages/core/echo/echo-pipeline/src/space/space.ts
```diff
@@ -210,21 +210,21 @@ export class Space extends Resource {
...
 /**
  * Space keys are generated by creating a keypair, and then taking the first 20 bytes of the SHA-256 hash of the public key and encoding them to multibase RFC4648 base-32 format (prefixed with B, see Multibase Table).
  * Inspired by how ethereum addresses are derived.
  */
 export const createIdFromSpaceKey = async (spaceKey: PublicKey): Promise<SpaceId> => {
-  // Sanity check.
-  invariant(spaceKey.length === EXPECTED_SPACE_KEY_LENGTH);
+  if (SPACE_IDS_CACHE.has(spaceKey)) {
+    return SPACE_IDS_CACHE.get(spaceKey)!;
+  }
```
dmaretskyi, line 222:
```suggestion
  const cachedValue = SPACE_IDS_CACHE.get(spaceKey);
  if (cachedValue !== undefined) {
    return cachedValue;
  }
```

this does 2 map lookups instead of 1

PR #6996, opened 2024-06-13
chore: tweak composer telemetry

wittjosiah:
- only send navigation events for new ids in deck
- update navigation typename data
- add events to key space plugin actions
- add identity creation/share events
- reduce the frequency of space metric events
- only send window click events for id'd elements


ZaymonFC (approved):
LGTM

PR #7001, opened 2024-06-14
feat!: Migrate reference storage format to use new DXNs

dmaretskyi:
### TODO

- [x] Prevent indexer from reading old documents
- [x] SpaceProxy.createEpoch should wait for the new epoch and the new automerge root
- [x] Typenames are getting serialized to the wrong DXN kind: ` '@type': { '/': 'dxn:echo:@:dxos.org/type/Properties' },`
- [x] Make sure document version gets set for all new documents (roots & leafs)


packages/common/keys/src/dxn.ts
```diff
@@ -0,0 +1,59 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { invariant } from '@dxos/invariant';
+
+export class DXN {
```
wittjosiah, line 7: would be good to include jsdoc on the class and include the basic spec

packages/common/keys/src/dxn.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (!(typeof kind === 'string' && kind.length > 0)) {
+      throw new Error('Invalid DXN');
+    }
+    if (!(parts.length > 0)) {
+      throw new Error('Invalid DXN');
+    }
+    return new DXN(kind, parts);
+  }
+
+  #kind: string;
```
wittjosiah, line 30: not seen this hash syntax before, what does this do?

packages/core/echo/echo-db/src/serializer.ts
```diff
@@ -155,21 +166,28 @@ export class Serializer {
       ...data,
       '@version': Serializer.version,
       '@meta': meta,
-      '@timestamp': new Date().toUTCString(),
+      '@timestamp': new Date().toISOString(),
```
wittjosiah, line 169: is this just the timestamp of when it was serialized?

packages/core/echo/echo-protocol/src/space-doc-version.ts
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+/**
+ * Denotes the data version of the space document.
```
wittjosiah, line 6: document here referring to "automerge document"?

packages/core/echo/indexing/src/index-schema.ts
```diff
@@ -4,7 +4,7 @@
 
 import { Event } from '@dxos/async';
 import { Resource } from '@dxos/context';
-import { type ObjectStructure } from '@dxos/echo-protocol';
+import { decodeReference, type ObjectStructure } from '@dxos/echo-protocol';
 import { EXPANDO_TYPENAME } from '@dxos/echo-schema';
```
wittjosiah, line 8: does this typename need to be updated as well?

packages/common/keys/src/dxn.ts
dmaretskyi: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

packages/core/echo/indexing/src/index-schema.ts
dmaretskyi: cc @richburdon 

Currently it is:

```ts
export const EXPANDO_TYPENAME = 'Expando';
```

wittjosiah: presumably it should be `dxos.org/type/Expando` and we might as well do that at the same time as the space properties

packages/apps/types/src/serializer/serializers/default.ts
```diff
@@ -87,7 +87,7 @@ const deserializeEchoObject = (parsed: any, newId?: boolean): Expando => {
     core.id = id;
   }
   getMeta(deserializedObject).keys.push(...(meta?.keys ?? []));
-  const typeRef = getTypeRef(type);
+  const typeRef = internalDecodeReference(type);
```
wittjosiah, line 90: maybe todo to remove usage? I think this was the only external usage of the api?

packages/gravity/proto-guard/src/check-snapshot.test.ts
```diff
@@ -30,35 +31,59 @@ describe('Load client from storage snapshot', () => {
     return testStoragePath;
   };
 
-  test('check if space loads for Automerge on nodeFS snapshot', async () => {
-    const snapshot = SnapshotsRegistry.getSnapshot('automerge-nodeFS');
-    invariant(snapshot, 'Snapshot not found');
-    log.info('Testing snapshot', { snapshot });
+  test
+    .skip('check if space loads for Automerge on nodeFS snapshot', async () => {
```
wittjosiah, line 35: why is this test being skipped?

packages/sdk/client/src/echo/space-proxy.ts
```diff
@@ -435,6 +451,12 @@ export class SpaceProxy implements Space {
     });
   }
 
+  private async _migrate() {
```
wittjosiah, line 454: why this vs space.internal.migrate?

packages/gravity/proto-guard/src/check-snapshot.test.ts
dmaretskyi: its getting removed in https://github.com/dxos/dxos/pull/7031

packages/sdk/client/src/echo/space-proxy.ts
dmaretskyi: This is exposed as `space.internal.migrate`

PR #7004, opened 2024-06-14
feat: make deck the default layout

wittjosiah:
Inverts the deck layout settings such that deck is the default layout, but leaves setting for now which allows it to be disabled.


thure (approved):
✨

ZaymonFC (approved):
✨

PR #7005, opened 2024-06-14
fix(shell): profile updates not persisted

wittjosiah:
Changes to the profile were changing in the UI because of stale values in callbacks.


thure (approved):
lgtm, thanks!

Issue #6969, opened 2024-06-11
[composer] switch to space.id for graph identifiers in composer


wittjosiah:
Currently composer uses `space.key` but this is now a deprecated api.

ZaymonFC:
@wittjosiah I'm assuming this issue includes all plugins?

PR #7006, opened 2024-06-14
chore(composer): Change usages of `space.key` -> `space.id` in composer graph identifiers

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/6969

Changes:
- **Change usages of `space.key` in graph**
- **Switch `SpacePlugin.enabled` to use space ids instead of keys**


wittjosiah (commented):
I think one file you missed is the util file in space plugin which is what adds the spaces to the graph

ZaymonFC:
> I think one file you missed is the util file in space plugin which is what adds the spaces to the graph

Good catch!

wittjosiah (approved):

PR #7007, opened 2024-06-17
feat(echo): flag for lazy opening of spaces

yarolegovich:
### Details

* Presence subclasses a resource and updates need to be explicitly scheduled (open vs start in constructor).
* `lazy_space_open` flag in `Config`.
* `DefaultSpace` credential is written to halo when a default space is created.
* `DataSpaceManager` never opens spaces, this is now handled by `SpaceListProxy` in accordance with `lazy_space_open` flag.


packages/core/mesh/teleport-extension-gossip/src/presence.ts
```diff
@@ -65,6 +61,18 @@ export class Presence {
       this._receiveAnnounces(message);
     });
 
+    // Remove peer state when connection is closed.
+    this._params.gossip.connectionClosed.on((peerId) => {
```
dmaretskyi, line 65: Will route errors and unsubscribe automatically.

```suggestion
    this._params.gossip.connectionClosed.on(this._ctx, (peerId) => {
```

Although this will make this class non-reentrant (since the context would be different on the second open) so it's better to move this subscription into `_open` -- which is right anyway since you don't want the callback to be called on a closed Presence

packages/core/protocols/src/proto/dxos/halo/credentials.proto
```diff
@@ -181,6 +181,11 @@ message DeviceProfile {
   DeviceProfileDocument profile = 1;
 }
 
+/// [ASSERTION]: Sets default space for identity.
+message DefaultSpace {
+  dxos.keys.PublicKey space_key = 1;
```
dmaretskyi, line 186: string spaceId

dmaretskyi: lets prefer space ids everywhere outside of internal halo state machiene when referring to a space

packages/sdk/client-services/src/packlets/spaces/data-space-manager.test.ts
```diff
@@ -168,5 +168,50 @@ describe('DataSpaceManager', () => {
...
+    });
   });
+
+  const reloadDataSpaces = async (peer: TestPeer) => {
+    await peer.dataSpaceManager.close();
+    await peer.dataSpaceManager.open();
+  };
+
+  const getFirstSpace = (peer: TestPeer) => {
+    return [...peer.dataSpaceManager.spaces.values()][0];
```
dmaretskyi, line 215: just as a code consistency thing:

```suggestion
    return Array.from(peer.dataSpaceManager.spaces.values())[0];
```

packages/sdk/client-services/src/packlets/identity/default-space-state-machine.ts
```diff
@@ -0,0 +1,40 @@
...
+export type ProfileStateMachineParams = {
+  identityKey: PublicKey;
+  onUpdate?: () => void;
+};
+
+/**
+ * Processes device invitation credentials.
+ */
+export class DefaultSpaceStateMachine implements CredentialProcessor {
+  private _spaceKey: PublicKey | undefined;
```
dmaretskyi, line 19: spaceId

packages/sdk/client-services/src/packlets/identity/default-space-state-machine.ts
```diff
@@ -0,0 +1,40 @@
+//
+// Copyright 2022 DXOS.org
+//
+
+import { type CredentialProcessor, getCredentialAssertion } from '@dxos/credentials';
+import { type PublicKey } from '@dxos/keys';
+import { log } from '@dxos/log';
+import { type Credential } from '@dxos/protocols/proto/dxos/halo/credentials';
+
+export type ProfileStateMachineParams = {
```
dmaretskyi, line 10: wrong name?

packages/sdk/client-services/src/packlets/identity/identity-service.ts
```diff
@@ -9,30 +9,50 @@ import { invariant } from '@dxos/invariant';
...
 import { type CreateIdentityOptions, type IdentityManager } from './identity-manager';
+import { type DataSpaceManager } from '../spaces';
 
 export class IdentityServiceImpl implements IdentityService {
   constructor(
     private readonly _createIdentity: (params: CreateIdentityOptions) => Promise<Identity>,
+    private readonly _dataSpaceManagerProvider: () => DataSpaceManager,
     private readonly _identityManager: IdentityManager,
     private readonly _keyring: Keyring,
     private readonly _onProfileUpdate?: (profile: ProfileDocument | undefined) => Promise<void>,
```
dmaretskyi, line 31:
```suggestion
    private readonly _identityManager: IdentityManager,
    private readonly _keyring: Keyring,
    private readonly _dataSpaceManagerProvider: () => DataSpaceManager,
    private readonly _createIdentity: (params: CreateIdentityOptions) => Promise<Identity>,
    private readonly _onProfileUpdate?: (profile: ProfileDocument | undefined) => Promise<void>,
```

packages/sdk/client-services/src/packlets/identity/identity-service.ts
```diff
@@ -9,30 +9,50 @@ import { invariant } from '@dxos/invariant';
...
   ) {}
 
-  async createIdentity(request: CreateIdentityRequest): Promise<Identity> {
+  async open() {
```
dmaretskyi, line 34: extends Resource?

dmaretskyi: lets do that every time we need open/close

dmaretskyi: will save us for acciedental idempotence & race-condition bugs

packages/sdk/client/src/testing/utils.ts
```diff
@@ -38,3 +44,32 @@ export const waitForSpace = async (
...
+  testBuilder.storage = options?.storage ? createStorage({ type: StorageType.RAM }) : undefined;
+  testBuilder.level = options?.storage ? createTestLevel() : undefined;
+
+  const clients = range(
+    count,
+    () =>
+      new Client({ config: options?.config, services: testBuilder.createLocalClientServices(options?.serviceConfig) }),
+  );
+  const initialized = await Promise.all(
+    clients.map(async (c, index) => {
```
dmaretskyi, line 67:
```suggestion
    clients.map(async (client, index) => {
```

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -211,6 +200,53 @@ export class DataSpaceManager {
     return space;
   }
 
+  async isDefaultSpace(space: DataSpace): Promise<boolean> {
+    const rootDoc = await this._getSpaceRootDocument(space);
+    for (const object of Object.values(rootDoc.docSync()?.objects ?? {}) as ObjectStructure[]) {
```
dmaretskyi, line 205: There's a function for that: `findPropertiesObject`

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -211,6 +200,53 @@ export class DataSpaceManager {
...
+    const space = await this.createSpace();
+    const document = await this._getSpaceRootDocument(space);
+
+    // TODO(dmaretskyi): Better API for low-level data access.
+    const properties: ObjectStructure = {
+      system: {
+        type: encodeReference(getTypeReference(PropertiesType)!),
+      },
+      data: {
+        [defaultKey]: this._signingContext.identityKey.toHex(),
```
dmaretskyi, line 226: is this for backwards compatibility? i suppose we woudn't need this with the new credential

yarolegovich: backward compatibility and was also thinking about using this for recovering from a case when a default space was created, but credential writing failed. we won't create another space but just write the missing credential on the next open. this however doesn't help if space was created but properties mutation failed, so not strongly against getting rid of it

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
yarolegovich: nice

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -211,6 +200,46 @@ export class DataSpaceManager {
     return space;
   }
 
+  async isDefaultSpace(space: DataSpace): Promise<boolean> {
+    const rootDoc = await this._getSpaceRootDocument(space);
+    const [_, properties] = findPropertiesObject(rootDoc.docSync()) ?? [];
+    return properties?.data?.[defaultKey] === this._signingContext.identityKey.toHex();
```
richburdon, line 206: change to `defaultSpaceKey`? (defaultKey could mean many things).

but, could spaces have some generic attributes (e.g., tags?) -- that other layers of the system can interpret as a "default space" -- rather than tying together these concepts directly? E.g., in Gmail, folders are represented as tags on a message, and there are special "system" tags for Inbox, Sent, Deleted, etc. So Sent is treated as a "special" folder by some subsystems, but the folders are all basically the same (just have different tags). It's decoupled and more flexible. (e.g., later perhaps we want "hidden" spaces, or "transient" (in-memory) spaces, etc.

PR #7008, opened 2024-06-17
chore: Export lint rules

richburdon:


wittjosiah (commented):
if you're going to publish this, it should probably be in `packages` and not `tools` which only internal monorepo tooling

richburdon:
> if you're going to publish this, it should probably be in `packages` and not `tools` which only internal monorepo tooling

i think it makes sense in tools since it's just to share across other monorepos.

PR #7012, opened 2024-06-17
fix: Update debug plugin

richburdon:


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -240,6 +240,7 @@ export const SpacePlugin = ({
           spaceSubscriptions.clear();
           client.spaces
             .get()
+            // TODO(burdon): Load all?
```
wittjosiah, line 243: what do you mean?

packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
```diff
@@ -137,47 +111,15 @@ export const DebugPlugin = (): PluginDefinition<DebugPluginProvides> => {
...
           return () => {
-            unsubscribe();
             subscriptions.clear();
           };
         },
       },
       intent: {
         resolver: async (intent, plugins) => {
           switch (intent.action) {
+            // TODO(burdon): Remove?
```
wittjosiah, line 122: I think yes

PR #7018, opened 2024-06-18
feat(plugin-space): add space names cache

wittjosiah:
Adds local storage cache of names to the space plugin to allow space names to be rendered even if a space isn't open currently.

Also moves the close space action to it's own condition so that spaces stuck loading can be closed.


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -154,8 +161,19 @@ export const SpacePlugin = ({
         await onFirstRun?.({ client, dispatch });
       }
 
+      // Cache space names.
+      subscriptions.add(
+        client.spaces.subscribe((spaces) => {
+          spaces
+            .filter((space) => space.state.get() === SpaceState.READY)
+            .forEach((space) => {
+              state.values.spaceNames[space.id] = space.properties.name;
```
dmaretskyi, line 170: you also wanna subscribe to the properties echo object itself

PR #7025, opened 2024-06-18
fix: sketch serializer

### LINKED ISSUE #6749: [composer] drag-and-drop between spaces is broken for sketches

wittjosiah:
During the clone operation only the sketch is serialized and the data isn't included.
Clone needs to not preserve ids right now so that you can copy things back and forth
between spaces and not have it fail because of id collision. Not preserving ids
makes references complicated because you need to update all the references.

The document serializer works because it collapses the references into a flat thing
(markdown) and then rebuilds the object structure afterwards. This does the same thing
but using json for sketch objects.

Resolves #6749

### DIFF
--- a/packages/apps/types/src/serializer/serializers/sketch.ts
+++ b/packages/apps/types/src/serializer/serializers/sketch.ts
@@ -15,11 +15,12 @@ export const serializer: TypedObjectSerializer<SketchType> = {
 
   serialize: async ({ object }): Promise<string> => {
     // TODO(wittjosiah): Implement `toJSON` method?
-    const sketch = { ...object, data: { ...object.data } };
+    const data = await loadObjectReferences(object, (s) => s.data);
+    const sketch = { ...object, data: { ...data } };
     return JSON.stringify(sketch, null, 2);
   },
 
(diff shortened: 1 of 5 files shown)


packages/apps/types/src/serializer/serializers/sketch.ts
```diff
@@ -12,10 +14,41 @@ export const serializer: TypedObjectSerializer<SketchType> = {
   }),
 
   serialize: async ({ object }): Promise<string> => {
-    throw new Error('Not implemented.');
+    // TODO(wittjosiah): Implement `toJSON` method?
+    const sketch = { ...object, data: { ...object.data } };
```
yarolegovich, line 18: I'd use `loadObjectReferences` on `object.data` first.

PR #7030, opened 2024-06-18
chore(plugin-stack): use fully qualified id for stack collapse

wittjosiah:


jessmartin (approved):
A++++ fast shipping would buy again

PR #7033, opened 2024-06-19
refactor: Reconcile EchoHandler link reference methods

mykola-vrmchk:
What changed in cyclic references handling:

Test: 
```
 task = create(Expando, { title: 'test' });
 task.previous = create(Expando, { title: 'another' });
 task.previous!.previous = task;
 db.add(task);
```

Before:
1. db.add would create EchoReactiveObject from `task` which is ReactiveObject
2. Creation would recursively link all fields as they are (as ReactiveObjects)
3. Init automerge doc for an object
4. save all linked objects triggering `db.add` with `task.previous`

After:
1. db.add would create EchoReactiveObject from `task` which is ReactiveObject
2. Creation would recursively link all fields and make sure they are EchoReactiveObject-s (if not it would create echo objects respectively)
  2.1 Creation of EchoReactiveObject for references would trigger properties validation
  2.2 Properties validation would access all properties recursively
  2.3 it would access circular reference
  2.4 it would try to decode automerge doc (not present yet so I changed behaviour so `undefined` here do not crash everything)
3. Init automerge doc for an object
4. db.add then would link to a parent object and save all linked objects triggering `db.add` with `task.previous`


packages/core/echo/echo-db/src/echo-handler/echo-handler.ts
```diff
@@ -472,35 +472,37 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
 
   /**
    * Store referenced object.
+   * Used when `set` and other mutating methods are called with a proxy.
+   * @param target - self
+   * @param proxy - the proxy that was passed to the method
    */
-  linkObject(target: ProxyTarget, obj: EchoReactiveObject<any>): Reference {
+  linkObject(target: ProxyTarget, proxy: any): Reference {
```
dmaretskyi, line 479: what about _linkReactiveHandler method?

packages/core/echo/echo-db/src/echo-handler/echo-handler.ts
```diff
@@ -472,35 +472,37 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
...
     const database = target[symbolInternals].database;
     if (database) {
-      // TODO(dmaretskyi): Fix this.
-      if (isReactiveObject(obj) && !isEchoObject(obj)) {
-        database.add(obj);
-      }
-
-      const foreignDatabase = (getProxyHandlerSlot(obj).target as ProxyTarget)[symbolInternals].database;
+      const foreignDatabase = (getProxyHandlerSlot(otherEchoObj).target as ProxyTarget)[symbolInternals].database;
       if (!foreignDatabase) {
```
dmaretskyi, line 487: pls rewrite this to reduce nesting and add comments

packages/core/echo/echo-db/src/core-db/load-object-references.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * EXPERIMENTAL - the API is subject to change.
+ * @param objOrArray - an echo object or collection of objects with references to other echo objects.
+ * @param valueAccessor - selector for a reference that needs to be loaded.
+ *                        if return type is an array the method exits when all entries are non-null.
+ *                        otherwise the method exits when valueAccessor is not null.
+ * @param timeout - loading timeout, defaults to 5s.
+ */
+export const loadObjectReferences = async <
```
dmaretskyi, line 18: should go to proxy db since it deals with proxies

dmaretskyi (approved):

PR #7040, opened 2024-06-20
[docs] Import create function from @dxos/client/echo

michielbdejong:
It seems the `create` function was undefined in this snippet and @dxos/client/echo does have an export by that name, so that's probably the one to use?


wittjosiah:
@michielbdejong Thanks for opening this. It looks like there's a bunch more issues with the snippets. They're supposed to be type-checked but it seems that broke at some point. I've filed this issue and we'll try to get this sorted for the next release https://github.com/dxos/dxos/issues/7041

PR #7056, opened 2024-06-22
fix: Identity panel styles.

richburdon:
Minor fix to identity controls.


packages/apps/plugins/plugin-grid/package.json
```diff
@@ -31,6 +31,8 @@
     "@dxos/react-client": "workspace:*",
     "@dxos/util": "workspace:*",
     "@preact/signals-core": "^1.6.0",
+    "grid": "^4.10.8",
```
thure, line 34: This package hasn’t been updated in 3 years and doesn’t appear to be very popular… if this truly works for `plugin-grid`, we may want to consider taking ownership of the package’s source code.

thure: …come to think of it, was it an accident that `plugin-grid` dependencies became part of this PR?

richburdon: yes this was leakage from another PR. i'll remove.

PR #7057, opened 2024-06-22
fix: Rename SketchType to more general DiagramType

richburdon:
Need to set the schema property for tldraw.


wittjosiah:
closing because would require a separate migration at this point, may revisit later

wittjosiah:
planning to land after manually reverting migrations on the few spaces which were migrated

Issue #6083, opened 2024-03-27
[presenter] stack does not enter presentation mode when it is not navigated to


zhenyasav:
if you're on some item and not the stack, but you click "present" from the actions of a stack in the nav tree, it will enter "full screen" but will not engage the presenter, the stack will just appear normally without sidebars and other chrome.

Issue #5192, opened 2024-01-13
[presenter] no way to exit fullscreen mode with pointer


zhenyasav:
I hit esc while in presenter mode and found myself looking at the stack with no sidebars or anything else

https://github.com/dxos/dxos/assets/3398896/e4f5d394-d115-4e16-825d-029d1057635f

ZaymonFC:
- Resolved by #7468

PR #7058, opened 2024-06-22
feat: Enable presenter plugin by default

richburdon:


wittjosiah:
probably should resolve at least https://github.com/dxos/dxos/issues/5192 before enabling by default with a playwright test ensure the core functionality doesn't break again

ideally https://github.com/dxos/dxos/issues/6083 as well

PR #7059, opened 2024-06-23
chore: Merge labs

richburdon:


packages/apps/plugins/plugin-sketch/package.json
```diff
@@ -46,6 +46,7 @@
     "@tldraw/tldraw": "^2.2.1",
     "@tldraw/tlschema": "^2.2.1",
     "@tldraw/utils": "^2.2.1",
+    "classnames": "^2.3.2",
```
wittjosiah, line 49: this shouldn't be needed, just use `mx` from react ui

packages/apps/plugins/plugin-sketch/src/components/custom/CustomGrid.tsx
wittjosiah, line 1: dupe

packages/apps/plugins/plugin-sketch/src/hooks/schema.ts
wittjosiah, line 1: dupe

packages/apps/types/src/schema/thread.ts
```diff
@@ -28,14 +29,14 @@ export class MessageType extends TypedObject({ typename: 'dxos.org/type/Message'
   text: S.String,
   /** Non-text content sent with a message (e.g. files, polls, etc.) */
   parts: S.optional(S.mutable(S.Array(ref(Expando)))),
-  /** Context of the application when message was created. */
-  // TODO(burdon): Evolve "attention object" to be current UX state? E.g., of Deck?
-  context: S.optional(ref(Expando)),
   /** Custom properties for specific message types (e.g. email subject or cc fields). */
   properties: S.optional(S.mutable(S.Record(S.String, S.Any))),
   // TODO(wittjosiah): Add read status:
```
wittjosiah, line 34: this todo should probably stay at the end otherwise it reads as though it applies to the field below it

packages/core/echo/echo-db/src/proxy-db/load-object.test.ts
```diff
@@ -1,14 +1,15 @@
...
 
-import { type EchoReactiveObject, Expando, create, TypedObject, S, ref } from '@dxos/echo-schema';
+import { create, type EchoReactiveObject, Expando, ref, S, TypedObject } from '@dxos/echo-schema';
 import { describe, test } from '@dxos/test';
 
-import { loadObjectReferences } from './load-object-references';
+import { loadObjectReferences } from './load-object';
 import { TestBuilder } from '../testing';
 
+// Copyright 2024 DXOS.org
```
wittjosiah, line 10: ??

packages/core/echo/echo-db/src/proxy-db/load-object.ts
```diff
@@ -7,6 +7,14 @@ import { type EchoReactiveObject } from '@dxos/echo-schema';
 
 import { getObjectCore } from '../core-db';
 
+/**
+ * @param obj
+ */
+// TODO(burdon): Rename/review SDK.
+export const loadObject = <T extends EchoReactiveObject<any>>(obj: T): T => {
```
wittjosiah, line 14: is this expecting a reference to be passed in? if you passed in a object isn't it already loaded?

packages/core/functions/src/types.ts
```diff
@@ -97,7 +97,8 @@ export class FunctionTrigger extends TypedObject({
   enabled: S.optional(S.Boolean),
   function: S.String.pipe(S.description('Function URI.')),
   // The `meta` property is merged into the event data passed to the function.
-  meta: S.optional(S.mutable(S.Any)),
+  // TODO(burdon): Record.
```
wittjosiah, line 100: done

packages/experimental/agent-functions/src/functions/gpt/response.ts
```diff
@@ -52,7 +52,8 @@ export class ResponseBuilder {
     //
     // Add to collection.
     //
-    if (this._context.object instanceof CollectionType) {
+    const stack = false;
```
wittjosiah, line 55: todo?

tsconfig.json
```diff
@@ -24,5 +24,15 @@
   "ts-node": {
     "swc": true,
   },
-  "exclude": ["dist", "node_modules", "out", "**/*.mts"],
+  "include": [
```
wittjosiah, line 27: I believe this was intentionally removed

packages/core/echo/echo-db/src/echo-handler/echo-handler.test.ts
```diff
@@ -344,6 +353,33 @@ describe('Reactive Object with ECHO database', () => {
...
+
+      // References serialized as IPLD.
+      {
+        const obj = JSON.parse(JSON.stringify(obj2));
+        expect(decodeReference(obj.nested).itemId).to.eq(obj2.nested?.id);
+      }
+
+      // Load refs.
+      {
+        const nested = await loadObjectReferences(obj2, (obj) => obj.nested);
```
dmaretskyi, line 378: this would have no effect in this test because references are already loaded since you've only just addedthe object to the database

packages/sdk/client/src/testing/data.ts
```diff
@@ -28,18 +28,19 @@ export class ContactType extends TypedObject({ typename: 'braneframe.Contact', v
   ),
 }) {}
 
-const _BlockSchema = S.Struct({
+const BlockSchema = S.Struct({
```
dmaretskyi, line 31: why not class?

packages/sdk/client/src/testing/data.ts
```diff
@@ -28,18 +28,19 @@ export class ContactType extends TypedObject({ typename: 'braneframe.Contact', v
...
   timestamp: S.String,
   content: S.optional(ref(TextV0Type)),
   object: S.optional(ref(Expando)),
 });
-export interface BlockType extends S.Schema.Type<typeof _BlockSchema> {}
+
+export interface BlockType extends S.Schema.Type<typeof BlockSchema> {}
```
dmaretskyi, line 37:
```suggestion
export const BlockType: S.Schema<BlockType> = BlockSchema;
```

packages/core/echo/echo-db/src/proxy-db/load-object.ts
dmaretskyi: I don't think this name makes sense. Let's discuss if this should be a part of the API

packages/apps/plugins/plugin-sketch/src/hooks/schema.ts
richburdon: i don't see the dupe.

packages/core/echo/echo-db/src/proxy-db/load-object.ts
richburdon: see the use case below?

tsconfig.json
richburdon: can you explain. @dmaretskyi ?

packages/core/echo/echo-db/src/echo-handler/echo-handler.test.ts
richburdon: we don't have a way to test eviction yet right?

packages/apps/plugins/plugin-sketch/src/hooks/schema.ts
wittjosiah: it was moved to braneframe/types with the migrations

tsconfig.json
dmaretskyi: This was changed 1-2 weeks ago, the idea being that the tsconfig files from individual projects should be including sources files, while the root one should include nothing. In thereory that should have improved the performance of tsserver, but personally it's still quite unstable for me

PR #7060, opened 2024-06-23
feat(deck): fix presence in plank headings

### LINKED ISSUE #7021: [deck] fix presence in plank headings
- clicking on avatars seems to open comments of a document in a plank but should open the object that peer is currently attending
- deck changes the "active" object dynamics, currently presence is showing "online" as a green ring and "has this object open" as a blue ring. Showing "online" members in every plank heading is a lot of duplication so that should go away and presence in the space heading should distinguish between "open" (as a grey ring?) and "attended" (as a blue ring?)
- the presence status being indicated only via coloured ring isn't ideal either but more design work is needed on that and changes there are out of scope for this issue

ZaymonFC:
- Partial #7021

This also changes how the presence system works in the old layout, but since we're making deck the default I think that's okay for now?

Minimal changes made until we do a full design pass on this.

- All member icons are now 'blue ringed' to indicate `present and viewing the current object`.
- Removed click action since it doesn't make sense to dispatch a navigation action to the currently open document.


wittjosiah (approved):
So this doesn't fully resolve #7021 because it's not showing the "open"/"idle" state. However it should help reduce the repetitiveness/brokeness feeling of the presence, so I think it still makes sense to merge, but perhaps that last thing needs to be broken out into it's own issue.

PR #7076, opened 2024-06-24
feat: Http param processor

richburdon:
typesafe http param parsing (effect demo)


packages/common/util/src/params.test.ts
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import * as S from '@effect/schema/Schema';
+import { expect } from 'chai';
+
+import { test } from '@dxos/test';
```
dmaretskyi, line 8: since `@dxos/util` is configured with vitest you should import those DSL methods from `vitest` from now on

packages/sdk/shell/src/panels/JoinPanel/JoinPanel.tsx
```diff
@@ -1,24 +1,24 @@
...
-// Copyright 2023 DXOS.org
-//
 import React, { useCallback, useEffect, useMemo } from 'react';
 
 import { log } from '@dxos/log';
 import { useClient, useMulticastObservable } from '@dxos/react-client';
 import { useIdentity } from '@dxos/react-client/halo';
 import { DensityProvider, useId, useThemeContext } from '@dxos/react-ui';
 
+// Copyright 2023 DXOS.org
```
dmaretskyi, line 9: !!!

PR #7077, opened 2024-06-25
feat(presence): Presence should leverage attended items

### LINKED ISSUE #7074: [attention] Attention should be a standalone plugin
Currently space presence is implemented at the space-plugin level. With `deck` becoming the default we'd like to be able to show:

- Who's looking at this plank / object actively?
- Who has this plank / object in their broader `awareness`?

It doesn't make sense to bring attention into the space plugin. At the same time `DeckPlugin` shouldn't care about attention. We need an orthogonal chunk of functionality in place.

Object presence should be driven from the new attention plugin, not the space plugin.

Related:
- #7060
- #5698

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/7021
- Resolves #7074 
- New plugin `AttentionPlugin`
- New package `react-ui-attention`
- Resolves various packages needing to depend on `react-ui-deck`


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -81,7 +82,7 @@ import {
 } from './types';
 import { SHARED, updateGraphWithSpace } from './util';
 
-const ACTIVE_NODE_BROADCAST_INTERVAL = 30_000;
+const ACTIVE_NODE_BROADCAST_INTERVAL = 30000;
```
thure, line 85: but… formatting!

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -223,31 +227,45 @@ export const SpacePlugin = ({
...
+              const parts = Array.from(activeIds(location.active)).filter((part) => part !== undefined);
 
+              // Group parts by space for efficient messaging.
+              const partsBySpace = reduceGroupBy(parts, (part) => {
                 const [spaceId] = part.split(':');
-                const spaces = client.spaces.get();
+                return spaceId;
+              });
+
+              // NOTE: Ensure all spaces are included so that we send the correct `removed`` object arrays.
```
thure, line 241:
```suggestion
              // NOTE: Ensure all spaces are included so that we send the correct `removed` object arrays.
```

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -266,21 +284,26 @@ export const SpacePlugin = ({
             .forEach((space) => {
               spaceSubscriptions.add(
                 space.listen('viewing', (message) => {
-                  const { added, removed } = message.payload;
+                  const { added, removed, attended } = message.payload;
```
thure, line 287: Very tidy, I like it.

packages/ui/react-ui-attention/src/components/AttentionProvider.tsx
```diff
@@ -59,6 +59,7 @@ const getAttendables = (selector: string, cursor: Element, acc: string[] = []):
   }
 };
 
+// TODO(Zan): There's a case to be made for moving this to a new package `react-ui-attention`
```
thure, line 62: Agreed this is much more comfortable here.

thure (approved):
Very nice! Looking forward to this.

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
ZaymonFC: I'm thinking we should bring this interval down, especially now that I'm batching updates per space. CC @wittjosiah? Would love to make it 3 or 5 seconds.

wittjosiah: it was made 30s because we were trying to reduce the noise on the network at one point. Not sure if that's still an issue or not (cc @nferch )

packages/apps/plugins/plugin-layout/src/components/MainLayout.tsx
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 
 import { Surface, type Toast as ToastSchema } from '@dxos/app-framework';
 import { Button, Main, Dialog, useTranslation, DensityProvider, Popover, Status } from '@dxos/react-ui';
-import { useAttendable } from '@dxos/react-ui-deck';
+import { useAttendable } from '@dxos/react-ui-attention';
```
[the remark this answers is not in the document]
ZaymonFC: Yeah I removed the dependency.

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
ZaymonFC: I'm happy to address this as another issue. In my video demo I was using 3000ms.

PR #7078, opened 2024-06-25
fix(plugin-help): Joyride interactivity

### LINKED ISSUE #5990: [composer] fix joyride for first time users

thure:
This PR:
- Resolves #5990
- Polishes the style, accessibility, keyboard access, and interactivity of the Joyride tooltip
- Adds accessors and listeners for shell display

https://github.com/dxos/dxos/assets/855039/c24d5e5b-6964-445b-9da9-cb70ac786999


ZaymonFC (approved):
Nice work. The shell changes look compatible and read-only so I'm not worried there.

packages/apps/composer-app/src/main.tsx
```diff
@@ -139,11 +139,13 @@ const main = async () => {
       isDeck ? DeckMeta : LayoutMeta,
       NavTreeMeta,
       SettingsMeta,
-      HelpMeta,
       StatusBarMeta,
 
-      // Data integrations
+      // Shell and help
```
wittjosiah, line 144: I'm not sure I understand the significance of this group

wittjosiah (approved):

packages/apps/composer-app/src/main.tsx
thure: Help now needs shell (so it knows when it’s open), but shell was part of the group below it and help is not a data persistence plugin… I’m amenable to other ideas if you have any.

wittjosiah: does help need to come after client in order to get access to shell?

thure: Yes exactly

wittjosiah: why exactly? does help not render inside all of the contexts regardless?

thure: In order to get the shell display status, it needs `client`, which it can only get beneath client’s context provider.

wittjosiah: I see, help's logic is also in a context not in a component so the context order matters. We can fiddle with these later, this seem fine for now. Maybe just leave a clearer note stating that help needs to be after client for the context ordering.

PR #7089, opened 2024-06-25
feat(observability): otel logs and metrics

nferch:
Experimental support for OpenTelemetry Logs, Metrics, and Traces submitted through an OpenTelemetry collector.

* Both Datadog and OpenTelemetry metrics are submitted for now.
* Log capture must be explicitly started, and is not enabled on the CLI, probably need some better way for adjusting features.

TODO:
* Remove Datadog
* Update `@baselime/node-opentelemetry` to fix peer dependency issues (https://github.com/baselime/node-opentelemetry/pull/7)


packages/sdk/observability/src/otel/metrics.ts
wittjosiah, line 1: what do you think about hooking these metrics up to this interface? https://github.com/dxos/dxos/blob/main/packages/sdk/observability/src/sentry/browser.ts#L61 then any metrics that get added within the sdk will get sent to sentry for the apps and otel for agents.

nferch: the API seems similar enough, so I don't see why not.

were you thinking we shouldn't send metrics for the apps to Grafana?

could also look at sending traces to Grafana, but I'm less familiar with how we're using those.

wittjosiah: > were you thinking we shouldn't send metrics for the apps to Grafana?

not necessarily, but let's discuss

packages/sdk/observability/src/otel/metrics.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    this._meterProvider = new MeterProvider({
+      resource,
+      readers: [grafanaMetricReader],
+    });
+    this._meter = this._meterProvider.getMeter('dxos-observability');
+
+    const metrics = {
+      // TODO: update metrics names and remove prefix?
```
nferch, line 63: @wittjosiah @yarolegovich could we prefix the metric names (e.g. `dxos.echo.space.count`) so they're less ambiguous and consistent with the trace names?

PR #7099, opened 2024-06-26
build: default context name from constructor call site

yarolegovich:
### Details

* Extended swc-log-plugin to support constructor calls
* Fixed tests for swc-log-plugin
* Added config for adding Context constructor call meta during build
* If name is not provided during Context construction default it to `filename#line_number`.


packages/apps/composer-app/vite.config.ts
```diff
@@ -94,20 +94,30 @@ export default defineConfig({
...
                 package: '@dxos/invariant',
                 param_index: 2,
                 include_args: true,
                 include_call_site: false,
+                include_scope: true,
+              },
+              {
+                name: 'Context',
+                package: '@dxos/context',
+                param_index: 0,
```
dmaretskyi, line 117: Putting it an an index 0 excludes the posibility of contexts with options from having names:

```ts
new Context({ onError })
```

The log transformer will be able to merge the options and add the metadata.

I recommend putting the generated meta into a separate, optional parameter:

```ts
new Context({ onError }, { F: ...., L: ... })
```

yarolegovich: I added the code for modifying object literals [here](https://github.com/dxos/dxos/pull/7099/files#diff-12a7aec988ef3a49326b52998c90b4eb009e6e3c43b5e230cfa99517c391cd81R31). 
But an optional param is maybe more robust, cause will handle non-literal cases like `new Context(errorHandler)`

PR #7101, opened 2024-06-26
chore: Proto-guard add protocols dump

mykola-vrmchk:
### Motivation / Background

This Pull Request ...


packages/gravity/proto-guard/src/check-snapshot.test.ts
```diff
@@ -66,4 +66,32 @@ describe('Load client from storage snapshot', () => {
 
     expect(await SpacesDumper.checkIfSpacesMatchExpectedData(client, expectedData)).to.be.true;
   }).timeout(10_000);
+
+  test
+    .only('load protocols snapshot', async () => {
```
wittjosiah, line 71: .only

PR #7105, opened 2024-06-26
test(composer-app): reduce flakiness of playwright tests

wittjosiah:
- skip test in webkit that was failing consistently in webkit but not chromium
- lower default timeout now that browser processes are being cleaned up and tests run faster
- stop using deprecated non-plank selector for markdown textbox
- don't show privacy notice toast when running e2e tests

Passed three times in a row in CI with these changes (with cache for e2e set to false):

image


packages/apps/composer-app/src/playwright/plugins/deck.ts
```diff
@@ -69,6 +69,7 @@ const getPlankPresence = (locator: Locator) => {
 };
 
 // Define a plank manager that takes a page and provides a way to interact with planks.
+// TODO(wittjosiah): Factor out to react-ui-deck.
```
ZaymonFC, line 72: Good idea.

ZaymonFC (approved):
LGTM ✨

PR #7117, opened 2024-06-27
chore: protect main from accidental pushes

yarolegovich:
### Details

Protect main from accidental pushes by requiring `push_to_main=true` environment variable.


ZaymonFC (approved):
Nice! This would have spared me last week.

.husky/pre-push
wittjosiah, line 1: maybe include `production` in here too :)

wittjosiah (approved):

PR #7125, opened 2024-06-28
fix(echo): don't return partial results from index queries

yarolegovich:
### Details

1. Added documentId to remote QueryResult.
2. Removed individual row load timeout.
3. Removed row-level exception swallowing, if row load fails - the whole query is failed.
4. Removed query-source-level exception swallowing. If a single query source fails empty results are returned with `success: false` (not to introduce any unexpected failures).
5. Fixed error propagation in one-short queries (exception was unhandled before).
6. Objects whose documentId is not equal to the documentId returned from index query results are ignored.


packages/core/echo/echo-pipeline/src/automerge/automerge-doc-loader.ts
```diff
@@ -107,6 +108,17 @@ export class AutomergeDocumentLoaderImpl implements AutomergeDocumentLoader {
...
   }
 
+  public getObjectDocumentId(objectId: string): string | undefined {
+    invariant(this._spaceRootDocHandle);
+    const spaceRootDoc = this._spaceRootDocHandle.docSync();
+    invariant(spaceRootDoc);
+    if (spaceRootDoc.objects?.[objectId]) {
+      return this._spaceRootDocHandle.documentId;
+    }
+    const documentUrl = (spaceRootDoc.links ?? {})[objectId];
```
dmaretskyi, line 118: extract documentUrlToId function

packages/core/echo/echo-db/src/proxy-db/database.ts
```diff
@@ -189,9 +195,9 @@ export class EchoDatabaseImpl extends Resource implements EchoDatabase {
    */
   async loadObjectById<T = any>(
     objectId: string,
-    { timeout }: { timeout?: number } = {},
+    options: LoadObjectOptions = {},
```
dmaretskyi, line 198: doesn't seem right that this is becoming a part of a public API. It's ok if this is just temporary to get the release out, but in that case I'd add a TODO to remove `expectedDocumentId` option

packages/core/echo/echo-db/src/core-db/core-database.ts
```diff
@@ -189,8 +189,16 @@ export class CoreDatabase {
   }
 
   // TODO(Mykola): Reconcile with `getObjectById`.
-  async loadObjectCoreById(objectId: string, { timeout }: { timeout?: number } = {}): Promise<ObjectCore | undefined> {
+  async loadObjectCoreById(
```
dmaretskyi, line 192: doesn't seem right that this is becoming a part of a public API. It's ok if this is just temporary to get the release out, but in that case I'd add a TODO to remove expectedDocumentId option

packages/core/echo/echo-db/src/client/index-query-source-provider.ts
```diff
@@ -20,12 +20,14 @@ import { getObjectCore } from '../core-db';
 import { OBJECT_DIAGNOSTICS, type QuerySourceProvider } from '../hypergraph';
 import { type Filter, type QueryResult, type QuerySource } from '../query';
 
+export type ObjectLoaderParams = {
```
dmaretskyi, line 23:
```suggestion
export type LoadObjectParams = {
```

dmaretskyi (changes requested):

packages/core/echo/echo-db/src/testing/echo-test-builder.ts
```diff
@@ -138,8 +139,13 @@ export const createDataAssertion = ({
...
     },
     verify: async (db: EchoDatabase) => {
-      const { objects } = await db.query().run();
-      const received = objects.find((object) => object.id === seedObject.id);
+      const findReceivedObject = async () => {
+        const { objects } = await db.query().run();
+        const received = objects.find((object) => object.id === seedObject.id);
+        return { objects, received };
+      };
+      await waitForCondition({ condition: async () => (await findReceivedObject()).received != null });
```
dmaretskyi, line 147: hmm, doesn't this PR fix this exact issue? queries not returning all of the objects on the initial call

packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -227,13 +227,32 @@ describe('Queries', () => {
...
   });
+
+  test('query fails if one of the results fails to load', async () => {
+    const kv = createTestLevel();
+    const spaceKey = PublicKey.random();
+
+    const builder = new EchoTestBuilder();
+    afterTest(() => builder.close());
+
+    const peer = await builder.createPeer(kv);
```
dmaretskyi, line 244: kv is optional and a test one will be created by default

packages/core/echo/echo-db/src/proxy-db/database.ts
yarolegovich: true, moved the logic up into echo-client private method.

packages/core/echo/echo-db/src/client/echo-client.ts
```diff
@@ -133,7 +134,14 @@ export class EchoClient extends Resource {
       throw err;
     }
 
-    const object = await db.loadObjectById(objectId, options);
-    return object;
+    if (localDataOnly) {
+      const objectDocId = db._coreDatabase._automergeDocLoader.getObjectDocumentId(objectId);
+      if (objectDocId?.split(':')[0] !== documentId) {
```
dmaretskyi, line 139: not sure I follow the logic behind `objectDocId?.split(':')[0]`. The document ID should never contain any `:` characters

packages/core/echo/echo-db/src/client/index-query-source-provider.ts
```diff
@@ -129,46 +130,38 @@ export class IndexQuerySource implements QuerySource {
...
-          loaded: results.length,
-        });
-
-        if (currentCtx === ctx) {
-          onResult(results);
-        } else {
-          log.warn('results from the previous update are ignored', { queryId });
+        try {
+          const fetchedFromIndexCount = response.results?.length ?? 0;
+          const processedResults: Array<QueryResult | null> =
```
dmaretskyi, line 140:
```suggestion
          const processedResults: (QueryResult | null)[] =
```

packages/core/echo/echo-db/src/client/index-query-source-provider.ts
```diff
@@ -129,46 +130,38 @@ export class IndexQuerySource implements QuerySource {
...
+            fetchedFromIndexCount > 0
```
dmaretskyi, line 141: this branch seems extraneous -- if `response.results.length` is 0 then `.map` will return an empty array anyway

packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -207,12 +205,53 @@ describe('Queries', () => {
...
+        doc.links![obj1.id] = anotherDocHandle.url;
+      });
+      await db.flush();
+      await peer.host.queryService.reindex();
+
+      root = rootDocHandle.url;
+      assertion = { objectId: obj2.id, documentUrl: anotherDocHandle.url };
+    }
+
+    {
```
dmaretskyi, line 244: there's a better `peer.reload()` api which similates a restart without you having to pass `kv` around

packages/core/echo/echo-db/src/client/echo-client.ts
yarolegovich: right, that's some copy-pasted leftover

packages/core/echo/echo-db/src/client/index-query-source-provider.ts
yarolegovich: yeah, can be replaced with `Promise.all((response ?? []).map(...))`

mykola-vrmchk (approved):

PR #7126, opened 2024-06-28
docs: ECHO design docs

dmaretskyi:
### Motivation / Background

This Pull Request ...


design/diagrams/echo-stack.drawio.svg
wittjosiah, line 1: - what is space metadata store? would that include something like cache of space properties?
- what would you expect to be cached in the document cache?
- I'm assuming the dotted line is the client/services rpc boundary?

design/collection-sync.md
```diff
@@ -0,0 +1,90 @@
+# Collection sync
+
+<img src="diagrams/collection-sync.drawio.svg" />
+
+## Goals
+
+Collection sync protocol would enable peers to ensure that a certain set of documents is fully synchronized in relation to other nodes without having to load and sync each document in the set individually.
+
+Collection is defined as a set of automerge documents with known versions (heads).
+The state of a collection is mapping from document ID to the current heads of that document.
```
richburdon, line 10: Define heads.

design/collection-sync.md
```diff
@@ -0,0 +1,90 @@
+# Collection sync
+
+<img src="diagrams/collection-sync.drawio.svg" />
+
+## Goals
```
richburdon, line 5: Overview
(Goals should be listed)

design/collection-sync.md
```diff
@@ -0,0 +1,90 @@
...
+  repeated Document documents = 2;
+}
+```
+
+Each peer would transmit its entire state in one message, completing the sync protocol in one round.
+The message size would be in the order of 100 bytes per document in the collection.
+With an example of 1000 documents per collection, this would result in approx 100 kB messages being exchanged.
+This solution is viable for collections of under 1000 documents as long as the sync has a built-in debounce mechanism.
+
+## Merkle-tree search
```
richburdon, line 50: Merkle Search Tree

design/indexed-document-store.md
```diff
@@ -0,0 +1,61 @@
...
+
+#### Replication with a peer
+
+Given `peerId` find all documents that need to be synced with that peer.
+
+TODO
+
+## Open questions
+
+- How is the list of documents replicated / is this a concern of the document manager?
```
richburdon, line 61: list issues (e.g., fully replicated peers).
merge with my docs?

PR #7134, opened 2024-06-28
feat(authoring): Link pasting

### LINKED ISSUE #5530: [authoring] make pasted links automatically clickable
if text matches a URL it should just be clickable (Cmd + Click?)

ZaymonFC:
- Resolves #5530

https://github.com/dxos/dxos/assets/12402727/07256901-88d5-435b-9b59-940efccb7045


packages/ui/react-ui-editor/src/extensions/markdown/highlight.ts
```diff
@@ -155,12 +144,6 @@ export const markdownHighlightStyle = (readonly?: boolean) => {
       { tag: tags.strong, class: bold },
       { tag: tags.strikethrough, class: strikethrough },
 
-      // Naked URLs.
```
wittjosiah, line 158: so this means it will no longer specially format raw urls?

ZaymonFC: Ah good point I forgot about that. It was interfering with URLs inside the text part of markdown links. Should we try to reinstate that functionality? I could potentially work on it as a follow-up issue?

richburdon: yes we just need i decorator (not just CSS) that recognizes links.

PR #7146, opened 2024-07-02
chore: navtree handling of cycles

### LINKED ISSUE #7075: [organization] dragging a collection into itself doesn't behave as expected

wittjosiah:
Resolves #7075

What is still a problem is if something shows up in multiple places in the navtree, it can still only be selected in one. This needs to be addressed in the upcoming work on navtree.


thure (approved):
lgtm, though I still wonder about the value of treenodes being distinct from graphnodes…

PR #7152, opened 2024-07-02
fix(sdk): Focus within iframe when shell opens

### LINKED ISSUE #7143: [shell] `ESC` should close the identity dialog
Repro:
- hit cmd-k and choose "Open HALO"
- try to close with escape or use the keyboard to navigate
- nothing happens

https://discord.com/channels/837138313172353095/1257425403883294901/1257431653274353694

thure:
This PR:
- resolves #7143
- fixes keyboard access to shell more broadly, by focusing the first focusable element in the shell when it opens

https://github.com/dxos/dxos/assets/855039/46eade83-d449-403b-af9a-1128daa1e5dd

### DIFF
--- a/packages/sdk/client/src/services/shell-manager.ts
+++ b/packages/sdk/client/src/services/shell-manager.ts
@@ -59,6 +59,7 @@ export class ShellManager {
     this._display = ShellDisplay.FULLSCREEN;
     this.contextUpdate.emit({ display: this._display });
     await this._shellRpc.rpc.ShellService.setLayout(request, { timeout: RPC_TIMEOUT });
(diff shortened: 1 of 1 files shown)


packages/sdk/client/src/services/shell-manager.ts
```diff
@@ -59,6 +59,11 @@ export class ShellManager {
     this._display = ShellDisplay.FULLSCREEN;
     this.contextUpdate.emit({ display: this._display });
     await this._shellRpc.rpc.ShellService.setLayout(request, { timeout: RPC_TIMEOUT });
+    (
```
wittjosiah, line 62: could you include a comment here explaining why this is needed/what it's intended to do?

wittjosiah (approved):

PR #7169, opened 2024-07-03
fix(plugin-space): create object translations

### LINKED ISSUE #6910: [organization] two add to space actions in deck
> I get 2 add to space options If i open commands from a block that is in a deck

image

wittjosiah:
Resolves #6910


ZaymonFC (approved):
Nice!

PR #7179, opened 2024-07-03
feat(sdk): Shell agent creation flow improvements

### LINKED ISSUE #7036: [shell] improve agent control
The agent launch UI needs a bit more work. It doesn't render a good "loading state" for "starting agent" and doesn't reflect errors when agent service refuses to do stuff

- after clicking create agent button it should be disabled and say "Requesting agent"
- if the request is successful, introduce a fake device into the device list as a placeholder for the starting agent
- if the request fails an error should be rendered to the user (and full error should be logged/captured for sentry)
- it will be known that the agent has started when a device shows up in the devices list with type `DeviceType.AGENT_MANAGED`
- once the agent is started, rather than replacing the button with an agent, the agent should just be specially designated within the device list and this device should replace the placeholder
- the stop agent button should just be a button in the agent device action menu

### LINKED ISSUE #7171: [shell] "go offline" icon is tin on iOS safari
Image

thure:
This PR:
- resolves #7036 
- resolves #7171

Some states likely need to be tested in staging.

https://github.com/dxos/dxos/assets/855039/8803dcfa-2f08-468f-937d-1b9e0982ba7a

IMG_9518


packages/sdk/client/src/services/fake-agent-hosting-provider.ts
```diff
@@ -2,29 +2,28 @@
 // Copyright 2023 DXOS.org
 //
 
-import { type Halo } from '@dxos/client-protocol';
-import { type Config } from '@dxos/config';
-
 import { type AgentHostingProviderClient } from './agent-hosting-provider';
 
+const wait = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
wittjosiah, line 7: you can `import { sleep } from '@dxos/async';`

packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
```diff
@@ -54,22 +47,30 @@ export const IdentityActionChooserImpl = ({
...
   devices,
   connectionState,
-  onAgentDestroy,
-  onAgentCreate,
-  agentStatus,
-  agentActive,
   agentHostingEnabled,
+  ...agentProps
 }: IdentityActionChooserImplProps) => {
+  const { nonAgentDevices, agentDevice } = devices.reduce(
```
wittjosiah, line 53: worth noting that there could still be other agents in here but likely no managed one so this maybe would be better name "nonManagedDevices" and "managedDevice"

packages/sdk/shell/src/panels/IdentityPanel/useAgentHandlers.ts
```diff
@@ -11,54 +11,91 @@ import { useAgentHostingClient, useClient } from '@dxos/react-client';
...
-          setAgentActive(true);
-          setAgentStatus(agentStatus);
+  const updateAgentStatus = async () => {
+    // TODO(nf): if agent exists, subscribe to updates instead of oneshot status
+    if (agentHostingProviderClient) {
+      setAgentStatus('getting');
+      try {
+        const agentId = await agentHostingProviderClient?.getAgent(identity?.identityKey.truncate() ?? 'never');
+        if (agentId) {
+          setAgentStatus('online');
```
wittjosiah, line 35: I think part of the idea of integrating the agent with the device list and having it show up as a special device is that it should use similar signals as well. I'm not sure what the difference between online/ready is, but I wouldn't expect it to be "online" unless I can see it's presence as active in my devices.

thure: I’ve adjusted how this works — the managed agent will appear in the usual list if it is in `devices`, otherwise `AgentConfig` will render.

PR #7182, opened 2024-07-04
feat(app-graph): new reactive graph builder

### LINKED ISSUE #4418: [composer] revamp Personal and Shared Spaces
## Motivation

The Personal Space is currently a bit confusing in Composer. 
- It's the _default_ space that's created automatically buy the SDK when initializing a new identity. But it can be shared via the SDK or in a non-Composer app, resulting in a broken/confusing Composer UX.
- When creating a new Shared Space in Composer, it starts off as _not shared_, and therefore, Personal. Which is also confusing.

Let's clarify this by simplifying and unifying Personal and Shared.

## Idea

- Personal Space becomes Personal Space**s** (with an extra s). You can create as many spaces in Personal Spaces as you would like. They have the same nesting and display as Shared Spaces. 
- As soon as you share a space, or one is shared with you, it moves to the Shared Spaces section.

## Open Questions / Design Challenges

- Should you have the option of "creating a new space" in the Shared Spaces section? As discussed, by default, it's not shared. So it's confusing to have that option be present. Maybe we go from Space creation straight to the "Share Space" dialogue when creating a space in Shared Spaces?
- Once there is an invite created, does the Space immediately move from Personal Spaces to Shared Spaces? How will that jump in the UI feel?

### LINKED ISSUE #6560: [composer] Remove perfomance-intensive queries
https://github.com/dxos/dxos/blob/286bbb6cdf33fb609190f4d7a0b82eb6c6062be5/packages/apps/plugins/plugin-space/src/util.tsx#L317

This query has an effect of querying ALL data inside echo (equivalent to `SELECT * FROM echo`).

we should query only the metadata typenames

### LINKED ISSUE #7044: [composer] reactive graph builder api
### High level goals
 
- conditional (e.g. only action nodes) + lazy expansion
- only subscribe to updates from specific nodes (+ N-degrees of relations)
  - graph builders which access signals trigger subscriptions via builders being run in `effect`s
  - since builders return proposed nodes and don't actually mutate the graph, they can be run as often as needed
  - given this, the effect subscriptions for a graph builder can be disposed and re-created later when re-subscribing just by running the graph builder again
- lru eviction of unsubscribed nodes from the graph to bound memory usage
- caching + hydration of the graph
  - caching would cache non-data fields of a node under the assumption they are serializable
  - nodes need `type` field
  - id/type pair would be used to re-hydrate `data`
  - graph builder args maybe stop being positional to allow `node` to be swapped for `id` without having to come up with a name that satisfies both
  - maybe helper functions for different types of graph builders
    - `hydrator((state, id, type) => ...)`
    - `connector((state, node, direction) => ...)`
- traversal without loading the whole thing into memory (for export)

### Rough Builder API Sketch

Markdown Plugin

```tsx
definePlugin({
  provides: {
    graph: {
      // Support builder being an array of builders to simplify the logic in each builder
      builders: [
        // State being plugin host state (i.e. other plugins & their provides)
        connector((state, node, direction) => {
          // Here considering Space a type of collection.
          // This would similarly to now add documents both to the space as well as
          //   folders which they exist in, expecting the tree to trim it as necessary.
          if (isCollection(node.data) && direction === 'outbound') {
            return [
              ...constructCollectionActions(state, node.data),
              ...getDocumentsInCollection(node.data).map(constructDocumentNode)
            ];
          }
        }),
        connector((state, node, direction) => {
          if (isDocument(node.data) && direction === 'outbound') {
            return constructDocumentActions(state, node.data);
          }
        })
      ]
    }
  }
});
```

Space Plugin

```tsx
definePlugin({
  provides: {
    graph: {
      builder: [
        connector((state, node, direction) => {
          if (!isSharedSpacesGroupNode(node)) {
            return;
          }
          
          if (direction === 'inbound') {
            return constructRootNode();
          } else {
            const client = getClient(state);
            // Teach builder about non-signals-based subscriptions
            manageSubscription(client.spaces.subscribe);
            return [
              constructSpacesGroupActions(state),
              ...client.spaces.get().map(constructSpaceNode)
            ];
          }
        }),
        connector((state, node, direction) => {
          if (!isSpace(node.data)) {
            return;
          }
  
          if (direction === 'inbound') {
            return constructSpacesGroupNode();
          } else {
            return constructSpaceActions(state, node.data);    
          }
        }),
        connector((state, node, direction) => {
          if (isTypedObject(node.data) && direction === 'inbound') {
            return constructSpaceNode(getSpaceFromObject(node.data));
          }
        })
      ]
    }
  }
})
```

### LINKED ISSUE #7238: [authoring] document doesn't fully render content on first load in space stack
Needs repro steps

Image

In the image above those are the same document.

wittjosiah:
Resolves #7044 
Resolves #4418 
Resolves #6560
Resolves #7238


packages/sdk/app-graph/src/graph-builder.ts
```diff
@@ -2,50 +2,236 @@
...
+import { type Signal, effect, signal } from '@preact/signals-core';
+// import { yieldOrContinue } from 'main-thread-scheduling';
+
+import { EventSubscriptions } from '@dxos/async';
+import { invariant } from '@dxos/invariant';
 
 import { Graph } from './graph';
+import { type EdgeDirection, type NodeArg, type Node } from './node';
+
+export type ConnectorExtension = (params: {
```
richburdon, line 14: Major concepts like this really need at least a sentence.

packages/sdk/app-graph/src/graph-builder.ts
```diff
@@ -2,50 +2,236 @@
...
+export const connector = (extension: ConnectorExtension): BuilderExtension => ({ type: 'connector', extension });
+export const hydrator = (extension: HydratorExtension): BuilderExtension => ({ type: 'hydrator', extension });
+
+export type GraphBuilderTraverseOptions = {
+  node: Node;
+  direction?: EdgeDirection;
+  visitor: (node: Node, path: string[]) => void;
+};
+
+class BuilderInternal {
```
richburdon, line 41: i don't understand why this needs to be static and it feels dangerous.

packages/sdk/app-graph/src/graph-builder.ts
```diff
@@ -2,50 +2,236 @@
...
+          type: arg.type,
+          data: arg.data ?? null,
+          properties: arg.properties ?? {},
+        }),
+      );
+
+    await Promise.all(nodes.map((n) => this.traverse({ node: n, direction, visitor }, [...path, node.id])));
+  }
+
+  private get _hydrators() {
```
richburdon, line 161: just maintain a multimap?

packages/sdk/app-graph/src/graph-builder.ts
```diff
@@ -2,50 +2,236 @@
...
+  | {
+      type: 'connector';
+      extension: ConnectorExtension;
+    }
+  | {
+      type: 'hydrator';
+      extension: HydratorExtension;
+    };
+
+export const connector = (extension: ConnectorExtension): BuilderExtension => ({ type: 'connector', extension });
```
richburdon, line 32: what is the purpose of this polymorhpism since these extensions do very different things?

packages/sdk/app-graph/src/graph-builder.ts
wittjosiah: strictly speaking it doesn't need to be static, but the alternative api feels very cumbersome (needing to pass an instance of a builder around everywhere to track the state). The goal here was to end up with a more React-like api where that wasn't necessary.

There are certainly potentials gotchas with this, for instance, everything needs to be using the same version of app-graph otherwise they could end up with different static classes which are misaligned. However, this problem of ensuring versions are aligned is also a problem that we already deal with in other packages that we use so that didn't feel like a blocker.

All that said, if there was a way to have this _not_ be static or module level state or something like but still have a nice clean builder api then I'd love that, I just haven't figured out a way to accomplish that to this point.

packages/sdk/app-graph/src/graph-builder.ts
wittjosiah: I don't think I'm hard set on them being the same thing, I think this mostly just evolved from convenience of not changing the api. One thing I'd say though is that I don't think they do very different things, they take different inputs but in the end they both add nodes to the graph.

packages/sdk/app-graph/src/graph-builder.test.ts
```diff
@@ -0,0 +1,279 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { signal } from '@preact/signals-core';
+import { expect } from 'chai';
+
+import { describe, test } from '@dxos/test';
```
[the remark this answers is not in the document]
wittjosiah: yeah I'm hoping to do that

packages/sdk/app-graph/src/graph-builder.test.ts
```diff
@@ -105,76 +118,92 @@ describe('GraphBuilder', () => {
...
-      expect(graph.nodes(graph.root)[0]?.properties.label).to.equal('updated');
+      expect(node.properties.label).to.equal('updated');
     });
 
     test('removes', () => {
       const builder = new GraphBuilder();
       const nodes = signal([
-        { id: 'first', type: 'type', data: 1 },
-        { id: 'second', type: 'type', data: 2 },
+        { id: 'first', type: EXAMPLE_TYPE, data: 1 },
```
richburdon, line 137: 'item-1', 'item-2' i think is clearer (more realistic and moves towards generating these for larger tests).

packages/sdk/app-graph/src/graph-builder.test.ts
```diff
@@ -8,41 +8,49 @@ import { expect } from 'chai';
 import { describe, test } from '@dxos/test';
 
 import { ACTION_TYPE } from './graph';
-import { GraphBuilder, connector, hydrator, memoize } from './graph-builder';
+import { GraphBuilder, defineExtension, memoize } from './graph-builder';
+
+const EXAMPLE_ID = 'dx:echo:BA25QRC2FEWCSAMRP4RZL65LWJ7352CKE:01J00J9B45YHYSGZQTQMSKMGJ6';
```
richburdon, line 13: TEST_

packages/sdk/app-graph/src/graph-builder.test.ts
```diff
@@ -240,34 +275,31 @@ describe('GraphBuilder', () => {
...
-      );
-      builder.addExtension(
-        'connector',
-        connector(({ node }) => {
-          // TODO(wittjosiah): Should subscribe to parent changes.
-          const a = memoize(() => Math.random());
-          const b = memoize(() => Math.random());
-          const c = Math.random();
-          return [{ id: 'second', type: 'type', data: { a, b, c, data: node.id } }];
+        'extension',
```
richburdon, line 278: you don't need the descriminating prop now?
also pref, createExtension (and need ID?)

packages/sdk/app-graph/src/graph-builder.ts
```diff
@@ -31,25 +31,21 @@ export type ConnectorExtension = (params: {
...
+  connector?: ConnectorExtension;
+  resolver?: ResolverExtension;
+};
 
-export const connector = (extension: ConnectorExtension): BuilderExtension => ({ type: 'connector', extension });
-export const hydrator = (extension: HydratorExtension): BuilderExtension => ({ type: 'hydrator', extension });
+/**
+ * Create a graph builder extension.
+ */
+export const defineExtension = (extension: BuilderExtension): BuilderExtension => extension;
```
richburdon, line 44: createExtension?

packages/sdk/app-graph/src/graph-builder.test.ts
wittjosiah: these are just identifiers for the extensions, but yes they should probably just be apart of the extension def 👍

packages/apps/composer-app/project.json
```diff
@@ -4,7 +4,7 @@
   "tags": [
     "scope:apps"
   ],
-  "sourceRoot": "packages/apps/composer-app/src",
+  "sourceRoot": "{projectRoot}/src",
```
thure, line 7: 🙌

Issue #7197, opened 2024-07-05
[composer] Inflexible scroll and focus system


ZaymonFC:
Our `scroll-into-view` and focus management system requires a significant overhaul:

1. Current approach lacks flexibility and features
2. Inconsistent auto-focus behavior across data types
3. Insufficient control over focus / `scroll into view` mechanics
4. Cannot re-trigger scroll for the same plank (e.g., subsequent internal link clicks)

ZaymonFC:
My current thinking is that every data type should have a way of locating and focusing it. We should be able to deterministically fire this on any object in any plank, stack, or collection.

If there are multiple instances of an object in the view, we should examine the current object and check if the new instance is contained within it (stack, grid, collection, etc.). We should enumerate the semantic cases and variations for scrolling and focusing, such as:

- Opening an object from the navtree or search
- Creating an object from the navtree or side-bar
  - Either open in a new plank OR scroll to the open collection where the newly created object was created
- Navigating an internal link

Related issues:
- #7011
- #7045

PR #7198, opened 2024-07-05
fix(authoring): Internal links work with deck

### LINKED ISSUE #7011: [authoring] internal links are broken
Links in a markdown document of the form:

```md
[A thing](/c49a27f8bbf524d10db1587e26985146daaa0eee4dad1ddbe267b200ba728192) is linked.
```

where `c49a27f8bbf524d10db1587e26985146daaa0eee4dad1ddbe267b200ba728192` is an object ID in ECHO.

They currently show an internal link icon: 

Image

But clicking the link doesn't navigate to the object.

ZaymonFC:
- Resolves #7011

As a follow up I'd like to simplify the active id manipulation that `SearchDialog` and `NavTreeContainer` by leveraging the new NavigationAction `ADD_TO_ACTIVE`.

Open to better names for the action, however, I'm not too fussed since they're ephemeral.

Clicking on an internal link multiple times in deck will only scroll and focus the linked object the first time. 
- More discussion here: #7197

### DIFF
--- a/packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
+++ b/packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
@@ -406,6 +406,56 @@ export const DeckPlugin = ({
               };
             }
 
+            case NavigationAction.ADD_TO_ACTIVE: {
+              const { id, scrollIntoView, pivot } = intent.data as NavigationAction.AddToActive;
+              batch(() => {
+                if (isActiveParts(location.active)) {
+                  const main = Array.isArray(location.active.main) ? [...location.active.main] : [location.active.main];
+
+                  if (!main.includes(id)) {
+                    // Check if the id is not already in the main array
+                    if (pivot) {
+                      const pivotIndex = main.indexOf(pivot.id);
+                      if (pivotIndex !== -1) {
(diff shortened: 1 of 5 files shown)


packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -406,6 +406,56 @@ export const DeckPlugin = ({
...
+                      } else {
+                        main.push(id);
+                      }
+                    } else {
+                      const newIds = [id];
+                      const partMembers = new Set<string>();
+                      switch (settings.values.newPlankPositioning) {
+                        case 'start': {
+                          newIds.forEach((newId) => partMembers.add(newId));
+                          main.forEach((existingId) => partMembers.add(existingId));
```
thure, line 430: For this to work as expected, you have to filter `newIds` to just the ones not already in `main`, otherwise the ids move around. The implementation of `NavigationAction.OPEN` adjacent this one does that.

thure: I see this is an array with one item, and moving it is probably the point of this action, so never mind 😅

thure (approved):
Into it!

Issue #7081, opened 2024-06-25
Replication errors


dmaretskyi:
![Image](https://github.com/dxos/dxos/assets/35851437/60575535-18a8-4915-a411-98dd204bb2f5)

dmaretskyi:
Turn those into a metric counter instead?

PR #7209, opened 2024-07-08
fix(echo): replication failure handling improvements

yarolegovich:
### Details

* AutomergeReplicator closes connection after exceeding max retries to send a sync message.
* Await sendSyncMessage result in MeshReplicatorConnection to enable WritableStream backpressure. For 10 seconds (control-hearbeat timeout) we were accumulating failed sendMessageSync execution paths all stuck in retry loop.
* Check connection presence in connectionPerPeer before calling onConnectionAuthScopeChanged to prevent invariant violation.
* Extracted MeshReplicatorConnection into a separate file.

re #7081


packages/core/echo/echo-pipeline/src/automerge/mesh-echo-replicator-connection.ts
```diff
@@ -0,0 +1,130 @@
...
+        readableStreamController = controller;
+        this._ctx.onDispose(() => controller.close());
+      },
+    });
+
+    this.writable = new WritableStream<Message>({
+      write: async (message: Message, controller) => {
+        invariant(this._isEnabled, 'Writing to a disabled connection');
+        try {
+          await this.replicatorExtension.sendSyncMessage({ payload: cbor.encode(message) });
```
dmaretskyi, line 49: does this mean that we would only be able to have 1 outgoing sync message in-transit for every space?

packages/core/mesh/teleport-extension-automerge-replicator/src/automerge-replicator.ts
```diff
@@ -122,13 +139,15 @@ export class AutomergeReplicator implements TeleportExtension {
         log('sendSyncMessage error', { err });
 
         retries++;
-        if (retries >= MAX_RETRIES) {
-          throw new Error(
-            `Failed to send sync message after ${MAX_RETRIES} retries. Last attempt failed with error: ${err}`,
+        if (retries >= retryPolicy.maxRetries) {
+          const numberOfRetriesExceededError = new Error(
```
dmaretskyi, line 143: nit: this could use error-cause property: https://v8.dev/features/error-cause

dmaretskyi (approved):

packages/core/echo/echo-pipeline/src/automerge/mesh-echo-replicator-connection.ts
yarolegovich: 1 outgoing sync message per peer

PR #7212, opened 2024-07-09
feat(comments): Disable comment button on selection overlap

ZaymonFC:


packages/ui/react-ui-editor/src/extensions/markdown/formatting.ts
```diff
@@ -1124,7 +1131,17 @@ export const getFormatting = (state: EditorState): Formatting => {
   };
 
   const { selection } = state;
+  const commentState = state.field(commentsState);
+
   for (const range of selection.ranges) {
+    if (!comment) {
+      const overlap = (a: Range, b: SelectionRange): boolean => {
+        return a.from <= b.to && a.to >= b.from;
+      };
```
ZaymonFC, line 1140: I wonder if we have one of these helpers already. I don't want to be re-creating this function every selection change. @richburdon thoughts on where this should live?

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -505,6 +514,22 @@ export const scrollThreadIntoView = (view: EditorView, id: string, center = true
   }
 };
 
+/**
+ * Query the editor state for the active formatting at the selection.
+ */
+export const doesSelectionOverlapComment = (state: EditorState): boolean => {
```
richburdon, line 520: selectionContainsComment?

richburdon: separately, perhaps we should disable comments if there's more than one selection range?

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -521,3 +546,23 @@ export const useComments = (view: EditorView | null | undefined, id: string, com
     }
   }, [id, view, comments]);
 };
+
+/**
+ * Hook provides an extension to compute the current comment state under the selection.
+ * NOTE(Zan): I think this conceptually belongs in 'formatting.ts' but we can't import ESM modules there atm.
```
richburdon, line 552: how is that ESM? i mean why are the two approaches done differently?

packages/ui/react-ui-editor/src/extensions/util/overlap.ts
```diff
@@ -0,0 +1,15 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+interface Range {
```
richburdon, line 5: this is already defined in src.extension/types.ts

packages/ui/react-ui-editor/src/extensions/comments.ts
ZaymonFC: Since formatting has an associated test file, and the tests for that project haven't been migrated to `vitest` we can't have any ESM imports in `formatting.ts`.

PR #7213, opened 2024-07-09
feat: Deck layout update

richburdon:
- Update react-ui status component
- New loading treatment with DXOS logo
- Updated deck layout (overscroll)
- Deck settings (for background)
- Updated storybook settings (for dark mode)

https://github.com/dxos/dxos/assets/3523355/f8854a0d-4aa9-4242-8d9e-467f1b46983f


packages/apps/composer-app/src/components/Placeholder.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <div className={mx('flex grow items-center scale-50 opacity-0 blur-xl', className)}>
+          <DXOS className={getSize(48)} />
+          {/* <ComposerLogo */}
+          {/*  size={200} */}
+          {/*  classNames={[ */}
+          {/*    'fill-[#ddd] dark:fill-[#1d1d1d]', */}
+          {/*    'fill-[#ccc] dark:fill-[#191919]', */}
+          {/*    'fill-[#bbb] dark:fill-[#161616]', */}
+          {/*  ]} */}
+          {/* /> */}
```
thure, line 32: Remove unused code

packages/apps/composer-app/src/components/Placeholder.tsx
```diff
@@ -0,0 +1,37 @@
...
+    setTimeout(() => {
+      setClassName('transition duration-1000 rotate-180 scale-100 opacity-25 blur-none');
+    }, 100);
+  }, []);
+
+  return (
+    <ThemeProvider tx={defaultTx}>
+      <div className='flex flex-col bs-[100dvh] justify-center items-center'>
+        <Status indeterminate aria-label='Initializing' classNames='block is-full' />
+        <div className={mx('flex grow items-center scale-50 opacity-0 blur-xl', className)}>
```
thure, line 23: Generally `div`s should not need to be `scale`’d, did something motivate this over setting a plain size?

packages/apps/composer-app/src/components/Placeholder.tsx
```diff
@@ -0,0 +1,37 @@
...
+// Copyright 2024 DXOS.org
+//
+
+import React, { useEffect, useState } from 'react';
+
+import { DXOS } from '@dxos/brand';
+import { Status, ThemeProvider } from '@dxos/react-ui';
+import { defaultTx, getSize, mx } from '@dxos/react-ui-theme';
+
+export const Placeholder = () => {
```
thure, line 11: The root node of this is `Status`, so the name of this component should reflect that. `Loader` is also a worthwhile term here, but `Placeholder` signals (for me) that something did not load or that this is a design prototype.

packages/apps/composer-app/src/components/Placeholder.tsx
```diff
@@ -0,0 +1,37 @@
...
+      setClassName('transition duration-1000 rotate-180 scale-100 opacity-25 blur-none');
+    }, 100);
+  }, []);
+
+  return (
+    <ThemeProvider tx={defaultTx}>
+      <div className='flex flex-col bs-[100dvh] justify-center items-center'>
+        <Status indeterminate aria-label='Initializing' classNames='block is-full' />
+        <div className={mx('flex grow items-center scale-50 opacity-0 blur-xl', className)}>
+          <DXOS className={getSize(48)} />
```
thure, line 24: I don’t think this improves usability much, and it uses our 'X' logo which is already a signifier in UX for ‘dismiss’ or ‘close’.

I would much prefer a loading skeleton, for which there are plenty of studies that show usability improvements. Now that Deck has a stable design, we should seriously consider skeletons instead of trying to optimize an atomic signifier like this.

thure: See: https://www.nngroup.com/articles/skeleton-screens/

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -45,19 +46,24 @@ const isSocket = !!(globalThis as any).__args;
...
+      'px-96',
+      // TODO(burdon): Option for opacity.
+      themeBackgroundDark.length ? '_dark:opacity-95 _dark:bg-transparent' : 'bg-neutral-50 dark:bg-neutral-950',
+    ),
   },
   plank: {
     classNames: 'mx-1 bg-neutral-25 dark:bg-neutral-900',
   },
-};
+});
```
thure, line 66: This implementation should be improved now that this isn’t for a demo — I think there are a variety of options we could go with, depending on what the feature is meant to work / how it is meant to scale. I think we should discuss.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -68,6 +78,28 @@ export const firstComplementaryId = (active: Location['active']): string | undef
...
+    return <div className='absolute inset-0 z-0 bg-neutral-50 dark:bg-neutral-950' />;
+  }
+
+  return (
+    <div
+      className='absolute inset-0 z-0 bg-cover bg-no-repeat'
+      style={{ backgroundImage: `url(${backgroundImage})` }}
+    />
+  );
+};
```
thure, line 101: This should not be a component or add an element to the DOM, there are already several elements that can apply a `background-image` in an equivalent way.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -384,9 +416,7 @@ export const DeckLayout = ({
         {(Array.isArray(activeParts.main) ? activeParts.main.filter(Boolean).length > 0 : activeParts.main) ? (
           <Main.Content bounce classNames={['grid', 'block-end-[--statusbar-size]']}>
             <div role='none' className='relative'>
-              {slots?.wallpaper?.classNames && (
-                <div className={mx('absolute inset-0 z-0', slots.wallpaper.classNames)} />
-              )}
+              <Wallpaper {...slots?.wallpaper} />
               <Deck.Root classNames={mx('absolute inset-0', slots?.deck?.classNames)}>
```
thure, line 420: `Deck.Root` can already support `style={{backgroundImage: …}}`

packages/ui/brand/src/components/ComposerLogo/ComposerLogo.stories.tsx
```diff
@@ -12,25 +13,35 @@ import { Button } from '@dxos/react-ui';
...
 };
 
+// TODO(burdon): Get from theme?
+const colors = {
+  gray: '#888888',
+  purple: '#AA23D3',
+  orange: '#CA6346',
+  green: '#4DA676',
+  blue: '#539ACD',
+};
```
thure, line 36: It’s okay to use absolute colors in stories IMO, though it is (or should be) trivial to get them from `react-ui-theme`.

packages/ui/react-ui-theme/src/config/tailwind.ts
```diff
@@ -146,13 +146,17 @@ export const tailwindConfig = ({
...
+              width: '50%',
+            },
+            '75%': {
+              left: '50%',
               width: '100%',
             },
             '100%': {
-              width: '0%',
               left: '100%',
+              width: '0%',
```
thure, line 159: I thought this PR removed the indeterminate status bar from Composer — are these changes still intentional?

tools/stories/.storybook-shared/preview.ts
```diff
@@ -25,35 +26,41 @@ const preview: Preview = {
...
     darkMode: {
+      classTarget: 'html',
+      stylePreview: true,
       dark: { ...themes.dark },
-      light: { ...themes.normal },
+      darkClass: 'dark',
+      light: { ...themes.light },
+      lightClass: 'dark',
+      // TODO(burdon): This is ignored.
+      // current: 'light'
```
thure, line 39: Thank you for fixing this 👍

thure (changes requested):
I have two significant concerns that I think merit negotiation, let’s please discuss.

packages/apps/composer-app/src/components/Placeholder.tsx
richburdon: it sets the initial size prior to the transition being applied in the timeout above. is there a better way to do this?

packages/apps/composer-app/src/components/Placeholder.tsx
richburdon: Let's discuss.

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
richburdon: Yep let's discuss and do this properly.

packages/apps/composer-app/src/components/Placeholder.tsx
thure: Ah I think I didn't realize this was part of the animation, it's fine if so.

packages/apps/composer-app/src/components/Placeholder.tsx
thure: Noting for posterity from the NN article:

> Spinners are typically best used on a single module, like a video or a card which is on a dashboard. Skeleton screens (with the exception of frame-display ones) are better when the full screen is loading because the wireframe gives users a sense of what the page will look like and, thus, minimizes cognitive load.

> progress bars are strongly recommended for any page that takes longer that 10 seconds to load because they give users a sense of the state of the system and of how much longer they have to wait. Anything above 10 seconds requires an explicit estimation of duration.

packages/ui/brand/src/components/ComposerLogo/ComposerLogo.stories.tsx
thure: fwiw this relates to my (and Daniel’s) idea to replace colors even in Tailwind with CSS variables, which would eliminate the need to implement semantic tokens in Tailwind and make it easier to reference them from TS. I have a system (vite plugin + package) ready to go for this.

Issue #7194, opened 2024-07-05
[comments] clicking on a comment should open the complementary sidebar


wittjosiah:
Currently you need to manually open the complementary sidebar to view comments which is not intuitive.

wittjosiah:
Consider doing this in a popover above the comment instead on directly clicking on the comment itself but start with just clicking on the comment because it should be simpler

PR #7226, opened 2024-07-09
feat(comments): Click to open

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/7194

This PR adds clickable comment highlights to our editor, enhancing user interaction with code comments.

https://github.com/user-attachments/assets/cdd1fa07-2269-49a4-8e66-bad945beca86

## Features
- Clicking a comment opens the c11y sidebar.
- Better styles for comments.

## Technical Approach
- Used CodeMirror's decoration system for styling
- Annotated each decoration with the comment ID to decouple presentation and click handling
- Created `commentClickedEffect` StateEffect for click events
- `useCommentClickListener` React hook that creates an `EditorView.updateListener` observer to handle the effect


packages/apps/plugins/plugin-markdown/src/components/EditorMain.tsx
```diff
@@ -101,16 +103,17 @@ export const EditorMain = ({
...
 
-  // Toolbar actions.
-  const handleAction = useActionHandler(editorView);
-  const [formattingState, formattingObserver] = useFormattingState();
 
-  // Comments
+  const [formattingState, formattingObserver] = useFormattingState();
   const [{ comment, selection }, commentObserver] = useCommentState();
+  const commentClickObserver = useCommentClickListener((id) => {
+    props.onCommentClick && props.onCommentClick(id);
```
wittjosiah, line 110:
```suggestion
    props.onCommentClick?.(id);
```

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -145,17 +164,33 @@ const commentsDecorations = EditorView.decorations.compute([commentsState], (sta
...
   return Decoration.set(decorations);
 });
 
+const commentClickedEffect = StateEffect.define<string>();
+
+const handleCommentClick = EditorView.domEventHandlers({
+  click: (event, view) => {
+    let target = event.target as HTMLElement;
+    // Traverse up the DOM tree looking for an element with data-comment-id
+    while (target && !target.hasAttribute('data-comment-id')) {
```
wittjosiah, line 181: will this limit itself to codemirror or would it traverse up the whole page if no comments exist?

wittjosiah (approved):

packages/ui/react-ui-editor/src/extensions/comments.ts
ZaymonFC: It will go all the way up to the html root. It's only on click events though so I'm not too bothered. Can implement a short circuit if you feel it's good to isolate it from the wider app?

wittjosiah: given the potentially dynamic nature of plugins it feels like it would be safer to be defensive and try to keep things from having unintended side-effects (i.e. another plugin having a `data-comment-id` attribute that this somehow finds triggers clicks against)

wittjosiah: if it's a huge bother to find get the codemirror root here then I'd just leave it as a todo

ZaymonFC: Looks like we can just use `view.dom` as a reference to the root element.

PR #7236, opened 2024-07-10
fix: storybook dark mode config

richburdon:


tools/stories/.storybook-shared/config.ts
```diff
@@ -25,6 +25,7 @@ export const config = (
     '@storybook/addon-interactions',
     '@storybook/addon-links',
     '@storybook/addon-themes',
+    // 'storybook-dark-mode'
```
thure, line 28: Can go ahead and completely remove I think

PR #7239, opened 2024-07-10
chore(native): Add socket config and bundle script to `testbench-app`

ZaymonFC:
Add socket config and bundle script to the testbench application.

image


wittjosiah (approved):
maybe add a note about how to run it to the testbench readme as well

PR #7242, opened 2024-07-11
chore: try bundling sourcemaps with releases

yarolegovich:
### Details

Source maps are uploaded, but not associated with releases properly.
image

### DIFF
--- a/packages/apps/composer-app/vite.config.ts
+++ b/packages/apps/composer-app/vite.config.ts
@@ -174,6 +174,9 @@ export default defineConfig({
       },
       authToken: process.env.SENTRY_RELEASE_AUTH_TOKEN,
       disable: process.env.DX_ENVIRONMENT !== 'production',
+      release: {
+        name: process.env.npm_package_version,
+      },
     }),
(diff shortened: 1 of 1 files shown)


packages/apps/composer-app/vite.config.ts
```diff
@@ -174,6 +174,9 @@ export default defineConfig({
       },
       authToken: process.env.SENTRY_RELEASE_AUTH_TOKEN,
       disable: process.env.DX_ENVIRONMENT !== 'production',
+      release: {
+        name: process.env.npm_package_version,
```
wittjosiah, line 178: how does this env get set?

yarolegovich: https://docs.npmjs.com/cli/v8/using-npm/scripts#packagejson-vars
npm (and pnpm as well) sets it

wittjosiah: 🤯 

excellent

wittjosiah (approved):

PR #7244, opened 2024-07-11
feat: Collection sync

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
```diff
@@ -327,12 +364,121 @@ export class AutomergeHost extends Resource {
...
+      }
+    }
+    return this._headsStore.getHeads(documentId);
+  }
+
+  //
+  // Collection sync.
+  //
+
+  getLocalCollectionState(collectionId: string): CollectionState | undefined {
```
yarolegovich, line 378: not sure if these wrapper methods are useful, just adds an extra hop during navigation to the actual logic

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
```diff
@@ -327,12 +364,121 @@ export class AutomergeHost extends Resource {
...
+    return result;
+  }
+
+  /**
+   * Update the local collection state based on the locally stored document heads.
+   */
+  async updateLocalCollectionState(collectionId: string, documentIds: DocumentId[]) {
+    const heads: Record<DocumentId, Heads> = {};
+    await Promise.all(
+      documentIds.map(async (documentId) => {
```
yarolegovich, line 419: not sure we want to crash the whole thing if we fail to retrieve heads for a document. we'll retry eventually, so I'd just log and swallow the error here.

also I think it's better to use getMany queries in such cases, even remembering how "nice" the implementation is in browsers. if the implementation changes we'll benefit from it. also they have parallel limit there which might be good for spreading the load in time if documentIds collection is big.

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
```diff
@@ -327,12 +364,121 @@ export class AutomergeHost extends Resource {
...
+      count: different.length,
+    });
+
+    // Load the documents that are different.
+    for (const documentId of different) {
+      this._repo.find(documentId);
+    }
+  }
+
+  private _onHeadsChanged(documentId: DocumentId, heads: Heads) {
```
yarolegovich, line 474: should it maybe happen afterSave? we don't use batch in any way for atomicity here, but if save fails we have non-existent heads recorded

packages/core/echo/echo-db/src/host/space-state-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+    await root.handle.whenReady();
+
+    const checkSpaceDocumentList = async () => {
+      const documentIds = root.getAllLinkedDocuments().map((url) => interpretAsDocumentId(url));
+      if (!isEqual(documentIds, this._lastSpaceDocumentList.get(spaceId))) {
+        this._lastSpaceDocumentList.set(spaceId, documentIds);
+        this.spaceDocumentListUpdated.emit(new SpaceDocumentListUpdatedEvent(spaceId, documentIds));
+      }
+    };
+    root.handle.addListener('change', checkSpaceDocumentList);
```
yarolegovich, line 73: should we maybe have a short debounce here? if we're creating a composite object Document+Text content there'll be two separate change events one after another I think

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -383,14 +383,17 @@ export class DataSpaceManager {
...
       onAuthorizedConnection: (session) => {
-        session.addExtension('dxos.mesh.teleport.admission-discovery', new CredentialServerExtension(space));
-        session.addExtension(
-          'dxos.mesh.teleport.gossip',
-          gossip.createExtension({ remotePeerId: session.remotePeerId }),
-        );
-        session.addExtension('dxos.mesh.teleport.notarization', dataSpace.notarizationPlugin.createExtension());
-        this._echoHost.authorizeDevice(space.key, session.remotePeerId);
-        session.addExtension('dxos.mesh.teleport.automerge', this._echoHost.createReplicationExtension());
+        // TODO(dmaretskyi): Is it safe to do this asynchronously.
```
yarolegovich, line 386: I'd do
```ts
        queueMicrotask(async () => {
          if (!session.isOpen) {
            return;
          }
          try {
            ...
          } catch (err) {
            await session.close(err as Error);
          }
        });
```

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
```diff
@@ -0,0 +1,197 @@
...
+  getLocalCollectionState(collectionId: string): CollectionState | undefined {
+    return this._getPerCollectionState(collectionId).localState;
+  }
+
+  setLocalCollectionState(collectionId: string, state: CollectionState) {
+    this._getPerCollectionState(collectionId).localState = state;
+
+    queueMicrotask(async () => {
+      for (const peerId of this._connectedPeers) {
+        await this._refreshInterestedPeers(collectionId, peerId);
```
yarolegovich, line 70: should it be async?

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
```diff
@@ -0,0 +1,197 @@
...
+    this._queryCollectionState = params.queryCollectionState;
+    this._shouldSyncCollection = params.shouldSyncCollection;
+  }
+
+  protected override async _open(ctx: Context): Promise<void> {
+    scheduleTaskInterval(
+      this._ctx,
+      async () => {
+        for (const collectionId of this._perCollectionStates.keys()) {
+          this.refreshCollection(collectionId);
```
yarolegovich, line 50: maybe async return after refreshCollection to yield cpu time to other tasks

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  /**
+   * Callback when a connection to a peer is established.
+   */
+  onConnectionOpen(peerId: PeerId) {
+    this._connectedPeers.add(peerId);
+
+    queueMicrotask(async () => {
```
yarolegovich, line 102: why do we refreshCollection for all peers when a new connection is open. maybe make it just
```ts
    queueMicrotask(async () => {
      for (const [collectionId, state] of this._perCollectionStates.entries()) {
        if (await this._shouldSyncCollection(collectionId, peerId)) {
          state.interestedPeers.add(peerId);
          state.lastQueried.set(peerId, Date.now());
          this._queryCollectionState(collectionId, peerId);
        }
      }
    });
```
instead?

yarolegovich (changes requested):

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
dmaretskyi: 👍

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
dmaretskyi: I thought about it too, but this seems like a lesser evil then exposing the internals -- in the future it will be much easier to swap out the implementation. Also if you think about the architecture as a bunch of layered components, then wrapper methods help make the API explicit at every layer. Of course none of this is universal, and sometimes it makes sense to expose a getter to a component (e.g. `space.db`), but I think in this case I'd prefer to hide the collection synchronizer

dmaretskyi: I also want to fully hide the Repo inside AutomergeHost, hence the methods for creating and querying documents on this class

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
dmaretskyi: getHeads does only local IO, im not sure if we need to gracefully handle storage failures

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
dmaretskyi: not sure, but I generally prefer something to be async, unless there's a strong need for it to be synchronous (e.g. UI/signal-related code)

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
dmaretskyi: to add jitter for collection queries?

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
dmaretskyi: TODO: test how getMany handles missing keys -- we want to return undefined for missing ones, and not fail the whole query

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
yarolegovich: >  in the future it will be much easier to swap out the implementation

you can always add a getter with limited-API interface return type. but not that I'm insisting

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
yarolegovich: I think the opposite unless it's smth very generic like `open/close` on `resource/extension`.
`async` is always more suspicious (can it fail, how long can it take, can it hang, can it timeout?) and makes the client code harder like [here with Promise.all](https://github.com/dxos/dxos/pull/7244/files#diff-755dd7515a12273099fb49476cbda4caaf738263c1f702b0138ee35897994dbeR166)

it also looks funny that send and query in `CollectionSynchronizerParams` are sync, but this one is async. you'd usually expect the first two to be async

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
yarolegovich: I think there should be kind of implicit jitter cause of MIN_QUERY_INTERVAL which is nice. async return is just for spreading iteration over all_document_count * peers over multiple callbacks iterating over document_per_collection * peers in each

yarolegovich (approved):

PR #7249, opened 2024-07-12
fix: storybook theme mode, status bar changes, key shortcuts text color changes, composer logo changes, D3 upgrade

richburdon:
- Update storybook dark mode switch
- Update storybook decorators


packages/apps/composer-app/src/main.tsx
```diff
@@ -123,8 +123,8 @@ const main = async () => {
     ),
     placeholder: (
       <ThemeProvider tx={defaultTx}>
-        <div className='flex bs-[100dvh] justify-center items-center'>
-          <Status indeterminate aria-label='Initializing' />
+        <div className='flex flex-col justify-end bs-[100dvh]'>
```
thure, line 126: Does this just put `Status` on the bottom? Does that not look quite empty?

packages/apps/plugins/plugin-help/src/components/Shortcuts/ShortcutsList.tsx
```diff
@@ -20,11 +21,16 @@ export const ShortcutsList = () => {
   });
 
   return (
-    <dl className='is-fit grid grid-cols-[min-content_minmax(12rem,1fr)] gap-2 mlb-4'>
+    <dl
+      className={mx(
+        'is-fit grid grid-cols-[min-content_minmax(12rem,1fr)] gap-2 mlb-4',
+        'text-neutral-700 dark:text-neutral-300',
```
thure, line 27: Prefer to use semantic tokens for this kind of thing — should we mint a semantic token for this value? What is the purpose of these colors?

packages/apps/plugins/plugin-help/src/components/Shortcuts/ShortcutsList.tsx
```diff
@@ -20,11 +21,16 @@ export const ShortcutsList = () => {
...
+      )}
+    >
       {bindings.map((binding, i) => (
         <Fragment key={i}>
           <Key binding={binding.shortcut} />
-          <span role='definition' aria-labelledby={binding.shortcut}>
+          <span role='definition' className='ml-4' aria-labelledby={binding.shortcut}>
```
thure, line 33: Prefer `mis-4`

packages/ui/brand/src/components/ComposerLogo/ComposerLogo.tsx
```diff
@@ -2,14 +2,16 @@
 // Copyright 2022 DXOS.org
 //
 
-import '@dxosTheme';
-
+import * as d3 from 'd3';
```
thure, line 5: Assets generally should not be animated with such a heavy library, though I know this is just something you’re experimenting with.

richburdon: yes currently experimental. but noted.

thure (changes requested):
This PR does substantially more than its title or description states. I would prefer other changes occur in separate PRs, also preferably with linked & prioritized issues.

packages/apps/composer-app/src/main.tsx
richburdon: let's try it and review.

thure: Please summarize all other changes in the description and mark this PR as `feat` or `fix` if it includes noticeable changes in the design. Screenshots & screen captures would also be very helpful when reviewing PRs like this.

PR #7266, opened 2024-07-16
fix(comments): Infinite loop

ZaymonFC:
The `scrollThreadIntoView` function was causing an infinite loop when selecting threads. This was due to unnecessary updates being triggered even when no actual changes were needed, leading to a cycle of re-renders and state updates.

Optimized the `scrollThreadIntoView` function to only trigger updates when absolutely necessary. Key changes include:

Added checks to determine if scrolling is actually needed.
Implemented a check to only update the selection if it's different from the current selection.

### DIFF
--- a/packages/ui/react-ui-editor/src/extensions/comments.ts
+++ b/packages/ui/react-ui-editor/src/extensions/comments.ts
@@ -134,15 +134,14 @@ const styles = EditorView.baseTheme({
   '&dark .cm-comment-current:hover': { backgroundColor: getToken('extend.colors.yellow.900') },
 });
 
-const createCommentMark = (id: string, isCurrent: boolean) => {
-  return Decoration.mark({
+const createCommentMark = (id: string, isCurrent: boolean) =>
+  Decoration.mark({
     class: isCurrent ? 'cm-comment-current' : 'cm-comment',
     attributes: {
       'data-testid': 'cm-comment',
       'data-comment-id': id,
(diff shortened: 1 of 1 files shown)


packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -545,17 +544,28 @@ export const scrollThreadIntoView = (view: EditorView, id: string, center = true
...
-      ],
-    });
+    const currentSelection = view.state.selection.main;
+    const currentScrollPosition = view.scrollDOM.scrollTop;
+    const targetScrollPosition = view.coordsAtPos(range.from)?.top;
+
+    const needsScroll =
+      targetScrollPosition !== undefined &&
+      (targetScrollPosition < currentScrollPosition ||
+        targetScrollPosition > currentScrollPosition + view.scrollDOM.clientHeight);
```
wittjosiah, line 556: wondering if we should have something slightly smaller than the full client height in case the thing is right at the bottom then you might still want to scroll a bit? maybe just leave as a todo...

wittjosiah (approved):

packages/ui/react-ui-editor/src/extensions/comments.ts
ZaymonFC: I think it's fine, it leaves the comments in the middle of the view when scrolling.

PR #7270, opened 2024-07-17
fix(comments): Remove non-null assertion from thread.anchor filter

ZaymonFC:
Throwing a lot of runtime errors (including in prod). Works just fine without this assertion, the type accepts `string | undefined`.


ZaymonFC:
Oh, looks like this just got fixed.

PR #7271, opened 2024-07-17
chore: graph fixes

wittjosiah:
- match nodes with slug by id to avoid slug/node mismatches when updating the deck
- trigger resolve node via waitForNode to ensure nodes are loaded when reloading the page


ZaymonFC (approved):
🔧 Nice

Issue #5544, opened 2024-02-05
[comments] comment should not be created until after first message is added to thread


wittjosiah:
Currently when a comment is added it shows an empty thread in the sidebar for everyone. When creating a comment, it shouldn't be persisted until the first message in the thread is added.

PR #7273, opened 2024-07-18
feat(comments): Stage threads until first comment

### LINKED ISSUE #7272: [comments] Comments should show up in editors in collections
```[tasklist]
### Tasks
- [ ] Show comment button in stack section text editors
- [ ] C11Y sidebar should update to reflect attended section
```

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/5544

As a side-effect of moving annotations source of truth into an extension on the thread plugin, comment annotations now show up in stack sections, without the plumbing to support them.

- Aiming to resolve #7272 as well.


packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -592,27 +603,41 @@ const hasActiveSelection = (state: EditorState): boolean => {
...
-  useEffect(() => {
-    if (view) {
-      // Check same document.
-      // NOTE: Hook might be called before editor state is updated.
+class ExternalCommentSync implements PluginValue {
+  private unsubscribe: () => void;
+
+  // TODO(Zan): How to achieve this with deep signal?
+  constructor(view: EditorView, id: string, threads: ReadonlySignal<ThreadType[]>) {
+    this.unsubscribe = effect(() => {
```
wittjosiah, line 611: I think it would be good to try to do this without directly depending on signals in the extension. Could it take a subscribe/get pair of callbacks?

ZaymonFC: Yeah that felt weird to me! Good call.

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -403,6 +413,11 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
             return [];
           }
 
+          // TODO(Zan): How to achieve this with deepsignal?
```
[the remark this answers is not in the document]
ZaymonFC: Okay I'll add more information to the to-do.

PR #7275, opened 2024-07-18
fix: Update protocol defs that make bufbuild fail.

richburdon:
- [x] resolves cyclical dependencies.
- [x] first enum value must be ZERO.
- [x] enum names must be unique within package (e.g., add SPACE_ prefix to SpaceState).


packages/core/protocols/src/proto/dxos/client/services.proto
```diff
@@ -374,7 +379,7 @@ message CreateEpochRequest {
     /// Replace the current automerge root with a new one specified by the user.
     REPLACE_AUTOMERGE_ROOT = 4;
 
-    /// Upgrade references data strucutre 
+    /// Upgrade references data strucutre
```
wittjosiah, line 382:
```suggestion
    /// Upgrade references data structure
```

packages/core/protocols/src/proto/dxos/echo/object.proto
```diff
@@ -5,13 +5,19 @@
...
+import "google/protobuf/descriptor.proto";
 
 import "dxos/keys.proto";
 import "dxos/halo/credentials.proto";
 import "dxos/echo/timeframe.proto";
 
 package dxos.echo.object;
 
+// TODO(burdon): preserve_any definition.
+extend google.protobuf.FieldOptions {
```
dmaretskyi, line 17: extract to `src/proto/field_options.proto`?

dmaretskyi:
Is there a reason this PR is based against a hotfix branch?

wittjosiah:
> Is there a reason this PR is based against a hotfix branch?

we had initially discussed putting it out as a hotfix last week

dmaretskyi:
> > Is there a reason this PR is based against a hotfix branch?
> 
> we had initially discussed putting it out as a hotfix last week

why?

wittjosiah:
> > > Is there a reason this PR is based against a hotfix branch?
> > 
> > 
> > we had initially discussed putting it out as a hotfix last week
> 
> why?

because @richburdon was trying to use it externally so needed them published. After making more substantial changes I believe he changed his mind on that

PR #7277, opened 2024-07-18
feat: beta credential

### LINKED ISSUE #6599: [composer] new user experience
we had an interstitial choice about "choosing an identity" before, and removing it proved to be "too far" away from ideal.

Custoemers say "i haven't signed into anything, what does it mean to sign out?" or "i haven't made this mine yet, where is my data, who am I in this system?". There are also technical problems like "what account should a document go into when opening an invitation link?" which are created by "default identity" from "progressive multiplayer" features.

Proposal:
A new landing page which does not scare users with "create identity" or "join existing device" but does better to explain the account system and uses clear "skip" language to take the default action (create identity). Choices about profile name and icon can be made here as well.

wittjosiah:
Resolves #6599


packages/apps/plugins/plugin-observability/src/ObservabilityPlugin.tsx
```diff
@@ -159,15 +159,14 @@ export const ObservabilityPlugin = (options: {
...
-        // TODO(burdon): Remove?
-        // if (clientPlugin.provides.firstRun) {
-        //   await dispatch({
-        //     action: ObservabilityAction.SEND_EVENT,
-        //     data: {
-        //       name: 'identity.created',
-        //     },
-        //   });
-        // }
+        if (!client.halo.identity.get()) {
```
nferch, line 162: under what circumstances would this be needed where https://github.com/dxos/dxos/pull/7277/commits/8d362975208c71462fddacf83cee4aa86a4e1fe6#diff-9e2f9960cf303a7ff57e192290d26318bb0405752c0b57b593b895e8650450c9R156 wouldn't work?

wittjosiah: this PR makes it such that identity isn't created during client plugin initialize anymore so this should handle the case when first landing on composer

richburdon (approved):

packages/apps/composer-app/src/plugins/welcome/components/Welcome/composer_splash.spline
wittjosiah, line 1: consider adding `*.spline` to git lfs (e.g. https://github.com/dxos/dxos/blob/main/.gitattributes#L14)

packages/apps/composer-app/src/plugins/welcome/components/Welcome/Welcome.tsx
```diff
@@ -0,0 +1,213 @@
...
+            {/* https://github.com/splinetool/react-spline?tab=readme-ov-file#spline-component-props */}
+            <Spline
+              className={mx(spline ? 'transition duration-1000 opacity-75' : 'opacity-0')}
+              scene={splineUrl}
+              onLoad={handleLoadEvent}
+            />
+          </div>
+
+          <div className='z-10 flex flex-col gap-8 p-8 bg-black bg-opacity-70'>
+            <h1 className="font-['Poiret One'] text-[80px]" style={{ fontFamily: 'Poiret One' }}>
```
wittjosiah, line 66: font doesn't seem to be loaded

PR #7299, opened 2024-07-22
refactor: graph optimizations to avoid long render tasks

wittjosiah:
- make graph expansion explicit
- don't expand the graph by default
- make graph expansion async
- swap setTimeout for requestAnimationFrame in a few couple effects

### DIFF
--- a/packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
+++ b/packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
@@ -143,8 +143,12 @@ const NodePlankHeading = ({
   const ActionRoot = node && popoverAnchorId === `dxos.org/ui/${DECK_PLUGIN}/${node.id}` ? Popover.Anchor : Fragment;
 
   useEffect(() => {
-    // Load actions for the node.
-    node && graph.actions(node);
+    const frame = requestAnimationFrame(() => {
+      // Load actions for the node.
+      node && graph.actions(node);
+    });
+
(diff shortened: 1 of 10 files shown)


packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -143,8 +143,12 @@ const NodePlankHeading = ({
...
   useEffect(() => {
-    // Load actions for the node.
-    node && graph.actions(node);
+    const frame = requestAnimationFrame(() => {
+      // Load actions for the node.
+      node && graph.actions(node);
+    });
+
+    return () => cancelAnimationFrame(frame);
   }, [node]);
```
thure, line 152: I’m somewhat wary of implementing this as a `useEffect`, though I can understand the motive here…

There is another similar implementation in this PR in `useNode` — would you be amenable to moving this logic into a named hook so we can iterate on the implementation there if needed?

thure (approved):
Stellar 🌟 — just wondering about DRYing the hook a bit, but not blocking.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
wittjosiah: I'm a little wary of factoring it out to specific hook because I agree it probably shouldn't happen in a hook ideally. I think you and I should review how this will in your new navtree work and try to make some changes there.

PR #7303, opened 2024-07-23
refactor(composer-app): Activate service credentials

richburdon:


packages/apps/composer-app/src/plugins/welcome/WelcomePlugin.tsx
```diff
@@ -95,43 +89,18 @@ export const WelcomePlugin = (): PluginDefinition<SurfaceProvides & Translations
       }
 
       // Existing beta users should be able to get the credential w/o a magic link.
-      if (identity) {
+      if (hubUrl && identity && token) {
```
wittjosiah, line 92: This needs to run even without a token to migrate identities over

packages/ui/react-ui-theme/src/config/colors.ts
```diff
@@ -225,6 +225,10 @@ export const semanticColors = {
     light: physicalColors.neutral['700'],
     dark: physicalColors.neutral['300'],
   },
+  error: {
```
wittjosiah, line 228: @thure does this already exist somewhere?

packages/apps/composer-app/src/plugins/welcome/credentials/header-codec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export type EncodedTimestamp = { __t: number };
+
+export interface KeyCodec<KeyType> {
+  isKey(value: any): value is KeyType;
+  encode(key: KeyType): EncodedKey;
+  decode(value: EncodedKey): KeyType;
+}
+
+export class HeaderCodec<Key> {
```
dmaretskyi, line 21: use `deepMapValues` from @dxos/util -- it handles recursion for you

packages/ui/react-ui-theme/src/config/colors.ts
thure: There is a fragment `messageValence` that could be refactored to use semantic tokens, however I have a better system ready to go with `@ch-ui/tokens` which would make both of these approaches obsolete. Basically this adds tech debt but not much.

PR #7304, opened 2024-07-24
feat: publish apps to Cloudflare Pages

nferch:
# TODO
## p0
* ~fold in npm publishing from `main.yml` and `publish.yml`~
* ~Disable CF CI for enabled branches~

## p1
* read secrets directly from 1p
* only conditionally publish main (or GC?)
* Unify with Docker build
* Automate/capture CF Pages environment config
* Output Github Actions summary with deployment URL and link to dash page


.github/workflows/cloudflare-publish.yml
```diff
@@ -2,16 +2,37 @@ on:
   push:
     branches:
       - staging
+      - nf/publish-apps
 
-name: staging
+name: Publish apps to Cloudflare Pages
```
wittjosiah, line 7: this also publishes to npm, no? or is that what you mean by dedupe npm publish?

.github/workflows/cloudflare-publish.yml
```diff
@@ -44,6 +65,8 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install
+        env:
+          CI: true
```
wittjosiah, line 69: does github not set this automatically?

.github/workflows/cloudflare-publish.yml
```diff
@@ -52,23 +75,21 @@ jobs:
...
-          DX_TELEMETRY_API_KEY: ${{ secrets.CLI_SEGMENT_API_KEY }}
 
-      - name: Bundle Shell
-        run: pnpm nx bundle shell
+      - name: Bundle Apps
+        run: pnpm nx run-many --target=bundle --parallel=3
         env:
           NODE_ENV: production
-          DX_ENVIRONMENT: production
+          NODE_OPTIONS: "--max_old_space_size=12288"
```
wittjosiah, line 86: does it actually need this much? maybe we shouldn't parallel 3 😅

.github/workflows/cloudflare-publish.yml
nferch: Yeah, wanted to look a bit on how to multistage/parallelize the job while still using the same build output, but wanted to ensure the CF Pages publishing bit worked as intended first.

.github/workflows/cloudflare-publish.yml
nferch: yeah, I guess it does

.github/workflows/cloudflare-publish.yml
nferch: I had just grabbed what I knew to work out of the Dockerfile. Is there any downside other than masking the issue of bundle getting too large?

What is the benefit of parallelization?

wittjosiah: yeah I think the main downside is masking the growing memory usage

main benefit of parallelization is that it runs faster

.github/workflows/env/main
```diff
@@ -0,0 +1,19 @@
...
+CLI_SEGMENT_API_KEY=${CLI_SEGMENT_API_KEY}
+COMPOSER_APP_SENTRY_DESTINATION=${COMPOSER_APP_SENTRY_DESTINATION}
+COMPOSER_APP_SEGMENT_API_KEY=${COMPOSER_APP_SEGMENT_API_KEY}
+DX_DISCORD_WEBHOOK_URL=${DX_DISCORD_WEBHOOK_URL}
+CLOUDFLARE_API_TOKEN=${CLOUDFLARE_API_TOKEN}
+CLOUDFLARE_ACCOUNT_ID=${CLOUDFLARE_ACCOUNT_ID}
+AGENT_HOSTING_SERVER="https://api.prod.agenthosting.dxos.network:8082/v1alpha1/"
+AGENT_HOSTING_TYPE=AGENTHOSTING_API
+# End common vars
+DX_HUB_URL="https://hub-staging.dxos.workers.dev/"
```
wittjosiah, line 17: we don't want this one set on `main` because we don't want the beta page to show up on that deployment

.github/workflows/env/production
```diff
@@ -0,0 +1,17 @@
...
+CLI_SEGMENT_API_KEY=${CLI_SEGMENT_API_KEY}
+COMPOSER_APP_SENTRY_DESTINATION=${COMPOSER_APP_SENTRY_DESTINATION}
+COMPOSER_APP_SEGMENT_API_KEY=${COMPOSER_APP_SEGMENT_API_KEY}
+DX_DISCORD_WEBHOOK_URL=${DX_DISCORD_WEBHOOK_URL}
+CLOUDFLARE_API_TOKEN=${CLOUDFLARE_API_TOKEN}
+CLOUDFLARE_ACCOUNT_ID=${CLOUDFLARE_ACCOUNT_ID}
+AGENT_HOSTING_SERVER="https://api.prod.agenthosting.dxos.network:8082/v1alpha1/"
+AGENT_HOSTING_TYPE=AGENTHOSTING_API
+# End common vars
+JWT_SECRET=${JWT_SECRET}
```
wittjosiah, line 17: is this used to decode the cookie? this will go away once you update that agent code?

wittjosiah, line 17: we should have hub url here as well, but point to prod hub

.github/workflows/publish-all.yml
```diff
@@ -0,0 +1,117 @@
...
+        env:
+          NODE_ENV: production
+          NODE_OPTIONS: "--max_old_space_size=12288"
+
+      - name: Publish Apps
+        run: .github/workflows/scripts/publish-apps.sh
+
+      - name: Publish to NPM
+        run: .github/workflows/scripts/publish.sh
+        if: env.BRANCH == 'production' || github.event.inputs.publish_npm == 'true'
```
wittjosiah, line 100: since staging is a manual push I think it's okay to include here as well. Usually we're testing npm packages on staging pushes as well

.github/workflows/publish-all.yml
```diff
@@ -0,0 +1,117 @@
...
+        if: env.BRANCH == 'production' || github.event.inputs.publish_devtools_extension == 'true'
+
+      - name: Upload Devtools extension
+        uses: fnkr/github-action-ghr@v1
+        env:
+          GHR_PATH: artifacts/
+          GITHUB_TOKEN: ${{ secrets.CREATE_PR_TOKEN }}
+        if: env.BRANCH == 'production' || github.event.inputs.publish_devtools_extension == 'true'
+
+      - name: Monitor publish
```
wittjosiah, line 116: this should only run if we published to npm

package.json
```diff
@@ -152,6 +152,7 @@
     "webdriverio": "^8.32.4",
     "wrangler": "^3.51.2"
   },
+  "packageManager": "pnpm@8.15.8",
```
wittjosiah, line 155: what's this for?

.github/workflows/staging.yml
wittjosiah, line 1: should main.yml be removed too?

.github/workflows/env/production
nferch: Same as above

package.json
nferch: If this is set, you don't have to declare it for the usage of `pnpm/action-setup`.

.github/workflows/staging.yml
nferch: Yeah, I had removed it but the removal must have gotten lost with a rebase or something.

PR #7308, opened 2024-07-24
chore(plugin-help): make sure that joyride starts even if the target isn't immediately available

wittjosiah:
The welcome plugin starts the joyride now but when it starts it the app hasn't rendered yet. This deals with that race condition by having the help context wait for the initial step to be available before setting running to true.

### DIFF
--- a/packages/apps/plugins/plugin-help/src/components/HelpContextProvider/HelpContextProvider.tsx
+++ b/packages/apps/plugins/plugin-help/src/components/HelpContextProvider/HelpContextProvider.tsx
@@ -29,6 +29,39 @@ const removeTargetClass = (target: string | HTMLElement) => {
   }
 };
 
+const getTarget = (step: Step) => {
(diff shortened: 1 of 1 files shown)


packages/apps/plugins/plugin-help/src/components/HelpContextProvider/HelpContextProvider.tsx
```diff
@@ -29,6 +29,39 @@ const removeTargetClass = (target: string | HTMLElement) => {
...
+              resolve();
+            }
+          }
+        });
+      });
+
+      observer.observe(document.body, { childList: true, subtree: true });
+    });
+  }
+};
```
thure, line 63: It might be nice to move this further upstream, it seems like it could be extremely useful, fine to start here of course.

thure (approved):
lgtm!

ZaymonFC (approved):
lgtm. Agree that wait for target could be a good util but maybe wait to move it until we need it again.

PR #7310, opened 2024-07-25
fix(threads): Thread sorting

### LINKED ISSUE #7309: [comments] Threads should be sorted by document position
Often threads on the same line show up in the wrong order.

ZaymonFC:
- Resolves #7309
- Adds a new color treatment for selected threads using our primary color
- Resolves graph action expansion bug


packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -273,7 +273,36 @@ export const DeckLayout = ({
...
+
+  const expandNode = useMemo(
+    () => async (node: Node) => {
+      await graph.expand(node, 'outbound', ACTION_GROUP_TYPE);
+      await graph.expand(node, 'outbound', ACTION_TYPE);
+    },
+    [graph],
+  );
+
+  // TODO(Zan): Maybe this should be a hook?
```
wittjosiah, line 286: I think we should aim to have graph expansion be triggered by user action...but that might not always be possible. In which case, yeah this would probably make sense as a hook.

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -357,24 +357,32 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
...
+                const getStartPosition = (cursor: string | undefined) => {
+                  const range = cursor ? getRangeFromCursor(accessor, cursor) : undefined;
+                  return range?.start ?? Number.MAX_SAFE_INTEGER;
+                };
+
+                const threads = doc.threads
                   .concat(state.staging[data.subject.id])
                   .filter(nonNullable)
-                  .toSorted((a, b) => state.threads[a.id] - state.threads[b.id]);
+                  .sort((a, b) => getStartPosition(a.anchor) - getStartPosition(b.anchor));
```
wittjosiah, line 375: sort vs toSorted? I suppose due to the concat/filter operations its not going to mutate the doc itself anyways

wittjosiah (approved):

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
ZaymonFC: Yeah `concat` and filter return new arrays so I'm not worried.

PR #7311, opened 2024-07-25
feat: Add object creation feature and Fix debug plugin in composer app

mykola-vrmchk:
https://github.com/user-attachments/assets/455cb720-e0bd-485c-b700-8657e23393ba


packages/apps/plugins/plugin-debug/package.json
```diff
@@ -25,9 +25,11 @@
   "dependencies": {
     "@braneframe/plugin-client": "workspace:*",
     "@braneframe/plugin-settings": "workspace:*",
+    "@braneframe/plugin-sketch": "workspace:*",
     "@braneframe/plugin-status-bar": "workspace:*",
     "@dxos/async": "workspace:*",
     "@dxos/automerge": "workspace:*",
+    "@dxos/client": "workspace:*",
```
wittjosiah, line 32: this shouldn't be needed `@dxos/react-client` re-exports everything

packages/apps/plugins/plugin-debug/src/components/DebugPanel.tsx
```diff
@@ -35,6 +39,18 @@ export const DebugPanel: FC<PropsWithChildren<{ menu: ReactNode }>> = ({ menu, c
...
+const MainOrArticle: FC<PropsWithChildren<{ role: 'main' | 'article' }>> = ({ role, children }) => {
+  return role === 'main' ? (
+    <Main.Content classNames={[baseSurface, fixedInsetFlexLayout, topbarBlockPaddingStart, bottombarBlockPaddingEnd]}>
+      {children}
     </Main.Content>
+  ) : (
+    <div role='none' className='row-span-2 rounded-t-md overflow-x-auto'>
+      {children}
+    </div>
   );
```
wittjosiah, line 55: @thure @ZaymonFC how have you handled this elsewhere?

packages/apps/plugins/plugin-debug/src/DebugPlugin.tsx
```diff
@@ -230,6 +231,40 @@ export const DebugPlugin = (): PluginDefinition<DebugPluginProvides> => {
...
                 return <DevtoolsArticle />;
               }
+
+              if (!object || typeof object !== 'object' || !settings.values.debug) {
+                return null;
+              }
+
+              if ('space' in object && isSpace(object.space)) {
+                return (
+                  <DebugSpace
```
wittjosiah, line 241: I think this should probably be combined w/ the `main` one to avoid the `onAddObjects` logic drifting. If you look at some of the other surface providers they do something like `const primary = data.active ?? data.object` prior to checking the types.

packages/apps/plugins/plugin-debug/src/components/DebugSpace.tsx
```diff
@@ -218,10 +189,7 @@ const DebugSpace: FC<{ space: Space; onAddObjects?: (objects: ReactiveObject<any
         </>
       }
     >
-      <div className={'shrink-0'}>
-        <SchemaList space={space} onCreate={handleCreate} />
-      </div>
-      <Json theme={themeMode} data={data} />
+      {<ObjectCreator space={space} onAddObjects={onAddObjects} />}
```
wittjosiah, line 192:
```suggestion
      <ObjectCreator space={space} onAddObjects={onAddObjects} />
```

packages/apps/plugins/plugin-debug/src/scaffolding/generator.ts
```diff
@@ -0,0 +1,131 @@
...
+          labelColor: 'black',
+          labelPosition: 0.5,
+          scale: 1,
+          size: 'm',
+        },
+        rotation: 0,
+        type: 'arrow',
+        typeName: 'shape',
+        x: 0,
+        y: 0,
```
wittjosiah, line 124: does this always draw it in the same place? maybe helpful to place them at random locations?

packages/core/echo/echo-generator/src/util.ts
```diff
@@ -7,3 +7,13 @@ export const range = <T>(fn: (i: number) => T | undefined, length: number): T[]
   Array.from({ length })
     .map((_, i) => fn(i))
     .filter(Boolean) as T[];
+
+export const randomText = (length: number) => {
```
wittjosiah, line 11: why here vs using a `@dxos/random` (i.e. faker) function?

packages/apps/plugins/plugin-debug/src/scaffolding/generator.ts
mykola-vrmchk: See video

wittjosiah: Right but that looks like it's only drawing one spiral per diagram? what happens if more generated, are they just drawn on top or at a new start point?

mykola-vrmchk: I do not get the question, it draws one spiral per diagram, the spiral size depends on the number of mutations, each mutation is a separate segment of the spiral, and currently, there is no way to mutate the diagram a second time. So it doesn't matter what starting coordinates we choose

wittjosiah: Ah I didn't realize that you couldn't mutate the diagram a second time. I though that the mutator chose a random object to mutate each time. Makes sense 👍

packages/apps/plugins/plugin-debug/src/components/DebugPanel.tsx
thure, line 55: Better, I think, for `DebugPanel` to implement everything both `main` and `article` have in common, then just wrap them differently in the `provides.components` perhaps.

Creating new interceding components like this can cause issues if e.g. we need to start forwarding refs — instead of calling `forwardRef` multiple times because of the interceding component here, you could just pass the ref directly to `DebugPanel`.

Not blocking, just something to consider.

packages/apps/plugins/plugin-debug/src/components/DebugPanel.tsx
```diff
@@ -35,6 +39,18 @@ export const DebugPanel: FC<PropsWithChildren<{ menu: ReactNode }>> = ({ menu, c
...
+  );
+};
+
+const MainOrArticle: FC<PropsWithChildren<{ role: 'main' | 'article' }>> = ({ role, children }) => {
+  return role === 'main' ? (
+    <Main.Content classNames={[baseSurface, fixedInsetFlexLayout, topbarBlockPaddingStart, bottombarBlockPaddingEnd]}>
+      {children}
     </Main.Content>
+  ) : (
+    <div role='none' className='row-span-2 rounded-t-md overflow-x-auto'>
```
thure, line 52: Does `row-span-2` have to do with the toolbar? If so, be sure this works when it’s both enabled and disabled.

packages/apps/plugins/plugin-debug/src/components/DevtoolsMain.tsx
```diff
@@ -14,10 +14,14 @@ import {
...
 
-  return (
+  return role === 'article' ? (
+    <div role='none' className='row-span-2 rounded-t-md overflow-x-auto'>
+      <Devtools client={client} services={client.services.services as ClientServices} />
+    </div>
+  ) : (
     <Main.Content classNames={[baseSurface, fixedInsetFlexLayout, topbarBlockPaddingStart, bottombarBlockPaddingEnd]}>
       <Devtools client={client} services={client.services.services as ClientServices} />
     </Main.Content>
```
thure, line 27: opportunity to DRY out the code here perhaps, since this is repeated in another file

thure (approved):
lgtm! a couple things to consider that could yield tidier componentry but nothing blocking.

PR #7316, opened 2024-07-25
chore: update publish for production

nferch:


.github/workflows/publish-all.yml
```diff
@@ -12,7 +12,8 @@ on:
       - staging
       - nf/publish-apps
       - main
-      # TODO: enable production and remove publish.yml
+  release:
+    types: [published]
```
wittjosiah, line 16: ooo one gotcha here is that I'm not sure the `branch` here is going to be set to `production`... release please runs on the `rc` branch and so the release gets tagged there. That commit then gets pushed to `production` but not until the next step.

The main reason we need to run on this trigger is to be able to upload things to the release so perhaps what we should do is:
- also include `production` in the branches list
- publish apps and npm packages from production but not release trigger
- upload artifacts from release trigger but not production
- in retrospect `publish_devtools_extension` is probably useless because I'm pretty sure it will only work if triggered from a release tag

nferch: What's the reason for uploading artifacts from the release trigger?

I had initially thought there'd be a downside in having to build twice, but now I'm thinking we should probably enable it for rc branches too?

wittjosiah: > What's the reason for uploading artifacts from the release trigger?

it's been quite a while since I first set this up, but iirc the action we're using for upload only works if run from the release trigger. Maybe that's how it associates the uploads to the correct release?

nferch: It looks like it bases the version on the version tag. so if we added the rc branches, I think it'd just work.

wittjosiah: @nferch wouldn't that trigger publish on every push to the `rc` branches though?

nferch: Yeah, but it shouldn't do any harm and is kinda what we want anyway? We'll know whether any of the commits to the rc branch fail to build or publish.

wittjosiah: for rc (and hotfix) I think we'd want to bundle & build but not publish apps/npm? but then do attempt to upload extension artifacts?

.github/workflows/env/production
```diff
@@ -14,4 +14,4 @@ CLOUDFLARE_ACCOUNT_ID=${CLOUDFLARE_ACCOUNT_ID}
 AGENT_HOSTING_SERVER="https://api.prod.agenthosting.dxos.network:8082/v1alpha1/"
 AGENT_HOSTING_TYPE=AGENTHOSTING_API
 # End common vars
-JWT_SECRET=${JWT_SECRET}
+DX_HUB_URL="https://hub-production.dxos.workers.dev/"
```
wittjosiah, line 17:
```suggestion
DX_HUB_URL="https://hub.dxos.network/"
```

richburdon, line 17: hub.dxos.network is the custom domain for the prod worker.

PR #7319, opened 2024-07-25
chore: add capabilities check to welcome plugin

wittjosiah:


packages/apps/composer-app/src/plugins/welcome/WelcomePlugin.tsx
```diff
@@ -112,9 +115,9 @@ export const WelcomePlugin = ({
 
       // Query for credential in HALO and skip welcome dialog if it exists.
       const subscription = client.halo.credentials.subscribe(async (credentials) => {
-        const credential = credentials.find(isServiceCredential);
+        const credential = credentials.filter(isServiceCredential).find(hasCapabilities(['composer:beta']));
```
richburdon, line 118: maybe change isServiceCredential to be matchServiceCredential('composer:beta') so that we evolve a single function (query) rather than many helpers?

richburdon (approved):

PR #7321, opened 2024-07-25
chore: remove datadog

nferch:


nferch:
cc @wittjosiah also removed Datadog from the CircleCI config, which didn't seem to be in use

PR #7324, opened 2024-07-25
feat: Undo object deletion

### LINKED ISSUE #6728: [composer] enable undo of deleted objects
Replace the deletion confirmation with an undo toast after the deletion has been run

ZaymonFC:
- Resolves #6728


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -1033,28 +1018,96 @@ export const SpacePlugin = ({
 
             case SpaceAction.REMOVE_OBJECT: {
               const object = intent.data?.object ?? intent.data?.result;
+              // TODO(Zan): Since we were only using caller to provide an anchor to the popover, do we still need this?
```
wittjosiah, line 1021: I think not

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -1033,28 +1018,96 @@ export const SpacePlugin = ({
...
+                      action: NavigationAction.ADD_TO_ACTIVE,
+                      data: { id: fullyQualifiedId(restoredObject) },
+                    });
+                  }
+
+                  return { data: true };
+                }
               }
-              break;
+              return { data: false };
```
wittjosiah, line 1110: based on the above if statement it seems like the if/elseif conditions cover 100% of cases so this will never be run? could second branch just be an `else`?

packages/apps/plugins/plugin-space/src/translations.ts
```diff
@@ -45,6 +45,8 @@ export default [
         'duplicate object label': 'Duplicate item',
         'delete collection label': 'Delete collection',
         'delete object label': 'Delete item',
+        'collection deleted label': 'Collection deleted',
+        'object deleted label': 'Item deleted',
```
wittjosiah, line 49: realizing that we use `item` in most user-facing language for objects...except `unnamed object label` 🤔

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -1033,28 +1018,96 @@ export const SpacePlugin = ({
...
-                            object,
-                            collection: intent.data?.collection,
-                          },
-                        },
-                      },
-                    ],
-                  ],
+                  data: true,
+                  undoable: {
+                    message: translations[0]['en-US'][SPACE_PLUGIN][undoMessageKey], // Consider using a translation key here
```
wittjosiah, line 1077: we're basically limited to this type of usage until we expose the translation utils outside of react. It is a known issue, but not one that we've encountered enough to prioritize fixing it

wittjosiah (approved):

packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
ZaymonFC: Oh, that comment is vestigial.

packages/apps/plugins/plugin-space/src/translations.ts
ZaymonFC: 🤔

PR #7325, opened 2024-07-25
chore(threads): Analytics

ZaymonFC:
- Resolves: https://github.com/dxos/devrel/issues/97


packages/apps/plugins/plugin-thread/src/hooks/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useOnTransition } from '@dxos/react-ui';
+
+export const useAnalyticsCallback = (space: string | undefined, name: string, meta?: any) => {
+  const dispatch = useIntentDispatcher();
+
+  return useCallback(
+    (dynamicMeta?: any) => {
+      void dispatch({
+        action: ObservabilityAction.SEND_EVENT,
+        data: { name, ...meta, ...dynamicMeta, space },
```
wittjosiah, line 20: also all of the non-event name properties need to be inside `properties` field

packages/apps/plugins/plugin-thread/src/hooks/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+    [dispatch, name, meta, space],
+  );
+};
+
+export const useOnEditAnalytics = (message: ReactiveObject<any>, editing: boolean) => {
+  const space = getSpace(message);
+
+  const onEditAnalytics = useAnalyticsCallback(space?.id, 'threads.comment-edited', {
+    messageId: message.id,
+    messageLength: message.text.length,
```
wittjosiah, line 32: will this capture the initial length or the edited length? probably the initial?

ZaymonFC: Pretty sure it's mutating the doc on keypress like our other editors do so it should be relatively up to date. Clicking done-editing is a bit of a formality right.

packages/apps/plugins/plugin-thread/src/hooks/analytics.ts
ZaymonFC: Ahh. Sweet.

PR #7334, opened 2024-07-29
feat(echo): metrics

yarolegovich:
### Details

* `EchoDataMonitor` class integration with `LevelDBStorageAdapter` and `EchoNetworkAdapter` for storage and replication metrics capture.
* `EchoDataMonitor` manages a set of remote and local metrics.
* `EchoDataMonitor` exports stats through `EchoHost.echo-stats` diagnostics channel. 
* New stats are displayed in `StatsPanel`. 
* `SlidingWindowSummary` util class for computing sliding window statistics.
* If `@trace.resource()` annotated object implements `TimeAware` interface its tick method is going to be invoked by trace processor every 1s.
* `CircularBuffer` returns evicted element on push.

Indexing should be covered in a separate PR.

### Sentry metrics.

* Network metrics are sometimes split into `replication` and `collection-sync` because `collection-sync` payloads grow with the number of spaces while replication messages should stay the same.
* `-rate` should be interpreted as "when there was activity in the app here's how many operations were performed". Related counters help us understand how often the app is not doing this type of operation, eg.: 
   * `dxos.echo.storage.load{status=busy} / dxos.echo.storage.load` is equal to 0.05. Out of all recorded app running time 5% of seconds had some storage reads.
   * `p90(dxos.echo.storage.load-rate)` is equal to 1000. During these seconds in 90% of cases load rate has not exceeded 1000 operations per second.

#### Distributions:
* `dxos.echo.storage.(load|store)-rate`
* `dxos.echo.network.(receive|send)-rate`
* `dxos.echo.storage.bytes-(loaded|stored)`
* `dxos.echo.(replication|collection-sync).bytes-(sent|received)` split by message type.
* `dxos.echo.(replication|collection-sync).send-duration` split by message type.

#### Counters:
* `dxos.echo.storage.(load|store)` split by `status: 'busy' | 'idle'`.
* `dxos.echo.network.(send|receive)` split by `status: 'busy' | 'idle'`.
* `dxos.echo.(replication|collection-sync).send-status` split by `success: boolean`.

### Local metrics.

image

Extended information can be retrieved from `EchoDataMonitor` using `dxos.tracing`, for example time series or split of last sent messages by peer id.
```ts
const monitor = dxos.tracing.findResourcesByClassName('EchoDataMonitor')[0].instance.deref()
```
image


dmaretskyi (approved):

dmaretskyi:
For comparison indexed-db supports around 3000 reads per second according to [some](https://github.com/raineorshine/indexeddb-benchmark) benchmarks

PR #7336, opened 2024-07-30
fix: Update exports in package manifests

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/apps/plugins/plugin-attention/package.json
```diff
@@ -9,15 +9,17 @@
...
+      },
+      "types": "./dist/types/src/index.d.ts"
     },
     "./meta": {
       "browser": "./dist/lib/browser/meta.mjs",
-      "import": "./dist/lib/browser/meta.mjs",
-      "require": "./dist/lib/node/meta.cjs",
-      "node": "./dist/lib/node/meta.cjs"
+      "node": {
+        "default": "./dist/lib/node/meta.cjs"
```
dmaretskyi, line 20: The idea here is to eventually have separate "import" and "require" exports for ESM and CJS respectively

packages/common/crypto-names/package.json
```diff
@@ -6,11 +6,19 @@
...
+      "browser": "./dist/lib/browser/index.mjs",
+      "node": {
+        "default": "./dist/lib/node/index.cjs"
+      },
+      "types": "./dist/types/src/index.d.ts"
+    }
+  },
+  "types": "dist/types/src/index.d.ts",
+  "typesVersions": {
+    "*": {}
```
wittjosiah, line 20: seen this in a couple packages, is it necessary for something even though it's empty?

wittjosiah (approved):

packages/common/crypto-names/package.json
dmaretskyi: Not necessary but not harmful, I just didn’t want to complicate the tool code to handle this case

dmaretskyi: If it’s too annoying I can add a case to remove it

PR #7339, opened 2024-07-30
chore(e2e): Tests for deletion undo

ZaymonFC:


packages/apps/composer-app/src/playwright/collections.spec.ts
```diff
@@ -91,5 +91,31 @@ test.describe('Collection tests', () => {
       await host.toggleCollectionCollapsed(0);
       await expect(host.getObjectLinks()).toHaveCount(2);
     });
+
+    test('deletion undo restores collection', async () => {
+      await host.closeToast();
```
wittjosiah, line 96: this won't work in CI, need to just figure out why you're needing to do this locally? what's your `DX_ENVIRONMENT` set to?

ZaymonFC: Hmm, for some reason my `DX_ENVIRONMENT` is my computer name.

wittjosiah: yup, that's what mine is as well. Does it not end in `.local` or `.lan` mine automatically does, hence the notify condition in the observability plugin being based on that

ZaymonFC: It's `zan.mbp`. I'll add an export!

PR #7344, opened 2024-07-31
fix(threads): React hooks rule violation

ZaymonFC:
### Issue
Documents with comments were crashing the app, particularly when refreshing with the 'complementary' part present in the URL. This was caused by a violation of React's Rules of Hooks in the Thread Plugin.

### Root Cause
Analytics hooks were being called within a case statement in the Thread Plugin, violating React's rule that hooks must be called at the top level of a component.

### Notes
- The error manifested far downstream, making initial debugging challenging.
- This highlights the importance of adhering strictly to React's Rules of Hooks.

### Future Improvements
Consider adding the ESLint Rules of Hooks plugin to catch similar issues earlier in the development process.

### DIFF
--- a/packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
+++ b/packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
@@ -58,7 +58,6 @@ import {
   ChatHeading,
   ThreadArticle,
 } from './components';
-import { useAnalyticsCallback } from './hooks';
 import meta, { THREAD_ITEM, THREAD_PLUGIN } from './meta';
 import translations from './translations';
 import { ThreadAction, type ThreadPluginProvides, type ThreadSettingsProps } from './types';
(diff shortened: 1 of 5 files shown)


packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -396,13 +395,20 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
 
                 const qualifiedSubjectId = fullyQualifiedId(doc);
                 const attention = attentionPlugin?.provides.attention?.attended ?? new Set([qualifiedSubjectId]);
-                const attendableAttrs = useAttendable(qualifiedSubjectId);
+                const attendableAttrs = createAttendableAttributes(qualifiedSubjectId);
                 const space = getSpace(doc);
                 const context = space?.db.getObjectById(firstMainId(location?.active));
                 const { showResolvedThreads } = getViewState(qualifiedSubjectId);
 
-                const onCreateAnalytics = useAnalyticsCallback(space?.id, 'threads.thread-created');
```
wittjosiah, line 404: fwiw a lot of this logic inside this if statement could also be factored out into a component to prevent the hooks violation but this seems fine too

wittjosiah (approved):

PR #7349, opened 2024-07-31
feat(plugin-space): add space limit dialog

### LINKED ISSUE #7233: [composer] add a restriction to composer to limit the size of spaces
In order to prevent spaces from getting beyond the current capabilities add a limit to prevent object creation beyond a certain size.

Once this limit is hit attempting to create an object in the space should not succeed and instead pop up a dialog explaining the limitation and suggesting that they contact the team if they're interested in larger spaces.

wittjosiah:
Resolves #7233

### DIFF
--- a/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
+++ b/packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
@@ -90,7 +90,7 @@ import {
 
 const ACTIVE_NODE_BROADCAST_INTERVAL = 30_000;
 const OBJECT_ID_LENGTH = 60; // 33 (space id) + 26 (object id) + 1 (separator).
-const SPACE_SIZE_LIMIT = 100;
+const SPACE_MAX_OBJECTS = 100;
 
 export const parseSpacePlugin = (plugin?: Plugin) =>
   Array.isArray((plugin?.provides as any).space?.enabled) ? (plugin as Plugin<SpacePluginProvides>) : undefined;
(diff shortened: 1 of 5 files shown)


packages/apps/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -89,6 +90,7 @@ import {
 
 const ACTIVE_NODE_BROADCAST_INTERVAL = 30_000;
 const OBJECT_ID_LENGTH = 60; // 33 (space id) + 26 (object id) + 1 (separator).
+const SPACE_SIZE_LIMIT = 100;
```
richburdon, line 93: SPACE_MAX_OBJECTS?

packages/apps/plugins/plugin-space/src/components/LimitDialog.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react';
+
+import { Button, Dialog, useTranslation } from '@dxos/react-ui';
+
+import { SPACE_PLUGIN } from '../meta';
+
+export const LimitDialog = () => {
+  const { t } = useTranslation(SPACE_PLUGIN);
+
+  return (
```
richburdon, line 14: this is good, but from a longer term perspective is there a way to get away from dialogs and move towards everything being in settings. e.g., status panel in settings that shows a usage bar (and later, actions that can increase your limits). i.e., how can we build incrementally towards where we are going rather than introducing new and temporary UX idioms (that might even be more work than extending what we have).

richburdon (approved):

PR #7351, opened 2024-07-31
feat: upgrade beta credentials

wittjosiah:


packages/apps/composer-app/src/plugins/welcome/credentials/util.ts
```diff
@@ -54,3 +54,56 @@ export const activateAccount = async ({
...
+  });
+  if (!response.ok) {
+    throw new Error('profile fetch failed', { cause: response.statusText });
+  }
+
+  return response.json();
+};
+
+/**
+ * Upgrade  credential.
```
richburdon, line 88: TODO to move to @dxos/edge hub-protocols

packages/apps/composer-app/src/plugins/welcome/credentials/util.ts
```diff
@@ -54,3 +54,56 @@ export const activateAccount = async ({
...
+ * @param params.credential
+ */
+export const getProfile = async ({
+  hubUrl,
+  credential,
+}: {
+  hubUrl: string;
+  credential: Credential;
+}): Promise<ProfileResponse> => {
+  const response = await fetch(new URL('/account/profile', hubUrl), {
```
richburdon, line 77: (generically), if we're abstracting the request/response, then return the profile, not the ProfileResponse?

packages/apps/composer-app/src/plugins/welcome/credentials/util.ts
```diff
@@ -54,3 +54,56 @@ export const activateAccount = async ({
...
+    headers: { Authorization: `Bearer ${codec.encode(credential)}` },
+  });
+  if (!response.ok) {
+    throw new Error('profile fetch failed', { cause: response.statusText });
```
richburdon, line 81: these are 400 errors, so bad requests, rather than exceptions? i'm not sure i'd do anything different here, but like the comment about it feels a bit leaky.

just something to consider when we evolve the hub-protocols client lib.

richburdon (approved):

Issue #7345, opened 2024-07-31
Deck horizontal overscroll


wittjosiah:
- Allow edge-most planks to be scrolled to the centre of the screen
- Should be able to be turned off but start with it on by default

PR #7356, opened 2024-08-01
feat(deck): Overscroll

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/7345

The initial mode keeps the plank centered within the deck. I'm wondering if we should have a strategy that looks at which sidebars are open, accounts for their size and keeps the plank in the center of the screen instead.

https://github.com/user-attachments/assets/d2c31b63-dd18-4426-8c2e-9ef80a262a1e


packages/apps/plugins/plugin-deck/src/types.ts
```diff
@@ -16,12 +16,16 @@ import type {
 export const NewPlankPositions = ['start', 'end'] as const;
 export type NewPlankPositioning = (typeof NewPlankPositions)[number];
 
+export const OverScrollOptions = ['none', 'centering'] as const;
+export type OverScroll = (typeof OverScrollOptions)[number];
```
thure, line 20: Teeny nit, I think you rightly treat this as a tightly-compounded word elsewhere, but loosely-compounded here — maybe:
```suggestion
export const OverscrollOptions = ['none', 'centering'] as const;
export type Overscroll = (typeof OverscrollOptions)[number];
```

thure (approved):
lgtm!

PR #7362, opened 2024-08-02
chore: fix react-ui-stack exports

wittjosiah:
Previously the exports were general so even though the build was an esm build "for browser" it was being used everywhere


thure (approved):
👍

PR #7368, opened 2024-08-02
feat: Darken deck background

### LINKED ISSUE #7346: Dark background/separation between planks (customizable)
- allow deck background to be changed via settings
- start with small number of presets
- default should be not the same colour as the plank background (as it is currently)

thure:
This PR:
- resolves #7346 
- mints a new semantic token for deck
- adjusts spacing to account for new separation.

https://github.com/user-attachments/assets/7cbc69f9-56b9-479f-9603-9e683954181e


ZaymonFC (approved):
✨ LGTM!

PR #7370, opened 2024-08-03
fix: Resolve fontsource import for specific patch version

richburdon:


thure (changes requested):
This breaks italics.

thure (approved):
Fixed.

PR #7371, opened 2024-08-03
feat: sheet plugin styles

richburdon:
- [x] Basic plugin.
- [x] Basic toolbar integration. 
- [x] Range selection (robust to scrolling).
- [x] Cell selection within formula (with bugs).

Screenshot 2024-08-05 at 11 16 25 AM


packages/apps/plugins/plugin-sheet/src/components/Sheet/Overlay.tsx
```diff
@@ -0,0 +1,63 @@
...
+    fromBounds.left += scrollProps.scrollLeft;
+    if (rangeBounds) {
+      rangeBounds.top += scrollProps.scrollTop;
+      rangeBounds.left += scrollProps.scrollLeft;
+    }
+  }
+
+  return (
+    <>
+      <div
```
thure, line 48: Is this presentational as I suspect? If so, please always:
```suggestion
      <div role='none'
```

packages/apps/plugins/plugin-sheet/src/components/Sheet/Overlay.tsx
```diff
@@ -0,0 +1,63 @@
...
+        className={mx('z-[10] absolute border border-black dark:border-white opacity-50', 'pointer-events-none')}
+        style={fromBounds}
+      />
+      {rangeBounds && (
+        <div
```
thure, line 53:
```suggestion
        <div role='none'
```

packages/apps/plugins/plugin-sheet/src/components/Sheet/Overlay.tsx
```diff
@@ -0,0 +1,63 @@
...
+          className={mx(
+            'z-[10] absolute border bg-neutral-900 dark:bg-neutral-100 border-black dark:border-white opacity-10',
```
thure, line 55: These styles will need to be cleaned up — we have semantic tokens that work in both themes.

packages/apps/plugins/plugin-sheet/src/components/Sheet/Sheet.stories.tsx
```diff
@@ -51,7 +54,42 @@ const Story = ({ cells, ...props }: SheetProps & { cells?: Record<string, CellSc
...
+const SheetToolbar = () => {
+  const { selected, setFormat } = useSheetContext();
+  const handleAction: ToolbarProps['onAction'] = ({ type }) => {
+    const styles: Record<string, string> = {
+      left: 'text-left',
+      center: 'text-center',
+      right: 'text-right',
+      bold: 'font-bold',
+      highlight: 'bg-blue-100/25 border-yellow-100',
+    };
```
thure, line 78: Define constants outside of render functions.

packages/apps/plugins/plugin-sheet/src/components/SheetComponent.tsx
```diff
@@ -1,7 +1,18 @@
 //
-// Copyright 2024 DXOS.org
+// Copyright 2023 DXOS.org
```
thure, line 2: ??

packages/apps/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+//
+// Alignment
+//
+
+// TODO(burdon): Detect and display current state.
+const alignmentOptions: ButtonProps[] = [
+  { type: 'left', Icon: TextAlignLeft, getState: (state) => false },
+  { type: 'center', Icon: TextAlignCenter, getState: (state) => false },
+  { type: 'right', Icon: TextAlignRight, getState: (state) => false },
```
thure, line 75: Passing icon components will soon be deprecated. Components should begin to use `iconSymbol` strings and render an `<svg><use href={`/icons.svg#{iconSymbol}`}/></svg>` (we should add a component for this).

These icons would be `ph--text-align-left--regular`, `ph--text-align-center--regular`, and `ph--text-align-right--regular`.

packages/apps/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+//
+// Actions
+//
+
+const Actions = () => {
+  // const { onAction } = useToolbarContext('Actions');
+  const { t } = useTranslation(SHEET_PLUGIN);
+  return (
+    <>
```
thure, line 153: Unneeded fragment.

packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -135,7 +134,19 @@ const ToolbarButton = ({ Icon, children, ...props }: ToolbarButtonProps) => {
...
+//
+
+const HeadingIcons: { [key: string]: Icon } = {
+  '0': Paragraph,
+  '1': TextHOne,
+  '2': TextHTwo,
+  '3': TextHThree,
+  '4': TextHFour,
+  '5': TextHFive,
+  '6': TextHSix,
```
thure, line 148: See my other remark about using strings instead of importing and passing around components.

packages/apps/plugins/plugin-sheet/src/components/Sheet/Overlay.tsx
richburdon: ACK. please drop references in the PR comments. i know these exist, just have trouble remembering how to find them when i'm in flow.

packages/apps/plugins/plugin-sheet/src/components/Sheet/Sheet.stories.tsx
richburdon: yeah, this is a very first cut hack. i'm not sure how we want to do this. i've tried to use some of the componentry and logic from the editor toolbar. lots of possibilities to factor out.

packages/apps/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
richburdon: ok cool -- that's close to what i'm doing via `twind` in the edge monorepo (purely server side).

packages/apps/plugins/plugin-sheet/src/components/Sheet/Overlay.tsx
thure: `colors.ts` in react-ui-theme has the current semantic colors near the bottom.

fwiw I have a better solution in-hand that should make dx for using semantic colors far easier, let's discuss soon.

PR #7372, opened 2024-08-04
chore: override react version

wittjosiah:
composer stopped loading due to signals using a different version of react


thure (approved):
lgtm 👍

PR #7375, opened 2024-08-05
feat: Spreadsheet cell editor (prototype)

bvaughn:
Sharing a prototype I put together over the weekend. Not sure how CI will react to these changes because I'm new to this repo.

- Lezer to parse spreadsheet syntax
- Lexical to render cell editor (with syntax highlighting and code-completion)

Both Lezer and Lexical integrations _could_ be used in other places as well. (Lexical has a lot of pre-built plugins, and I've written several other open source ones [here](https://github.com/replayio/devtools/tree/main/packages/replay-next/components/lexical/plugins).) The type-ahead plugin I've written here is generic enough to be used with other languages/syntaxes. The "code plugin" is currently hard-coded to handle spreadsheet syntax (though `parseTokens` could be updated to support more languages/syntaxes if we wanted).

cc @richburdon

### DIFF
--- a/packages/apps/plugins/plugin-sheet/package.json
+++ b/packages/apps/plugins/plugin-sheet/package.json
@@ -43,12 +43,22 @@
     "@dxos/react-client": "workspace:*",
     "@dxos/react-ui-editor": "workspace:*",
     "@dxos/util": "workspace:*",
+    "@lexical/react": "^0.6.5",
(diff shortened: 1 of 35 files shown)


bvaughn:
I _think_ the CI failures may be some resolution problem where two versions of the `@codemirror/view` package are being installed. Not sure.

PR #7379, opened 2024-08-06
feat: Use cloudflare stun/turn server

mykola-vrmchk:
- **feat: Add ice provider to config protobuf**
- **Add mock**
- **Add failure test**
- **Use helper function**
- **Fix config**


packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -110,7 +115,7 @@ export class ClientServicesHost {
     this._runtimeParams = runtimeParams ?? {};
 
     if (config) {
-      this.initialize({ config, transportFactory, signalManager });
+      void this.initialize({ config, transportFactory, signalManager });
```
dmaretskyi, line 118: if fetch fails this will throw and unhandled rejection

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -201,7 +206,7 @@ export class ClientServicesHost {
     const {
       connectionLog = true,
       transportFactory = createSimplePeerTransportFactory({
-        iceServers: this._config?.get('runtime.services.ice'),
+        iceServers: this._config ? await getIceServers(this._config) : undefined,
```
dmaretskyi, line 209: I think the fetch logic should be inside the transport factory. We should send this request as a part of opening a webrtc connection, but cache the response for the subsequent connections.

There's also a case where we might re-request the ice config for some reason (invalidation, network change). No need to handle this now, but the implementation should allow us to cover this case in the future

PR #7397, opened 2024-08-08
chore: ensure over item is never empty

wittjosiah:
if `event.over` is `null` then `pick` returns an empty object which is an invalid over item


thure (approved):
yes please

PR #7398, opened 2024-08-08
fix(deck): Disable plank soloing on mobile

ZaymonFC:


ZaymonFC:
@wittjosiah Can we merge this into the RC, was there anything blocking the merge?

PR #7402, opened 2024-08-10
chore: debug wireframes and style changes: dialog width, solo plank padding

richburdon:
NOTE: I have left some styles prefixed with "_" to temporarily disable them pending review.

- [x] debug plugin wireframe option.

- [x] wider default dialog width and style to break words for technical errors.

- [x] center stack button.

Screenshot 2024-08-12 at 1 33 52 PM

- [x] align section with article (remove padding)
- ISSUE: discuss borders

Screenshot 2024-08-12 at 1 34 31 PM

- [x] remove padding for plank solos.

Screenshot 2024-08-12 at 1 36 25 PM


wittjosiah:
marked as draft pending review of `NOTE` in description

packages/ui/react-ui-thread/src/Message/Message.tsx
```diff
@@ -59,7 +59,7 @@ export const Message = ({ timestamp, children, ...messageMeta }: MessageProps) =
   return (
     <MessageMeta {...messageMeta} continues>
       <p className='grid grid-cols-[1fr_max-content] gap-2 pie-2'>
-        <Avatar.Label classNames={['truncate text-sm font-medium', !messageMeta.authorName && 'fg-description']}>
+        <Avatar.Label classNames={['truncate text-sm font-thin', !messageMeta.authorName && 'fg-description']}>
```
thure, line 62:
```suggestion
        <Avatar.Label classNames={['truncate text-sm font-normal', !messageMeta.authorName && 'fg-description']}>
```
We should definitely not use `font-thin` and `text-sm` at the same time — no software in the ecosystem does this because it is very illegible.

thure (dismissed):
You can probably safely remove the spacing rules that you disabled, however we must not use `font-thin` in Composer unless the text size is quite large. I’ve suggested another weight as a workaround, though perhaps we should discuss at length.

packages/ui/react-ui-thread/src/Message/Message.tsx
richburdon: ok i'll try fg-subdued

PR #7403, opened 2024-08-10
feat: excalidraw plugin

richburdon:
- [x] New Excalidraw plugin
- [x] Refactor Automerge mapping with Tldraw into common adapter.
- [x] Change Tldraw to update when changes are complete (i.e., remove throttling).


packages/apps/plugins/plugin-excalidraw/src/components/SketchComponent/SketchComponent.tsx
```diff
@@ -23,27 +23,65 @@ export type SketchComponentProps = {
...
+  //  https://github.com/excalidraw/excalidraw/blob/master/packages/excalidraw/components/App.tsx
+  useEffect(() => {
+    const flash = () => {
+      setTimeout(() => {
+        window.dispatchEvent(new Event('resize'));
+      });
+    };
+
+    flash();
+    const deck = findAncestorBySelector(containerRef.current!, 'article')?.parentElement;
```
thure, line 63: This is a native DOM method now: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

packages/apps/plugins/plugin-excalidraw/src/SketchPlugin.tsx
```diff
@@ -115,6 +115,11 @@ export const SketchPlugin = (): PluginDefinition<SketchPluginProvides> => {
       },
       surface: {
         component: ({ data, role }) => {
+          // TODO(burdon): Fix.
```
wittjosiah, line 118: fix what?

packages/apps/plugins/plugin-excalidraw/src/SketchPlugin.tsx
```diff
@@ -115,6 +115,11 @@ export const SketchPlugin = (): PluginDefinition<SketchPluginProvides> => {
...
+          if (data.object instanceof DiagramType && data.object.canvas?.schema !== EXCALIDRAW_SCHEMA) {
```
wittjosiah, line 119: why not include this condition on the instance of checks below? currently this will break the settings panel because that surface doesn't pass `object`

packages/apps/plugins/plugin-excalidraw/src/components/SketchComponent/SketchComponent.tsx
```diff
@@ -23,27 +23,65 @@ export type SketchComponentProps = {
...
+    // Detect moved in deck.
+    const observer = new MutationObserver(() => flash());
+    observer.observe(deck, { attributes: false, childList: true, subtree: false });
+    return () => observer.disconnect();
+  }, []);
+
+  // Menu action.
+  const handleRefresh = () => {
+    // excalidrawAPIRef.current?.setToast({ message: 'Refresh' });
+    excalidrawAPIRef.current?.updateScene({ elements: adapter.getElements() });
```
wittjosiah, line 77: what does this do? I thought I saw in a video you shared that this was live, so why is a refresh button needed?

packages/apps/plugins/plugin-excalidraw/src/hooks/useStoreAdapter.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!object) {
+      return;
+    }
+
+    if (object.canvas?.schema !== EXCALIDRAW_SCHEMA) {
+      log.warn('invalid schema', { schema: object.canvas?.schema });
+      return;
+    }
+
+    setTimeout(async () => {
```
wittjosiah, line 30: store reference to timeout and cancel it in the effect return callback to help avoid some async races

packages/apps/plugins/plugin-sketch/src/SketchPlugin.tsx
```diff
@@ -115,6 +115,11 @@ export const SketchPlugin = (): PluginDefinition<SketchPluginProvides> => {
       },
       surface: {
         component: ({ data, role }) => {
+          // TODO(burdon): Fix.
+          if (data.object instanceof DiagramType && data.object.canvas?.schema !== TLDRAW_SCHEMA) {
```
wittjosiah, line 119: same as other plugin, this is going to break settings

packages/apps/plugins/plugin-sketch/src/hooks/useStoreAdapter.tsx
```diff
@@ -2,32 +2,35 @@
...
-      forceUpdate({});
-    } catch (err) {
-      // TODO(burdon): User error handling for corrupted data.
-      log.catch(err);
+    if (object.canvas?.schema !== TLDRAW_SCHEMA) {
+      log.warn('invalid schema', { schema: object.canvas?.schema });
+      return;
     }
 
+    setTimeout(async () => {
```
wittjosiah, line 27: same as other hook, cancel timeout on effect cleanup

packages/apps/plugins/plugin-sketch/src/meta.tsx
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 
 import { pluginMeta } from '@dxos/app-framework';
 
-// TODO(burdon): Rename TLDRAW_PLUGIN.
+// TODO(burdon): Rename TLDRAW_PLUGIN?
```
wittjosiah, line 10: I think tldraw aims less to be a recognizable product and wants to be white-labelled. Most people won't know what tldraw is so we should name this literally (sketch, diagram, whiteboard, etc.) Similarly, Excalidraw is a bit more of a product and people who want it will want it because they specifically like excalidraw, so that makes more sense to re-use their name.

packages/apps/types/src/schema/diagram.ts
```diff
@@ -4,12 +4,14 @@
...
 
+// TODO(burdon): Move defs to plugin.
+export const EXCALIDRAW_SCHEMA = 'excalidraw.com/2';
 export const TLDRAW_SCHEMA = 'tldraw.com/2';
 
 export class CanvasType extends TypedObject({ typename: 'dxos.org/type/Canvas', version: '0.1.0' })({
   /** Fully qualified external schema reference */
   schema: S.optional(S.String),
-  // TODO(burdon): Default doesn't seem to work with create().
+  // TODO(burdon): Default doesn't seem to work with create()?
```
wittjosiah, line 14: default didn't exist when `create` was written so that's not surprising to me

packages/apps/plugins/plugin-sketch/src/util/util.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return value;
+};
+
+// TODO(burdon): AM Utils.
+
+/**
+ *
+ */
+// TODO(burdon): Comment.
```
wittjosiah, line 64: 👀

packages/apps/plugins/plugin-excalidraw/src/components/SketchComponent/SketchComponent.tsx
richburdon: there's no refresh button -- this is triggered on pointer up (i.e., after drag).

packages/apps/plugins/plugin-sketch/src/meta.tsx
richburdon: ok good point.

packages/apps/plugins/plugin-sketch/src/util/util.ts
richburdon: @dmaretskyi what does this do?

PR #7383, opened 2024-08-07
feat: single shot export

wittjosiah:
Resolves #6490
Resolves #6100 
Resolves #6112 
Resolves #6113 

- file system directory structure matches the graph structure
- allow duplicate copies of objects on export
- reconcile changes to the duplicates on import
- fallback serializer for an object simply serializes that object and does not follow or - don't roll up references (this will likely result in missing data from exports for unsupported data types, but this can be backfilled later because we only care about docs and diagrams today)
- export wipes entire export directory on each export
- handling file name changes while keeping file linked to an object may not be possible from within the browser sandbox, so import as update will not yet be supported and can be subject to future research

PR #7406, opened 2024-08-12
chore(deck): Refactor layout representation and manipulation

ZaymonFC:


packages/apps/plugins/plugin-deck/project.json
```diff
@@ -1,27 +1,19 @@
...
-        "platforms": [
-          "browser"
-        ]
+        "entryPoints": ["{projectRoot}/src/index.ts", "{projectRoot}/src/meta.ts"],
+        "platforms": ["browser"]
       }
     },
-    "lint": {}
+    "lint": {},
+    "test": {}
```
wittjosiah, line 16: if it doesn't work quickly don't bother, but if it's just as easy, try using vitest instead. We aspire to move over to vitest and off of our custom test infra and the less there is to port the easier that will be.

wittjosiah: see https://github.com/dxos/dxos/pull/7383 for an example or `@dxos/random-access-storage`

wittjosiah: currently, yes

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -53,39 +56,19 @@ export type DeckLayoutProps = {
   flatDeck?: boolean;
   toasts: ToastSchema[];
   onDismissToast: (id: string) => void;
-  location: Location;
+  layoutParts: LayoutParts;
   attention: Attention;
+  // TODO(Zan): Deprecate slots.
```
wittjosiah, line 61: how would those things be controlled?

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -295,69 +294,73 @@ export const DeckLayout = ({
...
-    ? Object.keys(location.active).length < 1
-      ? { sidebar: NAV_ID }
-      : location.active
-    : { sidebar: NAV_ID, main: [location.active].filter(Boolean) as string[] };
-  const sidebarSlug = firstSidebarId(activeParts);
-  const sidebarNode = useNode(graph, sidebarSlug);
-  const sidebarAvailable = sidebarSlug === NAV_ID || !!sidebarNode;
-  const fullScreenSlug = firstFullscreenId(activeParts);
+  const sidebarNodeId = useMemo(() => firstIdInPart(layoutParts, 'sidebar'), [layoutParts]);
+  const sidebarAvailable = useMemo(() => sidebarNodeId === NAV_ID, [sidebarNodeId]);
```
wittjosiah, line 298: fwiw my understanding is that memoizing things like this which take a negligible amount of time to compute doesn't actually save any time because the memoization needs to do an equality check at minimum as well

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -295,69 +294,73 @@ export const DeckLayout = ({
   const { t } = useTranslation(DECK_PLUGIN);
   const { graph } = useGraph();
 
-  // TODO(wittjosiah): Finding nodes in the graph should probably not be done at the top-level of layout.
```
wittjosiah, line 298: doesn't seem like this has been resolved?

packages/apps/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
```diff
@@ -103,16 +95,14 @@ export const NavTreeContainer = ({
     await dispatch({
       action: NavigationAction.OPEN,
       data: {
-        // TODO(thure): don’t bake this in, refactor this to be a generalized approach.
```
wittjosiah, line 106: I think this comment was in reference to the comments popping open vs the deck model

ZaymonFC: I think it's legacy from when we did a whole bunch of logic here in manually splicing the active parts.

packages/apps/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -174,22 +189,17 @@ export const DeckPlugin = ({
...
-        if (activeParts) {
-          return intentPlugin?.provides.intent.dispatch({
-            // TODO(thure): handle popstate
-            action: NavigationAction.OPEN,
-            data: { activeParts },
-          });
-        }
+        const layoutFromUri = uriToActiveParts(window.location.pathname);
+        const startingLayout = location.active;
+        location.active = mergeLayoutParts(layoutFromUri, startingLayout);
```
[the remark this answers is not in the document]
ZaymonFC: We only do this once on page load so I'm okay with the direct merge and set.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
ZaymonFC: I was seeing an infinite re-render issue until I memoised. Didn't narrow in on the loop precisely, but we're going to be slicing this up soon?

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
ZaymonFC: @thure What's the new version of this, variants and css variables?

packages/apps/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
ZaymonFC: Maybe it's referring to the string formatting

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
ZaymonFC: 👍

packages/apps/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
wittjosiah: @ZaymonFC my recollection was that we didn't want the navtree to know specifically about comments and that there should be a generalized approach for "related planks" to automatically be opened.

Issue #7422, opened 2024-08-12
[navtree] smarter opening of branches


wittjosiah:
- [x] remember previously open branches
- [x] open branches when creating a new object
- [ ] #7481

PR #7430, opened 2024-08-13
feat(react-ui-navtree): Adjust signifier for current items & focus indication

### LINKED ISSUE #7420: [navtree] open state indication
- change open state indication to align with common design patterns (full width background change for the item)
- this effectively replaces the grey square in the navtree
- retain current attention & presence indicators (blue square + presence dots)

thure:
This PR:
- resolves #7420.
- partly #7422.

https://github.com/user-attachments/assets/2a3d3fed-cbc1-4a2b-9a6d-fe484facd298


packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -43,14 +44,23 @@ export type NavTreePluginProvides = SurfaceProvides &
...
 
+type NavTreeState = { root?: NavTreeItemGraphNode };
+
+type NavTreeSettings = { openItemIds?: OpenItemIds };
+
 export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
-  const state = create<{ root?: NavTreeItemGraphNode; openItemIds?: OpenItemIds }>({});
+  const state = create<NavTreeState>({});
+
+  const settings = new LocalStorageStore<NavTreeSettings>('dxos.org/settings/navtree', {
```
wittjosiah, line 54: A couple things on this:
- these aren't settings this is just state still
- local storage store doesn't automatically map everything to local storage so you can include the root here too and just not map it to local storage

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -70,9 +80,9 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
       getActions(graph, state.root);
 
       // TODO(thure): Do this dynamically.
```
wittjosiah, line 82: move comment up to where openids are specified?

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -43,14 +44,23 @@ export type NavTreePluginProvides = SurfaceProvides &
...
+  const settings = new LocalStorageStore<NavTreeSettings>('dxos.org/settings/navtree', {
+    openItemIds: { root: true, 'dxos.org/plugin/space-spaces': true, 'dxos.org/plugin/files': true },
+  });
 
   let graphPlugin: Plugin<GraphProvides> | undefined;
   let graph: Graph | undefined;
 
   const handleOpenItemIdsChange = (nextOpenItemIds: OpenItemIds) => {
-    state.openItemIds = nextOpenItemIds;
+    // TODO(thure): This might become a localstorage leak; openItemIds that no longer exist should be removed from this map.
```
wittjosiah, line 62: if they're wholly overwritten every time they change then it's a bounded leakage and doesn't seem like a problem?

thure: It’s not wholly overwritten, I don’t think.

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -70,9 +80,9 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
       getActions(graph, state.root);
 
       // TODO(thure): Do this dynamically.
-      state.openItemIds = { root: true, 'dxos.org/plugin/space-spaces': true, 'dxos.org/plugin/files': true };
+      settings.prop({ key: 'openItemIds', storageKey: 'openItemIds', type: LocalStorageStore.json() });
```
wittjosiah:
```suggestion
      settings.prop({ key: 'openItemIds', storageKey: 'open-item-ids', type: LocalStorageStore.json() });
```
established pattern is to use kebab case for the storage key

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
wittjosiah: > settings.values.openItemIds = nextOpenItemIds

does this not replace the whole array? how does the leakage happen? does `nextOpenItemIds` leak forward item ids that aren't in the tree if they were passed in via props?

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
thure: I found instances of both patterns established — this probably shouldn’t be a dev’s responsibility to figure out, given naming is among the hardest of coding problems.

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
thure: > does nextOpenItemIds leak forward item ids that aren't in the tree if they were passed in via props?

Yes.

packages/apps/plugins/plugin-navtree/src/NavTreePlugin.tsx
wittjosiah: yeah, this is a temporary crutch. At some point this will be done automatically. The expectation is that `create` will have pluggable persistence (e.g. echo, local storage, etc.)

PR #7436, opened 2024-08-13
chore: clear url when reset is triggered from shell

### LINKED ISSUE #7414: [composer] reset should clear the url
this is a small task for large qol around development, after reset anything referenced in the url is gone

wittjosiah:
Resolves #7414


ZaymonFC (approved):
✨ LGTM

PR #7437, opened 2024-08-13
fix(plugin-registry): prevent plugins from jumping around when toggling enabled

### LINKED ISSUE #7415: [composer] toggling an experimental plugin off in the settings panel shouldn't cause it to disappear
currently toggling plugins on and off is quite jarring because of how they are sorted, toggling a plugin on or off shouldn't cause it disappear from the screen

wittjosiah:
Resolves #7415


packages/sdk/app-framework/src/plugins/PluginHost/PluginContext.tsx
```diff
@@ -26,6 +26,7 @@ export type PluginContext = {
   /**
    * All available plugins.
    */
+  // This is metadata rather just ids because it includes plugins which have not been fully loaded yet.
```
ZaymonFC, line 29:
```suggestion
  // This is metadata rather then just ids because it includes plugins which have not been fully loaded yet.
```

ZaymonFC (approved):
Fun solution.

thure (approved):
swell ✨

PR #7438, opened 2024-08-13
refactor(plugin-stack): stop exposing "stack" language to end-user

### LINKED ISSUE #7416: [composer] standardize stack/collection language across the app
- the language used in the app has not all kept up as we've evolved the interfaces and concepts in the app and needs an audit of how different terms are being presented and used
- proposal: remove `stack` from user-facing language, at this stage stack is just the component used to render a collection and so there's no need to introduce multiple terms

wittjosiah:
Resolves #7416


ZaymonFC (approved):
:shipit:

thure (approved):
👌

PR #7439, opened 2024-08-13
fix: unify deletion implementation

### LINKED ISSUE #7418: [composer] unclear when an action will be lossy or not
- 'dissolve' vs 'delete' when items within a collection will not also get deleted? 'remove' feels overloaded and consequences are unclear
- align language between removing objects in navtree vs in stack

wittjosiah:
Resolves #7418

- deletion now deletes objects within collections as well (+ restores them)
- deleting a section in a stack defers to the same deletion intent
- section removal translation updated to `Delete` as well to align
- cascades deletion to referenced objects
- action to remove deleted objects in space limit dialog


thure (approved):
lgtm, let’s hope we arrive at a moment where we can expose the true powers of collections

PR #7443, opened 2024-08-14
fix(client): broadcast to other workers to close on worker startup

### LINKED ISSUE #6087: [composer] updates break shared workers
we have a lock on the stored data which probably didn't allow the new worker to initiallize?

image

### LINKED ISSUE #6298: Worker does not exit when tab crashes

wittjosiah:
Resolves #6087
Resolves #6298

### DIFF
--- a/packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
+++ b/packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
@@ -3,6 +3,7 @@
 //
 
 import { Trigger } from '@dxos/async';
+import { DEFAULT_WORKER_BROADCAST_CHANNEL } from '@dxos/client-protocol';
 import { type Config } from '@dxos/config';
 import { Context } from '@dxos/context';
 import { invariant } from '@dxos/invariant';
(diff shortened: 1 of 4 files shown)


packages/sdk/client/src/worker/onconnect.ts
```diff
@@ -11,6 +11,8 @@ import { TRACE_PROCESSOR } from '@dxos/tracing';
 import { mountDevtoolsHooks } from '../devtools';
 import { LOCK_KEY } from '../lock-key';
 
+// change hash 2
```
dmaretskyi, line 14: 👁️ 👁️

packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
```diff
@@ -50,15 +56,26 @@ export class WorkerRuntime {
...
+    acquireLock,
+    releaseLock,
+    onStop,
+  }: WorkerRuntimeOptions) {
+    this._configProvider = configProvider;
     this._acquireLock = acquireLock;
     this._releaseLock = releaseLock;
+    this._onStop = onStop;
+    this._broadcastChannel = new BroadcastChannel(channel);
+    this._broadcastChannel.onmessage = async (event) => {
```
dmaretskyi, line 71: make sure its its disposed https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel/close

packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
```diff
@@ -50,15 +56,26 @@ export class WorkerRuntime {
...
+      if (event.data === 'stop') {
```
dmaretskyi, line 72: make it a `{ action: 'stop' }` object for extensibility

dmaretskyi (approved):
LGTM

Issue #6402, opened 2024-04-24
Wrapped reference API proposal


dmaretskyi:
### Motivation

Make working with lazy-loaded refrences more intuitive

### Current behaviour

RIght now when accesing a reference field on a proxy we automatically dereference it and return the target object as another proxy.
If the object is not loaded, we return `undefined` and load the object in the background.

### Suggested behaviour

This is gonna be a **breaking** API change.

Reading references from proxies will return wrapper objects:

```ts
type Ref<T> = {
  // load and return the target
  deref(): Promise<T>

  // read current state but might return undefined, reactive with signal semantics
  // This is primarly for usage inside React components and provides feature-parity with the current API.
  get target(): T | undefined

  // todo: some way to read raw reference data without dereferencing: space key, object id
}
```

### Examples

```ts
export class DocumentType extends TypedObject({ typename: 'braneframe.Document', version: '0.1.0' })({
  title: S.optional(S.string),
  content: ref(TextV0Type),
  comments: S.optional(S.mutable(S.array(CommentSchema))),
}) {}

declare const document: DocumentType;


const documentContent = await document.content.deref()
```

```ts

export class FolderType extends TypedObject({ typename: 'braneframe.Folder', version: '0.1.0' })({
  name: S.optional(S.string),
  objects: S.mutable(S.array(ref(Expando))),
}) {}


declare const folder: FolderType;


const allObjectsLoaded = await Promise.all(folder.objects.map(ref => ref.deref()))
```

PR #7446, opened 2024-08-14
feat: API to read a wrapped reference

dmaretskyi:
based on https://github.com/dxos/dxos/issues/6402


packages/core/echo/echo-db/src/echo-handler/echo-handler.test.ts
```diff
@@ -620,4 +620,15 @@ describe('Reactive Object with ECHO database', () => {
...
     expect(getDatabaseFromObject(another)).not.to.be.undefined;
   });
+
+  test('explicit reference API', async () => {
+    const { db, graph } = await builder.createDatabase();
+    graph.schemaRegistry.addSchema([TestSchemaType]);
+
+    const obj = db.add(create({ string: 'Object 1', another: create(Expando, { string: 'Object 2' }) }));
+
+    expect(readReference(obj, 'another')!.dxn.toString()).to.eq(`dxn:echo:@:${obj.another.id}`);
```
wittjosiah, line 630: is there a reason `obj.another` can't return the reference rather than needing this helper?

dmaretskyi: The eventual goal is for it to do exactly that, this helper is in-place to allow us to experiment with the new API without breaking backwards compatibility.

PR #7449, opened 2024-08-14
fix: IPFS plugin

nferch:
* render videos
* display errors when upload occurs


nferch:
@thure some placeholder UI when there's upload failures so the user isn't left hanging.

Will be swapping out the IPFS backend at some point, but I assume the UI and user flows will remain the same.

cc @wittjosiah

PR #7460, opened 2024-08-15
fix(deck): Editor toolbar breaks instrinsic plank sizing on mobile

ZaymonFC:


packages/ui/react-ui-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -72,9 +72,9 @@ const ToolbarRoot = ({ children, onAction, classNames, state }: ToolbarProps) =>
     <ToolbarContextProvider onAction={onAction} state={state}>
       <DensityProvider density='fine'>
         <ElevationProvider elevation='chrome'>
-          <NaturalToolbar.Root classNames={['is-full shrink-0 overflow-x-auto overflow-y-hidden p-1', classNames]}>
-            {children}
-          </NaturalToolbar.Root>
+          <div role='none' className='is-full shrink-0 overflow-x-auto overflow-y-hidden' style={{ contain: 'layout' }}>
+            <NaturalToolbar.Root classNames={['p-1', classNames]}>{children}</NaturalToolbar.Root>
```
thure, line 76: Is it possible to use `NaturalToolbar.Root asChild`? Would prefer extra nodes if possible, but if that doesn’t work here this is fine.

thure (approved):
Excellent, just one non-blocking question ✨

PR #7463, opened 2024-08-16
fix(composer): fix tldraw crash when bindings are not found

yarolegovich:
### Details

Was happening consistently when I was trying to connect block centers with an arrow.
You can see in database update log there's a batch with the same elements in added and deleted lists.
`decode(map[id])` in `updateModel` was returning undefined resulting into a crash inside `TLStore`. 

image

### DIFF
--- a/packages/apps/plugins/plugin-sketch/src/util/base-adapter.ts
+++ b/packages/apps/plugins/plugin-sketch/src/util/base-adapter.ts
@@ -163,7 +163,7 @@ export abstract class AbstractAutomergeStoreAdapter<Element extends BaseElement>
           this.onUpdate({
             updated: Array.from(updated)
               .map((id) => decode(map[id]))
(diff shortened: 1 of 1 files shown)


packages/apps/plugins/plugin-sketch/src/util/base-adapter.ts
```diff
@@ -160,7 +161,9 @@ export abstract class AbstractAutomergeStoreAdapter<Element extends BaseElement>
 
         if (updated.size || deleted.size) {
           this.onUpdate({
-            updated: Array.from(updated).map((id) => decode(map[id])),
+            updated: Array.from(updated)
+              .map((id) => decode(map[id]))
+              .filter(nonNullable), // Modified elements that were eventually removed
```
richburdon, line 166: period

richburdon (approved):

PR #7468, opened 2024-08-16
feat(solo): Persistent solo mode

### LINKED ISSUE #7423: [deck] persistent solo mode
- The app can be toggled between Deck and Solo mode (via a button in the plank)
  - deck/solo mode is persisted in local storage
  - first run starts in solo mode with view of default space in a stack
- In Solo mode clicking an item in the navtree should change the current selection.
- When exiting Solo mode, if the currently visible item is not in the deck it is added to the deck (per deck insertion setting), otherwise scroll to that item
- When deck mode is active, clicking on a navtree item should toggle the visibility of the item in the deck
  - consider this vs scroll to plank behaviour, how else scroll to plank might be handled, and decide which is most useful/understandable behaviour (related to ui perf & navtree open state)
- when opening a child object, scroll to the section in the collection if the collection is already open
  - show currently attended object in navtree + share with peers (follow up)
- deck state is stored in local storage, url is either an fq object id (solo mode) or `/` (deck)
- plank sizes should be persisted with planks in localstorage
- when moving deck state out of url & into local storage state, remember that reset should clear that state so that the deck starts empty

ZaymonFC:
- Resolves #7423


packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -371,7 +365,6 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
                           threads={threads}
                           detached={detached}
                           currentId={attention.has(qualifiedSubjectId) ? state.current : undefined}
-                          context={context}
```
wittjosiah, line 374: (realizing this functionality was experimental and we don't need it fully working to land this) did you have a place for how this might be restored? if so, perhaps a todo would be useful?

packages/common/local-storage/src/store.ts
```diff
@@ -154,6 +154,15 @@ export class LocalStorageStore<T extends object> {
     return this;
   }
 
+  /**
+   * Expunges all store-related items from local storage.
+   */
+  expunge() {
+    Object.keys(localStorage)
```
wittjosiah, line 161: I think this should probably iterate over the keys in the store? Because some stores share the same prefix and this might clear unexpected things

wittjosiah (approved):

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
```diff
@@ -573,15 +587,95 @@ export const DeckLayout = ({
...
+                overscroll={overscroll === 'centering'}
+                classNames={mx(
+                  'absolute inset-0',
+                  !flatDeck && 'surface-deck',
+                  slots?.wallpaper?.classNames,
+                  slots?.deck?.classNames,
+                )}
+                solo={true}
+              >
+                {layoutParts.solo.map((layoutEntry, index) => {
```
wittjosiah, line 614: wondering if this should only take the first element considering its supposed to be "solo"?

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
ZaymonFC: Context was both never resolving, and also the prop was never being used so I axed it, admittedly without getting to the bottom of what it was used for in the past.

packages/common/local-storage/src/store.ts
ZaymonFC: Ah I see. I'll make that change.

packages/apps/plugins/plugin-deck/src/components/DeckLayout.tsx
ZaymonFC: There's some logic so that the array only ever contains 0 / 1 element. Doing the map is the simplest thing here in JSX I think.

packages/apps/plugins/plugin-thread/src/ThreadPlugin.tsx
wittjosiah: it was used for some of the AI demos which depended on knowing what you were currently looking at when you sent the message. Make sure @richburdon sees this, he may prefer is stay around commented out because hooking some of that back up is still on his todo list

PR #7469, opened 2024-08-16
refactor: share space button use intent

wittjosiah:


ZaymonFC (approved):
Nice. Have you done a search for other usages of `shell.shareSpace`?

wittjosiah:
> Have you done a search for other usages of shell.shareSpace?

looking now and I only see something it the github plugin which is currently broken anyways so I think its all good

PR #7498, opened 2024-08-20
refactor: break up deck layout

wittjosiah:


packages/apps/plugins/plugin-deck/src/components/DeckLayout/ActiveNode.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const ActiveNode = ({ id }: { id?: string }) => {
+  const { graph } = useGraph();
+  const activeNode = useNode(graph, id);
+  useNodeActionExpander(activeNode);
+
+  return (
+    <div role='none' className='sr-only'>
+      {/* TODO(wittjosiah): Weird that this is a surface, feel like it's not really render logic.
+            Probably this lives in React-land currently in order to access translations? */}
```
thure, line 20: I think this is from before we had a cogent way of resolving fallback titles etc.

PR #7504, opened 2024-08-20
feat(deck): Persist plank sizing in local storage

ZaymonFC:
- Resolves: https://github.com/dxos/dxos/issues/7501


thure (approved):
lgtm!

packages/sdk/app-framework/src/plugins/common/layout.ts
```diff
@@ -95,6 +97,7 @@ export enum LayoutAction {
   SET_LAYOUT = `${LAYOUT_ACTION}/set-layout`,
   SET_LAYOUT_MODE = `${LAYOUT_ACTION}/set-layout-mode`,
   SCROLL_INTO_VIEW = `${LAYOUT_ACTION}/scroll-into-view`,
+  UPDATE_PLANK_SIZE = `${LAYOUT_ACTION}/update-plank-size`,
```
wittjosiah, line 100: could this be framed as a layout update somehow? the idea of these common definitions is that they aren't coupled to our implementation

packages/sdk/app-framework/src/plugins/common/layout.ts
```diff
@@ -67,6 +67,8 @@ export const Layout = z.object({
     .string()
     .optional()
     .describe('The identifier of a component to scroll into view when it is mounted.'),
+
+  plankSizing: z.record(z.string(), z.number()).describe('The size of the planks in the layout.'),
```
wittjosiah, line 71: ideally this should be a deck extension of this type since it's pretty deck-specific as opposed to a more general layout component

packages/ui/react-ui-deck/src/components/Deck/Deck.tsx
```diff
@@ -80,12 +80,35 @@ const defaults = {
...
   defaultSize = defaults.size,
+  size,
+  setSize,
   children,
   boundary,
-}: PropsWithChildren<{ defaultSize?: number; boundary?: 'start' | 'end' }>) => {
-  const [size, setSize] = useState(defaultSize);
+}: PropsWithChildren<{
+  defaultSize?: number;
+  size?: number;
```
wittjosiah, line 89: do these need to be separate props? could default just be the default of size?

PR #7538, opened 2024-08-23
fix(plugin-deck): handle deletion better in solo mode

wittjosiah:
Applies the same logic in CLOSE as OPEN was using.


ZaymonFC (approved):
I like it.

PR #7543, opened 2024-08-23
refactor(plugin-markdown): Unify Markdown editor components

thure:


packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -53,36 +40,25 @@ import {
...
+const isDataMarkdownish = (data: any): data is { object: { id: string; text: string } } => {
+  return (
+    data.object &&
+    typeof data.object === 'object' &&
+    'id' in data.object &&
+    typeof data.object.id === 'string' &&
+    'text' in data.object &&
+    typeof data.object.text === 'string'
+  );
+};
```
wittjosiah, line 52:
```suggestion
/**
 * Checks if an object conforms to the interface needed to render an editor.
 */
const isEditorModel = (data: any): data is { id: string; text: string } => {
  return (
    data &&
    typeof data === 'object' &&
    'id' in data &&
    typeof data.id === 'string' &&
    'text' in data &&
    typeof data.text === 'string'
  );
};
```

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -244,135 +218,54 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
                 return (
                   <DocumentEditor
-                    viewMode={getViewMode(doc.id)}
-                    toolbar={settings.values.toolbar}
                     document={doc}
-                    extensions={extensions}
-                    onCommentSelect={handleCommentSelect}
-                    onFileUpload={handleFileUpload}
-                    onViewModeChange={handleViewModeChange}
+                    extensions={state.values.extensionProviders.reduce((acc: Extension[], provider) => {
```
wittjosiah, line 235: this isn't memoized, is that going to be a problem?

packages/apps/plugins/plugin-markdown/src/components/DocumentEditor.tsx
```diff
@@ -36,9 +64,10 @@ const DocumentEditor = ({ document: doc, extensions: _extensions = [], ...props
           setFallbackName(doc, text);
         },
       }),
-      _extensions,
+      providerExtensions,
+      baseExtensions,
     ],
-    [doc, doc.content, _extensions, identity],
+    [doc, doc.content, baseExtensions, identity],
```
wittjosiah, line 70: should include provide extensions too?

wittjosiah: also space if it's being computed outside the memo

packages/apps/plugins/plugin-markdown/src/components/DocumentEditor.tsx
```diff
@@ -58,17 +87,32 @@ const DocumentEditor = ({ document: doc, extensions: _extensions = [], ...props
     };
   }, [doc]);
 
-  if (!doc.content) {
```
wittjosiah, line 61: this wasn't necessary?

packages/apps/plugins/plugin-markdown/src/components/DocumentSection.tsx
wittjosiah, line 1: 👋

packages/apps/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
```diff
@@ -50,28 +51,31 @@ const useTest = (view?: EditorView) => {
...
   toolbar?: boolean;
+  scrollPastEnd?: boolean;
+  inputMode?: EditorInputMode;
+  viewMode?: EditorViewMode;
   onViewModeChange?: (mode: EditorViewMode) => void;
   onCommentSelect?: (id: string) => void;
   onFileUpload?: (file: File) => Promise<FileInfo | undefined>;
 } & Pick<UseTextEditorProps, 'initialValue' | 'selection' | 'scrollTo' | 'extensions'>;
 
 // TODO(wittjosiah): Factor out main styles, reuse for all markdown editors, rename to MarkdownEditor.
```
wittjosiah, line 65: this PR should resolve this comment

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
thure: Maybe… I waffled back and forth about passing `extensionProviders` directly. I may just do that.

packages/apps/plugins/plugin-markdown/src/components/DocumentEditor.tsx
thure: I moved this to the surface renderer, we shouldn’t be returning `null` so late and we can check this value a lot sooner.

packages/apps/plugins/plugin-markdown/src/components/DocumentEditor.tsx
thure: https://www.npmjs.com/package/eslint-plugin-react-hooks 👀

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
thure: Memoized now, though it’s only handled by `DocumentEditor` — should the upstream `MarkdownEditor` do this instead and filter by a possibly falsy `document`?

wittjosiah: I think that's likely necessary. Unfortunately comments is somewhat a required extension at the moment because it's still a little coupled so that needs to always be available.

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -244,135 +221,49 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
-                    toolbar={settings.values.toolbar}
                     document={doc}
-                    extensions={extensions}
-                    onCommentSelect={handleCommentSelect}
-                    onFileUpload={handleFileUpload}
-                    onViewModeChange={handleViewModeChange}
+                    extensionProviders={state.values.extensionProviders}
+                    settings={settings.values}
+                    viewMode={getViewMode(doc.id)}
+                    onViewModeChange={(viewMode: EditorViewMode) => {
```
wittjosiah, line 241: since doc is defined outside the switch I think I'd leave this out there too so it can be wrapped in `useCallback` the downstream effect of recreating this every render worry me

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -244,135 +221,49 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
                 );
-              } else if (
-                data.object &&
-                typeof data.object === 'object' &&
-                'id' in data.object &&
-                typeof data.object.id === 'string' &&
-                'text' in data.object &&
-                typeof data.object.text === 'string'
-              ) {
+              } else if (isEditorModel(data)) {
```
wittjosiah, line 251:
```suggestion
              } else if (isEditorModel(data.object)) {
```

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
wittjosiah: nit: put callback as last prop (also I'd put role/coordinate at the top)

packages/apps/plugins/plugin-markdown/src/components/DocumentEditor.tsx
```diff
@@ -6,28 +6,73 @@ import { EditorView } from '@codemirror/view';
...
+type DocumentEditorProps = {
+  document: DocumentType;
+  settings: MarkdownSettingsProps;
+} & Omit<MarkdownEditorProps, 'id' | 'inputMode' | 'toolbar' | 'extensions'> &
+  Pick<MarkdownPluginState, 'extensionProviders'>;
 
 /**
  * Editor for a `DocumentType`.
  */
 // TODO(wittjosiah): Reconcile with DocumentSection & DocumentCard.
```
wittjosiah, line 34: This can be removed now (or maybe moved over to the card component, which should be removed eventually.

packages/apps/plugins/plugin-markdown/src/components/index.ts
```diff
@@ -5,20 +5,18 @@
...
 export * from './HeadingMenu';
 export * from './Layout';
 export * from './MarkdownSettings';
 
 // Lazily load components for content surfaces.
 export const DocumentCard = React.lazy(() => import('./DocumentCard'));
 export const DocumentEditor: LazyExoticComponent<DocumentEditorType> = React.lazy(() => import('./DocumentEditor'));
-export const DocumentSection: LazyExoticComponent<DocumentSectionType> = React.lazy(() => import('./DocumentSection'));
-export const EditorMain = React.lazy(() => import('./EditorMain'));
+// export const DocumentSection: LazyExoticComponent<DocumentSectionType> = React.lazy(() => import('./DocumentSection'));
```
wittjosiah, line 21:
```suggestion
```

packages/apps/plugins/plugin-markdown/src/components/EditorMain.stories.tsx
wittjosiah, line 1: I forgot to rename this file to match the component

wittjosiah (commented):
Looking good, nice work, thanks for jumping in here!

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
thure: The guard returns a type at the `data` scope

packages/apps/plugins/plugin-markdown/src/MarkdownPlugin.tsx
thure: I disagree with the callbacks-last style in this case, rather here I prefer callbacks are paired with the prop they control, which is true here. Callbacks-last might be reasonable for a single-element component like Button which don’t have controllable state, but that isn’t always the case and thus not always helpful IME.

We could put controlled props last and satisfy both styles *sometimes*, but my preference would interleave callbacks with their props.

thure: This is further complicated by another preference (of mine) in prop order, that flags occur last.

PR #7553, opened 2024-08-26
ci: bundle after env vars are set

wittjosiah:


nferch (approved):
Should fix the missing env vars issue.

Makes me wonder whether we should run this script once per app, and move the loop into Github Actions, which would enable a parallel bundle/deploy and (slightly?) less complicated handling of env vars.

Issue #7539, opened 2024-08-23
[table] fast data entry


wittjosiah:
- align with keyboard controls on sheet
- enter creates a new row below if editing the last row
- otherwise, enter moves to next line

PR #7560, opened 2024-08-26
feat(tables): Fast data entry

ZaymonFC:
- Resolve: https://github.com/dxos/dxos/issues/7539

https://github.com/user-attachments/assets/a312f176-a712-473d-9631-ee5e3de93661

### DIFF
--- a/packages/ui/react-ui-table/src/components/Cell/Cell.tsx
+++ b/packages/ui/react-ui-table/src/components/Cell/Cell.tsx
@@ -4,7 +4,7 @@
 
 import { useFocusableGroup } from '@fluentui/react-tabster';
 import { type Cell as CellType, type RowData } from '@tanstack/react-table';
-import React, { type PropsWithChildren } from 'react';
+import React, { useCallback, type PropsWithChildren } from 'react';
 
 import { tdRoot } from '../../theme';
 import { useTableContext } from '../Table/TableContext';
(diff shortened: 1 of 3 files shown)


packages/ui/react-ui-table/src/components/Cell/Cell.tsx
```diff
@@ -18,11 +18,45 @@ const CELL_NAME = 'Cell';
...
+      const pollForNewRow = () => {
+        const newNextSibling = currentRow.nextElementSibling;
+        if (newNextSibling instanceof HTMLTableRowElement) {
+          focusNextCell(newNextSibling);
+        } else {
+          requestAnimationFrame(pollForNewRow);
+        }
+      };
+      requestAnimationFrame(pollForNewRow);
+    }
```
thure, line 55: Oof, a shame this is necessary 🥲

ZaymonFC: And yet... we must.

packages/ui/react-ui-table/src/components/Table/Table.stories.tsx
```diff
@@ -323,15 +323,15 @@ export const InsertDelete = {
     };
 
     return (
-      <div>
+      <div className='flex flex-col gap-4'>
```
thure, line 326:
```suggestion
      <div className='space-y-4'>
```
Should avoid `flex` for these use-cases, it has caused subtle bugs in the past.

packages/ui/react-ui-table/src/components/Cell/Cell.tsx
ZaymonFC: Actually I should probably add a max depth to this because it will break the editable storybooks.

thure (approved):
Just a few thoughts, otherwise lgtm!

packages/ui/react-ui-table/src/components/Cell/Cell.tsx
richburdon: no way to set state so that when the new row is rendered it's autofocuesd?

packages/ui/react-ui-table/src/components/Cell/Cell.tsx
```diff
@@ -18,11 +18,45 @@ const CELL_NAME = 'Cell';
 const FocusableCell = <TData extends RowData, TValue>({ cell, children }: CellProps<TData, TValue>) => {
   const tableContext = useTableContext();
   const domAttributes = useFocusableGroup({ tabBehavior: 'limited' });
-
   const className = tdRoot(tableContext, cell.column.columnDef.meta?.cell?.classNames);
 
+  const handleKeyDown = useCallback((event: React.KeyboardEvent<HTMLTableCellElement>) => {
```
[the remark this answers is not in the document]
thure: I thought rule-of-thumb was not to abbreviate? I recall Josiah encouraging me to use `event`.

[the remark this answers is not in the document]
ZaymonFC: Either one doesn't impact readability. As it's a short name I'd prefer `event` personally.

PR #7568, opened 2024-08-27
feat: script editor

wittjosiah:
Revives script plugin in Composer to be used for deploying user defined functions. Deprecates existing script block component.

image


packages/apps/plugins/plugin-script/src/components/ScriptEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [error, setError] = useState<string>();
+
+  const handleBindingChange = useCallback(
+    (binding: string) => {
+      script.binding = binding;
+    },
+    [script],
+  );
+
+  const handleDeploy = useCallback(async () => {
```
richburdon, line 66: if this gets more complicated consider making a local file hook: useDeploy, etc?

packages/apps/plugins/plugin-script/src/components/Toolbar.tsx
```diff
@@ -0,0 +1,163 @@
...
+  Tooltip,
+  useTranslation,
+} from '@dxos/react-ui';
+import { getSize, mx } from '@dxos/react-ui-theme';
+
+import { SCRIPT_PLUGIN } from '../meta';
+
+export type ToolbarProps = {
+  binding: string;
+  onBindingChange: (binding: string) => void;
```
richburdon, line 23: prefer grouping callbacks at the end

packages/apps/plugins/plugin-script/src/components/TypescriptEditor.tsx
```diff
@@ -0,0 +1,68 @@
...
+        autocomplete(),
+      ].filter(nonNullable),
+    [_extensions, themeMode],
+  );
+
+  const { parentRef, focusAttributes } = useTextEditor(
+    () => ({
+      id,
+      initialValue,
+      extensions,
```
richburdon, line 58: create extensions inline here rather than additional useMemo above.

richburdon (approved):

Issue #7571, opened 2024-08-27
[map] plugin design


seagreen:
## Goals

1. Quickly make the map plugin useful to people
2. Do (1) in a way unlikely to require breaking changes later (thus we should have a long term idea for where the plugin is headed)

### Option 1: Query the parent Collection for tables with `latitude` and `longitude` columns

(or something similar like a single `location` column)

Then use those rows to populate the map. The result would look like this (not actually implemented yet):

![image](https://github.com/user-attachments/assets/50cd1f2a-4b5f-49a1-a0ac-30e010c69d75)

I think this is the favored plan, judging by this comment:

https://github.com/dxos/dxos/blob/6b2704bedef181e8d07d818f6b893c7f20711e2a/packages/apps/plugins/plugin-map/src/components/MapSection.tsx#L12

One question this brings up-- what data do we display if the map is viewed as a single object, not as part of the collection? Querying ECHO for all the collections it's part of might bring up multiple collections, which is awkward.

### Option 2: Explicitly link to a table

Have an "edit" mode for the app with some input fields, such as a link to the table to use, as well as what fields to find latitude and longitude in.

Straightforward, but more work for the user that Option 1.

Also, even with Option 1 above we might want to have an option to name what fields latitude and longitude should be found in.

## Question: should the plugin support saving viewport location and zoom level?

This might be useful. One alternative is to automatically put the viewport location in the middle of the displayed points.

richburdon:
i really like the automatic search for tables in the stack. that would be a great default.
after that we might have some metadata editor for the map (@thure and i discussed using the aux sidebar for this) to specifically create an ECHO query.

seagreen:
Sounds great, let's go with automatic search.

I like how that leads to a future where map creation defaults to as easy as possible, but then (if we write the query builder) is also fully controllable.

The only wrinkle is what to do when the map is viewed as a single object, without an enclosing collection. That doesn't seem too important though. Most of the time it will have one parent collection. We can query for that, autodetect the relevant rows, and display them.

If it actually has 2+ parent collections than we can either display a notice that it's an unusual situation, or we can auto-detect rows from both collections and combine them. Or both.

seagreen:
> If it actually has 2+ parent collections than we can either display a notice that it's an unusual situation, or we can auto-detect rows from both collections and combine them. Or both.

I didn't realize being in multiple collections at once is disabled in the UI right now. That makes this not very important, we can revisit it if we re-enable having multiple collection parents.

[...]

PR #7583, opened 2024-08-28
feat(map): add neighboring rows with latitude/longitude as pins

seagreen:
Working off of: https://github.com/dxos/dxos/issues/7571

Add a new hook to detect the rows we're interested in, then uses it to display them in Article and Section views.

Hook details:
```
/**
 * Find table rows that are in the same collection as the map.
 * Return the ones with `latitude` and `longitude` fields,
 * converted to `Marker`s.
 *
 * NOTE: This collects rows from all the collections the map is in,
 * even if there's more than one. At the time of writing this behavior
 * won't be exercised since being included in multiple collections
 * is disabled in the UI.
 *
 * If we enable being in more than one collection again we should
 * reconsider what behavior we actually want in that case.
 */
export const useMapDetectLocations = (map: MapType): Marker[] => {
```


packages/apps/plugins/plugin-map/src/components/MapControl.tsx
```diff
@@ -21,7 +21,7 @@ export type Marker = {
 
 const defaults = {
   center: { lat: 37.970833, lng: 23.72611 } as LatLngExpression,
-  zoom: 13,
+  zoom: 1,
```
seagreen, line 24: Zoom way out to guarantee including the pins. A future improvement would be to set zoom and pan to center the pins.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  const siblingTables = useQuery<TableType>(
+    space,
+    [Filter.schema(TableType), (table: any) => table.id in mapSiblingIds],
+    undefined,
+    [parentCollections],
+  );
+
+  // TODO: filter these for only schemas with latitude and longitude fields.
```
seagreen, line 63: Still got this to do...

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * If we enable being in more than one collection again we should
+ * reconsider what behavior we actually want in that case.
+ */
+export const useMapDetectLocations = (map: MapType): Marker[] => {
+  const space = getSpace(map);
+  if (!space) {
+    // TODO: not sure if throwing error here is right.
+    // Am I missing a space getter function that always succeeds?
+    throw new Error('MapSection Space is undefined');
```
seagreen, line 28: Is `getSpace` always guaranteed to succeed? Would appreciate suggestions on what to do here.

ZaymonFC: The pattern I prefer is just to return undefined. Then the UI can handle that gracefully. Missing data should never result in the user seeing the error boundary component.

Letting things reactively flow like this has benefits. For instance, if you're querying `map`, it will probably get the initial value `undefined` which will trigger this until it resolves, causing a re-flow of reactivity.

ZaymonFC: In this case, you could also return empty list until space has a value.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,79 @@
...
+      }
+    });
+  });
+
+  const siblingTables = useQuery<TableType>(
+    space,
+    [Filter.schema(TableType), (table: any) => table.id in mapSiblingIds],
+    undefined,
+    [parentCollections],
+  );
```
ZaymonFC, line 61:
```suggestion
  const siblingIds = new Set(
    parentCollections.flatMap(collection => 
      collection.objects
        .filter(obj => obj?.id)
        .map(obj => obj.id)
    )
  );

  const siblingTables = useQuery<TableType>(
    space,
    [Filter.schema(TableType), (table: any) => siblingIds.has(table.id)],
    undefined,
    [parentCollections],
  );
```

I bit of a micro-optimization, but this is cleaner with a set imo.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
ZaymonFC: Actually, space can be undefined in `useQuery` for this exact reason. So you can just pass it and let that empty query result flow.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
seagreen: Great! I've gotten into a habit of using `Record<string, null>` from situations where I needed a `toJSON` method, but `Set` is certainly meant for this and I'll switch to it in the future.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
seagreen: Ah, I'm catching on now. Is it best practice to have a no-op option for all custom hooks?

ZaymonFC: You can think of these things as compositional parts of reactive machines which will re-trigger when params and dependencies change. So yeah you often want to have a path for undefined / 'data not ready yet'. When the upstream thing gets data, it should flow through the reactivity graph and trigger updates along the way.

seagreen: Got it, that totally makes sense.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const schemaIds: string[] = siblingTables.flatMap((table) => (table.schema ? [table.schema.id] : []));
+
+  const rows = useQuery(
+    // If there are no schema IDs that we're interesting in return nothing.
+    // (this is necessary since the or-based filter will always return true in this case):
+    schemaIds.length === 0 ? undefined : space,
+    Filter.or(...schemaIds.map((id) => Filter.typename(id))),
```
seagreen, line 68: Does filter have an idiomatic "reject all" option? If so I'll replace the above with:
```
  const rows = useQuery(
    space,
    schemas.length === 0 ? Filter.rejectAll : Filter.or(...schemaIds.map((id) => Filter.typename(id))),
```

seagreen: (Not looking for a literal `rejectAll`, just something with the same effect that's performant)

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
seagreen: Done in this commit: https://github.com/dxos/dxos/pull/7583/commits/1f29911a413259f4cb5264574c17ea9d97c0bd64

seagreen:
I've busted something with how I'm using `useQuery`, looking into it now. This is the error:
```
Warning: The final argument passed to useMemo changed size between renders. The order and size of this array must remain constant.

Previous: [, options, , type, , properties, , text, , predicate, , not, false, and, or, ]
Incoming: [[object Object], options, , type, , properties, , text, , predicate, , not, false, and, or, , , [object Object],[object Object]]
    at MapSection (http://localhost:5173/@fs/Users/ian/code/dxos/packages/apps/plugins/plugin-map/src/components/MapSection.tsx?t=1724944654037:19:23)
```

seagreen:
I think the above may be due to me changing the number of filters passed to `useQuery` in my code (since I'm turning each detected schema into an `or`), combined without how `useQuery` works:
https://github.com/dxos/dxos/blob/e1cc207280cb469215711a83034c6a06ac13afe6/packages/sdk/react-client/src/echo/useQuery.ts#L44

There's also the issue of the arrays that I'm passing as deps too like here:
https://github.com/dxos/dxos/pull/7583/files#diff-517ad06639b0e181c204301747b34f39593ab62bbc62b8a47058f2ed77e4040eR70
...so right now there are two causes of variable length deps in the `useMemo` inside of `useQuery`.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,97 @@
...
+      (table: any) => siblingIds.has(table.id),
+      (table: any) => {
+        if (!(table instanceof TableType)) {
+          return false;
+        }
+        return table.schema ? schemaHasLatitudeAndLongitude(table.schema) : false;
+      },
+    ],
+    undefined,
+    [Array.from(siblingIds).sort().join('_')],
```
seagreen, line 59: If there's a more idiomatic way to deterministically stringify here, let me know

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
```diff
@@ -52,11 +52,9 @@ export const MapControl: FC<{ zoom?: number; center?: LatLngExpression; markers?
 
       {/* Markers. */}
       {/* TODO(burdon): Marker icon doesn't load on mobile. */}
-      {markers
-        .filter(({ location }) => location?.lat !== undefined && location?.lng !== undefined)
```
seagreen, line 56: Removed this filter since `lat` and `lng` should always be present.

Context from `@types/leaflet`:
```
export interface LatLngLiteral {
    lat: number;
    lng: number;
}
```

seagreen:
Added markup popups showing titles, screenshot:
image

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
ZaymonFC: https://github.com/PaulLeCam/react-leaflet/issues/190 Looks like it might be somewhat trivial to achieve this?

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
ZaymonFC: I would call `JSON.stringify(Array.from(siblingsIds))` personally.

seagreen: Done! Doubly good you mentioned this because I was using `siblingIds` down below where I should have been using `schemaIds`. Now fixed.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,98 @@
...
+    [JSON.stringify(Array.from(siblingIds))],
+  );
+
+  const schemaIds: string[] = siblingTables.flatMap((table) => (table.schema ? [table.schema.id] : []));
+
+  const rows = useQuery(
+    // Short circuit if there are no `schemaIds` to find rows for:
+    schemaIds.length === 0 ? undefined : space,
+    // PERFORMANCE: checks every object.
+    // Can't pass `Filter.or`s to `useQuery` because we need to a dynamic number of them.
```
wittjosiah, line 68:
```suggestion
    // TODO(seagreen): This checks every object.
    //   Can't pass list of `Filter.or`s to `useQuery` because we need to a dynamic number of them.
```

wittjosiah: @dmaretskyi we need to consider how to combine these filters in a way where there's just one root so that the react deps array size doesn't change

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
wittjosiah: this seems like it would make this much more useful

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,98 @@
...
+      Filter.schema(CollectionType),
+      (collection: any) => {
+        if (!(collection instanceof CollectionType)) {
+          return false;
+        }
+        return collection.objects.some((obj) => obj?.id === map.id);
+      },
+    ],
+    undefined,
+    [space],
```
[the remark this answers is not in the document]
seagreen: Nice!

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
seagreen: Agreed, I'll take care of it. I didn't want to put much polish into the PR until I knew it was overall on the right track.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,98 @@
...
+ * reconsider what behavior we actually want in that case.
+ */
+export const useMapDetectLocations = (map: MapType): Marker[] => {
+  const space = getSpace(map);
+
+  const parentCollections = useQuery<CollectionType>(
+    space,
+    [
+      Filter.schema(CollectionType),
+      (collection: any) => {
```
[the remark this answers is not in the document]
seagreen: Whoa, much nicer!

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
seagreen: Have something for it here: https://github.com/dxos/dxos/pull/7583/commits/e24455c8763b80ee5c7196443eb77ef13c64b21b

seagreen: Thank you for the `latLngBounds` tip!

packages/apps/plugins/plugin-map/src/components/MapControl.tsx
```diff
@@ -21,29 +22,30 @@ export type Marker = {
...
-  zoom: 13,
+  zoom: 1,
 };
 
 /**
  * https://react-leaflet.js.org/docs/api-map
  */
-export const MapControl: FC<{ zoom?: number; center?: LatLngExpression; markers?: Marker[] }> = ({
-  zoom = defaults.zoom,
-  center = defaults.center,
```
seagreen, line 32: Remove unused API options from `MapControl`, this way we don't have to deal with cases like whether provided `zoom` and `center` arguments or zooming in on the markers takes precedence. We can always add them back later if desired.

packages/apps/plugins/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -0,0 +1,88 @@
...
+    undefined,
+    [JSON.stringify(Array.from(siblingIds))],
+  );
+
+  const schemaIds: string[] = siblingTables.flatMap((table) => (table.schema ? [table.schema.id] : []));
+
+  const rows = useQuery(
+    // Short circuit if there are no `schemaIds` to find rows for:
+    schemaIds.length === 0 ? undefined : space,
+    // TODO(seagreen): performance issue since this checks every object.
```
seagreen, line 51: Rewrote this comment, hopefully it makes the situation more clear now.

packages/plugins/experimental/plugin-map/src/components/MapControl.tsx
```diff
@@ -21,29 +22,33 @@ export type Marker = {
...
+    if (markers.length > 0) {
+      const bounds = latLngBounds(markers.map((marker) => marker.location));
+      map.fitBounds(bounds);
+    } else {
+      map.setView(defaults.center, defaults.zoom);
     }
-  }, [zoom, center]);
+    // Using plain `[markers]` here causes the effect to trigger extraneously,
+    // overwriting the user's zoom when it shouldn't:
+  }, [markers.length]);
```
seagreen, line 51: We could also serialize the markers to JSON here, which would have the advantage of picking up when markers lat/long is changed, not just when they're added/removed.

packages/plugins/experimental/plugin-map/package.json
```diff
@@ -52,6 +52,7 @@
     "@dxos/plugin-graph": "workspace:*",
     "@dxos/plugin-space": "workspace:*",
     "@dxos/plugin-stack": "workspace:*",
+    "@dxos/plugin-table": "workspace:*",
```
seagreen, line 55: When I merged in `main` it broke the `TableType` import, added this as a dep to get it back.

packages/plugins/experimental/plugin-map/src/components/MapControl.tsx
ZaymonFC: I would do this.

ZaymonFC: If markers is changing too often, that means there's a bug higher up the graph. Why is markers not stable for a constant set of values is the better question?

packages/plugins/experimental/plugin-map/package.json
thure: cc @richburdon

thure (approved):
lgtm!

ZaymonFC (approved):
Looking good! Nice work.

packages/plugins/experimental/plugin-map/src/components/MapControl.tsx
seagreen: `useMemo` to the rescue

PR #7584, opened 2024-08-28
fix(plugin-navtree): Restore keyboard access to tree item tooltips and navigation

### LINKED ISSUE #7421: [navtree] keyboard navigation
- [x] introduce arrow navigation to navtree so that it can be much more quickly traversed from the keyboard without stepping through all menus
- [x] up and down arrows should traverse tree items in order
- [x] left and right arrows?
  - [x] expand/collapse on branch items?

thure:
This PR:
- resolves #7421.
- resolves an issue where overflowing tree item labels were no longer available by cursor or keyboard.


packages/ui/react-ui-navtree/src/components/NavTreeItem.tsx
```diff
@@ -158,6 +164,7 @@ const NavTreeItemImpl = forwardRef<HTMLDivElement, MosaicTileComponentProps<NavT
             setMenuOpen(true);
           }}
           {...draggableProps}
+          aria-describedby=''
```
wittjosiah, line 167: what's the significance of this?

wittjosiah (approved):

packages/ui/react-ui-navtree/src/components/NavTreeItem.tsx
thure: I wanted to leave a comment about this but couldn’t find a way that our estlint rules would allow… basically `draggableProps` includes an aria description with instructions on how to drag the item using the keyboard, but that is not supported here for now.

I think I’d notice this when returning to fix the DnD issues here, but if you have any recommendations on how to annotate this I’d be happy to leave an annotation.

thure: Figured it out

PR #7597, opened 2024-08-29
feat: ECHO API in workerd

dmaretskyi:
- [ ] Effect schema for queries
- [ ] Unify types for proxies and snapshots
- [ ] Mongo API for updates and queries


packages/core/echo/echo-db/src/core-db/core-database.ts
```diff
@@ -323,7 +324,50 @@ export class CoreDatabase {
...
   private _query(filter?: FilterSource, options?: QueryOptions) {
-    return new Query(new CoreDatabaseQueryContext(this, this._queryService), Filter.from(filter));
+    return new Query(
+      new CoreDatabaseQueryContext(this, this._queryService),
+      Filter.from(filter, { ...options, spaceIds: [this.spaceId] }),
+    );
+  }
+
+  // TODO(dmaretskyi): Mongo syntax.
+  async update(data: { id: string } & { [key: string]: any }) {
```
yarolegovich, line 334: what about deep updates `nested.object.field`?

packages/core/echo/echo-db/src/core-db/core-database-query-context.ts
```diff
@@ -80,26 +90,69 @@ export class CoreDatabaseQueryContext implements QueryContext {
...
       return null;
     }
 
-    const core = await this._coreDatabase.loadObjectCoreById(result.id);
-    if (!core || ctx.disposed) {
-      return null;
+    /**
+     * Ignore data in the query result and fetch documents through DataService & RepoProxy.
+     */
+    const FORCE_DATA_SERVICE_FETCH = true;
```
yarolegovich, line 101: not sure I understand how this is going to be used

yarolegovich (approved):

packages/core/echo/echo-db/src/core-db/core-database.ts
dmaretskyi: TBD. I wanna roll out this as a POC for people to experiment. Let's discuss API ideas in the afternoon today. I'm thinking we will just use the mongo syntax.

packages/core/echo/echo-db/src/core-db/core-database-query-context.ts
dmaretskyi: I'm thinking about ways to implement this where it doesn't require two separate RPCs: query to get the ideas, and then a separate document fetch. This flag is just to avoid deleting or commenting out the other branch of the code that experiments with doing a single query RPC that holds the data too. It doesn't full work because it doesn't seed the CRDT data, which means that you won't be able to mutate the queried objects later.

PR #7600, opened 2024-08-30
docs: Fix getting-started worker plugin config

wilkerlucio:
I noticed while going over the plugin's definition for worker that it is missing a function wrap.


ZaymonFC:
https://github.com/dxos/dxos/blob/e5cd96a0a544e06eb9da5109eb730b7c93cef277/packages/apps/composer-app/vite.config.ts#L81-L147

We don't seem to need this in composer app. Is this necessary?

wilkerlucio:
When I tried without it I got complaints at it didn't worked (some linter error, not have at hand now but if you want I can repro and get it back here later).

ZaymonFC:
> When I tried without it I got complaints at it didn't worked (some linter error, not have at hand now but if you want I can repro and get it back here later).

Oh, I see that in composer we also use a closure for the worker. Thanks for this!

ZaymonFC (approved):
✨ Thanks for the contribution!

wilkerlucio:
Thanks, and just for ref for future readers, this is the code at composer doing the same:

https://github.com/dxos/dxos/blob/08ac8a9c4592a01c47232c1ba8b4b3cc4a6203d9/packages/apps/composer-app/vite.config.ts#L79

PR #7602, opened 2024-08-30
refactor: move types to plugins

### LINKED ISSUE #4854: [composer] remove @braneframe namespace
## Motivation

The `@braneframe` namespace is increasing complexity and maintenance cost to work around multiple namespaces.

Specifically, the longer we use this namespace, the more users out there are creating entities with type strings that contain these words, the more migration complexity we will have later.

It's also hard to write docs and ship the app framework when all the plugins are in this weird namespace

e.g.: `@braneframe/plugin-markdown => @dxos/app-plugin-markdown`

wittjosiah:
Resolves #4854 

- moves echo schema definitions to plugins
- drops usage of braneframe namespace
- moves plugins to top-level and separates experimental plugins
- removes composer plugin npm create template

Notes:

- temporarily moved helpers from `@braneframe/types` into plugin metadata (loadReferences + serializer)
- thread types create circular dependencies so they've landed in space plugin for now


dmaretskyi (changes requested):
update toolbox.json

dmaretskyi (approved):

PR #7611, opened 2024-08-31
fix: Editor extensions: folding, numbered headings

richburdon:
- [x] Folding option
- [x] Numbered heading option
- [x] Misc
  - [x] NOTE: Remove default component props that are used as deps!
  - [x] Trim borders
  - [x] Fix chess layout
  - [x] Fix explorer plugin dark mode styles

## Editor options

https://github.com/user-attachments/assets/5e10439e-0901-4dd2-9583-b34d17ce9038

## Borders

Screenshot 2024-08-31 at 1 40 22 AM


packages/plugins/experimental/plugin-excalidraw/src/hooks/useStoreAdapter.ts
```diff
@@ -28,7 +28,7 @@ export const useStoreAdapter = (
     }
 
     const t = setTimeout(async () => {
-      await adapter.open(createDocAccessor(object, ['content']));
+      await adapter.open(createDocAccessor(object, ['canvas', 'content']));
```
wittjosiah, line 31:
```suggestion
      await adapter.open(createDocAccessor(object.canvas, ['content']));
```
is there a difference between these two? if the canvas isn't loaded will creating a deeply nested doc accessor load the object? shouldn't merge this without the corresponding migration?

packages/plugins/plugin-chess/src/components/index.ts
```diff
@@ -5,5 +5,4 @@
 import React from 'react';
 
 // Lazily load components for content surfaces.
-export const ChessMain = React.lazy(() => import('./ChessMain'));
-export const ChessArticle = React.lazy(() => import('./ChessArticle'));
+export const ChessWrapper = React.lazy(() => import('./ChessWrapper'));
```
wittjosiah, line 8: ChessContainer? if aligning with other plugins

wittjosiah (commented):
love to see folding making it's way into the editor (looking forward to it in lists next 😉). Not sure if you ran into a technical limitation here or not, but I think it would be more useable if the collapse button was right next to the content instead of over off to the side

richburdon:
> love to see folding making it's way into the editor (looking forward to it in lists next 😉). Not sure if you ran into a technical limitation here or not, but I think it would be more useable if the collapse button was right next to the content instead of over off to the side

:). i'm not sure how to do that given the content margins. let's discuss with the team.

packages/plugins/experimental/plugin-excalidraw/src/hooks/useStoreAdapter.ts
richburdon: ok, i cleaned this up before this commit -- and i think i had confused myself. no, this is correct. and a migration won't be needed. but it doesn't work in the storybook.

packages/plugins/plugin-chess/src/components/index.ts
richburdon: ah yes, that was the intention. thanks.

PR #7618, opened 2024-09-03
fix: Prevent `scrollIntoView` from scrolling elements that should not be scrolled

thure:
This PR uses overflow’s `clip` property instead of `hidden` to prevent `scrollIntoView` calls from scrolling elements which should not be scrollable.


ZaymonFC (approved):
**Ship it** 👾

PR #7629, opened 2024-09-04
fix: Scroll into view and numbered headings

richburdon:
- [x] Fix scroll into view.
- [x] Fix numbered headings.
- [x] Remove most of the shaky custom styles.

NOTES:
- the markdown plugin listens for scroll into view (select?) events and uses a CM effect to scroll vertically (this doesn't involve the DOM's scroll into view method).
- the deck layout listens for the event and scrolls the article into view.
- this seems to incur a race condition when returning from solo mode.

@ZaymonFC previously switching to/from solo seemed to be seamless (no noticeable re-render -- did that change?)


packages/plugins/plugin-markdown/src/components/DocumentEditor.tsx
```diff
@@ -120,12 +118,11 @@ const DocumentEditor = ({
...
   return (
     <MarkdownEditor
-      id={fullyQualifiedId(doc)}
+      id={id}
       initialValue={initialValue}
       extensions={extensions}
       scrollTo={scrollTo}
-      // TODO(wittjosiah): Ensure selection is within the document.
-      // selection={selection}
+      selection={selection}
```
wittjosiah, line 125: is there a fix for this somewhere? this broke a bunch of my documents because the stored selection was larger than the document length and so I couldn't open them anymore

packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
```diff
@@ -96,19 +96,18 @@ export const MarkdownEditor = ({
     onCommentSelect?.(id);
   });
 
-  // Focus comment.
+  // Focus the space that references the comment.
   useIntentResolver(MARKDOWN_PLUGIN, ({ action, data }) => {
     switch (action) {
+      // TODO(burdon): Use fully qualified ids everywhere.
```
wittjosiah, line 102: I know you just found a couple places where it wasn't being used but generally this is already the case and occasionally we find stragglers that we've missed causing issues.

packages/plugins/plugin-markdown/src/components/DocumentEditor.tsx
wittjosiah: okay I think I see a fix to only allow selection within the document below in `useTextEditor` below 👍

packages/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -432,16 +426,14 @@ export const DeckPlugin = ({
                   return;
                 }
 
-                const newPlankPositioning = settings.values.newPlankPositioning;
-
                 const processLayoutEntry = (partName: string, entryString: string, currentLayout: any) => {
+                  // TODO(burdon): Option to toggle?
+                  const toggle = false;
```
ZaymonFC, line 431: Did you see that you can pass `noToggle` in the intent data?

packages/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -578,16 +568,15 @@ export const DeckPlugin = ({
...
+                    location.values.active = incrementPlank(location.values.active, {
                       type: adjustment.type,
                       layoutCoordinate: adjustment.layoutCoordinate,
                     });
-                    location.values.active = nextActive;
                   }
 
                   if (adjustment.type === 'solo') {
+                    // TODO(burdon): fullyQualifiedId?
                     const entryId = adjustment.layoutCoordinate.entryId;
```
ZaymonFC, line 580: This should be fully-qualified at the dispatch site.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -81,9 +83,32 @@ export const DeckLayout = ({
...
+  const deckRef = useRef<HTMLDivElement>(null);
+  useLayoutEffect(() => {
+    if (layoutMode === 'deck' && activeId) {
+      const el = deckRef.current?.querySelector(`article[data-attendable-id="${activeId}"]`);
+      // TODO(burdon): Generalize scroll into view by parent.
+      setTimeout(() => {
+        el?.scrollIntoView({ behavior: 'smooth', inline: 'center' });
+      }, 100); // TODO(burdon): Hack. The element already exists but scrolling doesn't happen without the delay. Calculate or memo offset?
+    }
+  }, [layoutMode, activeId]);
```
ZaymonFC, line 96: A few things here:

It feels funky having a default value from attention but then getting subsequent values from the scroll into view event. What's the intention with always supplying a value?

Since we're in deck land here, you can watch the `scrollIntoView` property from the layout context instead. It's in scope here, just not destructured.

Make sure to return a clean up function that clears the timeout. It's a shame we need to wait so long.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -179,18 +227,16 @@ export const DeckLayout = ({
...
-            </div>
+            </Deck.Root>
           </Main.Content>
         )}
 
         {/* Solo main content surface. */}
-        {layoutMode === 'solo' && layoutParts.solo && layoutParts.solo.length > 0 && (
+        {/* Main content surface. */}
+        {/*
+        {layoutMode === '__solo' && layoutParts.solo && layoutParts.solo.length > 0 && (
```
ZaymonFC, line 237: If this is unused now please delete.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -199,20 +245,44 @@ export const DeckLayout = ({
...
+                      flatDeck={flatDeck}
+                      searchEnabled={searchEnabled}
+                    />
+                  );
+                })}
+              </Deck.Root>
+            </div>
           </Main.Content>
         )}
+        */}
```
ZaymonFC, line 285: Same here.

ZaymonFC (commented):
A few questions / suggestions

packages/plugins/plugin-markdown/src/components/DocumentEditor.tsx
richburdon: yes and i added a specific story also.

packages/plugins/plugin-deck/src/DeckPlugin.tsx
richburdon: i didn't. thanks. let's discuss how to ux this.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
richburdon: let's discuss how to do this properly.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
richburdon: all of this is pending our discussion.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
richburdon: can you take this. i think i have things working again, but i would be good to remove the need for this timeout altogether.

packages/plugins/experimental/plugin-debug/src/components/Wireframe.tsx
```diff
@@ -13,7 +13,7 @@ export const Wireframe = ({ label, data, className }: { label?: string; data?: a
   const attendableAttrs = createAttendableAttributes(data);
   const { width, height, ref } = useResizeDetector();
   return (
-    <div ref={ref} className={mx('relative grow', className)} {...attendableAttrs}>
+    <div ref={ref} className={mx('relative grow min-h-[400px]', className)} {...attendableAttrs}>
```
thure, line 16:
```suggestion
    <div ref={ref} className={mx('relative grow min-bs-96', className)} {...attendableAttrs}>
```

Unless it must be exactly `400px` in which case `min-bs-[400px]`.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -81,21 +83,46 @@ export const DeckLayout = ({
     }
   }, [layoutParts]);
 
-  const searchEnabled = !!usePlugin('dxos.org/plugin/search');
-  const activeId = useMemo(() => Array.from(attention.attended ?? [])[0], [attention.attended]);
+  const [activeId, setActiveId] = useState<string | undefined>(Array.from(attention.attended ?? [])[0]);
+  const deckRef = useRef<HTMLDivElement>(null);
+  useLayoutEffect(() => {
+    let t: NodeJS.Timeout | undefined;
```
thure, line 89: If timeouts are just numbers, maybe just `number` since this is a browser runtime? Unclear by its name whether it’s portable.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -81,21 +83,46 @@ export const DeckLayout = ({
...
+    if (layoutMode === 'deck' && activeId) {
+      const el = deckRef.current?.querySelector(`article[data-attendable-id="${activeId}"]`);
+      // TODO(burdon): Generalize scroll into view by parent.
+      t = setTimeout(() => {
+        el?.scrollIntoView({ behavior: 'smooth', inline: 'center' });
+      }, 100); // TODO(burdon): Hack. The element already exists but scrolling doesn't happen without the delay. Calculate or memo offset?
+    }
+    return () => clearTimeout(t);
+  }, [layoutMode, activeId]);
```
thure, line 98: Maybe JSDoc this effect or move it into a separate hook file to declare/describe what it’s for.

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -152,21 +180,43 @@ export const DeckLayout = ({
...
+        {isEmpty && (
+          <Main.Content>
+            <ContentEmpty />
+          </Main.Content>
+        )}
+
+        {/* TODO(burdon): Preserve DOM structure between solo/deck mode. */}
+        {!isEmpty && (
           <Main.Content bounce classNames={['grid', 'block-end-[--statusbar-size]']}>
-            <div role='none' className='relative'>
```
thure, line 158: I tried removing this node in my PR and it complicated things — does Deck still work (scrolls horizontally, responsive layout is correct) when this is removed?

packages/ui/react-ui-deck/src/components/Deck/Deck.tsx
```diff
@@ -123,25 +123,21 @@ const DeckPlankRoot = ({
...
     const articleElement = useRef<HTMLDivElement | null>(null);
     const ref = useComposedRefs(articleElement, forwardedRef);
     const { unit = 'rem', size = PLANK_DEFAULTS.size } = usePlankContext('DeckPlankContent');
     const { solo } = useDeckContext('DeckPlankContext');
     const inlineSize = solo ? undefined : isSm ? `${size}${unit}` : '100dvw';
 
-    // TODO(burdon): Experiment to fade in. Works well, except when toggling between solo mode.
     // Opacity transition prevents flicker when content is first rendered.
-    // const [visible, setVisible] = useState('opacity-0 transition duration-1000');
+    const [visible, setVisible] = useState<string>();
```
thure, line 140: `visible` is a different property that isn’t transitionable, consider `opaque` or `transparent`.

thure (approved):
I suppose I’m mostly concerned with whether Deck layout works without the interceding div with `relative`, otherwise this seems fine, just a few things to consider.

packages/plugins/plugin-deck/src/util/overscroll.ts
```diff
@@ -67,6 +67,7 @@ export const calculateOverscroll = (
    * └────────────────────────────────────────────────────────────────────────────────────────────────────┘
    */
 
+  // TODO(burdon): Factor out constants.
```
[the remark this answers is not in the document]
ZaymonFC: 👍

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
richburdon: it seems to work -- my goal was to make the deck and solo keep as much similarity in dom as possible.

packages/ui/react-ui-deck/src/components/Deck/Deck.tsx
richburdon: i am using opacity!?

PR #7633, opened 2024-09-04
feat(plugin-sheet): dynamic bindings

wittjosiah:
TODO: because this introduces echo deps into the sheet model it's not working in vitest

https://github.com/user-attachments/assets/741e9f28-d695-4c0c-a12a-cb6dd7134a47


packages/plugins/experimental/plugin-script/src/components/ScriptEditor.tsx
```diff
@@ -79,26 +83,30 @@ export const ScriptEditor = ({ script, role }: ScriptEditorProps) => {
...
       script.changed = false;
     } catch (err: any) {
       log.catch(err);
       setError(t('upload failed label'));
     }
-  }, [existingFunctionUrl, script, script.name, script.source]);
+  }, [fn, existingFunctionUrl, script, script.name, script.source]);
 
   return (
     <div role='none' className={mx(role === 'article' && 'row-span-2', 'is-full pli-2')}>
```
richburdon, line 107: i think this far into ./components we should not be checking roles or dealing with spans. would love to start nailing this down.

packages/plugins/experimental/plugin-script/src/components/Toolbar.tsx
```diff
@@ -27,20 +27,34 @@ export type ToolbarProps = {
...
+  functionUrl,
+  error,
+}: ToolbarProps) => {
   const { t } = useTranslation(SCRIPT_PLUGIN);
   const [pending, setPending] = useState(false);
   const [editing, setEditing] = useState(false);
+  const [binding, setBinding] = useState(_binding);
   const inputRef = useRef<HTMLInputElement>(null);
 
+  useEffect(() => {
```
richburdon, line 44: see useDefaultValue

richburdon (approved):

packages/plugins/experimental/plugin-script/src/components/ScriptEditor.tsx
wittjosiah: this is the top level script component

packages/plugins/experimental/plugin-script/src/components/Toolbar.tsx
wittjosiah: forgot about that one, will update

wittjosiah: actually not sure it handles this case because:
1. there's not a stable default value
2. want to make sure it only swaps it when not in editing mode

PR #7635, opened 2024-09-04
fix(comments): Restore click to view thread

ZaymonFC:
- Resolves: https://github.com/dxos/dxos/issues/7632


packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
```diff
@@ -92,9 +94,10 @@ export const MarkdownEditor = ({
 
   // TODO(Zan): Move these into thread plugin as well?
   const [commentsState, commentObserver] = useCommentState();
-  const commentClickObserver = useCommentClickListener((id) => {
-    onCommentSelect?.(id);
```
wittjosiah, line 96: is this prop now unused?

ZaymonFC: Removed.

PR #7637, opened 2024-09-04
fix(comments): Bidirectional thread highlighting

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/7636

https://github.com/user-attachments/assets/ef97bf96-529f-4d1a-8887-8916387ca6d5

### DIFF
--- a/packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
+++ b/packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
@@ -35,6 +35,8 @@ import {
   editorContent,
   editorGutter,
   Cursor,
+  selectThread,
+  setSelection,
 } from '@dxos/react-ui-editor';
 import { sectionToolbarLayout } from '@dxos/react-ui-stack';
 import { textBlockWidth, focusRing, mx } from '@dxos/react-ui-theme';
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
```diff
@@ -105,8 +106,23 @@ export const MarkdownEditor = ({
...
-            editorView.dispatch({ effects: EditorView.scrollIntoView(range.from, { y: 'start', yMargin: 96 }) });
+            const currentSelection = editorView.state.selection.main;
+            const needsSelectionUpdate = currentSelection.from !== range.from || currentSelection.to !== range.from;
+
+            const effects = [
+              // NOTE: This does not use the DOM scrollIntoView function.
+              EditorView.scrollIntoView(range.from, { y: 'start', yMargin: 96 }),
+            ];
+
+            // NOTE: Update the editor selection to get bi-directional highlighting.
```
richburdon, line 117: i'm using "NOTE" above for emphasis, not required here?

packages/plugins/plugin-markdown/src/components/MarkdownEditor.tsx
```diff
@@ -105,8 +106,23 @@ export const MarkdownEditor = ({
           // TODO(burdon): We need typed intents.
           const range = Cursor.getRangeFromCursor(editorView.state, data.cursor);
           if (range?.from) {
-            // NOTE: This does not use the DOM scrollIntoView function.
-            editorView.dispatch({ effects: EditorView.scrollIntoView(range.from, { y: 'start', yMargin: 96 }) });
+            const currentSelection = editorView.state.selection.main;
```
richburdon, line 109: reduce the number of interim vars?

```
const selection = editorView.state.selection.main.from !== range.from ? { anchor: range.from } : undefined;
effects = [...];
if (selection) {
  effects.push(...);
}
```

PR #7642, opened 2024-09-04
chore: don't unmount planks

richburdon:

PR #7643, opened 2024-09-05
feat: Landmark keyboard access

### LINKED ISSUE #7424: [deck] keyboard navigation
- when plank is focused should be able to left/right arrow between planks
- when section is focused should be able to up/down focus between sections
- figure out how to handle stack planks in an intuitive way
  - potentially if focused on a section left/right skips plank and goes right to sections if the plank is a stack
- `Enter`/`Esc` should focus content of a plank/blur to plank from content
- add key binding for solo/un-solo of planks
- stop scrolling from taking precedence over left/right arrows when navigating via keyboard.
- #6777 

Example interaction demo https://stack-matrix.vercel.app/

### LINKED ISSUE #7578: [sketch] grid disappears after plank resize
After resizing a sketch plank grid and controls disappear and reappear only after I click on the diagram. 

https://github.com/user-attachments/assets/d66d7e20-5aea-4b6c-83d5-fdfc52405dd1

thure:
This PR:
- resolves #7578 
- resolves #7424

https://github.com/user-attachments/assets/1114ed8d-6368-4326-ba12-b5f9818437da

Some caveats
- debt: tabster sometimes moves inside a group on 'tab'
- debt: tabster does not always focus a group on 'escape'
- because of this, it isn’t possible to move to the next item in the DOM using tab when focus is within the navtree


packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -125,6 +125,66 @@ export const DeckLayout = ({
...
+                  layoutMode === 'deck' && [
+                    'absolute inset-0',
+                    'transition-[padding] duration-200 ease-in-out',
+                    slots?.wallpaper?.classNames,
+                  ],
+                ]}
+                solo={layoutMode === 'solo'}
+                style={{ ...overscrollAmount }}
+              >
+                {layoutMode === 'solo' &&
```
richburdon, line 154: changed a lot of this in
https://github.com/dxos/dxos/pull/7642

packages/ui/react-ui-theme/src/styles/layers/focus-ring.css
```diff
@@ -1,13 +1,13 @@
 @layer base {
-  .ch-focus-ring {
+  .ch-focus-ring, .ch-focus-ring-inset {
```
richburdon, line 2: +1

packages/ui/react-ui/src/components/Main/Main.tsx
```diff
@@ -41,6 +45,30 @@ type MainContextValue = {
   setComplementarySidebarOpen: Dispatch<SetStateAction<boolean | undefined>>;
 };
 
+const landmarkAttr = 'data-main-landmark';
+
+const useLandmarkMover = (propsOnKeyDown: ComponentPropsWithoutRef<'div'>['onKeyDown'], landmark: string) => {
```
richburdon, line 50: interesting name -- can you cite a precedent or just add a comment?

packages/ui/react-ui-theme/src/styles/fragments/focus.ts
```diff
@@ -2,14 +2,7 @@
...
 
-import { mx } from '../../util';
-
-// TODO(burdon): Add m-1/2 to make focusRing visible.
-
-export const focusRing = mx(
-  'focus:outline-none focus-visible:z-[1] focus-visible:hover:outline-none dark:focus-visible:hover:outline-none focus-visible:ring-2 focus-visible:ring-offset-0',
-  'focus-visible:ring-primary-350 focus-visible:ring-offset-white dark:focus-visible:ring-primary-450 dark:focus-visible:ring-offset-black',
-);
+export const focusRing = 'ch-focus-ring';
```
richburdon, line 5: remove `mt-[2px]` in MarkdownEditor.tsx and Section.tsx?
assume Message.tsx is okay with it's current margin?

richburdon (approved):

packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
thure: I noticed, I rearranged these initially so their DOM order would allow tabbing, but I ended up needing it to do something different and now the DOM order doesn’t matter, so this should be easy enough to de-conflict.

packages/ui/react-ui/src/components/Main/Main.tsx
thure: 'mover' is from tabster: https://tabster.io/docs/mover
'landmark' is a standard term for this kind of element: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/landmark_role
can certainly add a comment to cite these

packages/ui/react-ui-theme/src/styles/fragments/focus.ts
thure: That would creep scope in a way this PR probably shouldn’t attempt to resolve since here it’s just deferring to the utility classname for the same styles. I’d prefer to land this and follow-up on that design debt separately.

richburdon: ok -- you'd just asked me to remind you about the places I added margins -- as long as you have this in mind i'll merge this now.

PR #7649, opened 2024-09-05
fix(react-ui-stack,react-ui-editor): Remove max-width from stack, apply max-widths to MarkdownEditor toolbar

thure:
Screenshot 2024-09-05 at 09 00 12


ZaymonFC (approved):
LGTM

PR #7656, opened 2024-09-05
fix(map): improve hasId check to not use `any`

### LINKED ISSUE #7653: [map] crash on adding second table
Occurs when you already have a map in a collection and you add another table to it:
```
TypeError: Cannot read properties of null (reading 'id')
```

Location:
https://github.com/dxos/dxos/blob/97aa72da8aa750067332bb37b706de81038bfb98/packages/plugins/experimental/plugin-map/src/components/MapDetectLocations.ts#L87

seagreen:
Close #7653. There might be other problems in the module if objects without `id` fields are coming out of `useQuery`, but this will fix the immediate error that occurs within `hasId`.


seagreen:
Closing in favor of https://github.com/dxos/dxos/pull/7658

Issue #7655, opened 2024-09-05
[composer] crash on adding second object to a collection


seagreen:
In a fresh account, if the first thing you do is create a collection with two tables in it, you get a crash on the second table creation.

The error is:
```
src-CCE_E-6O.js:2 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'id')
```

Here's a video of it happening in a fresh incognito identity:

https://github.com/user-attachments/assets/1ef55b63-953a-47f8-9e6d-2a803dca33fa

This seemed so similar to https://github.com/dxos/dxos/issues/7653 that I thought they were the same, but the error message here mentioned `undefined` while that issue mentions `none`, and the error locations are different.

PR #7657, opened 2024-09-05
fix(stack): Stack creator intent might be an array

ZaymonFC:
- Resolves: #7655


wittjosiah:
@ZaymonFC maybe rebase so this only includes your commit

ZaymonFC:
> @ZaymonFC maybe rebase so this only includes your commit

image

wittjosiah:
huh... I don't love that stack needs to inspect all these things...

looking at all the plugins, all of the stack creators are just a single intent anyways some are just wrapped in `[]`. I'd propose just removing the `[]` from all of those as a solution to this until we can remove stack creators altogether?

ZaymonFC:
> huh... I don't love that stack needs to inspect all these things...
> 
> looking at all the plugins, all of the stack creators are just a single intent anyways some are just wrapped in `[]`. I'd propose just removing the `[]` from all of those as a solution to this until we can remove stack creators altogether?

That might be a better idea!

wittjosiah (approved):

PR #7659, opened 2024-09-05
fix(map): show markers again

seagreen:
Turns out my previous PR https://github.com/dxos/dxos/pull/7658 stopped that crash by disabling markers entirely. `'id' in obj` will always return false for the ECHO objects we're interested in.

This still uses an `any`, but I think safely. It displays markers and the crash is still gone.

### DIFF
--- a/packages/plugins/experimental/plugin-map/src/components/MapDetectLocations.ts
+++ b/packages/plugins/experimental/plugin-map/src/components/MapDetectLocations.ts
@@ -85,7 +85,7 @@ export const useMapDetectLocations = (map: MapType): Marker[] => {
 export default useMapDetectLocations;
 
 const hasId = (obj: unknown): obj is { id: string } =>
-  typeof obj === 'object' && obj !== null && 'id' in obj && typeof obj.id === 'string';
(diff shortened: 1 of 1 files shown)


packages/plugins/experimental/plugin-map/src/components/MapDetectLocations.ts
```diff
@@ -85,7 +85,10 @@ export const useMapDetectLocations = (map: MapType): Marker[] => {
 export default useMapDetectLocations;
 
 const hasId = (obj: unknown): obj is { id: string } =>
-  typeof obj === 'object' && obj !== null && 'id' in obj && typeof obj.id === 'string';
```
wittjosiah, line 88: what the scenario where it fails `'id' in obj` but passes `Object.hasOwnProperty(obj, 'id')`?

seagreen: Great question. Strange stuff is afoot:
image

I'll investigate it more monday.

seagreen: Now that `hasId` has been removed in https://github.com/dxos/dxos/pull/7681 we don't need to look into this more.

PR #7660, opened 2024-09-06
feat: restore gem globe lib

richburdon:


packages/plugins/experimental/plugin-explorer/data/countries-110m.json
wittjosiah, line 1: is this meaningful to have in version control? not sure what best practice is here (the diff line count scared me 😂 )

PR #7673, opened 2024-09-06
feat(devtools): additional info for feed and echo explorer devtools 

yarolegovich:
### Details

* Show credential issuer and type for feed blocks in Feeds panel, as well as in the feed picker.
image

image

* Added filtering by space, content text search and object type column for Automerge panel.
image

* Added name / trimmed content display on objects panel.
image


packages/devtools/devtools/src/hooks/useContactBook.ts
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { useEffect, useState } from 'react';
+
+import { type Contact } from '@dxos/protocols/proto/dxos/client/services';
+import { useClient } from '@dxos/react-client';
+
+export const useContactBook = () => {
```
wittjosiah, line 10: if `client.halo.contacts` works now, I would put this hook in `@dxos/react-client` itself

packages/devtools/devtools/src/panels/echo/ObjectsPanel/ObjectsPanel.tsx
```diff
@@ -28,18 +28,32 @@ const textFilter = (text?: string) => {
...
-    meta: { cell: { classNames: textPadding } },
-    size: 220,
+    ...builder.string(),
+  }),
+  helper.accessor((item) => describeContent(item), {
+    id: 'content',
+    ...builder.string(),
   }),
-  helper.accessor((item) => (item.__deleted ? 'deleted' : ''), {
+  helper.accessor((item) => (isDeleted(item) ? '✅' : '❎'), {
```
wittjosiah, line 53: I might use a red X here to make it easier to distinguish by colour?

yarolegovich: though about it, but was too much red. I'll probably replace it with smth colorless

packages/devtools/devtools/src/hooks/useContactBook.ts
yarolegovich: yeah, a very basic (union all identities across all spaces recording which spaces are shared with them) implementation's been there since May

dmaretskyi (approved):

PR #7682, opened 2024-09-08
feat: Map/globe plugin with editor style updates.

richburdon:
- Tighten deps by reducing exports from react-ui-theme (partition plugin/lib code by directory).
- Attempts to fix style issues with plugins (missing classes, icons).


packages/apps/composer-app/vite.config.ts
```diff
@@ -88,7 +88,11 @@ export default defineConfig({
       content: [
         resolve(__dirname, './index.html'),
         resolve(__dirname, './src/**/*.{js,ts,jsx,tsx}'),
-        resolve(__dirname, '../plugins/*/src/**/*.{js,ts,jsx,tsx}'),
+        resolve(__dirname, '../../experimental/*/src/**/*.{js,ts,jsx,tsx}'),
+        resolve(__dirname, '../../plugins/*/src/**/*.{js,ts,jsx,tsx}'),
+        resolve(__dirname, '../../plugins/experimental/*/src/**/*.{js,ts,jsx,tsx}'),
+        resolve(__dirname, '../../sdk/*/src/**/*.{js,ts,jsx,tsx}'),
+        resolve(__dirname, '../../ui/*/src/**/*.{js,ts,jsx,tsx}'),
```
thure, line 95: We may benefit instead from putting these in `resolveContent.ts` as known peers. See: https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-theme/src/config/resolveContent.ts#L10

packages/plugins/plugin-table/src/components/TableSection.tsx
```diff
@@ -10,7 +10,7 @@ import { ObjectTable, type ObjectTableProps } from './ObjectTable';
 
 const TableSection: FC<Omit<ObjectTableProps, 'getScrollElement'>> = ({ table }) => (
   <Table.Root>
-    <Table.Viewport classNames='max-bs-96 is-full sticky-top-0 !bg-[--surface-bg]'>
+    <Table.Viewport classNames='max-bs-96 is-full sticky-top-0 !bg-[--surface-bg] -ml-[1px] -mt-[1px]'>
```
thure, line 13:
```suggestion
    <Table.Viewport classNames='max-bs-96 is-full sticky-top-0 !bg-[--surface-bg] -mis-px -mbs-px'>
```

packages/experimental/gem-globe/src/hooks/useTour.ts
```diff
@@ -60,37 +61,31 @@ const startTour = async (controller: GlobeController, features: Features) => {
 
     await d3
       .transition()
-      .duration(Math.max(1_000, distance * 2_000))
+      .duration(Math.max(2_000, distance * 2_000))
       .tween('render', () => (t) => {
-        projection.rotate(iv(t));
-        setRotation(projection.rotate());
+        const t1 = Math.max(0, Math.min(1, t * 2 - 1));
+        const t2 = Math.min(1, t * 2);
```
thure, line 67: What is `gem-globe` and why are changes to it part of a PR named “Theme plugin tweaks”? Can these concerns be separated?

packages/ui/react-ui-theme/src/config/tokens.ts
```diff
@@ -162,8 +162,6 @@ const physicalSeries = {
   ...huePalettes,
 };
 
-export type HuePalette = keyof typeof huePalettes;
```
thure, line 165: I believe this type was being used somewhere — did removing this not break the build?

thure: Ah nevermind, I see it’s exported from another file.

thure (commented):
Seems fine to me, though we should consider moving Composer’s content into react-ui-theme’s known peers. Also the changes to `gem-globe` look unrelated, can those be separated?

packages/plugins/plugin-table/src/components/TableSection.tsx
richburdon: We need to standardize all of this of course, but good for now.

packages/experimental/gem-globe/src/hooks/useTour.ts
richburdon: yes could be separated, but fairly minor and ran into problems with them trying to debug style issues here with this. gem is the SVG/canvas lib; globe is the canvas 3d map.

packages/ui/react-ui-theme/src/config/tokens.ts
richburdon: refactored so that we can avoid exporting lots of defs to a lib that should really only be part of the plugin. recommend splitting up packages like this into plugin and lib directories.

packages/plugins/plugin-table/src/components/TableSection.tsx
thure: What do you mean by standardize exactly? `px` is a “standard” length in Tailwind, and logical properties are meanginfully distinct from physical ones (see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)) and take precedence in the cascade.

richburdon:
> Seems fine to me, though we should consider moving Composer’s content into react-ui-theme’s known peers. Also the changes to `gem-globe` look unrelated, can those be separated?

yes for sure re splitting things up, but consider this PR a WIP -- i'm trying to debug some style issues (currently in maps, and fixing things along the way). i won't commit until we have a chance to discuss.

packages/ui/react-ui-theme/src/config/tokens.ts
thure: > a lib that should really only be part of the plugin

Do you mean to say the hue palettes should not be defined in `react-ui-theme`? We already have `react-ui-theme` and `plugin-theme`, and I’m amenable to moving definitions between them as needed, though I think there are good reasons hue palettes are defined in this package and not elsewhere.

PR #7690, opened 2024-09-09
ci: add github environments

wittjosiah:
Per environment env vars for Composer deploy using https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-environments-for-deployment

### DIFF
--- a/.github/workflows/publish-all.yml
+++ b/.github/workflows/publish-all.yml
@@ -6,12 +6,9 @@ on:
         default: false
   push:
     branches:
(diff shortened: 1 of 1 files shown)


nferch:
What are the advantages for using Github environments for deployment-specific variables over e.g. https://github.com/dxos/dxos/blob/295261af5faae951326261e14f9d93565379591e/.github/workflows/env/staging ? Is the intention to migrate everything to that?

PR #7696, opened 2024-09-10
fix: update vite-plugin-icons

richburdon:
- Recurses into dependencies while processing files in serve mode.
- Please leave in tools at least until we do a monorepo tools review.


thure (approved):
lgtm!

PR #7702, opened 2024-09-10
fix: shortcuts dialog

wittjosiah:
- change shortcut to avoid conflicts with comments shortcut
- fix dialog rendering

image


thure (approved):
lgtm

PR #7704, opened 2024-09-11
feat(lit-stories,lit-spreadsheet): Spreadsheet Lit component

thure:
This PR:
- lays the groundwork for the new grid webcomponent
- begins resolution of #7615.

Issue #7615, opened 2024-09-02
[composer] universal grid control


richburdon:
# Motivation

We need a universal grid component that is shared between Tables and Sheets (and possibly other tabular data plugins). It should be fast and able to deal with large datasets.

## Features
- works with sparse data (e.g., sheets).
- works with dense typed row-based data (e.g., tables).
- frozen rows and columns
- frozen columns on left and right (e.g., delete button at end of table row).
- resizable rows and columns.
- draggable rows and columns (reorder).
- able to select rows, columns, ranges.
- editable cells via CM.

## Demo

<img width="1728" alt="Screenshot 2024-09-02 at 1 19 35 PM" src="https://github.com/user-attachments/assets/8ad24b43-7522-47c1-8c88-80aa944f7b9b">

## Prior art
- glide

thure:
Resolved by #7900 and #7853.

PR #7713, opened 2024-09-11
feat(react-ui-grid): Intermediate React component for DxGrid

thure:
This PR
- succeeds #7704
- continues resolution of #7615
  - successor will implement content editing, then a further successor will integrate with `SheetPlugin`.


packages/ui/lit-grid/src/position.ts
```diff
@@ -2,58 +2,35 @@
 // Copyright 2024 DXOS.org
 //
 
-export type CellPosition = { i: number; j: number };
+export type PhysicalPosition = { c: number; r: number };
```
wittjosiah, line 5: c/r is column/row?

thure: Yes, though this is an earlier implementation that could use improvement. There is `DxAxis` which could be this object’s keys.

thure: Actually very little of the code in this file got used, I’m removing it.

PR #7720, opened 2024-09-12
chore: lit stories update

richburdon:
I recently updated the main storybook configs, updating this to match.


thure (approved):
thanks for this 👍

PR #7723, opened 2024-09-13
chore: Codemirror semantic styles.

richburdon:
NOTE: Given the late hour I have dark mode reader on so I haven't checked how all the specific colors look yet! :).

- [x] Replaced getToken with semantic colors.
- [x] Moved markdown styles out of base theme.
- [x] Removed light/dark selectors from styles and converted to regular theme (not base theme).
- [x] Updated styles for search panel. 
- [x] Toolbar search button. 

Questions (@thure):
- how to deal with fonts?
- how to normalize inputs/buttons, etc. (e.g., .cm-panel).

Screenshot 2024-09-13 at 3 26 47 PM


packages/ui/react-ui-theme/src/config/tokens.ts
```diff
@@ -301,6 +302,49 @@ const semanticColors = {
       light: ['red', 700],
       dark: ['red', 300],
     },
+
+    //
+    // Codemirror
```
wittjosiah, line 307: I imagine that we may want to factor many of these out from "cm specific styles" to "general theme styles" over time, but given they're specifically being used for cm editor right now this makes that quite clear 👍

packages/plugins/plugin-mermaid/src/extensions/mermaid.ts
```diff
@@ -155,6 +152,6 @@ const styles = EditorView.baseTheme({
   // },
   '& .cm-mermaid-error': {
     display: 'inline-block',
-    color: getToken('extend.colors.red.500'),
+    color: 'var(--dx-error)',
```
thure, line 155: I meant to remove/deduplicate/qualify this `error` token with the rest in `messageValence.ts` — I think the scope of this PR could, maybe should, be extended to handle that too.

packages/ui/react-ui-theme/src/config/tokens.ts
thure: Yeah this is an interesting use-case… maybe in a later iteration `react-ui-editor` or MarkdownPlugin can export these sememes and composer could provide it to the Vite config.

packages/plugins/plugin-mermaid/src/extensions/mermaid.ts
thure: Maybe not, though, this PR seems to just address CM and that’s fine, can remove the bare `error` token in another pass that handles `messageValence`.

PR #7725, opened 2024-09-13
feat(mesh): unified webrtc

yarolegovich:
### Details

* Removed libdatachannel and simplepeer webrtc implementations (`node-datachannel/polyfill` still used).
* `RtcTransportChannel` is what implements `Transport` backed by `RTCDataChannel`.
* `RtcPeerConnection` is the class we'll later use as a `Muxer`. It acts as `RtcTransportChannel` factory, opening a WebRTC connection when the first channel is open and destroying it after the last channel is closed. Marked as a resource.
* `rtc-connection-factory` encapsulates polyfill vs native selection.
* `rtc-transport-factory` currently simply creates an `RtcPeerConnection` instance and uses it to create a `RtcTransportChannel` instance. We can keep a `peerId -> RtcPeerConnection` cache when signals are routed by `peerId`.
* `initiator` role is chosen based on the key ordering. Initiator peer is responsible for creating an offer and data channels. 
* Perfect negotiation doesn't give us much and will be quite tricky to implement given the differences in polyfill and native implementation behavior when it comes to data-channels. 
* Changed `Transport` open/close signature to match `Resource` class.
* Increased reliable message delivery retry timeout 300ms -> 1s.

Added tests that match what we had for the old implementations, tested with cli and two browsers on invitations and data replication.
Probably best to wait before I'm back to go through the implementation once again and add some extra tests.


packages/core/mesh/network-manager/src/transport/memory-transport.test.ts
```diff
@@ -72,3 +78,15 @@ describe('MemoryTransport', () => {
     );
   });
 });
+
+const createTransportOptions = (options: Partial<TransportOptions>): TransportOptions => {
+  return {
+    initiator: false,
```
dmaretskyi, line 84: is the flag ignored now?

packages/core/mesh/network-manager/package.json
```diff
@@ -57,13 +57,12 @@
     "@dxos/protocols": "workspace:*",
```
dmaretskyi, line 57: Is the "browser" field above up-to-date?

I also think we should switch to [Subpath Imports](https://nodejs.org/api/packages.html#subpath-imports) for all partform-specific functionality it's supported natively everywhere without requiring extra tooling.

packages/core/mesh/network-manager/src/transport/memory-transport.test.ts
yarolegovich: yeah, considered removing it, but it's used in other implementations

PR #7728, opened 2024-09-14
feat: react-ui Icon class

richburdon:


packages/ui/react-ui-theme/src/styles/components/icon.ts
```diff
@@ -0,0 +1,15 @@
...
+// Copyright 2023 DXOS.org
+//
+
+import { type ComponentFunction, type Theme } from '@dxos/react-ui-types';
+
+import { mx } from '../../util';
+
+export type IconStyleProps = {};
+
+export const iconRoot: ComponentFunction<IconStyleProps> = (_props, etc) => mx(etc);
```
thure, line 11: Could maybe build in `getSize` since we use that so very often, then we can pass it in through the props and fall back to some default.

PR #7740, opened 2024-09-16
refactor(threads): Thread plugin implementation shouldn't be tied to documents

ZaymonFC:


packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -327,16 +327,8 @@ export const createComment: Command = (view) => {
...
-    const id = options.onCreate?.({ cursor, from, location: view.coordsAtPos(from) });
-    if (id) {
-      // Update range.
-      view.dispatch({
-        effects: setSelection.of({ current: id }),
-        selection: { anchor: to },
-      });
-
-      return true;
-    }
```
ZaymonFC, line 339: This selection gets set in the bi-directional linking based on `scroll_into_view`

packages/plugins/plugin-thread/src/ThreadPlugin.tsx
```diff
@@ -310,6 +311,12 @@ export const ThreadPlugin = (): PluginDefinition<ThreadPluginProvides> => {
             case 'article':
             case 'complementary': {
               const location = navigationPlugin?.provides.location;
+              const { plugins } = usePlugins();
+
+              // TODO(Zan): This feels a bit clunky.
```
wittjosiah, line 316: not significantly better but you can factor this logic out and use `useResolvePlugin` instead

wittjosiah (approved):

PR #7756, opened 2024-09-18
fix: vite-icons-manifest

richburdon:
- cache list of icons across builds (since chunks are cached).


thure (approved):
Very interesting, lgtm

packages/apps/composer-app/vite.config.ts
```diff
@@ -232,7 +237,20 @@ export default defineConfig({
           },
         ]
       : []),
-  ],
+    // Copy post-build assets to dist.
```
wittjosiah, line 240: if this is the fix shouldn't this go in the icons plugin?

Issue #7758, opened 2024-09-18
CF Demo


richburdon:
## PR

- https://github.com/dxos/dxos/compare/burdon/function-triggers?expand=1

## Tasks

- [ ] Update FunctionPlugin to create triggers
  - [x] NOTE: currently only visible in stack
  - [x] Combo picker should select from published functions
  - [x] When enabled should instantiate local subscription in browser (reuse logic from core/functions)
    -  https://github.com/dxos/dxos/blob/main/packages/core/functions/src/trigger/trigger-registry.ts#L50
  - [ ] Trigger function invocation in TriggerEditor
    -  
  - [ ] Create chess function script
- [ ] make separate browser and node exports for @dxos/functions (or factor out)
- [ ] Script editor; dark background when attended
- [ ] Stretch
  - [ ] Trigger meta property editor
  - [ ] Update ChainPlugin to create prompts
  - [ ] Create GPT function script that imports prompt object
- [ ] Deploy to labs.composer.space

## Demo

https://github.com/user-attachments/assets/11a961f7-caaa-4ebc-84a6-856d3173ef9b

richburdon:
![Image](https://github.com/user-attachments/assets/a627a269-17c8-4451-8394-4495853b061c)

PR #7760, opened 2024-09-18
feat: Function triggers

richburdon:
Resolves https://github.com/dxos/dxos/issues/7758#issue-2534827901


packages/ui/lit-grid/package.json
wittjosiah, line 1: @dmaretskyi I think this change was generated by toolbox, is there anyway we could have it ignore this package?

packages/plugins/experimental/plugin-script/src/types/script.ts
```diff
@@ -5,16 +5,23 @@
...
+  // TODO(burdon): Change to has of deployed content.
   // Whether source has changed since last deploy.
   changed: S.optional(S.Boolean),
 }) {}
 
 export class FunctionType extends TypedObject({ typename: 'dxos.org/type/Function', version: '0.1.0' })({
+  name: S.optional(S.String),
   version: S.Number,
   // Local binding to a function name.
+  // TODO(burdon): Move binding to inner type.
```
wittjosiah, line 24: Binding should not be on the `ScriptType`, those are only for authoring and not for invoking

packages/plugins/experimental/plugin-script/src/types/script.ts
```diff
@@ -5,16 +5,23 @@
 import { ref, S, TypedObject } from '@dxos/echo-schema';
 import { TextType } from '@dxos/plugin-markdown/types';
 
+// TODO(burdon): Which is the outer type? I.e., in the navbar?
+
 export class ScriptType extends TypedObject({ typename: 'dxos.org/type/Script', version: '0.1.0' })({
+  // TODO(burdon): Change to URI?
```
wittjosiah, line 11: name is user-visible, why would it be a uri?

packages/plugins/experimental/plugin-script/src/types/script.ts
```diff
@@ -5,16 +5,23 @@
 import { ref, S, TypedObject } from '@dxos/echo-schema';
 import { TextType } from '@dxos/plugin-markdown/types';
 
+// TODO(burdon): Which is the outer type? I.e., in the navbar?
```
wittjosiah, line 8: these types were not intended to be hierarchical, the script reference for functions is an optional reference denoting origin rather than ownership

packages/experimental/gem-globe/src/components/Globe.tsx
```diff
@@ -256,8 +256,8 @@ const GlobeZoomControls = ({ classNames, position = 'bottom-left', onAction }: G
 };
 
 const GlobeActionControls = ({ classNames, position = 'bottom-right', onAction }: GlobeControlsProps) => {
+  // TODO(wittjosiah): This is a hack to get the theme to work. Gem isn't getting global theme context.
```
wittjosiah, line 259: this comment makes more sense directly beside the component to which it is referring

PR #7768, opened 2024-09-20
chore: Add LOKI event firing on composer assets request

mykola-vrmchk:
### Motivation / Background

This Pull Request ...


packages/apps/composer-app/src/functions/_worker.ts
```diff
@@ -8,9 +8,40 @@
...
+        })
+      }).then(async (response) => {
+        if (!response.ok) {
+          console.log("LOKI telemetry fetch failed:", await response.text());
+        }
+      }).catch((error) => console.error("LOKI telemetry fetch error:", error));
+    }
+
+    // Wait for both promises to complete.
+    await Promise.all([assetsPromise, telemetryPromise]);
```
wittjosiah, line 42: does it need to wait for the telemetry promise to resolve before serving the assets? The request for assets shouldn't fail just because of an issue with loki

mykola-vrmchk: It is necessary, I tested it locally, and if it is not awaited the event is not propagated to grafana, I can add timeout on `telemetryPromise`

PR #7771, opened 2024-09-20
feat(map): get locations from throughout the space, not just neighboring plugins

### LINKED ISSUE #7769: [map] display all locations in the space, not just ones from neighboring tables

seagreen:
Close #7769


packages/plugins/experimental/plugin-map/src/hooks/useMarkers.ts
```diff
@@ -4,55 +4,32 @@
...
+    space,
     (obj) => {
+      const schemaNames = schemas.map((schema) => schema.id);
       const typename = getTypename(obj);
-      return typename ? schemaIds.includes(typename) : false;
+      return typename ? schemaNames.includes(typename) : false;
     },
     undefined,
-    [JSON.stringify(Array.from(schemaIds))], // TODO(burdon): Hack.
+    [schemas],
```
seagreen, line 32: @wittjosiah You mentioned the possibility of removing this `[schemas]` dep. At the moment if I remove it the map doesn't populate with icons, neither on initial load nor if you add new map data rows. Any idea what could be causing that?

A related question: is the expected behavior of `useQuery` to re-run when any data in the db that might affect the result of the query changes, even if the filter or dependency arguments haven't changed?

wittjosiah: In this case it's definitely needed because the query is just a callback. Where I was suggesting it might not be was when using `Filter` because iirc if the filter changes it would rerun the query

seagreen: Ah it makes sense now, thank you!

packages/plugins/experimental/plugin-map/src/hooks/useMarkers.ts
```diff
@@ -4,55 +4,32 @@
...
-    //  is forbidden since the number of 'or's will vary as `schemaIds` changes,
-    //  and the internals of `useQuery` require it to be fixed:
+    space,
     (obj) => {
+      const schemaNames = schemas.map((schema) => schema.id);
       const typename = getTypename(obj);
-      return typename ? schemaIds.includes(typename) : false;
+      return typename ? schemaNames.includes(typename) : false;
     },
     undefined,
```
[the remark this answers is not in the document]
wittjosiah: undefined is an acceptable value so I think that's probably what I'd do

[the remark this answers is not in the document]
seagreen: Sounds great!

packages/plugins/experimental/plugin-map/src/hooks/useMarkers.ts
```diff
@@ -66,13 +43,13 @@ export const useMarkers = (map: MapType): MapMarker[] => {
           lng: row.longitude,
         },
       }));
-  }, [objects]);
+  }, [JSON.stringify(objects.map((obj) => obj.id))]); // TODO(seagreen): Hack.
```
seagreen, line 46: This way adding non-location ECHO objects doesn't trigger a change to our returned object reference.

This might not be great, but it's not new, previously we did the deleted line above:

> [JSON.stringify(Array.from(schemaIds))], // TODO(burdon): Hack.

richburdon: ok to merge, but pls ask @dmaretskyi about this.

richburdon (approved):

packages/plugins/experimental/plugin-map/src/hooks/useMarkers.ts
seagreen: Done!

PR #7772, opened 2024-09-20
chore: fix storybook crash

wittjosiah:
ensures that node deps in resolveContent are not pulled in when importing browser code from the config directory


thure (approved):
so bizarre, but lgtm yes

PR #7781, opened 2024-09-22
fix: remove react-async

richburdon:
moved into react-hooks.
prompting discussion on best practice, esp. for async effects.


packages/ui/primitives/react-hooks/esbuild-server.config.js
```diff
@@ -2,6 +2,8 @@
 // Copyright 2020 DXOS.org
 //
 
+// TODO(burdon): Are these required?
```
wittjosiah, line 5: esbuild-server is long dead I think? so probably not

packages/ui/primitives/react-hooks/src/useAsyncCallback.ts
```diff
@@ -4,6 +4,9 @@
 
 import { useEffect, useState } from 'react';
 
+/**
+ * @deprecated
```
wittjosiah, line 8: why?

wittjosiah (approved):

packages/ui/primitives/react-hooks/src/useAsyncCallback.ts
richburdon: I was planning to merge with useAsyncEffect but decided against pending a longer discussion.

PR #7801, opened 2024-09-23
fix(maps): correctly add new markers to the live map view

### LINKED ISSUE #7792: [map] useMemo caching is too aggressive
The result is that new markers added in another tab won't appear live in an already loaded map correctly.

I know what's causing this, I'll make a PR.

seagreen:
Close #7792

When the user creates a location row it appears in the DB first partially completed. The final version only appears later once the user has filled out all the columns.

Previously we were caching too early in this process. This captured partially completed objects which lacked both `latitude` and `longitude`. Thus they didn't appear on the map.


packages/plugins/experimental/plugin-map/src/hooks/useStabilizedObject.ts
```diff
@@ -0,0 +1,17 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { useState, useEffect } from 'react';
+
+export const useStabilizedObject = <T>(obj: T, customEqual: (oldObject: T, newObject: T) => boolean) => {
```
wittjosiah, line 7: when making a generic util like this try to include a jsdoc about the intended use case/what it's for

packages/plugins/experimental/plugin-map/src/hooks/useStabilizedObject.ts
```diff
@@ -0,0 +1,17 @@
...
+  const [stableObj, setStableObj] = useState(obj);
+
+  useEffect(() => {
```
wittjosiah, line 10: beware that when using an effect here the returned markers above will be a render cycle behind the map and so if the map input changes there will be a render where the returned markers are for the previous map

packages/plugins/experimental/plugin-map/src/hooks/useStabilizedObject.ts
seagreen: Good idea, done.

packages/plugins/experimental/plugin-map/src/hooks/useStabilizedObject.ts
seagreen: Ooof. Added a TODO comment for now. I'd also be down for trying to get a correct implementation of `useStabilizedObject`, but maybe not in this PR unless you know exactly what to do.

PR #7803, opened 2024-09-23
feat(functions): make ws trigger work in browser

yarolegovich:
### Details

* Websocket trigger works in browser.
* Fix for trigger reload.


richburdon (approved):
nice

PR #7809, opened 2024-09-24
feat: Joins in queries

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-db/src/core-db/core-database-query-context.ts
```diff
@@ -163,25 +164,53 @@ export class CoreDatabaseQueryContext implements QueryContext {
...
       spaceId: core.database!.spaceId,
       spaceKey: core.database!.spaceKey,
-      object: core.toPlainObject(),
+      object: data,
       match: result && { rank: result.rank },
       resolution: { source: 'remote', time: Date.now() - queryStartTimestamp },
     } satisfies QueryResult;
   }
+
+  private async _recursivelyJoinFields(
```
yarolegovich, line 189: todo: handle recursive references?
```ts
const object = { parent: null, child: {} };
object.child.parent = object;
```

packages/core/echo/echo-db/src/core-db/core-databasse.test.ts
```diff
@@ -517,6 +517,35 @@ describe('CoreDatabase', () => {
...
+    test('query with join', async () => {
+      await using testBuilder = await new EchoTestBuilder().open();
+      const { crud } = await testBuilder.createDatabase();
+
+      const { id: id1 } = await crud.insert({ title: 'Inner' });
+      const { id: id2 } = await crud.insert({ title: 'Outer', inner: { '/': id1 } });
+      await crud.flush({ indexes: true });
+
+      {
+        const object = await crud.query({ id: id2 }, { include: { inner: true } }).first();
```
yarolegovich, line 530: let's also add a test for nested spec `{ inner: { nested: true } }`

packages/core/protocols/src/proto/dxos/echo/filter.proto
```diff
@@ -52,7 +55,7 @@ message QueryOptions {
    * Query only in specific spaces.
    * @deprecated Use `spaceIds` instead.
    */
-  repeated dxos.keys.PublicKey spaces = 3;
+  repeated dxos.keys.PublicKey spaces = 6;
```
yarolegovich, line 58: why did this change? 🤔

yarolegovich (approved):

packages/core/echo/echo-db/src/core-db/core-database-query-context.ts
dmaretskyi: This will require the JoinSpec object to be recursive which we would wanna prohibit

PR #7820, opened 2024-09-25
fix(perf): Deck re-renders when scrolling deck layout

ZaymonFC:


packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -73,42 +73,36 @@ export const DeckLayout = ({
   const searchPlugin = usePlugin('dxos.org/plugin/search');
   const fullScreenSlug = useMemo(() => firstIdInPart(layoutParts, 'fullScreen'), [layoutParts]);
 
-  const [scrollLeft, setScrollLeft] = useState<number | null>(null);
```
ZaymonFC, line 76: Updating this on scroll tick caused unnecessary `DeckLayout` renders.

thure (approved):
Seems fine to me, thanks!

PR #7830, opened 2024-09-25
feat: Edge sync indicator

dmaretskyi:
image
image


packages/plugins/plugin-space/src/components/SyncStatus.tsx
```diff
@@ -0,0 +1,136 @@
...
+    });
+  }, []);
+
+  const needsToUpload = state.missingOnRemote > 0 || state.differentDocuments > 0;
+  const needsToDownload = state.missingOnLocal > 0 || state.differentDocuments > 0;
+  const notConnectedToEdge = state.spacesNotConnectedToEdge.size > 0;
+
+  return (
+    <StatusBar.Item title={JSON.stringify(state, null, 2)}>
+      {notConnectedToEdge ? (
```
richburdon, line 33:
```tsx
import { Icon } from 'react-ui';
<Icon icon={'ph--cloud-slash--regular' || ...} size={4} />
```

packages/plugins/plugin-space/src/components/SyncStatus.tsx
```diff
@@ -0,0 +1,136 @@
...
+      setState(state);
+    });
+  }, []);
+
+  const needsToUpload = state.missingOnRemote > 0 || state.differentDocuments > 0;
+  const needsToDownload = state.missingOnLocal > 0 || state.differentDocuments > 0;
+  const notConnectedToEdge = state.spacesNotConnectedToEdge.size > 0;
+
+  return (
+    <StatusBar.Item title={JSON.stringify(state, null, 2)}>
```
richburdon, line 32: TODO(burdon): Click for popup status

packages/plugins/plugin-space/src/components/SyncStatus.tsx
```diff
@@ -0,0 +1,136 @@
...
+  install(client.spaces.get());
+
+  return () => {
+    for (const unsubscribe of Object.values(unsubscribeCallbacks)) {
+      unsubscribe();
+    }
+  };
+};
+
+type EdgeSyncState = {
```
richburdon, line 74: is there a better place to put all of the edge deps? (outside of ux?)

richburdon (approved):

packages/plugins/plugin-space/src/components/SyncStatus.tsx
dmaretskyi: so its not in plugin-space?

yeah, probably we should make an edge plugin

PR #7838, opened 2024-09-26
feat: Composer CRX

richburdon:
Screenshot 2024-09-26 at 9 49 12 PM


dmaretskyi (approved):

packages/apps/composer-crx/.eslintrc.cjs
wittjosiah, line 1: I think this should go in experimental

PR #7841, opened 2024-09-26
fix: Source maps for production build

dmaretskyi:
- Remove `top-level-await-plugin` as it is breaking sourcemaps
- Add `rollup-plugin-sourcemaps` to make vite read sourcemaps of workspace packages
- Preserve class and function names during minification (helps preserve log and context names)
- Only enable tsconfig paths mappings in local dev mode
- Improve error messages in sdk

### DIFF
--- a/packages/apps/composer-app/vite.config.ts
+++ b/packages/apps/composer-app/vite.config.ts
@@ -6,6 +6,7 @@ import { sentryVitePlugin } from '@sentry/vite-plugin';
 import ReactPlugin from '@vitejs/plugin-react-swc';
 import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
 import { join, resolve } from 'node:path';
+import sourceMaps from 'rollup-plugin-sourcemaps';
 import { visualizer } from 'rollup-plugin-visualizer';
 import { defineConfig, searchForWorkspaceRoot } from 'vite';
 import Inspect from 'vite-plugin-inspect';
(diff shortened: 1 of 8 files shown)


packages/apps/composer-app/vite.config.ts
```diff
@@ -81,13 +85,15 @@ export default defineConfig({
   },
   worker: {
     format: 'es',
-    plugins: () => [TopLevelAwaitPlugin(), WasmPlugin()],
+    plugins: () => [WasmPlugin()],
   },
   plugins: [
+    sourceMaps(),
```
wittjosiah, line 91: what does this do that `sourcemap: true` doesn't?

packages/apps/composer-app/vite.config.ts
```diff
@@ -49,9 +49,13 @@ export default defineConfig({
       ],
     },
   },
+  esbuild: {
+    keepNames: true,
+  },
   build: {
     sourcemap: true,
     minify: !isFalse(process.env.DX_MINIFY),
+    target: ['chrome89', 'edge89', 'firefox89', 'safari15'],
```
wittjosiah, line 58: we need to comment this to explain why these targets

packages/apps/composer-app/vite.config.ts
wittjosiah: ah I see it helps read the sourcemaps of workspace packages

PR #7842, opened 2024-09-26
feat: Sync status indicator

richburdon:
- new indicator for sync
- fix icons for perf panel (and closed by default)

Screenshot 2024-09-26 at 5 02 18 PM


packages/plugins/plugin-space/src/components/SyncStatus/SyncStatus.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <div className='flex flex-col text-xs'>
+      <h1 className='p-2'>Progress</h1>
+      <div className='flex flex-col gap-[2px] my-[2px]'>
+        {Object.entries(state).map(
+          ([spaceId, { localDocumentCount, remoteDocumentCount, missingOnLocal, missingOnRemote }]) => (
+            <Candle
+              key={spaceId}
+              classNames='h-2'
+              up={{ count: remoteDocumentCount, total: remoteDocumentCount + missingOnRemote }}
```
dmaretskyi, line 70: might wanna account for `differentDocuments` which are the documents that both peers have different versions of

dmaretskyi (approved):

PR #7832, opened 2024-09-25
fix: set contents of complementary sidebar based on attention

wittjosiah:
Resolves #7650
Resolves #7825

PR #7843, opened 2024-09-27
fix: always maintain attention to a plank

### LINKED ISSUE #7833: [composer] always maintain attention to something
1) it's jarring when losing attention on large planks just because of clicking a button (solo mode especially)
2) complementary sidebar is now driven by attention and would be preferable to reduce the amount of time it is empty (ideally only when the deck is empty)

[7:09 PM]Will: Attention should always mean something within that surface has focus, or if focus is outside any content entity then the attention treament signifies the surface of the content that last had focus.

wittjosiah:
Resolves #7833 

includes e2e test fixes for this and #7832


packages/plugins/plugin-deck/src/components/DeckLayout/DeckLayout.tsx
```diff
@@ -75,6 +75,15 @@ export const DeckLayout = ({
   const scrollLeftRef = useRef<number | null>();
   const deckRef = useRef<HTMLDivElement>(null);
 
+  // Ensure the first plank is attended when the deck is first rendered.
+  useEffect(() => {
+    const firstId = layoutMode === 'solo' ? firstIdInPart(layoutParts, 'solo') : firstIdInPart(layoutParts, 'main');
+    if (attention.attended.size === 0 && firstId) {
+      // TODO(wittjosiah): This is resulting in a focus ring around the plank.
```
thure, line 82: This is intended, though should only be if you’re using the keyboard.

PR #7844, opened 2024-09-27
fix: comments don't load if object wasn't loaded first

wittjosiah:


ZaymonFC (approved):
Works for me. We should workout how to do this without a timeout at somepoint.

PR #7845, opened 2024-09-27
fix(cli): imports and default configs

yarolegovich:
### Details

cli is broken on main.
esm imports are not working:
```
Error: Cannot find module 'packages/common/hypercore/node_modules/sodium-universal/memory' imported from packages/common/hypercore/dist/lib/node-esm/index.mjs
Did you mean to import "sodium-universal/memory.js"?
    at finalizeResolution (node:internal/modules/esm/resolve:265:11)
    at moduleResolve (node:internal/modules/esm/resolve:933:10)
    at defaultResolve (node:internal/modules/esm/resolve:1157:11)
```

and different signaling prevents connecting a local agent with browser

### DIFF
--- a/packages/sdk/client/src/services/local-client-services.ts
+++ b/packages/sdk/client/src/services/local-client-services.ts
@@ -59,9 +59,9 @@ const setupNetworking = async (
 
   const signals = config.get('runtime.services.signaling');
   const edgeFeatures = config.get('runtime.client.edgeFeatures');
-  if (signals) {
+  if (signals || edgeFeatures?.signaling) {
     const {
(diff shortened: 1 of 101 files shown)


packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -59,9 +59,9 @@ const setupNetworking = async (
 
   const signals = config.get('runtime.services.signaling');
   const edgeFeatures = config.get('runtime.client.edgeFeatures');
-  if (signals) {
+  if (signals || edgeFeatures?.signaling) {
```
mykola-vrmchk, line 62:
```suggestion
  if (signals && !edgeFeatures?.signaling) {
```

PR #7848, opened 2024-09-28
feat: Compute graph and CM sheet extension

richburdon:
- [x] Restructure to isolate HF defs.
- [x] ComputeGraphRegistry root class.
- [x] Functionmanager abstraction.
- [x] CM extension for graph. 
- [x] Storybooks
- [x] Async import dep from plugin to HF.
- [x] Factor out HF dependency from SheetPlugin.
- [ ] Browser and non-browser tests.
- [ ] Factor out Space dependency from SheetModel and generalize compute dependency graph.

<br />

Screenshot 2024-09-28 at 6 32 57 PM

<br />

Screenshot 2024-09-29 at 12 49 08 PM

<br />

https://github.com/user-attachments/assets/a529b05e-9d75-400a-a205-471053fecb18


packages/plugins/plugin-sheet/src/graph/compute-node.ts
```diff
@@ -0,0 +1,52 @@
...
+  // TODO(burdon): Remove?
+  get graph() {
+    return this._graph;
+  }
+
+  get hf() {
+    return this._graph.hf;
+  }
+
+  getValue(cell: CellAddress): CellScalarValue {
```
dmaretskyi, line 38: Is there a way to support other addressing formats for non-sheet objects? For example acessing table data

wittjosiah: you just need to be able to map to a cell address

packages/plugins/plugin-sheet/src/model/sheet-model.browser.test.ts
```diff
@@ -4,43 +4,50 @@
...
+    const sheet = createSheet({ rows: 5, columns: 5 });
+    const model = new SheetModel(graph, sheet);
+    await model.open();
+    return { graph, model };
   };
 
   test('create', async () => {
-    const model = await createModel();
+    const { model } = await createModel();
+    console.log(model);
```
wittjosiah, line 35: 👀

packages/plugins/plugin-table/src/components/ObjectTable/ObjectTable.tsx
```diff
@@ -29,9 +29,15 @@ export type ObjectTableProps = Pick<TableProps<any>, 'stickyHeader' | 'role'> &
 export const ObjectTable = ({ table, role, stickyHeader }: ObjectTableProps) => {
   const space = getSpace(table);
   const [showSettings, setShowSettings] = useState(false);
-
   useEffect(() => setShowSettings(!table.schema), [table.schema]);
 
+  const [schemas, setSchemas] = useState<DynamicSchema[]>([]);
+  useEffect(() => {
+    if (space) {
+      void space.db.schema.list().then(setSchemas).catch();
```
wittjosiah, line 37: todo don't don't void promise in useEffect

packages/sdk/client/src/echo/index.ts
```diff
@@ -12,8 +12,9 @@ export {
   getTypename,
   Expando,
   type EchoReactiveObject,
-  type ReactiveObject,
   type ObjectMeta,
+  type ReactiveObject,
+  type S, // TODO(burdon): Export as Schema?
```
wittjosiah, line 17: wasn't it suggested that this shouldn't be reexported?

packages/plugins/plugin-sheet/src/SheetPlugin.tsx
```diff
@@ -6,50 +6,43 @@ import { GridNine, type IconProps } from '@phosphor-icons/react';
...
-
+      invariant(client);
+      let remoteFunctionUrl: string | undefined;
       if (client.config.values.runtime?.services?.edge?.url) {
         const url = new URL('/functions', client.config.values.runtime?.services?.edge?.url);
         url.protocol = 'https';
         remoteFunctionUrl = url.toString();
       }
+
+      // Async import removes direct dependency on hyperformula.
```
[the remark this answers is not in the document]
dmaretskyi: everything in this file will be loaded during composer startup. Hyperformula is not small, so we shouldn't block startup on it, especcially if the user is not opening a spreadsheet

packages/plugins/plugin-table/src/components/ObjectTable/ObjectTable.tsx
richburdon: why?

packages/sdk/client/src/echo/index.ts
richburdon: we're currently re-exporting it from echo-schema.
do you think we shoudldn't do that?

packages/plugins/plugin-sheet/src/graph/compute-node.ts
richburdon: no, for tables, etc. we'd just use ECHO queries, no?

packages/plugins/experimental/plugin-debug/src/components/ObjectCreator.tsx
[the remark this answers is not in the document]
richburdon: i would but seemed too complex to get working.

PR #7857, opened 2024-10-01
fix: plugin icons

richburdon:


packages/plugins/experimental/plugin-outliner/src/OutlinerPlugin.tsx
```diff
@@ -27,7 +27,6 @@ export const OutlinerPlugin = (): PluginDefinition<OutlinerPluginProvides> => {
           [TreeType.typename]: {
             placeholder: ['object placeholder', { ns: OUTLINER_PLUGIN }],
             icon: (props: IconProps) => <TreeStructure {...props} />,
-            iconSymbol: 'ph--tree-structure--regular',
```
wittjosiah, line 30: removed the wrong one?

packages/plugins/plugin-navtree/src/experimental/Tree.tsx
```diff
@@ -29,7 +29,7 @@ export const IconButton = ({
   // TODO(burdon): Density aware.
   return (
     <div className={mx('flex w-6 h-6 items-center justify-center select-none', classNames)} onClick={onClick}>
-      <Icon size={size} className='cursor-pointer' {...props} />
```
wittjosiah, line 32: this was correct?

richburdon: i decided to leave this alone since it's experimental and has the duotone issue. so i'll fix in another pass when we figure out if/how we want to handle that.

PR #7866, opened 2024-10-02
refactor: multiple mosaic roots

wittjosiah:
improves performance for resizing planks when stacks are open by
removing the global mosaic root for the mean time


thure (approved):
lgtm ✨

PR #7877, opened 2024-10-03
fix: comments type icon

wittjosiah:
- comments type icons shouldn't be the same as the object
- comments type icon menu shouldn't include all the objects actions


ZaymonFC (approved):
LGTM

PR #7883, opened 2024-10-04
feat(lit-grid): Intrinsic sizing & col/row limits

thure:
This PR implements col/row limits in `dx-grid` and handles intrinsic sizing on those axes in those cases.


ZaymonFC (approved):
shipit

PR #7885, opened 2024-10-06
fix: react-ui-editor deps

richburdon:


docs/content/guide/dxos-ui/snippets/vite-config.ts
```diff
@@ -14,6 +14,6 @@ export default defineConfig({
         resolve(__dirname, './index.html'),
         resolve(__dirname, './src/**/*.{js,ts,jsx,tsx}'),
       ],
-    }),
+    }) as Plugin,
```
wittjosiah, line 17: was does this need to be cast?

packages/ui/react-ui-editor/project.json
```diff
@@ -11,10 +11,12 @@
     "compile": {
       "options": {
         "entryPoints": [
-          "{projectRoot}/src/index.ts"
+          "{projectRoot}/src/index.ts",
+          "{projectRoot}/src/state/index.ts"
         ],
         "platforms": [
```
wittjosiah, line 17: both should be the default, so this can just be removed

PR #7887, opened 2024-10-06
fix: Fix effect deps

richburdon:
# Motivation
- consistency of using Effect as a peer dependency
- consistency of deps (remove strange build errors)
- place for Effect utils beyond echo-schema

# Summary
- new @dxos/effect package that exports: AST, JSONSchema, S
  - moved effect url util out of common/util
  - declared effect as peer dependency
- composer provides actual effect dependencies
- normalize all effect imports (AST, S, JSONSchema, etc.)
- changed "optionalDependenices" to "peerDependencies" (in about half of the plugins)

# Commentary/Rationale

I'm a bit uncomfortable re-exporting third-party packages: how far do we go?

However, the minimally viable set of deps to be fully functional with echo-schema are: `Schema (S), AST, and JSONSchema`. echo-schema itself depends on all of these, and so we should consider these as part of our API.

Some plugins like debug and table, require other effect types, and they will import these directly. Patterns may emerge where we create our own re-usable utils in dxos/effect.


packages/core/echo/echo-schema/package.json
```diff
@@ -43,17 +43,11 @@
...
     "ulidx": "^2.3.0"
   },
-  "devDependencies": {
-    "@effect/schema": "^0.67.16",
-    "effect": "^3.2.7"
-  },
-  "peerDependencies": {
-    "@effect/schema": "^0.67.16",
-    "effect": "^3.2.7"
-  },
```
wittjosiah, line 56: Since it's no longer bundled, the dev+peer dependency is correct. Previously it should rightly have been a full dependency

PR #7891, opened 2024-10-07
feat(tables): Prototype object table

ZaymonFC:
- **Rename table v2 -> table-next**
- **Integrate row and column length limits**
- **Wire in updating column sizing**
- **Fix axis resize event propagation in dx-grid**
- **Switch table to a mutable model (bounded by a reducer-ish update pattern)**


packages/plugins/plugin-table/src/next/components/Table.tsx
```diff
@@ -0,0 +1,55 @@
...
+          ref={gridRef}
+          limitRows={data.length}
+          limitColumns={table.columnDefinitions.length}
+          initialCells={gridCells}
+          columnDefault={{ size: 120, resizeable: true }}
+          rowDefault={{ size: 32, resizeable: true }}
+          columns={columnMeta}
+          onAxisResize={(event) => {
+            if (event.axis === 'col') {
+              const columnIndex = parseInt(event.index, 10);
```
richburdon, line 43: is the index the ordinal position (in which case the event type should be a number) or and ID, which needs to be mapped? Unclear why need to parse here.

packages/plugins/plugin-table/src/next/components/table.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { registerSignalRuntime } from '@dxos/echo-signals';
+import { type WithClientProviderProps, withClientProvider } from '@dxos/react-client/testing';
+import { withLayout, withTheme } from '@dxos/storybook-utils';
+
+import { Table } from './Table';
+import { type ColumnDefinition } from '../table';
+
+registerSignalRuntime();
+
+const makeData = (n: number) => {
```
richburdon, line 19: use @dxos/faker?

packages/plugins/plugin-table/src/next/components/table.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+      }
+    };
+
+    const intervalId = setInterval(mutateRandomCell, intervalMs);
+    return () => clearInterval(intervalId);
+  }, [data, columnDefinitions, intervalMs, isEnabled]);
+};
+
+const DefaultStory = () => {
+  const data = React.useMemo(() => makeData(100), []);
```
richburdon, line 76: Don't use `React.` directly (import useMemo, etc).

packages/plugins/plugin-table/src/next/table.ts
```diff
@@ -0,0 +1,150 @@
...
+      });
+    });
+
+    return cellMap;
+  });
+
+  const rowCount = computed(() => data.length);
+  const updateTracker = new CellUpdateTracker(cells);
+
+  return create({
```
richburdon, line 136: why are you not using ECHO objects directly? the data should be the result of a query and not require mapping. similarly for other state (see Sheet)?

packages/plugins/plugin-table/src/next/hooks/useTable.ts
```diff
@@ -0,0 +1,65 @@
...
+import { type RefObject, useCallback, useEffect, useMemo } from 'react';
+
+import { type DxGridElement, type DxGridAxisMeta, type DxGridCells } from '@dxos/react-ui-grid';
+
+import { type TableEvent, type ColumnDefinition, createTable, updateTable } from '../table';
+
+// TODO(Zan): Take ordering here (or order based on some stored property).
+// When the order changes, we should notify the consumer.
+export const useTable = (columnDefinitions: ColumnDefinition[], data: any[], gridRef: RefObject<DxGridElement>) => {
+  const table = useMemo(() => createTable(columnDefinitions, data), [columnDefinitions, data]);
```
richburdon, line 15: the seems to create a datastructure that should be passed in from external state.
i.e., data here should be the already reactive result of a query.

packages/plugins/plugin-table/src/next/table.ts
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * Performance characteristics:
+   * - Cell edits: Optimal (only affected cell recalculates)
+   * - Array-level changes: More expensive (all cell `computed` recreated)
+   */
+
+  const cells = computed(() => {
+    const cellMap = new Map();
+
+    data.forEach((row, rowIndex) => {
```
richburdon, line 120: why do we need to process all records here, rather than map them when required?

packages/plugins/plugin-table/src/next/components/table.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+import { create } from '@dxos/echo-schema';
+import { registerSignalRuntime } from '@dxos/echo-signals';
+import { type WithClientProviderProps, withClientProvider } from '@dxos/react-client/testing';
+import { withLayout, withTheme } from '@dxos/storybook-utils';
+
+import { Table } from './Table';
+import { type ColumnDefinition } from '../table';
+
+registerSignalRuntime();
```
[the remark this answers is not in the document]
ZaymonFC: Nice. Thanks!

packages/plugins/plugin-table/src/next/components/Table.tsx
```diff
@@ -0,0 +1,55 @@
...
+// NOTE: The table model manages both ephemeral and persistent state.
+// - Ephemeral state (e.g., sorting, filtering, selection) is handled internally.
+// - Persistent state (e.g., column order, widths) is propagated to the parent for storage.
+
+// TODO(Zan): Callback for changing column width.
+// TODO(Zan): Callback for re-arranging columns.
+// TODO(Zan): Callbacks for editing column schema.
+// TODO(Zan): Remove column axis labels.
+// TODO(Zan): Custom header labels and buttons.
+export const Table: React.FC<TableProps> = ({ columnDefinitions, data }) => {
```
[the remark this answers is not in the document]
ZaymonFC: I prefer your suggestion too.

packages/plugins/plugin-table/src/next/components/Table.tsx
ZaymonFC: I believe everything passed in and out of dx-grid is a string since it's a web component. The event contract currently has the type as a string and not a number.

I too would prefer just receiving a number for the ordinal position.

packages/plugins/plugin-table/src/next/table.ts
ZaymonFC: Thinking further on this, unless we move sorting into the query layer (which I would love) then we will always need to process the entire data-set.

packages/plugins/plugin-table/src/next/components/Table.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React, { useRef } from 'react';
+
+import { type DxGridElement, Grid } from '@dxos/react-ui-grid';
+
+import { useTable } from '../hooks';
+import { type ColumnDefinition } from '../table';
+
+type TableProps = {
+  columnDefinitions: ColumnDefinition[];
+  data: any[];
```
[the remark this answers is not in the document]
ZaymonFC: The generic component aspect of the old table was one of the pain points for me. Generics are great when you're instantiating against known types at build time, but get in the way when you're operating over arbitrary data at runtime.

PR #7892, opened 2024-10-07
refactor: attention

wittjosiah:
- nested attention providers to distinguish between different paths
- stop using focus-within to style attendables, instead use a data
attribute
- introduce styles for elements related to current attendable
- remove autofocus from text inputs was overactive and causing more
issues than it solved


packages/plugins/plugin-thread/src/components/CommentContainer.tsx
```diff
@@ -200,7 +190,13 @@ export const CommentContainer = ({
           onDelete={(id: string) => onMessageDelete?.(id)}
         />
       ))}
-      <MessageTextbox extensions={extensions} autoFocus={autoFocus} onSend={handleCreate} {...textboxMetadata} />
+      {/*
+        TODO(wittjosiah): Can't autofocus this generally.
+          There can be multiple threads with inputs and they can't all be focused.
+          Also, it steals focus from documents when first rendered.
+          Need to find a way to autofocus in one scenario only: when a new thread is created.
```
thure, line 197: 👌

packages/ui/react-ui-theme/src/styles/layers/attention.css
```diff
@@ -1,24 +1,29 @@
 @layer utilities {
+  /* If an element is attendable, it should have this class. */
```
thure, line 2:
```suggestion
  /* If an element should apply attention-related color tokens, it should use this class. */
```

packages/ui/react-ui-attention/src/components/AttentionProvider.tsx
```diff
@@ -6,42 +6,62 @@
...
+  return useMemo(() => {
+    const attributes: Record<string, string | undefined> = { 'data-attendable-id': attendableId };
+
+    if (hasAttention) {
+      attributes['data-is-attention-source'] = 'true';
+    }
+
+    return attributes;
+  }, [attendableId, hasAttention]);
 };
```
ZaymonFC, line 65: Nice.

ZaymonFC (approved):
Nothing jumped out at me, although it's hard to review PR's like this.

packages/ui/react-ui-theme/src/styles/layers/attention.css
```diff
@@ -1,24 +1,29 @@
...
-    --surface-separator: var(--dx-separator);
+  /* Used to style elements which are _related_ to the currently attended element. */
+  .current-related[aria-current] {
+    /*
+      NOTE(wittjosiah): Current these colors are the same, but they could be different.
+        If they are different, a background transition is likely required to avoid flashing.
+    */
+    --surface-bg: var(--dx-currentRelated);
+    &:focus-within {
+      --surface-bg: var(--dx-attention);
```
thure, line 26: Ah interesting… I was anticipating this should be a different selector or rule, but this actually seems more cogent, and if you’ve tested it and it works then this seems ideal.

thure (approved):
lgtm!

PR #7897, opened 2024-10-08
feat: edge notarization

yarolegovich:
### Details

* Added the first iteration of Edge http client to `edge-client`.
* Added a generalized `EdgeHttpResponse` to `protocols`.
* Extended `NotarizationPlugin` to work with edge.


packages/core/protocols/src/edge.ts
```diff
@@ -8,3 +8,36 @@ export enum EdgeService {
...
+export type EdgeHttpSuccess<T> = {
+  success: true;
+  data: T;
+};
+
+export type EdgeErrorData = { type: string } & Record<string, any>;
+
+export type EdgeHttpFailure = {
+  success: false;
+  reason: string;
```
dmaretskyi, line 21: what's the relation ship between `reason` and `errorData.type`?

dmaretskyi: perhaps worth documenting possible values in code

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
```diff
@@ -156,13 +205,32 @@ export class NotarizationPlugin implements CredentialProcessor {
...
+    credentials: Credential[],
+    timeouts: NotarizationTimeouts & { jitter?: number },
+  ) {
+    const encodedCredentials = credentials.map((credential) => {
+      const binary = credentialCodec.encode(credential);
+      return Buffer.from(binary).toString('base64');
+    });
+    scheduleTask(ctx, async () => {
+      try {
+        await client.post(`/spaces/${this._spaceId}/notarization`, {
```
dmaretskyi, line 222: im wondering if we could make every request it's own method on the client with apporpriate TS types

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
```diff
@@ -180,6 +248,39 @@ export class NotarizationPlugin implements CredentialProcessor {
   setWriter(writer: FeedWriter<Credential>) {
     invariant(!this._writer, 'Writer already set.');
     this._writer = writer;
+    if (this._edgeClient) {
+      this._notarizePendingEdgeCredentials(this._edgeClient, writer);
+    }
+  }
+
+  private _notarizePendingEdgeCredentials(client: EdgeHttpClient, writer: FeedWriter<Credential>) {
+    setTimeout(async () => {
```
dmaretskyi, line 257: `scheduleTask(this._ctx, ...)` (make this class a resource)

Every there's a naked setTimeout like this you're risking a race condition: the timeout gets scheduled, and then the class gets closed immediatelly after while the timeout is queued to run.

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
```diff
@@ -67,11 +88,27 @@ export class NotarizationPlugin implements CredentialProcessor {
...
+  }
+
   get hasWriter() {
     return !!this._writer;
   }
 
-  async open() {}
+  async open() {
+    if (this._edgeClient && this._writer) {
+      this._notarizePendingEdgeCredentials(this._edgeClient, this._writer);
```
dmaretskyi, line 109: Is it ok that this gets only called on `open`? Do we want to poll periodically in case there are any new notarization requests?

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
dmaretskyi: scheduleMicrotask even -- if you don't specifically need it to be a macro task

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
```diff
@@ -156,13 +205,32 @@ export class NotarizationPlugin implements CredentialProcessor {
...
+    scheduleTask(ctx, async () => {
+      try {
+        await client.post(`/spaces/${this._spaceId}/notarization`, {
+          body: { credentials: encodedCredentials },
+          retry: { count: MAX_EDGE_RETRIES, timeout: timeouts.retryTimeout, jitter: timeouts.jitter },
+        });
+
+        log('edge notarization success');
+
+        await sleep(timeouts.successDelay);
```
dmaretskyi, line 229: what's the reason for the sleep, this ivocation happens in a separate task and doesn't seem it be awaited

yarolegovich: copy-pasting artifact, removed

packages/sdk/client-services/src/packlets/spaces/notarization-plugin.ts
yarolegovich: Left an explanatory comment:
> The method is used only for adding agent feeds to spaces.
When an agent is created we can admit them into all the existing spaces. In case the operation fails
this method will fix it on the next space open.
Given how rarely this happens there's no need to poll the endpoint.

PR #7899, opened 2024-10-08
chore: Factor out common container for editor

richburdon:
First stage in unification of MarkdownEditor and DocumentEditor.

- co-locates extension creation
- tighten overgeneralized large function params by removing dead code
- reduces surface area of type definitions
- improve testing/stories

Next:

- simplify type handling further in MarkdownPlugin.
- simplify extension creation.
- single container for ECHO/non-ECHO and different surface locations (section, article, etc.)


packages/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -267,72 +243,51 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
...
-                  />
-                );
-              }
-              break;
-            }
+              /**
+               * Checks if an object conforms to the interface needed to render an editor.
+               */
+              // TODO(burdon): Remove need for this; intents should provide DocumentType.
+              const isEditorModel = (data: any): data is { id: string; text: string } =>
```
wittjosiah, line 254: I think this should go back where it was, no reason it needs to be recreated in the render cycle

wittjosiah: (same for functions below, they don't depend on any react state so no reason for them to be (re)defined inside render function)

packages/plugins/plugin-markdown/src/extensions.tsx
```diff
@@ -2,63 +2,117 @@
...
+  viewMode?: EditorViewMode;
 };
 
-/**
- * Create extension instances for editor.
- */
-export const createBaseExtensions = ({
-  viewMode,
-  settings,
+// TODO(burdon): Merge with createBaseExtensions above.
```
wittjosiah, line 40: below?

packages/plugins/plugin-markdown/src/extensions.tsx
```diff
@@ -155,25 +206,31 @@ const onRenderLink = (onSelectObject: (id: string) => void) => (el: Element, url
...
+        classNames={style.icon}
+      />
+    </a>,
   );
 };
 
 const renderLinkTooltip = (el: Element, url: string) => {
   const web = new URL(url);
-  createRoot(el).render(
-    <StrictMode>
```
wittjosiah, line 172: no more strict mode for these?

packages/plugins/plugin-markdown/src/extensions.tsx
```diff
@@ -155,25 +206,31 @@ const onRenderLink = (onSelectObject: (id: string) => void) => (el: Element, url
...
+    <a href={url} rel='noreferrer' target='_blank' className={style.hover}>
+      {web.origin}
+      <Icon icon='ph--arrow-square-out--bold' size={4} classNames={style.icon} />
+    </a>,
   );
 };
+
+// TODO(burdon): Factor out. Reconcile with react-ui-editor.
+const renderRoot = <T extends Element>(root: T, node: ReactNode): T => {
+  createRoot(root).render(<ThemeProvider tx={defaultTx}>{node}</ThemeProvider>);
```
wittjosiah, line 234: what's the difference between a new root and a portal for these things?

richburdon: portals gets removed from the DOM by CM.

packages/plugins/plugin-markdown/src/extensions.tsx
richburdon: temporary; these will be replaced by non-react DOM directly.

PR #7902, opened 2024-10-09
chore: small attention cleanup

wittjosiah:
- handle path id for space collections
- ensure sheet stays attended when input is focused


ZaymonFC (approved):
LGTM

PR #7903, opened 2024-10-09
feat: add object settings panel

wittjosiah:
Screenshot 2024-10-09 at 14 03 35
Screenshot 2024-10-09 at 14 03 45


packages/plugins/plugin-deck/src/components/DeckLayout/ComplementarySidebar.tsx
```diff
@@ -23,15 +30,57 @@ export type ComplementarySidebarProps = {
   flatDeck?: boolean;
 };
 
-export const ComplementarySidebar = ({ context, layoutParts, flatDeck }: ComplementarySidebarProps) => {
+const panels = ['comments', 'settings'] as const;
+type Panel = (typeof panels)[number];
```
ZaymonFC, line 34: 👍 Love this pattern.

packages/plugins/plugin-deck/src/components/DeckLayout/ComplementarySidebar.tsx
```diff
@@ -23,15 +30,57 @@ export type ComplementarySidebarProps = {
...
+const getPanel = (part?: string): Panel => {
+  if (part && panels.findIndex((panel) => panel === part) !== -1) {
+    return part as Panel;
+  } else {
+    return 'settings';
```
ZaymonFC, line 39: Consider `log/warning` for unknown value?

packages/ui/react-ui-deck/src/components/PlankHeading/PlankHeading.tsx
```diff
@@ -149,44 +156,52 @@ const PlankHeadingActionsMenu = forwardRef<HTMLButtonElement, PlankHeadingAction
...
-                  const shortcut =
-                    typeof action.properties.keyBinding === 'string'
-                      ? action.properties.keyBinding
-                      : action.properties.keyBinding?.[getHostPlatform()];
-
-                  const menuItemType = action.properties.menuItemType;
-                  const Root = menuItemType === 'toggle' ? DropdownMenu.CheckboxItem : DropdownMenu.Item;
-
+                {actionGroups?.map((actions, index) => {
+                  const separator = index > 0 ? <DropdownMenu.Separator /> : null;
```
ZaymonFC, line 160: Cheeky.

ZaymonFC (approved):
Looking good.

packages/plugins/plugin-deck/src/components/DeckLayout/ComplementarySidebar.tsx
wittjosiah: I'm specifically expecting this happen as an initial state in some cases because of cruft in the cached deck layout data so I think it's better for it to just fall through. If we find it harmful I wouldn't mind having a log there eventually, but would prefer to be conservative about what we're firing into Sentry for now.

PR #7923, opened 2024-10-14
feat: Add context deadlines

dmaretskyi:
### Motivation / Background

This Pull Request ...

### DIFF
--- a/packages/common/context/src/context-deadline-exceeded-error.ts
+++ b/packages/common/context/src/context-deadline-exceeded-error.ts
@@ -1,0 +1,9 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+export class ContextDeadlineExceededError extends Error {
+  constructor() {
+    super('Context deadline exceeded.');
+  }
+}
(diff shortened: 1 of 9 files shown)


packages/common/context/src/context-deadline-exceeded-error.ts
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+export class ContextDeadlineExceededError extends Error {
```
yarolegovich, line 5: nit: ContextTimedOutError?

packages/common/context/src/context.ts
```diff
@@ -111,16 +175,26 @@ export class Context {
    */
   onDispose(callback: DisposeCallback): () => void {
     if (this.#isDisposed) {
+      let reason = ContextDisposeReason.DISPOSED;
```
yarolegovich, line 178: should it we maybe instead save the actual reason when a context is disposed? I mean if a context gets disposed 1ms before deadline and then onDispose gets called 2 ms after, t will be invoked with DEADLINE_EXCEEDED, even though it was DISPOSED.

PR #7925, opened 2024-10-15
feat: Use `dx-grid` for `SheetPlugin`

thure:
Not currently working in Composer due to decorators.


packages/plugins/plugin-sheet/src/SheetPlugin.tsx
```diff
@@ -126,7 +126,7 @@ export const SheetPlugin = (): PluginDefinition<SheetPluginProvides> => {
             testId: 'sheetPlugin.createSection',
             type: ['plugin name', { ns: SHEET_PLUGIN }],
             label: ['create sheet section label', { ns: SHEET_PLUGIN }],
-            icon: (props: any) => <GridNine {...props} />,
+            icon: (props: any) => <Icon icon='ph--grid-nine--regular' {...props} />,
```
wittjosiah, line 129: todo just have string as property

ZaymonFC (approved):
Awesome work. Looks good to me! Glad the transition ended up being relatively straightforward apart from the decorator shenanigans.

PR #7935, opened 2024-10-16
feat(tables): Cut over to table/next

ZaymonFC:
There are still some remaining issues, however I'd like to get this merged in since I introduced some cross-cutting utilities that sheet will need to leverage.

Known issues (to be addressed ASAP in follow ups):

* ViewEditor Issues
  * Can't edit property values after adding, have to re-render
  * Order doesn't persist
  * Underlying schema should be modified when adding, deleting properties, changing property types
  

* Remaining Table Issues
  * Enter on last row should add a new row
  * New row button in the toolbar
  * Container should seal against the top and bottom of the viewport
  * Header row shouldn't be content editable
  * Actions button in header row
  * Initial size is not working column meta


packages/plugins/plugin-table/src/components/ObjectTable.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import React, { useCallback } from 'react';
+
+import { type EchoReactiveObject } from '@dxos/echo-schema';
+import { useGlobalFilteredObjects } from '@dxos/plugin-search';
+import { create, getSpace, useQuery, Filter } from '@dxos/react-client/echo';
+import { Button, Icon } from '@dxos/react-ui';
+
+import { Table } from './Table';
+import { useStarterTable } from '../hooks/useStarterTable';
```
richburdon, line 13: no deep imports

packages/plugins/plugin-table/src/components/ObjectTable.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const filteredObjects = useGlobalFilteredObjects(queriedObjects);
+
+  const handleNewRow = useCallback(() => {
+    if (!table.schema || !space) {
+      return;
+    }
+    space.db.add(create(table.schema, {}));
+  }, [table.schema, space]);
+
+  useStarterTable({ table, space });
```
richburdon, line 38: i think it's better to change this hook into a util (create default table) and then use an explicit useEffect here to set it (i.e., don't obfuscate that decision).

packages/plugins/plugin-table/src/components/ObjectTable.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  // TODO(Zan): Do we need this?
+  if (!table.schema) {
+    return <p>No schema</p>;
```
richburdon, line 42: remove -- return null.

packages/plugins/plugin-table/src/components/ObjectTable.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return (
+    <div className='border border-separator is-full max-is-max min-is-0 mli-auto'>
+      <Table table={table} data={filteredObjects} />
+      <Button classNames='w-full' onClick={() => handleNewRow()}>
```
richburdon, line 48: move to toolbar

packages/sdk/schema/src/formatting.ts
```diff
@@ -4,6 +4,83 @@
 
 import { FieldValueType } from './types';
 
+/**
+ * Parse value by field value type.
+ * Used by Table, Sheet, etc.
+ * Handles various data types including booleans, numbers, dates, and strings.
+ * Returns undefined for empty or null inputs.
+ */
+export const parseValue = (type: FieldValueType, value: any) => {
```
richburdon, line 13: this might be better in react-ui-data since it's UX?

packages/plugins/plugin-table/src/components/ObjectTable.tsx
ZaymonFC: Fair point.

packages/sdk/schema/src/formatting.ts
ZaymonFC: Yeah I was thinking that for the `cellClassesForFieldType` function as well.

packages/plugins/plugin-table/src/components/ObjectTable.tsx
ZaymonFC: Done.

PR #7936, opened 2024-10-16
chore: echo-schema/db API and refactor.

richburdon:
- Tighten echo-schema proxy.
- Moved EchoReactiveObject type to echo-db.
- Cleaned tests (don't over share instances across packages -- e.g., => echo-schema/testing). Leads to brittle tests and obfuscation.

## To discuss (DX and testing)
- Rename TypedObject (not an object)?
- Rename EchoReactiveObject?
- Reconcile ObjectInternals and Typed/UntypedReactiveHandler.ProxyTarget (e.g., signals)
- EchoReactiveObject remove prefix and reconcile with ProxyTarget
  - push ProxyTarget to echo-schema? Simplify getObjectCore?
- Remove schema/getType from API?
- Make createEchoObject internal only (can we deprecate use of this in tests/stories)?
  - Merge ReactiveObject and EchoReactiveObject
- Move PROPERTY_ID to sdk.
- dispose callbacks
- reconcile echo-schema/handler vs proxy
- documentation and API-level tests
  - need low-level schematic for Proxy, ObjectCore, ObjectInternals, etc.
  - esp. interop between echo-schema/echo-db
- figure out how to do something like packlets to PREVENT cross sub-folder deps
  - use `@internal` to prevent package exports! This prevents the need to cherry pick exports or do deep imports within a package?
- import from '@dxos/echo' (make self contained rather than `@dxos/react-client/echo`)
- Review effect API for cues
- CONSOLIDATE ALL data generators
  - plugin-debug/echo-generator
  - schema, react-ui-data
  - edge?


packages/plugins/plugin-markdown/src/MarkdownPlugin.tsx
```diff
@@ -52,7 +52,8 @@ export const MarkdownPlugin = (): PluginDefinition<MarkdownPluginProvides> => {
   const settings = new LocalStorageStore<MarkdownSettingsProps>(MARKDOWN_PLUGIN, {
     defaultViewMode: 'preview',
     toolbar: true,
-    folding: false,
+    numberedHeadings: true,
+    folding: true,
```
wittjosiah, line 56: I don't think this behaves well enough to be on by default

richburdon: example?

wittjosiah: - they don't sit next to the text so if the plank is wider than max text width they look broken
- they seem to render after the text does and shift it over on first load
- there was also the z-index issue with them, but I think that one was fixed?

richburdon: reverted since we're trying to make the initial experience fast and clean.

PR #7946, opened 2024-10-16
refactor: schedule effects

wittjosiah:


ZaymonFC (approved):
Nice! This makes sense to me.

packages/plugins/plugin-deck/src/DeckPlugin.tsx
```diff
@@ -283,23 +287,13 @@ export const DeckPlugin = ({
       window.addEventListener('popstate', handleNavigation);
```
richburdon, line 287: should plugins all extend @dxos/context/Resource to make unregistering simpler.
e.g.,

```ts
this._ctx.onDispose(Event.addEventListener(window, 'popstate', () => {});
```

richburdon (approved):

PR #7947, opened 2024-10-17
chore: echo-schema refactor

richburdon:


packages/core/echo/echo-schema/src/json/json-schema.ts
```diff
@@ -8,22 +8,24 @@ import { AST, JSONSchema, S } from '@dxos/effect';
...
 } from '../ast';
-import { createEchoReferenceSchema } from '../ref-annotation';
+import { createEchoReferenceSchema } from '../handler';
 
 const ECHO_REFINEMENT_KEY = '$echo';
+
 interface EchoRefinement {
-  type?: EchoObjectAnnotation;
-  reference?: EchoObjectAnnotation;
+  type?: ObjectAnnotation;
```
dmaretskyi, line 22: TypenameAnnotation?

packages/core/echo/echo-schema/src/ast/annotations.ts
```diff
@@ -8,9 +8,13 @@ import { type Simplify } from 'effect/Types';
 import { AST, S } from '@dxos/effect';
 
 import { checkIdNotPresentOnSchema } from './schema-validator';
-import { type Identifiable } from '../types';
 
-// TODO(burdon): Standardize names.
+/**
+ * Marker interface for object with an `id`.
+ */
+export interface HasId {
```
dmaretskyi, line 15: why in ast/annotations.ts?

Issue #7974, opened 2024-10-18
[tables] Add row toolbar button stopped working after merge


ZaymonFC:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'annotations')
    at @effect_schema.js?v=ca743a75:495:96
    at chunk-7TFQN3DK.js?v=45294875:50:18
    at getObjectAnnotation (annotations.ts:55:41)
    at createObject (object.ts:49:50)
    at create (object.ts:30:12)
    at TableContainer.tsx:46:26
    at onClick (Toolbar.tsx:57:24)
    at overrideProps.<computed> (chunk-MY2LENNB.js?v=45294875:139:11)
    at HTMLUnknownElement.callCallback2 (chunk-EP455A3S.js?v=45294875:3674:22)
    at HTMLUnknownElement.sentryWrapped (@sentry_browser.js?v=cafc60e6:7350:17)
    at Object.invokeGuardedCallbackDev (chunk-EP455A3S.js?v=45294875:3699:24)
    at invokeGuardedCallback (chunk-EP455A3S.js?v=45294875:3733:39)
    at invokeGuardedCallbackAndCatchFirstError (chunk-EP455A3S.js?v=45294875:3736:33)
    at executeDispatch (chunk-EP455A3S.js?v=45294875:7014:11)
    at processDispatchQueueItemsInOrder (chunk-EP455A3S.js?v=45294875:7034:15)
    at processDispatchQueue (chunk-EP455A3S.js?v=45294875:7043:13)
    at dispatchEventsForPlugins (chunk-EP455A3S.js?v=45294875:7051:11)
    at chunk-EP455A3S.js?v=45294875:7174:20
    at batchedUpdates$1 (chunk-EP455A3S.js?v=45294875:18881:20)
    at batchedUpdates (chunk-EP455A3S.js?v=45294875:3579:20)
    at dispatchEventForPluginEventSystem (chunk-EP455A3S.js?v=45294875:7173:11)
    at dispatchEventWithEnableCapturePhaseSelectiveHydrationWithoutDiscreteEventReplay (chunk-EP455A3S.js?v=45294875:5478:13)
    at dispatchEvent (chunk-EP455A3S.js?v=45294875:5472:13)
    at dispatchDiscreteEvent (chunk-EP455A3S.js?v=45294875:5449:13)
```

PR #7976, opened 2024-10-18
chore(tables): Increment

### LINKED ISSUE #7972: [tables] Cell editor positioning bug

ZaymonFC:
- Resolves #7972 
- Resolves https://github.com/dxos/dxos/issues/7970
- Resolves https://github.com/dxos/dxos/issues/7974
- Resolves https://github.com/dxos/dxos/issues/7980


packages/plugins/plugin-table/src/table-model.ts
```diff
@@ -70,15 +85,51 @@ export class TableModel extends Resource {
...
+
+      for (let displayIndex = 0; displayIndex < sorted.length; displayIndex++) {
+        const { index: dataIndex } = sorted[displayIndex];
+        if (displayIndex !== dataIndex) {
+          this.displayToDataIndex.set(displayIndex, dataIndex);
+        }
+      }
+
+      return sorted.map(({ item }) => item);
+    });
```
ZaymonFC, line 127: It might be better to add some metadata to the cells themselves indicating their original row index, but this works quite well for now.

packages/plugins/plugin-table/src/components/Table.tsx
```diff
@@ -2,37 +2,34 @@
...
-
 const blockEndLine =
   '[&>.dx-grid]:before:absolute [&>.dx-grid]:before:inset-inline-0 [&>.dx-grid]:before:-block-end-px [&>.dx-grid]:before:bs-px [&>.dx-grid]:before:bg-separator';
 
 const frozen = { frozenRowsStart: 1 };
 
-// NOTE: The table model manages both ephemeral and persistent state.
-// - Ephemeral state (e.g., sorting, filtering, selection) is handled internally.
-// - Persistent state (e.g., column order, widths) is propagated to the parent for storage.
+type TableProps = { table: TableType; data: any[] };
```
richburdon, line 26: i think data => rows?

richburdon: also should this move to react-ui-table next?
i guess we should address how existing apps like devtools should handle the transition?

packages/plugins/plugin-table/src/table-model.ts
```diff
@@ -33,6 +34,13 @@ export type TableModelProps = {
   rowSelection?: number[];
 };
 
+// TODO(Zan): Is there a better place for this to live?
+export const columnSettingsButtonAttr = 'data-table-column-settings-button';
+const columnSettingsButtonClasses = 'ch-button is-6 pli-0.5 min-bs-0 absolute inset-block-1 inline-end-2';
+const columnSettingsIcon = 'ph--caret-down--regular';
+const columnSettingsButtonHtml = (columnId: string) =>
```
richburdon, line 41: we prob want a util for this in react-ui (e.g., at least factoring out the svg).

packages/plugins/plugin-table/src/table-model.ts
```diff
@@ -43,10 +51,17 @@ export class TableModel extends Resource {
   public readonly table: TableType;
   public readonly data: any[];
   private onCellUpdate?: (col: number, row: number) => void;
-  public sorting: SortConfig[];
   public pinnedRows: { top: number[]; bottom: number[] };
   public rowSelection: number[];
```
richburdon, line 55: move to getters (since we shouldn't expose public setter?)

packages/plugins/plugin-table/src/table-model.ts
```diff
@@ -70,15 +85,51 @@ export class TableModel extends Resource {
...
+      }
+
+      const field = this.table.view?.fields.find((f) => f.id === sort.columnId);
+      if (!field) {
+        return this.data;
+      }
+
+      const dataWithIndices = this.data.map((item, index) => ({ item, index }));
+      const sorted = sortBy(dataWithIndices, [(wrapper) => wrapper.item[field.path]]);
+
```
richburdon, line 114: remove line (grouping or related statements)

richburdon (approved):

PR #7989, opened 2024-10-21
fix: ThemeProvider issue with multiple react-ui deps

richburdon:
# Post Mortem
- multiple instances of the ThemeProvider context were getting bundled such that non-deterministically, different components would access an uninstantiated context global variable (created by createContext) and then not be able to resolve classes. For example, Map buttons would remain unstyled with the fallback `icon` classname.

# Action
- changed `react-ui` and `react-ui-theme` to dev and peer dependencies across ALL non-app packages
- used `pnpm why react-ui` to verify that only composer-app's root package directly depends on it.
- created bundle to inspect that only a single copy of the context was generated across all chunks.
- remove default contexts (if a default does not make sense) and throw if not set.

```ts
export const useThemeContext = () => useContext(ThemeContext) ?? raise(new Error('Missing ThemeContext'));
```


packages/plugins/plugin-table/src/components/ColumnActionsMenu.tsx
```diff
@@ -18,7 +18,7 @@ type ColumnActionsMenuProps = {
   onOpenChange: (open: boolean) => void;
 };
 
-export const ColumnActionsMenu: React.FC<ColumnActionsMenuProps> = ({ tableModel, columnId, open, onOpenChange }) => {
+export const ColumnActionsMenu = ({ tableModel, columnId, open, onOpenChange }: ColumnActionsMenuProps) => {
```
thure, line 21: Thanks for this — I think this yields less lossy typing (should the type of `FC` or the implied type of what this returns ever change) and feels more intuitive to me.

packages/sdk/app-framework/src/plugins/SurfacePlugin/SurfaceDebug.tsx
```diff
@@ -0,0 +1,52 @@
...
+          }),
+      );
+    }
+  };
+
+  return (
+    <div className={mx('flex flex-col border border-separator overflow-hidden bg-modalSurface', classNames)}>
+      <div className='flex flex-col h-full w-full p-2'>
+        {surfaces.map(({ id, name, renderCount }) => (
+          <div key={id} className='grid grid-cols-[1fr_3rem_3rem] items-center text-xs font-mono whitespace-nowrap'>
```
thure, line 38: Yay grid! ✨

richburdon: have to learn sub-grid next!

packages/sdk/app-framework/package.json
```diff
@@ -36,10 +36,14 @@
...
   },
   "devDependencies": {
+    "@dxos/react-ui": "workspace:*",
+    "@dxos/react-ui-theme": "workspace:*",
     "@types/react": "~18.2.0",
     "react": "~18.2.0"
   },
   "peerDependencies": {
+    "@dxos/react-ui": "workspace:*",
+    "@dxos/react-ui-theme": "workspace:*",
```
wittjosiah, line 46: I've specifically been trying to avoid introducing these dependencies here. App framework shouldn't take any opinion on your UI toolchain.

packages/sdk/app-framework/src/plugins/SurfacePlugin/Surface.tsx
```diff
@@ -84,6 +87,20 @@ export type SurfaceProps = PropsWithChildren<{
...
+    const [id] = useState<string>(Math.random().toString(36).slice(2));
+    useEffect(() => {
+      debugInfo?.set(id, { id, created: Date.now(), name, role, renderCount: 0 });
+      return () => {
+        debugInfo?.delete(id);
+      };
+    }, [id]);
+    if (debugInfo?.get(id)) {
+      debugInfo.get(id)!.renderCount++;
+    }
```
wittjosiah, line 102: do you have a sense of how much overhead this adds when it's not being used?

wittjosiah: would react's `useId` work instead of creating a custom id?

docs/content/guide/dxos-ui/snippets/vite-config.ts
```diff
@@ -7,6 +7,7 @@ import { defineConfig } from 'vite';
 
 import { ThemePlugin } from '@dxos/react-ui-theme/plugin';
 
+// https://vitejs.dev/config
```
wittjosiah, line 10: I don't think this need to be included into the snippet for the docs

packages/apps/composer-app/vite.config.ts
```diff
@@ -52,12 +53,15 @@ export default defineConfig((env) => ({
   esbuild: {
     keepNames: true,
   },
+  // optimizeDeps: {
+  //   include: ['react', 'react-dom', '@dxos/react-ui', '@dxos/react-ui-theme'],
+  // },
```
wittjosiah, line 58: remove?

packages/sdk/app-framework/package.json
richburdon: moved to plugin-debug.

packages/sdk/app-framework/src/plugins/SurfacePlugin/Surface.tsx
richburdon: i think it should be unmeasurable. there's deliberately no state change or refresh.

docs/content/guide/dxos-ui/snippets/vite-config.ts
richburdon: i thought snippets were compiled to check they were legit?

richburdon: i've been fighting this for an hour:

```ts
      docs/content/guide/dxos-ui/snippets/vite-config.ts:13:5 - error TS2769: No overload matches this call.
        The last overload gave the following error.
          Type 'Plugin<any>' is not assignable to type 'PluginOption'
```

wittjosiah: they are, but they're also rendered on the docs site

richburdon: oh sorry, i thought you meant the import.

richburdon: NOTE: I had to update @types/node to "@types/node": "^22.7.7" to get the error to go away.

wittjosiah: ^ FYI @dmaretskyi

PR #7992, opened 2024-10-21
feat: edge http auth & edge replicator reconnect backoff

yarolegovich:
### Details

* Extracted EdgeIdentity for reuse between edge ws and http clients. 
* Added real challenge fetch before ws connection open. Browser doesn't give access to ws error response headers.
* Added edge http client 401 handling.
* Added retry backoff to automerge replicator.

### DIFF
--- a/packages/core/mesh/edge-client/src/edge-http-client.ts
+++ b/packages/core/mesh/edge-client/src/edge-http-client.ts
@@ -19,6 +19,7 @@ import {
   type GetAgentStatusResponseBody,
 } from '@dxos/protocols';
 
+import { type EdgeIdentity, handleAuthChallenge } from './edge-identity';
 import { getEdgeUrlWithProtocol } from './utils';
 
 const DEFAULT_RETRY_TIMEOUT = 1500;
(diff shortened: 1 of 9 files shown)


packages/core/mesh/edge-client/src/edge-identity.ts
```diff
@@ -0,0 +1,31 @@
...
+  identityKey: string;
+  /**
+   * Returns credential presentation issued by the identity key.
+   * Presentation must have the provided challenge.
+   * Presentation may include ServiceAccess credentials.
+   */
+  presentCredentials({ challenge }: { challenge: Uint8Array }): Promise<Presentation>;
+}
+
+export const handleAuthChallenge = async (failedResponse: Response, identity: EdgeIdentity): Promise<Uint8Array> => {
```
dmaretskyi, line 20: I suggest separating this into `extractChallenge` and `generateAuthPresentation(identity, challenge)`

packages/core/mesh/edge-client/src/edge-http-client.ts
```diff
@@ -107,14 +124,18 @@ export class EdgeHttpClient {
...
-    body: args.body && JSON.stringify(args.body),
-  };
-};
+  private async _handleUnauthorized(response: Response) {
+    if (!this._edgeIdentity) {
+      log.warn('edge unauthorized response received before identity was set');
+      throw EdgeCallFailedError.fromHttpFailure(response);
+    }
+    const challenge = await handleAuthChallenge(response, this._edgeIdentity);
+    this._authHeader = `VerifiablePresentation pb;base64,${Buffer.from(challenge).toString('base64')}`;
```
dmaretskyi, line 134: extact

dmaretskyi (approved):

packages/core/mesh/edge-client/src/edge-identity.ts
yarolegovich: If these two methods are **always** going to be called one after another, is there a point?
I'd do this if one of the methods needs to be reused in other context.

PR #7996, opened 2024-10-22
refactor: attention reactivity

wittjosiah:
- introduce AttentionManager which includes a reactivity structure
similar to app-graph
- the signals are segmented per path so that components only subscribe
to their own attention changing
- refactor some components in the deck layout to reduce unnecessary
re-renders, largely via memoization

Before:

https://github.com/user-attachments/assets/5eb7fa35-e62a-4404-9c64-982caa126da4

After:

https://github.com/user-attachments/assets/41bb2b37-b233-4142-9683-074dfa11204c


packages/plugins/plugin-deck/src/components/DeckLayout/NodePlankHeading.tsx
```diff
@@ -22,137 +19,138 @@ import { TextTooltip } from '@dxos/react-ui-text-tooltip';
...
+  ({
+    coordinate,
+    node,
+    canIncrementStart,
+    canIncrementEnd,
+    popoverAnchorId,
+    pending,
+    flatDeck,
+    actions = [],
+  }: {
```
richburdon, line 32: prefer separate type def.

packages/plugins/plugin-deck/src/components/DeckLayout/NodePlankHeading.tsx
```diff
@@ -22,137 +19,138 @@ import { TextTooltip } from '@dxos/react-ui-text-tooltip';
...
-                      state: false,
-                    },
-                  }
-                : {
-                    action: NavigationAction.CLOSE,
-                    data: {
-                      activeParts: {
-                        [layoutPart]: [id],
+            // TODO(Zan): Update this to use the new layout actions.
+            return dispatch(
```
richburdon, line 130: prefer

```
if () {
  dispatch();
} else {
  dispatch();
}
```

packages/ui/react-ui-attention/src/attention.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  get current() {
+    return this._state.current;
+  }
+
+  keys() {
+    return this._map.keys();
+  }
+
+  get(key: string[]): ReactiveObject<Attention> {
```
richburdon, line 38: getAttended (match other verbs)

packages/ui/react-ui-attention/src/attention.ts
```diff
@@ -0,0 +1,108 @@
...
+        return getAttendables(selector, trigger, acc);
+      }
+    } else {
+      acc.push(attendableId);
+      // (attempt tail recursion)
+      return !closestAttendable.parentElement ? acc : getAttendables(selector, closestAttendable.parentElement, acc);
+    }
+  }
+
+  return [...new Set(acc)];
```
richburdon, line 107: `return Array.from(acc.values())`

packages/ui/react-ui-attention/src/testing/decorators/withAttention.ts
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { type Decorator, type StoryFn } from '@storybook/react';
+import { createElement } from 'react';
+
+import { RootAttentionProvider } from '../../components';
+
+export const withAttention: Decorator = (Story: StoryFn) => {
```
richburdon, line 10: (hint don't need to type Story since Decorator types the entire function)

richburdon (approved):

packages/plugins/plugin-deck/src/components/DeckLayout/Plank.tsx
```diff
@@ -100,27 +119,17 @@ export const Plank = ({ entry, layoutParts, part, flatDeck, searchEnabled, layou
         {node ? (
           <>
             <NodePlankHeading
-              id={entry.id}
+              coordinate={coordinate}
               node={node}
-              layoutPart={coordinate.part}
-              layoutParts={layoutParts}
+              canIncrementStart={canIncrementStart}
+              canIncrementEnd={canIncrementEnd}
```
thure, line 125: Nice ✨

thure (approved):
Very nice ✨

packages/plugins/plugin-deck/src/components/DeckLayout/NodePlankHeading.tsx
wittjosiah: fair enough

packages/ui/react-ui-attention/src/attention.ts
wittjosiah: that doesn't do the same thing? set ensures that the array only has unique values

packages/ui/react-ui-attention/src/testing/decorators/withAttention.ts
thure: Better yet, `satisfies Decorator` at the end? Might not give story an inference perhaps though...

PR #7998, opened 2024-10-22
chore: Table refactor

richburdon:
- organize components into sub folders (move next to react-ui-table)
- factor out common types and utils with grid (e.g., GridCell)
- simplify handlers and seed constructors
- TODOs


packages/plugins/plugin-table/src/model/table-model.ts
```diff
@@ -86,7 +88,7 @@ export class TableModel extends Resource {
       return Object.fromEntries(
         fields.map((field, index: number) => {
           return [
-            getCellKey(index, 0),
+            fromCellKey({ col: index, row: 0 }),
```
ZaymonFC, line 91: I find this name confusing, it doesn't indicate what's being created.

packages/sdk/schema/src/types/types.ts
```diff
@@ -83,7 +83,7 @@ export const QuerySchema = S.Struct({
 
 // TODO(burdon): Are views always flat projections?
 export const ViewSchema = S.Struct({
-  query: QuerySchema,
+  query: S.optional(QuerySchema),
```
ZaymonFC, line 86: Nice, I've been wondering about this.

ZaymonFC (approved):
Looks good to me.

packages/plugins/plugin-table/src/model/table-model.ts
richburdon: ok to rename this but follow existing standards (see sheet) and make consistent (i.e., toX, fromX, etc.)

ZaymonFC: Works for me. Agree get was suboptimal.

ZaymonFC: A pattern I have used in the past is defining to and from as object properties. Then you can invoke them like:

`gridCellKey.fromCoordinates({ row, col })` or something like that.

PR #7999, opened 2024-10-22
refactor: Echo object subscription management

richburdon:


packages/core/echo/echo-db/src/core-db/object-core.ts
```diff
@@ -77,7 +80,9 @@ export class ObjectCore {
   /**
    * Handles link resolution as well as manual changes.
    */
-  public updates = new Event();
+  public readonly updates = new Event();
+
+  public readonly subscriptions: UnsubscribeCallback[] = [];
```
dmaretskyi, line 85: why not re-use `updates` event above?

packages/common/async/src/events.ts
```diff
@@ -73,6 +73,23 @@ export type ListenerOptions = {
...
 // TODO(burdon): Rename EventSink? (Event is a built-in type).
 export class Event<T = void> implements ReadOnlyEvent<T> {
+  /**
+   * Wraps `addEventListener` and return an unsubscription handler.
+   * @param target
+   * @param type
+   * @param listener
+   * @param options
+   */
+  static on<K extends keyof HTMLElementEventMap>(
```
dmaretskyi, line 83: consider making this a standalone function

packages/core/echo/echo-db/src/echo-handler/create.ts
```diff
@@ -40,36 +40,37 @@ export const isEchoObject = (value: unknown): value is EchoReactiveObject<any> =
...
-
-    // TODO(dmaretskyi): Does this need to be disposed? RB: Probably!
-    core.updates.on(() => target[symbolInternals].signal.notifyWrite());
-
     target[symbolPath] = [];
     target[symbolNamespace] = DATA_NAMESPACE;
     slot.handler._proxyMap.set(target, proxy);
 
+    // TODO(dmaretskyi): Does this need to be disposed?
+    core.subscriptions.push(core.updates.on(() => target[symbolInternals].signal.notifyWrite()));
```
dmaretskyi, line 72: I'd manage subscriptions on the proxy then (in ObjectInternals) -- this is only needed if we wanna drop the proxy without dropping the object-core, otherwise both objects could be garbage collected together with the callback still attached

packages/core/echo/echo-db/src/core-db/object-core.ts
dmaretskyi, line 85: see my other comment + I'd use Context here intead

packages/common/async/src/events.ts
richburdon: +1

PR #8010, opened 2024-10-22
fix(grid): Intrinsic sizing reactivity and refocus bug

ZaymonFC:
- **Update intrinsic sizing when column or row sizing prop changes**
- Resolves: https://github.com/dxos/dxos/issues/8011


thure (approved):
lgtm!

PR #8014, opened 2024-10-22
chore: Remove DensityProvider usages and make default fine.

richburdon:
- Update withVariants decorator for elevation/density.
- Fix foldGutter (jitter) and make folding on by default.


thure (approved):
lgtm!

PR #8035, opened 2024-10-24
refactor: Use DXNs in the filter protocol

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/common/keys/src/dxn.ts
```diff
@@ -47,10 +50,18 @@ export class DXN {
     return new DXN(kind, parts);
   }
 
+  static equals(a: DXN, b: DXN) {
+    return a.kind === b.kind && a.parts.length === b.parts.length && a.parts.every((part, i) => part === b.parts[i]);
+  }
+
   static isDXNString(dxn: string) {
```
richburdon, line 57: isDXN

packages/core/echo/echo-db/src/query/filter-match.ts
```diff
@@ -102,12 +90,7 @@ const filterMatchInner = (
   }
 
   if (filter.text !== undefined) {
-    const objectText = legacyGetTextForMatch(core);
-
-    const text = filter.text.toLowerCase();
-    if (!objectText.toLowerCase().includes(text)) {
-      return false;
-    }
+    log.warn('Text filter is not supported yet');
```
richburdon, line 93: throw?

packages/core/echo/echo-pipeline/src/db-host/query-state.ts
```diff
@@ -188,13 +188,27 @@ const filterToIndexQuery = (filter: FilterProto): IndexQuery => {
...
+              .filter(nonNullable),
       inverted: filter.not,
     };
   } else {
     // Query all objects.
     return { typenames: [] };
   }
 };
+
+const mapDxnToIndexQueryTypename = (dxn: DXN) => {
```
richburdon, line 205: just dxnToTypename?

richburdon (approved):

PR #8038, opened 2024-10-24
fix: Restore Thread-Sheet integration, fix Sheet model interactions

### LINKED ISSUE #7957: [sheet] Delete should clear the content of all cells within the selection

### LINKED ISSUE #7982: [sheet] Restore comments (as ranges)

### LINKED ISSUE #8024: [sheet] Model operations are buggy

thure:
This PR:
- resolves #7957
- resolves #8024
- resolves #7982

https://github.com/user-attachments/assets/150e3137-1f65-4785-8f82-93c1bc95e7db

https://github.com/user-attachments/assets/b2a5168a-7183-4a23-82cf-07c688a040b6


packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
```diff
@@ -0,0 +1,101 @@
...
+      });
+
+      if (closestThread) {
+        void dispatch([
+          { action: 'dxos.org/plugin/thread/action/select', data: { current: fullyQualifiedId(closestThread) } },
+        ]);
+      }
+    },
+    [dispatch, threads],
+  );
```
ZaymonFC, line 89: We used to actively highlight in the sidebar whichever thread anchor was closest to the cursor, not just 1:1 overlap. This worked like the editor. Now we only select in the sidebar if we overlap?

ZaymonFC (approved):
The code looks solid.

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
ZaymonFC: You could calculate distance to the range boundary fairly easily i'd say. It's a nice little sprinkle.

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
```diff
@@ -0,0 +1,101 @@
...
+          // TODO(Zan): Everywhere we refer to the cursor in a thread context should change to `anchor`.
+          const range = parseThreadAnchorAsCellRange(data.cursor);
+          range && grid.current?.setFocus({ ...range.to, plane: 'grid' }, true);
+        }
+      }
+    },
+    [model.sheet, setActiveRefs],
+  );
+
+  useIntentResolver(SHEET_PLUGIN, handleScrollIntoView);
```
richburdon, line 52: i think useIntentResolver should be removed since it obfuscates what's actually happening. i wrote this and it was intended to be short lived.

better to replace the callback above with:

```ts
  const { registerResolver } = useIntent();
  useEffect(() => {
    return registerResolver()
  }, []);
```

same pattern in the 3 other places.

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
```diff
@@ -0,0 +1,101 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { type MutableRefObject, useCallback, useEffect, useMemo } from 'react';
```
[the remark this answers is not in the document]
thure: I don’t prefer to have a hooks folder intentionally, see: https://discord.com/channels/837138313172353095/1110334757126017076/1299088207434219603

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
thure: Agreed this is potentially a helpful increment, worth filing as an additional issue?

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
thure: The scope of this PR might not be great for this refactor as I’m not familiar with this pattern and just moved it to a new file.

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import { useSheetContext } from '../components';
+import { type CellAddress, type CompleteCellRange, inRange } from '../defs';
+import { SHEET_PLUGIN } from '../meta';
+
+export const completeCellRangeToThreadCursor = (range: CompleteCellRange): string => {
+  return `${range.from.col},${range.from.row},${range.to.col},${range.to.row}`;
+};
+
+export const parseThreadAnchorAsCellRange = (cursor: string): CompleteCellRange | null => {
```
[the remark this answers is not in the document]
thure: I can’t find any logic in TablePlugin that deal with ranges — has this indeed been reimplemented somewhere?

packages/plugins/plugin-sheet/src/integrations/thread-ranges.ts
thure: I do think Sheet’s understanding of ranges (using `to` and `from`) are meaningfully distinct from `dx-grid`’s (using `start` and `end`), since either may or may not sort one coordinate relative to the other…

PR #8040, opened 2024-10-25
feat: Storybook toolbar to change logging level

richburdon:
- [x] Custom decorator.
- [x] Update node-std bundled deps.
- [x] Fix storybook exports.
- [x] react-ui stories now prefixed with react-ui-core.
- TODO: Stories using ClientProvider fail to unmount cleanly.

https://github.com/user-attachments/assets/f4d7ac34-1138-471f-97fc-677808275e0c


packages/apps/testbench-app/src/components/Main.tsx
```diff
@@ -24,19 +24,19 @@ import { exportData, importData } from '../util';
 
 export const Main = () => {
   const client = useClient();
-  // TODO(wittjosiah): Why filter out the default space?
+  // Filter default so that the first space visible is the shared space.
```
wittjosiah, line 27: Note that there's nothing currently stopping the default space from being shared as well

PR #8048, opened 2024-10-25
fix: Update View to remove query and add schema

richburdon:
- NOTE: ViewEditor current requires schemaResolver to get the scheme object from the view type. @dmaretskyi i assume we can't create a transformation since it would require the registry state?


ZaymonFC (approved):
Nice!

PR #8050, opened 2024-10-25
feat: Settings reset

richburdon:
- NOTE: Compute hyphenated settings key; breaking change for small number of props.


packages/common/local-storage/src/store.ts
```diff
@@ -116,6 +123,7 @@ export class LocalStorageStore<T extends object> {
     private readonly _prefix: string,
     defaults?: T,
   ) {
+    // TODO(burdon): Should this be externalized.
```
wittjosiah, line 126: what do you mean externalized?

wittjosiah (approved):

PR #8070, opened 2024-10-28
feat: WNFS plugin improvements

icidasset:
As mentioned in #6636, the WNFS plugin still needs some additional work.

- [x] Offline/online tracking
- [x] A queuing system for the block store. Push data to the remote Blob service when an internet connection is available, and keep track of which blobs didn't make it over the boundary.
- [x] Loading animations for the images.
- [ ] Make components in `src/components` actually do something.
- [ ] Ability to upload a file and associate it to an Echo object (`FileType` echo object)
- [x] Add debouncing to queue flushing to reduce load on Blob service (depends on **https://github.com/dxos/edge/issues/157)**

Not happening in this PR:
- Codemirror plugin rewrite (current one I wrote is 💩)


packages/plugins/experimental/plugin-wnfs/src/WnfsPlugin.tsx
```diff
@@ -58,3 +59,86 @@ export const WnfsPlugin = () =>
...
+            }
+          }
+
+          return null;
+        },
+      },
+    },
+  };
+};
+>>>>>>> a8315dbb39 (feat: Track blocks that haven't been stored remotely yet)
```
wittjosiah, line 144: I think these can just be dropped. This was refactored so all the bits and pieces of this are in other files now.

icidasset: So fast! Just noticed I forgot this merge conflict sorry. Should've tried to build after I did the rebase on main before I pinged you 😅 In any case, the above should be resolved now, thanks!

Still doesn't build though, I'm getting a storybook error for some reason:
```
packages/sdk/app-framework/src/playground/playground.stories.tsx:9:39 - error TS2307: Cannot find module '@dxos/storybook-utils' or its corresponding type declarations.

9 import { withLayout, withTheme } from '@dxos/storybook-utils';
```

wittjosiah: that's weird... those are installed at the repo root I think so they should be available for all packages, but could try adding them to app-framework to see if it resolves the issue?

icidasset: Adding `@dxos/storybook-utils` to `app-framework` seems to work 👍 Everything should work as it's supposed to now.

Issue #8019, opened 2024-10-23
[tables] View editor should change the associated schema property path / type


ZaymonFC:

ZaymonFC:
Resolved by #8093

PR #8073, opened 2024-10-29
feat(ViewEditor): Schema integration

### LINKED ISSUE #8006: [tables] Column deletion with undo

### LINKED ISSUE #8009: [tables] Column settings popover

ZaymonFC:
- Resolves #8006 
- Partial #8019
- Resolves #8009


packages/ui/react-ui-data/src/hooks/useForm.ts
```diff
@@ -4,54 +4,87 @@
...
 interface FormOptions<T> {
   initialValues: T;
-  validate: (values: any) => ValidationError[] | undefined;
+  schema?: S.Schema<T>;
+  /**
+   * Custom validation function that runs only after schema validation passes.
+   * Use this for complex validation logic that can't be expressed in the schema.
+   * @returns Array of validation errors, or undefined if validation passes
+   */
+  additionalValidation?: (values: T) => ValidationError[] | undefined;
```
dmaretskyi, line 26:
```suggestion
  validate?: (values: T) => ValidationError[] | undefined;
```

packages/sdk/schema/src/types/field.ts
```diff
@@ -61,6 +62,29 @@ export const toFieldValueType = (type: AST.AST): FieldValueType => {
   return FieldValueType.JSON;
 };
 
+// TODO(ZaymonFC): Find all the appropriate annotations.
```
dmaretskyi, line 65: let's establish how we're gonna store the schema.

For the discussion with @richburdon this morning, we want to remove `FieldType` in favour of storing everything in the [json-schema](https://json-schema.org/learn/getting-started-step-by-step) format.

There's an import disticution between the type schema and view schema:
Type schema defines the schema of the data stored in the database -- this is what the current `DynamicSchema`/`StoredSchema` is.
View schema defines the field config of the view which is backing the table.
Reusing the same storage format for both of those cases should make the code concceptually simpler.

For example:

packages/sdk/schema/src/types/view.ts
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import { type MutableSchema } from '@dxos/echo-schema';
+import { getProperty } from '@dxos/effect';
+import { invariant } from '@dxos/invariant';
+
+import { schemaForType } from './field';
+import { type FieldType, type ViewType } from './types';
+
+export const addFieldToView = (
```
dmaretskyi, line 12: I think this is right, you add a field to both view schema and type schema, but:

1. If view schema and type schema are in the same format, the code will look more similar.
2. I'm not sure about the current `schema.addFields` API as it doesn't allow for concurrent editing, perhaps we should just manipulate the JSONSchema stored in echo directly:

```ts
schema.jsonSchema.properties[newField] = { type: 'string', ... }
```

packages/ui/react-ui-data/src/hooks/useForm.ts
ZaymonFC: We use the provided schema for the first pass of validation, and if that passes then this function has a chance to do additional refinements so we need a name to reflect that.

PR #8082, opened 2024-10-29
fix(lit-grid): Virtualization & viewport snapping

### LINKED ISSUE #7959: [grid] `snapIntoPos` should apply to any cells that aren’t entirely within-view

### LINKED ISSUE #8016: [grid] `snapIntoPos` sometimes doesn’t act, causing the focused cell to be out of visible range

thure:
This PR:
- resolves #7959.
- resolves #8016.
- makes several improvements to grid’s virtualization implementation

https://github.com/user-attachments/assets/406572fd-f931-4c19-9e98-654abd8d9880


packages/ui/lit-grid/src/dx-grid.ts
```diff
@@ -894,32 +900,32 @@ export class DxGrid extends LitElement {
...
-      ? this.focusedCell.col - (this.visColMin + minDelta)
-      : this.focusedCell.col >= this.visColMax - maxDelta
-        ? -(this.focusedCell.col - this.visColMax - maxDelta)
+  private focusedCellColOutOfVis() {
+    return this.focusedCell.col <= this.visColMin
+      ? this.focusedCell.col - this.visColMin - 1
+      : this.focusedCell.col >= this.visColMax - 1
+        ? this.focusedCell.col + 2 - this.visColMax
         : 0;
   }
```
ZaymonFC, line 917: So many magic numbers.

ZaymonFC (approved):
Nice work! As things crystallize I wouldn't mind trying to make some nice ascii diagrams for how some of the internals work.

packages/ui/lit-grid/src/dx-grid.ts
ZaymonFC: 🧙

thure: I know… the value here was consistently +2, probably due to two stacks of values that are “off by one” from the value this is meant to return.

PR #8085, opened 2024-10-30
feat!: Remove CJS build from packages

dmaretskyi:
### Motivation / Background

This Pull Request ...


toolbox.json
wittjosiah, line 1: I believe lit grid also has custom exports, we've had to revert toolbox changes there before because it exports a css file like shell

wittjosiah (approved):
should there also be a change to esbuild to stop generating the cjs versions?

PR #8093, opened 2024-10-31
feat: ECHO dynamic schema overhaul

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/sdk/schema/src/types/view.ts
```diff
@@ -2,33 +2,83 @@
...
+  S.nonEmptyString({ message: () => 'Property is required.' }),
+  S.pattern(/^[a-zA-Z_$][\w$]*(?:\.[a-zA-Z_$][\w$]*)*$/, { message: () => 'Invalid property path.' }),
+);
+
+/**
+ * Stored field metadata.
+ */
+export const FieldSchema = S.mutable(
+  S.Struct({
+    path: PathSchema,
```
dmaretskyi, line 21: seems extraneos to have path both on `field` and `schema` parts of the view. perhaps fields in the array should just reference schema fields by id?

richburdon (approved):
LGTM

PR #8105, opened 2024-10-31
fix(react-ui-list): Fix rearranging when `getId` isn't specified

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/8104

Fixes rearranging of list items when no `getId` function is provided by implementing a more robust equality check. The issue occurred when comparing drag target data where object references differed despite having the same content.

Previously, the comparison relied on direct reference equality. Now falls back to JSON string comparison when `getId` isn't specified.

### DIFF
--- a/packages/sdk/schema/src/types/types.ts
+++ b/packages/sdk/schema/src/types/types.ts
@@ -60,7 +60,7 @@ export const FieldSchema = S.mutable(
     // TODO(ZaymonFC): Should validations be common? Think about translations?
     path: S.String.pipe(
       S.nonEmptyString({ message: () => 'Property is required.' }),
-      S.pattern(/^\w+$/, { message: () => 'Invalid property name.' }),
+      S.pattern(/^[a-zA-Z_$][\w$]*(?:\.[a-zA-Z_$][\w$]*)*$/, { message: () => 'Invalid property path.' }),
     ),
 
     // TODO(burdon): Replace with annotations?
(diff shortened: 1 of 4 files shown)


packages/ui/react-ui-list/src/components/List/ListRoot.tsx
```diff
@@ -45,7 +45,22 @@ export const ListRoot = <T extends ListItemRecord>({
   onMove,
   ...props
 }: ListRootProps<T>) => {
-  const isEqual = useCallback((a: T, b: T) => (getId ? getId(a) === getId(b) : a === b), [getId]);
+  const isEqual = useCallback(
```
richburdon, line 48: move out of function to static funciton.

packages/sdk/schema/src/types/types.ts
```diff
@@ -60,7 +60,7 @@ export const FieldSchema = S.mutable(
     // TODO(ZaymonFC): Should validations be common? Think about translations?
     path: S.String.pipe(
       S.nonEmptyString({ message: () => 'Property is required.' }),
-      S.pattern(/^\w+$/, { message: () => 'Invalid property name.' }),
+      S.pattern(/^[a-zA-Z_$][\w$]*(?:\.[a-zA-Z_$][\w$]*)*$/, { message: () => 'Invalid property path.' }),
```
ZaymonFC, line 63: Pattern for valid js object paths concatenated with `.`.

packages/ui/react-ui-list/src/components/List/ListRoot.tsx
richburdon: but i don't think this works. can't we just rely on value equivalence?

richburdon (changes requested):

richburdon (approved):

PR #8120, opened 2024-11-01
feat(plugin-sheet): Range management

### LINKED ISSUE #8103: [spreadsheet] Ranges panel

thure:
This PR:
- resolves #8103.

https://github.com/user-attachments/assets/b67ea75b-e44a-4233-9647-ea392e5532eb


packages/plugins/plugin-sheet/src/SheetPlugin.tsx
```diff
@@ -144,22 +144,19 @@ export const SheetPlugin = (): PluginDefinition<SheetPluginProvides> => {
...
-          const graph = useComputeGraph(space);
-          const handleSelectRange = () => {};
-          const handleDeleteRange = () => {};
 
           switch (role) {
             case 'article':
             case 'section':
-              if (graph && data.object instanceof SheetType) {
-                return <SheetContainer graph={graph} sheet={data.object} role={role} />;
+              if (space && data.object instanceof SheetType) {
```
richburdon, line 151: maybe return early if no space (since other things could technically work, but we're def in an invalid state)?

packages/plugins/plugin-sheet/src/components/RangeList/RangeList.tsx
```diff
@@ -2,37 +2,51 @@
...
 export type RangeListProps = {
   sheet: SheetType;
-  onSelect?: (range: Range) => void;
-  onDelete?: (range: Range) => void;
 };
 
-export const RangeList = ({ sheet, onSelect, onDelete }: RangeListProps) => {
+export const RangeList = ({ sheet }: RangeListProps) => {
+  const { t } = useTranslation(SHEET_PLUGIN);
+  const handleSelectRange = (range: Range) => {};
```
richburdon, line 22: TODO to back-prop this to the grid to show the currently active range? I think this is like comments select.

richburdon (approved):

packages/plugins/plugin-sheet/src/components/RangeList/RangeList.tsx
thure: Good idea, will leave a comment and make an issue.

PR #8133, opened 2024-11-05
feat(plugin-sheet): Text wrap style

### LINKED ISSUE #8101: [spreadsheet] Soft wrapping formatting

thure:
This pr:
- resolves #8101
- fixes a variety of issues around Sheet’s Toolbar.


packages/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -214,7 +219,7 @@ const Alignment = () => {
             ({ range, key }) => key === alignKey && inRange(rangeFromIndex(model.sheet, range), cursor),
           )?.value
         : undefined,
-    [cursor, model.sheet.ranges],
```
wittjosiah, line 217: this doesn't need to be recreated if the ranges change?

packages/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
```diff
@@ -235,12 +240,17 @@ const Alignment = () => {
...
+const styleOptions: ButtonProps<StyleValue>[] = [
+  { value: 'highlight', icon: 'ph--highlighter--regular' },
+  { value: 'softwrap', icon: 'ph--paragraph--regular' },
+];
 
 const Styles = () => {
   const { cursor, model } = useSheetContext();
   const { onAction } = useToolbarContext('Styles');
   const { t } = useTranslation(SHEET_PLUGIN);
+  // TODO(thure): Why is neither model.sheet nor model.sheet.ranges not usable as a memo dependency?
```
wittjosiah, line 252: What happens without this?

packages/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
thure: I think I was trying to figure out reactivity here, will revert and create a new issue to address toolbar reactivity.

packages/plugins/plugin-sheet/src/components/Toolbar/Toolbar.tsx
thure: The toolbar item isn’t reactive even though the range is added.

thure: May just remove the `useMemo`…

PR #8134, opened 2024-11-06
feat: Programmatic grid selection

### LINKED ISSUE #8099: [spreadsheet] Entire column & row selection

thure:
This PR:
- resolves #8099
- adjusts how Sheet’s Toolbar resolves ranges

https://github.com/user-attachments/assets/00344e70-e209-4dd6-9e64-2e36d5440beb


thure:
> @thure noticing that the horizontal selection changes don't override the vertical one, is that intentional?

@wittjosiah Yes, there are two ranges depicted here, so modifying/deleting the horizontal range leaves the vertical range unaffected. The cell at the intersection gets the text alignment style determined by precedence among ranges (currently most recent added has precedence), otherwise falling back to the default for its format.

wittjosiah:
> currently most recent added has precedence

this doesn't seem to be the case though because the cell at the intersection doesn't update in the video based on the horizontal range being applied

thure:
The cell at the intersection of the ranges is C7 — it centers correctly when a range for that is applied on its column, then is unresponsive as designed when I apply and un-apply centering on its row, since its column’s text alignment range still applies.

ZaymonFC (approved):
LGTM.

PR #8140, opened 2024-11-07
fix: edge-client concurrency issues

yarolegovich:
### Details

Some of the invitation tests were flaky because of the delayed edge-client state changes, which were all performed in DeferredTask. For example:
```ts
client.setIdentity(newIdentity);
client.send(message);
```
Could send the message with the new identity on the old connection if identity was changed after there was already an active connection.

Fixed reconnect timeout for the case when websocket error was received during connection attempt. Previously we would wait for ready trigger or timeout, blocking the restart. Now we race with restart-requested trigger to unblock persistent lifecycle and schedule a restart immediately when requested.

Extracted websocket management into a separate `EdgeWsConnection` class.
EdgeClient now manages restarts, identity and notification dispatch.
EdgeWsConnection contains heartbeat logic and websocket callbacks. 

Replaced connected and reconnected events with a method `onReconnected` which also invokes a listener if it was added after a connection was already established.
It was decided to call it `onReconnected` to make is clear that cleanup of the previous state is needed.


packages/core/mesh/edge-client/src/edge-client.ts
```diff
@@ -98,13 +94,32 @@ export class EdgeClient extends Resource implements EdgeConnection {
...
 
-  public addListener(listener: MessageListener): () => void {
-    this._listeners.add(listener);
-    return () => this._listeners.delete(listener);
+  public onMessage(listener: MessageListener): () => void {
+    this._messageListeners.add(listener);
+    return () => this._messageListeners.delete(listener);
+  }
+
+  public onReconnected(listener: () => void): () => void {
```
dmaretskyi, line 107: I would call this `onConnected` or `withConnection` and add a brief jsdoc to explain the behaviour. The name `onReconnected` suggests that the callback will not be called if it is already connected

packages/core/mesh/edge-client/src/edge-client.ts
```diff
@@ -122,97 +137,96 @@ export class EdgeClient extends Resource implements EdgeConnection {
...
-    // TODO(dmaretskyi): Potential leak: context re-assigned without disposing the previous one.
-    this._keepaliveCtx = new Context();
-    scheduleTaskInterval(
-      this._keepaliveCtx,
-      async () => {
-        // TODO(mykola): use RFC6455 ping/pong once implemented in the browser?
-        // Cloudflare's worker responds to this `without interrupting hibernation`. https://developers.cloudflare.com/durable-objects/api/websockets/#setwebsocketautoresponse
-        this._ws?.send('__ping__');
+        },
+        onRestartRequired: () => {
```
dmaretskyi, line 172: `onClose` is better I think -- callback names should descrive the event happening in the class defining the callback, not what the implementation does

packages/core/mesh/edge-client/src/edge-ws-connection.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export type EdgeWsConnectionCallbacks = {
+  onConnected: () => void;
+  onMessage: (message: Message) => void;
+  onRestartRequired: () => void;
+};
+
+export class EdgeWsConnection extends Resource {
+  private _inactivityTimeoutCtx: Context | undefined;
+  private _ws: WebSocket | undefined;
```
dmaretskyi, line 28: readonly

packages/core/mesh/edge-client/src/edge-ws-connection.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  constructor(
+    private readonly _identity: EdgeIdentity,
+    private readonly _connectionInfo: { url: URL; protocolHeader?: string },
```
dmaretskyi, line 32: I'd move the header formatting here as well and pass in the encoded credentail - seems like header format is closely releated to the websocket protocol

PR #8141, opened 2024-11-07
refactor: Schema TS types

dmaretskyi:
### Motivation / Background

This Pull Request ...


packages/core/echo/echo-schema/src/formats/types.ts
```diff
@@ -21,7 +23,7 @@ export type ScalarType =
   | JSONSchema.JsonSchema7Ref;
 
 // TODO(burdon): Ref.
-export const getTypeEnum = (property: JSONSchema.JsonSchema7): TypeEnum | undefined => {
+export const getTypeEnum = (property: JsonSchemaType): TypeEnum | undefined => {
```
richburdon, line 26: should we detect refs here?

packages/sdk/schema/src/types/format.ts
```diff
@@ -26,16 +28,25 @@ const extend = (format: FormatEnum, type: TypeEnum, fields = {}) =>
     }),
   ).pipe(S.mutable);
 
+interface FormatSchemaCommon extends BaseProperty {
```
richburdon, line 31: can't this be imputed from the union?
e.g., if we add new format variants we have to manually create this mixin? (it's not "common" it's the "union").

richburdon (approved):

PR #8143, opened 2024-11-07
feat(plugin-script): add settings panel

wittjosiah:
Screenshot 2024-11-07 at 10 29 21


packages/plugins/experimental/plugin-script/src/components/ScriptSettingsPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+      spaceId: space?.id,
+    });
+
+  // TODO(wittjosiah): Use ClipboardProvider.
+  const handleCopy = useCallback(() => {
+    functionUrl && navigator.clipboard.writeText(functionUrl);
+  }, [functionUrl]);
+
+  return (
+    <div role='form' className='flex flex-col w-full p-2 gap-4'>
```
richburdon, line 60: use DeprecatedFormInput?
(in common with other settings so that we can easily replace these all later?)

richburdon (approved):

packages/plugins/experimental/plugin-script/src/components/ScriptSettingsPanel.tsx
wittjosiah: that lays things out in a way that doesn't work for the sidebar

PR #8146, opened 2024-11-07
refactor: provide panels via plugins

wittjosiah:


packages/plugins/experimental/plugin-automation/src/AutomationPlugin.tsx
```diff
@@ -37,6 +37,15 @@ export const AutomationPlugin = (): PluginDefinition<AutomationPluginProvides> =
       echo: {
         schema: [ChainType, ChainPromptType, FunctionDef, FunctionTrigger],
       },
+      complementary: {
+        panels: [
+          {
+            id: 'automation',
+            label: ['open automation panel label', { ns: AUTOMATION_PLUGIN }],
+            icon: 'ph--flow-arrow--regular',
```
wittjosiah, line 45: calling out that I changed this icon, let me know if you think it should be a different one

PR #8162, opened 2024-11-08
feat(plugin-navtree): open branches when direct linked to an object

### LINKED ISSUE #7481: [navtree] open branches when direct linked to an object
`graph.getPath` isn’t working even after `graph.waitForNode` because it isn’t connected/expanded yet.

Determine whether/how this should work, since linking and refreshing are difficult to distinguish.

wittjosiah:
Resolves #7481

### DIFF
--- a/packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
+++ b/packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
@@ -18,6 +18,7 @@ import {
   type GraphProvides,
   parseGraphPlugin,
   NavigationAction,
+  parseNavigationPlugin,
 } from '@dxos/app-framework';
 import { createExtension, type Graph, isAction, isGraphNode, type Node, type NodeFilter } from '@dxos/app-graph';
 import { invariant } from '@dxos/invariant';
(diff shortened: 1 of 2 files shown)


packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -130,18 +131,21 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
...
+
       graphPlugin = resolvePlugin(plugins, parseGraphPlugin);
       graph = graphPlugin?.provides.graph;
       if (!graph) {
         return;
       }
 
-      state.values.root = graph.root as NavTreeItemGraphNode;
-      getChildren(graph, state.values.root);
-      getActions(graph, state.values.root);
```
wittjosiah, line 141: these were old cruft, they used to have side-effects like "expand" call below but no longer do so they were doing nothing

packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -130,18 +131,21 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
   return {
     meta,
     ready: async (plugins) => {
+      state.prop({ key: 'open', type: LocalStorageStore.json<string[]>() });
```
wittjosiah, line 134: this needs to be first so that state is restored from storage before anything acts on it

Issue #7984, opened 2024-10-19
[tables] support ref column / cells


ZaymonFC:

thure:
Depends on #7962

PR #8170, opened 2024-11-09
feat: Cell references

richburdon:
- Resolves https://github.com/dxos/dxos/issues/7984
- Resolves https://github.com/dxos/dxos/issues/7968

## Tasks
- select items by reference prop
- backspace deletes cell value
- callback to create new row

## Notes
- Table imperative handle to avoid setters on model.
- Tests are currently disabled since plugin-table has deps on other plugins. Factor out to react-ui-table (and deprecate existing).
- added typedefs, generics, push down refs, etc. to reduce deps.
- standardize column => field

## Demo
https://github.com/user-attachments/assets/163f6d2f-f92a-4ff2-8b4f-96fe68d27905


packages/ui/react-ui-grid/src/CellEditor/CellEditor.tsx
```diff
@@ -174,7 +175,10 @@ export const CellEditor = ({
     <div
       ref={parentRef}
       className={editorVariants[variant].root}
-      style={box}
+      style={{
+        ...box,
+        ...{ '--dx-gridCellWidth': box ? `${box.inlineSize - 1}px` : undefined },
```
thure, line 180: Why -1?

packages/ui/react-ui-grid/src/Grid/Grid.tsx
```diff
@@ -109,7 +112,9 @@ const GridContent = forwardRef<NaturalDxGrid, GridScopedProps<GridContentProps>>
   const { id, editing, setEditBox, setEditing } = useGridContext(GRID_CONTENT_NAME, props.__gridScope);
   const [dxGrid, setDxGridInternal] = useState<NaturalDxGrid | null>(null);
 
-  // NOTE(thure): using `useState` instead of `useRef` works with refs provided by `@lit/react` and gives us a reliable dependency for `useEffect` whereas `useLayoutEffect` does not guarantee the element will be defined.
+  // TODO(burdon): Can we use useImperativeHandle here?
```
thure, line 115: I believe I demonstrated that `useImperativeHandle` causes a regression to the original issue; consumers still need to use `useState` to get a reliable effect dependency.

thure (approved):
The grid bits look fine to me, just left a question and a remark. The other business superficially looks fine to me, though I'm less familiar and would defer to someone more familiar.

packages/ui/react-ui-card/project.json
```diff
@@ -14,7 +14,8 @@
           "{projectRoot}/src/index.ts"
         ],
         "platforms": [
-          "browser"
+          "browser",
+          "node"
         ]
```
wittjosiah, line 19: this is the default, just remove

packages/ui/react-ui-data/project.json
```diff
@@ -14,7 +14,8 @@
           "{projectRoot}/src/index.ts"
         ],
         "platforms": [
-          "browser"
+          "browser",
+          "node"
         ]
```
wittjosiah, line 19: remove

packages/ui/react-ui-list/project.json
```diff
@@ -14,7 +14,8 @@
           "{projectRoot}/src/index.ts"
         ],
         "platforms": [
-          "browser"
+          "browser",
+          "node"
         ]
       }
```
wittjosiah, line 20: remove

packages/sdk/schema/src/types/view.ts
```diff
@@ -8,6 +8,9 @@ import { create, JsonProp, JsonSchemaType, QueryType, type ReactiveObject, S, Ty
  * Stored field metadata (e.g., for UX).
  */
 export const FieldSchema = S.Struct({
+  // TODO(burdon): We should add a fieldId distinct from the property since in principle we might have
```
wittjosiah, line 11: could the fieldId just be the property & format concatenated together?

packages/plugins/plugin-table/src/components/TableContainer.tsx
```diff
@@ -39,7 +39,14 @@ const TableContainer = ({ role, table }: LayoutContainerProps<{ table: TableType
   const queriedObjects = useQuery(space, schema ? Filter.schema(schema) : () => false, undefined, [schema]);
   const filteredObjects = useGlobalFilteredObjects(queriedObjects);
 
-  const handleDeleteRow = useCallback((row: any) => space?.db.remove(row), [space]);
+  // TODO(burdon): Undo action.
```
wittjosiah, line 42: use space plugin intent instead of directly removing, it has undo

wittjosiah (approved):
How does reference validation work? If it's being done via typeahead rather than a select input, what happens if you stop typing before there's a valid value? Does the cell get cleared on blur or set with an invalid state somehow?

packages/ui/react-ui-grid/src/CellEditor/CellEditor.tsx
richburdon: removed.

packages/ui/react-ui-grid/src/Grid/Grid.tsx
richburdon: let's discuss today.

richburdon:
> How does reference validation work? If it's being done via typeahead rather than a select input, what happens if you stop typing before there's a valid value? Does the cell get cleared on blur or set with an invalid state somehow?

yes, that's what would happen.
as a next step i'm going to add "create new record" as an option also.
and we need to consider how to nav to the other table.

packages/plugins/plugin-table/src/components/TableContainer.tsx
richburdon: @ZaymonFC

packages/ui/react-ui-grid/src/Grid/Grid.tsx
thure: Even if this method can be made to be equivalent, what value does it deliver exactly?

packages/plugins/plugin-table/src/components/Table/Table.tsx
```diff
@@ -25,32 +35,98 @@ const blockEndLine =
...
+          if (cell) {
+            dxGrid.updateIfWithinBounds(cell);
+          } else {
+            dxGrid.updateCells(true);
+          }
+        },
+      };
+    },
+    [model, dxGrid],
+  );
```
ZaymonFC, line 97: Nice!

ZaymonFC (approved):
Looking good!

packages/plugins/plugin-table/src/model/table-model.ts
```diff
@@ -6,108 +6,124 @@ import { computed, effect, signal, type ReadonlySignal } from '@preact/signals-c
...
   pinnedRows?: { top: number[]; bottom: number[] };
   rowSelection?: number[];
-  onDeleteColumn?: (field: string) => void;
-  onDeleteRow?: (row: any) => void;
   onInsertRow?: (index?: number) => void;
+  onDeleteRow?: (row: number, object: T) => void;
+  onDeleteColumn?: (fieldId: string) => void;
   onCellUpdate?: (cell: GridCell) => void;
   onRowOrderChanged?: () => void;
 };
```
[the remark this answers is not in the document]
richburdon: all of the devtools use tables with definite types, so in that case i think it's useful.

PR #8173, opened 2024-11-10
chore: imperative handle to simplify grid reference.

richburdon:


thure:
⚠️ This must be tested on labs before merging to be sure it doesn't regress back to the original issue.

I think I might still prefer the state setter because it gives React components a clear dependency to use in effects.

thure:
This PR does not work when deployed and regresses to refs being unstable. Closing this for now, please reopen if you feel it is important to pursue this.

https://github.com/user-attachments/assets/626b4817-870a-469d-8555-bf61d1eaa2bb

PR #8180, opened 2024-11-11
refactor: disable the stack view of collections by default

wittjosiah:
currently in draft as this should not be merged until there's an alternative way of accessing space settings, otherwise this is ready for review


packages/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -637,6 +646,7 @@ export const SpacePlugin = ({
                   .map((space) =>
                     constructSpaceNode({
                       space,
+                      selectable: state.values.selectableCollections,
```
thure, line 649: `navigable` perhaps? I still feel we use the term ‘select’ for too many orthogonal interactions…

thure (approved):
I think this could really mitigate a lot of confusion for new users, let me know if I can help find another way to surface space settings.

Issue #8027, opened 2024-10-24
[tables] Selection (checkbox column)


ZaymonFC:

ZaymonFC:
Support multi-delete when there are many rows selected.

PR #8183, opened 2024-11-12
chore(tables): Refactor modal state management

ZaymonFC:
- In service of #8027

Previously, menu/modal interactions in the table were managed by the `useTableMenuController` hook, which lived outside the model and required complex state synchronization. This change moves all modal/menu state management into the model layer where it belongs.

## Rationale

- Click handling and its corresponding state changes were split between the hook and model, making the interaction flow hard to follow
- Future interactive elements (checkboxes, toggles) would require additional coordination between the hook and model
- Having interaction logic outside the model made it harder to maintain a single source of truth

## Changes

- Introduced `ModalController` class to encapsulate menu/modal state management
- Removed `useTableMenuController` hook
- Modal state and trigger refs now live in the model layer
- Click handling unified under the model's `handleGridClick`
- Simplified component layer to just read and interact with modal state

## Architecture

The new architecture follows a clearer separation of concerns:

- Model layer owns all state and interaction handling
- Components are purely presentational, reading state from model
- Click handling flows through a single path in the model

## Future Work

This refactor sets us up for:

- Adding new interactive cell types (checkboxes, toggles, etc) without needing external hooks
- Cleaner integration of future modal/popup features
- Better testability as all interaction logic is contained in the model
- Potential to add interaction middleware


packages/plugins/plugin-table/src/components/Table/ColumnActionsMenu.tsx
```diff
@@ -2,50 +2,37 @@
...
-
-export const ColumnActionsMenu = ({
-  triggerRef,
-  model,
-  fieldId,
-  open,
-  onOpenChange,
-  onShowColumnSettings,
-}: ColumnActionsMenuProps) => {
+export const ColumnActionsMenu = ({ model }: { model?: TableModel }) => {
```
richburdon, line 12: always prefer an exported type definition as a standard.
model prob shouldn't be optional; i don't think we're consistent about this, but i'd try to make things non-optional if the component wouldn't function without it (different for hooks, since we might not have one available and you can't have optional hooks). callbacks should all be optional.

packages/plugins/plugin-table/src/components/Table/Table.tsx
```diff
@@ -173,8 +172,6 @@ const TableMain = forwardRef<TableController, TableMainProps>(({ model }, forwar
     [model],
   );
 
-  const { state: menuState, triggerRef, handleClick, close, showColumnSettings } = useTableMenuController();
-
   return (
     <>
       {/* TODO(burdon): Is this required to be unique? */}
```
richburdon, line 177: remove TODO?

packages/plugins/plugin-table/src/components/Table/Table.tsx
```diff
@@ -192,36 +189,15 @@ const TableMain = forwardRef<TableController, TableMainProps>(({ model }, forwar
           limitRows={model?.getRowCount() ?? 0}
           limitColumns={model?.table.view?.fields?.length ?? 0}
```
richburdon, line 190: i think bette to extract this to a local variable: `numColumns`.
(Q: doesn't this need to be +1 for the fixed column?)

packages/plugins/plugin-table/src/model/modal-controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public get trigger() {
+    return this._triggerRef as RefObject<HTMLButtonElement>;
+  }
+
+  public handleClick = (event: MouseEvent): boolean => {
+    const target = event.target as HTMLElement;
+
+    const columnButton = target.closest(`button[${tableButtons.columnSettings.attr}]`);
```
richburdon, line 33: perhaps keep these in order `row, column, columnSettings, close` and match the variable name to the state  (`newColumnButton => columnSettingsButton`)

packages/plugins/plugin-table/src/model/modal-controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  public close = () => {
+    this._state.value = { type: 'closed' };
+  };
+
+  public openColumnSettings = () => {
+    if (this._state.value.type === 'column') {
+      const fieldId = this._state.value.fieldId;
+      // Let the first modal close completely
```
richburdon, line 74: period.

packages/plugins/plugin-table/src/model/table-model.ts
```diff
@@ -398,6 +409,15 @@ export class TableModel<T extends BaseTableRow = {}> extends Resource {
     }
   }
 
+  //
+  // Interactions
+  //
+  public handleGridClick = (event: React.MouseEvent): void => {
```
richburdon, line 415: i don't think this is right. the model shouldn't be directly involved in the UX. (in principle the model could be used by a terminal app.) otherwise it'll blur the line of responsibility. i think the table should create and manage the ModalController separately, just like it does to manage the toolbar. splitting this also reduces the exposure of model to other components like RowActionMenu.

(also, all `React` defs should be imported in the `react` import above.)

richburdon (changes requested):

packages/plugins/plugin-table/src/model/table-model.ts
ZaymonFC: I'm going to push back on this.

1. Table model is already tied intimately with grid and constructing grid cells.
2. Table model currently wires up and renders the accessory html for each modal trigger button including setting the attributes.
3. Table model is about to become even more integrated with the UX, when we start rendering checkboxes for selection, toggles for Boolean cells, better triggers for refs at best, and tag lists in multi-select cells. We need a standard way of managing interactions and reacting to each of these being clicked since they have stateful changes and I think it should live in the model, unless we workout a way to abstract the entire presentational layer (which I think we should do all at once if we want that).

ZaymonFC: fwiw one of the reasons I made it a separate class is so that presentation can get abstracted later.

packages/plugins/plugin-table/src/components/Table/Table.tsx
ZaymonFC: I think the limit rows and columns is for the central grid plane.

PR #8193, opened 2024-11-12
chore: clean-up echo-schema

richburdon:


packages/core/echo/echo-schema/src/handler/ref.ts
```diff
@@ -14,7 +14,28 @@ import { type Ref } from '../types';
 /**
  * The `$id` field for an ECHO reference schema.
  */
-export const JSON_SCHEMA_ECHO_REF_ID = '/schemas/echo/ref';
+const JSON_SCHEMA_ECHO_REF_ID = '/schemas/echo/ref';
+
+const DXN = /dxn:type:(.+)/;
```
dmaretskyi, line 19: `DXN.isDXNString`

packages/core/echo/echo-schema/src/handler/ref.ts
```diff
@@ -14,7 +14,28 @@ import { type Ref } from '../types';
...
+
+export const getSchemaReference = (property: any): string | undefined => {
+  const { $id, reference: { schema: { $ref } = {} } = {} } = property;
+  if ($id === JSON_SCHEMA_ECHO_REF_ID) {
+    const match = $ref?.match(DXN);
+    return match?.[1];
+  }
+};
+
+export const setSchemaReference = (property: any, schema: string) => {
```
dmaretskyi, line 29:
```suggestion
export const setSchemaReference = (property: JsonSchemaType, schema: string) => {
```

packages/core/echo/echo-schema/src/handler/ref.ts
```diff
@@ -14,7 +14,28 @@ import { type Ref } from '../types';
...
+  Object.assign(property, {
```
dmaretskyi, line 30: This is not right, because this will leave any other fields on `property` intact.

e.g. 
```ts
const property = { type: 'number' };
Object.assign(property, { $id, reference })
property === { type: 'number', $id, reference }
```

i would instead re-assign the whole object: `schema.properties[key] = { $id, reference }`

packages/core/functions/src/types.ts
```diff
@@ -12,56 +16,48 @@ import { RawObject, S, TypedObject } from '@dxos/echo-schema';
...
-        deep: S.optional(S.Boolean),
-        // Debounce changes (delay in ms).
-        delay: S.optional(S.Number),
-      }),
-    ),
-  }),
-);
+const SubscriptionTriggerSchema = S.Struct({
+  type: S.Literal('subscription'),
+  // TODO(burdon): Define query DSL (from ECHO).
```
dmaretskyi, line 21: Use query DSL (from ECHO)

packages/sdk/schema/src/types/projection.ts
```diff
@@ -150,33 +135,21 @@ export class ViewProjection {
...
-      if (isRename) {
-        delete this._schema.jsonSchema.properties![sourcePropertyName!];
-      }
-
-      let $id;
-      let reference;
+      const jsonProperty: JsonSchemaType = {};
       if (referenceSchema) {
-        $id = JSON_SCHEMA_ECHO_REF_ID;
+        setSchemaReference(jsonProperty, referenceSchema);
```
dmaretskyi, line 140: Using the like this nullifies my previous comment, but still makes this function a poor abstraction, perhaps it would be better to do: `Object.assign(jsonProperty, getSchemaReference(referenceSchema))`

dmaretskyi (commented):
,

packages/core/functions/src/types.ts
richburdon: where is that defined -- need to chat about this. also needed for triggers.

packages/sdk/schema/src/types/projection.ts
richburdon: let's discuss i'm not sure about this.

PR #8194, opened 2024-11-13
chore(view): Limit number of fields in a view

ZaymonFC:


packages/plugins/plugin-table/src/util/table-buttons.ts
```diff
@@ -16,9 +16,17 @@ const ICONS: { [K in TableButton]: string } = {
...
+  value = '',
+  disabled = false,
+}: {
+  button: TableButton;
+  value?: string;
+  disabled?: boolean;
+}): string => {
   const buttonClasses = 'ch-button is-6 pli-0.5 min-bs-0 absolute inset-block-1 inline-end-2' as const;
-  return `<button class="${buttonClasses}" ${TABLE_ATTRS[button]}="${value}"><svg><use href="/icons.svg#${ICONS[button]}"/></svg></button>`;
+  return `<button class="${buttonClasses}" ${TABLE_ATTRS[button]}="${value}" ${disabled ? 'disabled' : ''}><svg><use href="/icons.svg#${ICONS[button]}"/></svg></button>`;
```
thure, line 29: Oof, I really want to avoid rendering strings for this, but maybe it’s inevitable for now… I know `@lit/react` supports binding to `<slot/>`s, I just don’t know if slots can repeat… to investigate certainly.

thure (approved):
lgtm!

packages/plugins/plugin-table/src/util/table-buttons.ts
ZaymonFC: 💯

PR #8200, opened 2024-11-13
feat: icon-button

richburdon:
We need to standardize our use of forms/buttons i'm suggesting with utils like this.

## Next
- dialogs (heading and button bar)
- forms

## Example

Screenshot 2024-11-13 at 5 55 27 PM


packages/ui/react-ui/src/components/Buttons/IconButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+// Copyright 2024 DXOS.org
+//
+
+import React from 'react';
+
+import { Button, type ButtonProps } from './Button';
+import { useThemeContext } from '../../hooks';
+import { Icon, type IconProps } from '../Icon';
+
+type IconButtonProps = Omit<ButtonProps, 'children'> & Pick<IconProps, 'icon'> & { label?: string };
```
thure, line 11:
```suggestion
type IconButtonProps = Omit<ButtonProps, 'children'> & Pick<IconProps, 'icon'> & { label: string };
```
Label must be required, it is user-hostile to exclude it. If a developer disagrees, they can provide an empty string.

packages/ui/react-ui/src/components/Buttons/IconButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const IconButton = ({ icon, label, ...props }: IconButtonProps) => {
+  const { tx } = useThemeContext();
+  return (
+    <Button {...props} classNames={tx('iconButton.root', 'iconButton')}>
+      <Icon icon={icon} />
+      {label && <span>{label}</span>}
```
thure, line 18:
```suggestion
      {label && <span className='sr-only'>{label}</span>}
```
I assume this is what you meant to use?

packages/ui/react-ui-data/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -102,22 +102,23 @@ export const ViewEditor = ({
...
-                  {view.fields.length >= VIEW_FIELD_LIMIT ? t('field limit reached') : t('add field')} <Tooltip.Arrow />
-                </Tooltip.Content>
-              </Tooltip.Portal>
-            </Tooltip.Root>
-          </div>
-        </div>
+        <ButtonGroup classNames='p-2 justify-center'>
+          <Tooltip.Root>
+            <Tooltip.Trigger asChild>
+              <IconButton
```
thure, line 108: Why use `ButtonGroup` if there is just the one button? I can see the argument that it will scale to accommodate more buttons certainly, just curious.

packages/ui/react-ui-data/src/components/Form/Form.tsx
```diff
@@ -92,18 +92,16 @@ export const Form = <T extends object>({
...
-            <Icon icon='ph--x--regular' />
-          </Button>
-        </div>
-      </div>
+      <ButtonGroup classNames='justify-center'>
+        {!readonly && <IconButton icon='ph--x--regular' label={'Cancel'} onClick={onCancel} />}
+        <IconButton
+          icon='ph--check--regular'
+          type='submit'
+          label={'Save'}
```
thure:
```suggestion
          label={t('save label')}
```

thure (changes requested):
Label shouldn’t be optional, otherwise just a few questions & suggestions.

wittjosiah:
> Label shouldn’t be optional

agree, though it would probably be good to be able to have icon-only buttons. In that case I would expect that the label be shifted to a tooltip

packages/ui/react-ui-data/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -102,22 +102,23 @@ export const ViewEditor = ({
...
+                icon='ph--plus--regular'
+                label={t('button add property')}
+                onClick={handleAdd}
+                disabled={view.fields.length >= VIEW_FIELD_LIMIT}
+              />
+            </Tooltip.Trigger>
+            <Tooltip.Portal>
+              <Tooltip.Content classNames='z-[12]'>
+                {view.fields.length >= VIEW_FIELD_LIMIT ? t('field limit reached') : t('add field')} <Tooltip.Arrow />
+              </Tooltip.Content>
```
thure, line 118: Why should the `label` be different from the tooltip? I’d assume they should align. If you agree, should `IconButton` simply always render a tooltip?

richburdon: i don't like the tooltips but yes we shouldn't require the developer to do this.

packages/ui/react-ui/src/components/Buttons/IconButton.tsx
richburdon: but this would make it invisible?

thure: If this should be an option, perhaps provide that as a flag property of `IconButton`

packages/ui/react-ui-data/src/components/ViewEditor/ViewEditor.tsx
thure: *always render a tooltip _when it sets `sr-only` on its in-flow label_?

wittjosiah: the tooltips are not great, but I think not having them is even worse because then you're just left guessing what the icon means. We should work on the tooltips rather than removing them

packages/ui/react-ui/src/components/Buttons/IconButton.tsx
thure: > the intention is in the image above
But I see about as many instances of icon-only buttons, and based on the name of the component I’d assumed it would have that capability. You gave this moose a muffin and he is asking for some jam 😅

packages/ui/react-ui-data/src/components/ViewEditor/ViewEditor.tsx
thure: > We should work on the tooltips rather than removing them

Yes, please let’s not leave icon-only buttons totally unlabeled. If tooltips are a blocker, let’s prioritize getting that right rather than giving end-users headaches. We have a path forward with `virtualRef` poppers.

richburdon: Agreed, but in this case the limit should be displayed as a caption inside the form.

packages/ui/react-ui-data/src/translations.ts
```diff
@@ -16,6 +16,12 @@ export default [
         'delete field': 'Delete field',
         'field limit reached': 'Maximum number of fields reached',
 
+        'button add property': 'Add property',
+
+        // TODO(burdon): Factor out?
```
thure, line 21: A lot of common translations are in the `'os'` namespace defined in the shell, which has `'cancel label'` but currently no ‘save’.

thure (approved):
Made adjustments so this could be used much more broadly.

PR #8210, opened 2024-11-14
chore: Unused dependencies

ZaymonFC:
This isn't an exhaustive cleanup, but I've rectified the worst offenders in `packages/plugins` and `packages/ui`.


ZaymonFC:
Too out of date now. Although we should revisit this at some point.

PR #8211, opened 2024-11-14
refactor(composer-app): explode shell

wittjosiah:
- Stops using turnkey sdk shell and uses shell components within the plugins
- this is a necessary followup to #8180 to expose space settings
- it will also prevent sdk code from being loaded twice due to the removal of the shell iframe in composer


packages/ui/react-ui-list/src/components/Tree/TreeItemToggle.tsx
```diff
@@ -19,6 +19,7 @@ export const TreeItemToggle = memo(
       <Button
         ref={forwardedRef}
         data-testid='treeItem.toggle'
+        aria-expanded={open}
```
thure, line 22: Thank you!

packages/plugins/plugin-space/src/components/SpaceSettings/SpaceSettingsDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Tabs.Viewport classNames='flex-1 min-bs-0'>
+          <div role='none' className='overflow-y-auto pli-3 @md:pis-2 @md:pie-0 mbe-4 border-r border-separator'>
+            <Tabs.Tablist classNames='flex flex-col max-bs-none min-is-[200px] gap-4 overflow-y-auto'>
+              <div role='none' className='flex flex-col ml-1'>
+                <Tabs.Tab value='settings'>{t('settings tab label')}</Tabs.Tab>
+                <Tabs.Tab value='members' disabled={locked}>
+                  {t('members tab label')}
+                </Tabs.Tab>
+              </div>
+            </Tabs.Tablist>
```
thure, line 86: I think we may need to iterate on this, but that can come in a follow-up.

thure (approved):
looks good! lesdoit!

packages/ui/react-ui-list/src/components/Tree/TreeItemToggle.tsx
wittjosiah: needed this for the e2e tests :)

PR #8218, opened 2024-11-14
feat(view): Editable typename

ZaymonFC:


packages/core/echo/echo-db/src/echo-handler/create.ts
```diff
@@ -140,7 +140,7 @@ const validateSchema = (schema: S.Schema<any>) => {
   SchemaValidator.validateSchema(schema);
 };
 
-const saveTypeInAutomerge = (internals: ObjectInternals, schema: S.Schema<any> | undefined) => {
+const setTypeOnObject = (internals: ObjectInternals, schema: S.Schema<any> | undefined) => {
```
richburdon, line 143: setObjectType

packages/core/echo/echo-db/src/proxy-db/mutable-schema-registry.ts
```diff
@@ -166,9 +166,21 @@ export class MutableSchemaRegistry implements SchemaResolver {
       return existing;
     }
 
+    let previousTypename: string | undefined;
+
     const mutableSchema = new MutableSchema(schema);
     const subscription = getObjectCore(schema).updates.on(() => {
       mutableSchema.invalidate();
+
+      if (previousTypename != undefined && schema.typename !== previousTypename) {
```
richburdon, line 175: I think != null? (i don't love the != syntax, but I've only seen it used with null -- to mean both null and undefined)
can we avoid the check here with just schema.typename !== previousTypename?

packages/core/echo/echo-schema/src/mutable/manipulation.ts
```diff
@@ -50,3 +52,13 @@ export const updateFieldNameInSchema = (
     ),
   );
 };
+
+export const updateTypenameInSchema = (schema: S.Schema<any>, typename: string): S.Schema<any> => {
```
richburdon, line 56: I see you following the convention here, but we should find another way to namespace these methods otherwise this suffix is becoming clumsy.

But it should be `setTypename(InSchema)` not update i think.

@dmaretskyi should we consider namespaces for packages with lots of public definitions? If these are internal, then they should be marked `@internal`.

packages/core/echo/echo-schema/src/mutable/manipulation.ts
```diff
@@ -50,3 +52,13 @@ export const updateFieldNameInSchema = (
...
+  const existingAnnotation = schema.ast.annotations[ObjectAnnotationId] as ObjectAnnotation;
+  invariant(existingAnnotation, "Schema doesn't have ObjectAnnotation");
```
richburdon, line 58: `Missing ${ObjectAnnotationId}`

all errors messages should be as terse (and consistent) as is reasonable. Just enough to give the missing runtime information since the static code context will be apparent).

packages/core/echo/echo-schema/src/mutable/mutable-schema.test.ts
```diff
@@ -91,4 +91,46 @@ describe('dynamic schema', () => {
...
+  test('updates typename', async ({ expect }) => {
+    // Create schema with some fields and annotations
+    const registered = createMutableSchema(EmptySchemaType);
+    const originalVersion = registered.storedSchema.version;
+    registered.addFields({
+      field1: S.String.pipe(PropertyMeta('test', { maxLength: 10 })),
+      field2: S.Number,
+    });
+
+    // First update
```
richburdon, line 104: period.

packages/core/echo/echo-schema/src/mutable/mutable-schema.test.ts
```diff
@@ -91,4 +91,46 @@ describe('dynamic schema', () => {
     });
     expect(getPropertyMetaAnnotation(registered.getProperties()[0], metaNamespace)).to.deep.eq(metaInfo);
   });
+
+  test('updates typename', async ({ expect }) => {
+    // Create schema with some fields and annotations
+    const registered = createMutableSchema(EmptySchemaType);
+    const originalVersion = registered.storedSchema.version;
+    registered.addFields({
+      field1: S.String.pipe(PropertyMeta('test', { maxLength: 10 })),
```
richburdon, line 100: name, age might be better to help comprehend.
ideally re-use existing test use cases.
contact, org, etc.

richburdon (approved):

richburdon (approved):

PR #8221, opened 2024-11-15
feat(tables): Selection

### LINKED ISSUE #8027: [tables] Selection (checkbox column)

ZaymonFC:
- Resolves #8027 
- Supports bulk deletion
- Selection is durable through sort and row order changes, insertions and deletions etc.

https://github.com/user-attachments/assets/b279a2cf-a49f-4953-9e9e-b1a787f6f0f9

### DIFF
--- a/packages/plugins/plugin-space/src/meta.ts
+++ b/packages/plugins/plugin-space/src/meta.ts
@@ -25,7 +25,7 @@ export enum SpaceAction {
   CLOSE = `${SPACE_ACTION}/close`,
   MIGRATE = `${SPACE_ACTION}/migrate`,
   ADD_OBJECT = `${SPACE_ACTION}/add-object`,
-  REMOVE_OBJECT = `${SPACE_ACTION}/remove-object`,
+  REMOVE_OBJECTS = `${SPACE_ACTION}/remove-objects`,
   RENAME_OBJECT = `${SPACE_ACTION}/rename-object`,
   DUPLICATE_OBJECT = `${SPACE_ACTION}/duplicate-object`,
   WAIT_FOR_OBJECT = `${SPACE_ACTION}/wait-for-object`,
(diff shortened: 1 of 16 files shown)


packages/plugins/plugin-space/src/meta.ts
```diff
@@ -25,7 +25,7 @@ export enum SpaceAction {
   CLOSE = `${SPACE_ACTION}/close`,
   MIGRATE = `${SPACE_ACTION}/migrate`,
   ADD_OBJECT = `${SPACE_ACTION}/add-object`,
```
wittjosiah, line 27: todo: add objects

packages/plugins/plugin-space/src/translations.ts
```diff
@@ -47,6 +47,7 @@ export default [
         'delete object label': 'Delete item',
         'collection deleted label': 'Collection deleted',
         'object deleted label': 'Item deleted',
+        'objects deleted label': 'Items deleted',
```
wittjosiah, line 50: see https://github.com/dxos/dxos/blob/da7e0b3e001c14e9c78cb47add1c95ab01a5c651/packages/plugins/plugin-space/src/translations.ts#L70-L72

ZaymonFC: Hmm, none of the `{{count}}` translations I could find seem to have usages. How do I supply the params?

wittjosiah: https://www.i18next.com/translation-function/plurals

ZaymonFC: https://github.com/dxos/dxos/blob/a597fa5190f0fe057687eecb8d5bfea610eda224/packages/plugins/plugin-space/src/SpacePlugin.tsx#L1347-L1359

Hmmm.

wittjosiah: Ah yeah that would be problematic, just TODO I guess

PR #8264, opened 2024-11-20
chore(tables): Basic e2e tests

ZaymonFC:
image


packages/ui/react-ui-table/src/playwright/smoke.spec.ts
```diff
@@ -0,0 +1,123 @@
+// Copyright 2021 DXOS.org
```
wittjosiah, line 1: suspect this will fail lint

wittjosiah (approved):
I'd suggest adding a TODO for factoring out either helper functions and/or manager class for table similar to other playwright tests as I find it helps w/ readability of the tests (and also reusability when going up a layer to composer). This looks like a great start though!

Maybe also worth triggering e2e tests to run on this branch in CI before merging to make sure they're passing there.

PR #8283, opened 2024-11-24
test: Schema generator utils.

richburdon:
## Data generators using annotations and effect pipeline
- [x] See `sdk/schema/generator.test.ts`
- [x] Create references 

## Refactors
- [x] `T extends object/{}` => `T extends BaseObject` everywhere (with optional `= {}` not `any`).
- [x] `EchoReativeObject` => `ReactiveEchoObject`.

## TODO/FIXME
- [ ] `generator.ts` `Filter.typename` doesn't work.

## Demo
Screenshot 2024-11-25 at 12 55 15 AM


packages/core/echo/echo-db/src/echo-handler/clone.ts
```diff
@@ -32,16 +32,16 @@ const requireAutomergeCore = (obj: EchoReactiveObject<any>) => {
  * Returns new unbound clone of the object.
  * @deprecated
  */
-export const clone = <T extends {}>(
-  obj: EchoReactiveObject<T>,
+export const clone = <T extends BaseObject>(
```
dmaretskyi, line 35: this is a good increment, but I think we should start typing those functions as:

```ts
export const clone = <O extends BaseObject>(obj: O): O
```

packages/sdk/schema/src/testing/generator.ts
```diff
@@ -0,0 +1,162 @@
...
+export const noop = (obj: any) => obj;
+
+export const logObject = (message: string) => (obj: any) => log.info(message, { obj });
+
+export const createObjectArray = <T extends BaseObject<T>>(n: number): ExcludeId<T>[] =>
+  Array.from({ length: n }, () => ({}) as any);
+
+export const createArrayPipeline = <T extends BaseObject<any>>(
+  n: number,
+  pipeline: (obj: ExcludeId<T>) => Effect.Effect<ExcludeId<T>>,
```
dmaretskyi, line 122: What do you think about having a non-effect API for this so it's consistent with the rest of the codebase?

packages/sdk/schema/src/testing/generator.ts
```diff
@@ -0,0 +1,162 @@
...
+) => {
+  return Effect.forEach(createObjectArray<T>(n), pipeline);
+};
+
+/**
+ * Create an object creation pipeline.
+ */
+export const createObjectPipeline = <T extends BaseObject>(type: S.Schema<T>, db?: EchoDatabase) => {
+  // TODO(burdon): Keep as functions and wrap with effect here?
+  const f1 = setProps(type);
```
dmaretskyi, line 132: f1, f2?

packages/sdk/schema/src/testing/generator.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Set references.
+ */
+export const setReferences = <T extends BaseObject>(type: S.Schema<T>, db: EchoDatabase) => {
+  return (obj: BaseObject<T>) => {
+    return Effect.promise(async () => {
+      for (const property of getSchemaProperties<T>(type.ast)) {
+        if (property.format === FormatEnum.Ref) {
+          const jsonSchema = findAnnotation<JsonSchemaType>(property.ast, AST.JSONSchemaAnnotationId);
```
dmaretskyi, line 75: I don't think we should be accessing `JSONSchemaAnnotationId` anywhere but the serializer.

I would just convert the effect schema to JSON-schema using the standard serializer and work with that.

packages/core/echo/echo-db/src/echo-handler/clone.ts
richburdon: just changing T to O?

packages/core/echo/echo-schema/src/types.ts
```diff
@@ -45,7 +45,21 @@ export const ObjectMetaSchema = S.mutable(
...
+// Objects
+//
+
+/**
+ * Base type for all data objects (reactive, ECHO, and other raw objects).
+ * NOTE: This describes the base type for all database objects.
+ * It is stricter than `T extends {}` or `T extends object`.
+ */
+// TODO(burdon): Consider moving to lower-level base type lib.
+export type BaseObject<T = any> = Record<keyof T, any>;
```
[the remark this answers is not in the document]
richburdon: i thought this provided stronger typing for indexing the object.

packages/sdk/schema/src/testing/generator.ts
richburdon: see TODO here: increment would be to revert the transform functions (e.g., setProps) to vanilla TS and wrap them as effects here.

PR #8288, opened 2024-11-25
feat(forms): Support array types

### LINKED ISSUE #8287: [forms] Form should handle arrays

ZaymonFC:
- Resolves #8287 

https://github.com/user-attachments/assets/eabc0484-18b7-4e33-9939-00095adc218a

We should do another increment on this and make the returned properties from `getSchemaProperties` hierarchical in order to move a lot of the processing logic out of the componentry and improve testability.


packages/ui/react-ui-form/src/components/Form/Form.tsx
```diff
@@ -128,6 +129,127 @@ export const Form = <T extends BaseObject>({
...
+                }
+
+                return (
+                  <div key={name} role='none'>
+                    Nested form not supported in arrays, yet.
+                  </div>
+                );
+              }
+
+              return (
```
richburdon, line 205: this inline function is getting a bit long. suggest splitting it out a local function at the bottom of this file to generate the component based on the type.

packages/ui/react-ui-form/src/components/Form/Form.tsx
```diff
@@ -101,12 +101,13 @@ export const Form = <T extends BaseObject>({
     <div role='form' className={mx('flex flex-col w-full gap-2 py-2', classNames)}>
       {properties
         .map((property) => {
-          const { ast, name, type, format, title, description, examples, options } = property;
+          const { ast, name, type, format, title, description, examples, options, array } = property;
           const key = [...path, name];
           const label = title ?? pipe(name, capitalize);
```
richburdon, line 106: here's the logic we discussed re table headers. perhaps factor this out (to sdk schema).
it's a trivial function, but having something in common with table might be helpful as things evolve.

richburdon (approved):

PR #8317, opened 2024-11-27
fix: resolve scroll into view

wittjosiah:
- fixes issue where first plank was always attended when closing a plank, now attends neighbouring plank when closing
- resolves scroll into view once it has been executed so that the state doesn't grow stale
- revives scroll into view when clicking on already open items in navtree
- retains functionality to close planks from navtree, toggled by holding alt/option while clicking

https://github.com/user-attachments/assets/3647e6af-7044-49d9-95a7-a9c8bd7ee1cf


packages/plugins/plugin-deck/src/components/DeckLayout/Plank.tsx
```diff
@@ -75,8 +75,16 @@ export const Plank = memo(({ entry, layoutParts, part, layoutMode, order, last }
...
-      rootElement.current?.focus({ preventScroll: true });
-      layoutMode === 'deck' && rootElement.current?.scrollIntoView({ behavior: 'smooth', inline: 'center' });
+      // TODO(wittjosiah): When focused on page load, the focus is always visible.
+      //   Forcing focus to something smaller than the plank prevents large focus ring in the interim.
+      const focusable =
+        rootElement.current?.querySelector('input') ||
+        rootElement.current?.querySelector('button') ||
+        rootElement.current;
+      focusable?.focus({ preventScroll: true });
+      layoutMode === 'deck' && focusable?.scrollIntoView({ behavior: 'smooth', inline: 'center' });
```
thure, line 85: This seems offset by some amount, but we should fix that by implementing a deck-aware method later, this is great for now.

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
```diff
@@ -196,7 +199,7 @@ export const RawTreeItem = <T = any,>({
           isBranch && open && handleOpenChange();
           break;
         case ' ':
-          handleSelect();
+          handleSelect(event.altKey);
```
thure, line 202: Did you choose alt for a particular reason? I might have tried shift…

thure (approved):
Awesome, thank you ✨

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
wittjosiah: mostly because "alternate" behaviour/secondary "option", would be happy with shift as well if folks think that makes more sense

thure: I don’t necessarily have an opinion, would defer to user testing

thure: While using mouse and keyboard, I guess I’d want to change the cursor to indicate what it will do – maybe leave a todo about that?

thure: Maybe this is more of a global feature worth a separate issue

richburdon: add hint to tooltip?

richburdon (approved):

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
thure: > add hint to tooltip?

This is the best solution so far, yes ✨

PR #8330, opened 2024-12-01
fix: focus indicator clipping.

richburdon:
Is there a more systematic fix?

Screenshot 2024-11-30 at 11 39 45 PM

### DIFF
--- a/packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
+++ b/packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
@@ -64,7 +64,7 @@ export const IdentityActionChooserImpl = ({
   );
   return (
     // TODO(burdon): Create box container to handle padding for focus indicators so they are not clipped.
-    <div role='none' className='bs-40 grow overflow-y-auto px-1'>
+    <div role='none' className='bs-40 grow overflow-y-auto pli-1'>
       <DeviceList
         devices={devices}
         connectionState={connectionState}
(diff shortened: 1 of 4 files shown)


packages/sdk/shell/src/panels/IdentityPanel/steps/IdentityActionChooser.tsx
```diff
@@ -63,7 +63,8 @@ export const IdentityActionChooserImpl = ({
     (device) => device.profile?.type === DeviceType.AGENT_MANAGED && device.profile?.os?.toUpperCase() === 'EDGE',
   );
   return (
-    <div role='none' className='bs-40 grow overflow-y-auto overflow-x-hidden'>
+    // TODO(burdon): Create box container to handle padding for focus indicators so they are not clipped.
+    <div role='none' className='bs-40 grow overflow-y-auto px-1'>
```
thure, line 67:
```suggestion
    <div role='none' className='bs-40 grow overflow-y-auto pli-1'>
```

thure (approved):

thure:
The only silver bullet is to inset all focus rings.

richburdon:
> The only silver bullet is to inset all focus rings.
> 
> Another strategy might be to enforce padding on overflow values other than visible, though that could cause other issues.

Wouldn't the following work?

- theme adds sufficient padding to all container components that contain focusable elements
- i.e., Toolbar, List, Tree

thure:
> - i.e., Toolbar, List, Tree

It would certainly be a bug if any of those components didn't accommodate focus rings themselves, but the element you added spacing to here isn't an intermediate component like these.

I don't think there is a cross-cutting solution to this that wouldn't make debugging this more difficult than simply adjusting spacing where we neglected to as in this PR.

PR #8332, opened 2024-12-01
feat: Creation flow

richburdon:


packages/sdk/client-protocol/src/types.ts
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { TYPE_PROPERTIES, S, TypedObject } from '@dxos/echo-schema';
+
+// TODO(burdon): Wire up class and remove TYPE_PROPERTIES from echo-schema and echo-db (@dima).
+export class SpaceProperties extends TypedObject({
```
dmaretskyi, line 8: see packages/sdk/client-protocol/src/schema.ts

PR #8350, opened 2024-12-03
fix: resolve scroll intents when applied

wittjosiah:
These intent handlers were not returning anything so the intent system
thought the intent wasn't handled, this lead to the layout handling it
and focus on the thread input being lost when it shouldn't be.


thure (approved):
Aha, noted

PR #8354, opened 2024-12-03
chore: enable edge replication for spaces by default

wittjosiah:


packages/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -359,11 +361,27 @@ export const SpacePlugin = ({
     );
   };
 
+  const setEdgeReplicationDefault = async (client: Client) => {
+    try {
+      await Promise.all(
+        client.spaces.get().map(async (space) => {
+          await space.waitUntilReady();
+          await space.internal.setEdgeReplicationPreference(EdgeReplicationSetting.ENABLED);
```
yarolegovich, line 369: might be better to set this before waitUntilReady. 
if I got admitted by someone's agent and there are no online peers, the space won't become ready until automerge root doc is replicated, which won't happen until `setEdgeReplicationPreference(true)`

wittjosiah: interesting, okay, I'll just remove the wait until ready call. I wasn't sure if it would be safe to change the setting before the space was fully setup or not

PR #8373, opened 2024-12-06
feat: space/object creation flow

wittjosiah:
https://github.com/user-attachments/assets/6fc9a03c-a9d0-4d12-83c1-09112455fabf


thure (changes requested):
The use of `Form` here results in an experience that doesn't meet some very basic usability criteria:
- not all fields are visible, so it’s unclear to the user when they’ll finally be able to create the object
- there’s no way to go back without just closing the dialog
- when something is filled in, the result gets added to the form in an unexpected order

I understand we want to use `Form` for this, but if it is not able to deliver a good experience then it needs to be augmented. I would suggest:
- this must display all fields together in a stable order
  - if not all fields are knowable, fields should be grouped into additional steps in a wizardlike experience and the user should be able to go back to make changes
- `Form` needs to be able to render a combobox field so editing can occur in-place

If these augmentations are out of scope for this, then I think this experience should not use `Form` and instead use components that can deliver something that meets a reasonable bar for usability.

thure:
If we should deliver an experience that presents one autofocused combobox at a time rather than all of them, the dialog should show a step count or breadcrumb somewhere that outlines the process so it's clear where the user is.

wittjosiah:
I'm not sure what you expect from a breadcrumb that the input values themselves are not providing. I'm not saying the design of it is great, but all of the information that you input is there for you.

wittjosiah:
That sounds like something that could be pretty easily added on afterwards and not a reason that I'd want to prevent this from landing given that I think it's already providing a clearer picture of what you're doing then before.

thure:
I think I don't agree, this doesn't seem shippable in its current state.

If we don't need things to be shippable to merge to main, I will clear my review and file several P0's about addressing the usability issues here.

thure (approved):
Given now the field order is stable and the user can revise previous choices, this now seems reasonable to me, I’ll file an issue about steps as follow-up.

richburdon (approved):
Please merge unless there are specifics you'd like me to focus on.

PR #8376, opened 2024-12-06
feat: Schema registry API upgrade

dmaretskyi:


packages/core/echo/echo-db/src/proxy-db/mutable-schema-registry.ts
```diff
@@ -19,17 +21,167 @@ import { createStoredSchema } from '@dxos/live-object';
...
-export interface SchemaResolver {
+export interface SchemaRegistry {
+  query(query: SchemaRegistryQuery): SchemaRegistryLiveQuery<SchemaRecord>;
+
+  /**
+   * Registers the provided schema.
+   *
+   * @returns Schema records after registration.
+   *
+   * The behavior of this method depends on the state of the database. The general
```
richburdon, line 37: nit. easier to read if one sentence per line. forget the wrapping.

packages/core/echo/echo-db/src/proxy-db/mutable-schema-registry.ts
```diff
@@ -19,17 +21,167 @@ import { createStoredSchema } from '@dxos/live-object';
...
+   */
+  typename?: string[];
+
+  /**
+   * [Semver Range](https://docs.npmjs.com/cli/v6/using-npm/semver#ranges) for the schema version.
+   */
+  version?: string;
+};
+
+export interface SchemaRegistryLiveQuery<T> {
```
richburdon, line 75: SchemaRegistrySubscription?

packages/core/echo/echo-db/src/proxy-db/mutable-schema-registry.ts
```diff
@@ -19,17 +21,167 @@ import { createStoredSchema } from '@dxos/live-object';
...
+
+  /**
+   * Get backing ECHO object for this schema.
+   */
+  getBackingObject(): Promise<ReactiveEchoObject<StoredSchema> | undefined>;
+
+  /**
+   * Updates typename.
+   *
+   * Only callable if the schema is mutable.
```
richburdon, line 152: `@throws Error if schema is not mutable` (jsdoc)

richburdon: this is fine, but it might be more precise to have update verbs on the registry?

registry.update(schema, {
  upsert: Fields,
  rename: [{ from, to }],
  delete: string[],
});

OR perhaps just set the new struct/schema directly (not fine grained?)

richburdon (approved):

packages/core/echo/echo-db/src/proxy-db/mutable-schema-registry.ts
dmaretskyi: the type is reused for one-shot queries too

dmaretskyi: the idea behind this (and also data queries in ECHO) is that you first create this LiveQuery object -- it sort-of acts like a prepared query in SQL databases, and then you either run it one-shot, or subscribe to it.

dmaretskyi: I'll rename it PreparedQuery

dmaretskyi:
Frozen for now

PR #8389, opened 2024-12-09
chore: SpacePlugin/Navtree update

richburdon:
- [x] Break long node constructors in SpacePlugin into separate functions.
- [x] Normalize `renderItem` (formerly `renderColumns`) for nav tree rendering.


packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
```diff
@@ -77,7 +77,7 @@ export const NavTreePlugin = (): PluginDefinition<NavTreePluginProvides> => {
     getInitialState() ?? [
       // TODO(thure): Initialize these dynamically.
       ['root', create({ open: true, current: false })],
-      ['root~dxos.org/plugin/space-spaces', create({ open: true, current: false })],
+      ['root~dxos.org/plugin/space/spaces', create({ open: true, current: false })],
```
wittjosiah, line 80: you can't change this without changing the actual id of the graph node

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
```diff
@@ -54,7 +54,7 @@ export type TreeItemProps<T = any> = {
   path: string[];
   last: boolean;
   draggable?: boolean;
-  renderColumns?: FC<{ item: T; path: string[]; menuOpen: boolean; setMenuOpen: (open: boolean) => void }>;
+  renderItem?: FC<{ item: T; path: string[]; menuOpen: boolean; setMenuOpen: (open: boolean) => void }>;
```
wittjosiah, line 57: why render item? imo render columns makes more sense because this allows a consumer to control the rendering of the additional grid columns for each item. The rendering of the item itself is not overridable and always controlled by the tree. Perhaps a jsdoc would be helpful here.

packages/plugins/plugin-navtree/src/components/NavTree.tsx
```diff
@@ -86,7 +83,7 @@ const NavTreeColumns = ({
 
   const ActionRoot = popoverAnchorId === `dxos.org/ui/${NAV_TREE_ITEM}/${node.id}` ? Popover.Anchor : Fragment;
 
-  // TODO(thure): Ideally this should not be necessary.
+  // TODO(thure): Ideally this should not be necessary. (Explain why it's required?)
```
wittjosiah, line 86: We've not found another way to load the graph at the right time than to load the children of any node which is renderer so that if it's expanded it's children are available. Ideally this would not be done within React as an useEffect hook but everything else to this point has proven either too complicated or didn't work.

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
richburdon: the tree has no concept of columns. i found this really confusing since this function renders a row!

wittjosiah: it's a treegrid, it has columns https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-list/src/components/Tree/Tree.tsx#L23

I can see how it could be interpreted as rendering the whole column as a single component, but it renders the column one row at a time. It's explicitly not the whole item though, only the additional column content for each row

packages/plugins/plugin-navtree/src/NavTreePlugin.tsx
wittjosiah: https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-space/src/util.tsx#L40

wittjosiah: the state is cached so it would only break on fresh profile

wittjosiah: https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-space/src/util.tsx#L40

richburdon: ok. i'll revert, but let's push to enforce consistency.

packages/ui/react-ui-list/src/components/Tree/TreeItem.tsx
richburdon: i still think this is a leaky abstraction since tree doesn't know anything about columns per se. i.e., gridTemplateColumns should just be getRowClassName?

wittjosiah: it's explicitly a tree grid, so not so sure about that

richburdon: but let me take another look since i do see what you mean; i need to understand this better (which was the point of the exercise)

PR #8397, opened 2024-12-11
feat: recovery flow

wittjosiah:
Screenshot 2024-12-10 at 20 32 12
Screenshot 2024-12-10 at 20 32 34
Screenshot 2024-12-10 at 20 36 16


packages/plugins/plugin-client/src/translations.ts
```diff
@@ -9,6 +9,12 @@ export default [
     'en-US': {
       [CLIENT_PLUGIN]: {
         'open shell label': 'Open HALO',
+        'recovery code dialog title': 'Recovery Code',
+        // TODO(wittjosiah): Update this.
```
wittjosiah, line 13: @Raymmar @richburdon this is intended as placeholder copy, open to suggestions for best thing to say here.

Raymmar:
Suggestion for text"

---

Recovery Code

This is your identity recovery code, store it in a safe place. You can use it to recover your identity if you ever lose access to your devices.


[ ] I have saved the code

* Please note: This code will never be displayed again and cannot be recovered. It is your private key for recovering DXOS data. Anyone with this key will also be able to gain access to your account.

PR #8404, opened 2024-12-11
chore: Reinstate tsconfig project refs

richburdon:


wittjosiah (changes requested):
this shouldn't merge without upgrading to TS 5.7

richburdon:
> this shouldn't merge without upgrading to TS 5.7

right. i was testing and it still kills vscode, but it DOES resolve the import issue for webstorm.

PR #8409, opened 2024-12-13
chore: Adjust packaging so non-VSCode IDEs can resolve some imports

thure:
This PR:
- extends `toolbox` to support `devKeys`
  - this removes mentioned keys from the root of the package.json and moves them to `publishConfig` where it is able to do so and the package is not mentioned in `devKeysExclude`
  - adds `"types": "./src/index.ts"` to the root of the package.json, which some IDEs use for module resolution
  - adds `"type": "module"` to the root of the package.json, since some builds were failing without this
  - renames `.eslintrc.js` to `.eslintrc.cjs` to clarify its dialect in those packages


dmaretskyi (approved):

thure:
I think this is a non-starter — the proposed solution doesn’t work for our e2e test runners, e.g.:
```
[WebServer] The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
Error: require() of ES Module /Users/willdx/Projects/dxos/packages/ui/lit-grid/dist/src/testing/index.js from /Users/willdx/Projects/dxos/packages/plugins/plugin-sheet/src/testing/playwright/sheet-manager.ts not supported.
Instead change the require of index.js in /Users/willdx/Projects/dxos/packages/plugins/plugin-sheet/src/testing/playwright/sheet-manager.ts to a dynamic import() which is available in all CommonJS modules.

   at sheet-manager.ts:7

   5 | import { type Locator, type Page } from '@playwright/test';
   6 |
>  7 | import { DxGridManager } from '@dxos/lit-grid/testing';
     |                ^
   8 | import { type DxGridPosition, type DxGridAxis } from '@dxos/react-ui-grid';
   9 |
  10 | /**
```

PR #8410, opened 2024-12-13
feat!: declarative surface api

wittjosiah:


thure (approved):
Very neat, I like it!

packages/sdk/app-framework/src/plugins/plugin-surface/SurfaceContext.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export type SurfaceComponent<T extends Record<string, any> = Record<string, unknown>> = (
+  props: SurfaceComponentProps<T>,
+  forwardedRef: ForwardedRef<HTMLElement>,
+) => JSX.Element | null;
+
+/**
+ * Determines the priority of the surface when multiple components are resolved.
+ */
+export type SurfaceDisposition = 'default' | 'hoist' | 'fallback';
```
richburdon, line 74: define hoist
remove 'default' as an option? just used undefined as default?

packages/plugins/plugin-table/src/TablePlugin.tsx
```diff
@@ -91,25 +97,20 @@ export const TablePlugin = (): PluginDefinition<TablePluginProvides> => {
...
-            }
-            default:
-              return null;
-          }
-        },
+        definitions: () => [
+          createSurface({
+            id: `${TABLE_PLUGIN}/table`,
+            role: ['article', 'section', 'slide'],
+            filter: (data): data is { subject: TableType } => data.subject instanceof TableType,
```
richburdon, line 104: i wonder if target would be a better label than subject?

packages/plugins/plugin-stack/src/components/AddSectionDialog.tsx
```diff
@@ -18,7 +18,9 @@ import { nonNullable } from '@dxos/util';
 import { STACK_PLUGIN } from '../meta';
 import { StackViewType, type StackPluginProvides, type AddSectionPosition } from '../types';
 
-type AddSectionDialogProps = { path?: string; position: AddSectionPosition; collection: CollectionType };
+export const ADD_SECTION_DIALOG = `${STACK_PLUGIN}/AddSectionDialog`;
+
+export type AddSectionDialogProps = { path?: string; position: AddSectionPosition; collection: CollectionType };
 
 export const dataHasAddSectionDialogProps = (data: any): data is { subject: AddSectionDialogProps } => {
   return (
```
richburdon, line 26: i think we can ultimately move away from using these "sniffers" by using effect schema for all types.

richburdon (approved):
amazing

packages/sdk/app-framework/src/plugins/plugin-surface/SurfaceContext.ts
thure: I think there’s value in explicitly making a positive space for a behavior that to us seems “default” in situations like these, it may need layers of override where `disposition: undefined` looks weird or we may want to rename it and the “default” behavior becomes something else.

thure: To that end, I think the term ‘default’ should be replaced by something that describes more what it does than its current utility. Maybe ‘static‘ or ‘normal‘?

wittjosiah: yeah it should be easy to change because it's only used internally currently. It could just as easily be undefined but I thought having it be a string value might be clearer for future debugging tools

packages/plugins/plugin-table/src/TablePlugin.tsx
wittjosiah: can you explain target? subject is the thing that's made the most sense to me as a generic term for an arbitrary thing passed in for the surface to render: it's the subject of the surface

packages/plugins/plugin-stack/src/components/AddSectionDialog.tsx
wittjosiah: > for all types

maybe, though as @dmaretskyi has pointed out, effect schema's don't boil away like TS types do so they do make the app bigger. I think there's lots of types where the schema wouldn't be that useful

> ultimately move away from using these "sniffers"

I do think that we could have typed surfaces so that the data passed in is type-checked too, similar to what I'm doing w/ intents. This cuts down on a lot of the need for guards/data parsing, however I'm still finding some cases where I wish I had the guards in the intents (e.g. schema's w/ optional fields) so I suspect they'll still around and probably be ported over to intents as well, but they'll be relied on less heavily.

PR #8411, opened 2024-12-13
feat(plugin-kanban): KanbanPlugin

### LINKED ISSUE #8360: [kanban] stack-next & form-based kanban
- deprecate existing kanban code (generally unused? might be able to just delete it)
- use existing table View to layout object cards (Forms) using field selection to determine column
- start with kanban being off by default to ensure table doesn't depend on it's functionality
- start with it being marked as experimental, then once it's stable either make it recommended or enabled by default

thure:
This PR:
- resolves #8360
- creates a kanban with test content, does not yet support CRUD on columns or items

Screenshot 2024-12-13 at 15 44 32


packages/apps/composer-app/src/plugins.tsx
```diff
@@ -137,6 +137,7 @@ export const recommended = ({ isDev, isLabs }: PluginConfig): PluginMeta[] =>
     CallsMeta,
     ExcalidrawMeta,
     ExplorerMeta,
+    KanbanMeta,
```
richburdon, line 140: Nice! (but K comes after I)

packages/plugins/experimental/plugin-kanban/src/components/KanbanContainer.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { type MutableSchema } from '@dxos/echo-schema';
+import { useGlobalFilteredObjects } from '@dxos/plugin-search';
+import { Filter, useQuery, getSpace } from '@dxos/react-client/echo';
+import { type KanbanType, useKanbanModel, Kanban } from '@dxos/react-ui-kanban';
+import { StackItem } from '@dxos/react-ui-stack';
+
+import { stateColumns } from './create-kanban';
+
+export const KanbanContainer = ({ kanban, role }: { kanban: KanbanType; role: string }) => {
```
richburdon, line 15: i think it's worth export KanbanContainerProps

packages/plugins/experimental/plugin-kanban/src/components/create-kanban.ts
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { faker } from '@dxos/random';
```
richburdon, line 5: pls move this to the types folder.
or ../testing if that what it's used for?

richburdon (approved):

richburdon (approved):

packages/plugins/experimental/plugin-kanban/src/components/KanbanContainer.tsx
thure: Most of the main plugins are already using `StackItem.Content` instead of other grid logic. Container is also a good opportunity to load models and display a fallback meanwhile if needed.

PR #8413, opened 2024-12-15
feat(echo)!: Add reference objects to live-object API

dmaretskyi:
- **Breaking API change**
- Non-breaking in-terms of stored data 

### Notable Code Changes
1. Reference System Updates
```ts
import { makeRef } from '@dxos/live-object';
// Changed from direct assignments to using makeRef:
list.todos.push(makeRef(create(TodoType, { title, completed: false })))
```

2. Schema Updates:

```ts
import { Ref } from '@dxos/echo-schema';
// Changed from 'ref' to 'Ref' in type definitions:
schema: S.optional(Ref(MutableSchema))
```

3. Target Access Updates:

```ts
// Changed property access to use target:
list?.todos.map((todo) => todo.target).filter(nonNullable)
```

### TODO:

- [ ] Remove usages of `loadObjectReferences`
- [ ] Move `handleMutableSchema` out of echo-handler
- [ ] `@dxos/echo-schema/typed-object.ts`: Reconcile with isTypedObjectProxy
- [ ] `@dxos/echo-schema/typed-object.ts`: Reconcile with isEchoObject
- [ ] `@dxos/echo-schema/hypergraph.ts`: Put refs into proxy cache
- [ ] `@dxos/echo-schema/hypergraph.ts`: Cache per hostDb
- [ ] `@dxos/echo-schema/hypergraph.ts`: Respect load flag
- [ ] `@dxos/echo-schema/ref.ts`: Validate reference target
- [x] `@dxos/echo-schema/testing/load-object-references.test.ts`: Test Ref.load() functionality
- [ ] `@dxos/echo-schema/testing/load-object-references.test.ts`: Test ref DXN doesn't load target


packages/core/echo/echo-schema/src/object/object-id.ts
```diff
@@ -13,7 +13,21 @@ export const ObjectIdSchema = S.ULID;
...
+  isValid(id: string): id is ObjectId;
+  random(): ObjectId;
+}
+
+export const ObjectId: ObjectIdClass = class extends ObjectIdSchema {
+  static isValid(id: string): id is ObjectId {
+    try {
+      S.decodeSync(ObjectId)(id);
+      return true;
+    } catch (e) {
```
richburdon, line 26: `err` or `ignore`

packages/core/echo/echo-schema/src/testing/schema.ts
```diff
@@ -90,7 +90,7 @@ export class Contact extends TypedObject<Contact>({
     name: S.String,
     username: S.String,
     email: S.String,
-    tasks: S.suspend((): S.mutable<S.Array$<ref<Task>>> => S.mutable(S.Array(ref(Task)))),
+    tasks: S.suspend((): S.mutable<S.Array$<Ref$<Task>>> => S.mutable(S.Array(Ref(Task)))),
```
richburdon, line 93: what does suspend do?

packages/core/echo/live-object/src/ref/ref-impl.ts
```diff
@@ -0,0 +1,140 @@
...
+   * Not the same as `target` which is resolved from the resolver.
+   * @internal
+   */
+  _getSavedTarget() {
+    return this.#target;
+  }
+}
+
+/**
+ * Internal API for setting the reference resolver.
```
richburdon, line 127: `@internal`

packages/core/echo/live-object/src/ref/ref-impl.ts
```diff
@@ -0,0 +1,140 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { Reference } from '@dxos/echo-protocol';
```
richburdon, line 5: we have both Ref and Reference?

richburdon (approved):

packages/core/echo/echo-schema/src/testing/schema.ts
dmaretskyi: It's necessary to allow self-referential schema

packages/core/echo/live-object/src/ref/ref-impl.ts
dmaretskyi: can't because internal will make it package-private

packages/core/echo/live-object/src/ref/ref-impl.ts
dmaretskyi: Reference is an internal implementation detail

packages/core/echo/live-object/src/ref/ref-impl.ts
wittjosiah: perhaps worth changing the function name to make it more apparent, e.g.: `__setRefResolver`

packages/core/echo/echo-db/src/echo-handler/echo-handler.ts
```diff
@@ -176,7 +176,7 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
         return array;
       });
 
-      return createProxy(newTarget, this);
+      return internProxy(newTarget, this);
```
wittjosiah: intern?

packages/core/echo/echo-schema/src/object/object-id.ts
wittjosiah: if this isn't used I think it can just be `} catch {`

packages/apps/composer-app/src/migrations/migrations.node.test.ts
```diff
@@ -206,26 +224,27 @@ describe('Composer migrations', () => {
...
+    expect(migratedThread2?.messages?.[0].target instanceof MessageType).to.be.true;
+    expect(migratedThread2?.messages?.[0].target?.text).to.equal('hello world');
 
     const migratedSketch1 = space.db.getObjectById<DiagramType>(sketch1.id);
     expect(migratedSketch1 instanceof DiagramType).to.be.true;
     expect(migratedSketch1?.name).to.equal('My Sketch');
-    expect(migratedSketch1?.canvas?.schema).to.equal(TLDRAW_SCHEMA);
-    expect(migratedSketch1?.canvas?.content?.id).to.equal('test string');
+    // TODO(dmaretskyi): Broke with references API update.
+    // expect(migratedSketch1?.canvas.target?.schema).to.equal(TLDRAW_SCHEMA);
```
wittjosiah, line 246: why doesn't this work?

packages/apps/composer-app/src/plugins.tsx
```diff
@@ -274,12 +275,12 @@ export const plugins = ({
 
       const readme = create(DocumentType, {
         name: INITIAL_DOC_TITLE,
-        content: create(TextType, { content: INITIAL_CONTENT.join('\n\n') }),
+        content: makeRef(create(TextType, { content: INITIAL_CONTENT.join('\n\n') })),
```
wittjosiah, line 278: not a blocker but I think in the end it would be nice if object assignments were automatically wrapped in `makeRef` internally to avoid having to do this. afaik assigning an object to a field in another object wouldn't ever make sense as anything other than a ref so I don't think it would be ambiguous

packages/core/echo/live-object/src/ref/ref-array.ts
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { type BaseObject, type ObjectId, Ref } from '@dxos/echo-schema';
```
wittjosiah, line 5: wait, live object depends on echo schema? why is that?

dmaretskyi: conceptually that seems better to me: you first define the schema of the data, and then you have the API to manipulate that data. But im not set on this

packages/apps/composer-app/src/migrations/migrations.node.test.ts
dmaretskyi: I didn't dig too deep on this, but are those migrations still relevant?

PR #8414, opened 2024-12-15
chore: tailwind config viewer

richburdon:
Screenshot 2024-12-15 at 1 36 05 PM


thure (changes requested):
I think this could be very misleading — I don’t know where these shades are coming from since we don’t use `slate` or `gray` anymore, nor do we use RGB, it’s all `display-p3` now.

I think we’d be better served by a Storybook story that demonstrates all our physical and semantic color tokens. I estimate it wouldn’t be too hard to do.

PR #8417, opened 2024-12-16
chore(tables): Split out controls and presentation from table model

ZaymonFC:


packages/ui/react-ui-table/src/model/table-controls.ts
```diff
@@ -0,0 +1,42 @@
...
+import { tableControls } from '../util/table-controls';
+
+export class TableControls {
+  public readonly modals = new ModalController();
+
+  constructor(private readonly model: TableModel) {}
+
+  public handleClick = (event: MouseEvent): void => {
+    const modalHandledClick = this.modals.handleClick(event);
+
```
richburdon, line 18: no space between declaration and check.

packages/ui/react-ui-table/src/model/table-controls.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (selectionCheckbox) {
+        this.handleCheckboxClick(selectionCheckbox);
+      }
+    }
+  };
+
+  private handleCheckboxClick = (checkbox: HTMLElement): void => {
+    const isHeader = checkbox.hasAttribute(tableControls.checkbox.attributes.header);
+
```
richburdon, line 34: same

packages/ui/react-ui-table/src/model/table-controls.ts
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { type MouseEvent } from 'react';
+
+import { type TableModel } from './';
```
richburdon, line 7: bad imports:
- no import '.' or 'foo/bar'

since this happens quite frequently (i assume via zed?) can you consider creating a lint rule for this.

packages/ui/react-ui-table/src/model/table-controls.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ModalController } from './modal-controller';
+import { tableControls } from '../util/table-controls';
+
+export class TableControls {
```
richburdon, line 11: i don't understand this and i think it's just exporting what was currently a very leaky abstraction from model into another model (which might be incremental progress, but still is covering up a crime).

IMO event handlers should not be moved out of the class that owns the UX. E.g., the following is good:

```ts
cons MyComponent = ({ item, onSelect }) => {
  const handleClick = (ev) => onSelect(item);
  return <div onClick={handleClick} />
}
```

i.e., "click" (and any subsequent event type) is an implementation detail of the component implementation. Its callbacks should abstract that and involve higher level semantics -- and almost never expose events, unless they are also in the intimate business of handling the specific UX implementation.

here, this is just plumbing that seems to add another level of indirection. how would you test this class?

i think these handlers would be better off inside the table itself.

packages/ui/react-ui-table/src/model/table-presentation.ts
```diff
@@ -0,0 +1,202 @@
...
+export class TablePresentation<T extends BaseTableRow = { id: string }> {
+  constructor(
+    private readonly model: TableModel<T>,
+    private readonly _visibleRange = signal<DxGridPlaneRange>({
+      start: { row: 0, col: 0 },
+      end: { row: 0, col: 0 },
+    }),
+  ) {}
+
+  public getCells = (range: DxGridPlaneRange, plane: DxGridPlane): DxGridPlaneCells => {
```
richburdon, line 32: if these are methods, please use the standard method syntax.

packages/ui/react-ui-table/src/model/table-presentation.ts
```diff
@@ -0,0 +1,202 @@
...
+  toPlaneCellIndex,
+  type DxGridPlaneRange,
+  type DxGridCellValue,
+} from '@dxos/react-ui-grid';
+import { mx } from '@dxos/react-ui-theme';
+import { VIEW_FIELD_LIMIT, type FieldType } from '@dxos/schema';
+
+import { type BaseTableRow, type TableModel } from './table-model';
+import { tableButtons } from '../util';
+import { tableControls } from '../util/table-controls';
```
richburdon, line 21: imports issue

packages/ui/react-ui-table/src/model/table-presentation.ts
```diff
@@ -0,0 +1,202 @@
...
+        return this.getActionColumnCells(range);
+      }
+      case 'fixedStartStart': {
+        return this.getSelectAllCell();
+      }
+      case 'fixedStartEnd': {
+        return this.getNewColumnCell();
+      }
+      default: {
+        return {};
```
richburdon, line 54: is this a valid result?

packages/ui/react-ui-table/src/model/table-presentation.ts
```diff
@@ -0,0 +1,202 @@
...
+  type DxGridCellValue,
+} from '@dxos/react-ui-grid';
+import { mx } from '@dxos/react-ui-theme';
+import { VIEW_FIELD_LIMIT, type FieldType } from '@dxos/schema';
+
+import { type BaseTableRow, type TableModel } from './table-model';
+import { tableButtons } from '../util';
+import { tableControls } from '../util/table-controls';
+
+export class TablePresentation<T extends BaseTableRow = { id: string }> {
```
richburdon, line 23: contrary to TableControls, this seems like a good abstraction, but please add a one line comment in the `/** */` block.

richburdon, line 23: perhaps create a type for `{ id: string }` since it's in a lot of places (e.g., `table.model.ts`)

```
type SimpleTableRow = { id: string }
export type BaseTableRow = Record<JsonProp, any> & SimpleTableRow;
```

richburdon (changes requested):

packages/ui/react-ui-table/src/model/table-controls.ts
ZaymonFC: Resolved! I agree and you make some good points.

I've done a cleanup pass over all the click handling / modal buttons. Moved the event handler into the Table component.

packages/ui/react-ui-table/src/model/table-presentation.ts
ZaymonFC: Yes. It's typed.

ZaymonFC: We supply empty record for all grid planes we aren't interested in filling.

packages/ui/react-ui-table/src/model/modal-controller.ts
```diff
@@ -37,55 +35,24 @@ export class ModalController {
...
-    const rowButton = target.closest(`button[${tableButtons.rowMenu.attr}]`);
-    if (rowButton) {
-      this._triggerRef.current = rowButton as HTMLElement;
-      this._state.value = {
-        type: 'row',
-        rowIndex: Number(rowButton.getAttribute(tableButtons.rowMenu.attr)),
-      };
-      return true;
-    }
+  public setTrigger = (trigger: HTMLElement) => {
```
richburdon, line 38: pls use normal method syntax or consider using `set` accessor.

packages/ui/react-ui-table/src/model/modal-controller.ts
```diff
@@ -37,55 +35,24 @@ export class ModalController {
...
-        targetId,
-        typename: schemaId,
-      };
-    }
+  public showColumnCreator = () => {
+    this._state.value = { type: 'columnSettings', mode: { type: 'create' } };
+  };
 
-    return false;
+  public showReferencePanel = (targetId: string, schemaId: string) => {
```
richburdon, line 54: method syntax

richburdon (approved):

PR #8444, opened 2024-12-22
feat: low-level compute graph API

richburdon:
- Factor out low-level graph API
- Factor out hyperformula to dxos/compute


packages/core/compute/src/types/hyperformula.ts
```diff
@@ -0,0 +1,14 @@
+//
+// Copyright 2024 DXOS.org
+//
+
+import { type SimpleCellRange } from 'hyperformula/typings/AbsoluteCellRange';
+import { type SimpleCellAddress } from 'hyperformula/typings/Cell';
+import { type SimpleDate, type SimpleDateTime } from 'hyperformula/typings/DateTimeHelper';
+
+// TODO(burdon): Explain special import.
+import { DetailedCellError, ExportedCellChange } from '#hyperformula';
```
dmaretskyi, line 10: Hyperformula had issues during bundling with vite, so we pre-bundle hyperformula and place it in the vendor directory of this package. Also see the deifinition in the package.json.

PR #8448, opened 2024-12-25
feat: Object migrations

dmaretskyi:


dmaretskyi:
Merged together with https://github.com/dxos/dxos/pull/8452

PR #8453, opened 2025-01-04
fix(tables): Refs and selection issues

ZaymonFC:
- **Fix selection updates and add smoke test**
- **Add test for boolean toggles**
- **Fix references and add tests**
- **Fix test**


packages/ui/react-ui-table/src/model/table-model.ts
```diff
@@ -490,7 +305,12 @@ export class TableModel<T extends BaseTableRow = { id: string }> extends Resourc
     const { props } = this._projection.getFieldProjection(field.id);
     switch (props.format) {
       case FormatEnum.Ref: {
-        setValue(this._rows.value[rowIdx], field.path, value);
+        // TODO(ZaymonFC): This get's called an additional time by the cell editor onBlur, but with the cell editors
+        // plain string value. Maybe onBlur should be called with the actual value?
```
richburdon, line 309: nit, indent TODO (IDE picks that up).

packages/ui/react-ui-table/src/model/table-model.ts
```diff
@@ -490,7 +305,12 @@ export class TableModel<T extends BaseTableRow = { id: string }> extends Resourc
...
+        if (!isReactiveObject(value)) {
```
richburdon, line 310: nit: to avoid extract break, negate the if and put the setValue inside (slightly clearer)

richburdon (approved):

PR #8454, opened 2025-01-06
fix: Switch onCheckedChange callback

richburdon:


thure:
Excellent, thank you ✨

PR #8457, opened 2025-01-06
feat!: new plugin api

wittjosiah:


thure (approved):
Very neat, excited to start using this ✨

packages/sdk/app-framework/src/plugins/next/plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+   */
+  tags?: string[];
+};
+
+/**
+ *
+ */
+export type InterfaceDef<T> = {
+  // TODO(dima): Pattern borrowed from effect, they use symbols which I ommited for brewity.
+  _TypeId: {
```
dmaretskyi, line 69:
```ts
const InterfaceDefTypeId: unique symbol = Symbol.for(blah)

export type InterfaceDef<T> = {
  // TODO(dima): Pattern borrowed from effect, they use symbols which I ommited for brewity.
  [IntefaceDefTypeId]: {
  
  }
}
```

basically making it a `unique symbol` will hide it for intelisense in vscode

packages/sdk/app-framework/src/plugins/next/plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ *
+ */
+export type Capability<T> = {
+  interface: InterfaceDef<T>;
+  implementation: T;
+};
+
+export type AnyCapability = Capability<any>;
```
dmaretskyi, line 84:
```suggestion
export declare namespace Capability {
   export type Any = Capability<any>;
   }
```

packages/sdk/app-framework/src/plugins/next/plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+export const defineEvent = (id: string, specifier?: string) => {
+  return { id, specifier } as ActivationEvent;
+};
+
+type PluginsContextOptions = {
+  activate: (event: ActivationEvent) => MaybePromise<boolean>;
+  reset: (event: ActivationEvent) => void;
+};
+
+class CapabilityDef<T> {
```
dmaretskyi, line 106:
```suggestion
class CapabilityImpl<T> {
```

packages/sdk/app-framework/src/plugins/next/manager.ts
```diff
@@ -0,0 +1,305 @@
...
+   * Ids of plugins that are currently registered.
+   */
+  get enabled(): readonly string[] {
+    return this._state.enabled;
+  }
+
+  /**
+   * Ids of modules that are currently active.
+   */
+  get active(): readonly string[] {
```
dmaretskyi, line 74: are those arrays reactive?

how do you feel about wrapping them like `ReactiveObject<readonly string[]>` (to be renamed to just Live) so its explicit in the api?

packages/sdk/app-framework/src/plugins/next/plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+    if (index !== -1) {
+      current.splice(index, 1);
+      log('capability removed', { id: interfaceDef.identifier, count: untracked(() => current.length) });
+    }
+  }
+
+  /**
+   *
+   */
+  requestCapability<T>(interfaceDef: InterfaceDef<T>) {
```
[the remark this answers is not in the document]
dmaretskyi: to be renamed Live<T[]>

packages/sdk/app-framework/src/plugins/next/manager.ts
wittjosiah: they are and we probably should do that

packages/sdk/app-framework/src/plugins/next/plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    const index = current.findIndex((i) => i.implementation === implementation);
+    if (index !== -1) {
+      current.splice(index, 1);
+      log('capability removed', { id: interfaceDef.identifier, count: untracked(() => current.length) });
+    }
+  }
+
+  /**
+   *
```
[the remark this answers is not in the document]
wittjosiah: is that a thing?

packages/sdk/app-framework/src/plugins/next/plugin.ts
ZaymonFC: btw, this pattern in type level programming is called a `Phantom Type`

packages/sdk/app-framework/src/plugins/next/plugin.ts
richburdon: very effect.

PR #8468, opened 2025-01-08
fix(tables): Sorting overhaul

ZaymonFC:
- Pull sorting out into its own module
- Supports #8256

### DIFF
--- a/packages/ui/react-ui-table/src/model/table-sorting.ts
+++ b/packages/ui/react-ui-table/src/model/table-sorting.ts
@@ -15,6 +15,27 @@ import { type TableType } from '../types';
 export type SortDirection = 'asc' | 'desc';
 export type SortConfig = { fieldId: string; direction: SortDirection };
 
+/**
+ * Manages table sorting functionality through display and data index mapping.
+ *
+ * When a table is sorted, rows visually reorder but we need to maintain references
+ * to the original data positions. This is handled through two types of indices:
+ *
+ * - Display indices: The visual positions after sorting
+ * - Data indices: The original positions in the source data
+ *
+ * The _displayToDataIndex map maintains these relationships, enabling lookups
+ * between visual and source positions.
+ *
+ * Example mapping between display and data indices:
+ *
+ * Original Data:          Sorted Display:          Mapping: (Display → Data)
+ * ┌──────────────┐        ┌──────────────┐         ┌────────┐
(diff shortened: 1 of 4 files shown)


packages/ui/react-ui-table/src/model/table-sorting.ts
```diff
@@ -0,0 +1,124 @@
...
+      return !!value; // Coerce null/undefined to false for booleans.
+    }
+    if (type === TypeEnum.Number) {
+      return value;
+    }
+    if (value == null) {
+      return value;
+    }
+
+    // TODO(ZaymonFC): Should we format this returned value for display based on the referenced prop format?
```
wittjosiah, line 114: so this supports sorting on a reference column but only on the raw value currently?

packages/ui/react-ui-table/src/model/table-sorting.ts
```diff
@@ -0,0 +1,124 @@
...
+  constructor(
+    rows: Signal<T[]>,
+    private readonly _table: TableType,
+    private readonly _projection: ViewProjection,
+  ) {
+    this._rows = rows;
+    this.sortedRows = this.initialiseSortedRows();
+  }
+
+  private initialiseSortedRows(): ReadonlySignal<T[]> {
```
wittjosiah, line 54: nit: I tend to prefer the public api coming at the top of the class after the constructor

wittjosiah (approved):

packages/ui/react-ui-table/src/model/table-sorting.ts
```diff
@@ -0,0 +1,124 @@
...
+        if (displayIndex !== dataIndex) {
+          this._displayToDataIndex.set(displayIndex, dataIndex);
+        }
+      });
+
+      return ordered.map(({ item }) => item);
+    });
+  }
+
+  public get sorting(): ReadonlySignal<SortConfig | undefined> {
```
wittjosiah, line 88: consider whether you want to expose the signal in the api here or whether it could just call `.value` inside the getter

wittjosiah: Dima was proposing that for such apis we could add a jsdoc that includes `@reactive` in it

packages/ui/react-ui-table/src/model/table-sorting.ts
ZaymonFC: Correct.

PR #8483, opened 2025-01-14
feat(tables): Persist sort in view query

ZaymonFC:
- Resolves https://github.com/dxos/dxos/issues/8480


packages/ui/react-ui-table/src/model/table-sorting.ts
```diff
@@ -38,13 +34,12 @@ export type SortConfig = { fieldId: string; direction: SortDirection };
  */
 export class TableSorting<T extends BaseTableRow> {
```
ZaymonFC, line 35: This sorting module became a whole lot more general purpose now that we only rely on the view.

packages/ui/react-ui-table/src/model/table-model.ts
```diff
@@ -105,8 +105,8 @@ export class TableModel<T extends BaseTableRow = { id: string }> extends Resourc
   /**
    * @reactive
    */
-  public get sorting(): SortConfig | undefined {
-    return this._sorting.sorting;
+  public get sorting(): TableSorting<T> | undefined {
```
wittjosiah, line 108: seems like this should always be defined now?

packages/ui/react-ui-table/src/model/table-sorting.ts
```diff
@@ -38,13 +34,12 @@ export type SortConfig = { fieldId: string; direction: SortDirection };
...
   private readonly _displayToDataIndex = new Map<number, number>();
-  private readonly _sorting = signal<SortConfig | undefined>(undefined);
   private readonly _rows: Signal<T[]>;
   public readonly sortedRows: ReadonlySignal<T[]>;
 
   constructor(
     rows: Signal<T[]>,
-    private readonly _table: TableType,
+    private readonly _view: ViewType | undefined,
```
wittjosiah, line 42: is this really optional? if so it should probably be the last arg?

wittjosiah (approved):

packages/ui/react-ui-table/src/model/table-sorting.ts
ZaymonFC: The view is optional on table type so that propagates to here.

PR #8487, opened 2025-01-15
refactor: Use `react-ui-menu` for EditorToolbar

### LINKED ISSUE #8502: [composer] Refactor MarkdownPlugin to use `react-ui-menu`

thure:
This PR:
- resolves #8502.
- makes significant improvements to `react-ui-menu`’s dx, usability, clarity.

Screenshot 2025-01-20 at 14 31 41


packages/ui/react-ui-editor/src/components/EditorToolbar/EditorToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  type EditorToolbarProps,
+  editorToolbarSearch,
+} from './util';
+import { mountViewMode } from './viewMode';
+
+//
+// Root
+//
+// TODO(thure): Derive actions from the reactive state
+const useEditorToolbarActionGraph = ({ state }: EditorToolbarActionGraphProps) => {
```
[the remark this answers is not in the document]
thure: The result of calling `getActions` on `graph.root` immediately on mount is an empty array, though if I wait for a node that should be in there the array becomes nonempty…

packages/plugins/plugin-markdown/src/components/MarkdownEditor.stories.tsx
```diff
@@ -24,18 +30,31 @@ type StoryProps = MarkdownEditorProps & {
...
 
   return (
-    <Main.Content
-      bounce
-      data-toolbar={toolbar ? 'enabled' : 'disabled'}
-      classNames={[topbarBlockPaddingStart, editorWithToolbarLayout]}
-    >
-      <MarkdownEditor id='test' initialValue={doc.content} extensions={extensions} toolbar={toolbar} />
-    </Main.Content>
+    <IntentProvider value={storybookIntentValue}>
```
wittjosiah, line 47: TODO: Use `IntentPlugin` with `withPluginManager`

packages/plugins/plugin-navtree/src/components/NavTreeItemAction.tsx
```diff
@@ -40,23 +40,20 @@ export const NavTreeItemActionDropdownMenu = ({
...
-        <IconButton
-          {...actionButtonProps}
-          icon={icon ?? fallbackIcon}
-          label={toLocalizedString(label, t)}
-          data-testid={testId}
-          suppressNextTooltip={suppressNextTooltip}
-        />
-      </DropdownMenu.Trigger>
-    </DropdownMenu.Root>
+    <MenuProvider onAction={onAction}>
```
wittjosiah, line 44: Why is this separate from the root now? Not using radix-style?

packages/ui/react-ui-editor/src/components/EditorToolbar/viewMode.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const createViewMode = (state: EditorToolbarState) => {
+  const value = state.viewMode ?? 'source';
+  const viewModeGroupAction = createViewModeGroupAction(value);
+  const viewModeActions = createViewModeActions(value);
+  return {
+    nodes: [viewModeGroupAction as NodeArg<any>, ...viewModeActions],
+    edges: [
+      { source: 'root', target: 'viewMode' },
+      ...viewModeActions.map(({ id }) => ({ source: viewModeGroupAction.id, target: id })),
```
wittjosiah, line 45: I think this might be simplified by utilizing this convenience of the graph api https://github.com/dxos/dxos/blob/main/packages/sdk/app-graph/src/node.ts#L60

packages/ui/react-ui-editor/src/components/EditorToolbar/util.ts
```diff
@@ -0,0 +1,115 @@
...
+export const editorToolbarGap = {
+  id: 'gap',
+  type: '@dxos/react-ui-toolbar/separator',
+  properties: { variant: 'gap' },
+  data: undefined as never,
+} satisfies MenuSeparator;
+
+export const editorToolbarSearch = createEditorAction({ type: 'search' }, 'ph--magnifying-glass--regular');
+
+export const edgesArrayToRecord = (edges: { source: string; target: string }[]): Record<string, string[]> => {
```
wittjosiah, line 90: TODO: Factor out?

packages/ui/react-ui-editor/src/components/EditorToolbar/EditorToolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+    nodes.push(...viewMode.nodes);
+    edges.push(...viewMode.edges);
+  }
+  return { nodes, edges };
+};
+
+//
+// Root
+//
+const useEditorToolbarActionGraph = ({ onAction, ...props }: EditorToolbarProps) => {
```
wittjosiah, line 77: would be good to factor the internals of this out to clarify the reactive toolbar api

packages/ui/react-ui-editor/src/extensions/comments.ts
```diff
@@ -608,26 +610,17 @@ export const createExternalCommentSync = (
...
   );
 
-export const useCommentState = (): [{ comment: boolean; selection: boolean }, Extension] => {
-  const [state, setState] = useState<{ comment: boolean; selection: boolean }>({
-    comment: false,
-    selection: false,
-  });
-
-  const observer = useMemo(
+export const useCommentState = (state: ReactiveObject<EditorToolbarState>): Extension => {
```
wittjosiah, line 613: given the changes to how the editor state is being managed it would be good to verify the e2e tests still pass on composer before landing this

packages/ui/react-ui-menu/src/components/Toolbar/Toolbar.tsx
```diff
@@ -8,17 +8,13 @@ import { Toolbar as NaturalToolbar } from '@dxos/react-ui';
...
 
-const ToolbarItem = ({
-  iconSize,
-  action,
-  onAction,
-}: Pick<ToolbarProps, 'iconSize' | 'onAction'> & {
-  action: MenuAction;
-}) => {
+const ToolbarItem = ({ action }: { action: MenuAction }) => {
+  const { onAction, iconSize } = useMenu();
```
wittjosiah, line 17: is the menu context intended to be optional or required for this to work? it seems to me like required, in which case I would follow the pattern we have elsewhere of having the default value of the context be `null` rather than noops and `raise`ing if the context is missing

packages/ui/react-ui-menu/src/components/MenuContext.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const menuContextDefaults: MenuContextValue = {
+  iconSize: 5,
+  onAction: () => {},
+  resolveGroupItems: () => null,
+};
+
+export const MenuContext = createContext<MenuContextValue>(menuContextDefaults);
+
+export const MenuProvider = ({
```
wittjosiah, line 17: I still find it a little odd that this package is named `menu` and all the top level concepts are "menu", in my mind "toolbar" is the top level concept. How do you see the two in relation to each other?

packages/ui/react-ui-menu/src/components/Toolbar/ToolbarToggleGroup.tsx
```diff
@@ -2,36 +2,44 @@
...
+      onClick={handleClick}
+      variant='ghost'
     />
   );
 };
 
-export const ToolbarToggleGroup = ({ actionGroup, graph }: ToolbarActionGroupProps) => {
-  const menuActions = useMemo(() => (graph ? (graph.actions(actionGroup) as MenuAction[]) : []), [actionGroup, graph]);
-  const { selectCardinality, value } = actionGroup.properties;
+export const ToolbarToggleGroup = ({ group, items: propsItems }: ToolbarActionGroupProps) => {
```
wittjosiah, line 35: if maintaining the secondary `propsItems` path is or becomes tedious you might consider only supporting the `resolveGroupItems` path which could still support static structures fine

packages/ui/react-ui-menu/src/components/Toolbar/ToolbarToggleGroup.tsx
```diff
@@ -2,36 +2,44 @@
...
+  const items = useMenuItems(group, propsItems);
+  const { selectCardinality, value } = group.properties;
   return (
-    // TODO(thure): This is extremely difficult to manage, what do?
+    // TODO(thure): The type here is difficult to manage, what do?
     // @ts-ignore
     <NaturalToolbar.ToggleGroup type={selectCardinality} value={value}>
-      {menuActions.map((action) => (
+      {(items as MenuAction[]).map((action) => (
```
wittjosiah, line 42: is this cast because currently only supporting one level of nesting?

packages/ui/react-ui-menu/src/testing/index.ts
```diff
@@ -67,10 +66,15 @@ export const createNestedActionGraph = (groupParams?: CreateActionsParams, param
...
     // @ts-ignore
-    graph._addEdges(actions.map((action) => ({ source: group.id, target: action.id })));
+    graph._addEdges([
+      { source: 'root', target: group.id },
+      ...actions.map((action) => ({ source: group.id, target: action.id })),
+    ]);
     void graph.expand(group);
   });
-  return { graph, topLevelActions: actionGroups as ToolbarItem[] };
+  const resolveGroupItems = (groupNode?: MenuItemGroup) =>
```
wittjosiah, line 75: I think we should unify this logic with that which is factored out of the editor

wittjosiah (commented):
liking how this is coming together, great work!

packages/ui/react-ui-menu/src/components/MenuContext.tsx
thure: Toolbar is a kind of menu — this package will render any arrangement of menus within menus, it may even be possible for a dropdown menu content to render a toolbar.

If it helps, it could be renamed `ToolbarMenu` so it is more akin with `DropdownMenu`, `ContextMenu`, etc.

wittjosiah: I guess I was just thinking of the toolbar as the top-level thing where a menu is triggered from toolbar rather than the toolbar being in a menu? It seems to me like you can have a toolbar without a menu, but you couldn't have a menu without a toolbar (considering the menu trigger a toolbar group)

thure: I think it’s not important how they are nested, they are all collections of actions, and this trait is what defines these components.

thure: A toggle group or dropdown menu occurring inside a toolbar is incidental to the nesting of actions – they could be nested differently, or hosted in a context menu rather than a toolbar, etc, they are still ontologically a tree of actions.

packages/ui/react-ui-menu/src/components/Toolbar/ToolbarToggleGroup.tsx
thure: I did at first, but I added the `items` back as optional because it was a lot easier for consumers in certain situations e.g. NavTreeItemAction.

thure: The `useMenuItems` hook I think sufficiently mitigates the complexity here and clarifies to consumers a best practice for determining items.

packages/ui/react-ui-menu/src/components/MenuContext.tsx
wittjosiah: Maybe it's just me, but the reason I'm a little hung up on it is that I wouldn't think to look in the menus package to get toolbar components unless I already knew they were there, so it just feels like the package is named wrong. I'd be curious what other folks think to know if its just me or not, but I agree its not particularly important.

packages/ui/react-ui-menu/src/components/Toolbar/ToolbarToggleGroup.tsx
thure: No, I was wrestling with TS over types being narrower/broader than other types, and sometimes not handling things like separators yet… TS seems to complain in both cases, like if I pass `MenuItem` in a `Node<any>` situation, it complains even though `Node<any>` is broader than `MenuItem`… it’s type debt, which seemed low enough priority for now, but I can leave TODO’s in all such cases if it would help.

packages/ui/react-ui-menu/src/testing/index.ts
thure: I’ll see what shakes out and make this a model of how to use that, certainly.

packages/ui/react-ui-menu/src/components/Toolbar/ToolbarToggleGroup.tsx
wittjosiah: todo about why it needs to be cast currently would be beneficial I think

packages/plugins/plugin-navtree/src/components/NavTreeItemAction.tsx
thure: Because of nesting… Toolbar itself uses DropdownMenu, and it’s theoretically possible DropdownMenu could render a Toolbar, so there’s no obvious root component, hence the explicit provider.

packages/ui/react-ui-editor/src/components/EditorToolbar/viewMode.ts
thure: Is there a good example of where this is used? It’s not clear from the type definition you linked which function or method does the work the comment describes.

packages/ui/react-ui-editor/src/components/EditorToolbar/util.ts
thure: I wanted to ask you about this — where should it go?
Can certainly leave a TODO for now.

packages/ui/react-ui-menu/src/components/Toolbar/Toolbar.tsx
thure: Sure, I was considering refactoring this to use Radix’s scoped context, which does a better job of enforcing the use of the context.

packages/plugins/plugin-navtree/src/components/NavTreeItemAction.tsx
wittjosiah: I guess this is another place where I look at the toolbar as the root thing. I might consider naming the `MenuProvider` `Toolbar.Root` instead? It just seems weird to me that the toolbar actions are coming from a hook about menus.

packages/ui/react-ui-menu/src/components/MenuContext.tsx
thure: I’m amenable on the name, we should just bear in mind this package is for components that surface a tree of actions, toolbar being among them. ‘Menu’ seemed like the right hypernym for this group, but maybe it’s something else, or we should list all the most salient components of the package in the package’s name perhaps.

packages/plugins/plugin-navtree/src/components/NavTreeItemAction.tsx
thure: I can see the naming is causing a lot of issues, maybe this should not be called toolbar at all? HorizontalStaticMenu? What name would bring better clarity to toolbar’s relatedness with all the other menus?

packages/ui/react-ui-editor/src/extensions/comments.ts
thure: For sure

packages/plugins/plugin-navtree/src/components/NavTreeItemAction.tsx
wittjosiah: I don't want to get too hung up on it because renaming things is pretty easy. Maybe just stick with what you've got for now and get feedback from Rich/Zaymon on the names as well before deciding if they should be changed (and to what).

packages/ui/react-ui-menu/src/hooks.ts
```diff
@@ -0,0 +1,68 @@
...
+          targetEdges.inbound.push(source);
+        }
+
+        return acc;
+      }, {}),
+    ).map(([id, { outbound }]): [string, string[]] => [id, outbound]),
+  );
+};
+
+export const useMenuGraph = (actionCreator: () => { nodes: NodeArg<any>[]; edges: Edge[] }) => {
```
wittjosiah, line 41: I think I'd call this `useMenuActions` and not return the `graph`. Is there a reason you were thinking the graph needs to be exposed?

thure: Not really and fair enough, can make that change.

PR #8497, opened 2025-01-17
feat(tables): Create table from existing schema

### LINKED ISSUE #8391: [tables] Views should ignore 'compound' fields when initialised with a schema
## Overview
Currently, table views break when initialized with schemas containing compound objects (e.g., `address` fields). We need to modify the table view initialization to gracefully handle compound fields by ignoring them during the initial view creation.

## Current Behavior
When creating a table view from a schema with compound objects (like https://github.com/dxos/dxos/blob/main/packages/sdk/schema/src/testing/types.ts#L67), the table rendering breaks due to inability to handle nested data structures.

## Proposed Behavior
- Table view initialization should skip compound fields
- These fields will still be accessible in the form view/sidebar
- Future iterations may support displaying compound fields with derived display values

### LINKED ISSUE #8491: [tables] Create table dialogue should let you pick a starting schema
- Depends on #8391

ZaymonFC:
- Resolves #8391 
- Resolves #8491


packages/plugins/plugin-table/src/capabilities/react-surface.tsx
```diff
@@ -24,4 +28,33 @@ export default () =>
...
+      id: `${TABLE_PLUGIN}/create-initial-schema-form`,
+      role: 'form-input',
+      filter: (data): data is { prop: string; schema: S.Schema<any>; target: Space | CollectionType | undefined } => {
+        if (data.prop !== 'initialSchema') {
+          return false;
+        }
+
+        // TODO(ZaymonFC): use an annotation on the schema.
+        return true;
+      },
```
wittjosiah, line 41: reminder to use annotation

packages/sdk/app-framework/src/react/Surface.tsx
```diff
@@ -9,13 +9,29 @@ import { byDisposition } from '@dxos/util';
...
+const findCandidates = (surfaces: SurfaceDefinition[], { role, data }: Pick<SurfaceProps, 'role' | 'data'>) => {
+  return Object.values(surfaces)
+    .filter((definition) =>
+      Array.isArray(definition.role) ? definition.role.includes(role) : definition.role === role,
+    )
+    .filter(({ filter }) => (filter ? filter(data ?? {}) : true))
+    .toSorted(byDisposition);
+};
+
+export const isSurfaceAvailable = (context: PluginsContext, { role, data }: Pick<SurfaceProps, 'role' | 'data'>) => {
```
wittjosiah, line 29:
```suggestion
/**
 * @returns `true` if there is a contributed surface which matches the specified role & data, `false` otherwise.
 */
export const isSurfaceAvailable = (context: PluginsContext, { role, data }: Pick<SurfaceProps, 'role' | 'data'>) => {
```

packages/ui/react-ui-form/src/components/Form/Form.tsx
```diff
@@ -43,6 +43,11 @@ export type FormProps<T extends BaseObject> = ThemedClassName<
     testId?: string;
     onCancel?: () => void;
 
+    lookupComponent?: (args: {
```
wittjosiah, line 46: mark `Custom` as deprecated

packages/ui/react-ui-table/src/util/table.ts
```diff
@@ -13,81 +13,108 @@ import {
...
-      name: 'Test',
-      typename: contactSchema.typename,
-      jsonSchema: contactSchema.jsonSchema,
-      fields: ['name', 'active', 'email', 'salary'],
-    }),
-  );
+    table.view = makeRef(
+      createView({
+        // TODO(ZaymonFC): Don't hardcode name?
+        name: 'View',
```
wittjosiah, line 55: could the name be omitted?

wittjosiah (approved):

packages/ui/react-ui-table/src/util/table.ts
ZaymonFC: It's required in the schema at the moment.

PR #8500, opened 2025-01-20
fix: typescript editor folding & selection

wittjosiah:
- enable code folding in typescript editor
- stop active line from hiding text selection


thure (approved):
lgtm ✨

PR #8505, opened 2025-01-22
feat(plugin-script): handle x-typescript-types headers

wittjosiah:
Screenshot 2025-01-22 at 15 07 56

it also handles fairly complex type traversals though they don't always come out looking pretty...
Screenshot 2025-01-22 at 15 07 07


dmaretskyi:
@cursor-bot please fix the CI failures

PR #8506, opened 2025-01-22
chore: Refactor grid plugins to use ToolbarMenu

thure:
This PR:
- [x] refactors SheetPlugin
- [x] refactors TablePlugin

#### Sheet

https://github.com/user-attachments/assets/7a91aaf7-f4e0-482f-aab5-5d69950e125a

Screenshot 2025-01-21 at 18 58 02

#### Table

Screenshot 2025-01-22 at 00 55 17
Screenshot 2025-01-22 at 00 42 28


packages/plugins/plugin-sheet/src/components/SheetToolbar/SheetToolbar.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useAlignState(state);
+  useStyleState(state);
+  useCommentState(state);
+
+  const actionsCreator = useCallback(() => createToolbarActions(state), [state]);
+  const menu = useMenuActions(actionsCreator);
+  const handleAction = useToolbarAction(state);
+
+  return (
+    <MenuProvider {...menu} onAction={handleAction as MenuActionHandler}>
```
wittjosiah, line 50: could this cast happen within `useToolbarAction`?

packages/plugins/plugin-sheet/src/components/SheetToolbar/comment.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useSheetContext } from '../SheetContext';
+
+export type CommentAction = { key: CommentKey; value: CommentValue; cellContent?: string };
+
+export type CommentState = { commentEnabled: 'comment' | 'no cursor' | 'selection overlaps existing comment' };
+
+export const useCommentState = (state: Partial<CommentState>) => {
+  const { cursorFallbackRange, model } = useSheetContext();
+
+  // TODO(thure): Can this O(n) call be memoized?
```
wittjosiah, line 21: this is already in `useMemo` what do you mean by memoized here?

packages/plugins/plugin-table/src/components/TableContainer.tsx
```diff
@@ -107,11 +104,7 @@ const TableContainer = ({ role, table }: LayoutContainerProps<{ table: TableType
 
   return (
     <StackItem.Content toolbar role={role}>
-      <Toolbar.Root onAction={handleAction} classNames={!hasAttention && 'opacity-20'}>
-        <Toolbar.Editing />
-        <Toolbar.Separator />
-        <Toolbar.Actions viewDirty={model?.isViewDirty} />
-      </Toolbar.Root>
+      <TableToolbar onAction={handleAction} classNames={['p-1 attention-surface', !hasAttention && 'opacity-20']} />
```
wittjosiah, line 107: thinking that these class names should be factored out to the intermediate component since we will want all our toolbars to have this behaviour. Attention is one of the few context's which also totally makes sense for it to not `raise` when requesting it. If the context is not available then naturally attention is false so it works fine at lower levels without adding the overhead of needing an attention provider in every story.

packages/sdk/app-framework/src/core/manager.test.ts
```diff
@@ -78,7 +78,7 @@ describe('PluginManager', () => {
     });
     const Test = definePlugin(testMeta, [Hello]);
 
-    const manager = new PluginManager({ plugins: [Test], enabled: [Test.meta.id], pluginLoader });
+    const manager = new PluginManager({ plugins: [Test], core: [], enabled: [Test.meta.id], pluginLoader });
```
wittjosiah, line 81: I think this one is not necessary, the only difference between core and enabled is that core cannot be disabled however this test doesn't attempt to disable the plugin

packages/sdk/app-framework/src/testing/withPluginManager.tsx
```diff
@@ -3,37 +3,53 @@
...
+        root: () => <Story />,
+      });
+
+      pluginManager.context.contributeCapability({
+        ...capability,
+        module: 'dxos.org/app-framework/withPluginManager',
+      });
+
+      return () => pluginManager.context.removeCapability(capability.interface, capability.implementation);
+    });
```
wittjosiah, line 51:
```suggestion
    }, []);
```
this is my bad

packages/ui/react-ui-menu/src/util.ts
```diff
@@ -2,13 +2,60 @@
...
 export const fallbackIcon = 'ph--placeholder--regular';
+
+const noop = () => {};
+
+export const createMenuAction = <P extends {} = {}>(id: string, properties: P & MenuActionProperties) =>
+  ({
+    id,
+    type: ACTION_TYPE,
+    properties,
+    data: noop,
```
wittjosiah, line 31: why is the action a noop?

packages/ui/react-ui-table/src/components/TableToolbar/TableToolbar.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+};
+
+export const TableToolbar = ({ classNames, viewDirty, onAction }: TableToolbarProps) => {
+  const state = useMemo(() => create<TableToolbarState>({ viewDirty }), []);
+  const actionsCreator = useCallback(() => createTableToolbarActions(state), [state]);
+  const menu = useMenuActions(actionsCreator);
+
+  return (
+    <MenuProvider {...menu} onAction={onAction as MenuActionHandler}>
```
wittjosiah, line 68: same thought as sheet

packages/plugins/plugin-table/src/components/TableContainer.tsx
thure: My first impression is this feels like not quite the right separation of concerns, this would mean react-ui-menu should depend on react-ui-attention. I’m amenable to trying it, though.

packages/plugins/plugin-sheet/src/components/SheetToolbar/comment.ts
thure: I think I mean dx-grid could have some way of optimizing these queries, i.e. the solution would go further upstream from here. Not within scope of this PR certainly.

packages/plugins/plugin-table/src/components/TableContainer.tsx
wittjosiah: yeah, I think that's okay, I'd argue our intermediate components should be attention-aware where it makes sense if there's an attention provider in the tree

packages/ui/react-ui-menu/src/util.ts
wittjosiah: I guess I should would have expected that instead of a noop it was a function which toggled some state. Where does the state get toggled?

packages/ui/react-ui-table/src/components/TableToolbar/TableToolbar.tsx
thure: This one is different from sheet, it does need the consumer to give it an action handler for a narrower type, since the handler is defined by the plugin but the logic is here in the intermediate component. Will leave as-is.

packages/plugins/plugin-table/src/components/TableContainer.tsx
thure: I can appreciate this essentially means that attention as a concept is upstream of even intermediate components, sitting between low-level UI (react-ui) and those, that does seem reasonable.

packages/ui/react-ui-menu/src/util.ts
thure: The action handler can and often does manage state so consumers don’t need to rig up an intent resolver.

thure: In `react-ui-menu`, the action handler is given the node itself rather than calling the node’s `data`, since the node is considered to be the relevant data.

wittjosiah: I see... so basically there's a single action handler and it takes the action node itself as the input to decide on the action to perform. I'm not sure how I missed that before but I get it now.

I feel a little uneasy about high-jacking the `Action` type in this way, it would be surprising to me that the action's data does nothing. FWIW the action's callback doesn't depend on intent dispatcher in any way, that's just how most of our graph actions execute, it's intentionally a callback rather than a static intent definition so that you can put logic in there. I think it would work totally fine to break up the monolithic action reducers into the node callbacks. That being said, I don't think you need to do that now since you've already got this working, but I do think we should probably leave a todo about it.

PR #8510, opened 2025-01-23
chore: Refactor ScriptPlugin’s toolbar to use `react-ui-menu`

thure:
This PR:
- refactors ScriptPlugin to use `react-ui-menu`

https://github.com/user-attachments/assets/0ebfaff0-4f75-4b54-a83b-904dd32c7af3


packages/plugins/experimental/plugin-script/src/components/ScriptToolbar/ScriptToolbar.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const format = createFormat();
+  const gap = createGapSeparator('gap');
+  const deploy = createDeploy(state);
+  const view = createView(state);
+  return {
+    nodes: [...templateSelect.nodes, ...format.nodes, gap, ...deploy.nodes, ...view.nodes],
+    edges: [
+      ...templateSelect.edges,
+      ...format.edges,
+      { source: 'root', target: gap.id },
```
richburdon, line 37: couldn't we have a shortcut for this? no need for the user to create a new instance?

packages/plugins/experimental/plugin-script/src/components/ScriptToolbar/useToolbarAction.ts
```diff
@@ -0,0 +1,51 @@
...
+        case 'format':
+          void handleFormat();
+          break;
+        case 'deploy':
+          void handleDeploy();
+          break;
+        case 'copy':
+          void handleCopy();
+          break;
+      }
```
richburdon, line 47: either put a default error trap or return true/false (library expectation of the callers) so that the caller can determine if something is not handled?

richburdon (approved):
looks great. MUCH cleaner.

PR #8515, opened 2025-01-24
feat: Complementary sidebar increment

thure:
This PR refactors the complentary sidebar to use a tab pattern and implements safe zone padding on relevant elements.

The top theoretically applies safe zone padding as well, though I wasn’t able to elicit a nonzero value from my iPhone or iPad since the statusbar space is reserved even for PWAs.

#### iPhone

IMG_1347
IMG_1346

#### iPad

IMG_0204
IMG_0205


ZaymonFC (approved):
I like it.

PR #8518, opened 2025-01-24
feat!: promote wnfs plugin

wittjosiah:
- deletes ipfs plugin
- align cm extension with basic image extension
- update space properties used by plugin (breaking change)
- add intents for creating and uploading files
- integrate file schema with object creation flow (images added in documents also create objects)
image


packages/plugins/plugin-wnfs/src/components/FileInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+    onDropAccepted,
+  });
+
+  return (
+    <div
+      {...getRootProps()}
+      className={mx(
+        'flex flex-col items-center p-8 border border-separator',
+        isFocused && 'focus-ring',
+        isDragAccept && 'bg-attention',
```
thure, line 38: If this has text in it, also do `text-inverse`

thure (approved):
Looking forward to this!

packages/plugins/plugin-wnfs/src/components/FileInput.tsx
thure: Wait, nevermind, I misread this as `bg-accent`, `bg-attention` is fine on its own of course.

packages/plugins/plugin-wnfs/src/capabilities/intent-resolver.ts
```diff
@@ -0,0 +1,26 @@
...
+import { create } from '@dxos/react-client/echo';
+
+import { WnfsCapabilities } from './capabilities';
+import { upload } from '../helpers';
+import { FileType, WnfsAction } from '../types';
+
+export default (context: PluginsContext) =>
+  contributes(Capabilities.IntentResolver, [
+    // TODO(wittjosiah): Deleting the object doesn't delete the wnfs blob.
+    //  Consider ways to trigger blob deletion based on object deletion and/or adding file system manager.
```
icidasset, line 15: Good shout yeah. I should also note that a WNFS never shrinks because it keeps all the history/versions of everything. Also something something about edge blob service keeping all the DAGs. WNFS doesn't have any pruning/snapshots built-in at the moment, but you can kind of replicate that behaviour by creating an entirely new WNFS and adding back the stuff you need (which is doable in this case, as the space just keeps around the access/encryption key).

icidasset (approved):
One minor typo that resulted in an error when uploading an image, but otherwise looking good! 🙌 

Also, the upload-file taskbar item doesn't seem to have an icon?
Screenshot 2025-01-25 at 14 04 15

icidasset:
PS. The code mirror extension seems to work much better now 🎉

packages/plugins/plugin-wnfs/src/helpers/load.ts
```diff
@@ -10,16 +10,40 @@ import { AccessKey, PrivateForest, PrivateDirectory, PrivateNode } from 'wnfs';
...
+  instances?: WnfsCapabilities.Instances;
+}) => {
+  // TODO(wittjosiah): Remove.
+  // Delete old properties if they exist.
+  if (space.properties.wnfs_access_key !== undefined && space.properties.wnfs_private_forest_cid !== undefined) {
+    delete space.properties.wnfs_access_key;
+    delete space.properties.wnfs_private_forest_cid;
+  }
+
+  const cacheKey = space.properties.wnfs.privateForestCid;
```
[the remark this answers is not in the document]
wittjosiah: whoops, classic last second untested refactor breaking things. Thanks for trying it out!

wittjosiah:
> The code mirror extension seems to work much better now

The key realization there was that by shifting where the async code lived we could essentially just copy the basic image plugin's structure. It was easy once I realized I could just copy something that was already working reasonably well :)

wittjosiah:
> Also, the upload-file taskbar item doesn't seem to have an icon?

oh... there's a long-standing bug where in the dev server occasionally icons are missing because of a weird sprite caching issue. It _does_ look like I'm missing a translation there though, will take a look at that

PR #8522, opened 2025-01-25
test: Compute graph

richburdon:


packages/plugins/experimental/plugin-canvas/src/types/schema.ts
```diff
@@ -2,89 +2,54 @@
...
-
 export class CanvasBoardType extends TypedObject({
   typename: 'dxos.org/type/CanvasBoard',
   version: '0.1.0',
 })({
   name: S.optional(S.String),
-  layout: Layout,
+
+  /**
+   * Graph of shapes positioned on the canvas.
```
wittjosiah, line 52: it occurs to me that in the future it would make sense for this to utilize our other canvas type once it's using [Open Canvas Interchange Format (OCIF)](https://www.canvasprotocol.org/) instead of tldraw's internal schema

PR #8525, opened 2025-01-27
refactor(app-framework): dispatcher defers to effect for error handling

wittjosiah:


packages/sdk/app-framework/src/plugin-intent/intent-dispatcher.ts
```diff
@@ -54,6 +55,8 @@ export type IntentEffectResult<Fields extends IntentParams> = {
    * An error that occurred while performing the action.
    *
    * If the intent is apart of a chain of intents and an error occurs, the chain will be aborted.
+   *
+   * Return caught error instead of throwing to trigger other intent to be triggered prior to returning the error.
```
ZaymonFC, line 59: Say that 5 times fast.

packages/sdk/app-framework/src/plugin-intent/intent-dispatcher.ts
```diff
@@ -206,36 +207,38 @@ export const createDispatcher = (
...
+      return result.data;
     });
   };
 
   const dispatchPromise: PromiseIntentDispatcher = (intentChain) => {
-    const program = dispatch(intentChain);
-    return Effect.runPromise(program);
+    return Effect.runPromise(dispatch(intentChain))
+      .then((data) => ({ data }))
+      .catch((error) => ({ error }));
```
ZaymonFC, line 241: Is there a way to factor this out? I see you're doing this operation twice.

packages/sdk/app-framework/src/plugin-intent/intent-dispatcher.ts
```diff
@@ -206,36 +207,38 @@ export const createDispatcher = (
...
+      });
+
+      if (result.undoable && isUndoable(results)) {
+        // TODO(wittjosiah): Is there a better way to handle showing undo for chains?
+        yield* pipe(
+          dispatch(createIntent(IntentAction.ShowUndo, { message: result.undoable.message })),
+          Effect.catchSome((err) =>
+            err instanceof NoResolversError ? Option.some(Effect.succeed(undefined)) : Option.none(),
+          ),
+        );
```
ZaymonFC, line 231: Do we need to log here?

ZaymonFC (approved):
Looks good to me.

packages/sdk/app-framework/src/plugin-intent/intent-dispatcher.ts
wittjosiah: I don't imagine doing it anywhere else so didn't feel necessary because it's internal but it could be pretty easily if it's needed elsewhere later 👍

packages/sdk/app-framework/src/plugin-intent/intent-dispatcher.ts
wittjosiah: I wondered about that but decided it wasn't worth it because there'd be no way to suppress it if you actually didn't want to provide a handler

PR #8528, opened 2025-01-30
feat: Popper safe area collision paddings

thure:
This PR
- ensures all `react-ui` Popper components include safe area values in their collision padding

Screenshot 2025-01-29 at 17 17 18
Screenshot 2025-01-29 at 17 14 44
Screenshot 2025-01-29 at 17 14 25

### DIFF
--- a/packages/ui/primitives/react-hooks/src/useResize.ts
+++ b/packages/ui/primitives/react-hooks/src/useResize.ts
@@ -1,0 +1,16 @@
+//
+// Copyright 2023 DXOS.org
+//
+
+import { useLayoutEffect } from 'react';
+
+export const useResize = (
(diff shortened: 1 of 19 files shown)


packages/ui/primitives/react-hooks/src/useResize.ts
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright 2023 DXOS.org
+//
```
ZaymonFC, line 3: 2025?

thure: Huh… maybe the lint rule is acting strangely?

ZaymonFC (approved):
Wonderful!

PR #8564, opened 2025-02-10
feat(plugin-deck): have a deck for each space

wittjosiah:
https://github.com/user-attachments/assets/dff0b7b9-92bd-43ea-b74e-5d9d773f96a0


thure (approved):
An interesting idea. I think this is worth a shot.

PR #8568, opened 2025-02-10
refactor: distinguish between position and disposition

wittjosiah:
- position is used to describe a fuzzy ordering
- disposition is used to describe the how the node should be rendered
- allows settings to be included in the L0


thure (approved):
lgtm! ✨

PR #8578, opened 2025-02-12
feat: dynamic enable/disable of plugins without reload

wittjosiah:
- Also enables modules to be loaded separately from activation being fired
- minimizes the usage of the ReactContext capability to a minimum


packages/sdk/app-framework/src/playground/generator/Toolbar.tsx
```diff
@@ -6,34 +6,40 @@ import React, { useCallback } from 'react';
...
   }, [manager]);
 
-  const handleCount = useCallback(async () => {
-    if (manager.pendingReset.includes(CountEvent.id)) {
-      await manager.reset(CountEvent);
-    } else {
-      await manager.activate(CountEvent);
-    }
-  }, [manager]);
+  const handleReset = useCallback(() => {}, []);
```
ZaymonFC, line 24: Empty callback? Document?

packages/sdk/app-framework/src/playground/generator/generator.ts
```diff
@@ -2,14 +2,27 @@
...
 import { contributes, defineEvent, defineCapability, defineModule, definePlugin } from '../../core';
+import { createResolver, type IntentSchema } from '../../plugin-intent';
 
 export const Number = defineCapability<number>('dxos.org/test/generator/number');
 
 export const CountEvent = defineEvent('dxos.org/test/generator/count');
 
 export const createPluginId = (id: string) => `dxos.org/test/generator/${id}`;
 
+export const createGeneratorIntent = (id: string) => {
```
ZaymonFC, line 17: What does this do? Can you explain the Alert section of this?

ZaymonFC (approved):
Looking good, I'm still getting my head around some of the new Effect stuff though. Nice!

packages/sdk/app-framework/src/playground/generator/generator.ts
wittjosiah: it's an intent which window.alerts the generated capability, was just using this to test that intents could be added dynamically

packages/sdk/app-framework/src/core/manager.ts
```diff
@@ -395,12 +401,21 @@ export class PluginManager {
...
+        concurrency: 'unbounded',
+      });
+
+      log('activating module', { module: module.id });
       // TODO(wittjosiah): This is not handling errors thrown if this is synchronous.
-      const program = module.activate(this.context);
-      const maybeCapabilities = yield* Match.value(program).pipe(
-        Match.when(Effect.isEffect, (effect) => effect),
+      const maybeCapabilities = typeof getCapabilities === 'function' ? getCapabilities() : getCapabilities;
+      const someCapabilities = yield* Match.value(maybeCapabilities).pipe(
```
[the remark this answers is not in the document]
ZaymonFC: I would avoid usage of the name 'some' in effect code because Options use `Some / None`.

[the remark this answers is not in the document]
wittjosiah: yeah... those prefixes are not that helpful. Need to think of better ones

PR #8592, opened 2025-02-14
refactor: move plugin registry out of settings

wittjosiah:
https://github.com/user-attachments/assets/66b6e1ac-07c1-4bca-a3ab-d68970004814

Of note:
- introduced a way to have planks describe mutual exclusivity in the deck (via key property)
- includes the pinned L0 items in the tab list
- update solo/unsolo icons
- update object settings icon


packages/plugins/plugin-registry/src/components/PluginList.tsx
```diff
@@ -2,111 +2,23 @@
...
-  installed?: readonly string[];
-  enabled?: readonly string[];
-  onChange?: (id: string, enabled: boolean) => void;
-  onReload?: () => void;
 };
 
 export const PluginList = ({ plugins = [], ...props }: PluginListProps) => {
   return (
-    <List classNames='mb-4 select-none'>
+    <List classNames='grid grid-cols-[repeat(auto-fill,minmax(240px,1fr))] gap-3 p-3'>
```
thure, line 18: Excellent layout choice ✨

packages/plugins/plugin-registry/src/components/RegistryContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+        part: 'main',
+        // TODO(wittjosiah): `/` currently is not supported in ids.
+        subject: [pluginId.replaceAll('/', ':')],
+        options: { pivotId: id, positioning: 'end' },
+      }),
+    );
+
+  return (
+    <div className='overflow-y-auto'>
+      <PluginList plugins={plugins} enabled={manager.enabled} onClick={handleClick} onChange={handleChange} />
```
thure, line 47: Maybe `PluginList` should handle its own overflow?

thure: One of these needs to render a `StackItem.Content`.

packages/plugins/plugin-deck/src/layout.ts
```diff
@@ -15,9 +15,18 @@ export const openEntry = (deck: string[], entryId: string, options?: OpenLayoutE
...
     }
 
+    const key = options?.key;
     const plankPositioning = options?.positioning ?? 'start';
     const pivotId = options?.pivotId;
 
+    if (key) {
+      const index = draft.findIndex((id) => id.startsWith(key));
+      if (index !== -1) {
+        draft.splice(index, 1, entryId);
```
richburdon, line 25: what does draft mean?

richburdon (approved):

packages/plugins/plugin-deck/src/layout.ts
wittjosiah: @ZaymonFC wrote this util is using `immer`, my understanding is that `draft` is a clone of the incoming value that can be mutated and then at the end it will be returned as the new value

PR #8595, opened 2025-02-15
feat: IDE-style statusbar

thure:
This PR:
- implements the IDE-style statusbar
- moves `role='status'` surface into R0 and the statusbar CTAs into R1
- clears some other technical debt

https://github.com/user-attachments/assets/18eea78f-eb0c-4996-aae7-bd5d9522de00


ZaymonFC (approved):
LGTM

Issue #8262, opened 2024-11-20
[dxosui] Composer color token audit and adjustment


thure:
Punchlist:
- [x] attention contrast (light mode is broken)
- [ ] spaces apply colors that the user can configure (not on creation necessarily)
- [x] kanban levels
- [x] CM gutters occlude peer cursors

PR #8605, opened 2025-02-18
chore: Clear color token cruft, add alias token layer

thure:
This PR:
- prepares for a resolution to #8262.
- renames the `base` color token as `baseSurface` to resolve the overload for `text-base` (now a font size)
- adds a token alias layer so components with special variables can be configured by the theme system to point to semantic or physical tokens
- cleans up other technical debt in the lower-level design system

No changes should be apparent in the Composer experience.


ZaymonFC (approved):
Looking good! I'd be keen to talk to you about aliasing to understand the system better.

PR #8607, opened 2025-02-18
docs: fix command to run tailwind config viewer

HerbCaudill:
The command provided 

```shell
pnpm nx run tailwind-config-viewer
```

gives the error

```shell
 NX   Both project and target have to be specified
 ```
 
 This command works:
 
 ```shell
 pnpm nx run react-ui-theme:tailwind-config-viewer
 ```

### DIFF
--- a/packages/ui/react-ui-theme/README.md
+++ b/packages/ui/react-ui-theme/README.md
@@ -11,7 +11,7 @@ pnpm i @dxos/react-ui-theme
 To view the interactive Tailwind CSS configuration viewer, run the following command:
 
 ```bash
-pnpm nx run tailwind-config-viewer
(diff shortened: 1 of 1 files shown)


thure:
Hi @HerbCaudill, thank you for bringing this to our attention. This tool has been unable to support how we currently handle our design system tokens (without fairly [expensive interventions](https://github.com/rogden/tailwind-config-viewer?tab=readme-ov-file#themereplacements)) for a long time and we’ve overlooked fixing or removing it, so I’ll remove it as part of #8605.

We’re introducing a theme editor to Composer and it should be able to enumerate the design system tokens like this tool does. There is also a color token story in [the Storybook](https://dxos-storybook.pages.dev/?path=/story/ui-react-ui-theme-tokens--tokens).

PR #8611, opened 2025-02-18
fix: Fix storybook for Calls; merge calls from labs into feature/compute

mykola-vrmchk:


packages/plugins/experimental/plugin-calls/src/hooks/useHandleTranscription.ts
```diff
@@ -0,0 +1,58 @@
...
+  const transcriptionFolder = space.properties[CollectionType.typename]?.target;
+  const handleTranscription = async () => {
+    const newTranscription: TranscriptionType = {
+      ...ai.transcription,
+      enabled: !ai.transcription.enabled,
+      lamportTimestamp: ai.transcription.lamportTimestamp! + 1,
+    };
+    let document: DocumentType | undefined;
+    if (!ai.transcription.enabled && !ai.transcription.objectId) {
+      document = (
```
richburdon, line 31: this feels cumbersome; assign the await to a variable first.

packages/plugins/experimental/plugin-calls/src/components/Call/Call.tsx
```diff
@@ -56,32 +62,37 @@ export const Call: FC<ThemedClassName> = ({ classNames }) => {
 
   useBroadcastStatus({ userMedia, peer, updateUserState, identity, pushedTracks, ai, speaking: isSpeaking });
 
+  const transcriptionProps = useHandleTranscription();
```
richburdon, line 65: the number 5 is hardcoded above; if that's indented move it to a const so it's clearer what that it.

richburdon (changes requested):

PR #8624, opened 2025-02-20
content: update plugin meta details

Raymmar:
First draft of plugin meta details


packages/plugins/plugin-chess/src/meta.ts
```diff
@@ -9,7 +9,8 @@ export const CHESS_PLUGIN = 'dxos.org/plugin/chess';
 export const meta = {
   id: CHESS_PLUGIN,
   name: 'Chess',
-  description: 'Play chess.',
+  description: `Chess is a simple plugin which allows you to render an interactive chessboard inside of a plank. This chessboard can be used for collaborative chess games, or for practice experiments with the LLM. `,
```
richburdon, line 12: avoid the term plank (i.e., don't assume knowledge of internal product details).

packages/plugins/plugin-debug/src/meta.ts
```diff
@@ -9,8 +9,9 @@ export const DEBUG_PLUGIN = 'dxos.org/plugin/debug';
 export const meta = {
   id: DEBUG_PLUGIN,
   name: 'Debug',
-  description: 'DXOS debugging tools.',
+  description: `The debug plugin is useful for troubleshooting inside of Composer. You can also use the Debug plugin to create test data inside of sheets or tables or explore pre-built automations. Non-technical users will likely not spend much time here, but this plugin is invaluable for developers and technical users.`,
```
richburdon, line 12: Non-developers

packages/plugins/plugin-map/src/meta.ts
```diff
@@ -9,8 +9,9 @@ export const MAP_PLUGIN = 'dxos.org/plugin/map';
 export const meta = {
   id: MAP_PLUGIN,
   name: 'Maps',
-  description: 'Display objects on maps.',
+  description: `Activate Maps to plot data from your tables on an interactive globe. Just add Lat/Long fields to your schema to see them on a map. You can also plot points on the map while chatting with your AI assistant.`,
```
richburdon, line 12: interactive map or globe.
i think this is an example where we would really benefit from 2 screenshots at least, but for later.

packages/plugins/plugin-markdown/src/meta.ts
```diff
@@ -9,7 +9,12 @@ export const MARKDOWN_PLUGIN = 'dxos.org/plugin/markdown';
 export const meta = {
   id: MARKDOWN_PLUGIN,
   name: 'Markdown',
-  description: 'Text editor supporting extended Markdown.',
+  description: `Composer offers a Modern Markdown editor that is collaborative and fully extensible. It provides rich text editing as well as read only and markdown view.
```
richburdon, line 12: remove "Composer offers" (too flowery).

richburdon (changes requested):
Something isn't right with the PR -- there are lots of code changes that I assume you didn't make.

richburdon (approved):

PR #8629, opened 2025-02-20
feat(table): Detailed edit rows in sidebar

ZaymonFC:
https://github.com/user-attachments/assets/0914d9d2-36cb-4ebc-af96-ac68529c7fc4

### DIFF
--- a/packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
+++ b/packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
@@ -2,7 +2,7 @@
 // Copyright 2025 DXOS.org
 //
 
-import React, { useMemo, useSyncExternalStore } from 'react';
+import React, { useCallback, useMemo, useSyncExternalStore } from 'react';
 
 import { type JsonPath, type EchoSchema, setValue } from '@dxos/echo-schema';
 import { invariant } from '@dxos/invariant';
@@ -15,7 +15,8 @@ import { type ViewType } from '@dxos/schema';
 
 import { TABLE_PLUGIN } from '../meta';
 
-export const useSchema = (space: Space | undefined, typename: string | undefined): EchoSchema | undefined => {
+// TODO(ZaymonFC): Factor this out and use it where we query for schemas with typename.
+const useSchema = (space: Space | undefined, typename: string | undefined): EchoSchema | undefined => {
   const { subscribe, getSchema } = useMemo(() => {
     if (!typename || !space) {
       return {
@@ -43,10 +44,7 @@ export const useSchema = (space: Space | undefined, typename: string | undefined
   return useSyncExternalStore(subscribe, getSchema);
 };
 
-type RowDetailsPanelProps = {
-  objectId: string;
-  view: ViewType;
-};
+type RowDetailsPanelProps = { objectId: string; view: ViewType };
 
 const ObjectDetailsPanel = ({ objectId, view }: RowDetailsPanelProps) => {
   const { t } = useTranslation(TABLE_PLUGIN);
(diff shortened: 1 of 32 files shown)


packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { t } = useTranslation(TABLE_PLUGIN);
+  const selectedRows = useSelectedItems(objectId);
+  const space = getSpace(view);
+  const schema = useSchema(space, view.query?.type);
+  const effectSchema = useMemo(() => schema?.getSchemaSnapshot(), [JSON.stringify(schema?.jsonSchema)]);
+
+  const queriedObjects = useQuery(space, schema ? Filter.schema(schema) : Filter.nothing());
+  const selectedObjects = queriedObjects.filter((obj) => selectedRows.has(obj.id));
+
+  // TODO(ZaymonFC): This is a bit unsophisticated.
```
wittjosiah, line 61: maybe hint at how you think it should be improved?

packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { getSpace, Filter, type Space } from '@dxos/react-client/echo';
+import { useQuery } from '@dxos/react-client/echo';
+import { useTranslation } from '@dxos/react-ui';
+import { useSelectedItems } from '@dxos/react-ui-attention';
+import { Form } from '@dxos/react-ui-form';
+import { type ViewType } from '@dxos/schema';
+
+import { TABLE_PLUGIN } from '../meta';
+
+export const useSchema = (space: Space | undefined, typename: string | undefined): EchoSchema | undefined => {
```
wittjosiah, line 18: todo factor out

packages/ui/react-ui-form/src/hooks/useForm.ts
```diff
@@ -10,6 +10,7 @@ import { invariant } from '@dxos/invariant';
 import { log } from '@dxos/log';
 import { validateSchema, type ValidationError } from '@dxos/schema';
 import { type MaybePromise } from '@dxos/util';
+import { cond } from 'effect/STM';
```
wittjosiah, line 13: unused? missing import?

packages/ui/react-ui-kanban/src/components/Kanban.tsx
```diff
@@ -84,6 +95,7 @@ export const Kanban = ({ model, onAddCard, onRemoveCard }: KanbanProps) => {
                             label={t('card drag handle label')}
                           />
                         </StackItem.DragHandle>
+                        <AttentionGlyph attended={selectedItems.has(card.id)} />
```
wittjosiah, line 98: I know this is a pattern we have elsewhere but I don't think it's helpful as implemented here for a couple reasons:
1. the glyph on its own isn't noticeable enough
2. there's no glyph in the sidebar to pair with it

I could see the glyph being useful at some point here but I'd say in addition to a selection ring for the same reason that we added the selection background colour change in the navtree, it's just much clearer what's going on.

thure: I think we must leave rings as exclusive signifiers of focus, otherwise this will get very confusing.

I think we could also change the background color as an added signifier, especially to remain consistent with the navtree, but it should still come with the relatedness signifier, otherwise we should reconsider using the relatedness signifier at all.

thure: > there's no glyph in the sidebar to pair with it

The panel of the R1 should display such a glyph anyway in order to carry out a design consistent with the rest of the app, I would actually recommend following through on that.

thure: If we don’t have consensus that this should be consistent with the navtree, then it should be consistent with table by using checkboxes.

There is not sufficient cause to implement a third way of signifying `current` state.

wittjosiah: My feedback is that the glyph signifier as demonstrated above is not clear enough as an indication of selection, I don't have too strong an opinion on how that's addressed, ring is just a proposed option.

thure (approved):
Worth revisiting the design of how(/whether) the c11y sidebar is used in this case since I think we need to give real design attention to “master-detail” patterns across the app, but this seems fine for now.

packages/ui/react-ui-kanban/src/components/Kanban.tsx
thure: Agreed that it is too subtle a signifier on its own. I think there are two reasonable options:
- use checkboxes like table
- use a background change and an associated glyph like navtree

Personally I would prefer this use the same signifiers as navtree since I think selection state in table has conflated selection with `current`, which are orthogonal and should be signified separately, i.e. table should also adopt the same pattern as navtree eventually.

ZaymonFC: I get the feeling this entire side-bar editing pattern is a stop-gap measure. Given that we're pushing to get *something* working I think we should go with the current implementation and circle back with more design thought and a more considered ontology later.

The way this is written, it's very orthogonal and doesn't complect either the table or kanban implementations. Very easy to rip back out.

PR #8640, opened 2025-02-21
chore: Remove deprecated mosaic-kanban

ZaymonFC:


wittjosiah (commented):
this whole package should be nuked, I think the only thing still being used anywhere is the `Path` util and that could be factored out

ZaymonFC:
> this whole package should be nuked, I think the only thing still being used anywhere is the `Path` util and that could be factored out

Fair enough. Closing for now.

PR #8651, opened 2025-02-24
feat(kanban): Create kanban from existing type

ZaymonFC:
Adds type selection and initial pivot column options (from available `single-select` fields) during Kanban creation.

https://github.com/user-attachments/assets/50aa1f74-7ce7-421f-9861-194d3f50ade9


thure (approved):
lgtm!

PR #8670, opened 2025-02-26
feat: functions in the compute graph

wittjosiah:


packages/core/echo/echo-schema/src/schema/echo-schema.ts
```diff
@@ -265,18 +265,21 @@ const unwrapOptionality = (property: AST.PropertySignature): AST.PropertySignatu
   } as any;
 };
 
-// TODO(dmaretskyi): Change to `getSnapshot`.
-const unwrapProxy = (jsonSchema: any): any => {
+/**
+ * Returns a non-reactive snapshot of the given live object.
+ */
+// TODO(wittjosiah): Types.
+export const getSnapshot = (jsonSchema: any): any => {
```
dmaretskyi, line 272: pls move out of this file

PR #8675, opened 2025-02-27
feat: Add resize handle to calls grid

richburdon:
https://github.com/user-attachments/assets/8984e749-5429-40b1-a528-b3a0d5dad1d3

### DIFF
--- a/packages/ui/react-ui-dnd/src/components/ResizeHandle.tsx
+++ b/packages/ui/react-ui-dnd/src/components/ResizeHandle.tsx
@@ -77,7 +77,7 @@ export const ResizeHandle = ({
   const client = orientation === 'horizontal' ? 'clientX' : 'clientY';
 
   useLayoutEffect(() => {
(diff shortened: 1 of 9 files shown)


packages/ui/react-ui-dnd/src/components/ResizeHandle.tsx
```diff
@@ -65,52 +72,52 @@ export const ResizeHandle = ({
...
-          dragStartSize.current = nextSize;
-        },
-      });
-    },
-    [
-      // Note that `size` should not be a dependency here since dragging this adjusts the size.
-    ],
-  );
+  useLayoutEffect(() => {
+    if (!buttonRef.current || buttonRef.current.hasAttribute('draggable')) {
```
thure, line 80:
```suggestion
    if (!buttonRef.current) {
```
That predicate was an early misunderstanding that `return draggable` fixes.

thure (approved):
lgtm 👍

PR #8685, opened 2025-02-28
feat: Calls summaries

dmaretskyi:


packages/plugins/experimental/plugin-calls/src/components/TranscriptionContainer.tsx
```diff
@@ -2,24 +2,66 @@
...
   const queue = useQueue<TranscriptBlock>(edge, transcript.queue ? DXN.parse(transcript.queue) : undefined, {
     pollInterval: 1_000,
   });
 
+  // TODO(dmaretskyi): Pending state and errors should be handled by the framework!!!
+  const [isSummarizing, setIsSummarizing] = useState(false);
+  const handleSummarize = useCallback(async () => {
+    setIsSummarizing(true);
+    try {
+      const document = await summarizeTranscript(edge, aiService, transcript);
```
mykola-vrmchk, line 43: It should be done in intent-resolver.ts

PR #8720, opened 2025-03-05
fix: dynamic settings store

wittjosiah:


ZaymonFC (approved):
Looks good. I'll keep an eye out for places where we perform side-effects and don't use `untracked`.

PR #8725, opened 2025-03-05
refactor: add object form capability

wittjosiah:
- factor object form out of metadata
- register schemas from object form removing the need for two different
client schema capabilities
- remove the default form schema, if no schema is provided, immediately
create the object


packages/plugins/plugin-kanban/src/KanbanPlugin.tsx
```diff
@@ -26,18 +27,23 @@ export const KanbanPlugin = () =>
...
+      activate: () =>
+        contributes(
+          SpaceCapabilities.ObjectForm,
+          defineObjectForm({
+            objectSchema: KanbanType,
+            formSchema: CreateKanbanSchema,
+            getIntent: (props, options) => createIntent(KanbanAction.Create, { ...props, space: options.space }),
+          }),
+        ),
     }),
```
ZaymonFC, line 47: Nice!

ZaymonFC (approved):
Awesome ✨

PR #8735, opened 2025-03-06
refactor: devtools deck

wittjosiah:
image


packages/devtools/devtools/src/panels/client/LoggingPanel/LoggingPanel.tsx
```diff
@@ -83,11 +85,20 @@ export const LoggingPanel = () => {
     setLogs((logs) => [...logs.slice(-MAX_LOGS), logEntry]);
   }, [logEntry]);
 
+  const presets = [
```
wittjosiah, line 88: @yarolegovich @dmaretskyi could you add some useful presets here? One in particular that we'd like to see is edge replication logs

richburdon: separately, we've all discussed getting a feed of events from the edge services that we could surface here.

richburdon (approved):
aweome!

richburdon:
i know the bottom of R0 is getting busier, but devtools belongs there rather than up top with the spaces.

PR #8745, opened 2025-03-07
feat(markdown): Add document artifact

ZaymonFC:


packages/plugins/experimental/plugin-transcription/src/components/TranscriptContainer.tsx
```diff
@@ -129,12 +129,32 @@ const summarizeTranscript = async (
...
 // TODO(dmaretskyi): Add example to set consistent structure for the summary.
 const SUMMARIZE_PROMPT = `
-  Create a summary of the transcript provided.
-  Reference specific people by name when possible.
-  Format the summary as a list of key points and takeaways.
-  Include action items in the summary.
-  Use markdown formatting for headings and bullet points.
-  Format the summary as a markdown document without extra comments like "Here is the summary of the transcript:".
+  <GOAL>
+  Create a markdown summary of the transcript provided.
```
richburdon, line 133: we've not been using xml in the prompts (only in the responses), so might be better to just keep it as markdown with level-2 headings.

packages/plugins/experimental/plugin-transcription/src/components/TranscriptContainer.tsx
```diff
@@ -129,12 +129,32 @@ const summarizeTranscript = async (
...
+  - Include short quotes verbatim where appropriate. Especially when concerned with design decisions and problem descriptions.
+  </NOTE_TAKING>
+
+  <TASKS>
+  At the end of the summary include tasks.
+  Extract only the tasks that are:
+  - Directly actionable
+  - Clearly assigned to a person or team (or can easily be inferred)
+  - Strongly implied by the conversation and/or user note (no speculative tasks)
+  - Specific enough that someone reading them would know exactly what to do next
```
richburdon, line 154: require markdown formatting -- i.e., `- [ ] xxx.`
this might be referred to as GFM (github flavored markdown?)

packages/plugins/plugin-markdown/src/capabilities/artifact-definition.ts
```diff
@@ -0,0 +1,84 @@
...
+          id: S.String.annotations({
+            description: 'The fully qualified ID of the document `spaceID:objectID`',
+          }),
+        }),
+        execute: async ({ id }, { extensions }) => {
+          invariant(extensions?.space, 'No space');
+          const space = extensions.space;
+          const { objects: documents } = await space.db.query(Filter.schema(DocumentType)).run();
+          const document = documents.find((doc) => fullyQualifiedId(doc) === id);
+
```
richburdon, line 64: remove space (set then test)

packages/sdk/schema/src/util.ts
```diff
@@ -80,3 +80,23 @@ const toFieldValueType = (type: AST.AST): { format?: FormatEnum; type: TypeEnum
...
+ */
+export const makeSingleSelectAnnotations = (
+  jsonProperty: JsonSchemaType,
+  options: Array<{ id: string; title?: string; color?: string }>,
+) => {
+  jsonProperty.enum = options.map(({ id }) => id);
+  jsonProperty.format = FormatEnum.SingleSelect;
+  jsonProperty.echo = {
+    annotations: {
+      singleSelect: {
```
richburdon, line 95: we're we going to call these `oneOf`?

richburdon (approved):

packages/sdk/schema/src/util.ts
ZaymonFC: @dmaretskyi and I decided on a new format which would play better with the effect-json serializer and not cause us the issues we've been seeing.

I think `singleSelect` inside the annotation works better because JsonSchema still has it's own `oneOf` concept and I'd like to not mix the two. I'm happy to change it over if you guys think it's worth it.

PR #8746, opened 2025-03-07
build: Clean-up tsconfig

dmaretskyi:


.eslintrc.js
```diff
@@ -51,6 +51,10 @@ module.exports = {
     {
       files: '**/*.{tsx,jsx}',
       extends: ['plugin:@dxos/react'],
+      rules: {
```
wittjosiah, line 54: these should go in the react preset

packages/apps/composer-crx/tsconfig.json
```diff
@@ -5,70 +5,65 @@
...
+    "rootDir": "src",
   },
-  "include": [
-    "src"
-  ],
+  "include": ["src"],
   "references": [
     {
-      "path": "../../../tools/vite-plugin-icons"
+      "path": "../../../tools/vite-plugin-icons",
```
wittjosiah, line 15: if this is acceptable, can you update toolbox not to undo it? it creates a lot of diff spam, especially when it gets merged

packages/core/compute/tsconfig.json
```diff
@@ -1,70 +1,57 @@
...
-    "lib": [
-      "ESNext"
-    ],
-    "module": "ESNext",
-    "moduleResolution": "Bundler",
-    "outDir": "./dist/types",
-    "types": [
-      "node"
-    ]
+    "lib": ["ESNext"],
```
wittjosiah, line 4: is this a thing?

packages/sdk/config/tsconfig.json
```diff
@@ -2,33 +2,24 @@
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
     "allowJs": true,
-    "lib": [
-      "DOM",
-      "ESNext"
-    ],
-    "outDir": "./dist/types"
```
wittjosiah, line 9: I've noticed this removed in a bunch of places, is that because it's correctly pulling it from the project.json now?

packages/plugins/plugin-wnfs/tsconfig.json
```diff
@@ -1,83 +1,69 @@
...
   "compilerOptions": {
-    "lib": [
-      "DOM",
-      "ESNext"
-    ],
-    "outDir": "dist/types",
-    "types": [
-      "node"
-    ]
+    "lib": ["DOM", "ESNext"],
```
wittjosiah, line 4: I've noticed this removed in lots of packages, but not others, is there a reason it's needed in some places or could it be removed from them as well?

packages/plugins/plugin-wnfs/src/blockstore.ts
```diff
@@ -3,7 +3,7 @@
 //
 
 import { CarWriter } from '@ipld/car';
-import type { Block } from '@ipld/car/api';
+import type { Block } from '@ipld/car/reader';
```
wittjosiah, line 6: did this package bump or something?

PR #8747, opened 2025-03-08
chore(maps): Selection aware maps

ZaymonFC:
- **Stub sharing selection with map (allow setting multiple context ids).**
- **Fix formatting in tsconfigs**

### DIFF
--- a/packages/plugins/plugin-map/src/components/MapContainer.tsx
+++ b/packages/plugins/plugin-map/src/components/MapContainer.tsx
@@ -7,6 +7,7 @@ import React, { useEffect, useState } from 'react';
 
 import { useQuery, getSpace, useSchema, Filter } from '@dxos/react-client/echo';
 import { useControlledState } from '@dxos/react-ui';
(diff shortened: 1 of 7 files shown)


packages/plugins/plugin-map/src/components/MapContainer.tsx
```diff
@@ -30,6 +31,12 @@ export const MapContainer = ({ role, type: _type = 'map', map, ...props }: MapCo
   const schema = useSchema(space, map?.view?.target?.query.type);
   const rowsForType = useQuery(space, schema ? Filter.schema(schema) : undefined);
 
+  // TODO(burdon): Do something with selected items (ids). (Correlate against `rowsForType`).
+  const selectedItems = useSelectedItems(map!.view!.target!.query.type!);
+  useEffect(() => {
+    console.log('Selected items changed:', selectedItems);
+  }, [selectedItems]);
```
ZaymonFC, line 38: @richburdon, the selection for type is now available. This will show you the selection from any Kanban / Table that shares the typename.a

PR #8753, opened 2025-03-10
feat: calendar demo

richburdon:


packages/experimental/artifact/src/types/tools.ts
```diff
@@ -66,8 +66,11 @@ export const ToolResult = Object.freeze({
 export const Tool = S.Struct({
   /**
    * Unique name.
+   * ^[a-zA-Z0-9_-]{1,64}$
    */
   name: S.String,
+  namespace: S.optional(S.String),
```
dmaretskyi, line 72: add jsdoc for those pls

packages/experimental/assistant/src/ai-service/client.ts
```diff
@@ -12,14 +12,16 @@ import { log } from '@dxos/log';
...
+export interface AIService {
+  exec(request: GenerateRequest): Promise<GenerationStream>;
+}
+
+export interface AIServiceClient extends AIService {
   getSpace(spaceId: SpaceId): Promise<Space>;
   getThread(spaceId: SpaceId, threadId: string): Promise<Thread>;
   getMessages(spaceId: SpaceId, threadId: string): Promise<Message[]>;
   appendMessages(messages: Message[]): Promise<void>;
   updateMessage(spaceId: SpaceId, threadId: string, messageId: string, message: Message): Promise<void>;
```
dmaretskyi, line 24: those CRUD methods are depreacted since we've moved away from storing context in ai-service and moved to queues

packages/experimental/assistant/src/ai-service/types.ts
```diff
@@ -34,25 +34,25 @@ export const GenerateRequest = S.Struct({
...
    */
+  // TODO(burdon): Rename messages.
   history: S.optional(S.Array(Message)),
 
   /**
-   * Tools available for the LLM.
+   * Current request.
    */
-  tools: S.optional(S.Array(Tool).pipe(S.mutable)),
+  prompt: S.optional(Message),
```
dmaretskyi, line 55: does this require changes on the service side to respect `promp`?

packages/experimental/artifact/src/types/util.ts
```diff
@@ -9,20 +9,29 @@ import { toJsonSchema, type JsonSchemaType } from '@dxos/echo-schema';
...
-  description,
-  schema,
-  execute,
-}: DefineToolParams<Params>): Tool => {
+export const parseToolName = (name: string) => {
+  return name.split('_').pop();
+};
+
+export const defineTool = <Params extends S.Schema.AnyNoContext>(
+  namespace: string,
```
dmaretskyi, line 27: intuitively I thought tool ID would be the first argument - that's what I though when looking at examples first, and then at the definition. I would move the namespace into the params, so there's no ambiguity

packages/ui/react-ui-pickers/src/components/HuePicker.tsx
```diff
@@ -20,6 +20,8 @@ export type HuePickerProps = {
 export const HuePicker = (props: ThemedClassName<HuePickerProps>) => {
   const { t } = useTranslation('os');
 
+  console.log('hues', { hues });
```
dmaretskyi, line 23: remove

dmaretskyi (changes requested):

packages/experimental/artifact-testing/src/plugins.tsx
```diff
@@ -62,14 +62,14 @@ export const capabilities: AnyCapability[] = [
...
-    Images:
-    - When presenting an image, you must use an artifact.
-    - Nest the <image> tag inside the <artifact> tag.
-    - Image tags are always self-closing and must contain an id attribute.
-    (Example: <artifact><image id="unique_identifier" prompt="..." /></artifact>)
-    `,
+        - When presenting an image, you must use an artifact.
+        - Nest the <image> tag inside the <artifact> tag.
+        - Image tags are always self-closing and must contain an id attribute.
+          (Example: <artifact><image id="unique_identifier" prompt="..." /></artifact>)
```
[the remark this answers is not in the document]
richburdon: ok. i think this is legacy for testing. there's a bunch of stuff like this.

packages/experimental/artifact/src/types/tools.ts
richburdon: ideally we'd replace the name with these; otherwise it's redundant -- need to discuss.

packages/experimental/assistant/src/ai-service/types.ts
richburdon: not currently; to discuss how to evolve in common.

PR #8769, opened 2025-03-11
feat: store invocation trace queue dxn in props

yarolegovich:
### Details 

Store `invocationTraceQueue` dxn in space properties, where function and workflow invocation events will be stored.


packages/sdk/client-protocol/src/schema.ts
```diff
@@ -13,9 +13,10 @@ export class PropertiesType extends TypedObject({
 })(
   {
     name: S.optional(S.String),
+    invocationTraceQueue: S.optional(DXN),
```
dmaretskyi, line 16:
```suggestion
    invocationTraceQueue: Ref(Expando)
```

store dxns as refs

packages/sdk/client/src/echo/space-list.ts
```diff
@@ -300,7 +301,9 @@ export class SpaceList extends MulticastObservable<Space[]> implements Echo {
     const spaceProxy = this._findProxy(space);
 
     await spaceProxy._databaseInitialized.wait({ timeout: CREATE_SPACE_TIMEOUT });
-    spaceProxy.db.add(create(PropertiesType, meta ?? {}), { placeIn: 'root-doc' });
+    const invocationTraceQueue =
```
dmaretskyi, line 304: I'd rather do it somewhere upstream, such as plugin automation

PR #8779, opened 2025-03-12
feat: presenter syntax highlgihting

richburdon:
- Rewrite of ECHO db explorer graph.


packages/plugins/experimental/plugin-explorer/src/components/Graph/graph-model.ts
```diff
@@ -50,40 +50,51 @@ export class SpaceGraphModel extends GraphModel<EchoGraphNode> {
...
             }
 
-            return { type: 'object', id: object.id, object };
+            return { type: 'object', id: object.id, data: object };
           });
 
           this._graph.links = objects.reduce<GraphLink[]>((links, object) => {
-            const objectSchema = getSchema(object);
+            const objectSchema = getSchema(object); // TODO(burdon): ???
             const typename = getType(object)?.objectId;
```
dmaretskyi, line 79: don't use getType

PR #8790, opened 2025-03-12
fix(table): resolve storybook crash

ZaymonFC:
Fix yet another bug caused by calling hooks with array literals (new reference every render).

### DIFF
--- a/packages/plugins/plugin-table/src/components/TableContainer.tsx
+++ b/packages/plugins/plugin-table/src/components/TableContainer.tsx
@@ -96,7 +96,7 @@ const TableContainer = ({ role, table }: { role?: string; table: TableType }) =>
         }
       }
     },
-    [dispatch, space, schema, table, model, handleThreadCreate, handleInsertRow],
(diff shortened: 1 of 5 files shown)


packages/plugins/plugin-table/src/components/TableContainer.tsx
```diff
@@ -96,7 +96,7 @@ const TableContainer = ({ role, table }: { role?: string; table: TableType }) =>
         }
       }
     },
-    [dispatch, space, schema, table, model, handleThreadCreate, handleInsertRow],
+    [space, schema, table, model, handleThreadCreate, handleInsertRow],
```
wittjosiah, line 99: are space, schema or table used in here either?

wittjosiah (approved):

packages/plugins/plugin-table/src/components/TableContainer.tsx
ZaymonFC: negative!

PR #8791, opened 2025-03-12
feat: view v2

wittjosiah:
todo: `runMigrations` is not available on space proxy db


packages/sdk/schema/src/view.ts
```diff
@@ -76,6 +78,33 @@ export class ViewType extends TypedObject({
...
+    [AST.TitleAnnotationId]: 'Name',
+    [AST.ExamplesAnnotationId]: ['Contact'],
+  }),
+  query: S.Struct({
+    type: S.optional(S.String),
+    sort: S.optional(S.Array(FieldSortType)),
+  }).pipe(S.mutable),
+  schema: S.optional(JsonSchemaType),
+  fields: S.mutable(S.Array(FieldSchema)),
+  metadata: S.optional(S.Record({ key: S.String, value: S.Any }).pipe(S.mutable)),
```
dmaretskyi, line 96: not a huge issue but lets think of a pattern on how to avoid repetiion when versioning schema. e.g. using S.Struct(...).fields and spreading them

dmaretskyi (approved):

packages/sdk/schema/src/view.ts
wittjosiah: agreed

PR #8795, opened 2025-03-12
feat(views): Hidden fields

### LINKED ISSUE #8591: [views] Show and hide fields in views
Adding the ability to **show and hide views** will address key limitations in Composer:

1. Currently, when a user creates a field in one view (such as a *table*), that field doesn't appear in other related views (like *Kanban*) that reference the same type name, despite being part of the base schema.

2. Users have no way to selectively display or hide specific fields across different view types, preventing them from customising which schema fields are visible in each UI representation without permanently altering the underlying data structure.

Fields that are not referenced in the current view field array, but exist in the schema, should show up in the UI as `hidden`.

ZaymonFC:
- Resolves #8591 
---
- [x] Update the view projection to return hidden fields (fields that exist in the schema but not in the view)
- [x] View projection should check if its view field still exists in the schema
  - [x] Normalization pass on view projection construction?
- [x] Check which fields exist in the schema but not in the view and add those to hidden fields on view projection construction
- [x] Distinguish the difference between hiding a view field and deleting it
  - Deletion removes the field from the schema and will *impact* all other views
- [x] Test that snapshotting when deleting works as expected
- [x] View editor
  - Delineate between hiding and deleting
  - Show hidden fields (no rearrange?)
- [x] Fields should have a 'hidden' category
  - This will allow us to preserve field IDs so that if the user toggles hide/unhide in a kanban, the column won't get nuked
  - [x] We will coalesce fields that are there but 'hidden' and fields that are in the underlying schema
  - [x] Get field projections / get field projection shouldn't return hidden fields
- [x] Remove the separator
- [x] Resolve panic when hiding kanban pivot column

### DIFF
--- a/packages/sdk/schema/src/projection.ts
+++ b/packages/sdk/schema/src/projection.ts
@@ -140,7 +140,7 @@ export class ViewProjection {
   hideFieldProjection(fieldId: string): void {
     const index = this._view.fields.findIndex((field) => field.id === fieldId);
     if (index !== -1) {
-      const fieldToHide = this._view.fields[index];
+      const fieldToHide = getSnapshot(this._view.fields[index]);
       this._view.fields.splice(index, 1);
       if (!this._view.hiddenFields) {
         this._view.hiddenFields = [];
@@ -163,7 +163,7 @@ export class ViewProjection {
       const hiddenIndex = this._view.hiddenFields.findIndex((field) => field.path === property);
 
       if (hiddenIndex !== -1) {
-        const fieldToUnhide = this._view.hiddenFields[hiddenIndex];
+        const fieldToUnhide = getSnapshot(this._view.hiddenFields[hiddenIndex]);
         this._view.hiddenFields.splice(hiddenIndex, 1);
         this._view.fields.push(fieldToUnhide);
         return;
--- a/packages/sdk/schema/src/projection.ts
+++ b/packages/sdk/schema/src/projection.ts
@@ -126,32 +126,55 @@ export class ViewProjection {
     const pathsInView = new Set(this._view.fields.map((field) => field.path));
     const schemaProperties = Object.keys(this._schema.jsonSchema.properties ?? {});
 
-    // schema properties that are not in the view and are not 'id'.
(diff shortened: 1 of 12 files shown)


packages/sdk/schema/src/projection.ts
```diff
@@ -112,9 +116,90 @@ export class ViewProjection {
...
+    invariant(this._schema.jsonSchema.properties);
+    const field = this._view.fields.find((field) => field.id === fieldId);
+    if (!field) {
+      return undefined;
+    }
+
+    return this.getFieldProjection(fieldId);
+  }
+
+  private _fieldProjectionsComputed = computed(() =>
```
wittjosiah, line 133: I'd put this at the top of the class with the rest of the class properties I think

packages/sdk/schema/src/projection.ts
```diff
@@ -112,9 +116,90 @@ export class ViewProjection {
...
+    );
+    const hiddenFieldPaths = this._view.hiddenFields?.map((field) => field.path as string) ?? [];
+    return [...new Set([...hiddenSchemaProperties, ...hiddenFieldPaths])].sort();
+  }
+
+  /**
+   * Hides a field from the view by moving it from visible fields to hiddenFields.
+   * This preserves the field's structure and ID for later unhiding.
+   */
+  hideFieldProjection(fieldId: string): void {
```
wittjosiah, line 164: should this and unhide be `untracked`?

packages/sdk/schema/src/projection.ts
```diff
@@ -112,9 +116,90 @@ export class ViewProjection {
...
+      this._view.fields.splice(index, 1);
+      if (!this._view.hiddenFields) {
+        this._view.hiddenFields = [];
+      }
+
+      this._view.hiddenFields.push(fieldToHide);
+    }
+  }
+
+  unhideFieldProjection(property: JsonProp): void {
```
wittjosiah, line 177: show?

wittjosiah (approved):

packages/sdk/schema/src/projection.ts
ZaymonFC: Yep!

PR #8818, opened 2025-03-16
feat: Browser extension

richburdon:


packages/apps/composer-crx/assets/img/icon-512.png
thure, line 1: The aliasing artifacts are getting pretty grisly — is there a chance to change the logo before we go live?

richburdon: yes we have plenty of time.

PR #8831, opened 2025-03-19
chore(schema): Move schema-tool to space-plugin

ZaymonFC:
Acting on @wittjosiah's suggestion this PR moves the schema-tool to the `space-plugin`.


packages/plugins/plugin-space/src/SpacePlugin.tsx
```diff
@@ -178,5 +179,10 @@ export const SpacePlugin = ({
       ),
       activate: SpacesReady,
     }),
+    defineModule({
+      id: `${meta.id}/module/schema-tools`,
+      activatesOn: Events.SetupArtifactDefinition,
+      activate: SchemaTools,
+    }),
```
wittjosiah, line 186: probably just Tools? there's likely other echo tools which will be added here

packages/plugins/plugin-space/package.json
```diff
@@ -63,8 +64,11 @@
     "@dxos/react-ui-stack": "workspace:*",
     "@dxos/react-ui-syntax-highlighter": "workspace:*",
     "@dxos/react-ui-tabs": "workspace:*",
+    "@dxos/react-ui-theme": "workspace:*",
+    "@dxos/schema": "workspace:*",
     "@dxos/shell": "workspace:*",
     "@dxos/util": "workspace:*",
+    "@effect/schema": "^0.75.5",
```
wittjosiah, line 71: this package is deprecated, import schema from `effect`

packages/ui/react-ui-table/src/components/testing.ts
```diff
@@ -75,6 +75,7 @@ export const useSimulator = ({ items, table, insertInterval, updateInterval }: S
       const rowIdx = Math.floor(Math.random() * items.length);
       const fields = table.view?.target?.fields ?? [];
       const columnIdx = Math.floor(Math.random() * fields.length);
+      // TODO(ZaymonFC): ... This is borked.
```
wittjosiah, line 78: What's not working? Due to this pr or already broken?

PR #8837, opened 2025-03-23
feat: persist folded sections locally

wittjosiah:


packages/plugins/experimental/plugin-script/src/components/TypescriptEditor/TypescriptEditor.tsx
```diff
@@ -73,6 +73,7 @@ export const TypescriptEditor = ({
           '.cm-gutters': {
             // Match margin from content.
             marginTop: '16px',
+            background: 'var(--surface-bg)',
```
thure, line 76: Make sure this works with attention, I’ve seen issues before. Let me know if it doesn’t, I’ve been meaning to find a repro so I can fix the issue.

thure (approved):
lgtm ✨

PR #8841, opened 2025-03-24
feat: Queue protocol test bench

dmaretskyi:


packages/core/echo/echo-db/src/queue/sync-prototype.test.ts
```diff
@@ -0,0 +1,593 @@
...
+
+import { describe, expect, test } from 'vitest';
+
+import { log } from '@dxos/log';
+
+/**
+ * # Queue design
+ *
+ * Queue is an ordered sequence of items.
+ * Items can be edited and deleted, but concurrent edits are not merged.
```
yarolegovich, line 13: why do we need to edit items? I thought the idea was to have an append-only log.

packages/core/echo/echo-db/src/queue/sync-prototype.test.ts
```diff
@@ -0,0 +1,593 @@
...
+ * New items have a shape: { seq, actor, data }
+ * Items confirmed by the master sync server have a shape: { globalId, seq, actor, data }
+ * Items that mutate other items have a shape: { globalId?, seq, actor, pred, replace: true, data }
+ * Items that delete other items have a shape: { globalId?, seq, actor, pred, replace: true }
+ *
+ * # Ordering rules
+ *
+ * If both items have globalId set, order by globalId ASC.
+ * If only one item has globalId set, this item should be ordered before.
+ * If no items have globalId set, order by [seq; actor] ASC (lamport timestamp rules).
```
yarolegovich, line 39: lamport timestamp rules means items from different actors not acked by sync server are always considered concurrent? why not order by lexicographic actor id?

packages/core/echo/echo-db/src/queue/sync-prototype.test.ts
```diff
@@ -0,0 +1,593 @@
...
+ *
+ * To edit a previous item the writer would create an item that references the item being edited in its predecessor fields, with replace flag set, and data containing the new data.
+ * Deleting an item is done in the same fashion but the data field is left blank.
+ * These new items that mutate and delete the previous ones are synced to the queue with the regular protocol.
+ *
+ * # Queue compaction
+ *
+ * One of the main design decisions around queues, is that clients can load a subrange of items from disk.
+ * This means that the entire queue might not be in memory at the same time.
+ * Since items that mutate previous items might be located far after the original items, how do we return up-to-date data when the range containing the originals is queried?
```
yarolegovich, line 71: queues are usually read in batches till the end is reached, so that we know the latest state is reconstructed. it's hard to think of a usecase where I would want to read a past range only and then execute some logic based on the data I got. the implication is that I don't need to know whether an item I read was overwritten by a later time, I would just continue reading and find this override, when no items are returned I'll start acting on the state

dmaretskyi: One usecase i'm considering is using a queue as a backend for a discussion thread (e.g. slack, discord), with ability to edit past messages. In that case you might want to query prior range and get the latest state of an item.

packages/core/echo/echo-db/src/queue/sync-prototype.test.ts
dmaretskyi: i'm thinking through situations where this would matter:
If we have peers syncing through a master server then peers should never see concurent messages since everything coming from the sync server has already been ordered by globalId.

Considering how our apps operate with a separate tab thread and a worker thread, you should consider the tab and the worker to be two separate peers (so that tab can generate ids locally for synchronous local optimistic mutations). Now we have 3 peers: tab, worker, server. In this case tab peers could seed concurent messages that don't have the global id set.

I think you're right in the way that to achieve total order we should order by actor id lexicographically. But conceptually those items are still considered concurent. Here's where this is important.

Imagine alice appends A, B, C concurent with bob appending D, E, F

The [seq; actor] ids of the items would be

```
A 1: alice
B: 2:alice
C 3:alice

D: 1:bob
E: 2:bob
F 3:bob
```

Sorted by [seq; actor] they would be ordered as: A, D, B, E, C, F - there's an interleaving problem here.

This is somthing that would need to be adressed for proper P2P sync, but for now it can only occur locally within one device, so we can postpone this.

packages/core/echo/echo-db/src/queue/sync-prototype.test.ts
```diff
@@ -0,0 +1,1692 @@
...
+ * Queue compaction forwards the new data into the original records.
+ * When a new item, that has the replace flag set is pushed into the store, the peer/server will lookup the record from the predecessor field, and update the data field on it.
```
yarolegovich, line 74: data or succ?

PR #8842, opened 2025-03-24
feat: variable height transcript grid

richburdon:


packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
```diff
@@ -169,70 +140,78 @@ export const Transcript: FC<TranscriptProps> = ({ attendableId, ignoreAttention,
...
-  });
-};
+  }, [rows]);
 
-export const formatElapsed = (start: Date, end: Date) => {
-  const { hours, minutes, seconds } = intervalToDuration({ start, end });
-  const pad = (n = 0) => String(n).padStart(2, '0');
-  return `${pad(hours)}:${pad(minutes)}:${pad(seconds)}`;
+  return (
+    <div ref={ref} className='absolute top-0 z-10 invisible'>
```
thure, line 209:
```suggestion
    <div ref={ref} className='fixed top-full invisible' inert='' aria-hidden='true'>
```
This will ensure the browser definitely ignores this.

packages/plugins/experimental/plugin-transcription/src/components/TranscriptContainer.tsx
```diff
@@ -56,16 +60,15 @@ export const TranscriptionContainer: FC<{ transcript: TranscriptType }> = ({ tra
     }
   }, [transcript, edge]);
 
-  // TODO(dmaretskyi): Move action to menu.
   return (
     <StackItem.Content toolbar={false} classNames='relative'>
       <Toolbar.Root classNames='absolute block-start-1 inline-end-1 z-[1] is-min'>
         <IconButton
-          icon='ph--pen--regular'
+          icon='ph--subtitles--regular'
```
thure, line 67: This is the same icon as the type, I think actions and the entity type should have a distinct icon which is why I chose `pen`. If you don’t like `pen`, I’d still recommend choosing something other than `subtitles` since summaries and transcripts are distinct entity types.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
```diff
@@ -169,70 +140,78 @@ export const Transcript: FC<TranscriptProps> = ({ attendableId, ignoreAttention,
...
+  width: number;
+  onUpdate: (rows: TranscriptRow[]) => void;
+}) => {
+  const ref = useRef<HTMLDivElement>(null);
+  useLayoutEffect(() => {
+    const div = ref.current;
+    if (div) {
+      for (let i = 0; i < div.children.length; i++) {
+        const row: HTMLDivElement = div.children[i].children[0] as HTMLDivElement;
+        rows[i].size = row.offsetHeight + cellSpacing;
```
thure, line 201: This really works? More frequently than the arithmetical method?

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
richburdon: what do you mean -- are you saying that this shouldn't work? (it seems to).

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
```diff
@@ -169,70 +140,78 @@ export const Transcript: FC<TranscriptProps> = ({ attendableId, ignoreAttention,
...
+  }, [rows]);
 
-export const formatElapsed = (start: Date, end: Date) => {
-  const { hours, minutes, seconds } = intervalToDuration({ start, end });
-  const pad = (n = 0) => String(n).padStart(2, '0');
-  return `${pad(hours)}:${pad(minutes)}:${pad(seconds)}`;
+  return (
+    <div ref={ref} className='absolute top-0 z-10 invisible'>
+      {rows.map(({ author, text }, i) => (
+        <div key={i} className='p-1 border' style={{ width }}>
```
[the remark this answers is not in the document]
richburdon: is there a fragment we can include?

packages/plugins/experimental/plugin-transcription/src/components/TranscriptContainer.tsx
richburdon: oh ok i missed this. i'll change it.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
richburdon: yes this works. what do you mean arithmetical?
the resize/observer approach didn't work reliably.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
thure: If it works, it’s probably fine to leave it for now, though maybe leave a TODO about ensuring this never receives focus and definitely won’t occlude canonical content.

e.g. `z-10` would make this appear on top of other things, and `absolute` could cause this to become visible if its offset parent happens to have a range of positions, also without `inert` this could receive focus if it has focusable content.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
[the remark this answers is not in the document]
thure: You could define CSS variables in `dx-grid.pcss` that it uses for the cells and also which could be used here.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
thure: Very interesting, wow. I’m glad the method that works also happens to be incredibly simple, here’s hoping it scales as we need it to.

packages/plugins/experimental/plugin-transcription/src/components/Transcript/Transcript.tsx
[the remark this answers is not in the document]
thure: lmk if you’d like me to handle that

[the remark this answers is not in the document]
richburdon: yes please.

Issue #8847, opened 2025-03-25
Queues


dmaretskyi:
## API

- `space.queues.get(dxn): Queue`
- `queue.getRange({ from, to }): Promise<T>` gets range without downloading
- `queue.setReplicationSettings()` updates the settings on how this queue should be replicated to this device
- `queue.download({ from, to }): Stop` actively download and keep refreshing this range.
- `queue.append(item)` add item to queue
- `queue.items` signal that returns data in the queue
- `useQueueRange(queue, { from, to })` - to be used in react components that require a specific range of data in the queue

## 

- [x] Sync protocol spec
- [ ] API update
- [ ] Locally preserving data

PR #8848, opened 2025-03-25
feat: Queues api update

dmaretskyi:
https://github.com/dxos/dxos/issues/8847


packages/core/echo/echo-db/src/queue/queue-service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Service for managing queues.
+ */
+export interface QueuesService {
+  queryQueue(subspaceTag: string, spaceId: SpaceId, query: QueueQuery): Promise<QueryResult>;
+
+  insertIntoQueue(subspaceTag: string, spaceId: SpaceId, queueId: ObjectId, objects: unknown[]): Promise<void>;
+
+  deleteFromQueue(subspaceTag: string, spaceId: SpaceId, queueId: ObjectId, objectIds: ObjectId[]): Promise<void>;
```
wittjosiah, line 18: it seems odd to me that these apis don't align with the ones on `Queue` which call them

packages/core/echo/echo-db/src/queue/queue-service.ts
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error('Not available.');
+  }
+
+  insertIntoQueue(subspaceTag: string, spaceId: SpaceId, queueId: ObjectId, objects: unknown[]): Promise<void> {
+    throw new Error('Not available.');
+  }
+
+  deleteFromQueue(subspaceTag: string, spaceId: SpaceId, queueId: ObjectId, objectIds: ObjectId[]): Promise<void> {
+    throw new Error('Not available.');
+  }
```
wittjosiah, line 54: once these write locally first, wouldn't there be no difference? just difference in latency of when items are pushed to edge?

dmaretskyi: This code is far from final form, so don't read too much into it. I think eventual queues will go through the worker like ECHO-automerge

PR #8849, opened 2025-03-25
refactor: calls -> meetings

### LINKED ISSUE #8853: Meeting bugs & polish
- [x] The toolbars in meetings aren’t getting attention
- [x] The toolbars are overlapping the plank resize handle
- [x] Transcript layout causes tabs to overflow (#8849)
- [x] Transcript is duplicating messages (#8849)
- [ ] De-conflict summary tab and summarize action

wittjosiah:
- resolves #8853
- rename the calls plugin to meeting plugin
- add channel activities
- make meeting a channel activity
- add meeting data type
- replace global call context with call manager capability
- sidebar for active meeting
- ui for meeting notes

https://github.com/user-attachments/assets/15e587d5-9071-4b15-9a02-e99efc6d2fe1

TODO:
- transcription capability
- make transcription non-experimental
- use meeting notes in the summary prompt


thure:
> make meeting a channel activity

Is that what that camera iconOnly button is doing? If so, I recommend displaying the full label (there are no other actions so there is plenty of room) and wording it as a CTA e.g. “Start video meeting” or similar.

thure:
Does this make the video meeting mutually exclusive with the chat thread? Users will expect to be able to do both since you can do both in most similar experiences in the ecosystem, so I think we should implement this in a way that supports multitasking.

That can come in a follow-up but I think we should log and prioritize the issue.

wittjosiah:
@thure yeah sorry I should have included that in the todos. This is 100% not final UI, just trying to limit scope and not rewrite more than necessary immediately.

Follow-ups will include at least:
1. Unifying thread container components
2. Using plank groups for channel activities

richburdon:
> Does this make the video meeting mutually exclusive with the chat thread? Users will expect to be able to do both since you can do both in most similar experiences in the ecosystem, so I think we should implement this in a way that supports multitasking.
> 
> That can come in a follow-up (I’d understand wanting to use companion planks for this); I think we should log and prioritize the issue.

+1 we must be able to chat, participate in the video -- and view the transcript at the same time. in fact all activities related to the meeting.

thure (approved):
This is ready to go, I think, pending any other items you wanted to take care of @wittjosiah.

PR #8857, opened 2025-03-26
fix: Remove `lit-grid` from `tsconfig.paths.json`

ZaymonFC:


ZaymonFC:
CC @dmaretskyi @thure

PR #8869, opened 2025-03-29
feat: outliner plugin redux

richburdon:
Outline and Journal.

Screenshot 2025-03-31 at 11 14 34 PM
Screenshot 2025-03-31 at 11 13 49 PM


packages/plugins/experimental/plugin-outliner/src/types/types.ts
```diff
@@ -1,28 +1,49 @@
...
+export class JournalEntryType extends TypedObject({ typename: 'dxos.org/type/JournalEntry', version: '0.1.0' })({
+  date: S.Date,
+  root: TreeNodeType,
+}) {}
+
+export class JournalType extends TypedObject({ typename: 'dxos.org/type/JournalType', version: '0.1.0' })({
+  name: S.optional(S.String),
+
+  // TODO(burdon): Refs vs. objects?
+  entries: S.mutable(S.Array(JournalEntryType)),
```
dmaretskyi, line 48: Can't have an echo object be embedded in another typed object (we can discuss how to allow that separatelly). This either needs to be a ref or JournalEntryType to not be an echo object.

richburdon: the storybook throws: not an ECHO object error.
http://localhost:9009/?path=/story/plugins-plugin-outliner-outliner--large

PR #8880, opened 2025-04-01
feat: edge oauth integration

yarolegovich:
### Details

* Updated `useEdgeClient` to create a client capable of passing edge auth.
* Added `initiateOAuthFlow` to edge client.
* Integrated `plugin-token-manager` with edge oauth.

OAuth flow works in the following way:
1. Composer creates an `AccessToken` object, but doesn't add it to space immediately.
2. Composer makes a call to `/oauth/initiate` passing auth provider enum, desired scopes, access token `objectId` and `spaceId` where the token will later be stored.
3. Edge stores auth context (spaceId, accessTokenId, origin), generates a provider-specific auth url, and returns it to the caller. The url contains a passthrough value used to locate the stored context.
4. Composer the received url in a popup window.
5. User decision (grant or not) gets passed to edge `/oauth/callback` url which returns a simple self-closing page that uses `postMessage` to pass result to `window.opener` (composer).
6. If a successful response is received `composer` sets the received token on the object created in step 1 and adds it to a space.


packages/plugins/plugin-token-manager/src/components/OAuthPresetSelector.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const edgeUrl = new URL(edgeClient.baseUrl);
+
+    const listener = (event: MessageEvent) => {
+      if (event.origin === edgeUrl.origin) {
+        const data = event.data as OAuthFlowResult;
+        if (data.success) {
+          const token = tokenMap.get(data.accessTokenId);
+          if (token) {
+            token.token = data.accessToken;
+            dispatch(createIntent(SpaceAction.AddObject, { object: token, target: space })).catch((e) => log.catch(e));
```
wittjosiah, line 44: does lint not require this to be awaited?

wittjosiah, line 44: specify `hidden: true` to ensure the tokens are not added to the root space collection (just landed this option on `main`)

packages/sdk/client/package.json
```diff
@@ -25,6 +25,11 @@
       "browser": "./dist/lib/browser/halo/index.mjs",
       "node": "./dist/lib/node-esm/halo/index.mjs"
     },
+    "./edge": {
```
wittjosiah, line 28: we should think about how edge is exposed from/integrated with client overall, cc @dmaretskyi

wittjosiah (approved):

packages/plugins/plugin-token-manager/src/components/OAuthPresetSelector.tsx
yarolegovich: if there's a catch it doesn't

PR #8885, opened 2025-04-02
refactor: unified message type

wittjosiah:


packages/sdk/schema/src/common/message.ts
```diff
@@ -0,0 +1,148 @@
...
+}) {}
+
+enum MessageV1State {
+  NONE = 0,
+  ARCHIVED = 1,
+  DELETED = 2,
+  SPAM = 3,
+}
+
+export class MessageTypeV1 extends TypedObject({ typename: 'dxos.org/type/Message', version: '0.1.0' })({
```
dmaretskyi, line 120: the EchoObject transformer should be widely supported by this point

packages/sdk/schema/src/common/message.ts
```diff
@@ -0,0 +1,148 @@
...
+  created: S.String.annotations({
+    description: 'ISO date string when the message was sent.',
+  }),
+  sender: ActorSchema.annotations({
+    description: 'Identity of the message sender.',
+  }),
+  blocks: S.Array(MessageContentBlock).annotations({
+    description: 'Inline content of the message.',
+  }),
+  attachments: S.optional(
```
dmaretskyi, line 97: why noy unify with blocks?

packages/sdk/schema/src/common/message.ts
wittjosiah: I tried, build fails if I use for certain cases currently so I've reverted those cases back to TypedObject for now

packages/sdk/schema/src/common/message.ts
wittjosiah: something like this?
```ts
export const AttachmentContentBlock = S.extend(
  AbstractContentBlock,
  S.Struct({
    type: S.Literal('attachment'),
    attachment: Ref(Expando),
  }),
).pipe(S.mutable);
export type AttachmentContentBlock = S.Schema.Type<typeof AttachmentContentBlock>;
```

thure (approved):
lgtm!

PR #8889, opened 2025-04-02
feat: select static schema for table

richburdon:


packages/core/echo/echo-db/src/proxy-db/echo-schema-registry.ts
```diff
@@ -278,7 +278,7 @@ export class EchoSchemaRegistry extends Resource implements SchemaRegistry {
   // TODO(dmaretskyi): Figure out how to migrate the usages to the async `register` method.
   private _addSchema(schema: S.Schema.AnyNoContext): EchoSchema {
     if (schema instanceof EchoSchema) {
-      schema = schema.getSchemaSnapshot().annotations({
```
dmaretskyi, line 281: dont agree, as this creates a snapshot from the current state

packages/sdk/react-client/src/echo/useSchema.ts
```diff
@@ -19,16 +39,25 @@ export const useSchema = (space: Space | undefined, typename: string | undefined
...
+      return {
+        subscribe: () => () => {},
+        getSchema: () => new ImmutableSchema(staticSchema),
+      };
+    }
+
     const query = space.db.schemaRegistry.query({ typename });
     const initialResult = query.runSync()[0];
-    let currentSchema = initialResult;
+    let currentSchema = initialResult as BaseSchema;
```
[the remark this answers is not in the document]
richburdon: yes although i don't fully understand why it's required. i changed it to which is safer.

```ts
    const initialResult: BaseSchema = query.runSync()[0];
    let currentSchema = initialResult;
```

packages/core/echo/echo-db/src/proxy-db/echo-schema-registry.ts
richburdon: agree?

PR #8898, opened 2025-04-03
refactor: use queue for channel messages

wittjosiah:


packages/plugins/plugin-deck/src/components/DeckLayout/Plank.tsx
```diff
@@ -189,7 +191,7 @@ const SplitFrame = ({ children }: PropsWithChildren<{}>) => {
   return (
     <div
       role='none'
-      className={mx('grid grid-cols-[1fr_1fr] absolute inset-0', railGridHorizontal, mainIntrinsicSize)}
+      className={mx('grid grid-cols-[1fr_1fr] absolute inset-0 min-bs-0', railGridHorizontal, mainIntrinsicSize)}
```
thure, line 194: Did this not do anything? We can remove it if it had no effect, but it'll do no harm to leave in-place.

thure (approved):
fantastic ✨

packages/plugins/plugin-deck/src/components/DeckLayout/Plank.tsx
wittjosiah: oh right, yeah I forgot about this, will revert

PR #8902, opened 2025-04-05
refactor: separate meeting from channel

wittjosiah:
- removes channel activities and makes meeting a primary object
- makes meeting extensible with artifacts, starting with notes (outline), transcript and summary (document)
- decouples transcription from calls via generic activities for the swarm data
- allow all meeting artifacts to be serialized into the summarization prompt
- specify tabs for meetings via graph builder as companion nodes which the expectation to integrate with companion planks work being done currently which will ensure they do not show up in the navtree

TODO:
- track active meetings


packages/plugins/experimental/plugin-transcription/src/capabilities/app-graph-builder.ts
```diff
@@ -3,28 +3,150 @@
...
+            },
+            properties: {
+              label: state.enabled
+                ? ['stop transcription label', { ns: TRANSCRIPTION_PLUGIN }]
+                : ['start transcription label', { ns: TRANSCRIPTION_PLUGIN }],
+              icon: 'ph--record--regular',
+              // TODO(wittjosiah): Disambiguate from navtree toolbar and content toolbar.
+              disposition: 'toolbar',
+              // TODO(wittjosiah): Declare a semantic color.
+              classNames: state.enabled ? 'text-red-500' : '',
```
thure, line 92: Use `text-errorText` — these are now defined in `semems-hue.ts`

packages/plugins/plugin-meeting/src/components/Lobby/Lobby.tsx
```diff
@@ -48,9 +56,8 @@ export const Lobby: FC<LobbyProps> = ({ classNames, roomId }) => {
 
       <Toolbar.Root classNames='justify-between'>
         <IconButton variant='primary' label={t('join call')} onClick={handleJoin} icon='ph--phone-incoming--regular' />
-        {/* <div className='grow text-sm text-subdued'>
-          {sessionError ?? `${numUsers} ${numUsers === 1 ? t('lobby participant') : t('lobby participants')}`}
-        </div> */}
+        {count !== undefined && <div className='text-sm text-subdued'>{t('lobby participants', { count })}</div>}
+        <div className='grow' />
```
thure, line 60:
```suggestion
        <Toolbar.Separator variant='gap' />
```

thure (approved):
Looks great ✨

packages/plugins/experimental/plugin-transcription/src/capabilities/app-graph-builder.ts
wittjosiah: any issue with making an equivalent one called "active"? it feels wrong/confusing to add an error class to something that's not an error

thure: No, you could mint that as an alias alongside 'error' if you like, then that way if it should look different from error it’s trivial to do that.

wittjosiah: active is definitely overloaded... but this is more general than recording. The other actions like screen sharing and raised hand also use this sort of styling

thure: I still think we’d benefit from using a different term even if it needs to apply to all these things… furthermore I don’t think these should be a shade of red when active…

thure: The term could use a qualifier on active, `activeInCall` perhaps.

wittjosiah: all the more reason to mint a semantic token to make it easier to change later :)

packages/plugins/experimental/plugin-transcription/src/transcriber/transcription-manager.ts
```diff
@@ -11,8 +11,8 @@ import { DXN } from '@dxos/keys';
 import { type EdgeHttpClient } from '@dxos/react-edge-client';
 import { type HuePalette } from '@dxos/react-ui-theme';
 
-import { MediaStreamRecorder, Transcriber } from './transcriber';
-import { TranscriptBlock, type TranscriptSegment } from './types';
+import { MediaStreamRecorder, Transcriber } from '.';
```
richburdon, line 14: remove `.` imports

packages/plugins/experimental/plugin-transcription/src/capabilities/app-graph-builder.ts
```diff
@@ -3,28 +3,148 @@
...
-      id: `${TRANSCRIPTION_PLUGIN}/transcription`,
-      filter: (node): node is Node<TranscriptType> => node.data instanceof TranscriptType,
-      actions: () => [
-        {
-          id: `${TRANSCRIPTION_PLUGIN}/transcription`,
-          data: async () => {
-            const { dispatchPromise: dispatch } = context.requestCapability(Capabilities.IntentDispatcher);
-            await dispatch(createIntent(TranscriptionAction.Summarize));
+      id: `${TRANSCRIPTION_PLUGIN}/meeting-transcript`,
+      filter: (node): node is Node<MeetingType> => node.data instanceof MeetingType,
```
richburdon, line 63: Does `S.is(MeetingType)(node.data)` work?

packages/plugins/plugin-meeting/src/components/CallContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export type CallContainerProps = {
+  meeting?: MeetingType;
+  roomId?: string;
+};
+
+export const CallContainer: FC<CallContainerProps> = ({ meeting, roomId: _roomId }) => {
+  const call = useCapability(MeetingCapabilities.CallManager);
+  const roomId = meeting ? fullyQualifiedId(meeting) : _roomId;
+  invariant(roomId);
```
richburdon, line 26: avoid invariants in react components main function (i.e., move inside hook callbacks)?

packages/plugins/plugin-meeting/src/components/Lobby/Lobby.tsx
```diff
@@ -35,10 +33,20 @@ export const Lobby: FC<LobbyProps> = ({ classNames, roomId }) => {
...
     }
-  }, [joinSound, roomId, call.joined, call.leave, call.setRoomId, call.join]);
+  }, [joinSound, roomId, call.joined, call, onJoin]);
+
+  // TODO(wittjosiah): Leaving the room doesn't remove you from the swarm.
+  useEffect(() => {
+    void call.peek(roomId).then((count) => setCount(count));
+    const interval = setInterval(() => {
+      void call.peek(roomId).then((count) => setCount(count));
+    }, 1000);
```
richburdon, line 47: at least move const to file-local const at top of files (self documenting).

packages/plugins/plugin-meeting/src/translations.ts
```diff
@@ -3,30 +3,35 @@
...
         'create summary message': 'A summary doesn’t exist for this meeting yet. Create one now?',
 
         'join call': 'Join',
         'leave call': 'Leave',
 
-        'lobby participant': 'participant',
-        'lobby participants': 'participants',
+        'lobby participants_zero': 'No participants',
+        'lobby participants_one': '1 participant',
+        'lobby participants_other': '{{count}} participants',
```
richburdon, line 34: fancy

packages/plugins/plugin-meeting/src/types/schema.ts
```diff
@@ -4,25 +4,40 @@
...
+      key: S.String,
+      value: Ref(Expando),
+    }),
+  ),
 });
 
 export class MeetingType extends TypedObject({
   typename: 'dxos.org/type/Meeting',
-  version: '0.1.0',
+  version: '0.2.0',
```
richburdon, line 42: while plugins are in experimental do we need to do major bumps?

richburdon (approved):
amazing

packages/plugins/plugin-meeting/src/types/schema.ts
wittjosiah: we have existing meeting objects in our production spaces because we're using them on labs now, so it's necessary. I don't think we need to be shy about version bumps when they're necessary

packages/plugins/experimental/plugin-transcription/src/capabilities/app-graph-builder.ts
wittjosiah: that would be a structural check, I'm assuming that this will change to `Echo.isInstanceOf(MeetingType, node.data)` once we finish migrating to the new apis for all the schemas

packages/sdk/client/src/client/client.ts
```diff
@@ -213,6 +213,14 @@ export class Client {
     return this._runtime.mesh;
   }
 
+  /**
+   * EDGE client.
+   */
+  get edge(): EdgeHttpClient {
```
mykola-vrmchk, line 219: This is a leaky abstraction, I believe Dima added _edgeClient to Client as a quick hack. It will be moved into  ClientServices in the near future, so it is better to not expose it like that. cc @dmaretskyi

wittjosiah: my impression was actually the opposite of that: that edge client was already in client services but needed to be exposed on the client which is why it was added here.

The reason for this is that currently we have this `useEdgeClient` hook which is creating new edge clients all over the place, having it in client at least centralizes this so there's only two (one on frontend, one in services). If we wanted to take it a step further we could then replace the edge on the frontend client with a proxy, but maintain the interface, however I don't think that should stop us from moving usage from separate edge clients to this one in the short term.

packages/plugins/experimental/plugin-transcription/src/types/schema.ts
```diff
@@ -18,7 +18,7 @@ export const TranscriptSchema = S.Struct({
    * Queue containing TranscriptBlock objects.
    */
   // TODO(wittjosiah): Should be a ref to a queue.
-  queue: S.optional(S.String),
+  queue: Ref(Expando),
```
mykola-vrmchk, line 21: Could it cause any problems, if ECHO tries to load it?

packages/plugins/plugin-meeting/src/state/call-manager.ts
```diff
@@ -25,6 +23,11 @@ export type GlobalState = {
  * Top level manager for call state.
  */
 export class CallManager extends Resource {
+  public readonly callStateUpdated = new Event<CallState>();
+  // TODO(wittjosiah): Consolidate isSpeaking into the MediaState type.
+  public readonly mediaStateUpdated = new Event<[MediaState, boolean]>();
+  public readonly left = new Event<string>();
```
mykola-vrmchk, line 29: Is it used anywhere?

packages/plugins/experimental/plugin-transcription/src/types/schema.ts
wittjosiah: yes, if you try to call `.target` on this ref, it throws. However this is the preferred way to reference queues in ECHO right now as it best represents the structure of the data and the expectation is that the api will evolve to handle this case better. We've used this case in a number of other places already

packages/sdk/client/src/client/client.ts
mykola-vrmchk: Yes, it is better than what we had before. Could you add an experimental/deprecated notice in JSDoc?

mykola-vrmchk (approved):

PR #8908, opened 2025-04-07
feat: `lit-ui`, Avatar refactor

thure:
This PR:
- begins a resolution of #8878.
- refactors Avatar to be rendered by a new `dx-avatar` webcomponent exported by a new `lit-ui` package
- clears a lot of tech debt related to Avatar
- makes adjustments to the tokens applied by `dx-avatar`


packages/ui/lit-ui/src/defs.ts
```diff
@@ -0,0 +1,43 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+// TODO(thure): I was unable to bring this in from `react-ui-types` because toolbox.ts would not acknowledge
+//  `"@dxos/lit-ui": [ "packages/ui/lit-ui/src/index.ts" ]` in tsconfig.paths.json. How is this meant to work? How is it
+//  okay with `react-hooks`?
```
wittjosiah, line 7: because `lit-grid` was breaking in tsconfig-paths, `toolbox.json` is configured to only take `react-* packages from the `ui` dir

packages/apps/composer-app/vite.config.ts
```diff
@@ -32,7 +33,7 @@ const isFalse = (str?: string) => str === 'false' || str === '0';
  */
 export default defineConfig((env) => ({
   // Vitest config.
-  test: baseConfig({ cwd: __dirname })['test'] as any,
+  test: mergeConfig(baseConfig({ cwd: __dirname }), defineConfig({ test: { environment: 'jsdom' } }))['test'] as any,
```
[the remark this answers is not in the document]
thure: It fails because `lit-ui` is in its stack (because Avatar is part of Shell), which expends `HTMLElement` to be defined. Certainly amenable to other ways of solving the problem, this was just the easiest way to do so.

packages/ui/lit-ui/src/defs.ts
thure: Okay — shall we leave this this way for now then?

packages/apps/composer-app/vite.config.ts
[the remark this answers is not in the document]
wittjosiah: maybe just leave a todo that explains why it's needed so we know what to look at when we come in to clean it up later

packages/ui/lit-ui/src/defs.ts
wittjosiah: you can update the toolbox.json to include lit-ui and then I think it should work as you expected?

thure: That actually didn’t work when I tried to do that — that’s what I mention in the comment.

PR #8910, opened 2025-04-07
fix: AI assistant fixes

dmaretskyi:


packages/experimental/artifact-testing/src/plugins.tsx
```diff
@@ -62,7 +62,7 @@ export const capabilities: AnyCapability[] = [
   contributes(
     Capabilities.ArtifactDefinition,
     defineArtifact({
-      id: 'dxos.org/plugin/image',
+      id: 'artifact:dxos.org/plugin/image',
```
dmaretskyi, line 65: TODO: rename to blueprints

PR #8911, opened 2025-04-07
chore: Refactor telemetry.

richburdon:
- [x] Fix basic types for page/track events.
- [x] Consolidate track actions to single user event (one table).
- [x] Common base class for node/browser (normalize event logging).


docs/src/plugins/telemetry/client-config.ts
```diff
@@ -27,7 +27,9 @@ export default defineClientConfig({
     router.afterEach((to, from, failure) => {
       if (!isNavigationFailure(failure)) {
         observability.page({
+          name: to.path,
```
wittjosiah, line 30: do page events need a name?

packages/sdk/observability/src/segment/types.ts
```diff
@@ -2,26 +2,48 @@
...
+export type TrackOptions = CommonOptions & {
+  event?: Event;
+  action: string;
 };
+
+/**
+ * Each event will be mapped to a different Postgres table via segment.
+ */
+export enum Event {
+  USER = 'user',
```
wittjosiah, line 48: I would say that this event https://github.com/dxos/dxos/pull/8911/files#diff-d457dd0159c665b772f2e7a6a8e0697e212cb9f4c0f8f15ab3ff80f3bb4abc59R410-R414 is not a user event, however iirc its also super spammy, so we maybe want to disable it anyways and consider another way to collect that

packages/plugins/plugin-observability/src/capabilities/intents-resolver.ts
wittjosiah, line 1: it looks like this file should have been renamed but was deleted

packages/sdk/observability/src/observability.ts
```diff
@@ -62,6 +62,19 @@ export type ObservabilityOptions = {
...
+ *
+ * Segment:
+ * https://app.segment.com/dxos/sources/composer-app/debugger
+ *
+ * Testing:
+ * https://app.segment.com/dxos/sources/composer-app/settings/keys
+ * set DX_TELEMETRY_API_KEY for local dev.
+ *
+ * Sentry:
+ * https://sentry.io/organizations/dxos/issues
```
wittjosiah, line 74: fyi set DX_SENTRY_DESTINATION for local dev

packages/sdk/observability/src/segment/browser.ts
```diff
@@ -6,20 +6,18 @@ import snippet from '@segment/snippet';
 
 import { log } from '@dxos/log';
 
-import type { EventOptions, SegmentTelemetryOptions, PageOptions } from './types';
+import { AbstractSegmentTelemetry } from './base';
```
wittjosiah, line 9: this class doesn't seem to be included in this PR

PR #8915, opened 2025-04-07
chore: max container sememe and components layer

richburdon:
- [x] add container-max-width class and sememe.
- [x] move outliner out of experimental and fix journal/outliner callbacks.


packages/apps/composer-crx/src/options.tsx
```diff
@@ -14,7 +14,7 @@ import { Container, Options } from './components';
 const Root = () => {
   return (
     <Container classNames='absolute inset-0 flex justify-center overflow-hidden bg-modalSurface'>
-      <div className='flex flex-col max-w-[50rem] grow'>
+      <div className='container-max-width'>
```
thure, line 17: Love to see it ✨

thure (approved):
Looks great, thanks a lot for pairing on the tokens ✨

PR #8922, opened 2025-04-08
feat(scripts): Invocation explorer

ZaymonFC:
**Companion plank**
image
**Devtools**
image


packages/devtools/devtools/src/panels/edge/InvocationTracePanel/ExceptionPanel.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <div
+            key={`log-${index}`}
+            className={mx('mb-2 border border-red-200 dark:border-red-900 rounded overflow-hidden')}
+          >
+            <div className={mx('p-2')}>
+              <div className={mx('flex justify-between items-start')}>
+                <div className={mx('font-medium')}>{errorName}</div>
+                <div className={mx('text-xs font-mono opacity-80')}>{time}</div>
+              </div>
+              <div className={mx('mt-1 text-xs font-mono whitespace-pre-wrap')}>{errorMessage}</div>
```
thure, line 68: `mx` definitely should not be called unless needed — these can just be strings since no merging is necessary.

Massive div-itis, probably too much and/or needs semantics, e.g. any divs need a role so we know what to expect. If this was moved from elsewhere, not necessary for you to clean up, just mentioning.

thure (approved):
lgtm!

packages/devtools/devtools/src/panels/edge/InvocationTracePanel/ExceptionPanel.tsx
ZaymonFC: Good call!

This was originally only dev-tools facing so I was blasting through it.

PR #8927, opened 2025-04-10
test: Review ECHO API

richburdon:
## Priorities
- [x] Add ECHO Namespace
- [ ] Review proposals in `api.test` and `api.test.ts`

## API issues
- [ ] Build error when changing from TypedObject using Ref$: Type 'Task' does not satisfy the constraint 'WithId' (see echo-schema/testing/schema.ts)
- [ ] Rename ObjectAnnotation => TypeAnnotation?
- [ ] Remove TypedObject? Currently second arg is `options` but these are not available in the pipe(EchoObject) althernative?
- [ ] What is the pipe alternative for TypedRelation?
- [ ] Reconcile all schema variants (Base, Immutable, TypedObject, EchoObject, etc.)
- [ ] Consolidate getters (getType, getSchema, getTypename, getSchemaTypename, etc.)
- [ ] ReactiveObject should specify `id` property? Reconcile BaseObject, ReactiveObject, HasId, WithId, etc.
- [ ] Can we us S.is(MyType) to detect objects with our types system? (Branding?)

## Misc
- [ ] Remove legacy types (@wittjosiah)


packages/common/keys/src/dxn.ts
```diff
@@ -138,7 +138,7 @@ export class DXN {
     return this.#parts;
   }
 
-  toTypename() {
+  get typename() {
```
dmaretskyi, line 141: I don't like having getters that could fail.

packages/core/echo/echo-db/src/api.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    typename: 'example.com/type/Contact',
+    version: '0.1.0',
+  }),
+);
+
+interface Contact extends S.Schema.Type<typeof Contact> {}
+
+const Message = S.Struct({
+  // TODO(burdon): Support S.Date; Custom Timestamp (with defaults).
+  // TODO(burdon): Support defaults (update create and createStatic).
```
dmaretskyi, line 61: Defaults should be applied on read, not on write, to support schema migrations with new properties and default values.

richburdon: i think defaults should be applied when the object is constructed -- and when read from disk (for schema updates).

.vscode/settings.json
```diff
@@ -40,6 +40,12 @@
     "build-cache.json": "jsonc",
     "command-versions.json": "jsonc"
   },
+  "files.exclude": {
```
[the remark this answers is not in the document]
richburdon: i'm not sure; my IDE was getting slow and gpt suggested doing this. not sure if it made any difference.

PR #8933, opened 2025-04-11
fix(react-client): prevent `useQueue` zombie polling

ZaymonFC:
The `useQueue` hook had a subtle bug where polling could continue indefinitely after component unmount. This happens when:

1. The component unmounts during an active `queue.refresh()` call
2. The cleanup function runs and clears the current timeout
3. The `refresh()` promise `finally()` handler completes afterward
4. A new timeout gets scheduled despite component unmount which leaks the timeout.

This PR fixes the issue by:
- Adding a ref to track component mounted state
- Checking this mounted state before scheduling new polls in the `finally()` handler
- Ensuring all polling stops immediately when the component unmounts

The key change is the conditional check in the `finally()` handler that only schedules the next poll if the component is still mounted.


packages/sdk/react-client/src/echo/useQueue.ts
```diff
@@ -44,17 +44,28 @@ export const useQueue = <T extends BaseEchoObject>(
 
   useEffect(() => {
     let timeout: NodeJS.Timeout;
-    if (options.pollInterval) {
-      const poll = () => {
-        void queue?.refresh().finally(() => {
```
ZaymonFC, line 49:
```ts
void queue?.refresh().finally(() => {
```

This was being invoked after react called the cleanup function.

wittjosiah (approved):

PR #8940, opened 2025-04-12
feat(script-templates): Queue consumption, contact extraction template

ZaymonFC:


packages/plugins/experimental/plugin-script/src/templates/queue.ts
```diff
@@ -0,0 +1,149 @@
...
+  inputSchema: S.Any,
+  outputSchema: S.Struct({ success: S.Boolean }),
+  handler: ({ event, context: { space } }: any) =>
+    Effect.gen(function* () {
+      console.log('Starting queue function', JSON.stringify(event.data));
+
+      console.log('Upserting contact stored schema');
+      const [contactStoredSchema] = yield* Effect.tryPromise({
+        try: () => space.db.schemaRegistry.register([ContactType]),
+        catch: (e: any) => e,
```
wittjosiah, line 24: we should try excluding this again because afaik it should be the default and it's not clean to look at. iirc @yarolegovich added this originally because we were having trouble with errors in functions, but I don't think it should be necessary unless we actually want to do something with the error

packages/plugins/experimental/plugin-script/src/templates/queue.ts
```diff
@@ -0,0 +1,149 @@
...
+      const addedObject = space.db.add(contact);
+      console.log('Added contact to space', JSON.stringify(addedObject));
+
+      yield* Effect.tryPromise({
+        try: () => space.db.flush(),
+        catch: (e: any) => e,
+      });
+
+      return { success: true };
+    }).pipe(Effect.provide(FetchHttpClient.layer)),
```
wittjosiah, line 55: you can remove this provide, I don't think you're using it

packages/plugins/experimental/plugin-script/src/templates/queue.ts
```diff
@@ -0,0 +1,149 @@
...
+// Types
+//
+
+const ContactType = S.Struct({
+  name: S.optional(S.String),
+  email: S.optional(S.String),
+}).pipe(EchoObject('example.org/TempContactType', '0.1.0'));
+type ContactType = S.Schema.Type<typeof ContactType>;
+
+const ActorRoles = ['user', 'assistant'] as const;
```
wittjosiah, line 68: probably remove all these unused schemas

packages/plugins/experimental/plugin-script/src/templates/queue.ts
```diff
@@ -0,0 +1,149 @@
...
+});
+
+//
+// Types
+//
+
+const ContactType = S.Struct({
+  name: S.optional(S.String),
+  email: S.optional(S.String),
+}).pipe(EchoObject('example.org/TempContactType', '0.1.0'));
```
wittjosiah, line 65: assuming this is intentionally changed for testing purposes?

packages/plugins/experimental/plugin-script/src/templates/queue.ts
ZaymonFC: These are all required to type the incoming data as `MessageType`

packages/plugins/experimental/plugin-script/src/templates/queue.ts
ZaymonFC: We can't (yet) register and use system types!

packages/plugins/experimental/plugin-script/src/templates/queue.ts
wittjosiah: ah I missed that cast

packages/plugins/experimental/plugin-script/src/templates/queue.ts
wittjosiah: if you match the typename and schema though echo won't be able to tell the difference once it's replicated to composer?

ZaymonFC: creating a stored schema with the same typename and version as a static schema might have unintended consequences?

PR #8941, opened 2025-04-12
feat: Gmail sync defaults and test

richburdon:


packages/plugins/experimental/plugin-script/src/templates/gmail.ts
```diff
@@ -11,40 +11,42 @@ import {
...
+// @ts-ignore
 import { pipe, Chunk, Effect, Ref, Schedule, Stream } from 'https://esm.sh/effect@3.13.3';
 
-// TODO(ZaymonFC): Calculate this dynamically and expose a parameter.
-const DEFAULT_AFTER = '2025-01-01';
-
 export default defineFunction({
   inputSchema: S.Struct({
+    mailboxId: S.String,
+    userId: S.optional(S.String).pipe(S.withDecodingDefault(() => 'me')),
```
wittjosiah, line 21: do defaults like this work for echo objects too?

packages/plugins/experimental/plugin-script/src/templates/gmail.ts
```diff
@@ -11,40 +11,42 @@ import {
...
-    context: { space },
   }: any) =>
     Effect.gen(function* () {
       const { token } = yield* Effect.tryPromise({
         try: () => space.db.query(Filter.typename('dxos.org/type/AccessToken', { source: 'gmail.com' })).first(),
         catch: (e: any) => e,
       });
 
-      const makeRequest = (path: string) =>
+      // TODO(burdon): Use `googleapis`.
```
wittjosiah, line 46: Here’s a summary explaining why you’re not using the googleapis package:

⸻

The team encountered a deployment error (Uncaught TypeError: Cannot convert object to primitive value) when trying to deploy a script that included the googleapis and google_auth_library packages. The error occurred during the module evaluation phase of deployment, even though the function wasn’t being run yet. After investigating, it became clear that the issue only happened when either of those libraries were included in the bundle.

Additionally, the bundle size with googleapis was extremely large (~18MB), which isn’t ideal for edge environments. Based on these findings, the decision was made to avoid using the googleapis package and instead implement the needed functionality using fetch and custom helper functions to keep things lightweight and compatible with the deployment platform.

wittjosiah: (summary of my conversation w/ Dima about this)

packages/plugins/experimental/plugin-script/src/templates/gmail.ts
```diff
@@ -11,40 +11,42 @@ import {
...
 
   outputSchema: S.Struct({
     newMessages: S.Number,
   }),
 
   handler: ({
+    context: { space },
     event: {
-      data: { mailboxId, userId = 'me', after = DEFAULT_AFTER, pageSize = 100 },
+      data: { mailboxId, userId, after, pageSize = 100 },
```
wittjosiah, line 37: why moving some defaults into the schema but not all?

packages/plugins/experimental/plugin-script/src/templates/gmail.ts
richburdon: still not yet

PR #8946, opened 2025-04-14
feat(grid): Readonly cell text selection

ZaymonFC:


packages/ui/lit-grid/src/dx-grid.ts
```diff
@@ -1228,19 +1228,52 @@ export class DxGrid extends LitElement {
 
     const cellReadonly = this.cell(col, row, plane)?.readonly;
     if (cellReadonly !== undefined) {
-      return cellReadonly;
+      // Convert the possible string values to boolean
+      return cellReadonly === 'no-text-select' || cellReadonly === 'text-select' || cellReadonly === true;
```
thure, line 1232: Strings will be `true`, so maybe just `!!cellReadonly`?

packages/ui/lit-grid/src/dx-grid.ts
```diff
@@ -1228,19 +1228,52 @@ export class DxGrid extends LitElement {
...
     }
 
-    return (
-      (this.columns?.[colPlane]?.[col]?.readonly ?? this.columnDefault?.[colPlane]?.readonly) ||
-      (this.rows?.[rowPlane]?.[row]?.readonly ?? this.rowDefault?.[rowPlane]?.readonly)
-    );
+    const colDefault = this.columns?.[colPlane]?.[col]?.readonly ?? this.columnDefault?.[colPlane]?.readonly;
+    const rowDefault = this.rows?.[rowPlane]?.[row]?.readonly ?? this.rowDefault?.[rowPlane]?.readonly;
```
thure, line 1236: Names are a little confusing, maybe `colReadonly` / `rowReadonly`?

packages/ui/lit-grid/src/dx-grid.ts
```diff
@@ -1228,19 +1228,52 @@ export class DxGrid extends LitElement {
...
+
+    // Check cell-specific setting first.
+    const cellReadonly = this.cell(col, row, plane)?.readonly;
+    if (cellReadonly !== undefined) {
+      return cellReadonly === 'text-select' || cellReadonly === false;
+    }
+
+    // Check column/row defaults.
+    const colDefault = this.columns?.[colPlane]?.[col]?.readonly ?? this.columnDefault?.[colPlane]?.readonly;
+    const rowDefault = this.rows?.[rowPlane]?.[row]?.readonly ?? this.rowDefault?.[rowPlane]?.readonly;
```
thure, line 1258: Ditto on naming here

packages/ui/lit-grid/src/dx-grid.ts
```diff
@@ -1228,19 +1228,52 @@ export class DxGrid extends LitElement {
...
+      return true;
+    }
+    if (colDefault === true || colDefault === 'no-text-select') {
+      return false;
+    }
+    if (rowDefault === true || rowDefault === 'no-text-select') {
+      return false;
+    }
+
+    return !this.cellReadonly(col, row, plane);
```
thure, line 1269: text is only selectable if the `readonly` value is exactly `text-select`, if the cell is not readonly then it’s up to the consumer to handle displaying an editor since dx-grid doesn’t support `contenteditable` on its own.

packages/ui/lit-grid/src/dx-grid.ts
ZaymonFC: That's a little magical for me.

thure: Fair enough! This is still fine.

packages/ui/lit-grid/src/dx-grid.ts
ZaymonFC: 100%. Fixed this. Thanks!

thure (approved):
lgtm!

PR #8948, opened 2025-04-15
feat: Companion tabs

thure:
This PR:
- refactors companion planks to use graph for their ontology
- allows multiple companion nodes to render as tabs


ZaymonFC (approved):
LGTM ✨

PR #8966, opened 2025-04-16
fix(plugin-meetings): Fix clean-up on track closing

mykola-vrmchk:
- **Add create space method**
- **Define types**
- **Add testing panel**
- **Add template**
- **Misc**
- **Lock**
- **Extend abilities of TestingPanel**
- **Lock file**
- **Config**
- **Lock**
- **Fix script types**
- **fix: Fix tracks closing handling**


packages/plugins/plugin-meeting/src/peers.stories.tsx
```diff
@@ -149,6 +148,97 @@ export default meta;
...
 
 export const Default: Story = {
+  render: Render,
+  args: {
+    videoSrc: video,
+  },
+};
+
+export const RePull: Story = {
+  render: (args) => {
```
richburdon, line 158: better to pull this out into a local function i think; makes the meta more readable.

richburdon (approved):

PR #8984, opened 2025-04-21
fix(plugin-inbox): Consume css in composer bundle directly

ZaymonFC:


packages/apps/composer-app/src/plugins.tsx
```diff
@@ -19,6 +19,8 @@ import { FilesPlugin, FILES_PLUGIN } from '@dxos/plugin-files';
 import { GraphPlugin, GRAPH_PLUGIN } from '@dxos/plugin-graph';
 import { HelpPlugin, HELP_PLUGIN } from '@dxos/plugin-help';
 import { InboxPlugin } from '@dxos/plugin-inbox';
+// NOTE(ZaymonFC): This is a workaround. See: https://discord.com/channels/837138313172353095/1363955461350621235
+import '@dxos/plugin-inbox/css';
```
wittjosiah, line 23: could you put the naked imports at the top of the file?

wittjosiah (approved):

packages/ui/react-ui-theme/src/plugins/plugin.ts
```diff
@@ -72,7 +72,7 @@ export const ThemePlugin = (options: ThemePluginOptions): Plugin => {
     name: 'vite-plugin-dxos-ui-theme',
     config: async ({ root }, env): Promise<UserConfig> => {
       environment = env.mode;
-      const content = root ? await resolveKnownPeers(config.content ?? [], root) : config.content;
+      const content = root ? await resolveKnownPeers([...(config.content ?? []), '**/*.pcss'], root) : config.content;
```
thure, line 75: I think this should be reverted for now

thure (approved):
lgtm! should probably revert one thing.

PR #8985, opened 2025-04-22
feat: Setting to enable/disable deck.

richburdon:


packages/plugins/plugin-deck/src/translations.ts
```diff
@@ -8,6 +8,7 @@ export default [
   {
     'en-US': {
       [DECK_PLUGIN]: {
+        'app name': 'Composer',
```
wittjosiah, line 11: app name is defined by the entrypoint https://github.com/dxos/dxos/blob/main/packages/apps/composer-app/src/plugins.tsx#L188 and is accessible via the theme plugin's translations https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-theme/src/translations/en-US/appkit.ts#L8

packages/plugins/plugin-deck/src/components/DeckLayout/Banner.tsx
```diff
@@ -21,7 +23,7 @@ export const Banner = ({ variant, classNames }: ThemedClassName<{ variant?: 'top
       )}
     >
       {variant === 'sidebar' ? <CloseSidebarButton /> : <ToggleSidebarButton />}
-      <span className='self-center grow mis-1'>Composer</span>
+      <span className='self-center grow mis-1'>{t('app name')}</span>
```
wittjosiah, line 26: 👍 see other comment for where this translations already exists

packages/plugins/plugin-deck/src/components/Plank/PlankControls.tsx
```diff
@@ -24,6 +24,8 @@ export type PlankControlHandler = (event: DeckAction.PartAdjustment) => void;
 export type PlankCapabilities = {
   incrementStart?: boolean;
   incrementEnd?: boolean;
+  // TODO(burdon): Variants.
+  deck?: boolean;
```
wittjosiah, line 28: I don't understand this, what is the "deck" capability/action?

richburdon: if not enabled, then it would lock solo mode.

PR #8990, opened 2025-04-24
feat: react-ui-list accordion first iteration

richburdon:
Quick first cut of a headless scrolling accordion based loosely on the List component.

https://github.com/user-attachments/assets/47b1bee0-bd9f-4ded-997b-8d58c1480eb3

### DIFF
--- a/packages/ui/react-ui-list/src/components/Accordion/AccordionRoot.tsx
+++ b/packages/ui/react-ui-list/src/components/Accordion/AccordionRoot.tsx
@@ -3,7 +3,7 @@
 //
 
 import { createContext } from '@radix-ui/react-context';
-import React, { useCallback, useMemo, type ReactNode } from 'react';
+import React, { useCallback, useState, type ReactNode } from 'react';
 
 import { type ThemedClassName } from '@dxos/react-ui';
 import { mx } from '@dxos/react-ui-theme';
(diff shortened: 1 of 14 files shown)


packages/ui/react-ui-list/src/components/Accordion/AccordionRoot.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: AccordionRootProps<T>) => {
+  const [openItems, setOpenItems] = useState<Record<string, boolean>>({});
+
+  const handleSetItemOpen = useCallback((id: string, open: boolean) => {
+    setOpenItems((prev) => ({ ...prev, [id]: open }));
+  }, []);
+
+  return (
+    <AccordionProvider {...{ openItems, setItemOpen: handleSetItemOpen, getId }}>
+      <div className={mx('flex flex-col w-full overflow-x-hidden overflow-y-auto scrollbar-thin', classNames)}>
```
thure, line 50: Please use Radix’s `Accordion` primitives, unless there is a compelling reason to make these from scratch.

thure: The `flex flex-col w-full` incantation is almost certain to cause issues — I still recommend using only exactly the interventions necessary to achieve what you need here. Default block layout already achieves these 3 things unless you’re actually using shrink or grow in the flex items.

thure: `overflow-x-hidden` is implied when `overflow-y` is anything other than `visible`.

thure (changes requested):
Let me know if Radix’s `Accordion` primitive is not appropriate for this component. I’m happy to help get this across the line if needed.

packages/ui/react-ui-list/src/components/Accordion/AccordionRoot.tsx
richburdon: agreed. this was a quick experiment. i think if we decide we want the accordion pattern then we'll book some time to redo this using Radix.

PR #8995, opened 2025-04-24
refactor: move space settings dialog into planks

wittjosiah:
Notes:
- adjusts how variants are handled by the deck, in particular they are not a companion-specific thing but any node can be specified as a variant
- surfaces which render variants should be position: 'hoist' to be prioritized above the default renderer, this allows the default render to not have to be concerned with variants
- removed the menu indicator from the sigil when it has no actions
- don't automatically hide companion nodes, hide based on disposition of 'hidden'
- copied a bunch of components between client and space plugin, these should be reconciled sometime soon

TODO:
- [x] form cleanup
- [x] update e2e tests
- [x] scroll to plank not working for variants

Screenshot 2025-04-24 at 13 51 06
Screenshot 2025-04-24 at 15 06 08


packages/plugins/plugin-space/src/components/ControlSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const ControlFrame = ({ children }: ControlGroupProps) => (
+  <div
+    role='none'
+    className='p-4 border border-separator rounded-lg container-max-width grid grid-cols-1 md:grid-cols-[1fr_1fr] gap-2 md:gap-8'
+  >
+    {children}
+  </div>
+);
+
+export type ControlItemProps = PropsWithChildren<{
```
richburdon, line 48: pls put the Props directly before the class when that's possible.

packages/plugins/plugin-space/src/components/SpaceSettings/SpacePropertiesForm.tsx
```diff
@@ -0,0 +1,139 @@
...
+              classNames='min-is-64'
+            />
+          </ControlItemInput>
+        );
+      },
+      icon: ({ type, label, getValue, onValueChange }) => {
+        const handleChange = useCallback((nextEmoji: string) => onValueChange(type, nextEmoji), [onValueChange, type]);
+        const handleEmojiReset = useCallback(() => onValueChange(type, undefined), [onValueChange, type]);
+        return (
+          <ControlItem title={label} description={t('icon description')}>
```
richburdon, line 90: can you make the icon and color picker consistent (not full width and right aligned).

richburdon (approved):

packages/plugins/plugin-space/src/components/SpaceSettings/SpacePropertiesForm.tsx
wittjosiah: yes, that's why this is a draft, I need to get Will's help to cleanup this form

PR #9023, opened 2025-04-29
feat: Codemirror transcript scroller

richburdon:
- [x] Reimplement transcription component using Codemirror to support previews.
- [x] Should diff TranscriptBlocks if changed.

## Demo
<br>

https://github.com/user-attachments/assets/7a43ce01-4399-4ebb-acc6-708f87ca3fb1


thure (approved):
Clever to use gutter for the timestamps! lgtm ✨

PR #9024, opened 2025-04-29
feat(functions): Use function input schema in trigger editor

ZaymonFC:
https://github.com/user-attachments/assets/26d33abe-5990-40ec-a55e-ed2aa176c074

NOTE(Zaymon): We cherry picked the form ref object lookup code from this PR in #9145. So we'll need to drop the changes here before merge.


ZaymonFC:
Moved to #9155

PR #9032, opened 2025-04-29
feat: DXN link tags and target lookup components

thure:
This PR:
- begins a resolution of #9006
- implements `dx-ref-tag`
- uses `dx-ref-tag` in react-ui-editor preview stories
- refactors react-ui-editor to have smaller story files

https://github.com/user-attachments/assets/ac7ab21b-2cb7-4b9f-a717-50c97a10667e

https://github.com/user-attachments/assets/b6baf195-19c3-43d5-89c6-fbebfe77a0aa


richburdon:
Popover stays on top inside CM client.

https://github.com/user-attachments/assets/9b4ddc24-2891-4977-b600-8dca149c0753

packages/plugins/plugin-markdown/src/capabilities/intent-resolver.ts
```diff
@@ -48,44 +47,23 @@ export default (context: PluginsContext) =>
...
+        const target = await resolveRef(client, targetRef.dxn, space);
+        if (target && isInstanceOf(DocumentType, target)) {
+          const accessor = createDocAccessor(target, ['content']);
+          // TODO(burdon): Should be a cursor that references a selected position.
+          const index = 0;
+          accessor.handle.change((doc) => {
+            // TODO(burdon): Throws error:
+            // intent-dispatcher.ts:270 Cannot read properties of undefined (reading 'annotations') (FiberFailure) TypeError: Cannot read properties of undefined (reading 'annotations')
+            const ref = `[${label ?? 'Generated content'}]](${objectRef.dxn.toString()})\n`;
+            A.splice(doc, accessor.path.slice(), index, 0, ref);
```
richburdon, line 62: @dmaretskyi this is currently throwing

packages/ui/react-ui-editor/src/components/RefPopover.tsx
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  useEffect(() => {
+    return rootRef
+      ? addEventListener(rootRef, 'dx-ref-tag-activate', handleDxRefTagActivate, customEventOptions)
+      : undefined;
+  }, [rootRef]);
+
+  return (
+    <RefPopoverProvider pending={value.pending} link={value.link} target={value.target}>
```
richburdon, line 58: @thure can you constrain the popover so that it has a max size if there are not extrinsic constraints (e.g., the current default is full width -- and I think a better default would be say 20rem width (phone width).

richburdon:
- scrolling bug [@thure]
- startup storybook bug [@wittjosiah]
- insert into document bug [@dmaretskyi]

https://github.com/user-attachments/assets/989d4ba8-11fc-4f43-a8f2-80b3fd79e5fc

packages/plugins/plugin-markdown/src/capabilities/intent-resolver.ts
wittjosiah: mutating the am doc needs to be within a `change` doesn't it?

PR #9047, opened 2025-04-30
feat: Triggers schema proposal

dmaretskyi:


packages/core/functions/src/types/types.ts
```diff
@@ -108,21 +109,34 @@ export type TriggerType = S.Schema.Type<typeof TriggerSchema>;
...
   // TODO(burdon): Flatten entire schema.
   spec: S.optional(TriggerSchema),
 
-  // TODO(burdon): Get schema as partial from function.
-  // TODO(wittjosiah): Rename to payload.
-  // The `meta` property is merged into the event data passed to the function.
-  meta: S.optional(S.mutable(S.Record({ key: S.String, value: S.Any }))),
+  /**
+   * Passed as the input data to the function.
+   * Must match the function's input schema.
```
wittjosiah, line 136: when in conductor this isn't necessarily true, because other data could be bound to the inputs that doesn't come from the trigger, correct? but this is true if they are bound directly together

packages/core/functions/src/handler.ts
```diff
@@ -22,13 +23,28 @@ import { isNonNullable } from '@dxos/util';
...
+  /**
+   * Data passed as the input to the function.
+   * Must match the function's input schema.
+   */
+  data: TData;
+
+  /**
+   * Event data from the trigger.
+   */
+  event: EventType;
```
wittjosiah, line 47: should this be a part of the generic? that way you can provide strong types on it later

wittjosiah, line 47: in a workflow does this also come from the trigger for the overall workflow?

packages/core/functions/src/types/types.ts
dmaretskyi: good catch. I don't love how those are done separatelly for conductor and for functions

packages/core/functions/src/handler.ts
yarolegovich: not sure I understand why we need both event and trigger

wittjosiah: I think you're right that likely you only need the event and not both

packages/core/functions/src/types/types.ts
dmaretskyi: @yarolegovich @richburdon this seems like a major point to discuss

packages/core/functions/src/handler.ts
dmaretskyi: for example for queue triggers you'd get the queueId from `trigger.spec.queueId`. or would you do it differently?

dmaretskyi: yes

wittjosiah: I think you could include that sort of thing in the event to avoid having two places the trigger information could be

packages/core/functions/src/types/types.ts
dmaretskyi: Here's my proposal -- the binding operates in 2 modes:

## Mode 1 `trigger.payload === undefined`:

the input to the function is the TriggerOutput type, i.e. the event. The function must declare its input schema to match the event type.
This is how wokflows will operate, for workflows you can read everything above with function meaning workflow.

## Mode 2 `trigger.payload !== undefined`:

In this case `payload` is passed to the function input, but the even is still acceesible from the event field.
This is the mode we discussed recently. 
This will also be used for direct function to trigger bindings.

dmaretskyi: ## Webhooktriggers

For webhook triggers I propose that the runtime parses the body and matches it to the input schema.
A function that has input of `S.Struct({ message: S.String })` you would call with JSON body of `{ "message": "hello" }`
If the body doesn't match the schema runtime should error.

You can also set the input schema to be S.Void, and then access the unparsed body text from the `event.bodyText`

yarolegovich: > Webhooktriggers

Sounds useful and makes sense.

What about queue triggers? Should we map inserted items to match input schema? Can also potentially be useful. I can invoke a function synchronously if I get a result, or I can just publish a message or a batch to a queue and let the function process items in background. 

Subscription triggers? Maybe getting the changed object snapshot as payload will make it easier to write useful functions. 

My point is, given `event.trigger` will contain everything you might want to know about the invocation reason (queueId when applicable, trigger type etc), I like the idea of keeping `data` of function input trigger-agnostic. 
So that I can write a function that operate on `Foo` and make it run when:
1. A `Foo` changes (subscription).
2. A `Foo` item is appended to a queue.
3. A sync http call is made with `Foo` in body.
4. Periodically for some `Foo`. Sounds weird, but if a function sends discord messages I might want to be able to invoke it using webhook and also periodically for some channel (timer trigger with channel data in `payload`). 

Email is kind of a special case.

Also `payload` seems very timer-specific, and the question is should we maybe just move it timer spec?

dmaretskyi: To summarize, instead of two modes, we're going to put the payload only on the cron trigger, and for all of the triggers, the data is the event for cron trigger data is the payload, and then we don't need a separate event channel for the event. The function signature is just data, trigger, and context. 

We also need `trigger.meta` thats a place to store configuration that's only accessible in functions but not workflows

dmaretskyi:
### Alternative proposal to unifying configuration and event input

Add `trigger.input` field that defines the input data for the function invocation. 
It's a record of properties where the values are either constants (`{ const: 'XXXX' }`) or expresions (`{ get: '$.json.path' }`)

Example:

```ts
const trigger = {
   function: { '/': 'dxn:echo:AAA:XXX' }
   spec: {
      kind: 'queue',
      queue: { '/': 'dxn:queue:AAA:YYY' }
   },
   input: {
      item: {
         get: '$.trigger.event',
      },
      instructions: {
         const: 'Summarize and perform entity-extraction'
      }, 
      mailbox: {
        const: { '/': 'dxn:echo:AAA:ZZZ' }
      } 
   }
}
```

PR #9062, opened 2025-05-01
fix: Fix race condition in transcription state updates

mykola-vrmchk:
- **Fix race condition**
- **Add context**
- **Add activities**
- **Change calls payload structure**

### DIFF
--- a/packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
+++ b/packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
@@ -117,7 +117,7 @@ export default (context: PluginsContext) =>
                 state.transcriptionManager = undefined;
                 state.enabled = false;
               },
-              onCallStateUpdated: async (callState: CallState) => {
+              onCallStateUpdated: (callState: CallState) => {
                 const typename = getSchemaTypename(TranscriptType);
                 const transcription = typename ? callState.activities?.[typename] : undefined;
                 if (!transcription?.payload) {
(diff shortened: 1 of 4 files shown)


packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
```diff
@@ -128,15 +117,15 @@ export default (context: PluginsContext) =>
                 state.transcriptionManager = undefined;
                 state.enabled = false;
               },
-              onCallStateUpdated: (callState: CallState) => {
+              onCallStateUpdated: async (callState: CallState) => {
```
wittjosiah, line 120: doesn't look like you ended up needing this to be async?

packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
```diff
@@ -69,17 +58,17 @@ export default (context: PluginsContext) =>
           {
             id: `${fullyQualifiedId(meeting)}/action/start-stop-transcription`,
             data: async () => {
-              const call = context.requestCapability(MeetingCapabilities.CallManager);
-              invariant(state.transcriptionManager);
-              state.enabled = !state.enabled;
-              void state.transcriptionManager.setEnabled(state.enabled);
+              // NOTE: We are not saving the state of the transcription manager here.
```
wittjosiah, line 61: so this is waiting to update the state until you get the call state update from the swarm because otherwise there's a race between you local interaction and the swarm state?

wittjosiah (approved):

packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
mykola-vrmchk: Yes

PR #9066, opened 2025-05-01
fix(tables): References in relational table storybook

ZaymonFC:


packages/sdk/schema/src/testing/generator.ts
```diff
@@ -149,27 +147,27 @@ export const createObjectPipeline = <T extends BaseObject>(
...
-      );
+      const pipeline: Effect.Effect<ReactiveEchoObject<any>, never, never> = Effect.gen(function* (_) {
+        // logObject('before')(obj);
+        const withProps = createProps(generator, type, optional)(obj);
+        const liveObj = createReactiveObject(type)(withProps);
+        const withRefs = yield* Effect.promise(() => createReferences(type, db)(liveObj));
+        const dbObj = addToDatabase(db)(withRefs);
+        // logObject('after')(dbObj);
+        return dbObj;
+      });
```
wittjosiah, line 170: what's the change here? did this just change so you could console log and could be reverted? I suspect they must have a logger that could be used in a pipeline

wittjosiah (approved):

packages/sdk/schema/src/testing/generator.ts
ZaymonFC: I switched to the generator syntax, because the `Effect.` `map` / `flatMap` aspect is just noise.

wittjosiah: ah I see now, github's diff highlighting is brutal here

ZaymonFC: Yeah it looks way more gnarly than it is.

PR #9068, opened 2025-05-01
refactor: factor out preview from story to plugins

wittjosiah:
todo: integrate w/ deck


wittjosiah:
was merged into #9070

PR #9069, opened 2025-05-02
feat: render enriched markdown in mailbox messages

ZaymonFC:
https://github.com/user-attachments/assets/278cef52-060b-48d0-9f12-6f2b6c6ff86f


packages/plugins/plugin-inbox/src/components/Message/MessageContainer.tsx
```diff
@@ -2,23 +2,78 @@
...
+        preview({
+          onLookup: async ({ label, ref }) => {
+            const dxn = DXN.parse(ref);
+            if (!dxn) {
+              return null;
+            }
+
+            const object = await resolveRef(client, dxn, space);
+            return { label, object };
+          },
```
wittjosiah: from what I could tell this is no longer needed, I think this was an old way this extension worked, you should be able to remove this

wittjosiah (approved):

packages/plugins/plugin-inbox/package.json
```diff
@@ -54,9 +54,11 @@
     "@dxos/plugin-deck": "workspace:*",
     "@dxos/plugin-graph": "workspace:*",
     "@dxos/plugin-markdown": "workspace:*",
+    "@dxos/plugin-preview": "workspace:*",
```
wittjosiah, line 57: does this need to be a full dep?

packages/plugins/plugin-inbox/src/components/Message/MessageContainer.tsx
ZaymonFC: I was copying `Transcript`, how does the extension get the client and space otherwise?

wittjosiah: the preview plugin binds that, that callback is never actually called anymore as far as I could tell

wittjosiah: check with @thure / @richburdon, I wasn't sure if it was new or old code so I left it

ZaymonFC: I'll leave it until we overhaul `TextEditorPreview.stories.tsx`.

PR #9075, opened 2025-05-02
feat: add common types

wittjosiah:


packages/plugins/plugin-inbox/src/components/index.ts
```diff
@@ -5,7 +5,6 @@
 import { lazy } from 'react';
 
 export const EventsContainer = lazy(() => import('./Calendar'));
```
richburdon, line 7: what do you think about putting these into a namespace? E.g., `Types`.

richburdon (approved):

packages/plugins/plugin-inbox/src/components/index.ts
wittjosiah: hmmm, these are components?

PR #9085, opened 2025-05-02
refactor: move transcript to use messages

wittjosiah:


packages/plugins/plugin-transcription/src/components/Transcript/Transcript.tsx
```diff
@@ -20,26 +21,37 @@ import {
...
-  // TODO(burdon): Use link/reference markup for users (with popover).
-  // TODO(burdon): Color and avatar.
-  return [
-    `###### ${block.authorName}` + (debug ? ` (${block.id})` : ''),
-    block.segments.map((segment) => segment.text.trim()).join(' '),
-    '',
-  ];
-};
+export const renderMarkdown =
+  (identities: Identity[]) =>
```
richburdon, line 32: i'm adding Start/Stop messages, which will be anonymous. so this might need special handling.

packages/plugins/plugin-transcription/src/components/Transcript/testings.ts
```diff
@@ -9,16 +9,14 @@ import { AI_SERVICE_ENDPOINT } from '@dxos/assistant/testing';
...
-import { hues } from '@dxos/react-ui-theme';
-import { ContactType } from '@dxos/schema';
+import { ContactType, MessageType, type TranscriptionContentBlock } from '@dxos/schema';
 
-import { processTranscriptBlock } from '../../entity-extraction';
+import { processTranscriptMessage } from '../../entity-extraction';
 import * as TestData from '../../testing/test-data';
-import { TranscriptBlock } from '../../types';
 
 // TODO(burdon): Reconcile with plugin-markdown. Move to schema-testing.
```
richburdon, line 21: this file was renamed and moved to `../../testing/testing.ts`

packages/plugins/plugin-transcription/src/components/transcription.stories.tsx
```diff
@@ -4,184 +4,163 @@
...
-  };
-  useEffect(() => {
-    manageChunkRecording();
-  }, [transcriber, track?.readyState, transcriber?.isOpen]);
-
-  return <TranscriptionStory model={model} running={running} onRunningChange={setRunning} />;
-};
-
-const meta: Meta<typeof AudioFile> = {
+// import { TranscriptBlock } from '../types';
```
richburdon: TODO for mykola?

PR #9097, opened 2025-05-03
feat: Entity extraction in transcript

dmaretskyi:


packages/core/echo/echo-pipeline/src/db-host/query-state.ts
```diff
@@ -181,21 +181,21 @@ export class QueryState extends Resource {
 
 // TODO(burdon): Process Filter DSL.
 const filterToIndexQuery = (filter: FilterProto): IndexQuery => {
-  invariant(!(filter.type && (filter.or ?? []).length > 0), 'Cannot mix type and or filters.');
+  invariant(!((filter.type ?? []).length > 0 && (filter.or ?? []).length > 0), 'Cannot mix type and or filters.');
```
wittjosiah, line 184:
```suggestion
  const { type = [], or = [] } = filter;
  invariant(!(type.length > 0 && or.length > 0), 'Cannot mix type and or filters.');
```
fallback once rather than on every access

packages/plugins/plugin-transcription/src/capabilities/app-graph-builder.ts
```diff
@@ -98,11 +110,33 @@ export default (context: PluginsContext) =>
               schema: TranscriptType,
               getIntent: ({ space }: { space: Space }) =>
                 createIntent(TranscriptionAction.Create, { spaceId: space.id }),
-              onJoin: () => {
-                const transcriptionManager = new TranscriptionManager(client.edge);
+
+              // TODO(burdon): Is this the right place to inisitalize the state? Change to a capability?
```
wittjosiah, line 114: this could (maybe should) be an intent - there's already a capability which is storing the state, but as the logic gets more complex it would be better to have it factored out of the graph builder and into the intent resolver

PR #9099, opened 2025-05-03
chore: vscode settings (intellisense loading...)

richburdon:
Make VSCode use the same TS version as the workspace.
So far this has fixed the constant intellisense error that has made the IDE slow.
- https://www.codejourney.net/how-to-fix-visual-studio-code-intellisense-loading-infinitely/
- https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-newer-typescript-versions


dmaretskyi:
There are no changes to vscode configs in this PR

PR #9105, opened 2025-05-05
fix: Fix max video size

richburdon:


packages/plugins/plugin-meeting/src/components/Call/Call.tsx
```diff
@@ -45,14 +45,14 @@ CallAudio.displayName = 'CallAudio';
 
 type CallRoomProps = {};
 
-const CallRoom: FC<CallRoomProps> = () => {
+const CallMain: FC<CallRoomProps> = () => {
```
wittjosiah, line 48: why `Main`? is this not also used in the sidebar?

packages/plugins/plugin-meeting/src/components/Call/Call.tsx
```diff
@@ -45,14 +45,14 @@ CallAudio.displayName = 'CallAudio';
...
   const debug = useDebugMode();
   const call = useCapability(MeetingCapabilities.CallManager);
 
   return <ParticipantGrid self={call.self} users={call.users} debug={debug} />;
 };
 
-CallRoom.displayName = 'CallRoom';
+CallMain.displayName = 'CallRoom';
```
wittjosiah, line 55: if changing component name change the display name too

packages/plugins/plugin-meeting/src/components/Call/Call.tsx
richburdon: Changed to Call.Grid

PR #9132, opened 2025-05-07
feat: Schema serialization & API improvements

dmaretskyi:
- [x] Move getMeta, getTypename, isDeleted out of live-object and into echo-schema
- [x] consider splitting out API changes into a separate PR
- [x] Resolve circular deps in echo-schema
- [x] Relations
- [x] Schema explorer

image


packages/core/echo/echo-schema/src/json/annotations.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Annotations that go into ECHO namespace in json-schema.
+ */
+// TODO(dmaretskyi): Consider removing ECHO namespace and putting them at the top level.
+// TODO(dmaretskyi): Move to format.ts when circular imports are solved
+export const EchoAnnotations: Partial<Record<NamespacedJsonSchemaProperty, symbol>> = {
+  meta: PropertyMetaAnnotationId,
+  generator: GeneratorAnnotationId,
+  labelProp: LabelAnnotationId,
+
+  // TODO(dmaretskyi): `FieldLookupAnnotationId` might go here, but lets remove it entirely and use LabelAnnotation instead.
```
ZaymonFC, line 55: If `LabelAnnotation` does the same thing that's fine with my, I can update consumers.

packages/devtools/devtools/src/panels/echo/SchemaPanel/SchemaPanel.tsx
```diff
@@ -0,0 +1,179 @@
...
+      { name: 'typename', format: FormatEnum.String },
+      { name: 'version', format: FormatEnum.String, size: 100 },
+      {
+        name: 'kind',
+        format: FormatEnum.SingleSelect,
+        size: 100,
+        config: {
+          options: [
+            { id: 'object', title: 'OBJECT', color: 'red' },
+            { id: 'relation', title: 'RELATION', color: 'green' },
```
ZaymonFC, line 102: `red` should generally be reserved for errors and failures. Maybe `indigo`?

PR #9134, opened 2025-05-07
fix: check for missing artifact

wittjosiah:


mykola-vrmchk (approved):

packages/plugins/plugin-meeting/src/components/MissingArtifact.tsx
```diff
@@ -24,10 +24,13 @@ export const MissingArtifact = ({ meeting, typename }: MissingArtifactProps) =>
...
-      invariant(space);
-      const { data } = await dispatch(getIntent({ space, meeting }));
-      meeting.artifacts[typename] = makeRef(data!.object);
+      // TODO(wittjosiah): This check shouldn't be necessary, this component should only render if it's missing.
+      if (meeting.artifacts[typename] == null) {
+        const space = getSpace(meeting);
+        invariant(space);
+        const { data } = await dispatch(getIntent({ space, meeting }));
+        meeting.artifacts[typename] = makeRef(data!.object);
+      }
```
dmaretskyi, line 33: if (meeting.artifacts[typename] != null) {
 return;
}
        const space = getSpace(meeting);
        invariant(space);
        const { data } = await dispatch(getIntent({ space, meeting }));
         if (meeting.artifacts[typename] != null) {
 return;
}
        meeting.artifacts[typename] = makeRef(data!.object);

PR #9136, opened 2025-05-07
fix: support kanban of static schema

wittjosiah:


packages/ui/react-ui-kanban/src/testing/initialize-kanban.ts
```diff
@@ -42,29 +55,35 @@ const createDefaultTaskSchema = () => {
...
 export const initializeKanban = async ({
+  client,
   space,
   name,
   typename,
   initialPivotColumn,
-}: InitializeKanbanProps): Promise<{ kanban: KanbanType; schema: EchoSchema }> => {
+}: InitializeKanbanProps): Promise<{ kanban: KanbanType; schema?: EchoSchema }> => {
   if (typename) {
+    const staticSchema = client.graph.schemaRegistry.schemas.find((schema) => getTypenameOrThrow(schema) === typename);
```
ZaymonFC, line 65: Doesn't this break using stored schema since it will throw?

ZaymonFC (approved):
Looks good to me. Hopefully we can unify static and stored schema querying soon.

PR #9137, opened 2025-05-07
feat: Add connection state per peer to EDGE swarm protocol

mykola-vrmchk:
- **Add connection state in Messenger proto**
- **Change protocol**


packages/core/protocols/src/proto/dxos/edge/messenger.proto
```diff
@@ -19,16 +19,42 @@ package dxos.edge.messenger;
...
+
+  /**
+   * State of the connection to the swarm.
+   */
+  optional ConnectionState connection_state = 5;
+  /**
+   * UNIX timestamp of the connection.
+   * @example 1715142000000
+   */
+  // Note: it is not `int64` because `int64` is compiled as a `string` in 'buf' and as a `number` in 'proto-js'.
```
dmaretskyi, line 50: need to enable bigint support somwhere

dmaretskyi (approved):

PR #9161, opened 2025-05-10
Sunggles & Struggles 

dmaretskyi:


packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const Org = Schema.Struct({
+  id: Type.ObjectId,
+  name: Schema.String,
+}).pipe(Type.def({ typename: 'dxos.org/type/Org', version: '0.1.0' }));
+
+const WorksFor = Schema.Struct({
+  id: Type.ObjectId,
+  since: Schema.String,
+  ...Relation.connects(Contact, Org),
```
dmaretskyi, line 31: @richburdon I can work on making echo combinators passthrough if you're good with this style of an API + making ids explicit

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
+import { Schema } from 'effect';
+/// ISSUE(dm): Is it okay to have one entry point for all of the types like object id and type definitions and will it prohibit us from building isolated storybooks that don't pull in echo? DX vs tree shaking.
+// So that type and relation API should be separate from echo database API.
+// Potentially, some part of the database API could be included in there. Things like query builders. So only the declarative API, not the actual execution layer.
```
richburdon, line 4: - we should keep separate aspects in different packages
- @dxos/echo is just the main API below @dxos/client; it mainly just aggregates other aspects under public namespaces

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+declare const Relation: any;
+declare const setupDb: any;
+declare const Filter: any;
+declare const Query: any;
+
+///
+
+const Contact = Schema.Struct({
+  // ISSUE(dm): Forcing explicit object IDs for every schema?
+  id: Type.ObjectId,
```
richburdon, line 18: it's ok to add sugar here (as we do for Relation)

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  name: Schema.String,
+  // ISSUE(dm): Reconcile "type" and "typename" i.e. the dxn and the name of the type.
```
richburdon, line 20: i think `dxos.org/type/Contact` is the typename and `dxn:....:dxos.org/type/Contact` is the FQ typename. and Type is the object representation of the type.

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  id: Type.ObjectId,
+  name: Schema.String,
+}).pipe(Type.def({ typename: 'dxos.org/type/Org', version: '0.1.0' }));
+
+const WorksFor = Schema.Struct({
+  id: Type.ObjectId,
+  since: Schema.String,
+  ...Relation.connects(Contact, Org),
+  // ISSUE(dm): "dxos.org/type" vs "dxos.org/relation"
+}).pipe(Relation.def({ typename: 'dxos.org/type/WorksFor', version: '0.1.0' }));
```
richburdon, line 33: depends on add/get/query; if the API is the same, then i would make this `type` the same.

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      since: '2021-01-01',
+      ...Relation.make(contact, org),
+      ...Type.meta({
+        keys: [],
+      }),
+    }),
+  );
+
+  {
+    // ISSUE(dm): Don't like get API. This won't return an object if it's not loaded.
```
richburdon, line 61: let's evolve it. perhaps a wrapper. but suggest getting the rest working first. i.e., second phase.

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    const contact2 = db.get(contact.id);
+    const org2 = db.get(org.id);
+    const worksFor2 = db.get(worksFor.id);
+    // ISSUE(dm): `db.ref(dxn).target`
+  }
+
+  {
+    const contact2 = await db.query(Query.type(Contact, { id: contact.id }).related(WorksFor)).run();
```
richburdon, line 69: - how can we query for all relations given a particular source/target?
- how do we filter traversal (e.g., all WorksFor where the date range is now)

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  const org = db.add(
+    create(Org, {
+      name: 'Acme Inc.',
+    }),
+  );
+
+  const worksFor = db.add(
+    create(WorksFor, {
+      since: '2021-01-01',
+      ...Relation.make(contact, org),
```
richburdon, line 53: live/make/create

packages/core/echo/echo/src/struggle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/*
+- Open PR
+- ISSUE(wj): Conversion between `create` <-> `live`
+- ISSUE(dm): Readonly APIs
+- ISSUE(burdon): Make source and target immutable.
+- ISSUE(burdon): Add `created`, `updated` (timestamps), `author`, etc to meta
+- ISSUE(burdon): Support dates
+- ISSUE(wj): How to get a DXN of an object. `getDXN`
+- ISSUE(dm): For cross-schema references, e.g. relation schema referencing a type schema, do DXN specify the target schema version or not?
```
richburdon, line 124: Many Relations will have a from/to (date) pattern and provenance (incl. possibly probability). Do we add these via a `...Relation.dateRange` sugar or consider this to be part of all relations?

richburdon (approved):

dmaretskyi:
the scope of dxos/echo:
 - echo-schema
 - live-object - maybe? trying to keep it out
 - automerge / echo-db - no

PR #9185, opened 2025-05-14
feat: Filter API in queries

dmaretskyi:


packages/core/echo/echo/src/query/api.ts
```diff
@@ -74,17 +83,80 @@ export interface Query<T> {
 }
 
 interface QueryAPI {
```
richburdon, line 85: I think QueryBuilder is appropriate.

packages/core/echo/echo/src/query/api.ts
```diff
@@ -94,143 +166,283 @@ interface QueryAPI {
...
    * @param to - End of the range (exclusive).
    */
-  range<T>(from: T, to: T): Predicate<T>;
+  between<T>(from: T, to: T): Filter<T>;
 
-  // TODO(dmaretskyi): Add `Query.match` to support pattern matching on string props.
+  /**
+   * Combine filters with a logical AND.
+   */
+  and<FS extends Filter.Any[]>(...filters: FS): Filter<Filter.And<FS>>;
```
richburdon, line 222: add `not`?

packages/core/echo/echo/src/query/api.ts
```diff
@@ -94,143 +166,283 @@ interface QueryAPI {
...
+      type: 'object',
       typename: dxn.toString(),
-      predicates: predicates ? predicateSetToAst(predicates) : undefined,
+      props: props ? propsFilterToAst(props) : {},
+    });
+  }
+
+  /**
+   * Internal.
+   */
```
richburdon, line 270: `/** @internal */`
?

packages/core/echo/echo/src/query/query.test.ts
```diff
@@ -48,88 +50,97 @@ describe('query api', () => {
     const getAllPeople = Query.type(Person);
 
     log.info('query', { ast: getAllPeople.ast });
+    Schema.validateSync(QueryAST.Query)(getAllPeople.ast);
   });
 
   test('get all people named Fred', () => {
     // Query<Person>
-    const getAllPeopleNamedFred = Query.type(Person, { name: 'Fred' });
+    const getAllPeopleNamedFred = Query.select(Filter.type(Person, { name: 'Fred' }));
```
richburdon, line 58: just `peopleNamedFred` (drop getAll prefix)

richburdon (approved):
Nice

PR #9190, opened 2025-05-15
fix: CSS `@layer` order race condition

thure:
This PR:
- fixes a CSS layer order issues in bundled versions of Composer

Screenshot 2025-05-14 at 17 09 14

### DIFF
--- a/packages/apps/composer-app/index.html
+++ b/packages/apps/composer-app/index.html
@@ -33,9 +33,11 @@
     <link rel="apple-touch-icon" href="/apple-touch-icon-180x180.png">
     -->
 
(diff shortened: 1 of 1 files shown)


packages/apps/composer-app/index.html
```diff
@@ -34,7 +34,7 @@
     -->
 
     <!-- TODO(wittjosiah): Add hash/nonce and remove unsafe-inline from content security policy. -->
-
+    <style>@layer dx-tokens, user-tokens, tw-base, dx-base, tw-components, dx-components, utilities;</style>
```
wittjosiah, line 37: this makes it look like my todo is for this, I'd suggest putting above the todo

PR #9199, opened 2025-05-15
chore: Remove dependencies on ImmutableSchema type

richburdon:
## API Goals

- [ ] `plugin-xxx` should depend on new `@dxos/echo` types and methods
- [ ] `react-ui-xxx` should not depend on `@dxos/echo`, `@dxos/echo-db`, or `@dxos/client` -- instead `@dxos/schema` which should export `@dxos/echo-schema` defs.
  - [ ] How should `react-ui-table` define the Table type without requiring ECHO schema?
  - [ ] Should we selectively import from `@dxos/echo`, e.g., `@dxos/echo/Type`?

## Tasks

- [x] Remove dependencies on deprecated `ImmutableSchema`
- [ ] Replace `isMutable(schema)` hack.
- [ ] JSONSchema reactivity (see KanbanContainer.stories.tsx, KanbanViewEditor.tsx)
- [ ] Move Table ECHO schema definition upstream into plugin (i.e., leave low-level table def in react-ui, but pipe this into a named object at the plugin level).

## Next
- [ ] Update @dxos/echo/docs diagrams
- [ ] Remove `@dxos/echo-db` deps from react-ui-x and plugin-x.
- [ ] Remove `@dxos/echo-schema` deps from react-ui-x and plugin-x.


packages/plugins/plugin-kanban/src/components/KanbanContainer.stories.tsx
```diff
@@ -95,7 +101,9 @@ const StorybookKanban = () => {
   const handleTypenameChanged = useCallback(
     (typename: string) => {
       if (kanban?.cardView?.target) {
-        schema?.mutable.updateTypename(typename);
+        invariant(schema);
+        invariant(isMutable(schema));
+        (schema as EchoSchema).updateTypename(typename);
```
ZaymonFC, line 106: Not a fan of this. Consider instanceof?

packages/plugins/plugin-kanban/src/components/KanbanViewEditor.tsx
```diff
@@ -36,7 +37,9 @@ export const KanbanViewEditor = ({ kanban }: KanbanViewEditorProps) => {
         view.query.typename = newTypename;
       }
 
-      schema.mutable.updateTypename(newTypename);
+      invariant(schema);
+      invariant(isMutable(schema));
+      (schema as EchoSchema).updateTypename(newTypename);
```
ZaymonFC, line 42: > Not a fan of this. Consider instanceof?

Same here.

packages/plugins/plugin-table/src/components/TableViewEditor.tsx
```diff
@@ -34,7 +35,9 @@ const TableViewEditor = ({ table }: TableViewEditorProps) => {
         view.query.typename = newTypename;
       }
 
-      schema.mutable.updateTypename(newTypename);
+      invariant(schema);
+      invariant(isMutable(schema));
+      (schema as EchoSchema).updateTypename(newTypename);
```
ZaymonFC, line 40: > Not a fan of this. Consider instanceof?

Here too.

packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
```diff
@@ -23,9 +23,6 @@ const ObjectDetailsPanel = ({ objectId, view }: RowDetailsPanelProps) => {
   const space = getSpace(view);
   const schema = useSchema(client, space, view.query?.typename);
 
-  // TODO(burdon): Why is this needed?
-  const effectSchema = useMemo(() => schema?.snapshot, [JSON.stringify(schema?.jsonSchema)]);
-
```
ZaymonFC, line 28: Pretty sure this was here because schema?.jsonSchema changes weren't reactive before. We should check that.

packages/sdk/react-client/src/echo/useSchema.ts
```diff
@@ -15,7 +16,7 @@ export const useSchema = (
   client: Client,
   space: Space | undefined,
   typename: string | undefined,
-): BaseSchema | undefined => {
+): Schema.Schema.AnyNoContext | undefined => {
```
ZaymonFC, line 19: What's the reason to pick `AnyNoContext` vs `Any` or `All`?

packages/sdk/schema/src/projection.ts
```diff
@@ -44,6 +44,7 @@ export class ViewProjection {
   private _fieldProjections = computed(() => this._view.fields.map((field) => this.getFieldProjection(field.id)));
   private _hiddenProperties = computed(() => this._view.hiddenFields?.map((field) => field.path as string) ?? []);
 
+  // TOOD(burdon): Should this take an instane of S.S.AnyNoContext and derive the JsonSchemaType itselft (and watch for reactivity)?
```
ZaymonFC, line 47: Maybe, but we manipulate the underlying `JsonSchema` directly in a few cases here. If we derive `JsonSchema` and only take `EffectSchema` then we need to make sure the changes get pushed back into the underlying `JsonSchema`.

packages/ui/react-ui-table/src/components/Table/Table.stories.tsx
```diff
@@ -152,7 +152,8 @@ const StoryViewEditor = () => {
     (typename: string) => {
       if (table?.view?.target) {
         invariant(schema);
-        schema.mutable.updateTypename(typename);
+        invariant(isMutable(schema));
+        (schema as EchoSchema).updateTypename(typename);
```
ZaymonFC, line 156: > Not a fan of this. Consider instanceof?

Here too?

packages/ui/react-ui-table/src/components/Table/relations.stories.tsx
```diff
@@ -30,14 +31,14 @@ const generator: ValueGenerator = faker as any;
...
 // TODO(burdon): Reconcile schemas types and utils (see API PR).
 // TODO(burdon): Base type for T (with id); see ECHO API PR?
-const useTestModel = <T extends BaseObject & HasId>(schema: BaseSchema<T>, count: number) => {
+const useTestModel = <T extends BaseObject & HasId>(schema: Schema.Schema<T>, count: number) => {
   const { space } = useClientProvider();
 
   const table = useMemo(() => {
     // const { typename } = pipe(schema.ast, SchemaAST.getAnnotation<TypeAnnotation>(TypeAnnotationId), Option.getOrThrow);
-    const typename = schema.typename;
+    const typename = getTypename(schema)!;
```
ZaymonFC, line 39: Prefer not to use the non-null assertion operator.

packages/ui/react-ui-table/src/components/Table/relations.stories.tsx
```diff
@@ -47,7 +48,7 @@ const useTestModel = <T extends BaseObject & HasId>(schema: BaseSchema<T>, count
     }
 
     // TODO(burdon): Just pass in view? Reuse same jsonSchema instance? View determines if mutable, etc.
-    return new ViewProjection(schema.jsonSchema, table.view.target);
+    return new ViewProjection(toJsonSchema(schema), table.view.target);
   }, [schema, table]);
```
ZaymonFC, line 52: Since we're calling `toJsonSchema` in both places we should probably memoize that then add it as a dependency to these two memos.

packages/ui/react-ui-table/src/util/dynamic-table.ts
```diff
@@ -36,13 +39,13 @@ export const getBaseSchems = ({
...
   jsonSchema?: JsonSchemaType;
-  schema?: BaseSchema;
+  schema?: Schema.Schema.AnyNoContext;
 }): { typename: string; jsonSchema: JsonSchemaType } => {
   if (typename && properties) {
     const schema = getSchemaFromPropertyDefinitions(typename, properties);
     return { typename: schema.typename, jsonSchema: schema.jsonSchema };
   } else if (schema) {
-    return { typename: schema.typename, jsonSchema: schema.jsonSchema };
+    return { typename: getTypename(schema)!, jsonSchema: toJsonSchema(schema) };
```
ZaymonFC, line 48: Prefer not to use non-null assertion operator.

packages/core/echo/echo/src/type/Obj.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { type AnyLiveObject as AnyLiveObject$ } from '@dxos/echo-db';
+import { type BaseEchoObject, type BaseObject, getSchema as getSchema$ } from '@dxos/echo-schema';
+import { live as live$ } from '@dxos/live-object';
+
+// TODO(burdon): Remove from Type?
+
+export declare namespace Obj {
+  export type Any = BaseEchoObject;
+  export type Live<T extends BaseObject> = AnyLiveObject$<T>;
```
dmaretskyi, line 13: I'm hesitant to include Live here -- it deals with reactivity

packages/core/echo/echo/src/type/index.ts
```diff
@@ -2,5 +2,7 @@
 // Copyright 2025 DXOS.org
 //
 
-export { Type } from './Type';
+export * as Obj from './Obj';
```
dmaretskyi, line 5: Those seem like a nice set of top-level nouns

Is the intent that Obj is the common base for Type and Relation?

I still don't like Type in here as it's hard to call instances of those objects "types". How about renaming it to Entity?

packages/core/echo/echo/src/type/Obj.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const create = live$;
```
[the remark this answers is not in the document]
dmaretskyi: technically I could see a way to combine `create` (static) and `live` but that would collapse those packages into a single one (again). But lets discuss this \w @wittjosiah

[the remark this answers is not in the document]
wittjosiah: imo, we should re-export both of them together `Object.live` and `Object.create`

packages/plugins/plugin-kanban/src/components/KanbanContainer.stories.tsx
richburdon: this is temporary; we need an API (see TODO on isMutable);

packages/plugins/plugin-table/src/components/ObjectDetailsPanel.tsx
richburdon: @dmaretskyi how do we deal with reactivity?

packages/ui/react-ui-table/src/components/Table/relations.stories.tsx
richburdon: @dmaretskyi i think this should throw -- the new API does, but i didn't backward port that behavior.

packages/ui/react-ui-table/src/util/dynamic-table.ts
richburdon: see above

packages/core/echo/echo/src/type/Obj.ts
richburdon: let's discuss.

packages/core/echo/echo/src/type/index.ts
richburdon: I think we need to decide if we can live with `Obj.Obj.Any` which is an outcome of the effect pattern.

wittjosiah: I think the way to clean that up is not importing from the index. In cases when you're just working w/ types you can import from `@dxos/echo/Object`

PR #9203, opened 2025-05-15
test: Color panels

richburdon:
Simple test for surface tokens.

## Current Story

Screenshot 2025-05-15 at 3 48 23 PM

## Kanban

Screenshot 2025-05-15 at 4 18 27 PM


packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
...
+
+const container = 'flex flex-col grow gap-2 p-4 rounded-md';
+
+const TestPanel = () => {
+  // TODO(burdon): Where should attention be applied?
+  const attentionAttrs = useAttendableAttributes('test');
+
+  return (
+    <StackItem.Content toolbar classNames='w-full'>
+      {/* TODO(burdon): Should the separator be applied by StackItem.Content? */}
```
thure, line 54: Bear in mind `StackItem` is upstream of, i.e. used by:
- Planks
- Stack sections
- Kanban columns
- Kanban cards

StackItem can apply styles relevant only to some of these, but I would assert we should apply typical separation of concerns, that it’s reasonable for Kanban, StackPlugin, and Deck to give concern-specific styles to their `StackItem`s and `StackItem.Content`s.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
 
 const StoryState = defineCapability<{ tab: string }>('story-state');
 
+// TODO(burdon): How to adjust existing surfaces for this test?
+// TODO(burdon): How to toggle attention in this test?
```
thure, line 42: I’ve been struggling with this too, need to investigate or build scaffolding for attention in storybook.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
...
+// TODO(burdon): Fix outline (e.g., button in sidebar nav is clipped when focused).
```
thure, line 43: Sidebar nav and any of its parents even up to the sidebar itself should not need any overflow other than visible — this should only ever occur for elements that scroll.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
...
+// TODO(burdon): Consider similar containment of: Table, Sheet, Kanban Column, Form, etc.
+
+const container = 'flex flex-col grow gap-2 p-4 rounded-md';
```
thure, line 46: I would still assert `flex flex-col` is not a good default, it should only apply where it is needed.

I think it’s actually the Planks themselves, rather than their content, which might benefit most from the encapsulation here, similar to Fibery’s approach.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
...
+
+const TestPanel = () => {
+  // TODO(burdon): Where should attention be applied?
```
thure, line 49: Anything can have attendable attributes, which is different from having an attention background or applying other attention-related signifiers (e.g. toolbars becoming partially opaque).

The attributes are for annotating the object id of which the element is rendering a representation, they don’t apply any styles.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
```diff
@@ -36,13 +38,54 @@ faker.seed(1234);
...
+  const attentionAttrs = useAttendableAttributes('test');
+
+  return (
+    <StackItem.Content toolbar classNames='w-full'>
+      {/* TODO(burdon): Should the separator be applied by StackItem.Content? */}
+      {/* TODO(burdon): Toolbar is not the same height as the sidebar here. */}
```
thure, line 55: That is intentional — toolbars are narrower than rails.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
richburdon: i'm just pulling this in as a devDependency for this story.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
richburdon: see first image above -- blue outline is cut off.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
richburdon: again, please just update to make this story a canonical example.
noting that sheet, table, and kanban all do this inconsistently (hence the bug in surface attention we discussed), so again i want to make this story definitive.

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
richburdon: I mean the TestPanel Toolbar is shorter than the NavTree Toolbar.

Screenshot 2025-05-16 at 10 38 13 AM

richburdon:
The tooltip stays present when the menu is closed.

Screenshot 2025-05-16 at 10 41 14 AM

packages/plugins/plugin-navtree/src/components/NavTree/NavTree.stories.tsx
richburdon: @wittjosiah

wittjosiah: @thure there's an attention decorator for stories already, I can help you set that up if you're not sure where to get started

PR #9212, opened 2025-05-16
feat: Safe mode

richburdon:
- Add `?safe` to the composer URL and it will prevent ALL but the necessary core plugins from loading.
- Memoizes video components for meetings (suspected memory/perf overhead).


thure (approved):
lgtm!

packages/sdk/app-framework/src/App.tsx
```diff
@@ -49,6 +50,7 @@ export type CreateAppOptions = {
  * @param params.placeholder Placeholder component to render during startup.
  * @param params.fallback Fallback component to render if an error occurs during startup.
  * @param params.cacheEnabled Whether to cache enabled plugins in localStorage.
+ * @param params.safeMode Whether to enable safe mode.
```
wittjosiah, line 53: would be helpful to describe what safe more actually does (e.g. ignores enabled plugins)

packages/apps/composer-app/src/main.tsx
```diff
@@ -98,7 +98,8 @@ const main = async () => {
     plugins: plugins(conf),
     core: core(conf),
     defaults: defaults(conf),
-    cacheEnabled: true,
+    cacheEnabled: false,
```
wittjosiah, line 101: I don't think this is intended to be merged

PR #9219, opened 2025-05-20
chore: Tinker with vitest configs for workspace execution

dmaretskyi:


wittjosiah (commented):
@dmaretskyi is this still relevant?

PR #9222, opened 2025-05-20
refactor: Reconcile SpaceId schema

dmaretskyi:


richburdon (approved):

packages/plugins/plugin-thread/src/types.ts
```diff
@@ -4,9 +4,8 @@
 
 import { Schema } from 'effect';
 
-import { SpaceIdSchema } from '@dxos/echo-schema';
 import { ChannelType, ThreadType } from '@dxos/plugin-space/types';
-import { EchoObjectSchema } from '@dxos/react-client/echo';
+import { EchoObjectSchema, SpaceId } from '@dxos/react-client/echo';
```
richburdon, line 8: for plugins import Type from @dxos/echo

richburdon (approved):

PR #9224, opened 2025-05-20
feat: hide members gallery when in fullscreen mode

richburdon:
- [x] autohide members gallery in fullscreen mode
- [x] bias size of pinned window to gallary by 2:1
- [x] by default turn deck off (advanced feature)
- [x] debounce speaking indicator
- [x] remove readonly flag from kanban preview forms


thure (approved):
lgtm ✨

PR #9234, opened 2025-05-24
chore: Color system mods

richburdon:
Small simplifications.


thure (approved):
Tried to answer your questions here

PR #9241, opened 2025-05-27
fix: add react-ui content to tailwind 

richburdon:


packages/ui/react-ui-theme/src/config/tokens/lengths.ts
```diff
@@ -1,4 +1,5 @@
 //
+
```
thure, line 2: ?

thure (approved):
If this works and unblocks you, I’m fine with it, though it bears mentioning our original intention with `react-ui` was to decouple it from tailwind utility classnames, hence the `tx` function and the separate `react-ui-theme` package.

This looks more like it just changes which packages are included in the contents configuration, and maybe `react-ui` gets lumped in coincidentally, which is also fine, just might be cause to adjust the title of the PR to avoid confusion when we look back at the record.

richburdon:
> If this works and unblocks you, I’m fine with it, though it bears mentioning our original intention with `react-ui` was to decouple it from tailwind utility classnames, hence the `tx` function and the separate `react-ui-theme` package.
> 
> This looks more like it just changes which packages are included in the contents configuration, and maybe `react-ui` gets lumped in coincidentally, which is also fine, just might be cause to adjust the title of the PR to avoid confusion when we look back at the record.

i think that's long since not been true though. i think the vast majority of react-ui packages require TW classes; in fact the lit and react-ui (core) packages may be the only ones that do not.

perhaps there's a way to evolve the semantic tokens into being the way to control the theme?

thure:
>  i think the vast majority of react-ui packages require TW classes

Which ones in particular? After a cursory look, I can only see that `CopyButton` uses tailwind directly.

We had established this was a matter of tech debt, IIRC, that components promoted upstream to `react-ui` should eventually conform to its theming devolution.

wittjosiah:
@thure you're probably correct that the core react-ui components don't but most of the packages in `ui/*` are "intermediate" react-ui-* packages at this point

wittjosiah:
Yes, exactly

PR #9254, opened 2025-05-31
refactor: channels, calls & meetings

wittjosiah:
https://github.com/user-attachments/assets/262f5373-b1d1-47c6-bb46-33f83974f2b9


wittjosiah:
@thure just experimenting with stack here, not set on it. The goal here is to make this the meeting "home" and will likely include other stuff in the future. The details of this still need a lot of refinement, this is primarily intended to put the broad strokes in place.

packages/plugins/plugin-thread/src/translations.ts
```diff
@@ -41,6 +40,48 @@ export default [
...
+        'summarize label': 'Summarize',
+        'summarizing label': 'Creating summary...',
+        'create summary message': 'A summary doesn’t exist for this meeting yet. Create one now?',
+
+        'join call': 'Join',
+        'leave call': 'Leave',
+
+        'lobby participants_zero': 'No participants',
+        'lobby participants_one': '1 participant',
+        'lobby participants_other': '{{count}} participants',
```
thure, line 67: Nice!

PR #9272, opened 2025-06-04
Remove @dxos/echo-db dependency from @dxos/echo package

### LINKED ISSUE #9271: Remove @dxos/echo-db dependency from @dxos/echo package

dmaretskyi:
Removes the @dxos/echo-db dependency from the @dxos/echo package by replacing AnyLiveObject imports with local type definitions using components from @dxos/live-object and @dxos/echo-schema.

Fixes #9271

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/core/echo/echo/package.json
+++ b/packages/core/echo/echo/package.json
@@ -22,7 +22,6 @@
   ],
   "dependencies": {
     "@dxos/debug": "workspace:*",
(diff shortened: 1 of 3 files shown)


dmaretskyi:
@cluade also do pnpm install and make sure ci passes

PR #9285, opened 2025-06-06
feat(react-ui-form): Readonly design increment

thure:
This PR:
- improves the experience of readonly Forms
- reduces design debt in Forms

https://github.com/user-attachments/assets/20b7ee67-80c7-4921-9ac1-c34f12524e1e


thure:
> I think the double layers here are a bit strange; i.e., border around the pills

Out of scope for readonly, this is a design for the interactive picker. Logged as #9295.

> Please remove the vertical button; can we normalize all icon buttons so that they are square and the same size?

I think you mean it should not span the full height of the list item rather than the delete button should be removed. I’ve applied a style which centers it vertically in this case. If we attempt to do so systematically, you’d be surprised how many buttons are actually _not_ square or otherwise have spacing issues and do rely on `self-stretch` as is default for grid items, e.g.:

Screenshot 2025-06-10 at 00 59 16
Screenshot 2025-06-10 at 00 53 23

> I don't think the border around the read-only content is helpful.

I think you mean that the read-only content should not be encapsulated the way tags are encapsulated; I’ve reverted to plain text. Please see the updated screen capture in the description of this PR.

> Also for the readonly form, can we try making the captions smaller?

I don’t think readonly is good cause to create a variation in label size since it might mix with interactive inputs at some point. I do agree the contrast could be improved, so I adjusted labels throughout the system to apply a smaller type size `xs` to increase contrast between them and their values everywhere. I think the result is visually appealing, but I’m open to further iteration of course.

> I'm not sure the X/tick icon is better than having a readonly switch and may be quite disorientating when switching.

Agreed, I’ve used simple ‘yes’ ‘no’ labels for this now.

> For cards, forms, and input control I'd like to have three densities: 40px, 32px, 28px; can we easily do that?

I wouldn’t describe the task as “easy”. I’ve logged this for follow-up as #9296.

PR #9288, opened 2025-06-08
chore: Use deep compare instead of JSON.stringify

richburdon:


packages/plugins/plugin-kanban/src/components/KanbanContainer.stories.tsx
```diff
@@ -60,14 +61,15 @@ const StorybookKanban = () => {
     }
   }, [kanbans]);
 
-  useEffect(() => {
+  useDeepCompareEffect(() => {
```
dmaretskyi, line 63: using echo objects in deep compare effect needs additional investingation to make sure it doesn't recurse too deeply, e.g. through references to neighbouring objects

dmaretskyi (changes requested):
comment

packages/plugins/plugin-kanban/src/components/KanbanContainer.stories.tsx
richburdon: how can we test this?

dmaretskyi: Keep the normal effect and selectively apply useDeepCompareMemo to the deps of the effect like they do in the impl of useDeepCompareEffect

dmaretskyi: Echo objects should be compared by reference

PR #9303, opened 2025-06-11
feat: Resolve queue <-> database references; research graph building

dmaretskyi:


packages/core/echo/echo-db/src/core-db/automerge-doc-loader.ts
```diff
@@ -111,6 +112,14 @@ export class AutomergeDocumentLoaderImpl implements AutomergeDocumentLoader {
     this._spaceRootDocHandle = existingDocHandle;
   }
 
+  objectPresent(id: ObjectId): boolean {
```
richburdon: change to getObject which would be more generally useful?

packages/core/echo/echo-db/src/core-db/core-database.ts
```diff
@@ -358,44 +366,62 @@ export class CoreDatabase {
...
+    const diagnostics: string[] = [];
+    try {
+      return await new Promise((resolve, reject) => {
+        let unsubscribe: CleanupFn | null = null;
+        let inactivityTimeoutTimer: any | undefined;
+        const scheduleInactivityTimeout = () => {
+          inactivityTimeoutTimer = setTimeout(() => {
+            unsubscribe?.();
+            if (failOnTimeout) {
+              diagnostics.push('inactivity-rejected');
```
richburdon: create a simple discriminated union of events?

packages/core/echo/echo-db/src/hypergraph.ts
```diff
@@ -165,8 +210,9 @@ export class Hypergraph {
    * @returns Result of `onLoad`.
    */
   // TODO(dmaretskyi): Restructure API: Remove middleware, move `hostDb` into context option. Make accessible on Database objects.
-  getRefResolver(hostDb: EchoDatabase, middleware: (obj: BaseObject) => BaseObject = (obj) => obj): Ref.Resolver {
-    // TODO(dmaretskyi): Cache per hostDb.
+  createRefResolver({ context = {}, middleware = (obj) => obj }: RefResolverOptions): Ref.Resolver {
```
richburdon: nice. i think this deserves a type def.

packages/core/echo/echo-db/src/testing/queues.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    await builder.close();
+  });
+
+  test('create and resolve an object from a queue', async () => {
+    await using peer = await builder.createPeer({ types: [Testing.Contact] });
+    const spaceId = SpaceId.random();
+    const queues = peer.client.constructQueueFactory(spaceId);
+    const queue = queues.create();
+
+    const obj = create(Testing.Contact, {
```
richburdon:
```ts
const [obj] = queue.append([create(Test.Contact, { name: 'test' })]);
```

(compact, standardize testing data)

packages/plugins/plugin-assistant/src/stories/Query.stories.tsx
```diff
@@ -381,14 +382,55 @@ const ItemList = <T extends AnyEchoObject>({ items = [] }: { items?: T[] }) => {
   );
 };
 
+const FlushButton = ({ db }: { db?: EchoDatabase }) => {
```
richburdon: this would be much cleaner using a custom local hook pattern that provided the handler and state, with a regular button.

```tsx
const [state, icon, handleFlush] = useFlush();

return <IconButton onClick={handleFlush} />
```

dmaretskyi: 👍

packages/core/echo/echo-db/src/core-db/core-database.ts
dmaretskyi: its just logging, is I didn't bother

packages/core/echo/echo-db/src/core-db/automerge-doc-loader.ts
dmaretskyi: not really since object might be in a different doc, and that's an async operation

PR #9307, opened 2025-06-11
feat: Replace outliner with new CM extension

richburdon:


packages/plugins/plugin-outliner/src/types/types.ts
```diff
@@ -32,11 +33,11 @@ export interface OutlineType extends Schema.Schema.Type<typeof OutlineType> {}
 
 export const JournalEntryType = Schema.Struct({
   date: Schema.String, // TODO(burdon): Date.
-  tree: Ref(TreeType),
+  content: Ref(DataType.Text), // TODO(burdon): Breaks unless this is a reference.
```
wittjosiah, line 36: `DataType.Text` is another object so has to be reference, if you just want a string field it should be `Schema.String`

wittjosiah, line 36: Is there a reason that this is a separate type from `OutlineType`? Could `JournalType` entries just be `OutlineType`s?

packages/plugins/plugin-outliner/src/types/types.ts
```diff
@@ -48,47 +49,38 @@ export const JournalType = Schema.Struct({
...
   }),
 );
 
 export interface JournalType extends Schema.Schema.Type<typeof JournalType> {}
 
+//
+// Helpers
+//
+
+export const createOutline = (name?: string, content?: string): OutlineType => {
```
wittjosiah, line 62: ideally we wouldn't need to abstract this initialization away the defaults could just be specified on the schema, right?

packages/plugins/plugin-outliner/src/types/types.ts
richburdon: not currently but i expect they might diverge since currently the only thing in common is the text model

packages/plugins/plugin-outliner/src/types/types.ts
richburdon: yes, still waiting for dima to add the new make method that would do this

PR #9313, opened 2025-06-12
fix: Skip indexing canvas

dmaretskyi:


packages/core/echo/indexing/src/indexes/text.ts
```diff
@@ -2,9 +2,14 @@
 // Copyright 2025 DXOS.org
 //
 
-import { isEncodedReference, type ObjectStructure } from '@dxos/echo-protocol';
+import { decodeReference, isEncodedReference, ObjectStructure } from '@dxos/echo-protocol';
 import { visitValues } from '@dxos/util';
 
+/**
+ * Types that are excluded from text indexing.
```
wittjosiah, line 9: `@deprecated`?

wittjosiah (approved):

PR #9331, opened 2025-06-16
feat: Consistent ECHO APIs across queue and DB objects

dmaretskyi:


packages/core/echo/echo-db/src/echo-handler/echo-handler.ts
```diff
@@ -822,3 +821,234 @@ interface DecodedValueAtPath {
...
+  if (schema != null) {
+    internals.core.setType(requireTypeReference(schema));
+
+    const kind = getEntityKind(schema);
+    invariant(kind);
+    internals.core.setKind(kind);
+  }
+};
+
+const setRelationSourceAndTarget = (
```
richburdon, line 989: "target" here is ambiguous. prefer to pass core first, then call target "proxy" since i don't think it's the "target" referenced below (or in the name of the function)?

packages/core/echo/echo-db/src/hypergraph.ts
```diff
@@ -201,7 +203,21 @@ export class Hypergraph {
...
       },
-    };
+
+      resolveSchema: async (dxn) => {
+        switch (dxn.kind) {
+          case DXN.kind.TYPE: {
+            return this.schemaRegistry.getSchemaByDXN(dxn);
+          }
+          case DXN.kind.ECHO: {
+            throw new Error('Not implemented: Resolving schema stored in the database');
```
richburdon, line 212: Ambiguous. 
If "not implements" then add TODO.
If invalid state then throw error with context info (e.g., obj.id)

packages/core/echo/echo-schema/src/object/model.ts
```diff
@@ -0,0 +1,123 @@
+//
```
richburdon, line 1: why model.ts?
use standard types.ts for general defs?

dmaretskyi: it was deliberate, the file defines a **model** of the ECHO object

dmaretskyi: types.ts is to generic - becomes a dumping ground for all types

PR #9347, opened 2025-06-17
feat: Add backpreasure to query updates

mykola-vrmchk:
before
image

after
image


packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * The timeout for the task.
+   * in [ms].
+   */
+  timeout: number;
+};
+
+export type RunParams<T> = {
+  run: Task<T>['run'];
```
dmaretskyi: just inline those

packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+  timeout?: Task<T>['timeout'];
+}[];
+
+export class BackpressureExecutor<T> extends Resource {
```
dmaretskyi: I think a scheduler is a better name for this

TimedTaskScheduler?

packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+  private _executor?: DeferredTask = undefined;
+
+  constructor(private readonly _params: BackpressureExecutorParams) {
+    super();
+  }
+
+  protected override async _open(ctx: Context): Promise<void> {
+    this._executor = new DeferredTask(ctx, async () => {
+      // Clean up old tasks.
+      this._processedTasks = this._processedTasks.filter((task) => task.end > Date.now() - this._params.cleanUpAfter);
```
dmaretskyi: calls Date.now() repeatedly. Getting the time does have a CPU impact, so please pull it out from the loop

packages/common/async/src/backpressure-executor.ts
dmaretskyi: jsdoc

packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * The timeout for the task.
+   * in [ms].
+   */
+  timeout: number;
+};
+
+export type RunParams<T> = {
+  run: Task<T>['run'];
+  timeout?: Task<T>['timeout'];
```
dmaretskyi: specify @default in jsodc

dmaretskyi (changes requested):
I'm not certain we need this, how about just scheduling queries directly inside the QueryService? My intuition is that it's to early to generalize schduling and its better to write custom code for query scheduling

packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+          } finally {
+            task.end = Date.now();
+            task.elapsed = task.end - task.start;
+            this._processedTasks.push(task);
+          }
+        }),
+      );
+
+      if (this._executionQueue.length > 0) {
+        this._executor!.schedule();
```
richburdon: invariant

packages/common/async/src/backpressure-executor.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  protected override async _close(ctx: Context): Promise<void> {
+    this._executor = undefined;
+  }
+
+  /**
+   * Info method for debugging purposes.
+   */
+  get processedTasks(): Array<Task<T>> {
```
richburdon: getter up front after constructor

packages/common/async/src/backpressure-executor.ts
dmaretskyi: duplicated

packages/core/echo/echo-schema/src/ref/ref.ts
```diff
@@ -291,7 +292,7 @@ export interface RefResolver {
   // TODO(dmaretskyi): Combine with `resolve`.
   resolveSchema(dxn: DXN): Promise<Schema.Schema.AnyNoContext | undefined>;
 }
-
+@trace.resource()
```
dmaretskyi: no, this shouldn't be a resource: resource is to heavy for a ref. Ref is a lightweight API construct

you can put the span on database or hypergraph instead

packages/core/echo/echo-pipeline/src/db-host/query-service.ts
```diff
@@ -44,20 +45,35 @@ type ActiveQuery = {
 
 export class QueryServiceImpl extends Resource implements QueryService {
   private readonly _queries = new Set<ActiveQuery>();
+  private readonly _backpressureExecutor = new TimedTaskScheduler({
```
dmaretskyi: field name

packages/core/echo/echo-pipeline/src/db-host/query-service.ts
```diff
@@ -44,20 +45,35 @@ type ActiveQuery = {
...
-        } catch (err) {
-          log.catch(err);
-        }
-      }),
-    );
+        },
+      };
+    });
+
+    await Promise.all(this._backpressureExecutor.schedule(tasks));
```
dmaretskyi: the api would look cleaner with `.schedule(cb, options)`

dmaretskyi: is there any benifit in `schedule` taking an array of tasks? 
alternatively the code could look like:

```ts
await Promise.all(Array.from(this._queries).map(query => this._backpressureExecutor.schedule( async () => {
          try {
            const { changed } = await query.executor.execQuery();
            if (changed) {
              query.sendResults(query.executor.getResults());
            }
          } catch (err) {
            log.catch(err);
          }	          }
        } catch (err) {	        },)))

**Review by dmaretskyi — DISMISSED** [review]

File: packages/core/echo/echo-pipeline/src/db-host/query-service.ts [hunk 10]
```diff
@@ -118,17 +135,21 @@ export class QueryServiceImpl extends Resource implements QueryService {
...
-        } catch (error: any) {
-          close(error);
-        }
-      });
+      this._backpressureExecutor
+        .schedule([
+          {
+            run: async () => {
+              // Run query executor.
+              await queryEntry.executor.open();
```
dmaretskyi: pull out open

packages/core/echo/echo-pipeline/src/db-host/query-service.ts
```diff
@@ -118,17 +135,21 @@ export class QueryServiceImpl extends Resource implements QueryService {
...
+              await queryEntry.executor.execQuery();
+              // Always send first result set.
+              queryEntry.sendResults(queryEntry.executor.getResults());
+            },
+          },
+        ])[0]
+        .catch((err) => {
```
dmaretskyi: i think try { } catch {} looks cleaner then callback chaining

packages/core/echo/echo-pipeline/src/scheduler/timed-task-scheduler.ts
```diff
@@ -0,0 +1,174 @@
...
+      this._processedTasks = this._processedTasks.filter((task) => task.end > now - this._params.cleanUpAfter);
+
+      // Early return if there are no tasks to run.
+      if (this._executionQueue.length === 0) {
+        return;
+      }
+
+      // Check if the budget is exceeded.
+      if (this._processedTasks.length > 0) {
+        const lastPeriodTasks = this._processedTasks.filter(
```
dmaretskyi: this will break down if you have multiple concurent tasks that are waiting for IO. You're gonna be counting the elapsed time for each one separatelly, even though they all executed concurently

packages/core/echo/echo-pipeline/src/scheduler/timed-task-scheduler.ts
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * The maximum number of tasks to run in parallel.
+   */
+  maxParallelTasks: number;
+
+  /**
+   * Clean executed tasks after time period.
+   * in [ms].
+   */
+  cleanUpAfter: number;
```
dmaretskyi: what is the idea behind this setting? what should it be set to?

packages/core/echo/echo-pipeline/src/scheduler/timed-task-scheduler.ts
dmaretskyi: is there a repro of the hanging issue so we can compare before and after?

packages/core/echo/echo-pipeline/src/scheduler/timed-task-scheduler.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  protected override async _close(ctx: Context): Promise<void> {
+    this._executor = undefined;
+  }
+
+  /**
+   * Adds tasks to execution queue.
+   * @returns An array of promises that resolve to the results of the tasks.
+   */
+  schedule(
```
dmaretskyi: could be just singular `schedule(cb, options)` - makes the api cleaner

packages/core/echo/echo-pipeline/src/db-host/query-service.ts
```diff
@@ -44,20 +45,35 @@ type ActiveQuery = {
 
 export class QueryServiceImpl extends Resource implements QueryService {
   private readonly _queries = new Set<ActiveQuery>();
+  private readonly _backpressureExecutor = new TimedTaskScheduler({
+    budget: 500,
+    budgetPeriod: 1000,
```
dmaretskyi, line 50: here's an example I want to test that works:

seed the data so there's 10 objects of type Task. Now run 10 concurrent queries of type Task. Each one should return the same 10 tasks. Now, lets assume that the IO time to load them is 50 milliseconds (accessing the disk). Perhaps we can model that with an AutomergeStorageAdapter middleware. Would this cause us to go into the cooldown in the executor?

mykola-vrmchk: No it should not

packages/core/echo/echo-pipeline/src/automerge/automerge-host.ts
```diff
@@ -86,6 +87,13 @@ export class AutomergeHost extends Resource {
   private readonly _db: LevelDB;
   private readonly _indexMetadataStore: IndexMetadataStore;
   private readonly _echoNetworkAdapter: EchoNetworkAdapter;
+  private readonly _syncScheduler = new TimedTaskScheduler({
+    budget: 500,
+    budgetPeriod: 1000,
+    cooldown: 500,
+    maxParallelTasks: 10,
+    saveHistoryFor: 30_000,
```
dmaretskyi, line 95: what does this mean?

PR #9361, opened 2025-06-18
feat: hashtag colors

richburdon:


packages/ui/react-ui-theme/src/util/hash.ts
```diff
@@ -6,30 +6,31 @@ export type Color = {
...
+  { color: 'emerald', bg: 'bg-emerald-500', text: 'text-emerald-500', tag: 'bg-emerald-300 dark:bg-emerald-700' },
+  { color: 'teal', bg: 'bg-teal-500', text: 'text-teal-500', tag: 'bg-teal-300 dark:bg-teal-700' },
+  { color: 'cyan', bg: 'bg-cyan-500', text: 'text-cyan-500', tag: 'bg-cyan-300 dark:bg-cyan-700' },
+  { color: 'sky', bg: 'bg-sky-500', text: 'text-sky-500', tag: 'bg-sky-300 dark:bg-sky-700' },
+  { color: 'indigo', bg: 'bg-indigo-500', text: 'text-indigo-500', tag: 'bg-indigo-300 dark:bg-indigo-700' },
+  { color: 'violet', bg: 'bg-violet-500', text: 'text-violet-500', tag: 'bg-violet-300 dark:bg-violet-700' },
+  { color: 'purple', bg: 'bg-purple-500', text: 'text-purple-500', tag: 'bg-purple-300 dark:bg-purple-700' },
+  { color: 'fuchsia', bg: 'bg-fuchsia-500', text: 'text-fuchsia-500', tag: 'bg-fuchsia-300 dark:bg-fuchsia-700' },
+  { color: 'rose', bg: 'bg-rose-500', text: 'text-rose-500', tag: 'bg-rose-300 dark:bg-rose-700' },
+  { color: 'pink', bg: 'bg-pink-500', text: 'text-pink-500', tag: 'bg-pink-300 dark:bg-pink-700' },
```
thure: We can easily mint semantic tokens for these, we should instead extend `sememes-hue.ts`. These particular shades may already be semantic tokens.

We do already have a way of applying styles like this in pure CSS, and we should probably augment that system for this. See https://github.com/dxos/dxos/blob/main/packages/ui/react-ui-theme/src/styles/layers/hues.css

thure (changes requested):
We should always prefer semantic tokens since physical tokens are untracked and not guaranteed to be available.

thure (approved):
Adjusted to apply semantic tokens.

PR #9362, opened 2025-06-18
feat: Research story fix

richburdon:
Screenshot 2025-06-18 at 4 12 22 PM


packages/core/ai/src/util/printer.ts
```diff
@@ -8,25 +8,40 @@ import type { Message, MessageContentBlock } from '../tools';
...
+  mode?: Mode;
+};
+
 export class ConsolePrinter {
+  logger: Logger;
   mode: Mode;
-  constructor({ mode = 'text' }: { mode?: Mode } = {}) {
+
+  // eslint-disable-next-line no-console
+  constructor({ logger = console.log, mode = 'text' }: ConsolePrinterOptions = {}) {
```
wittjosiah, line 24: is there a reason you don't want to use `@dxos/log` here?

packages/core/assistant/src/blueprint/logger.ts
```diff
@@ -2,41 +2,84 @@
...
 
 /* eslint-disable no-console */
 
-// Force chalk colors on for tests
+// Force chalk colors on for tests.
 chalk.level = 2;
 
-export const setConsolePrinter = (machine: BlueprintMachine, extraLogging = false) => {
+export class Logger {
+  private _messages = signal<string[]>([]);
```
wittjosiah, line 20: what's the use of this being a signal if you never update it?

packages/core/ai/src/util/printer.ts
richburdon: i'm not sure; refactoring @dmaretskyi; i don't think so.

packages/core/assistant/src/blueprint/logger.ts
richburdon: it is updated via the log method.

PR #9363, opened 2025-06-19
fix: JSON schema hack

richburdon:


packages/core/echo/echo-schema/src/json/json-schema.ts
```diff
@@ -76,11 +76,37 @@ export const toPropType = (type?: PropType): string => {
...
+      }
+
+      return false;
+    });
+  }
+
+  return jsonSchema;
+};
+
+export const _toJsonSchema = (schema: Schema.Schema.All): JsonSchemaType => {
```
dmaretskyi: better naming; don't export

packages/core/echo/echo-schema/src/json/json-schema.ts
```diff
@@ -76,11 +76,37 @@ export const toPropType = (type?: PropType): string => {
...
 /**
  * Convert effect schema to JSON Schema.
  * @param schema
  */
-export const toJsonSchema = (schema: Schema.Schema.All): JsonSchemaType => {
+export const toJsonSchema = (schema: Schema.Schema.All, options: JsonSchemaOptions = {}): JsonSchemaType => {
+  let jsonSchema = _toJsonSchema(schema);
+  if (options.strict) {
+    // TOOD(burdon): Workaround to ensure JSON schema is valid (for agv parsing).
+    jsonSchema = removeProperties(jsonSchema, (key, value) => {
```
dmaretskyi, line 91: this will not decode correctly into effect-schema, but probably fine to leave in the API

PR #9368, opened 2025-06-19
refactor: More more cardlike patterns into `Card` exemplar

thure:
This PR does what it says on the tin. It should cause no visible changes.

### DIFF
--- a/packages/plugins/plugin-preview/src/capabilities/react-surface.tsx
+++ b/packages/plugins/plugin-preview/src/capabilities/react-surface.tsx
@@ -22,6 +22,7 @@ import { descriptionMessage } from '@dxos/react-ui-theme';
 import { DataType } from '@dxos/schema';
 
 import { ContactCard, OrganizationCard, ProjectCard } from '../components';
+import { CardContainer } from '../components/CardContainer';
 import { PREVIEW_PLUGIN } from '../meta';
 
 export default () =>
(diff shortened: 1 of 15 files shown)


ZaymonFC (approved):
Looks good to me.

packages/plugins/plugin-preview/src/capabilities/react-surface.tsx
```diff
@@ -65,9 +65,11 @@ export default () =>
           [dispatch],
         );
         return (
-          <ContactCard subject={data.subject} onOrgClick={handleOrgClick} role={role}>
-            {role === 'popover' && <Surface role='related' data={data} />}
-          </ContactCard>
+          <CardContainer role={role}>
```
wittjosiah: any reason why this isn't within `ContactCard`?

thure: I gave this a lot of thought – one of the problems with any pattern that can be called "card" is where it gets its encapsulation, since that is a common pattern across the system. Encapsulated patterns might host a card, like a popover, so the card might not need to provide its own encapsulation.

We've already established a pattern of signaling surface situations through `role`, though since this concept diverges from the HTML attribute it seems most appropriate at first blush to handle that in pluginland.

I'm amenable to moving this or finding another way of tracking encapsulation, just don't want to optimize overly early, hence the current proposal.

thure: Ah, turns out we need this for other plugins so I’m already going to move it upstream.

wittjosiah: I didn't necessarily mean removing `CardContainer` but questioning why `<CardContainer role={role}>` wasn't a part of the `ContactCard`, which I would still ask even if its factor out.

PR #9376, opened 2025-06-21
fix: Reduction of border radii & shadows, adjustment of L0 main menu trigger

richburdon:
Screenshot 2025-06-23 at 4 54 57 PM


thure:
I think this discards a useful signifier, the larger L0 which acts as the pivot for the entire app. This pattern is present across the ecosystem, and I would argue users flock to pivot patterns with even increased size and functionality e.g. Arc & Zen.

I am closing this because a move in this direction diminishes our prioritization of function over form; I think we can address the form issues this PR set out to address without diminishing function. Please feel free to reopen, however, if we should continue the discussion.

richburdon:
> I think this discards a useful signifier, the larger L0 which acts as the pivot for the entire app. This pattern is present across the ecosystem, and I would argue users flock to pivot patterns with even increased size and functionality e.g. Arc & Zen.
> 
> I am closing this because a move in this direction diminishes our prioritization of function over form; I think we can address the form issues this PR set out to address without diminishing function. Please feel free to reopen, however, if we should continue the discussion.

we need to discuss this. the PR contains a number of other modification and clean-ups.

thure (changes requested):
We can't prioritize form over function, this is a user-hostile move. Let's discuss ways to resolve the alignment issue which gives us better opportunities to deliver value.

PR #9379, opened 2025-06-21
fix: debug panel

wittjosiah:


wittjosiah:
Succeeded by https://github.com/dxos/dxos/pull/9378

PR #9395, opened 2025-06-24
fix: minor form ux

richburdon:
- [x] System schema message padding
- [x] List padding (so that drag marker isn't partially cut-off)
- [x] Correctly handle readonly form (control visibility)
- [x] Consistent padding of labels and list rows (alignment of visible and non-visible items)

TOOD
- [ ] Should the View itself be readonly or just the schema be immutable (I think the latter, but currently if you try to edit the View title, it breaks, hence I've temporarily made those filed immutable also.
- [ ] IMPORTANT: Readonly fields should be ghosted (e.g., uneditable, but have the textfield outline taking up the same space).
- [ ] Make it simple to make Lists part of Forms without custom padding
<br/>

Screenshot 2025-06-24 at 10 30 21 AM


packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -48,7 +48,7 @@ export const ViewEditor = ({
   schema,
   view,
   registry,
-  readonly,
+  readonly: _readonly,
```
thure, line 51: I’ve been using  e.g. `propsReadonly` in cases like this, since prefixing with an underscore usually indicates you are discarding or otherwise not using a variable but need to declare it.

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -152,8 +152,8 @@ export const ViewEditor = ({
 
   return (
     <div role='none' className={mx('overflow-y-auto', classNames)}>
-      {immutable && (
-        <div role='none' className='mbe-card-spacing-block'>
+      {readonly && (
+        <div role='none' className='p-3'>
```
thure:
```suggestion
        <div role='none' className='plb-card-spacing-block pli-card-spacing-inline'>
```
This will connect this to future work on density.

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -162,52 +162,57 @@ export const ViewEditor = ({
...
+          readonly={readonly}
+          onMove={readonly ? undefined : handleMove}
         >
           {({ items: fields }) => (
             <>
               {showHeading && (
                 <div role='heading' className={grid}>
                   <div />
-                  <div className='flex items-center text-sm'>{t('field path label')}</div>
+                  <div className='flex px-2 items-center text-sm'>{t('field path label')}</div>
```
thure:
```suggestion
                  <div className='flex pli-card-spacing-inline items-center text-sm'>{t('field path label')}</div>
```

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -162,52 +162,57 @@ export const ViewEditor = ({
...
                 </div>
               )}
 
-              <div role='list' className='flex flex-col w-full'>
+              <div role='list' className='flex flex-col w-full px-2'>
```
thure:
```suggestion
              <div role='list' className='flex flex-col w-full pli-card-spacing-inline'>
```

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -224,7 +229,7 @@ export const ViewEditor = ({
 
         {hiddenProperties.length > 0 && (
           <div>
-            <div role='none' className='pli-card-spacing-inline mlb-card-spacing-block'>
+            <div role='none' className='pli-card-spacing-inline px-2'>
```
thure: `px-2` conflicts with `pli-card-spacing-inline`, and the latter has precedence. If this did not need margins on the block axis and the inline padding is okay, please remove `px-2`.

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -234,20 +239,22 @@ export const ViewEditor = ({
               getId={(property) => property}
             >
               {({ items: properties }) => (
-                <div role='list' className='flex flex-col w-full'>
+                <div role='list' className='flex flex-col w-full px-2'>
```
thure:
```suggestion
                <div role='list' className='flex flex-col w-full pli-card-spacing-inline'>
```

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
richburdon: noted. i've been doing this consistently across 100s of files, so it's an aspect we should discuss and standardize on.

PR #9398, opened 2025-06-24
fix(ViewEditor): Initialization issues

zaymonoid:
Fix `EchoSchema` handling in table/kanban components for proper schema editing.

- Ensures `EchoSchema` instances are handled correctly by accessing `.jsonSchema` property directly instead of converting through `Type.toJsonSchema()` which disconnects schema editing persistence.
- Fixes deletion not working in table view editor.

### DIFF
--- a/packages/apps/todomvc/src/components/SpaceList.tsx
+++ b/packages/apps/todomvc/src/components/SpaceList.tsx
@@ -39,7 +39,7 @@ export const SpaceList = ({ current }: { current?: Space }) => {
         <button onClick={() => current && client.shell.shareSpace({ spaceId: current.id })} data-testid='share-button'>
           ↸
         </button>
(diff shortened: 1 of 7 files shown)


packages/plugins/plugin-table/src/components/TableContainer.tsx
```diff
@@ -69,7 +70,8 @@ const TableContainer = ({ role, table }: { role?: string; table: TableType }) =>
       return;
     }
 
-    return new ViewProjection(Type.toJsonSchema(schema), table.view.target!);
+    const jsonSchema = schema instanceof EchoSchema ? schema.jsonSchema : Type.toJsonSchema(schema);
+    return new ViewProjection(jsonSchema, table.view.target);
   }, [schema, table.view?.target]);
```
wittjosiah: stringify the schema in deps in kanban but not here?

PR #9400, opened 2025-06-24
test: Effect HTTP request

richburdon:


packages/core/mesh/edge-client/src/edge-http-client.ts
```diff
@@ -64,96 +96,115 @@ export class EdgeHttpClient {
     }
   }
 
+  //
+  // Status
+  //
+
+  // TODO(burdon): Implement status endpoint.
+  public async getStatus(args?: EdgeHttpGetArgs): Promise<GetStatusResponseBody> {
+    return this._call(new URL('/status', this.baseUrl), { ...args, method: 'GET' });
```
mykola-vrmchk, line 109: Why do you move URL creation from `._call()`? Will there ever be a case when you do not want to use `this.baseUrl` for a request?

packages/core/mesh/edge-client/src/edge-http-client.ts
```diff
@@ -64,96 +96,115 @@ export class EdgeHttpClient {
...
-    }
-    if (query.objectIds != null) {
-      queryParams.set('objectIds', query.objectIds.join(','));
-    }
-    return this._call(`/spaces/${subspaceTag}/${spaceId}/queue/${queueId}/query?${queryParams.toString()}`, {
-      ...args,
-      method: 'GET',
-    });
+    return this._call(
+      createUrl(new URL(`/spaces/${subspaceTag}/${spaceId}/queue/${queueId}/query`, this.baseUrl), {
```
mykola-vrmchk, line 200: Move createUrl inside `._call`

packages/core/protocols/src/edge.ts
```diff
@@ -22,6 +22,7 @@ export type EdgeHttpSuccess<T> = {
 export type EdgeErrorData = { type: string } & Record<string, any>;
 
 export type EdgeHttpFailure = {
+  // TODO(burdon): Why is this required?
```
mykola-vrmchk, line 25: That is a discriminated union pattern
```
type Success = {
  success: true;
  message: string;
};

type Failure = {
  success: false;
  error: string;
};

type Result = Success | Failure;
```

richburdon: it felt strange to do this in the protocol (protobuf)?

packages/core/mesh/edge-client/src/edge-http-client.ts
richburdon: separation of concerns so that the call method takes a well formed URL

PR #9404, opened 2025-06-25
feat: Alert component

richburdon:
Screenshot 2025-06-24 at 9 35 12 PM

<hr/>
Screenshot 2025-06-24 at 9 34 55 PM

<hr/>
Screenshot 2025-06-24 at 9 37 05 PM


packages/ui/react-ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,25 @@
...
+    title?: string;
+  }>
+>;
+
+// TODO(burdon): Severity tag to color background?
+export const Alert = ({ classNames, title, children }: AlertProps) => {
+  const { tx } = useThemeContext();
+  return (
+    <div role='alert' className={tx('alert.root', 'alert', {}, classNames)}>
+      {title && <h2 className={tx('alert.title', 'title')}>{title}</h2>}
```
thure: `react-ui` core does not do subcomponents like this — please create a subcomponent `Alert.Title` which does this.

packages/ui/react-ui-theme/src/styles/components/alert.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { type Theme, type ComponentFunction } from '@dxos/react-ui-types';
+
+import { mx } from '../../util';
+import { descriptionMessage } from '../fragments';
+
+export type AlertStyleProps = {};
+
+export const alertRoot: ComponentFunction<AlertStyleProps> = (_, etc) => {
+  return mx('max-w-[calc(100%-2rem)] is-full m-4 place-self-center text-center rounded-md', descriptionMessage, etc);
```
thure: Components from `react-ui` core should not subtract from full width or apply outer spacing like this, it assumes too much about the layout/spacing environment it’s rendered within.

packages/ui/react-ui-form/src/components/ViewEditor/ViewEditor.tsx
```diff
@@ -152,16 +152,7 @@ export const ViewEditor = ({
...
-        <div role='none' className='plb-card-spacing-block pli-card-spacing-inline'>
-          <Message.Root valence='neutral' className='rounded'>
-            <Message.Title>
-              <Icon icon='ph--info--regular' size={5} classNames='inline' /> {t('system schema title')}
-            </Message.Title>
-            <Message.Body>{t('system schema description')}</Message.Body>
-          </Message.Root>
-        </div>
-      )}
+      {readonly && <Alert title={t('system schema title')}>{t('system schema description')}</Alert>}
```
thure: This discards all of the card spacing, which I’m working hard to enforce — please refactor `Alert` not to apply its own outer spacing, that is context-sensitive and should be applied by hosts as with all other components in `react-ui` core.

thure (changes requested):
This will be useful, thank you.

I’d like to land #9403, then we should address all of these physical spacing tokens, they need to be refactored to use density-sensitive alias tokens.

packages/ui/react-ui-theme/src/styles/components/alert.ts
thure: If this intends to *force* a specific spacing around the element, it should wrap everything in a `div role='none'` which applies internal spacing, rather than trying to be robust to whatever display mode the host might be applying.

packages/ui/react-ui/src/components/Alert/Alert.tsx
thure: `Alert.Content` would probably also be helpful, even if it just renders a `Fragment` for now.

thure (approved):
I’ve merged this with `Message` and left `Callout` as a synonym. I’ve renamed the story `Callout` in case that is a less ambiguous term for this component.

I’ve adjusted the anatomy of both slightly in an effort to merge the best of both implementations.

PR #9407, opened 2025-06-25
test: Blueprint to circuit compiler

dmaretskyi:
- [x] Blueprint to circuit compiler
- [x] GPT node conversation passthrough (needed to chain GPT steps)
- [x] GPT node should use `AiSession`
- [ ] GPT node should self-report status like blueprint does.

---

image

### DIFF
--- a/packages/common/async/src/events.test.ts
+++ b/packages/common/async/src/events.test.ts
@@ -22,7 +22,7 @@ describe('Event', () => {
       pureCount++;
     });
 
-    const debounced = event.debounce(80);
+    const debounced = event.debounce(200);
     debounced.on(() => {
       debounceCount++;
     });


packages/core/artifact-testing/src/experimental.test.ts
```diff
@@ -2,12 +2,188 @@
...
+  });
+
+  test('conversation', { timeout: 120_000 }, async () => {
+    const { client } = serviceContainer.getService(AiService);
+    const session = new AISession({
+      operationModel: 'configured',
+    });
+
+    // TODO(dmaretskyi): Use tool registry.
+    const sage = createTool('test', {
```
richburdon: funny but let's start to develop something mildly more useful in tests like the calculator in my ai.test.ts since it actually uses params and schema. building up a repertoire of tools via tests would also be useful.

richburdon (approved):

PR #9408, opened 2025-06-25
feat: Query.without

wittjosiah:


packages/core/echo/echo-pipeline/src/query/plan.ts
```diff
@@ -176,4 +183,14 @@ export namespace QueryPlan {
 
     plans: Plan[];
   };
+
+  /**
+   * Subtract the results of one plan from another.
```
dmaretskyi, line 188: Give them descriptive names: source, exclude

dmaretskyi (approved):
LGTM!

PR #9414, opened 2025-06-26
fix: Space settings UI

richburdon:
- [x] Full width Space list with border
- [ ] NOTE: ListHeading should center and truncate text.

Screenshot 2025-06-26 at 12 05 14 AM

### DIFF
--- a/packages/plugins/plugin-space/src/components/SpacePluginSettings.tsx
+++ b/packages/plugins/plugin-space/src/components/SpacePluginSettings.tsx
@@ -8,7 +8,7 @@ import { createIntent, useIntentDispatcher } from '@dxos/app-framework';
 import { useClient } from '@dxos/react-client';
 import { useSpaces } from '@dxos/react-client/echo';
 import { Input, toLocalizedString, useTranslation, List, ListItem, IconButton } from '@dxos/react-ui';
-import { DeprecatedFormContainer, DeprecatedFormInput } from '@dxos/react-ui-form';
+import { controlItemClasses, ControlPage, ControlSection, DeprecatedFormInput } from '@dxos/react-ui-form';
 
 import { SPACE_PLUGIN } from '../meta';
 import { SpaceAction, type SpaceSettingsProps } from '../types';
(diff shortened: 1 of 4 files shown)


packages/plugins/plugin-space/src/components/SpacePluginSettings.tsx
```diff
@@ -27,21 +27,28 @@ export const SpacePluginSettings = ({ settings }: { settings: SpaceSettingsProps
...
+            {spaces.map((space) => (
+              <ListItem.Root key={space.id} classNames='items-center gap-4'>
+                {/* TODO(burdon): Should auto center and truncate; NOTE truncate doesn't work with flex grow. */}
+                <ListItem.Heading classNames='flex w-full items-center truncate'>
+                  {toLocalizedString(getSpaceDisplayName(space, { personal: space === client.spaces.default }), t)}
+                </ListItem.Heading>
+                <ListItem.Endcap>
+                  <IconButton
+                    icon='ph--faders--regular'
+                    iconOnly
```
thure:
```suggestion                 
```
This is not the place to omit labels I think, there is plenty of room so this seems unnecessarily enigmatic.

packages/plugins/plugin-space/src/components/SpacePluginSettings.tsx
```diff
@@ -27,21 +27,28 @@ export const SpacePluginSettings = ({ settings }: { settings: SpaceSettingsProps
...
-              <ListItem.Endcap>
-                <Button onClick={() => dispatch(createIntent(SpaceAction.OpenSettings, { space }))}>
-                  {t('open space settings label')}
-                </Button>
-              </ListItem.Endcap>
-            </ListItem.Root>
-          ))}
-        </List>
+        <h2 className='text-xl my-4'>{t('space settings label')}</h2>
+        {/* TODO(burdon): Consider fragment for a callout border section. */}
```
thure: There is one (for settings pages like this, these are meaningfully distinct from callouts): `controlItemClasses`, it can be imported from `react-ui-form`.

I’d recommend using the component `ControlItem` instead, though, since it will apply the patterns used everywhere else in settings.

packages/plugins/plugin-space/src/components/SpacePluginSettings.tsx
```diff
@@ -27,21 +27,28 @@ export const SpacePluginSettings = ({ settings }: { settings: SpaceSettingsProps
...
+        <div className='-mis-4 -mie-4 p-4 border border-separator rounded-md'>
```
thure: In case `ControlItem` or `controlItemClasses` is not of use, please use `cardSpacing` tokens since they will be `coarse` in this context (`ControlItem` et al should be refactored to match):

```suggestion
        <div className='plb-cardSpacingBlock pli-cardSpacingInline border border-separator rounded-md'>
```

Negative margins should not be used here since that may cause this section to exceed the width of the page in narrower viewports — internal spacing should be applied to the element hosting all this, and each nonencapsulated element of the page should apply the same `pli-cardSpacingInline` so they all line-up while also fitting in the page.

thure (changes requested):
Actually I think this should be improved to connect it with ongoing work related to spacing and to settings pages.

thure (approved):
I’ve adjusted this to use existing componentry. I’ve adjusted the content to better align with other settings pages:

Screenshot 2025-06-25 at 22 21 35

PR #9415, opened 2025-06-26
fix: array reference indexing

wittjosiah:


packages/core/echo/indexing/src/indexes/index-graph.ts
```diff
@@ -177,11 +177,11 @@ export class IndexGraph extends Resource implements Index {
             continue;
           }
           if (property !== null) {
-            const source = sources.get(property);
-            if (!source) {
-              continue;
+            for (const [prop, source] of sources.entries()) {
+              if (prop === property || prop.startsWith(`${property}.`)) {
```
dmaretskyi: By using starts with this will also match paths as 'object.3.foo.bar.baz'.


Could you make this stricter so that it's just a single index segment that's allowed after the key and no more keys?

dmaretskyi (changes requested):

PR #9425, opened 2025-06-27
fix(tables): Ref query lookup for objects with system schema

ZaymonFC:


thure (approved):
lgtm ✨

packages/ui/react-ui-table/src/components/Table/Table.tsx
```diff
@@ -77,13 +78,14 @@ export type TableMainProps = {
   model?: TableModel;
   presentation?: TablePresentation;
   schema?: Schema.AnyNoContext;
+  client?: Client;
```
wittjosiah, line 81: this is a react-ui component, it really shouldn't know about client...

honestly it probably shouldn't know about space either. These should be callbacks that returns schemas, objects, etc.

PR #9429, opened 2025-06-28
chore: toolbar/tabs increment

richburdon:
- [x] Ghost inactive tabs
- [x] New var and size for toolbar height
 
image


packages/ui/react-ui-theme/src/styles/components/toolbar.ts
```diff
@@ -8,10 +8,11 @@ import { mx } from '../../util';
 
 export type ToolbarStyleProps = Partial<{ layoutManaged: boolean }>;
 
-export const toolbarLayout = 'flex flex-nowrap shrink-0 items-center gap-1 p-1 overflow-x-auto contain-layout';
+export const toolbarLayout =
+  'bs-full is-full flex flex-nowrap shrink-0 items-center gap-1 p-2 overflow-x-auto scrollbar-none contain-layout';
```
thure: Why `bs-full`? This will cause issues if the toolbar is anywhere other than a grid area or containing element that happens to apply the right size. A better default, if you want the consumer to _need_ to override the toolbar size, would be `bs-[--toolbar-size]`.

PR #9430, opened 2025-06-28
fix: Update component while rendering a different component

richburdon:
- [x] `fullyQualifiedId` returns a non-stable composite `id` which is used to trigger hooks while rendering.
- [x] resolves access of connections via what was wishfully being provided as a hook via a property :).
- [ ] TODO: Seems to re-introduce a slowdown in the editor!

## Fixes:

Screenshot 2025-06-28 at 1 58 08 PM

image

### DIFF
--- a/packages/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
+++ b/packages/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
@@ -17,8 +17,8 @@ import {
   useLayout,
 } from '@dxos/app-framework';
 import { isAction, isActionLike, ROOT_ID, type Node, type ReadableGraph } from '@dxos/app-graph';
+import { log } from '@dxos/log';
 import { PLANK_COMPANION_TYPE } from '@dxos/plugin-deck/types';
-import { useConnections } from '@dxos/plugin-graph';
 import { useMediaQuery, useSidebars } from '@dxos/react-ui';
 import { isTreeData, type TreeItemDataProps, type TreeData } from '@dxos/react-ui-list';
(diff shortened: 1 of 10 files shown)


packages/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
```diff
@@ -75,16 +67,26 @@ export const NavTreeContainer = memo(({ tab, popoverAnchorId, topbar }: NavTreeC
   const { dispatchPromise: dispatch } = useIntentDispatcher();
   const { graph } = useAppGraph();
   const { isOpen, isCurrent, isAlternateTree, setItem } = useCapability(NavTreeCapabilities.State);
-  const layout = useLayout();
   const { navigationSidebarState } = useSidebars(NAVTREE_PLUGIN);
+  const layout = useLayout();
 
   const getActions = useCallback((node: Node) => naturalGetActions(graph, node), [graph]);
+  const getChildItems = (node?: Node, { disposition, sort = false }: TraveralOptions = {}) => {
```
wittjosiah, line 74: `graph.getConnections` is not reactive, this is only "working" because this function isn't wrapped in `useCallback` and so it likely forcing excessive re-renders (which is probably the source of the perf issue you mentioned)

I was also skeptical of passing a custom hook in via props, but from testing it works fine as long as its stable

wittjosiah (changes requested):

packages/plugins/plugin-navtree/src/components/NavTreeContainer.tsx
richburdon: +1

PR #9433, opened 2025-06-29
feat: Schema records

richburdon:
StoredSchema => ViewType Relation


packages/core/echo/echo/src/Relation.ts
```diff
@@ -2,18 +2,44 @@
...
 
+import type * as Obj from './Obj';
+
 export type Any = EchoSchema.AnyEchoObject & EchoSchema.RelationSourceTargetRefs;
 
 export const Source = EchoSchema.RelationSourceId;
 export const Target = EchoSchema.RelationTargetId;
 
-export const make = LiveObject.live;
+type Obj<T extends EchoSchema.BaseObject, Source extends Obj.Any, Target extends Obj.Any> = Omit<
```
dmaretskyi: is Relation.Obj intentional?

packages/core/echo/echo/src/Relation.ts
```diff
@@ -2,18 +2,44 @@
...
+  EchoSchema.ExcludeId<T>,
+  typeof EchoSchema.RelationSourceId | typeof EchoSchema.RelationTargetId
+> & {
+  [Source]: Source;
+  [Target]: Target;
+};
+
+// TODO(burdon): Narrow generics to match type of source and target schemas.
+export const make = <T extends EchoSchema.BaseObject, Source extends Obj.Any, Target extends Obj.Any>(
```
wittjosiah: Source/Target key types are objects?

We had omitted this api intentionally because of its ability to be used interchangeably with `Obj.make`. I think if its going ship we should ensure that it only works for making relations

dmaretskyi: working on this rn

PR #9438, opened 2025-07-01
fix: command menu

wittjosiah:


packages/ui/react-ui-editor/src/extensions/command/command-menu.ts
```diff
@@ -92,7 +92,7 @@ export const commandMenu = (options: CommandMenuOptions) => {
           // Insert and select the trigger.
           view.dispatch({
             changes: { from: selection.head, insert: trigger },
-            selection: { anchor: selection.head, head: selection.head + 1 },
+            selection: { anchor: selection.head + 1, head: selection.head + 1 },
```
wittjosiah, line 95: this is the primary fix, without this the menu would close when typing

thure (approved):
lgtm!

PR #9461, opened 2025-07-04
chore: Test reporter tool

richburdon:
- [x] Experimental workflow via BuildJet (2x speed of GH Actions): Completes in 5m
  - https://app.buildjet.com/74398940
  - https://docs.google.com/spreadsheets/d/1Denfaa_K4igfULkaZC_zk25uRrRHxkP-yBgsoeG_x8k/edit?gid=0#gid=0 (cost model)
- [x] CLI tool to poll for JSON report artifact and display errors with (clickable file links).

```bash
./scripts/gh-action.mjs --watch --truncate
```

### ---watch

Screenshot 2025-07-05 at 1 28 09 AM

### --truncate

Screenshot 2025-07-05 at 1 40 49 AM

### --all

Screenshot 2025-07-05 at 2 07 59 AM


.github/actions/setup/action.yml
```diff
@@ -4,6 +4,10 @@ description: Setup the environment for the other workflows
 runs:
   using: 'composite'
   steps:
+    - name: Mark workspace as safe
```
wittjosiah, line 7: what caused this to be needed? I spend time yesterday trying to sort out the permissions in the container so that this and other workarounds weren't needed

.github/workflows/docker-image-circleci.yml
```diff
@@ -1,4 +1,4 @@
-name: CircleCI Docker Container
+name: Build and Push Docker Image
```
wittjosiah, line 1: note that the name here is what shows up in the github ui so if multiple workflows have the same name they won't be distinguishable in the ui (also if the names are too long it just truncates them)

package.json
```diff
@@ -34,6 +34,11 @@
     "watch-build": "./node_modules/.bin/nx watch --all -- pnpm -w nx build \"\\$NX_PROJECT_NAME\""
   },
   "dependencies": {
+    "@1password/op-js": "^0.1.13",
```
wittjosiah, line 37: why deps vs dev deps? probably doesn't matter at the root but seems inconsistent

scripts/gh-action.mjs
wittjosiah, line 1: nice

.github/workflows/docker-image-circleci.yml
richburdon: I"ll fix

.github/actions/setup/action.yml
richburdon: The job failed immediately without this.

PR #9469, opened 2025-07-07
feat: Board increment

richburdon:
- [x] Add and delete tiles (cards).
- [x] Drag to move cards.
- [x] Updated i18n translations definitions (make consistent).
- [x] Updated kanban type exports (make consistent).

Next
- [ ] Surfaces.
- [ ] Plugin with reactive ECHO objects.


packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
...
+      style={{
+        width: bounds.width,
+        height: bounds.height,
+      }}
+    >
+      {/* Scrollable container. */}
+      <div className={mx('absolute top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2')}>
+        {children}
+        {items?.map((item, index) => (
+          <Tile item={item} key={index} layout={layout?.tiles[item.id] ?? { x: 0, y: 0 }} />
```
thure: I think the naming here should be adjusted before this gets too far along, since it helps to have globally unique names, and we already have a `Grid.tsx`.

This package uses a unique name already, `Board` — I think in parallel with Stack, this should be renamed to `BoardItem`. As with Stack, `BoardItem` can manage DnD operations, and then styling the DOM nodes can be handled by `Card`.

packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
...
 
+// TODO(burdon): Create variant that can be housed outside of provider?
 const Controls = ({ classNames }: ControlsProps) => {
-  const { readonly, zoom, controller } = useGridContext('Controls');
+  const { readonly, zoom, controller } = useGridContext(Controls.displayName);
 
   return (
-    <div className={mx('fixed top-2 left-2 z-10', classNames)}>
+    <div className={mx('fixed top-4 left-4 z-10', classNames)}>
       <Toolbar.Root>
```
thure: No need for the extra div, you can give `Toolbar.Root` the fixed position and the classnames, then you also won’t need `mx`.

packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
...
+Content.displayName = 'Grid.Content';
+
 //
 // Controls
 //
 
 type ControlsProps = ThemedClassName;
 
+// TODO(burdon): Create variant that can be housed outside of provider?
 const Controls = ({ classNames }: ControlsProps) => {
```
thure: If this will always render a `Toolbar`, this should be renamed to `{Qualifier}Toolbar`, as we have with `EditorToolbar`, `Card.Toolbar`, etc.

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
+      <Card.Content classNames='h-full'>
+        <Card.Toolbar>
+          {/* TODO(burdon): How to set disabled? */}
+          <Card.DragHandle toolbarItem />
+          {/* TODO(burdon): Heading has strange padding (makes the Toolbar too tall). */}
+          {/* <Card.Heading classNames='grow truncate'>{title}</Card.Heading> */}
+          <h1 className='grow truncate pli-1'>{title}</h1>
+          {dragState !== 'dragging' && (
+            <Card.ToolbarIconButton
+              // TODO(burdon): Should be the same size/padding as the DragHandle (and square by default).
```
thure: Best to act on these TODO’s right away since it’s stragithforward to do so and we should eliminate classname copypasta as an urgent priority.

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
-      <div>
-        <h1 className='p-4'>{item.id}</h1>
-      </div>
-    </div>
+      {/* TODO(burdon): Should the header (toolbar) be inside Content? If so, why is Content separate from Root? (e.g., rather than a Body). */}
+      <Card.Content classNames='h-full'>
+        <Card.Toolbar>
+          {/* TODO(burdon): How to set disabled? */}
+          <Card.DragHandle toolbarItem />
+          {/* TODO(burdon): Heading has strange padding (makes the Toolbar too tall). */}
```
thure: Too tall in what way exactly?

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
+      onClick={() => onSelect?.(item.id)}
     >
-      <div>
-        <h1 className='p-4'>{item.id}</h1>
-      </div>
-    </div>
+      {/* TODO(burdon): Should the header (toolbar) be inside Content? If so, why is Content separate from Root? (e.g., rather than a Body). */}
+      <Card.Content classNames='h-full'>
+        <Card.Toolbar>
+          {/* TODO(burdon): How to set disabled? */}
```
thure: If that flag is not implemented, it certainly can be.

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
+      // TODO(burdon): Root should have no padding by default (leave that to Content?)
+      classNames='absolute p-0'
+      style={getGridRect(grid, layout)}
+      onClick={() => onSelect?.(item.id)}
     >
-      <div>
-        <h1 className='p-4'>{item.id}</h1>
-      </div>
-    </div>
+      {/* TODO(burdon): Should the header (toolbar) be inside Content? If so, why is Content separate from Root? (e.g., rather than a Body). */}
```
thure: Because `Content` is what applies the encapsulation styles (border, border radius, background, clipping, etc etc), and because sometimes `Content` does not need to be rendered because the card will already be encapsulated by another component, e.g. `Popover.Content`.

Note the parallelism, `Popover.Content` applies encapsulation, as does `Card.Content`.

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
+  const title = item.id;
 
   return (
-    <div
-      className={mx('absolute flex flex-col bg-inputSurface border border-separator rounded-sm shadow', classNames)}
-      style={getGridRect(grid, { x, y, width, height })}
-      onClick={onClick}
+    <Card.Root
+      ref={ref}
+      // TODO(burdon): Root should have no padding by default (leave that to Content?)
```
thure: `Card.Root` applies padding in order to facilitate flush DnD drop zones while giving the illusion of consistent spacing between the cards.

You don’t _need_ to use `Card.Root` (unlike with many other components), you can skip it and just render `Card.Content`.

thure: `Card.Root` can be renamed to something else, certainly.

packages/ui/react-ui-board/src/Grid/Tile.tsx
```diff
@@ -2,36 +2,86 @@
...
-      </div>
-    </div>
+      {/* TODO(burdon): Should the header (toolbar) be inside Content? If so, why is Content separate from Root? (e.g., rather than a Body). */}
+      <Card.Content classNames='h-full'>
+        <Card.Toolbar>
+          {/* TODO(burdon): How to set disabled? */}
+          <Card.DragHandle toolbarItem />
+          {/* TODO(burdon): Heading has strange padding (makes the Toolbar too tall). */}
+          {/* <Card.Heading classNames='grow truncate'>{title}</Card.Heading> */}
+          <h1 className='grow truncate pli-1'>{title}</h1>
```
thure: If `Card.Heading` doesn’t do what you want, I’d still recommend using it and overriding its styles as you need, that way it doesn’t become disconnected, i.e. using an `h1` here will add more tech debt.

packages/ui/react-ui-board/src/Grid/Tile.tsx
thure: `Card.Heading` applies `cardSpacing` tokens — these tokens adapt to context, so it’s possible a parent has `density-coarse`.

packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
...
+        zoom && 'scale-50',
+        classNames,
+      )}
+      style={{
+        width: bounds.width,
+        height: bounds.height,
+      }}
+    >
+      {/* Scrollable container. */}
+      <div className={mx('absolute top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2')}>
```
thure: We may (eventually) need to style subcomponents like this — I’d recommend moving this into a new `Board.Viewport` component which can apply `classNames` as other subcomponents do, then it would be consistent with other components with viewports.

packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
...
+        {children}
+        {items?.map((item, index) => (
```
thure: Why does this support both `children` and a higher-order `items`? Usually if the consumer is able to render `children`, it should iterate over `items` on its own so it doesn’t have to reimplement the iterator if the DOM order should change.

packages/ui/react-ui-board/src/Grid/Grid.tsx
thure: Wait, though, why should this have `fixed`? This makes it compete with Deck. Only stories and layout plugins should apply `fixed`, other components shouldn’t do that because it isn’t portable.

packages/ui/react-ui-board/src/Grid/Grid.tsx
```diff
@@ -158,49 +173,65 @@ const RootInner = forwardRef<GridController, RootProps>(
             padding: margin,
```
thure: Is the `margin` prop _only_ for styling the padding on this particular element?

Some `Root` subcomponents do render DOM nodes, I’d recommend using a more ordinary pattern, like applying styles by default and letting a `classNames` prop apply overrides.

richburdon: Changed to `overscroll`

packages/ui/react-ui-board/src/Grid/Grid.tsx
richburdon: Got it -- i separated out Content and Viewport (very helpful).

richburdon:
Addressed 90% of @thure's feedback. Some items left for him:
- Iconbutton (square/size)
- Refactor Card.Root/Content (e.g., to StackItemConatiner).

packages/plugins/plugin-conductor/package.json
```diff
@@ -50,7 +50,8 @@
     "@dxos/react-ui-canvas-editor": "workspace:*",
     "@dxos/react-ui-stack": "workspace:*",
     "@preact-signals/safe-react": "^0.9.0",
-    "effect": "3.14.21"
+    "effect": "3.14.21",
+    "i18next": "^21.10.0"
```
thure: Why would plugins need `i18next`?

thure: If this is just for the type, it should be re-exported from `react-ui`.

packages/plugins/plugin-debug/package.json
```diff
@@ -74,6 +74,7 @@
     "@preact/signals-core": "^1.9.0",
     "@tldraw/tldraw": "^3.0.0",
     "date-fns": "^3.3.1",
+    "i18next": "^21.10.0",
```
thure: Ditto here, re-export the type you need from `react-ui`.

packages/plugins/plugin-automation/package.json
```diff
@@ -50,7 +50,8 @@
     "@dxos/react-ui-stack": "workspace:*",
     "@dxos/util": "workspace:*",
     "@preact-signals/safe-react": "^0.9.0",
-    "@preact/signals-core": "^1.9.0"
+    "@preact/signals-core": "^1.9.0",
+    "i18next": "^21.10.0"
```
thure: Ditto here, re-export the type you need from `react-ui`.

packages/plugins/plugin-table/src/translations.ts
```diff
@@ -2,12 +2,14 @@
...
 
+import { type Resource } from 'i18next';
+
 import { Type } from '@dxos/echo';
 import { TableType } from '@dxos/react-ui-table';
 
 import { TABLE_PLUGIN } from './meta';
 
-export default [
+export const translations: Resource[] = [
```
thure: Change to `satisfies Resource[]`, import `Resource` from `react-ui`.

thure: Please move all of these into a separate PR if possible, this is making Board much more difficult to review.

packages/ui/lit-grid/package.json
```diff
@@ -36,6 +36,7 @@
   ],
   "dependencies": {
     "@atlaskit/pragmatic-drag-and-drop": "^1.4.0",
+    "i18next": "^21.10.0",
```
thure: This *definitely* should not have this dependency.

packages/ui/lit-theme-editor/package.json
```diff
@@ -29,6 +29,7 @@
     "@dxos/log": "workspace:*",
     "@dxos/react-hooks": "workspace:*",
     "@dxos/react-ui-theme": "workspace:*",
+    "i18next": "^21.10.0",
```
thure:
```suggestion
```

packages/ui/lit-grid/package.json
thure:
```suggestion
```

packages/ui/lit-ui/package.json
```diff
@@ -40,6 +40,7 @@
     "@dxos/react-hooks": "workspace:*",
     "@dxos/util": "workspace:*",
     "@lit/react": "^1.0.5",
+    "i18next": "^21.10.0",
```
thure:
```suggestion
```

packages/ui/primitives/react-hooks/package.json
```diff
@@ -27,7 +27,8 @@
     "@dxos/async": "workspace:*",
     "@dxos/log": "workspace:*",
     "@preact-signals/safe-react": "^0.9.0",
-    "alea": "^1.0.1"
+    "alea": "^1.0.1",
+    "i18next": "^21.10.0"
```
thure:
```suggestion
```

packages/ui/primitives/react-input/package.json
```diff
@@ -28,6 +28,7 @@
     "@radix-ui/react-primitive": "2.0.2",
     "@radix-ui/react-slot": "1.1.2",
     "@radix-ui/react-use-controllable-state": "1.1.0",
+    "i18next": "^21.10.0",
```
thure:
```suggestion
```

packages/ui/primitives/react-list/package.json
```diff
@@ -31,6 +31,7 @@
     "@radix-ui/react-primitive": "2.0.2",
     "@radix-ui/react-slot": "1.1.2",
     "@radix-ui/react-use-controllable-state": "1.1.0",
+    "i18next": "^21.10.0",
```
thure:
```suggestion
```

thure (dismissed):
Please move changes related to translations into a separate PR, re-exporting `Resource` from `react-ui` or some other package instead of adding it directly as a dependency to all packages that supply translations.

packages/plugins/plugin-conductor/package.json
richburdon: done

packages/plugins/plugin-table/src/translations.ts
richburdon: done throughout

richburdon:
> Please move changes related to translations into a separate PR, re-exporting `Resource` from `react-ui` or some other package instead of adding it directly as a dependency to all packages that supply translations.

thanks for catching all of this. i've resolved everything.

thure:
@richburdon Have you pushed your changes? I’m not seeing any since much earlier today…

packages/core/functions/package.json
```diff
@@ -75,6 +75,7 @@
     "esbuild-wasm": "^0.16.14",
     "express": "^4.19.2",
     "get-port-please": "^3.1.1",
+    "i18next": "^24.2.1",
```
thure, line 78: Missed one…

PR #9475, opened 2025-07-07
feat: Add Effect retries for functions htttp module loading

mykola-vrmchk:


packages/core/functions/src/bundler/bundler.ts
```diff
@@ -263,8 +266,39 @@ const httpPlugin: Plugin = {
     // When a URL is loaded, we want to actually download the content from the internet.
     // This has just enough logic to be able to handle the example import from unpkg.com but in reality this would probably need to be more complex.
     build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
-      const response = await fetch(args.path);
-      return { contents: await response.text(), loader: 'jsx' };
+      return retryExponentially(new Context(), async () => {
```
dmaretskyi: Context.default()

dmaretskyi (changes requested):

packages/core/functions/src/bundler/bundler.ts
```diff
@@ -263,8 +266,39 @@ const httpPlugin: Plugin = {
...
+        const response = await fetch(args.path);
+        if (response.ok) {
+          return { contents: await response.text(), loader: 'jsx' };
+        } else {
```
dmaretskyi: use effect for retries

dmaretskyi (dismissed):

packages/core/functions/src/bundler/bundler.ts
```diff
@@ -263,8 +264,50 @@ const httpPlugin: Plugin = {
     // When a URL is loaded, we want to actually download the content from the internet.
     // This has just enough logic to be able to handle the example import from unpkg.com but in reality this would probably need to be more complex.
     build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
-      const response = await fetch(args.path);
-      return { contents: await response.text(), loader: 'jsx' };
+      const a = await Effect.runPromise(
```
dmaretskyi:
```ts
onLoad(
  { filter: /.*/, namespace: "http-url" },
  (args) =>
    Effect.gen(
      function*() {
        const response = yield* HttpClient.get(args.path)
        if (response.status !== 200) {
          throw new Error(`failed to fetch: ${response.status}`)
        }

        const text = yield* response.text
        log.info("fetch complete", { text })
        return { contents: text, loader: "jsx" }
      }
    ).pipe(
      Effect.retry(
        pipe(
          Schedule.exponential(Duration.millis(1_000)),
          Schedule.jittered,
          Schedule.intersect(Schedule.recurs(4))
        )
      ),
      Effect.provide(FetchHttpClient.layer),
      Effect.runPromise // runAndForwardErrors
    )
)
```

packages/core/functions/src/bundler/bundler.ts
```diff
@@ -263,8 +265,28 @@ const httpPlugin: Plugin = {
     // When a URL is loaded, we want to actually download the content from the internet.
     // This has just enough logic to be able to handle the example import from unpkg.com but in reality this would probably need to be more complex.
     build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
-      const response = await fetch(args.path);
-      return { contents: await response.text(), loader: 'jsx' };
+      const maxRetries = 5;
```
dmaretskyi: constants at module top-level

good -- extracting them

dmaretskyi (dismissed):

PR #9486, opened 2025-07-08
refactor: rename ViewType to Projection

wittjosiah:
Part of DX-196

Followup will handle adjusting the view types themselves

### DIFF
--- a/packages/core/assistant/src/research/graph.ts
+++ b/packages/core/assistant/src/research/graph.ts
@@ -282,15 +282,21 @@ const SoftRef = Schema.Struct({
 const preprocessSchema = (schema: Schema.Schema.AnyNoContext) => {
   const isRelationSchema = getEntityKind(schema) === 'relation';
 
-  const go = (ast: SchemaAST.AST): SchemaAST.AST => {
+  const go = (ast: SchemaAST.AST, visited = new Set<SchemaAST.AST>()): SchemaAST.AST => {
+    if (visited.has(ast)) {
+      // Already visited this node, prevent infinite recursion.
+      return ast;
+    }
+    visited.add(ast);
+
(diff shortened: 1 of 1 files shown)


packages/sdk/schema/src/projection/projection-manager.ts
```diff
@@ -141,7 +141,7 @@ export class ViewProjection {
    */
   tryGetFieldProjection(fieldId: string): FieldProjection | undefined {
     invariant(this._schema.properties);
-    const field = this._view.fields.find((field) => field.id === fieldId);
+    const field = this._projection.fields.find((field) => field.id === fieldId);
```
richburdon, line 144: i don't think it makes sense to have both `get` and `try`. just make get return undefined (i realize this isn't being introduced this PR).

packages/sdk/schema/src/projection/projection-manager.ts
```diff
@@ -172,15 +172,15 @@ export class ViewProjection {
    */
   hideFieldProjection(fieldId: string): void {
     untracked(() => {
-      const index = this._view.fields.findIndex((field) => field.id === fieldId);
+      const index = this._projection.fields.findIndex((field) => field.id === fieldId);
```
richburdon, line 175: a very strange and inconsistent method name => setHiddenProperty

packages/sdk/schema/src/projection/projection-manager.ts
```diff
@@ -35,28 +35,28 @@ export type FieldProjection = {
 };
 
 /**
- * Wrapper for View that manages Field and Format updates.
+ * Wrapper for Projection that manages Field and Format updates.
  */
-export class ViewProjection {
+export class ProjectionManager {
```
richburdon, line 40: not a great name; manager is usually for collections?
ProjectionWrapper?

PR #9487, opened 2025-07-08
chore: Minor clean-up of translations to make consistent

richburdon:
- Basic standardization of translations
- NEXT: Figure out how to configure `eslint-plugin-i18next`
- Would require a build step to transform `translations.ts` into a JSON document.


thure (approved):
Looks fine to me — please also check for i18next dependencies that were not cleaned-up from your previous PR, there was at least one.

PR #9491, opened 2025-07-09
chore: Thread refresh

richburdon:


packages/plugins/plugin-thread/src/components/CommentsContainer.tsx
```diff
@@ -58,9 +59,13 @@ export const CommentsContainer = ({ anchors, currentId, showResolvedThreads, ...
     );
   }
 
-  return filteredAnchors.map((anchor) => {
-    const thread = Relation.getSource(anchor) as ThreadType;
-    const threadId = fullyQualifiedId(thread);
-    return <CommentContainer key={threadId} anchor={anchor} current={currentId === threadId} {...props} />;
-  });
+  return (
+    <div className='divide-y divide-subduedSeparator'>
```
thure: I don't recommend using `divide` for complex components like this, it depends a lot on knowledge of direct DOM node descendants which isn't reliable at this scope (this has caused issues in the past).

richburdon: it was strange to me that this Container didn't have a containin element; and it seemed appropriate (in the same level) for the container that maps the children to decide on the conditions between them (spacing, border, etc.) but point taken and i've moved it downstream.

PR #9495, opened 2025-07-10
moonrepo trial

wittjosiah:


tools/esbuild/src/cli.ts
```diff
@@ -0,0 +1,64 @@
...
+//
+// Copyright 2025 DXOS.org
+//
+
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+
+import main, { type EsbuildExecutorOptions } from './main';
+
+void (async () => {
```
dmaretskyi, line 12: I think our node version should support top-level await?

I can take a look at this later

thure (approved):
looks great, looking forward to this! ✨

.dockerignore
```diff
@@ -1,141 +0,0 @@
-# Logs
```
dmaretskyi, line 1: Is this not used when building our CI image?

.ncurc.js
```diff
@@ -1,4 +0,0 @@
-module.exports = {
```
dmaretskyi: afaik this is for npm-check-updates tool that im using regularly

Dockerfile
```diff
@@ -1,69 +0,0 @@
-FROM node:18-slim AS monorepo
```
dmaretskyi, line 1: is this not used to build our CI image?

.ncurc.js
wittjosiah: will restore

tools/esbuild/src/cli.ts
wittjosiah: cursor wrote this, it could probably use some cleanup

Dockerfile
wittjosiah: no was for something else

PR #9497, opened 2025-07-10
fix: JSON serialization of ECHO proxy objets

dmaretskyi:


packages/core/echo/echo/src/Type.ts
```diff
@@ -238,6 +238,7 @@ export const getMeta = (schema: Obj.Any | Relation.Any): Meta | undefined => {
   return EchoSchema.getTypeAnnotation(schema);
 };
 
+// TODO(dmaretskyi): Remove re-exports.
```
wittjosiah, line 241: `Keys` namespace perhaps?

wittjosiah (approved):
excellent

PR #9504, opened 2025-07-11
feat: Transition towards effect AI

dmaretskyi:
Part of DX-191

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -30809,6 +30809,7 @@ packages:
   source-map@0.8.0-beta.0:
     resolution: {integrity: sha512-2ymg6oRBpebeZi9UUNsgQ89bhx01TcTkmNTGnNO88imTmbSgy4nfujrgVEFKWpMTEGA11EDkTt7mqObTPdigIA==}
     engines: {node: '>= 8'}
+    deprecated: The work that was done in this beta branch won't be included in future versions
 
   sourcemap-codec@1.4.8:
     resolution: {integrity: sha512-9NykojV5Uih4lgo5So5dtw+f0JgJX30KCNI8gwhz2J9A15wD0Ml6tjHKwf6fTSa6fAdVBdZeNOs9eJ71qCk8vA==}


packages/sdk/schema/src/common/message.ts
```diff
@@ -10,107 +10,268 @@ import { GeneratorAnnotation, ObjectId, TypedObject } from '@dxos/echo-schema';
...
+    ...Base.fields,
+  }).pipe(Schema.mutable);
+  export interface Select extends Schema.Schema.Type<typeof Select> {}
+
+  /**
+   * Associates artifact (of a specific version) with this conversation.
+   * Used to track associated artifacts as well their changes during the conversation.
+   */
+  // TODO(dmaretskyi): What's the relation of this to the reference content block?
+  const ArtifactPin = Schema.TaggedStruct('artifactPin', {
```
richburdon: pinArtifact or bind/set/attachAritfact

packages/sdk/schema/src/common/message.ts
```diff
@@ -10,107 +10,268 @@ import { GeneratorAnnotation, ObjectId, TypedObject } from '@dxos/echo-schema';
...
+  // TODO(dmaretskyi): Consider handling this via a tool call.
+  export const Proposal = Schema.TaggedStruct('proposal', {
+    text: Schema.String,
+
+    ...Base.fields,
+  }).pipe(Schema.mutable);
+  export interface Proposal extends Schema.Schema.Type<typeof Proposal> {}
+
+  /**
+   * Model priniting info about the list of available tools.
```
richburdon: "printing" (sp.)

richburdon: Also, can we use anything from the AiToolkit defs here?

richburdon (approved):
nice and clean

packages/sdk/schema/src/common/message.ts
dmaretskyi: I was considering re-using their message-part types, but since this is part of persisted data (which we'd need to migrate) I think maintaining our own definition is beneficial

PR #9506, opened 2025-07-14
feat: Blueprints

dmaretskyi:


packages/core/artifact-testing/src/blueprints.ts
```diff
@@ -8,6 +8,7 @@ import { Obj } from '@dxos/echo';
 import { readDocument, writeDocument } from './tools';
 
 export const DESIGN_SPEC_BLUEPRINT = Obj.make(Blueprint, {
+  blueprintId: 'dxos.org/blueprint/design-spec',
```
dmaretskyi: name, identifier, key, tag, registryId, qualifiedName, canonical --- lets come-up with a standard name for this

packages/core/assistant/src/blueprint/blueprint.ts
```diff
@@ -6,12 +6,17 @@ import { Schema } from 'effect';
...
+import { Format, LabelAnnotation } from '@dxos/echo-schema';
 
 /**
  * Blueprint schema defines the structure for AI assistant blueprints.
  * Blueprints contain instructions, tools, and artifacts that guide the AI's behavior.
  */
 export const Blueprint = Schema.Struct({
+  // TODO(burdon): Move to main API.
+  // TODO(burdon): It's cumbersome not to call this `id` but would clash with runtime property.
+  blueprintId: Format.DXN,
```
dmaretskyi: The names in example blueprints are not DXNs: "dxos.org/blueprint/design-spec"

packages/core/artifact-testing/src/blueprints.ts
richburdon: i think either `key`, `name`, or `gid` or `uid`.

packages/core/assistant/src/blueprint/blueprint.ts
richburdon: isn't that a DXN? what do we call that? it's the same as types no?

dmaretskyi: DXN would be `dxn:blueprint:dxos.org/blueprint/design-spec`

this is just the typename portion

PR #9535, opened 2025-07-21
refactor: board uses create dialog

wittjosiah:


packages/plugins/plugin-space/src/capabilities/intent-resolver.ts
```diff
@@ -324,6 +324,7 @@ export default ({ context, observability, createInvitationUrl }: IntentResolverO
                 props: {
                   target,
                   typename,
+                  onCreateObject,
```
richburdon, line 327: ok interesting -- this was what i was missing.
would prefer all `on` callbacks to be defined at the end.

packages/plugins/plugin-space/src/types/types.ts
```diff
@@ -222,6 +222,8 @@ export namespace SpaceAction {
       target: Schema.Union(SpaceSchema, DataType.Collection),
       typename: Schema.optional(Schema.String),
       navigable: Schema.optional(Schema.Boolean),
+      // TODO(wittjosiah): This is a function, is there a better way to handle this?
+      onCreateObject: Schema.optional(Schema.Any),
```
richburdon, line 226: Schema.Unknown is possibly a bit more correct, but i don't have a better idea.

PR #9542, opened 2025-07-22
fix: Card system increment

thure:
This PR:
- begins a resolution of [DX-239](https://linear.app/dxos/issue/DX-239)
- omits `role` from the props of `StackItem.Content` to avoid confusion
- renames:
  - `card--kanban` to `card--intrinsic`
  - `card--board` (newly introduced) to `card--extrinsic`
  - `card--document` to `transclusion`
  - all subcomponents of `Board` to be prefixed with `Board` as is best practice to aid in global searches
- fixes KanbanContainer story
- adds BoardContainer story
- fixes an issue related to `StackItem.Content`’s toolbar styling functionality
- aligns Board subcomponents better with Kanban’s.

https://github.com/user-attachments/assets/c7a43bfd-0716-4f73-b663-ac379b90a44f


packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  return { board };
+};
+
+//
+// Story components
+//
+
+const rollOrg = () =>
```
richburdon, line 54: pls try to reuse sdk/schema/testing generators.
see generator.test.ts

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+    image: faker.image.url(),
+    website: faker.internet.url(),
+    status: faker.helpers.arrayElement(DataType.OrganizationStatusOptions).id,
+    // TODO(thure): Why is this so difficult to type?
+  }) as unknown as DataType.Organization;
+
+const StorybookBoard = () => {
+  const _client = useClient();
+  const spaces = useSpaces();
+  const space = spaces[spaces.length - 1];
```
richburdon, line 67: standard here for storybooks is to just use useSpace

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const boards = useQuery(space, Filter.type(Board.Board));
```
richburdon, line 68: const [board] = useQuery(...)

packages/plugins/plugin-board/src/components/BoardContainer.tsx
```diff
@@ -5,20 +5,20 @@
...
+import { type Obj, Ref, type Type } from '@dxos/echo';
 import { invariant } from '@dxos/invariant';
 import { SpaceAction } from '@dxos/plugin-space/types';
-import { Board as BoardComponent, type BoardController, type BoardRootProps } from '@dxos/react-ui-board';
+import { Board, type BoardController, type BoardRootProps } from '@dxos/react-ui-board';
 import { StackItem } from '@dxos/react-ui-stack';
 import { isNonNullable } from '@dxos/util';
 
-import { type Board } from '../types';
+import { type Board as BoardType } from '../types';
```
richburdon, line 18: why invert this? feels that the type should be dominant since it is used everywhere, whereas the component is "local" and should be renamed.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
```diff
@@ -82,22 +82,22 @@ export const BoardContainer = ({ role, board }: BoardContainerProps) => {
...
-      <StackItem.Content role={role} toolbar>
-        <BoardComponent.Controls />
-        <BoardComponent.Container>
-          <BoardComponent.Viewport classNames='border-none'>
-            <BoardComponent.Background />
-            <BoardComponent.Content items={items} getTitle={(item) => Obj.getLabel(item) ?? item.id} />
-          </BoardComponent.Viewport>
-        </BoardComponent.Container>
+    <Board.Root ref={controller} layout={board.layout} onAdd={handleAdd} onDelete={handleDelete} onMove={handleMove}>
+      <StackItem.Content toolbar>
```
richburdon, line 87: is "StackItem" the relevant cross-cutting container for all planks main content? i'm very glad to have a common container (that manages the toolbar), but for most planks "stack" doesn't seem like the right concept.

packages/plugins/plugin-outliner/src/components/JournalContainer.tsx
```diff
@@ -16,7 +16,7 @@ export type JournalContainerProps = {
 
 export const JournalContainer = ({ role, journal }: JournalContainerProps) => {
```
richburdon: remove role?
same comment across all containers.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { translations } from '../translations';
+import { Board } from '../types';
+
+faker.seed(0);
+
+//
+// Initialization utilities
+//
+
+const initializeBoard = async ({ space, client }: { space: Space; client: Client }) => {
```
richburdon, line 36: i think better to inline this since it isn't re-used

packages/ui/react-ui-board/src/components/Board/Board.tsx
```diff
@@ -368,30 +361,30 @@ const Controls = ({ classNames }: ControlsProps) => {
...
-  Viewport,
-  Content,
-  Background,
-  Controls,
-  Cell,
+  Root: BoardRoot,
+  Container: BoardContainer,
+  Viewport: BoardViewport,
+  Content: BoardContent,
+  Backdrop: BoardBackdropProps,
```
richburdon: Props?

thure: Must be a typo, thank you.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: I find that inverse more confusing personally but can revert this if you prefer. I seem to recall plenty of type names ending in `Type`, but maybe that has been refactored? `KanbanType` and `KanbanSchema` are both still present and helpfully disambiguate the term 'Kanban' when used in React files.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
thure: I think we can optimize this later, currently BoardContainer.story is written to be as similar to KanbanContainer.story, so refactors to one should apply to the other as well, and I’d like scope not to creep for this one PR.

I’ll leave a TODO.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: The Stack component in `react-ui-stack` has underpinned both StackPlugin and DeckPlugin for quite a long time as their principal upstream component. Whether it should have a different name can be a matter for discussion but, I think it’s not relevant to this PR.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
thure: If this is used in KanbanContainer.story, I will put off optimizations to these two stories as a follow-up to avoid scope-creep.

packages/plugins/plugin-outliner/src/components/JournalContainer.tsx
thure: We should use an ESLint rule for this.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
thure: If this is a lexical style we should enforce, we should find an ESLint rule for it.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
thure: This causes the story not to load.

packages/plugins/plugin-board/src/components/BoardContainer.stories.tsx
thure: Strangely, if I use `const space = useSpace()`, the story doesn’t load.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: Why is Board unlike Kanban in this way? In Kanban, we have `KanbanSchema` and `KanbanType`, which means we don’t need to rename things when using the `Kanban` component namespace.

If Board should set the best practice, Kanban should be refactored to apply it.

thure: Left a TODO on the matter for now.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: Why is Board unlike Kanban in this way? In Kanban, we have KanbanSchema and KanbanType, which means we don’t need to rename things when using the Kanban component namespace.

If Board should set the best practice, Kanban should be refactored to apply it.

Left a TODO on the matter for now.

PR #9556, opened 2025-07-28
fix: re-connect related orgs/contacts

wittjosiah:
Resolves DX-260

https://github.com/user-attachments/assets/ef800859-67b8-43e9-b783-9f2dea7a9691


packages/plugins/plugin-inbox/src/capabilities/react-surface.tsx
```diff
@@ -127,38 +130,54 @@ export default () =>
...
-            if (table) {
-              return dispatch(
-                createIntent(LayoutAction.Open, {
-                  part: 'main',
-                  subject: [fullyQualifiedId(table)],
-                  options: { workspace: space?.id },
+          (contact: DataType.Person) =>
+            Effect.gen(function* () {
+              const view = currentSpaceContacts.includes(contact) ? currentSpaceContactTable : defaultSpaceContactTable;
+              yield* dispatch(
```
thure, line 149: Ah, is this the new pattern we should use for this?

thure (approved):
lgtm!

PR #9567, opened 2025-07-29
refactor: update L1Panel header with improved back button and toolbar layout

richburdon:
Update back button.

NOTE: @wittjosiah @thure I can't figure out what `NavTreeItemColumns` is for.

https://github.com/user-attachments/assets/c0848b0e-45b0-40bd-911c-60d6ec4158ae


thure:
I thought we had planned to move to a stateless L1, which would host collections, records, and settings, schemas, etc, all in the same tree. In what way does this improve the design, exactly?

thure (changes requested):
I thought we had planned to move to a stateless L1, which would host collections, records, and settings, schemas, etc, all in the same tree.

In what way would you argue this improves the design? Is the improvement significant enough to warrant a PR?

richburdon:
> I thought we had planned to move to a stateless L1, which would host collections, records, and settings, schemas, etc, all in the same tree, eliminating the need for this pattern which has known discoverability & learnability issues.
> 
> Can we postpone minor iterations till we can resolve the bigger usability issues here, or would you be amenable to implementing that in this PR?

you're right, but this resolves a current pain point (back is not discoverable) and the other work is currently unscheduled so i want to land this.

richburdon:
@thure @wittjosiah do you have a comment on the NavTreeItemColumns?

PR #9575, opened 2025-07-30
refactor: Reorganize chat services

dmaretskyi:


packages/plugins/plugin-assistant/src/capability-definitions.ts
wittjosiah, line 1: I'd call this file "capabilities.ts" to align with the other plugins

packages/plugins/plugin-assistant/src/AssistantPlugin.tsx
```diff
@@ -95,4 +99,12 @@ export const AssistantPlugin = () =>
       activatesBefore: [Events.SetupArtifactDefinition],
       activate: ReactSurface,
     }),
+    defineModule({
+      id: `${meta.id}/module/ai-service`,
+      activatesOn: Events.Startup,
```
wittjosiah, line 104: does this cause Composer to completely re-render when turning the plugin on/off? Generally I've been trying to avoid using this event unless strictly necessary, particularly for user-toggleable plugins

packages/plugins/plugin-assistant/src/hooks/useChatServices.ts
```diff
@@ -43,49 +43,25 @@ export type UseChatServicesProps = {
  * Construct service layer.
  */
 export const useChatServices = ({ space }: UseChatServicesProps): Layer.Layer<ChatServices> | undefined => {
-  const toolRegistry = useToolRegistry();
-  // TODO(dmaretskyi): We can provide the plugin registry as a layer and then build the entire layer stack from there. We need to think how plugin reactivity affect our layer structure.
-  const toolResolver = useToolResolver();
-  const toolExecutionService = useToolExecutionService();
+  const aiServiceLayer =
+    useCapabilities(AssistantCapabilities.AiServiceLayer).at(0) ?? Layer.die('AiService not found');
```
wittjosiah, line 47: is this to handle the capability being missing than it throwing here? if so, TODO we should factor out a util for this

packages/plugins/plugin-assistant/src/AssistantPlugin.tsx
dmaretskyi: What's a better event for this?

packages/plugins/plugin-assistant/src/hooks/useChatServices.ts
dmaretskyi: This will make it throw inside chat-processor. Agree about factoring out, but I'm holding off on this until a pattern emerges.

packages/plugins/plugin-assistant/src/capability-definitions.ts
dmaretskyi: There's a folder called capabilities already. Happy to restructure, just give me the outline

PR #9579, opened 2025-07-30
feat: basic blueprint support in composer

wittjosiah:
- add blueprints to object form
- include blueprint smart collection in default space content
- flatten template object into blueprint object

also:
- include default smart collections in personal space on identity creation
- prevent records smart collection from being removed
- only allow tables/kanbans of configured schema to be created

Part of DX-279


packages/core/assistant-testing/src/blueprints/blueprints.ts
```diff
@@ -13,9 +14,10 @@ export const DESIGN_BLUEPRINT = Obj.make(Blueprint.Blueprint, {
   key: 'dxos.org/blueprint/design-spec',
   name: 'Design Spec',
   description: 'Preserve the conversation in the design spec.',
-  instructions: Ref.make(
-    Template.make({
-      source: trim`
+  instructions: {
+    source: Ref.make(
+      Obj.make(DataType.Text, {
+        content: trim`
```
richburdon, line 20: indent content +2

packages/core/assistant-testing/src/blueprints/blueprints.ts
```diff
@@ -13,9 +14,10 @@ export const DESIGN_BLUEPRINT = Obj.make(Blueprint.Blueprint, {
   key: 'dxos.org/blueprint/design-spec',
   name: 'Design Spec',
   description: 'Preserve the conversation in the design spec.',
-  instructions: Ref.make(
-    Template.make({
-      source: trim`
+  instructions: {
+    source: Ref.make(
+      Obj.make(DataType.Text, {
```
richburdon, line 19: FYI, my current PR will provide `DataType.text({ name, content })` as a util for this.

PR #9582, opened 2025-07-30
refactor: rename DocumentType to Document.Document across codebase

richburdon:


packages/plugins/plugin-markdown/src/capabilities/react-surface.tsx
```diff
@@ -74,7 +74,8 @@ export default () =>
     createSurface({
       id: `${MARKDOWN_PLUGIN}/editor`,
       role: ['article', 'section'],
-      filter: (data): data is { subject: { id: string; text: string } } => isEditorModel(data.subject),
```
wittjosiah, line 77: this should not be removed

packages/sdk/schema/src/common/index.ts
```diff
@@ -126,6 +127,7 @@ export namespace DataType {
 
   export const Text = Text$.Text;
   export type Text = Text$.Text;
+  export const text = (content = '') => Obj.make(DataType.Text, { content });
```
wittjosiah: makeText? would be nice if it followed the same pattern everywhere

richburdon: yeah, experimenting with this pattern.

packages/plugins/plugin-markdown/src/capabilities/react-surface.tsx
richburdon: sorry what do you mean -- I replaced it with the Effect type check -- is that not ok?

wittjosiah: they're not the same thing, this is handling the case where it is *not* a document

PR #9584, opened 2025-07-31
build: Set typescript target to ESNext

dmaretskyi:


packages/plugins/plugin-native/package.json
```diff
@@ -25,6 +25,8 @@
   "dependencies": {
     "@dxos/app-framework": "workspace:*",
     "@dxos/log": "workspace:*",
+    "@dxos/plugin-help": "workspace:*",
+    "@dxos/plugin-navtree": "workspace:*",
```
wittjosiah: these aren't needed anymore

dmaretskyi: not sure how they got here, prob a bad merge

dmaretskyi:
```suggestion
```

PR #9589, opened 2025-07-31
fix: Change ‘schema’ to ‘record’ in user-facing strings

thure:
This PR:
- does what it says on the tin
- fixes a callout instance’s spacing

Screenshot 2025-08-01 at 14 35 16
Screenshot 2025-08-01 at 14 34 07


packages/plugins/plugin-debug/src/translations.ts
```diff
@@ -50,7 +50,7 @@ export const translations = [
         'space label': 'Space',
         'feeds label': 'Feeds',
         'objects label': 'Objects',
-        'schema label': 'Schema',
+        'schema label': 'Record',
```
wittjosiah: imo debug plugin should stay saying schema for now, just as it says "objects"

wittjosiah (commented):
So one thing that I found confusing is that I feel like the objects themselves are the "records" and the schemas are actually the "type of the record". I actually have the fallback name for dynamic schema as "record type" I think.

The places I think that maybe doesn't hold is in the navtree, I think it makes sense just have the settings page be called "Records" to match the smart collection name, but then it could say "Manage record types"

packages/plugins/plugin-debug/src/translations.ts
thure: I see, fair enough

PR #9592, opened 2025-08-01
refactor: reorganize assistant stories with new testing utilities and components

richburdon:


tools/storybook/vitest.config.ts
```diff
@@ -2,13 +2,27 @@
 // Copyright 2025 DXOS.org
 //
 
+import { join } from 'node:path';
 import { defineConfig, mergeConfig } from 'vitest/config';
+import { storybookTest } from '@storybook/addon-vitest/vitest-plugin';
 
-import { baseConfig } from '../../vitest.base.config';
+import { baseConfig } from '../../vitest.storybook.config';
```
dmaretskyi, line 10: since this is only imported here, can we inline `vitest.storybook.config` and throw out the parts that are not needed?

richburdon: yes, it's not working yet, so i left it here -- i want to factor out the common areas.

PR #9605, opened 2025-08-03
feat: remove deprecated files from dxos/ai

richburdon:


packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -117,7 +145,15 @@ const MapCanvas = forwardRef<MapController, MapCanvasProps>(({ markers, center,
   }, [markers]);
 
   return (
-    <div ref={ref} className='flex w-full h-full overflow-hidden bg-baseSurface'>
+    <div ref={ref} role='none' className='grid inset-0 overflow-hidden bg-baseSurface'>
+      {/* Focus ring. */}
+      <div
+        className={mx(
+          'z-[9999] absolute inset-0 border border-transparent pointer-events-none',
```
thure: Highly recommend just using one of the `dx-focus-ring` classes rather than reinventing this wheel.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -117,7 +145,15 @@ const MapCanvas = forwardRef<MapController, MapCanvasProps>(({ markers, center,
...
+          hasFocus && 'border-primary-500 ',
```
thure: Focus has a semantic tokens, `--dx-neutralFocusIndicator` and `--dx-accentFocusIndicator`, by default elements apply the neutral one. We also have a token for the focus ring's width. Avoid defining something new and special for focus wherever possible.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -84,28 +94,46 @@ const MapCanvas = forwardRef<MapController, MapCanvasProps>(({ markers, center,
...
-      onChange?.({ center: map.getCenter(), zoom: map.getZoom() });
+      onChange?.({
+        center: map.getCenter(),
+        zoom: map.getZoom(),
+      });
     }, 100);
+
     map.on('move', handler);
     map.on('zoom', handler);
+    map.on('focus', () => setHasFocus(true));
```
thure: This is a risky pattern – best practice is to handle this using attention to determine what to do with wheel and pointer events, which I recommend doing here. react-ui-grid is an example.

thure (changes requested):
Highly recommend using more ordinary patterns for focus indication and interactivity. See react-ui-grid for examples of the latter. The `dx-focus-ring` utility classes handle the former.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
richburdon: @thure given the TZ difference, please suggest how to do that -- i struggled to get it working and it's not documented or used in a story that enabled me to figure out how to use it.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
richburdon: +1

packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -117,7 +145,15 @@ const MapCanvas = forwardRef<MapController, MapCanvasProps>(({ markers, center,
...
   return (
-    <div ref={ref} className='flex w-full h-full overflow-hidden bg-baseSurface'>
+    <div ref={ref} role='none' className='grid inset-0 overflow-hidden bg-baseSurface'>
+      {/* Focus ring. */}
+      <div
+        className={mx(
+          'z-[9999] absolute inset-0 border border-transparent pointer-events-none',
+          hasFocus && 'border-primary-500 ',
+        )}
+      />
```
thure:
```suggestion
```
To use `dx-fous-ring`, first delete this extra element.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -84,28 +94,46 @@ const MapCanvas = forwardRef<MapController, MapCanvasProps>(({ markers, center,
...
 
+  // Support zoom if focused.
+  const [hasFocus, setHasFocus] = useState(false);
+  useEffect(() => {
+    if (hasFocus) {
+      map.scrollWheelZoom.enable();
+    } else {
+      map.scrollWheelZoom.disable();
+    }
+  }, [map, hasFocus]);
```
thure: Then use the attention context to gate map’s wheel events, use that state from the context based on the same fully qualified id provided to `useAttentionAttributes`:

```suggestion
  const {hasAttention} = useAttention(props.id);
  useEffect(() => {
    if (hasAttention) {
      map.scrollWheelZoom.enable();
    } else {
      map.scrollWheelZoom.disable();
    }
  }, [map, hasAttention]);
```

packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -34,15 +33,26 @@ const defaults = {
...
+  classNames,
+  scrollWheelZoom = true,
+  doubleClickZoom = true,
+  touchZoom = true,
+  center = defaults.center,
+  zoom = defaults.zoom,
+  ...props
+}: MapRootProps) => {
   return (
     <MapContainer
```
thure: In order to register this element with the attention system, apply the attention attributes to it with the fully qualified id:

```suggestion
  const attentionAttrs = useAttentionAttributes(props.id);
return (
<MapContainer {...attentionAttrs}
```

packages/ui/react-ui-geo/src/components/Map/Map.tsx
thure: Next, apply `dx-focus-ring` or `dx-focus-ring-inset` to the element that is focusable. Is that `TileLayer`?

packages/ui/react-ui-geo/src/components/Map/Map.tsx
thure: Ensure the story for this includes the attention plugin, or add an `ignoreAttention` property to the container which, when set, will always cause `map.scrollWheelZoom` to be enabled.

richburdon: this is low level react-ui -- so we can't use attention here unless we move it downstream.

packages/ui/react-ui-geo/src/components/Map/Map.tsx
richburdon: ok. the MapRoot is the focusable div.

PR #9606, opened 2025-08-04
fix: action + label reactivity in navtree

wittjosiah:
also includes fixes to e2e tests to account for additional initial object

Resolves DX-314

### DIFF
--- a/packages/apps/composer-app/src/playwright/constants.ts
+++ b/packages/apps/composer-app/src/playwright/constants.ts
@@ -1,0 +1,9 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+// - Records smart collection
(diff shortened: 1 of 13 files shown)


packages/apps/composer-app/src/playwright/constants.ts
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+// - Records smart collection
+// - Channels smart collection
+// - Assistant smart collection
+// - Blueprint smart collection
+export const INITIAL_OBJECT_COUNT = 4;
```
thure, line 9: Ahaaa, I see now. Makes sense.

PR #9607, opened 2025-08-04
refactor: consolidate geo types and refactor leaflet map components; lock scroll if not active

richburdon:


packages/ui/react-ui-geo/src/components/Map/Map.tsx
```diff
@@ -2,109 +2,223 @@
...
-        map.setView(center, zoom);
-      },
-      setZoom: (cb) => {
-        map.setZoom(cb(map.getZoom()));
-      },
-    }),
-    [map],
-  );
+  ) => {
+    const [attention, setAttention] = useState(false);
```
thure, line 76: @richburdon this is not what I meant when I left [this comment](https://github.com/dxos/dxos/pull/9605#discussion_r2251852708), this just renames ‘focus’ to ‘attention’ which confuses the matter…

I’d recommend at least leaving a to-do to follow-up on this since it’s definitely tech debt.

PR #9610, opened 2025-08-05
refactor: remove deprecated tool defs

richburdon:


packages/core/ai/src/deprecated/tools/tool.ts
```diff
@@ -136,13 +139,6 @@ export interface ExecutableTool extends Tool {
   execute: (params: unknown, context: ToolExecutionContext) => Promise<ToolResult>;
 }
 
-export const ToolId = Schema.String.pipe(Schema.brand('ToolId')).annotations({
```
dmaretskyi, line 139: Is this the same one used in blueprints?

packages/core/assistant-testing/src/plugins.tsx
```diff
@@ -41,21 +40,21 @@ export const capabilities: AnyCapability[] = [
...
-        - When presenting an image, you must use an artifact.
-        - Nest the <image> tag inside the <artifact> tag.
-        - Image tags are always self-closing and must contain an id attribute.
-          (Example: <artifact><image id="unique_identifier" prompt="..." /></artifact>)
-      `,
-      schema: Schema.Struct({}), // TODO(burdon): Add schema.
-      tools: [],
-    }),
-  ),
+  // contributes(
```
dmaretskyi, line 43: Is the idea that those become blueprints?

packages/core/assistant/src/conversation/conversation.ts
```diff
@@ -48,6 +48,7 @@ export class AiConversation {
    *
    * @deprecated Pass in a session instead.
    */
+  // TODO(burdon): Fix!!!
```
dmaretskyi: Elaborate pls

packages/core/assistant/src/session/session.ts
```diff
@@ -24,17 +23,19 @@ import { type ObjectId } from '@dxos/echo-schema';
...
 
-export type SessionRunOptions<Tools extends AiTool.Any> = {
+export type SessionRunParams<Tools extends AiTool.Any> = {
   prompt: string;
-  history: DataType.Message[];
-  systemPrompt?: string;
-  toolkit?: AiToolkit.AiToolkit<Tools>;
+  system?: string;
+  history?: DataType.Message[];
+  objects?: Obj.Any[]; // TODO(burdon): Meta only (typename and id -- write to binder).
```
dmaretskyi, line 36: Elaborate pls

packages/core/assistant/src/session/session.ts
```diff
@@ -74,63 +72,79 @@ export class AiSession {
...
+                <typename>${Obj.getTypename(object)}</typename>
+              </object>
+            `,
+          ) ?? [];
+        if (context.length) {
+          context.splice(0, 0, 'Context objects:');
+        }
+
+        // TODO(burdon): Pass objects here? Should they be pre-processed?
+        const prompts = yield* AiPreprocessor.preprocessAiInput([...this._history, ...this._pending]);
```
dmaretskyi: My intuition is that the preprocessor should handle the formatting of context

packages/core/assistant/src/session/session.ts
```diff
@@ -188,52 +202,54 @@ export class AiSession {
...
-
-        const artifactDiff = yield* Effect.tryPromise({
-          try: () =>
-            artifactDiffResolver.value.resolve(
-              [...versions.entries()].map(([id, version]) => ({ id, lastVersion: version })),
-            ),
-          catch: AiAssistantError.wrap('Artifact diff resolution error'),
-        });
+  // TODO(burdon): Implement abort.
+  abort(): void {
```
dmaretskyi: No, effect framework handles aborts (you call Effect.interrupt). This function should appear in chat processor (that's where we terminate effect) but not below.

packages/core/assistant/src/session/session.ts
```diff
@@ -188,52 +202,54 @@ export class AiSession {
...
+    // this._stream?.abort();
+    // Effect.interrupt?
+  }
+}
 
-        log.info('vision', {
-          artifactDiff,
-          versions,
-        });
+const formatUserPrompt = (prompt: string, history: DataType.Message[]) =>
```
dmaretskyi, line 206: Considering moving this to aipreprocessor, but you can leave this as a todo for me

packages/core/conductor/src/compiler/fiber-compiler.ts
```diff
@@ -343,17 +342,19 @@ export class GraphExecutor {
       const layer = Layer.mergeAll(
         Layer.succeed(Scope.Scope, yield* Scope.Scope),
         Layer.succeed(ComputeEventLogger, yield* ComputeEventLogger),
-        Layer.succeed(AiService, yield* AiService),
+        // TODO(burdon): Fix !!!
+        // Layer.succeed(AiService.AiService, yield* AiService.AiService),
```
dmaretskyi: Why? I don't see any obvious issues here

packages/core/conductor/src/nodes/gpt/gpt-node.test.ts
```diff
@@ -86,47 +85,49 @@ describe.runIf(process.env.DX_RUN_SLOW_TESTS === '1')('gptNode', () => {
...
+        );
         log.info('conversationMessages', { conversationMessages });
-        expect(conversationMessages.at(-1)?.role).toEqual('assistant');
+        expect(conversationMessages.at(-1)?.sender.role).toEqual('assistant');
       }),
       60_000,
     );
   });
 
-  it.skip(
```
dmaretskyi, line 97: Let's kill this for now

packages/core/assistant/src/conversation/conversation.ts
richburdon: Fix the deprecated annotation. Is this now OK?

packages/core/assistant-testing/src/plugins.tsx
richburdon: Yes, but are these functions or "built-in tools"?

packages/core/assistant/src/session/session.ts
richburdon: Just need the `<id, typename>` tuple here not the full object. 
Do refs currently have the typename? That might help later on so that we don't need to hydrate them to get the relevant metadata?

packages/core/assistant/src/session/session.ts
richburdon: OK I'll need to pair with you on this.

packages/core/assistant/src/session/session.ts
richburdon: Yeah I thought so. I.e., adding the `<object>` context which is currently done above?

packages/core/conductor/src/compiler/fiber-compiler.ts
richburdon: I couldn't undertand the error when this was uncommented. And everything appears to work?!
My late night assumption was that this is provided elsewhere?

packages/core/ai/src/deprecated/tools/tool.ts
richburdon: Yes I've moved it out of deprecated

packages/core/conductor/src/nodes/gpt/gpt-node.test.ts
richburdon: ok

.vscode/settings.json
```diff
@@ -63,9 +64,328 @@
...
+    // Orange
+    "FAE7E2",
+    "F0A390",
+    "EB785E",
+    "CA5F45",
+    "9D3C26",
+    "6B2210"
+  ],
+  "hediet.vscode-drawio.resizeImages": null,
+  "eslint.rules.customizations": [{ "rule": "prettier/*", "severity": "off" }]
```
wittjosiah, line 390: hmm, does this stop eslint from formatting code in the editor? seems like not great thing to enforce project-wide

packages/core/assistant-testing/src/blueprints/planning/planning.test.ts
```diff
@@ -121,7 +121,7 @@ describe.runIf(process.env.DX_RUN_SLOW_TESTS === '1')('Planning Blueprint', { ti
           TestDatabaseLayer({ types: [DataType.Text, Markdown.Document, Blueprint.Blueprint] }),
           makeToolResolverFromFunctions([readTasks, updateTasks]),
           makeToolExecutionServiceFromFunctions([readTasks, updateTasks]),
-          AiService.model('@anthropic/claude-3-5-sonnet-20241022'),
+          AiService.AiService.model('@anthropic/claude-3-5-sonnet-20241022'),
```
wittjosiah: the AiService statics should probably just be moved into the namespace so this doesn't need to change?

richburdon: actually the model static can't be moved -- there's some magic TS/Effect where it seems to inherit from the class.

.vscode/settings.json
richburdon: not it just defers until you save.

packages/core/ai/src/deprecated/tools/tool.ts
richburdon: yes i moved it to ../tools

packages/core/assistant-testing/src/plugins.tsx
richburdon: yes

packages/common/keys/src/dxn.ts
```diff
@@ -44,6 +44,7 @@ export type QueueSubspaceTag = (typeof QueueSubspaceTags)[keyof typeof QueueSubs
  * ```
  */
 export class DXN {
+  // TODO(burdon): Rename to DXN (i.e., DXN.DXN).
```
dmaretskyi, line 47: I don't think that's a good idea

packages/core/ai/src/AiService.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * AI Model Factory.
+ */
+export class AiService extends Context.Tag('@dxos/ai/AiService')<
+  AiService,
+  {
+    readonly model: (model: LLMModel) => Layer.Layer<AiLanguageModel.AiLanguageModel, AiModelNotAvailableError, never>;
+  }
+>() {
+  static model: (model: LLMModel) => Layer.Layer<AiLanguageModel.AiLanguageModel, AiModelNotAvailableError, AiService> =
```
dmaretskyi: consider moving model to top-level too

packages/core/assistant-testing/src/blueprints/planning/planning.test.ts
dmaretskyi: I don't think there're any hard blockers to it being moved to the top level

packages/core/assistant/src/session/session.ts
dmaretskyi: What about passing in the label (Obj.getLabel) too? -- I think having a full object is more flexible

packages/core/assistant/src/templates/assistant.tpl
```diff
@@ -15,13 +15,13 @@ Before responding, explain your reasoning and include your detailed chain-of-tho
 Include the following steps:
 
 - Analyze the structure and type of the content in the user's message.
-- Can you complete the task using the available artifact definitions?
-- If you can't complete the task using the available artifact definitions, query the list of available artifact definitions using the appropriate tool.
-- Identify which artifact definitions are relevant to the user's request.
+- Determine if you complete the task using the available blueprint definitions?
```
dmaretskyi, line 18: Is this section still relevant? This calls out things like `require_artifact_definitions` tool. I belive we've moved away from that approach in favor of having the user bind blueprints manually.

packages/core/blueprints/src/blueprint/blueprint.ts
```diff
@@ -51,6 +51,7 @@ export const Blueprint = Schema.Struct({
   /**
    * Array of tools that the AI assistant can use when this blueprint is active.
    */
+  // TODO(burdon): Include meta (e.g., description)?
```
dmaretskyi: elaborate pls

packages/core/conductor/src/compiler/fiber-compiler.ts
dmaretskyi: this is not provided elsewhere

dmaretskyi: uncomment this before merging

packages/common/keys/src/dxn.ts
richburdon: let's discuss. main thing is we need the non-FQ variant.

packages/core/assistant/src/templates/assistant.tpl
richburdon: this needs a complete review. i only just touched the surface. that's what i'll do next.

packages/core/blueprints/src/blueprint/blueprint.ts
richburdon: based on previous misunderstanding.

PR #9611, opened 2025-08-05
feat: Research agent

dmaretskyi:
closes DX-318


packages/plugins/plugin-assistant/src/components/ChatThread/ChatMessage.tsx
```diff
@@ -250,6 +254,51 @@ const panelClasses = 'flex flex-col is-full bg-activeSurface rounded-sm';
 const marginClasses = 'pie-4 pis-4';
 const paddingClasses = 'pis-2 pie-2 pbs-0.5 pbe-0.5';
 
+// TODO(dmaretskyi): Extract.
+const useResolvedRef = <T,>(ref: Ref.Ref<T>): T | undefined => {
```
dmaretskyi, line 258: It might be better to attach a ref resolver to this ref and then use `ref.target` with standard signals.

```ts
const useResolvableRef = (ref) => {
  setRefResolver(ref, space.db.createRefResolver({}))
}

// later
useResolvableRef(ref)
ref.target
```

dmaretskyi: Alternatively we can ensure that refs come "hydrated" (with resolvers attached) by doing 

```ts
// imaginary api
const ref = space.db.ref(dxn)
// or
const ref = space.db.graph.ref(dxn)
```

PR #9613, opened 2025-08-05
feat: 1-up view of a record

wittjosiah:
https://github.com/user-attachments/assets/2941e431-496b-408d-9ea2-1d7e8b793d70

Resolves DX-315


thure (approved):
lgtm!

Presumably we’d possibly mount this as a companion, or maybe even bake it into TableContainer? Either way it’s a great start.

PR #9617, opened 2025-08-06
refactor: Ai session loop

richburdon:


packages/core/assistant/src/session/session.ts
```diff
@@ -202,8 +175,72 @@ export class AiSession {
...
 }
 
+/**
+ * Build a combined toolkit from the blueprint tools and the provided toolkit.
+ */
+const createToolkit = <Tools extends AiTool.Any>({
+  blueprints = [],
+  toolkit,
+}: Pick<SessionRunParams<Tools>, 'blueprints' | 'toolkit'>) =>
+  Effect.gen(function* () {
```
dmaretskyi, line 190: I suggest using `Effect.fn` (for this and in-general) so we have stack-trace info

packages/core/assistant/src/session/session.ts
```diff
@@ -202,8 +175,72 @@ export class AiSession {
...
+ * Formats the system prompt.
+ */
+// TODO(burdon): Move to AiPreprocessor.
+const formatSystemPrompt = ({
+  system,
+  blueprints,
+  objects = [],
+}: Pick<SessionRunParams<any>, 'system' | 'blueprints' | 'objects'>) =>
+  Effect.gen(function* () {
+    const b = yield* pipe(
```
dmaretskyi: `const b` ?? you're going against your core principles

packages/core/assistant/src/session/session.ts
```diff
@@ -180,6 +152,7 @@ export class AiSession {
         yield* this.messageQueue.offer(toolResultMessage);
       } while (true);
 
+      // TDOO(burdon): Comment required.
```
wittjosiah: the queues shutting down will signal to stream consumers that the session has completed and no more messages are coming

packages/core/assistant/src/session/session.ts
```diff
@@ -154,9 +122,11 @@ export class AiSession {
           Stream.runCollect,
           Effect.map(Chunk.toArray),
         );
+
+        // TODO(burdon): Comment.
```
wittjosiah, line 127: this was necessary to coordinate between the block and message queues so that the streaming Rx value has a signal to clear itself when the message was emitted and not render the message twice.

Dima also has some concerns about consumers coming in partway through a session and there being a backlog in the blocks queue, so prior to this anything unconsumed in the block queue should probably be flushed as well.

packages/core/assistant/src/session/session.ts
richburdon: ha ha. no logging mistake.

packages/core/assistant/src/session/session.ts
richburdon: ok i was fullowing the form of the structure below. will add todo and research.

PR #9620, opened 2025-08-06
feat: generic load function

wittjosiah:


packages/plugins/plugin-assistant/src/functions/load.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Effect, Option, Schema, pipe } from 'effect';
+
+import { ArtifactId } from '@dxos/assistant';
+import { Obj } from '@dxos/echo';
+import { DatabaseService, defineFunction } from '@dxos/functions';
+import { log } from '@dxos/log';
+
+// TODO(burdon): Common error handling.
+// TODO(burdon): Factor out to space plugin.
+export default defineFunction({
```
richburdon, line 14: i was thinking that we'd move this to the space plugin, but i think it's good to have it here for the time being.

richburdon (approved):

PR #9623, opened 2025-08-07
fix: Fix automerge document synchronizer (client <-> worker sync)

mykola-vrmchk:
- **Bump automerge repo deps**
- **Bump automerge repo utils**
- **Bump dependencies**
- **Add test**
- **Fix flush bug**

### DIFF
--- a/packages/core/echo/echo-db/src/automerge/repo-proxy.test.ts
+++ b/packages/core/echo/echo-db/src/automerge/repo-proxy.test.ts
@@ -130,7 +130,7 @@ describe('RepoProxy', () => {
     }
   });
 
-  test('document persists without flush', async () => {
+  test('new document persists without `flush`', async () => {
     const path = `/tmp/dxos-${PublicKey.random().toHex()}`;
     let url: AutomergeUrl;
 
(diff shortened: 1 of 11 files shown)


packages/core/echo/echo-db/src/automerge/repo-proxy.test.ts
```diff
@@ -141,9 +141,44 @@ describe('RepoProxy', () => {
       await openAndClose(clientRepo);
 
       const text = 'Hello World!';
-      const clientHandle = clientRepo.create<{ text: string }>({ text });
+      const clientHandle = clientRepo.create<{ text: string; mutation_text?: string }>({ text });
       url = clientHandle.url;
-      await sleep(500); // Wait for the object to be saved.
+      await sleep(200); // Wait for the object to be saved.
```
dmaretskyi: Can we call flush here instead of sleeping?

mykola-vrmchk: this is testing if 'new document persists without `flush`'

PR #9624, opened 2025-08-07
feat: plugin blueprints

richburdon:


packages/core/assistant-testing/src/functions/document/update.ts
```diff
@@ -22,11 +21,7 @@ export default defineFunction({
   }),
   outputSchema: Schema.Void,
   handler: Effect.fn(function* ({ data: { id, content } }) {
-    const doc = yield* DatabaseService.resolve(ArtifactId.toDXN(id));
-    if (!doc || !Obj.instanceOf(Markdown.Document, doc)) {
-      throw new Error('Document not found.');
-    }
-
+    const doc = yield* DatabaseService.resolve(ArtifactId.toDXN(id), Markdown.Document);
```
dmaretskyi, line 24: 👍

packages/core/echo/echo/src/Type.ts
```diff
@@ -80,13 +80,17 @@ export namespace Obj {
    * NOTE: This is not an instance type.
    */
   // TODO(dmaretskyi): If schema was covariant, we could specify props in here, like `id: ObjectId`.
-  export type Any = Schema.Schema.AnyNoContext;
+  // TODO(burdon): This erases the ECHO type info (e.g., id, typename).
+  // export type Any = Schema.Schema.AnyNoContext;
+  export type Type<T extends EchoSchema.HasId> = Schema.Schema<T, any, never>;
```
dmaretskyi: I dont think this type is useful

PR #9625, opened 2025-08-07
feat: ai toolkit capability

wittjosiah:
poc is show tool in the deck plugin

### DIFF
--- a/packages/plugins/plugin-assistant/src/hooks/useChatProcessor.ts
+++ b/packages/plugins/plugin-assistant/src/hooks/useChatProcessor.ts
@@ -38,18 +38,6 @@ export const useChatProcessor = ({
   const registry = useContext(RegistryContext);
   const { dispatchPromise: dispatch } = useIntentDispatcher();
 
-  // TODO(wittjosiah): Remove?
-  // Services.
-  // const remoteServices = useQuery(space, Filter.type(ServiceType));
-  // const [serviceTools, setServiceTools] = useState<ExecutableTool[]>([]);
-  // useEffect(() => {
-  //   log('creating service tools...');
-  //   queueMicrotask(async () => {
-  //     const tools = await Promise.all(remoteServices.map((service) => createToolsFromService(service)));
-  //     setServiceTools(tools.flat());
-  //   });
-  // }, [remoteServices]);
-
   // Tools and context.
   const chatId = useMemo(() => (chat ? fullyQualifiedId(chat) : undefined), [chat]);
   const extensions = useMemo(() => ({ space, dispatch, pivotId: chatId }), [dispatch, space, chatId]);


packages/core/assistant-testing/src/blueprints/design/design.test.ts
```diff
@@ -94,8 +95,8 @@ describe.runIf(process.env.DX_RUN_SLOW_TESTS === '1')('Design Blueprint', { time
       Effect.provide(
         Layer.mergeAll(
           TestDatabaseLayer({ types: [DataType.Text, Markdown.Document, Blueprint.Blueprint] }),
-          makeToolResolverFromFunctions([readDocument, updateDocument]),
-          makeToolExecutionServiceFromFunctions([readDocument, updateDocument]),
+          makeToolResolverFromFunctions([readDocument, updateDocument], testToolkit),
```
dmaretskyi, line 98: what is this for?

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
```diff
@@ -19,7 +19,11 @@ export default () => {
         key: 'dxos.org/blueprint/assistant',
         name: 'Assistant',
         instructions: templates.system,
-        tools: functions.map((tool) => ToolId.make(tool.name)),
+        tools: [
+          ...functions.map((tool) => ToolId.make(tool.name)),
+          // TODO(wittjosiah): Factor out.
+          ToolId.make('show'),
```
dmaretskyi, line 25: Can we import ShowToolkit and get the tool by name `ShowToolkit.tools.show.name` (or somthing like this

packages/plugins/plugin-assistant/src/hooks/useChatProcessor.ts
```diff
@@ -40,16 +38,17 @@ export const useChatProcessor = ({
...
-  const remoteServices = useQuery(space, Filter.type(ServiceType));
-  const [serviceTools, setServiceTools] = useState<ExecutableTool[]>([]);
-  useEffect(() => {
-    log('creating service tools...');
-    queueMicrotask(async () => {
-      const tools = await Promise.all(remoteServices.map((service) => createToolsFromService(service)));
-      setServiceTools(tools.flat());
-    });
-  }, [remoteServices]);
+  // const remoteServices = useQuery(space, Filter.type(ServiceType));
```
dmaretskyi: this is okay to remove

packages/plugins/plugin-deck/src/capabilities/toolkit.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { Capabilities, LayoutAction, type PluginContext, contributes, createIntent } from '@dxos/app-framework';
+
+class DeckToolkit extends AiToolkit.make(
+  AiTool.make('show', {
+    description:
+      'Show an item as a companion to an existing plank. This will make the item appear alongside the primary content. When supplying IDs, they must be fully qualified like space:object.',
+    parameters: {
+      // TODO(wittjosiah): Currently the chat is calling this with a dxn consistently.
+      id: Schema.String.annotations({
```
dmaretskyi: Use `ArtifactId` every time you want model to provide an object id or DXN

dmaretskyi: ArtifactId.toDXN(id).asEchoDXN().objectId

dmaretskyi (changes requested):

packages/core/assistant-testing/src/blueprints/design/design.test.ts
wittjosiah: what do you mean? we updated this to take a toolkit so its passing in a toolkit

packages/plugins/plugin-deck/src/capabilities/toolkit.ts
wittjosiah: so this isn't strictly an artifact id, its a graph node id, because you can open other things in the layout (e.g. settings)

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
wittjosiah: this is fair but if we're going to do that cross-plugin we need to separate the implementation so that plugins and not pull in the full implementation when importing. Do you have suggestions for how that should be setup?

packages/plugins/plugin-deck/src/capabilities/toolkit.ts
dmaretskyi: okay, makes sense. I think we'd need to add code to handle the case where its a DXN or just naked id, or <space>:<object id> tuple

packages/core/assistant-testing/src/blueprints/design/design.test.ts
dmaretskyi: oh, I got confused by the testToolkit naming

PR #9628, opened 2025-08-08
fix: Table relations story and e2e test

thure:
This PR:
- resolves DX-306

Screenshot 2025-08-07 at 17 30 09

### DIFF
--- a/packages/ui/react-ui-table/src/model/table-model.ts
+++ b/packages/ui/react-ui-table/src/model/table-model.ts
@@ -20,7 +20,7 @@ import {
 import { invariant } from '@dxos/invariant';
 import { ObjectId } from '@dxos/keys';
 import { isLiveObject } from '@dxos/live-object';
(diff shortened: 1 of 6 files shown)


packages/ui/react-ui-table/src/model/table-model.ts
```diff
@@ -118,6 +120,7 @@ export class TableModel<T extends TableRow = TableRow> extends Resource {
   constructor({
     view,
     schema,
+    space,
```
wittjosiah: why is this necessary? why can you not just use the space of the view? this makes the model more difficult to use as evidenced by needing to cast the props below. It seems perfectly valid to have a table model outside a space (e.g. dynamic tables)

thure: Because `getSpace` no longer returned a space for the `view`. Do you know why that would be?

PR #9630, opened 2025-08-08
chore: update moon version to 1.39.2 and add moon config for ai package

richburdon:


packages/core/ai/.prototools
```diff
@@ -0,0 +1 @@
+moon = "1.38.6"
```
dmaretskyi: I don't think you need a separate version specification here. Lets remove this file and use one version for entire repo

PR #9631, opened 2025-08-08
chore: Bump Vite@7.1.1

dmaretskyi:
## Summary

This PR bumps Vite and its related plugins to the latest versions across the monorepo.
Resolves DX-227

### Core Updates
- **Vite**: Updated from  to  
- **@vitejs/plugin-react-swc**: Updated to 
- **@vitest/* packages**: Updated all Vitest packages to 
- **vite-plugin-top-level-await**: Updated to 
- **vite-plugin-wasm**: Updated to 

### Patch Removal
- Removed  as it's no longer needed with the new version
- Updated  to remove the patch reference from 

### Configuration Changes
- Added  to  to handle compatibility issues with some plugins that haven't been updated to support Vite 7 yet

## Affected Packages
Updated Vite across 50+ packages in the monorepo, including:
- All React UI components packages
- Storybook tools
- Various plugins and applications

## Testing
- ✅ All builds pass
- ✅ All tests pass
- ✅ Pre-CI checks completed successfully

## Notes
Some Vite plugins (like , , ) haven't been updated to officially support Vite 7 yet, but they work correctly with the current setup. The  setting allows these to install without blocking the upgrade.

The removed patch was addressing source map issues in Vite 5.4.7 that appear to be resolved in Vite 7.1.1.


package.json
```diff
@@ -151,12 +151,12 @@
...
     "uuid": "^9.0.0",
-    "vite": "^5.4.7",
+    "vite": "^7.1.1",
     "vite-plugin-devtools-json": "^0.4.1",
-    "vite-plugin-glslify": "^2.2.1",
-    "vite-plugin-inspect": "0.8.9",
-    "vite-plugin-top-level-await": "^1.5.0",
-    "vite-plugin-wasm": "^3.4.1",
+    "vite-plugin-glslify": "^2.3.0",
+    "vite-plugin-inspect": "11.3.2",
```
wittjosiah: probably don't need this pinned anymore, it was before because we were on an old version and couldn't upgrade until we upgraded vite

pnpm-workspace.yaml
```diff
@@ -7,3 +7,5 @@ packages:
   - '!**/build/**'
   - '!**/dist/**'
   - '!**/out/**'
+
+strictPeerDependencies: false
```
wittjosiah: 👀 

did windsurf do this or did you? not sure this is a good idea

package.json
```diff
@@ -151,12 +151,12 @@
     "typescript-eslint": "^8.39.0",
     "unplugin": "^1.7.1",
     "uuid": "^9.0.0",
-    "vite": "^5.4.7",
+    "vite": "^7.1.1",
```
wittjosiah: using ^ here vs not using ^ elsewhere

PR #9634, opened 2025-08-08
feat: list objects from the database in via tools

wittjosiah:
Resolves DX-310

Of note:
- tool to list available schemas in the system for querying with
- tool to list available objects in the space based on a query by typename (currently references the schema list tool by name in the description in order to ensure valid typenames are used)

Given a prompt of just "list all organizations" it should be able to successfully query for the orgs in the space


packages/core/assistant/src/session/session.ts
```diff
@@ -154,7 +157,7 @@ export class AiSession {
         yield* this.messageQueue.offer(toolResultMessage);
       } while (true);
 
-      // The queues shutting down signals to stream consumers that the session has completed and no more messages are coming.
+      // Signal to stream consumers that the session has completed and no more messages are coming.
```
richburdon: Signals to...

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
```diff
@@ -22,6 +22,7 @@ export default () => {
         tools: [
           ...functions.map((tool) => ToolId.make(tool.name)),
           // TODO(wittjosiah): Factor out.
+          ToolId.make('list-schemas'),
```
richburdon: can we import this name from somewhere? or will that happen when it's factored out?

packages/plugins/plugin-assistant/src/capabilities/toolkit.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Capabilities, type PluginContext, contributes } from '@dxos/app-framework';
+import { Filter, Type } from '@dxos/echo';
+import { DatabaseService } from '@dxos/functions';
+import { ClientCapabilities } from '@dxos/plugin-client';
+import { SpaceCapabilities, getActiveSpace } from '@dxos/plugin-space';
+import { DataType } from '@dxos/schema';
+
+class SchemaToolkit extends AiToolkit.make(
+  AiTool.make('list-schemas', {
+    description: 'List the available schemas.',
```
richburdon: Retrieves the available schemas
(make from the point of view of the agent using the tool)
perhaps rename get-schemas?

packages/plugins/plugin-assistant/src/functions/list.ts
```diff
@@ -0,0 +1,38 @@
...
+//
+
+import { Effect, Schema } from 'effect';
+
+import { Filter, Obj, Query } from '@dxos/echo';
+import { DatabaseService, defineFunction } from '@dxos/functions';
+
+// TODO(burdon): Factor out to space plugin.
+export default defineFunction({
+  name: 'dxos.org/function/assistant/list',
```
richburdon, line 13: ${meta.id}?

packages/plugins/plugin-assistant/src/functions/list.ts
```diff
@@ -0,0 +1,38 @@
...
+  description:
+    // TODO(wittjosiah): Find a better way to prompt for looking up typenames before querying with them.
+    'Lists the objects of the given type. Check the list-schemas tool for available types before calling this function.',
```
richburdon: Prefer each sentence on a newline. with `trim` wrapper (dxos/util).

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
wittjosiah: we need to figure out we want to separate the definitions and implementations in order to do that, but yes ideally we import them from somewhere

PR #9639, opened 2025-08-11
chore: Update to node 24

dmaretskyi:
- **Test on node 24**
- **CI on specific node version**
- **Update node in prototools**


dmaretskyi:
have to build a CI container for node 24

PR #9640, opened 2025-08-11
chore: Bump effect deps

dmaretskyi:
closes DX-300


dmaretskyi (commented):
Closes DX-300

dmaretskyi (commented):
Closes DX-300

PR #9644, opened 2025-08-11
fix: name conflict leading to race condition

wittjosiah:
doesn't show up in the table plugin diff but if you expand the diff you'll note the duplicate name

Resolves DX-349


thure (approved):
Ah I see, subtle. I noticed some other duplicate id's, I definitely think it'd help if the framework were verbose about that.

PR #9653, opened 2025-08-12
feat: AssistantPlugin chat integration with card surfaces

thure:
This PR:
- resolves DX-342
- adjusts the ChatThread story to use PreviewPlugin and StorybookLayout to render popovers as they would work in Composer
- adjusts ChatThread to render `card--transclusion` Surfaces and `dx-ref-tag` webcomponents to connect to the cards infrastructure


packages/plugins/plugin-assistant/src/components/ChatThread/ChatThread.stories.tsx
```diff
@@ -232,11 +238,20 @@ const meta = {
   component: ChatThread,
   render: render(StoryContainer),
   decorators: [
-    withClientProvider({
-      createIdentity: true,
-      types: [DataType.Organization, DataType.Person],
+    withPluginManager({
```
richburdon, line 241: is it possible to just use `withClientProvider`; the plugin manager is really heavy. or do we need the full intent stack? @wittjosiah?

packages/plugins/plugin-assistant/src/components/ChatThread/ChatThread.stories.tsx
```diff
@@ -232,11 +238,20 @@ const meta = {
...
+      plugins: [
+        ClientPlugin({
+          onClientInitialized: async ({ client }) => {
+            await client.halo.createIdentity();
+          },
+          types: [DataType.Organization, DataType.Person],
```
richburdon, line 247: prefer defs before callbacks

richburdon (approved):

packages/plugins/plugin-assistant/src/components/ChatThread/ChatThread.stories.tsx
thure: I tried to leave `withClientProvider` there, but one of these plugins hit an RTE about the client context being unavailable…

PR #9656, opened 2025-08-13
feat: cli-next

wittjosiah:


packages/devtools/effect-cli/src/services/client-service.ts
```diff
@@ -0,0 +1,53 @@
...
+    Effect.gen(function* () {
+      // TODO(wittjosiah): Provide config via layer.
+      const client = new Client({ config: defaultConfig });
+      yield* Effect.tryPromise(() => client.initialize());
+      yield* Effect.addFinalizer(() => Effect.tryPromise(() => client.destroy()).pipe(Effect.orDie));
+      return client;
+    }),
+  );
+
+  static memoryLayer = Layer.scoped(
```
dmaretskyi:
```suggestion
  static layerMemory = Layer.scoped(
```

packages/devtools/cli-next/scripts/copy-package-json.ts
```diff
@@ -0,0 +1,34 @@
+import { FileSystem, Path } from "@effect/platform"
```
richburdon, line 3: lint

packages/devtools/cli-next/scripts/copy-package-json.ts
```diff
@@ -0,0 +1,34 @@
...
+import { NodeContext } from "@effect/platform-node"
+import { Effect } from "effect"
+
+const program = Effect.gen(function*() {
+  const fs = yield* FileSystem.FileSystem
+  const path = yield* Path.Path
+  yield* Effect.log("[Build] Copying package.json ...")
+  const json: any = yield* fs.readFileString("package.json").pipe(Effect.map(JSON.parse))
+  const pkg = {
```
richburdon, line 12: https://lodash.com/docs/#pick or effect equiv?

packages/devtools/cli-next/src/commands/functions/deploy.ts
```diff
@@ -0,0 +1,275 @@
...
+import { waitForSync } from '../../util';
+
+const file = Args.text({ name: 'file' }).pipe(Args.withDescription('The file to deploy'));
+
+const name = Options.text('name').pipe(Options.withDescription('The name of the function to deploy'), Options.optional);
+const version = Options.text('version').pipe(
+  Options.withDescription('The version of the function to deploy'),
+  Options.optional,
+);
+const spaceId = Options.text('spaceId').pipe(
```
richburdon, line 42: spaceKey?

packages/devtools/cli-next/src/commands/functions/deploy.ts
```diff
@@ -0,0 +1,275 @@
...
+  name,
+}: {
+  client: Client;
+  space: Space;
+  functionObject: FunctionType;
+  scriptFileName: string;
+  scriptFileContent: string;
+  name?: string;
+}) {
+  client.addTypes([ScriptType, DataType.Text]);
```
richburdon: file-local const for default types?

packages/devtools/cli-next/src/commands/halo/join.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Effect, Match } from 'effect';
+
+import { type AuthenticatingInvitationObservable, Invitation, InvitationEncoder } from '@dxos/client/invitations';
+import { invariant } from '@dxos/invariant';
+
+import { ClientService } from '../../services';
+
+const invitationCode = Args.text({ name: 'invitationCode' }).pipe(Args.withDescription('The invitation code'));
+
+export const join = Command.make('join', { invitationCode }, ({ invitationCode: encoded }) =>
```
richburdon: i think it's useful to have a `/** */` block for each command with a one line description just to orient the reader. But your call if you think it's useful or now.

packages/devtools/cli-next/src/commands/index.ts
```diff
@@ -0,0 +1,38 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import { Command, Options } from '@effect/cli';
+
+import { ENV_DX_PROFILE_DEFAULT } from '@dxos/client-protocol';
+
+import { ClientService, ConfigService } from '../services';
```
richburdon, line 9: i still prefer index files not to have anything except imports, but feel free to push back.

packages/devtools/cli-next/src/commands/spaces/query.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Effect, Option } from 'effect';
+
+import { Filter, SpaceId } from '@dxos/client/echo';
+
+import { ClientService } from '../../services';
+
+const rawSpaceId = Options.text('spaceId').pipe(Options.withDescription('The space ID to query'));
+const typename = Options.text('typename').pipe(Options.withDescription('The typename to query'));
+
+export const query = Command.make('query', { rawSpaceId, typename }, ({ rawSpaceId, typename }) =>
```
richburdon: is it possible to establish a test for any effect CLI commands?

packages/devtools/cli-next/src/commands/functions/deploy.ts
wittjosiah: spaceKey is supposed to be internal since we introduced spaceId

packages/devtools/cli-next/src/commands/functions/deploy.ts
wittjosiah: I ported all this logic directly from the old CLI, it's a mess, there's a todo above to refactor it

packages/devtools/cli-next/scripts/copy-package-json.ts
wittjosiah: ha, I didn't even write this, it's straight from the @effect/cli template 😅 will try to get it included in lint at least 👍

packages/devtools/cli-next/src/commands/index.ts
wittjosiah: yeah it started as just command which grouped the sub-commands but then during cleanup ended up being where the services were provided... can think about another way to organize the commands

packages/devtools/cli-next/src/commands/spaces/query.ts
wittjosiah: good question, I'll ask Max

packages/core/functions/src/edge/functions.ts
```diff
@@ -29,6 +29,7 @@ export const uploadWorkerFunction = async ({
   functionId,
   ownerPublicKey,
 }: UploadWorkerArgs): Promise<UploadFunctionResponseBody> => {
+  log('Uploading function', { functionId, name, version, source: source.length, ownerPublicKey });
```
richburdon: nit. for some reason we've always lower-cased our logs (distinct from errors).

packages/devtools/cli-next/src/commands/functions/deploy.ts
richburdon: oh right sorry i got that the wrong way around.

packages/devtools/cli-next/src/commands/functions/deploy.ts
richburdon: ok but i'm just suggesting a very low impact way to start to move this fwd.

packages/devtools/cli-next/src/commands/spaces/query.ts
richburdon: +1. yeah i think we've suffered from not having this -- i.e., cli often broken.

wittjosiah: 100% agree. It's a often unattended entrypoint with a lot of surface area, some minimal tests would do it a lot of good.<br><br>tldr from Max: there's not any utils for testing commands currently. The paths for testing look like:<br>a) what they've done in the cli repo itself [https://github.com/Effect-TS/effect/blob/main/packages/cli/test/Command.test.ts](https://github.com/Effect-TS/effect/blob/main/packages/cli/test/Command.test.ts)<br>b) separating the command effect from the command to be able to test the effect directly<br><br>which makes the most sense probably depends on the circumstances. I'll see if I can put together at least one example test for a command

PR #9668, opened 2025-08-14
feat: Prompt options modal design increment

thure:
This PR:
- resolves DX-331

https://github.com/user-attachments/assets/78b6cc09-b050-4b49-8ee4-ca3124e79729

https://github.com/user-attachments/assets/09521d68-5172-4d5d-958b-15d0424a0ae4

## Caveats

- reactivity for objects in context doesn’t appear to work, does `processor.context.bind`/`.unbind` work? Is there a signal the component needs to subscribe to for changes?
  - this is true for the extant `ChatReferences` as well, the deletion visually working there appears to be due to `TagPicker` keeping its own internal state

Screenshot 2025-08-13 at 18 21 33


packages/plugins/plugin-assistant/src/components/Chat/Chat.tsx
```diff
@@ -319,11 +320,17 @@ const ChatPrompt = ({
...
-    void processor.context.bind({ objects: dxns.map((dxn) => Ref.fromDXN(DXN.parse(dxn))) });
-  }, []);
+  // TODO(thure): The components depending on `processor.context.objects` are not reacting to changes, why is this?
 
+  const handleReferenceChange = useCallback<NonNullable<ChatOptionsProps['onObjectChange']>>(
+    (dxn, checked) => {
+      log.info('update', { dxn, checked });
+      return processor.context[checked ? 'bind' : 'unbind']({ objects: [Ref.fromDXN(DXN.parse(dxn))] });
+    },
+    [processor.context.objects],
```
wittjosiah: why is this needed as a dep? experiment to try to get reactivity?

packages/plugins/plugin-assistant/src/components/Chat/Chat.tsx
```diff
@@ -351,18 +358,17 @@ const ChatPrompt = ({
...
-        onUpdate={handleUpdateReferences}
-      />
-
       <ChatOptions
+        space={space}
         blueprintRegistry={processor.blueprintRegistry}
-        context={processor.context}
-        onUpdateBlueprint={onUpdateBlueprint}
+        context={
+          /* TODO(thure): Why are we prop-drilling `context` and `blueprintRegistry`? If it’s a context can we not just `useContext`? */ processor.context
```
wittjosiah: I think just because these concepts are rather in flux and shift around a lot. There's still a question as to how the binding happens and whether or not it should have access to the whole registry at that time (its fine when registry is static but maybe less so if its open/dynamic)

packages/plugins/plugin-assistant/src/components/Chat/Chat.tsx
thure: Right, to no avail yet, will remove this.

PR #9674, opened 2025-08-15
feat: Conversation branching

dmaretskyi:
Part of DX-354

### DIFF
--- a/packages/sdk/schema/src/common/message.ts
+++ b/packages/sdk/schema/src/common/message.ts
@@ -258,6 +258,23 @@ export namespace ContentBlock {
   }).pipe(Schema.mutable);
   export interface Toolkit extends Schema.Schema.Type<typeof Toolkit> {}
 
+  export const ConversationFork = Schema.TaggedStruct('threadFork', {
+    ...Base.fields,
+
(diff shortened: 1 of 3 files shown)


packages/sdk/schema/src/common/message.ts
```diff
@@ -299,8 +321,18 @@ export namespace ContentBlock {
 const MessageSchema = Schema.Struct({
   id: ObjectId,
 
+  /**
+   * ID of the preceding message in the conversation.
+   * For merges
```
richburdon, line 326: broken/truncated comment?

packages/sdk/schema/src/common/message.ts
```diff
@@ -258,6 +258,23 @@ export namespace ContentBlock {
   }).pipe(Schema.mutable);
   export interface Toolkit extends Schema.Schema.Type<typeof Toolkit> {}
 
+  export const ConversationFork = Schema.TaggedStruct('threadFork', {
```
richburdon, line 261: Since these are namespaced i think fork/merge is enough, but otherwise Fork/MergeThread? (or at least agree with the tagged string)

richburdon (approved):

PR #9675, opened 2025-08-15
feat: git timeline nav

richburdon:
https://github.com/user-attachments/assets/7eefcf48-694e-436d-92eb-e94da82eaffb


packages/plugins/plugin-assistant/src/components/Timeline/Timeline.tsx
```diff
@@ -68,163 +27,278 @@ export type Commit = {
...
+            const idx = getBranchIndex(commit.branch);
+            if (idx === -1) {
+              return null;
+            }
+
+            return (
+              <div
+                key={commit.id}
+                data-index={index}
+                aria-current={current === index}
```
thure, line 152: `aria-current` is for a master-detail pattern where the item that is displayed in the detail is marked with this attribute in the master.

If this is the idea of “active“, `aria-selected` is more appropriate, but it depends on the role, but this element has no role and isn’t focusable. What does selection do here?

If it’s just a visual style, Radix uses `data-selected` to avoid conflicts with a11y standards, you could do that here.

packages/plugins/plugin-assistant/src/components/Timeline/Timeline.tsx
```diff
@@ -68,163 +27,278 @@ export type Commit = {
...
+                className={mx(
+                  'group flex shrink-0 items-center gap-2 overflow-hidden',
+                  // TODO(burdon): Factor out fragment?
+                  'aria-[current=true]:!bg-primary-500 hover:bg-hoverSurface',
```
thure, line 156: Please use semantic tokens, either an existing one or mint new ones, since we no longer guarantee physical tokens will be present.


```suggestion
                  'aria-[current=true]:!bg-accentSurface aria-[current=true]:hover:!bg-accentSurfaceHover hover:bg-hoverSurface',
```

richburdon: please suggest how to do this for this specific use case -- which should be really prevalent -- i.e., in dozens of different places.

packages/plugins/plugin-assistant/src/components/Timeline/Timeline.tsx
richburdon: i'm not sure i understand this. 
i had always just used "selected" to mean current cursor, but thought i'd "learned" from you that "current" was more precise, but i'm not sure if we have a canonical example of this?

packages/plugins/plugin-assistant/src/components/Timeline/Timeline.tsx
thure: [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) has fairly specific use-cases, I can only think of two places we use this, both master-detail patterns (navtree and mailbox), and both use the attribute to style items differently to suit the particular situation.

The pattern of applying `accentSurface` conditionally is also one I thought we intended to deprecate because it was too loud, except in cases where the component has attention. I can certainly imagine a fragment for that but this appears to be the first such case where we would do this.

packages/plugins/plugin-assistant/src/components/Timeline/Timeline.tsx
thure: At the time iirc we’d been discussing master-detail scenarios like NavTree and Mailbox, where `aria-current` is appropriate. I can’t tell from the screenshot or code exactly how this is meant to work, but it looks like a `listbox` for which `aria-selected` would be the appropriate attribute: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role

PR #9686, opened 2025-08-18
refactor: reorganize CLI commands under edge namespace and improve test infrastructure

richburdon:


packages/devtools/cli-next/src/commands/edge/status/status.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { Command, Options } from '@effect/cli';
+import { Console, Effect } from 'effect';
+
+export const status = Command.make(
+  'status',
+  {
+    json: Options.boolean('json').pipe(Options.withDescription('Output in JSON format.')),
+  },
+  Effect.fn(function* ({ json }) {
```
wittjosiah, line 15: fyi unless you pipe `Options.withDefault` for `json` its going to be an `Option.Option<boolean>`

not a big deal here as its a stub

packages/devtools/cli-next/src/commands/spaces/list.ts
```diff
@@ -90,3 +86,7 @@ const getSyncIndicator = (up: boolean, down: boolean) => {
     }
   }
 };
+
+export const list = Command.make('list', {}, listSpaces).pipe(
```
wittjosiah:
```suggestion
export const list = Command.make('list', {}, handler).pipe(
```

packages/devtools/cli-next/src/testing/test-console.ts
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import util from 'util';
```
wittjosiah:
```suggestion
import util from 'node:util';
```

packages/devtools/cli-next/src/testing/test-logger.ts
wittjosiah, line 1: just delete this?

wittjosiah: or at least mark as deprecated with a note to migrate to test console

PR #9687, opened 2025-08-18
fix: deprecated `getSnapshot` causing schema RTE

thure:
This PR:
- switches to use the recommended `getSnapshot`
- adds an e2e test case so this doesn’t regress

https://github.com/user-attachments/assets/4bda2fd3-8618-4066-8423-bcf4995772e7

### DIFF
--- a/packages/ui/react-ui-table/src/playwright/smoke.spec.ts
+++ b/packages/ui/react-ui-table/src/playwright/smoke.spec.ts
@@ -5,6 +5,7 @@
 import { expect, test } from '@playwright/test';
 
 import { type DxGrid } from '@dxos/lit-grid';
+import { faker } from '@dxos/random';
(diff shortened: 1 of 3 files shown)


packages/ui/react-ui-table/src/playwright/smoke.spec.ts
```diff
@@ -242,7 +243,7 @@ test.describe('Table', () => {
     await page.close();
   });
 
-  test('new relations work as expected', async ({ browser, browserName }) => {
+  test.only('new relations work as expected', async ({ browser, browserName }) => {
```
wittjosiah: 👀

wittjosiah (approved):

packages/ui/react-ui-table/src/playwright/smoke.spec.ts
thure: 🤦‍♂️

PR #9691, opened 2025-08-19
fix: Kanban hook issue

thure:
This PR:
- resolves DX-367

### DIFF
--- a/packages/ui/react-ui-kanban/src/components/Kanban.tsx
+++ b/packages/ui/react-ui-kanban/src/components/Kanban.tsx
@@ -2,7 +2,7 @@
 // Copyright 2024 DXOS.org
 //
 
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect } from 'react';
 
 import { Surface } from '@dxos/app-framework';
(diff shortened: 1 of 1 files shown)


packages/ui/react-ui-kanban/src/components/Kanban.tsx
```diff
@@ -59,11 +53,10 @@ export const Kanban = ({ model, onAddCard, onRemoveCard }: KanbanProps) => {
         const uncategorized = columnValue === UNCATEGORIZED_VALUE;
         const prevSiblingId = index > 0 ? array[index - 1].columnValue : undefined;
         const nextSiblingId = index < array.length - 1 ? array[index + 1].columnValue : undefined;
-        const columnItem = useMemo(() => ({ id: columnValue }), [columnValue]);
         return (
           <CardStack.Root asChild key={columnValue}>
             <StackItem.Root
-              item={columnItem}
+              item={{ id: columnValue }}
```
wittjosiah, line 59: should this be memoized?

thure: I didn’t encounter any issues with overreactivity when this was not memoized. I think it should be fine for now, though if we encounter overreactivity this will be an obvious suspect.

PR #9697, opened 2025-08-20
feat: add ItemAnnotation to identify selectable types in AI options dialog

richburdon:
Query by item type.

- [x] Split object dialog from settings (blueprints, models)
- [x] Factor out tabbed panels.
- [x] Factor out hook to get types. 
- [ ] TODO: Add ItemAnnotation to all items.

Screenshot 2025-08-20 at 4 57 11 AM

### DIFF
--- a/packages/plugins/plugin-assistant/src/translations.ts
+++ b/packages/plugins/plugin-assistant/src/translations.ts
@@ -9,6 +9,7 @@ import { type Resource } from '@dxos/react-ui';
 import { meta } from './meta';
 import { Assistant } from './types';
 
+// TODO(burdon): Standardize translation names.
 export const translations = [
   {
     'en-US': {


packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatOptions.tsx
```diff
@@ -115,11 +62,10 @@ export const ChatOptions = ({
       <Popover.Root>
         <Popover.Trigger asChild>
           <IconButton
-            icon='ph--sliders-horizontal--regular'
             variant='ghost'
-            size={5}
+            icon='ph--sliders-horizontal--regular'
             iconOnly
-            label={t('context settings placeholder')}
+            label={t('button context settings')}
```
thure, line 63: This naming is nonstandard — throughout the repo the naming convention is “{purpose} {label|title|placeholder|etc}”, since the component to which it applies might change.

Is there a motivation for this naming change or is it unintentional?

richburdon: ah. sorry, i really wasn't aware of there being a standard so i was naively using my own ad hoc one -- that i've been using in this and other plugins, which is roughly "component type | purpose". i'll look into creating a claude code rule to fix this across the repo, but it would be important to document this.

how is this a "placeholders"?

richburdon: it would be helpful to make this more machine readable: i.e., "context.purpose.type` -- e.g., "assistant-container.cancel.button"). what do you think? if we can define a  deterministic standard, i'll create a simple workflow.

here's what we currently have in this package:

```
        'button suggest': 'Assistant suggestion',
        'button select option': 'Select option',
        'button cancel': 'Cancel',
        'button save': 'Save',
        'button run': 'Run',

        'calling tool label': 'Calling tool...',
        'calling label': 'Calling',
        'error label': 'Error',

        'search placeholder': 'Search...',
        'prompt placeholder': 'Ask or search for anything...',
        'button context objects': 'Add to context',
        'button context settings': 'Chat settings',
        'button microphone': 'Click to speak',
        'button cancel processing': 'Stop processing',
        'button scroll down': 'Scroll down',
        'blueprints in context title': 'Blueprints',
        'objects in context title': 'Content',
        'remove object in context label': 'Remove document',
        'chat model title': 'Model',

        'settings default label': 'Default',
        'settings custom prompts label': 'Use custom prompts',
        'settings llm provider label': 'LLM provider',
        'settings edge llm model label': 'Remote language model',
        'settings ollama llm model label': 'Ollama language model',
```

packages/plugins/plugin-assistant/src/hooks/useItemTypes.ts
```diff
@@ -0,0 +1,37 @@
...
+  useEffect(() => {
+    if (!space) {
+      return;
+    }
+
+    return space.db.schemaRegistry.query().subscribe(
+      (query) => {
+        const types = Array.from(
+          new Set(
+            [...space.db.graph.schemaRegistry.schemas, ...query.results].filter((type) =>
```
wittjosiah, line 24: are there ever "item schema" in the db right now?

packages/sdk/schema/src/annotations/item.ts
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import { createAnnotationHelper } from '@dxos/echo-schema';
+
+export const ItemAnnotationId = Symbol.for('@dxos/schema/annotation/Item');
+
+export const ItemAnnotation = createAnnotationHelper<boolean>(ItemAnnotationId);
```
wittjosiah, line 10: todo: reconcile w/ SpaceCapabilities.ObjectForm

packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatOptions.tsx
thure: > make this more machine readable

Are you referring to nesting keys for the sake of translation key management software? If you want to organize them that way, that’s an option, though we don’t have a tool to test within-plugin nesting.

thure: There are some good tips here: https://lokalise.com/blog/translation-keys-naming-and-organizing/

richburdon: i meant to make it tractable to claude code -- e.g., i could create a simple workflow to make sure we have consistent naming; i also looked into the lint tool that would check this.

thure: I’d imagine an LLM would be able to read this just fine, I’m less certain whether it will produce keys like this.

I guess we can ask claude code to produce some translations and adopt its key naming pattern if you want its keys to work seamlessly with ours (as long as that naming pattern is reasonably standard for i18n systems).

packages/plugins/plugin-assistant/src/hooks/useItemTypes.ts
richburdon: it's a static annotation

packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatOptions.tsx
richburdon: i'm just proposing it's something like "component-name.use.type" where type = button, etc.
but mainly: exactly three dot-separated parts (with hyphens as needed)

thure: Why exactly three? That would be more onerous than the current pattern (two encouraged but optional) and will be harder to stick to if we start e.g. renaming components.

thure: I’d rather no standard than one that is sure to add to tech- or naming-debt.

thure: fwiw our keys are already delineated by package, some packages already split them into multiple namespaces, which simplifies organization, if that is the concern.

I don’t know that adding more namespaces to the keys themselves will produce better results, I would be interested in a demonstration since otherwise I think there’s no good cause to change the standard.

PR #9698, opened 2025-08-20
feat: Debounce collection synchronizer updates

mykola-vrmchk:


packages/e2e/blade-runner/package.json
```diff
@@ -32,7 +32,7 @@
     }
   },
   "scripts": {
-    "run-tests": "tsx ./src/main.ts"
+    "run-tests": "moon run blade-runner:build && node ./dist/lib/node-esm/main.mjs"
```
dmaretskyi, line 35: no, instead setup a moon task for `run-tests` and make it depend on `blade-runner:build`

dmaretskyi (approved):

PR #9701, opened 2025-08-20
refactor: move common options to config provider

wittjosiah:


packages/devtools/cli-next/src/services/client-service.ts
```diff
@@ -2,20 +2,18 @@
...
 
-// TODO(burdon): How to get this from options?
-const verbose = true;
-
 // TODO(wittjosiah): Factor out.
 export class ClientService extends Context.Tag('ClientService')<ClientService, Client>() {
   static layer = Layer.scoped(
     ClientService,
     Effect.gen(function* () {
+      const verbose = yield* Config.boolean('VERBOSE').pipe(Config.withDefault(false));
```
richburdon, line 16: nice

PR #9705, opened 2025-08-21
feat: add tool for adding schemas

wittjosiah:


packages/plugins/plugin-assistant/src/capabilities/toolkit.ts
```diff
@@ -27,9 +28,23 @@ class SchemaToolkit extends AiToolkit.make(
     success: Schema.Any,
     failure: Schema.Never,
   }),
-  AiTool.make('create-record', {
+  AiTool.make('add-schema', {
```
richburdon, line 31: create-schema?

richburdon (approved):

packages/plugins/plugin-assistant/src/capabilities/toolkit.ts
wittjosiah: intent is already called AddSchema just trying to keep it consistent for the time being

PR #9707, opened 2025-08-21
chore: stub open tool in storybook

wittjosiah:


packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
```diff
@@ -10,10 +10,11 @@ import { Blueprint } from '@dxos/blueprints';
 import { analysis, list, load } from '../functions';
 
 const functions = [analysis, list, load];
-const tools: string[] = [
+const tools = [
+  // prettier-ignore
   'get-schemas',
   'create-record',
-  // 'open-item'
+  'open-item',
```
richburdon, line 19: TODO move to system plugin?

PR #9710, opened 2025-08-22
feat: add blueprint definitions with tools and functions to plugins

richburdon:
Partially resolves DX-401

Begin working on markdown diffs:

Screenshot 2025-08-25 at 2 30 42 PM


packages/core/assistant/src/session/session.ts
```diff
@@ -28,83 +27,36 @@ import { isNotFalsy } from '@dxos/util';
...
-  }),
+export type AiSessionRunRequirements<Tools extends AiTool.Any> =
+  | AiTool.ToHandler<Tools>
+  | AiLanguageModel.AiLanguageModel
+  | ToolResolverService
+  | ToolExecutionService
+  | TracingService;
 
-  noop: () => GenerationObserver.make(),
+export type AiSessionRunEffect<Tools extends AiTool.Any> = Effect.Effect<
```
dmaretskyi, line 41: I agree with extracting Error and Requirements but I think typedeffing effect is generally an anti-pattern

dmaretskyi: Its the same as typing `type SessionRunPromise = Promise<DataType.Message[]>` vs `Promise<SessionRunResult>`

packages/core/assistant/src/session/session.ts
```diff
@@ -300,3 +259,5 @@ const createToolkit = <Tools extends AiTool.Any>({
       Effect.provide(blueprintToolkitHandler),
     ) as Effect.Effect<AiToolkit.ToHandler<any>, never, AiTool.ToHandler<Tools>>;
   });
+
+const createSnippet = (text: string, len = 32) => [text.slice(0, len), '...', text.slice(-len)].join('');
```
dmaretskyi: btw, this breaks on inputs less then 2*len

packages/core/assistant/src/session/session.ts
richburdon: ok but the same types are used in session and conversation and i want to prevent them from accidentally diverging. i'm currently doubting session should be a separate class/folder.

PR #9718, opened 2025-08-26
fix: Fix AutomergeHost._afterSave race condition

mykola-vrmchk:
`Resolves DX-404`

### DIFF
--- a/packages/core/echo/echo-db/src/query/query.test.ts
+++ b/packages/core/echo/echo-db/src/query/query.test.ts
@@ -21,7 +21,7 @@ import { range } from '@dxos/util';
 import { getObjectCore } from '../echo-handler';
 import type { Hypergraph } from '../hypergraph';
 import { type EchoDatabase } from '../proxy-db';
-import { EchoTestBuilder, type EchoTestPeer } from '../testing';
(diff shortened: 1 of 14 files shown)


packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -154,7 +154,7 @@ describe('Query', () => {
   });
 
   test('query.run() queries everything after restart', async () => {
-    const kv = createTestLevel();
+    const tmpPath = `/tmp/dxos-${PublicKey.random().toHex()}`;
```
dmaretskyi: extract to a function in /testing

packages/core/echo/indexing/src/indexer.ts
```diff
@@ -219,6 +219,16 @@ export class Indexer extends Resource {
    */
   async updateIndexes(): Promise<void> {
     await this._run.runBlocking();
+    // Note: Indexing task might schedule itself again if it run over indexing budget.
+    let iterations = 0;
+    while (this._run.scheduled) {
+      await this._run.runBlocking();
```
dmaretskyi: use .join instead

PR #9723, opened 2025-08-26
feat: basic proposals in documents

wittjosiah:
https://github.com/user-attachments/assets/ab719989-ac0c-4cc9-ada3-270eab36f238


thure (approved):
Very exciting stuff ✨

We might consider for follow-up adjusting the accept/reject design here so it’s a bit more discoverable, or at least make some visible distinction between comments that are suggestions vs comments that are discursive. That ontological predicate would just be whether the comment has an associated diff, right?

packages/plugins/plugin-thread/src/capabilities/toolkit.ts
```diff
@@ -0,0 +1,77 @@
...
+    failure: Schema.Never,
+  }),
+) {
+  static layer = (context: PluginContext) =>
+    Toolkit.toLayer({
+      'add-proposals': ({ id, diffs: _diffs }) =>
+        Effect.gen(function* () {
+          // TODO(wittjosiah): Get capabilities via layers.
+          const state = context.getCapability(Capabilities.Layout);
+          const dxn = ArtifactId.toDXN(id, state.workspace as SpaceId);
```
richburdon, line 43: grouping with TODO to rethink DXN processing since it get quite verbose

```ts
const state == // capabilities.
<new line>
const dxn = 
const echoDXN = 
```

packages/plugins/plugin-thread/src/capabilities/toolkit.ts
```diff
@@ -0,0 +1,77 @@
...
+          }
+
+          const client = context.getCapability(ClientCapabilities.Client);
+          const space = client.spaces.get(echoDxn.spaceId!);
+          const object = space?.db.getObjectById<Markdown.Document>(echoDxn.echoId);
+          if (!object) {
+            throw new Error(`Object not found: ${id}`);
+          }
+
+          const { dispatch } = context.getCapability(Capabilities.IntentDispatcher);
```
richburdon: move all capabilities up front (lends itself to progressively factoring out code into functions)?

packages/plugins/plugin-thread/src/capabilities/toolkit.ts
```diff
@@ -0,0 +1,77 @@
...
+// TODO(wittjosiah): How to make this work for more than Documents?
+class Toolkit extends AiToolkit.make(
+  AiTool.make('add-proposals', {
+    description: trim`
+      Proposes a set of changes to a document.
+    `,
+    parameters: {
+      id: ArtifactId,
+      diffs: Schema.Array(Schema.String).annotations({
+        description: 'The diffs to propose for the document.',
```
richburdon, line 27: i'm surprised this works; i.e., it doesn't seem detailed enough with respect to the types.

packages/plugins/plugin-thread/src/components/MessageContainer.tsx
```diff
@@ -133,6 +156,14 @@ const TextboxBlock = ({
   return <div role='none' ref={parentRef} className='mie-4' {...focusAttributes} />;
 };
 
+const ProposalBlock = ({ block }: { block: DataType.MessageBlock.Proposal }) => {
+  return (
+    <div role='none' className='mie-4 italic'>
+      {block.text}
```
richburdon, line 162: should we be mindful of sanitizing responses to check they don't include <script> tags? TODO upstream in AI parser?

packages/plugins/plugin-thread/src/components/ThreadComplementary.tsx
```diff
@@ -114,6 +115,28 @@ export const ThreadComplementary = ({ subject }: { subject: any }) => {
     [dispatch, subject],
   );
 
+  const handleAcceptProposal = useCallback(
```
richburdon: `useCallback<NonNullable<CommentsContainer['onAcceptProposal'>>`

packages/plugins/plugin-thread/src/types/types.ts
```diff
@@ -77,6 +77,16 @@ export namespace ThreadAction {
     output: Schema.Void,
   }) {}
 
+  export class AddProposal extends Schema.TaggedClass<AddProposal>()(`${THREAD_ACTION}/add-proposal`, {
```
richburdon: progressive cleanup: remove THREAD_ACTION and import `meta` (meta.id)

packages/plugins/plugin-thread/src/capabilities/toolkit.ts
wittjosiah: I just copied this from the existing diff function, the format comes from the blueprint

packages/plugins/plugin-thread/src/components/MessageContainer.tsx
wittjosiah: that should just be rendered as <script> text unless we're using `dangerouslySetInnerHTML` or whatever that prop is

wittjosiah:
@thure yes the rendering is just a naive hookup currently, would welcome some improvements there

PR #9724, opened 2025-08-26
fix: Distinguish between mutability of view vs schema in `ViewEditor`

thure:
https://github.com/user-attachments/assets/48257564-b32e-4a2d-80b1-3de00d626480

### DIFF
--- a/packages/ui/react-ui-list/src/components/List/ListItem.tsx
+++ b/packages/ui/react-ui-list/src/components/List/ListItem.tsx
@@ -170,12 +170,7 @@ export const ListItem = <T extends ListItemRecord>({ children, classNames, item,
 
   return (
     <ListItemProvider item={item} dragHandleRef={dragHandleRef}>
-      <div
-        ref={ref}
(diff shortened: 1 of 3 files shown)


packages/ui/react-ui-list/src/components/List/ListItem.tsx
```diff
@@ -170,12 +170,7 @@ export const ListItem = <T extends ListItemRecord>({ children, classNames, item,
 
   return (
     <ListItemProvider item={item} dragHandleRef={dragHandleRef}>
-      <div
-        ref={ref}
-        role='listitem'
-        className={mx('flex overflow-hidden relative', classNames, stateStyles[state.type])}
-        {...props}
-      >
+      <div ref={ref} role='listitem' className={mx('flex relative', classNames, stateStyles[state.type])} {...props}>
```
richburdon, line 173: stateStyles (which may have been me?) seems a bit obfuscating and might better be done inline with semantic tokens?

PR #9732, opened 2025-08-29
feat: inbox blueprint

wittjosiah:
https://github.com/user-attachments/assets/0f9a9719-2230-46f1-b76e-9b736b268ce3

TODO: fix mailbox/message container styles in chat story so they fill out container

Resolves DX-311

### DIFF
--- a/packages/plugins/plugin-assistant/src/stories/Chat.stories.tsx
+++ b/packages/plugins/plugin-assistant/src/stories/Chat.stories.tsx
@@ -165,6 +165,20 @@ const MARKDOWN_DOCUMENT = trim`
   Many applications extend the core syntax with extras (e.g., tables, task lists, math notation), but the core idea remains the same—clean, minimal markup that stays readable even without rendering.
 `;
 
+const STYLE_GUIDE = trim`
+  # Style Guide
+
+  ## Clear Communication
+
+  - Use short, simple sentences.
+  - Organize content with headings and bullet points.
+  - Avoid jargon and explain technical terms.
+  - Use active voice whenever possible.
+  - Highlight key points in bold.
+  - Keep paragraphs brief and focused on one idea.
+  - Proofread for clarity and correctness.
+`;
+
 const addSpellingMistakes = (text: string, n: number): string => {
   const words = text.split(' ');
   for (let i = 0; i < n; i++) {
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-assistant/src/stories/Chat.stories.tsx
```diff
@@ -198,7 +202,7 @@ export const WithDocument = {
   }),
   args: {
     components: [ChatContainer, [SurfaceContainer, CommentsContainer, LoggingContainer]],
-    blueprints: ['dxos.org/blueprint/assistant'],
+    blueprints: [BLUEPRINT_KEY],
```
richburdon: better than a raw string, but still doesn't feel right. since we have a lot of keys (e.g., the translation key, plugin key, etc.) not sure what to do.

packages/plugins/plugin-assistant/src/stories/components/MailboxContainer.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useQuery } from '@dxos/react-client/echo';
+
+import { type ComponentProps } from './types';
+
+export const MailboxContainer = ({ space }: ComponentProps) => {
+  const [mailbox] = useQuery(space, Filter.type(Mailbox.Mailbox));
+  const data = useMemo(() => ({ subject: mailbox }), [mailbox]);
+
+  // TODO(wittjosiah): Fix styles to fill the container.
+  return (
```
richburdon: i'll redo the DefaultStory "stack".
put anything here for now so that something shows up. e.g., fixed height.

packages/plugins/plugin-assistant/src/testing/test-data.ts
```diff
@@ -55,3 +55,55 @@ export const testTranscriptMessages = (): DataType.Message[] => {
 
   return transcription;
 };
+
+// TODO(wittjosiah): Find way to use data generator to generate substantive messages that could be summarized.
+export const testMailboxMessages = (): DataType.Message[] => {
```
richburdon: createTestMessages, createTestTranscription, etc. (i made same comment to M)

packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
```diff
@@ -14,15 +14,15 @@ import { StackItem } from '@dxos/react-ui-stack';
 import { TagPicker } from '@dxos/react-ui-tag-picker';
 
 import { InboxCapabilities } from '../../capabilities/capabilities';
-import { InboxAction, type MailboxType } from '../../types';
+import { InboxAction, type Mailbox } from '../../types';
 
 import { EmptyMailboxContent } from './EmptyMailboxContent';
-import { Mailbox, type MailboxActionHandler } from './Mailbox';
+import { type MailboxActionHandler, Mailbox as MailboxComponent } from './Mailbox';
```
richburdon, line 20: @thure this is the standard i'd like to establish -- that the types have the "natural" name, and that clashing components adjust (adding a "Component" suffix).

packages/plugins/plugin-inbox/src/functions/open.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const renderMarkdown = (message: DataType.Message): string[] => {
+  const sender =
+    message.sender.contact?.target?.fullName ??
+    message.sender.name ??
+    message.sender.email ??
+    message.sender.identityDid;
+  const blocks = message.blocks.filter((block) => block._tag === 'text');
+  return [
+    // prettier-ignore
```
richburdon, line 46: I've gone back to "//" because i want to force a new line but not totally disable prettier.

packages/plugins/plugin-inbox/src/functions/summarize.ts
```diff
@@ -0,0 +1,82 @@
...
+        AiService.model('@anthropic/claude-sonnet-4-0'),
+        ToolResolverService.layerEmpty,
+        ToolExecutionService.layerEmpty,
+        TracingService.layerNoop,
+      ),
+    ),
+  ),
+});
+
+const SUMMARIZE_PROMPT = trim`
```
richburdon: Just `prompt = `

packages/plugins/plugin-inbox/src/functions/open.ts
wittjosiah: I find `//` very confusing because its not clear what its purpose is. If `prettier-ignore` is disabling too much, I'd probably want to put in a short string which made it clear the purpose (e.g. `// force-multiline`)

PR #9745, opened 2025-08-31
feat: chat summary messages

richburdon:


packages/common/effect/src/ast.ts
```diff
@@ -135,15 +135,15 @@ const visitNode = (
   path: Path = [],
   depth = 0,
 ): VisitResult | undefined => {
-  const _result = test?.(node, path, depth);
+  const $result = test?.(node, path, depth);
```
dmaretskyi, line 138: why the dollar notation?

packages/common/util/src/unit.ts
```diff
@@ -0,0 +1,52 @@
...
+type Format = (n: number, precision?: number) => string;
+
+const Formatter = (unit: Unit): Format => {
+  return (n: number, precision = 2) => {
+    const value = n / unit.quotient;
+    return `${value.toFixed(precision)}${unit.symbol}`;
+  };
+};
+
+export const Unit = {
```
dmaretskyi, line 19: 👍

packages/core/assistant/src/session/session.ts
```diff
@@ -212,7 +214,28 @@ export class AiSession {
         yield* this.messageQueue.offer(toolResultsMessage);
         yield* observer.onMessage(toolResultsMessage);
         yield* TracingService.emitConverationMessage(toolResultsMessage);
-      } while (true);
+        toolCount++;
+      } while (true); // Tool loop.
+
+      // Summary.
+      // TODO(burdon): Get token count.
+      const summaryMessage = Obj.make(DataType.Message, {
```
dmaretskyi, line 222: why do we write a new message over just reducing the data in the existing messages?

richburdon: i think this is a very important stat once we add the token counts

packages/sdk/schema/src/common/message.ts
```diff
@@ -122,6 +123,42 @@ export namespace ContentBlock {
   }).pipe(Schema.mutable);
   export interface ToolResult extends Schema.Schema.Type<typeof ToolResult> {}
 
+  /**
+   * GPT Summary
```
dmaretskyi, line 127: session summary?

packages/sdk/schema/src/common/message.ts
```diff
@@ -122,6 +123,42 @@ export namespace ContentBlock {
...
+        model: Schema.optional(Schema.String),
+        input: Schema.optional(Schema.Number),
+        output: Schema.optional(Schema.Number),
+        total: Schema.optional(Schema.Number),
+      }),
+    ),
+    duration: Schema.optional(Schema.Number).annotations({
+      description: 'Duration in ms.',
+    }),
+    toolCalls: Schema.optional(Schema.Number),
```
dmaretskyi, line 143: could be inferred from the mesage history, what's the value in duplicating it?

packages/sdk/schema/src/common/message.ts
```diff
@@ -122,6 +123,42 @@ export namespace ContentBlock {
   }).pipe(Schema.mutable);
   export interface ToolResult extends Schema.Schema.Type<typeof ToolResult> {}
 
+  /**
+   * GPT Summary
+   */
+  export const Summary = Schema.TaggedStruct('summary', {
+    mimeType: Schema.optional(Schema.String),
+    message: Schema.String,
+    usage: Schema.optional(
```
dmaretskyi, line 132: I'd rather include the usage under eath message with `role=assistant`

richburdon: to avoid having a new message or to have greater preceision (which you have to reduce)? i don't see the disadvantage in having a message for really important stuff like this.

packages/sdk/schema/src/common/message.ts
richburdon: i think it's fine to roll this up now to make processing simpler given that we have this new message type.

richburdon (commented):
> why the dollar notation?  

i've reverted this; instead adding a Param suffix.

packages/common/effect/src/ast.ts
richburdon: reverted

PR #9746, opened 2025-08-31
refactor: Refactor of AiSession with new AiSessionRequest object

richburdon:
## In particular

- AiConversation
- AiSession
- AiSessionRequest (new)
- AiChatProcessor (new logic)

## Issues

- What is ArtifactDiffResolver for?
- What is ToolContextExtensions for?

## Demo

https://github.com/user-attachments/assets/2d579b76-70eb-4788-976a-ff70d12e2b63

### DIFF
--- a/.junie/guidelines.md
+++ b/.junie/guidelines.md
@@ -1,0 +1,123 @@
+# DXOS Developer Guidelines
+
+## Project Overview
+
+**DXOS** is a Decentralized Operating System platform for building fully decentralized applications. The main product is **Composer** - an extensible IDE for real-time and async collaboration.
+
+**Tech Stack:**
+- TypeScript
+- React
+- pnpm (package manager)
+- moon (build system)
+- Vitest (testing)
+- Playwright (e2e testing)
+- Storybook (component documentation)
+
+## Project Structure
+
+```
+dxos/
+├── packages/           # Main codebase organized by domain
+│   ├── apps/          # Applications (Composer, tasks-app, etc.)
+│   ├── common/        # Shared utilities and libraries
+│   ├── core/          # Core DXOS functionality
+│   ├── devtools/      # Development tools
+│   ├── e2e/           # End-to-end test suites
+│   ├── plugins/       # Plugin ecosystem
+│   ├── sdk/           # Software Development Kit
+│   └── ui/            # UI components and themes
+├── tools/             # Build tools and development utilities
+├── scripts/           # Automation scripts
+└── docs/              # Documentation
+```
+
+## Development Setup
+
+### Prerequisites
+```bash
+# Install native libraries (macOS)
+brew install cairo giflib git-lfs jpeg libpng librsvg pango pkg-config python-setuptools git unzip gzip xz
+
+# Install proto (toolchain manager)
+bash <(curl -fsSL https://moonrepo.dev/install/proto.sh)
+
+# Setup proto for shell
+eval "$(proto activate zsh --config-mode all)"
+```
+
+### Initial Setup
+```bash
+# Install toolchain
+proto install
+
+# Install dependencies
+pnpm i
+
+# Build everything
+moon :build
+```
+
+## Development Workflow
+
+### Running Applications
+```bash
+moon composer-app:serve    # Main Composer IDE
+moon tasks-app:serve      # Tasks application
+moon docs:serve           # Documentation site
+```
+
+### Development Commands
+```bash
+moon <package>:lint -- --fix # Lint and fix code
+moon <package>:build         # Build a package
+```
+
+## Testing
+
+### Unit Tests
+```bash
+moon <package>:test       # Test specific package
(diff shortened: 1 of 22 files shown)


packages/core/assistant/src/conversation/request.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) {}
+
+  get session() {
+    return this._session;
+  }
+
+  /**
+   * Runs the request.
+   */
+  async run(services: Layer.Layer<AiSessionRunRequirements<Tools>, AiModelNotAvailableError>) {
```
dmaretskyi: Any downside for using effect-based APIs? We'd have more consistency with conversation code

richburdon: What do you mean?

packages/core/assistant/src/session/request.ts
```diff
@@ -0,0 +1,60 @@
...
+  ) {}
+
+  get session() {
+    return this._session;
+  }
+
+  /**
+   * Runs the request.
+   */
+  async run(services: Layer.Layer<AiSessionRunRequirements<Tools>, AiModelNotAvailableError>) {
```
dmaretskyi: those should be effects, not promises

packages/core/assistant/src/conversation/request.ts
wittjosiah: I think he means `run` returning an `Effect.Effect` rather than a `Promise`

dmaretskyi (changes requested):

packages/core/assistant/src/conversation/conversation.ts
```diff
@@ -29,66 +34,85 @@ export type AiConversationOptions = {
...
-    Effect.gen(this, function* () {
+  public createRequest<Tools extends AiTool.Any>(params: AiConversationRunParams<Tools>) {
+    const session = new AiSession();
+    return new AiSessionRequest<Tools>(this.exec<Tools>({ session, ...params }), session);
+  }
+
+  /**
+   * Executes a request.
+   */
+  exec<Tools extends AiTool.Any>({
```
wittjosiah: should this be internal if the preferred api is to create a request?

wittjosiah: oh I see, you've not moved everything over to this api yet

packages/core/echo/echo-db/src/proxy-db/database.ts
```diff
@@ -68,7 +68,7 @@ export interface EchoDatabase {
 
   toJSON(): object;
 
-  getObjectById<T extends BaseObject = any>(id: string, opts?: GetObjectByIdOptions): AnyLiveObject<T> | undefined;
+  getObjectById<T extends BaseObject = any>(id: string, opts?: GetObjectByIdOptions): Live<T> | undefined;
```
wittjosiah, line 71: I don't think this is equivalent

packages/devtools/cli-next/src/bin.ts
```diff
@@ -35,27 +35,29 @@ const run = Command.run(dx, {
 const EXIT_GRACE_PERIOD = 1_000;
 const FORCE_EXIT = true;
 
-run(process.argv).pipe(
+// Work around Effect type system limitation where Requirements type becomes overly restrictive.
```
wittjosiah, line 38: did something change here?

packages/devtools/cli-next/src/commands/smoke.test.ts
```diff
@@ -32,5 +32,9 @@ describe('smoke tests', () => {
         yield* run(args('dx --json hub status'));
         assert.containSubset(logger.logs.at(1), { level: 'log', args: [{ status: 'ok' }] });
       }
-    }).pipe(Effect.provide(Layer.mergeAll(TestConsole.layer, NodeContext.layer)), Effect.scoped, Effect.runPromise));
+    }).pipe(
+      Effect.provide(Layer.mergeAll(TestConsole.layer, NodeContext.layer)),
+      Effect.scoped,
+      Effect.runPromise,
+    ) as Promise<void>);
```
wittjosiah: why does this need to be cast?

packages/plugins/plugin-assistant/src/capabilities/toolkit.ts
```diff
@@ -56,7 +58,9 @@ class Toolkit extends AiToolkit.make(
     success: Schema.Any,
     failure: Schema.Never,
   }),
+
   AiTool.make('create-record', {
+    // TODO(burdon): How does the model know what "record" schemas are?
```
wittjosiah: currently `get-schemas` marks them as records vs not

wittjosiah (approved):

packages/devtools/cli-next/src/bin.ts
richburdon: i had type errors and had windsurf fix it

PR #9747, opened 2025-09-01
fix: typename not required when creating a table

wittjosiah:
part of dx-421, I think this should resolve the first three items


thure (approved):
Works for me, sure 👌

PR #9756, opened 2025-09-02
refactor: chat via observer

wittjosiah:


packages/plugins/plugin-assistant/src/processor/processor.ts
```diff
@@ -324,4 +273,30 @@ export class AiChatProcessor {
...
+            created: new Date().toISOString(),
+            sender: { role: 'assistant' },
+            blocks: [...blocks, block],
+          }),
+        );
+      });
+    }.bind(this),
+  );
+
+  private _onMessage = Effect.fn(
```
richburdon: nit -- i'd plut this above onBlock

packages/core/conductor/src/nodes/gpt/node.ts
```diff
@@ -126,20 +126,16 @@ export const gptNode = defineComputeNode({
...
-    const tokenStream = Stream.fromQueue(session.eventQueue).pipe(
-      Stream.tap((event) =>
-        Effect.sync(() => {
-          // TODO(wittjosiah): Consider using Effect logger.
-          logger.log({
-            type: 'custom',
-            nodeId: logger.nodeId!,
-            event,
-          });
+    const tokenPubSub = yield* PubSub.unbounded<AiResponse.Part>();
```
richburdon, line 129: interesting

packages/plugins/plugin-assistant/src/processor/processor.ts
```diff
@@ -72,93 +73,43 @@ export type AiRequest = {
  */
 export class AiChatProcessor {
   // TODO(burdon): Comment required.
+  // TODO(wittjosiah): Rename.
```
richburdon: perhaps just `_rx` (like `ctx`) would be better. "registry" implies something bigger (more relevant to the functionality) is going on -- or something directly connected to the observer (in terms of naming). i know that is their class name, but this is a very low-level thing we need to repeat?

richburdon (approved):

PR #9761, opened 2025-09-03
feat: add repl

wittjosiah:
https://github.com/user-attachments/assets/8d0cf515-2755-4821-8d8d-a0d85551d798

### DIFF
--- a/packages/devtools/cli-next/src/commands/repl.ts
+++ b/packages/devtools/cli-next/src/commands/repl.ts
@@ -5,7 +5,7 @@
 import { Command, Prompt } from '@effect/cli';
 import { Effect } from 'effect';
 
-import { run } from '../bin';
+import { run } from './dx';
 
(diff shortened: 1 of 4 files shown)


packages/devtools/cli-next/src/commands/repl.ts
```diff
@@ -0,0 +1,26 @@
...
+    Effect.gen(function* () {
+      while (true) {
+        const rawCommand = yield* Prompt.text({ message: '' }).pipe(Prompt.run);
+        if (rawCommand === 'quit' || rawCommand === 'q') {
+          break;
+        }
+        yield* run(args(rawCommand));
+      }
+    }) as Effect.Effect<void, unknown, never>,
+).pipe(Command.withDescription('Enter a REPL to interact with the DXOS CLI.'));
```
richburdon, line 26: enter a repl? if user facing then just "Enter command"

PR #9762, opened 2025-09-03
chore: update storybook to 9.1.4 and improve chat UI styling

richburdon:


packages/ui/react-ui-theme/src/styles/layers/focus-ring.css
```diff
@@ -1,15 +1,16 @@
...
+  .dx-focus-ring-inset {
     &:not([disabled]),
     &[disabled='false'] {
       &:focus {
         @apply outline-none;
       }
 
       &:focus-visible {
-        @apply ring-focusLine ring-offset-focusOffset z-[1] ring-neutralFocusIndicator;
+        @apply ring-focusLine ring-offset-focusOffset z-10 ring-neutralFocusIndicator;
```
thure: This was intentionally the arbitrary `[1]` value to avoid conflicting with elements using the more canonical `10` in the same stacking context — if this change was intentional, please check focusable elements around usages of `z-10` to ensure they still work as expected, otherwise please revert.

packages/ui/react-ui-theme/src/styles/layers/focus-ring.css
```diff
@@ -74,10 +78,10 @@
 
       &:focus-visible {
         &::after {
-          @apply ring-focusLine ring-offset-focusOffset ring-inset z-[1] ring-neutralFocusIndicator;
+          @apply ring-focusLine ring-offset-focusOffset ring-inset z-10 ring-neutralFocusIndicator;
```
thure: Ditto here.

tools/storybook/.storybook/preview-head.html
```diff
@@ -1,22 +1,39 @@
 <script>
   window.global = window;
 </script>
-<style>
-  @layer dx-tokens, user-tokens, tw-base, dx-base, tw-components, dx-components, utilities;
-</style>
```
thure: This layer order declaration is critical for the normal function of the theme — it intentionally comes first, before any other declarations.

Please triple-check the CSS layers are applied in the correct order in the storybook before merging this change that moves it to a later declaration.

thure (commented):
There are a few high-risk changes here that I’d just want to be sure don’t cause regressions, otherwise this is fine.

packages/ui/react-ui-theme/src/styles/layers/focus-ring.css
richburdon: interesting. no i didn't mean to change this. must have been windsurf editing when i asked it why the error with the check was happening.

PR #9767, opened 2025-09-04
refactor: gmail sync story

wittjosiah:


packages/plugins/plugin-inbox/src/functions/gmail-sync.ts
```diff
@@ -2,94 +2,116 @@
...
-        try: () => space.db.query({ id: mailboxId }).first(),
-        catch: (e: any) => e,
-      });
-      const { objects } = yield* Effect.tryPromise({
-        try: () => space.queues.queryQueue(mailbox.queue.dxn),
-        catch: (e: any) => e,
-      });
-      const newMessages = yield* Ref.make([]);
-      const nextPage = yield* Ref.make(null);
+        makeRequest(getUrl(userId, undefined, { q, pageSize, pageToken })) as Effect.Effect<
```
dmaretskyi: might be cleaner to add a generic parameter to `makeRequest`. Even better would be to use effect-schema for response

packages/plugins/plugin-inbox/src/functions/gmail-sync.ts
```diff
@@ -2,94 +2,116 @@
...
+        >;
+
+      const mailbox = yield* DatabaseService.resolve(ArtifactId.toDXN(mailboxId), Mailbox.Mailbox);
+      const queue = yield* QueueService.getQueue<DataType.Message>(mailbox.queue.dxn);
+      const newMessages = yield* Ref.make<DataType.Message[]>([]);
+      const nextPage = yield* Ref.make<string | null>(null);
 
       do {
-        const last = objects.at(-1);
+        const last = queue.objects.at(-1);
```
dmaretskyi: It better to use `queue.queryObjects` in the imperative code (like functions).

`queue.objects` is the local reactive cache of the queue state, but its not guaranteed to be up-to-date, unless you call `queue.refresh`. `queue.queryObjects()` is guranteed to be up-to-date

packages/plugins/plugin-inbox/src/functions/gmail-sync.ts
```diff
@@ -2,94 +2,116 @@
...
-
-  handler: ({ context: { space }, data: { mailboxId, userId, after, pageSize } }: any) =>
+  handler: ({
+    data: { mailboxId, userId = 'me', after = format(subDays(new Date(), 30), 'yyyy-MM-dd'), pageSize = 100 },
+  }) =>
     Effect.gen(function* () {
-      yield* Effect.log('running gmail sync', { mailboxId, userId, after, pageSize });
+      yield* Console.log('running gmail sync', { mailboxId, userId, after, pageSize });
+
+      const {
```
dmaretskyi: Use credentials service. See pattern in https://github.com/dxos/dxos/blob/a248013c2aea8691cdbdfcfa07a2fba287ec6a12/packages/core/assistant-testing/src/util/auth.ts

dmaretskyi: perhaps create another helper (or option) for bearer auth

packages/plugins/plugin-inbox/src/functions/gmail-sync.ts
```diff
@@ -2,94 +2,116 @@
...
+        const last = queue.objects.at(-1);
         const q = last
           ? `in:inbox after:${Math.floor(new Date(last.created).getTime() / 1000)}`
           : `in:inbox after:${after}`;
         const pageToken = yield* Ref.get(nextPage);
+        yield* Console.log('listing messages', { q, pageToken });
         const { messages, nextPageToken } = yield* listMessages(userId, q, pageSize, pageToken);
         yield* Ref.update(nextPage, () => nextPageToken);
         for (const message of messages) {
           const messageDetails = yield* getMessage(userId, message.id);
```
dmaretskyi: this is a nice candidate to extract into a separate JS function `mapEmail`

dmaretskyi (dismissed):
`queue.objects` is an important change, the rest is stylistic

packages/plugins/plugin-inbox/src/functions/gmail-sync.ts
wittjosiah: yes, I would like to do that, I have a todo for parsing responses in there

packages/core/mesh/edge-client/src/edge-http-client.ts
```diff
@@ -259,8 +259,26 @@ export class EdgeHttpClient {
     body: UploadFunctionRequest,
     args?: EdgeHttpGetArgs,
   ): Promise<UploadFunctionResponseBody> {
+    const formData = new FormData();
+    formData.append('name', body.name ?? '');
+    formData.append('version', body.version);
+    formData.append('ownerPublicKey', body.ownerPublicKey);
+    formData.append('entryPoint', body.entryPoint);
```
dmaretskyi, line 266:
```ts
formData.append('manifest', JSON.stringify({ name, version, ownerPublicKey, entrypoint });
```

wittjosiah: why?

packages/core/mesh/edge-client/src/edge-http-client.ts
```diff
@@ -259,8 +259,26 @@ export class EdgeHttpClient {
...
+    for (const [filename, content] of Object.entries(body.assets)) {
+      formData.append(
```
dmaretskyi, line 268: I'm not 100% sure this is required, but Cloudflare uses `filename` as the chunk name, so its safer to stick to that.
I believe this is also the key for the chunk when form data gets parsed on the server.

```ts
      formData.append(
        filename,
        new Blob([content as Uint8Array<ArrayBuffer>], { type: 'application/octet-stream' }),
        filename,
      );
      ```

PR #9781, opened 2025-09-06
feat: UI effects (incl. streaming text processing)

richburdon:


packages/ui/react-ui-components/src/components/MarkdownViewer/MarkdownViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return <h1 className='pbs-1 pbe-1 text-xl'>{children}</h1>;
+  },
+  h2: ({ children }) => {
+    return <h2 className='pbs-1 pbe-1 text-lg'>{children}</h2>;
+  },
+  h3: ({ children }) => {
+    return <h3 className='pbs-1 pbe-1 text-base'>{children}</h3>;
+  },
+  blockquote: ({ children, ...props }) => (
+    <blockquote className='pis-4 mbs-2 mbe-2 pbs-2 pbe-2 border-l-4 border-primary-500 text-primary-500' {...props}>
```
thure:
```suggestion
    <blockquote className='pis-4 mbs-2 mbe-2 pbs-2 pbe-2 border-l-4 border-accentSurface text-accentText' {...props}>
```

Please continue to use semantic tokens, physical tokens will become a source of design- and tech-debt.

packages/ui/react-ui-components/src/components/Typewriter/extensions/typewriter.ts
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        return decorations;
+      },
+      provide: (f) => EditorView.decorations.from(f),
+    }),
+
+    EditorView.theme({
+      '.cm-typewriter-append': {
+        animation: 'typewriter-fade-in 1s ease-out forwards',
```
thure: `1s` seems quite slow to me… `200ms` is more common for big motions. I think the easing function also could do with adjustment, exp7 or some more refined curve for a motion this frequent might look nicer.

fwiw, this much motion should be gated behind [`motion-safe:`](https://v3.tailwindcss.com/docs/animation#prefers-reduced-motion) for users who [declare a sensitivity to motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion).

packages/ui/react-ui-components/src/components/MarkdownViewer/MarkdownViewer.tsx
richburdon: +1

packages/ui/react-ui-components/src/components/Typewriter/extensions/typewriter.ts
richburdon: did you run it?

PR #9790, opened 2025-09-09
feat: Blueprint for generating Discord digest

dmaretskyi:


packages/core/assistant-testing/src/blueprints/discord/discord.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Agent prompt instructions for managing hierarchical task lists.
+ */
+const instructions = trim`
+  You are able to fetch messages from Discord servers.
+
+  Known servers:
+
+  DXOS serverId: 837138313172353095
```
wittjosiah, line 21: I'd say todo to pull this out to the space

packages/core/assistant-testing/src/functions/discord/fetch-messages.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      TracingService.layerLogInfo(),
+      FetchHttpClient.layer,
+    ),
+  ),
+);
+
+const DXOS_SERVER_ID = '837138313172353095';
+
+describe('Feed', { timeout: 600_000 }, () => {
+  it.effect(
```
wittjosiah, line 47: should this be running in CI? needs access token?

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
```diff
@@ -4,133 +4,195 @@
...
-    created: message.timestamp,
-    blocks: [{ _tag: 'text', text: message.content }],
-  });
+type DiscordChannel = GuildChannelResponse | PrivateChannelResponse | PrivateGroupChannelResponse | ThreadResponse;
+
+const DEFAULT_AFTER = 1704067200; // 2024-01-01
+const DEFAULT_LIMIT = 500;
+const DEFAULT_IGNORE_USERNAMES = ['GitHub', 'Needle'];
+
+// TODO(dmaretskyi): Align with standard thread type.
```
wittjosiah, line 65: I'd say, todo: use thread type from plugin-thread

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
```diff
@@ -4,133 +4,195 @@
...
+              afterTs,
+              afterSnowflake: options.after,
+              after: parseSnowflake(options.after),
+              limit: options.limit,
+            });
+            const messages = yield* rest.listMessages(channel.id, options).pipe(
+              Effect.map(Array.map(makeMessage)),
+              Effect.map(Array.reverse),
+              Effect.catchTag('ErrorResponse', (err) =>
+                err.cause.code === 50001 ? Effect.succeed([]) : Effect.fail(err),
```
wittjosiah, line 164: what is 50001?

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
```diff
@@ -140,3 +202,44 @@ export default defineFunction({
...
+  const discordEpoch = 1420070400000n; // Discord Epoch (ms)
+  return new Date(Number((BigInt(snowflake) >> 22n) + discordEpoch));
+};
+
+const makeMessage = (message: MessageResponse): DataType.Message =>
+  Obj.make(DataType.Message, {
+    [Obj.Meta]: {
+      keys: [
+        { id: message.id, source: 'discord.com' },
+        { id: message.channel_id, source: 'discord.com/thread' },
```
wittjosiah, line 224: thread vs channel? it seems like you're not using channels here?

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
```diff
@@ -4,133 +4,195 @@
...
-          if (allMessages.length >= limit) {
-            break;
+      yield* TracingService.emitStatus({ message: `Will fetch from channels: ${channels.length}` });
+
+      const threads = yield* Effect.forEach(
+        channels,
+        Effect.fnUntraced(function* (channel) {
+          const allMessages: DataType.Message[] = [];
+
+          let lastMessage: Option.Option<DataType.Message> = Option.none();
```
wittjosiah, line 140: given this always starts from none I assume this sync is one shot and can't be run on a schedule without creating duplicate data?

packages/plugins/plugin-assistant/package.json
```diff
@@ -121,6 +121,7 @@
     "use-deep-compare-effect": "^1.8.1"
   },
   "devDependencies": {
+    "@dxos/assistant-testing": "workspace:*",
```
wittjosiah, line 124: why? it's already in "dependencies"

wittjosiah: I think we could to with a todo somewhere to stop importing blueprints from assistant testing though

packages/core/assistant-testing/src/blueprints/discord/discord.ts
dmaretskyi: Yes, they should really go into the `context` but I had trouble hooking that up in the story

packages/core/assistant-testing/src/functions/discord/fetch-messages.test.ts
dmaretskyi: `TestHelpers.taggedTest('sync')` prevents it from running unless `DX_TEST_TAGS=sync` is set.

To get the token locally you do `eval $(pnpm -ws 1p-credentials)`

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
dmaretskyi: What do you think about extracting Thread do @dxos/schema?

dmaretskyi: This seems to be a part of a bigger discussion around data-types. My intuition is that there should be separation between data types and UX.

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
dmaretskyi: I with I remembered 😓  -- it's probably "unauthorized"

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
dmaretskyi: If I understand their API correctly -- messages are always stored in channels. Channels could be "Channels" from UX like `#general` or they could be threads created by users -- see the code above that combines threads and channels.

dmaretskyi: Fetching messages from a channel will not include messages in subthreads of that channel -- they have their own channel_id equal to thread.id

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
dmaretskyi: you can specify `after` timestamp manually in the call parameters, but this doesn't sync with a particular queue.

packages/plugins/plugin-assistant/package.json
dmaretskyi: something from our tooling keeps adding it there automatically

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
wittjosiah: we've oscillated between grouping data types together and grouping data types in plugins and in between, trying to find a good balance there. There's trade-offs either way. The current theory is that @dxos/schema would hold common primitives (e.g. text, canvas, message, etc.) whereas the plugins hold more use case specific types (e.g. document, sketch, etc). So using that as a guideline, the question is whether thread & channel are common primitives or use case specific types.

packages/core/assistant-testing/src/functions/discord/fetch-messages.ts
wittjosiah: I see, we probably just need to work on our data structures so that we can map these better

PR #9793, opened 2025-09-09
refactor: replace React components with native widgets in MarkdownStream

richburdon:
Further simplification, moves away from React (or lit) for simple tag/components (e.g., prompt, summary).

Screenshot 2025-09-09 at 11 38 37 PM

### DIFF
--- a/packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/PromptWidget.tsx
+++ b/packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/PromptWidget.tsx
@@ -23,8 +23,8 @@ export class PromptWidget extends WidgetType {
     const el = document.createElement('div');
     el.className = 'flex justify-end';
     const inner = document.createElement('div');
-    // TODO(burdon): Container must set var (see ChatThread.tsx).
-    inner.className = 'p-2 rounded-sm bg-[--user-fill]';
+    // NOTE: Container must set var based on user's identity.
+    inner.className = 'p-2 bg-[--user-fill] rounded-sm';
     inner.textContent = this.text;
     el.appendChild(inner);
     return el;


packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/ElementWidget.ts
```diff
@@ -17,9 +21,9 @@ export class ElementWidget extends WidgetType {
 
   override toDOM(): HTMLElement {
     const el = document.createElement(this.tag);
-    Object.entries(this.props).forEach(([key, value]) => {
+    Object.entries(this.props ?? {}).forEach(([key, value]) => {
       if (key === 'children') {
-        el.innerHTML = value as string;
+        el.innerText = String(value);
```
thure, line 26: Yes, I was thinking we should do this for safety reasons.

packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/PromptWidget.tsx
```diff
@@ -0,0 +1,35 @@
...
+export class PromptWidget extends WidgetType {
+  constructor(private text: string) {
+    super();
+  }
+
+  override toDOM(): HTMLElement {
+    const el = document.createElement('div');
+    el.className = 'flex justify-end';
+    const inner = document.createElement('div');
+    inner.className = 'p-2 rounded-sm bg-green-500';
```
thure: We have semantic tokens for this — we should also try to resolve this to the user’s chosen hue like we do elsewhere, if that’s a design we want to stick with. Maybe leave a TODO?

packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/SummaryWidget.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export class SummaryWidget extends WidgetType {
+  constructor(private text: string) {
+    super();
+  }
+
+  override toDOM(): HTMLElement {
+    const el = document.createElement('div');
+    el.className = 'text-sm text-subdued';
+    el.textContent = this.text;
```
thure, line 26: Is this all this is meant to do? If we should extend this to match a certain React component, maybe best to leave a TODO that refers to the specific component so we (I) don’t loose track.

packages/ui/react-ui-components/src/components/MarkdownStream/extensions/widgets/PromptWidget.tsx
richburdon: yes i think i accidentally dropped the TODO for this. putting it back.

PR #9807, opened 2025-09-14
fix: Fix transcription storybook rendering

mykola-vrmchk:


packages/plugins/plugin-transcription/src/components/Transcript/transcript-extension.ts
```diff
@@ -52,6 +53,83 @@ export const transcript = ({ model, started }: TranscriptOptions): Extension =>
...
+
+          const scroller = view.scrollDOM;
+          let isAutoScrolling = false;
+          let hasScrolled = false;
+
+          let timeout: NodeJS.Timeout | undefined;
+          const scrollToBottom = (smooth = false) => {
+            scroller.style.scrollBehavior = smooth ? 'smooth' : '';
+
+            // Temporarily hide scrollbar to prevent flicker.
```
thure, line 75: Great, thank you ✨

PR #9808, opened 2025-09-14
feat: Local queue triggers

dmaretskyi:


dmaretskyi:
Continued in https://github.com/dxos/dxos/pull/9813

PR #9812, opened 2025-09-16
feat: Stack perpendicular arrow focus movement

thure:
To do:
- [x] When a nonempty stack is focused, arrows along its orientation should focus the first or last item
- [x] Handle empty sibling stack case
- [x] See if there’s a reasonable way to `scrollIntoView` that doesn’t clip the focus ring
- [x] Protect elements that may be children of stack items that handle arrow keys from losing focus

https://github.com/user-attachments/assets/b7ad2c05-9bd7-47b9-9db6-b3ba62fdaed5

https://github.com/user-attachments/assets/1043953d-de4d-4afe-a226-ed8c997bc359

### DIFF
--- a/packages/ui/lit-grid/src/dx-grid.ts
+++ b/packages/ui/lit-grid/src/dx-grid.ts
@@ -1480,7 +1480,7 @@ export class DxGrid extends LitElement {
       <div
         role="none"
         class="dx-grid"
-        data-arrow-keys="handles-all"
+        data-arrow-keys="all"
         style=${styleMap({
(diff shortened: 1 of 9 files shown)


thure:
> @thure what happens if you `Enter` into one of the cards? Do the arrow controls still work?

Yes, as long as they aren’t already handled by a focus mover on the same axis, in which case moving between stack items is still facilitated by the Escape key or by tabbing to a focusable element inside the car that doesn’t declare that it handles arrow keys.

PR #9813, opened 2025-09-16
feat: Local triggers and managed compute runtime

dmaretskyi:


packages/core/functions/src/index.ts
```diff
@@ -8,8 +8,7 @@ export * from './schema';
 export * from './trace';
 export * from './types';
 export * from './url';
-
-// Trigger dispatcher
-export * from './triggers/trigger-dispatcher';
+export * from './triggers';
 export * from './services';
 export * from './executor';
+export * as exampleFunctions from './examples';
```
wittjosiah, line 14:
```suggestion
export * as Example from './examples';
```

packages/core/functions/src/handler.ts
```diff
@@ -85,7 +86,10 @@ const __assertFunctionSpaceIsCompatibleWithTheClientSpace = () => {
   // const _: SpaceAPI = {} as Space;
 };
 
+const typeId = Symbol.for('@dxos/functions/FunctionDefinition');
```
wittjosiah, line 89: what's this for?

packages/plugins/plugin-assistant/src/processor/processor.ts
```diff
@@ -168,7 +170,7 @@ export class AiChatProcessor {
           log.error('request failed', { cause });
           return Effect.void;
         }),
-        Effect.runFork, // Runs in the background.
+        Runtime.runFork(runtime), // Runs in the background.
```
wittjosiah, line 173: I think you can also do `runtime.runFork`?

dmaretskyi: afaik, no

packages/core/functions/src/handler.ts
dmaretskyi: marker to check that this type is function definition

PR #9815, opened 2025-09-17
feat: Project UI v0

thure:
This PR resolves DX-472.

https://github.com/user-attachments/assets/675a8629-0c21-49f9-8461-6727ca6c97e9

https://github.com/user-attachments/assets/7eb3936d-d720-40e8-aadc-429672b52e87

### DIFF
--- a/packages/plugins/plugin-project/src/components/ProjectContainer.tsx
+++ b/packages/plugins/plugin-project/src/components/ProjectContainer.tsx
@@ -2,17 +2,26 @@
 // Copyright 2023 DXOS.org
 //
 
-import React from 'react';
+import React, { useCallback } from 'react';
 
+import { Surface } from '@dxos/app-framework';
 import { StackItem } from '@dxos/react-ui-stack';
 import { type DataType } from '@dxos/schema';
 
-import { Project } from './Project';
+import { type ItemProps, Project } from './Project';
+
+const ProjectItem = ({ item, projectionModel }: ItemProps) => {
+  return <Surface role='card--intrinsic' data={{ subject: item, projection: projectionModel }} limit={1} />;
+};
 
 export const ProjectContainer = ({ project }: { project: DataType.Project; role: string }) => {
+  const handleAddItem = useCallback(() => console.log('[project container]', 'To implement: handle add item'), []);
+  const handleAddColumn = useCallback(() => console.log('[project container]', 'To implement: handle add column'), []);
   return (
     <StackItem.Content>
-      <Project project={project} />
+      <Project.Root Item={ProjectItem} onAddItem={handleAddItem} onAddColumn={handleAddColumn}>
+        <Project.Content project={project} />
+      </Project.Root>
     </StackItem.Content>
   );
 };


packages/plugins/plugin-project/src/components/ViewCollectionColumn.tsx
```diff
@@ -0,0 +1,57 @@
...
+export type ViewCollectionColumnProps = {
+  viewRef: Ref<View>;
+};
+
+export const ViewCollectionColumn = ({ viewRef }: ViewCollectionColumnProps) => {
+  // Resolve the view from the viewRef using useQuery
+  const client = useClient();
+  const space = getSpace(viewRef);
+  const echoDxn = viewRef.dxn.asEchoDXN();
+  const [view]: View[] = useQuery(space, echoDxn ? Filter.ids(echoDxn.echoId) : Filter.nothing());
```
dmaretskyi:
```suggestion
  const view = viewRef.target;
```

packages/plugins/plugin-project/src/components/ViewCollectionColumn.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { Surface } from '@dxos/app-framework';
+import { type AnyEchoObject, type Ref } from '@dxos/echo-schema';
+import { useClient } from '@dxos/react-client';
+import { Filter, getSpace, useQuery, useSchema } from '@dxos/react-client/echo';
+import { Card, CardStack, StackItem, cardStackHeading } from '@dxos/react-ui-stack';
+import { type View } from '@dxos/schema';
+
+export type ViewCollectionColumnProps = {
+  viewRef: Ref<View>;
```
dmaretskyi: I'd recommend resolving the Ref upstream and architecting the component API to just accept the `View` object unless there's a good reason to do otherwise.

packages/plugins/plugin-project/src/components/ProjectContainer.tsx
```diff
@@ -4,16 +4,28 @@
...
+
+import { ViewCollectionColumn } from './ViewCollectionColumn';
 
 export const ProjectContainer = ({ object }: { object: DataType.Project; role: string }) => {
   return (
     <StackItem.Content>
-      <pre className='m-4 p-2 ring'>
-        <span>{fullyQualifiedId(object)}</span>
-      </pre>
+      <Stack orientation='horizontal' size='contain' rail={false}>
```
wittjosiah: This should also be a separate component, say `ColumnBoard`, that's not at the container level so that there can be stories to demonstrate both.

packages/plugins/plugin-project/src/components/ProjectContainer.tsx
```diff
@@ -4,16 +4,28 @@
...
+        {(
+          object.collections.filter((collection) => {
```
wittjosiah: I'd suggest something like:
```ts
const views =
  object.collections
    .map((ref) => ref.target)
    .filter((object) => Obj.instanceof(DataType.View, object));
```

packages/plugins/plugin-project/src/components/Project.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+  render: DefaultStory,
+  decorators: [withLayout({ fullscreen: true }), withTheme],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  decorators: [
```
wittjosiah, line 150: seems like the decorators could just be on the meta and not need to be duplicated between stories

PR #9816, opened 2025-09-17
feat: add order to query dsl

wittjosiah:


packages/core/echo/echo-pipeline/src/query/plan.ts
```diff
@@ -193,4 +194,13 @@ export namespace QueryPlan {
...
     exclude: Plan;
   };
+
+  /**
+   * Order the results of the plan.
+   */
+  export type OrderStep = {
+    _tag: 'OrderStep';
+
+    order: readonly QueryAST.Order[];
```
dmaretskyi, line 205: // defaults to natural order if empty

packages/core/echo/echo/src/query/dsl.ts
```diff
@@ -85,6 +114,14 @@ export interface Query<T> {
    */
   target(): Query<Type.Relation.Target<T>>;
 
+  /**
+   * Order the query results.
+   * Orders are specified in priority order. The first order will be applied first, etc.
+   * @param order - Order to sort the results.
+   * @returns Query for the ordered results.
+   */
+  orderBy(...order: Order<T>[]): Query<T>;
```
dmaretskyi:
```suggestion
  orderBy(...order: [Order<T>, ...Order<T>[]]): Query<T>;
```

this will force there to be at lease one element.

you can also use effect's NonEmptyArray helper

packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -64,6 +64,47 @@ describe('Query', () => {
       expect(objects).to.have.length(10);
     });
```
dmaretskyi, line 66: hmm, I don't get why this would be the case. Where does the natural order get enforced if the user didn't call `query.orderBy(...)`

PR #9818, opened 2025-09-18
feat: Research on a trigger

dmaretskyi:


packages/core/blueprints/src/prompt/prompt.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Blueprints that the prompt may utilize.
+   */
+  blueprints: Schema.Array(Type.Ref(Blueprint)),
+
+  /**
+   * Additional context that the prompt may utilize.
+   */
+  context: Schema.Array(Schema.Any),
```
wittjosiah, line 54: shouldn't this also be refs?

packages/core/blueprints/src/prompt/prompt.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Natural language instructions for the prompt.
+   * These should provide concrete course of action for the AI to follow.
+   */
+  instructions: Schema.String,
+
+  /**
+   * Blueprints that the prompt may utilize.
+   */
+  blueprints: Schema.Array(Type.Ref(Blueprint)),
```
wittjosiah, line 49: what is the reason to separate blueprints from other objects in the context at the data level?

packages/sdk/app-framework/src/App.tsx
```diff
@@ -156,7 +156,8 @@ export const useApp = ({
   );
 };
 
-const DELAY_PLACEHOLDER = 2_000;
+// TODO(dmaretskyi): This was 2 seconds before and stories were loading significantly slower.
```
wittjosiah, line 159: 👀

packages/core/blueprints/src/prompt/prompt.ts
dmaretskyi: probably yes, but wouldn't providing values here be useful? ....or maybe they can just go into instructions

packages/core/blueprints/src/prompt/prompt.ts
dmaretskyi: blueprint tools are made accesssible to the agent. The context gets treated just as data.

packages/sdk/app-framework/src/App.tsx
dmaretskyi: This affects composer too, perhaps we could disable the fallback in stories only?

packages/core/blueprints/src/prompt/prompt.ts
wittjosiah: I was just wondering why they couldn't be separated at runtime, but I guess in order to do that you'd have to resolve all the refs.

packages/core/blueprints/src/prompt/prompt.ts
wittjosiah: I'd assume they're all objects for now unless you have a concrete example of why they wouldn't be, otherwise you're going to have to be dealing with the fact that you don't know the type of this whenever interacting with it

packages/sdk/app-framework/src/App.tsx
wittjosiah: This should fix it https://github.com/dxos/dxos/pull/9820

PR #9823, opened 2025-09-19
feat: project preset

wittjosiah:


packages/plugins/plugin-debug/src/components/SpaceGenerator/presets.ts
```diff
@@ -51,6 +61,75 @@ export const generator = () => ({
   schemas: [CanvasBoardType, FunctionTrigger],
   types: Object.values(PresetName).map((name) => ({ typename: name })),
   items: [
+    [
+      PresetName.ORG_RESEARCH_PROJECT,
+      async (space, n, cb) => {
+        const objects = range(n, () => {
+          const mailboxQueue = space.queues.create();
+          const _contactsQueue = space.queues.create();
```
richburdon: TODO

richburdon (approved):

packages/core/echo/echo-protocol/src/query/ast.ts
```diff
@@ -61,6 +61,13 @@ const FilterIn_ = Schema.Struct({
 export interface FilterIn extends Schema.Schema.Type<typeof FilterIn_> {}
 export const FilterIn: Schema.Schema<FilterIn> = FilterIn_;
 
+const FilterContains_ = Schema.Struct({
```
dmaretskyi, line 64: jsdoc pls

packages/core/echo/echo/src/query/dsl.test.ts
```diff
@@ -90,6 +100,26 @@ describe('query api', () => {
     console.log('PeopleNamedFred', JSON.stringify(PeopleNamedFred.ast, null, 2));
   });
 
+  test('get all people with field of "label" set to "Research"', () => {
+    const PeopleWithFieldLabelSetToResearch = Query.select(
+      Filter.type(Person, { fields: [{ label: 'label', value: 'Research' }] }),
```
dmaretskyi: I dont see `Filter.contains`

wittjosiah: This is currently treating the array as a short-hand for contains. I'm not in love with the syntax, but it's already valid in the inferred type so it seemed like it would be worthwhile to make it do something useful… very open to disagreement on that though

dmaretskyi: I think it's to ambigious -- I could read this as `fields === [{ label: 'label', value: 'Research' }]`

dmaretskyi: I'd push you on using the verbose syntax `Filter.type(Person, { fields: Filter.contains({ label: 'label', value: 'Research' }) })`

wittjosiah: in that case I think we need to update the types of the filter to make that a type error, because currently is not

packages/core/echo/echo/src/query/dsl.ts
```diff
@@ -290,6 +290,12 @@ interface FilterAPI {
    */
   in<T>(...values: T[]): Filter<T>;
 
+  /**
+   * Predicate for an array property to contain the provided value.
+   * @param value - Value to check against.
+   */
+  contains<T>(value: T): Filter<T>;
```
dmaretskyi:
```suggestion
  contains<T>(value: T): Filter<T[]>;
```

packages/core/echo/echo-pipeline/src/filter/filter-match.ts
```diff
@@ -87,6 +88,31 @@ export const filterMatchObject = (filter: QueryAST.Filter, obj: MatchedObject):
...
+const structuralMatch = (filterObj: any, targetObj: any): boolean => {
+  if (typeof filterObj !== 'object' || filterObj === null) {
+    return filterObj === targetObj;
+  }
+
+  if (typeof targetObj !== 'object' || targetObj === null) {
+    return false;
+  }
+
+  return Object.keys(filterObj).every((key) => {
```
dmaretskyi: do we wanna handle extra keys?

dmaretskyi: I would prohibit them by default

dmaretskyi: and then add another filter clause for `object containing`

PR #9830, opened 2025-09-22
feat: Queue query API

dmaretskyi:
Limitations:

- Only simple filtered selections are supported
- Ordering not supported
- Subscriptions are poll-based (every second)

### DIFF
--- a/packages/core/echo/echo-db/src/testing/queues.test.ts
+++ b/packages/core/echo/echo-db/src/testing/queues.test.ts
@@ -4,7 +4,7 @@
 
 import { afterEach, beforeEach, describe, expect, test } from 'vitest';
 
-import { Obj, Relation } from '@dxos/echo';
+import { Filter, Obj, Query, Relation } from '@dxos/echo';
 import { Testing } from '@dxos/echo/testing';
 import { Ref, getSchema } from '@dxos/echo-schema';
 import { DXN, SpaceId } from '@dxos/keys';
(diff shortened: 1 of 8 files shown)


packages/core/echo/echo-pipeline/src/filter/filter-match.ts
```diff
@@ -87,6 +88,82 @@ export const filterMatchObject = (filter: QueryAST.Filter, obj: MatchedObject):
   }
 };
 
+export const filterMatchObjectJSON = (filter: QueryAST.Filter, obj: ObjectJSON): boolean => {
```
wittjosiah, line 91: comment: used for queue filtering

packages/core/echo/echo-protocol/src/query/ast.ts
```diff
@@ -261,7 +261,23 @@ export type Query = Schema.Schema.Type<typeof Query_>;
...
+   * NOTE: Spaces and queues are unioned together if both are specified.
+   */
   spaceIds: Schema.optional(Schema.Array(Schema.String)),
+
+  /**
+   * The nested select statemets will select from the given queues.
+   *
+   * NOTE: Spaces and queues are unioned together if both are specified.
+   */
+  queues: Schema.optional(Schema.Array(DXN.Schema)),
```
wittjosiah, line 276: note: queues are not yet fully supported

packages/core/echo/echo-db/src/testing/queues.test.ts
```diff
@@ -172,4 +173,134 @@ describe('queues', (ctx) => {
       expect(Relation.getTarget(relation as Testing.WorksFor).name).toEqual('jane');
     }
   });
+
+  describe('Query', () => {
```
wittjosiah, line 177: given the memory queue doesn't implement query yet, does that mean that these tests depend on edge or is there some other level of mock?

dmaretskyi: yes, mocked queue service -- memory queue should be killed

PR #9832, opened 2025-09-22
refactor: update error handling with default messages and consistent constructor signatures

richburdon:
First increment to standardize messages.
Removes the need for error instances to specify the the same message each time.


packages/common/errors/src/base.ts
```diff
@@ -21,36 +26,33 @@ export type BaseErrorOptions = {
...
     };
   }
 
   #code: Code;
   #context: Record<string, unknown>;
 
-  constructor(code: Code, message: string, options?: BaseErrorOptions) {
-    super(message, options);
+  constructor(code: Code, options?: BaseErrorOptions) {
+    super(options?.message);
```
dmaretskyi: looses `cause`

packages/common/errors/src/base.ts
```diff
@@ -61,6 +63,11 @@ export class BaseError<Code extends string = string> extends Error {
     return this.#code;
   }
 
+  /** Fallback message. */
+  override get message() {
```
dmaretskyi, line 64: wouldn't this override use-provided message?

packages/core/echo/echo/src/errors.ts
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import { BaseError, type BaseErrorOptions } from '@dxos/errors';
+import { type DXN } from '@dxos/keys';
+
+export class SchemaNotFoundError extends BaseError.extend('SCHEMA_NOT_FOUND', 'Schema not found') {
+  constructor(schema: string, options?: BaseErrorOptions) {
```
dmaretskyi, line 9: I think this kind of context specialization is too restrictive and potentially harmful.
Since context is only read by the developer and not machines there's no need to enfore is to have a certain structure. Think about this as the second parameter to `log` -- we don't wanna have typed log context.

For attaching addition typed metadata to errors, like SchemaNotFoundError always specifying the DXN of the schema, consider adding a field in the derived class:


```
export class SchemaNotFoundError extends BaseError.extend('SCHEMA_NOT_FOUND', 'Schema not found') {
  public readonly schema: string;

  constructor(schema: string) {
      super({ context: { schema } });
       this.schema = schema;
  }
}
```

This field can later be accessed in the error handler.


```
try {
  ...
} catch(err) {
   if(SchemaNotFoundError.is(err)) {
      err.schema; // <---
   }
}

```

packages/core/functions/src/errors.ts
```diff
@@ -2,20 +2,20 @@
...
 
-import { BaseError } from '@dxos/errors';
+import { BaseError, type BaseErrorOptions } from '@dxos/errors';
 
-export class ServiceNotAvailableError extends BaseError.extend('SERVICE_NOT_AVAILABLE') {
-  constructor(serviceName: string) {
-    super(`Service not available: ${serviceName}`);
+export class ServiceNotAvailableError extends BaseError.extend('SERVICE_NOT_AVAILABLE', 'Service not available') {
+  constructor(service: string, options?: Omit<BaseErrorOptions, 'context'>) {
+    super({ context: { service }, ...options });
```
dmaretskyi, line 9: perhaps we should add a helper for this:

```
super(BaseError.mergeOptions({ context: { service } }, options))
```

packages/core/functions/src/errors.ts
```diff
@@ -2,20 +2,20 @@
...
 //
 
-import { BaseError } from '@dxos/errors';
+import { BaseError, type BaseErrorOptions } from '@dxos/errors';
 
-export class ServiceNotAvailableError extends BaseError.extend('SERVICE_NOT_AVAILABLE') {
-  constructor(serviceName: string) {
-    super(`Service not available: ${serviceName}`);
+export class ServiceNotAvailableError extends BaseError.extend('SERVICE_NOT_AVAILABLE', 'Service not available') {
+  constructor(service: string, options?: Omit<BaseErrorOptions, 'context'>) {
```
dmaretskyi, line 8: why prohibit additional context?

packages/common/errors/src/base.ts
richburdon: no it doesn't; although i don't 100% understand why

packages/core/functions/src/errors.ts
richburdon: i was trying to figure out a nice way to not require this but couldn't come up with it.

packages/core/echo/echo/src/errors.ts
richburdon: i think it depends. on the one hand, additional context is good (like with logging), but on the other, if the error is something like Object not found, we want to enforce that the caller provides a DXN (and that the context has a given name -- that could be machine readable)

packages/common/errors/src/base.ts
richburdon: fixed

packages/core/functions/src/errors.ts
richburdon: see above; i guess we could merge additional context.

dmaretskyi (approved):

PR #9858, opened 2025-09-24
feat: Create research notes

dmaretskyi:


packages/core/assistant-testing/src/functions/research/create-research-note.ts
```diff
@@ -0,0 +1,62 @@
...
+import { log } from '@dxos/log';
+import { Markdown } from '@dxos/plugin-markdown/types';
+
+import { ResearchOn } from './research-graph';
+
+export default defineFunction({
+  name: 'dxos.org/function/create-research-note',
+  description: 'Creates a note summarizing the research.',
+  inputSchema: Schema.Struct({
+    target: Schema.String.annotations({
```
dmaretskyi, line 20: I wanna change this to `ArtifactId` or a related type

dmaretskyi: or perhaps it should be a `Ref(x)` type and we optimize that to work with tool parameters so that models can easily fill refs

packages/core/echo/echo/src/Obj.ts
```diff
@@ -31,6 +31,13 @@ export type Obj<Props> = BaseObj & Props;
  */
 export interface Any extends BaseObj {}
 
+export const Any = Schema.Struct({}).pipe(
```
wittjosiah, line 34: will this actually match every object?

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
```diff
@@ -47,7 +49,8 @@ const blueprint = createBlueprint();
 // TODO(dmaretskyi): Consider splitting into multiple modules.
 export default (): Capability<any>[] => [
   contributes(Capabilities.Functions, functions),
-  contributes(Capabilities.Functions, [research]),
+  contributes(Capabilities.Functions, [agent]),
+  contributes(Capabilities.Functions, [research, createResearchNote]),
```
wittjosiah, line 53: is there a reason not to put them in the same array?

packages/core/echo/echo/src/Obj.ts
dmaretskyi: we don't do structural matching on objects. The goal is to define safe base type for "any" object. I initially wanted. to go with `unknown`, but it must be object type, so I decided to go with object that has no fields.

packages/plugins/plugin-assistant/src/capabilities/blueprint-definition.ts
dmaretskyi: they are supposed to be grouped by the blueprint they come from. but im open to changing this

wittjosiah: I don't think it matters too much, just wanted to make sure there wasn't a reason it needed to be that way

PR #9862, opened 2025-09-24
test: run storybook test in CI

wittjosiah:
Resolves DX-304

Approach:
- maintain unified storybook in tools
- setup storybook to be able to run from each package with stories as well
- storybook test should work from the unified storybook locally, but it was failing in CI due to wasm memory issues. It's also not the most useful thing to run tests for every story at once.
- each package with stories includes a storybook vitest project which will run the story tests for just that package. These run in CI successfully and provide better feedback as to which package is causing failures


packages/plugins/plugin-help/src/components/WelcomeTour/WelcomeTour.stories.tsx
```diff
@@ -108,7 +110,8 @@ const meta = {
   title: 'plugins/plugin-help/WelcomeTour',
   component: WelcomeTour,
   render: DefaultStory,
-  decorators: [withTheme],
+  // TODO(wittjosiah): Try to write story which does not depend on plugin manager.
```
thure, line 113: Is this because `withPluginManager` causes issues in CI?

packages/plugins/plugin-markdown/src/components/Toolbar.stories.tsx
```diff
@@ -41,7 +42,7 @@ type StoryProps = {
 
 const DefaultStory = ({ content = '' }: StoryProps) => {
   const { themeMode } = useThemeContext();
-  const [text] = useState(DataType.makeText(content));
+  const [text] = useState(createObject(DataType.makeText(content)));
```
thure, line 45: (Can I get an explainer at some point about the difference between `createObject` and `Obj.make`?)

tools/storybook/.storybook/main.ts
```diff
@@ -21,7 +21,7 @@ const rootDir = resolve(baseDir, '../../');
 const staticDir = resolve(baseDir, './static');
 const iconsDir = resolve(rootDir, 'node_modules/@phosphor-icons/core/assets');
 
-export const packages = resolve(rootDir, 'packages');
+export const packages = '../../../packages';
```
thure: What happened here?

thure (approved):
Fantastic, very much looking forward to this ✨

packages/plugins/plugin-markdown/src/components/Toolbar.stories.tsx
wittjosiah: `createObject` is an internal helper which initializes automerge persistence for the object. It's needed in these stories specifically because the components use automerge helpers internally but the test objects are not stored in a space

packages/plugins/plugin-help/src/components/WelcomeTour/WelcomeTour.stories.tsx
wittjosiah: No, just because of what we've talked about previously, stories for components shouldn't need it, it's intended to help test integration between different plugins, etc. The fact that this is needed in a bunch of stories means that app-framework business is leaking into presentation components which could be pure

tools/storybook/.storybook/main.ts
wittjosiah: storybook test required a relative path to find the stories...however I am wondering if this is always what's causing the issue with the missing styles

packages/plugins/plugin-assistant/vitest.config.ts
```diff
@@ -2,15 +2,20 @@
...
+
+export default defineConfig({
+  test: {
+    ...resolveReporterConfig({ cwd: dirname }),
+    projects: [
+      createNodeProject('jsdom'),
+      createStorybookProject(dirname),
+    ]
+  },
+});
```
dmaretskyi:
```suggestion
export default createConfig({
  directory: import.meta.url,
  node: { env: 'jsdom' },
  storybook: true
});
```

dmaretskyi: since we gonna have a lot of that, I'd prefer a very specific DSL for creating configs, and then extracting all code into the base config.

vitest.base.config.ts
```diff
@@ -43,6 +44,102 @@ export const baseConfig = (options: ConfigOptions): ViteUserConfig => {
   }
 };
```
dmaretskyi, line 47: there's also `vitest.storybook.config.ts` is that still needed?

vitest.base.config.ts
```diff
@@ -43,6 +44,102 @@ export const baseConfig = (options: ConfigOptions): ViteUserConfig => {
...
+                },
+              ],
+            },
+          ],
+        ],
+      }),
+    ],
+  });
+
 const createNodeConfig = (cwd: string) =>
```
dmaretskyi: This will probably work without issues

```suggestion
const createNodeConfig = (cwd: string) =>
defineConfig({
  test: {
    projects: [createNodeProject(...)]
    }
    });
```

vitest.base.config.ts
dmaretskyi: I had an attempt at projects here -- https://github.com/dxos/dxos/pull/9825/files#diff-2286002fbe2ab6e5b1c991e08d3d4ad24072a29a1ee4dd631a9c5722ebfe36f2

It worked but I didn't wanna merge it since it changed the behaviour of `pnpm vite run` -- it would run all projects.

Perhaps a good iteration would be to use projects, but have the `baseConfig` only provide the right projects (nodejs, browser, storybook) based on options and `VITEST_ENV`. This would match existing behaviour while transitioning us to project-based config.

PR #9886, opened 2025-09-25
chore: Update definePlugin

richburdon:


packages/plugins/plugin-assistant/src/stories/testing/testing.tsx
```diff
@@ -125,7 +125,7 @@ export const getDecorators = ({
       GraphPlugin(),
       IntentPlugin(),
       SettingsPlugin(),
-      SpacePlugin(),
+      SpacePlugin({}),
```
wittjosiah, line 128: can we not support default args?

packages/plugins/plugin-client/src/index.ts
```diff
@@ -2,8 +2,9 @@
 // Copyright 2023 DXOS.org
 //
 
-export * from './ClientPlugin';
-export * from './meta';
-
 export { ClientCapabilities } from './capabilities';
 export { ClientEvents } from './events';
+export * from './meta';
```
wittjosiah, line 7: should meta be exported still?

PR #9891, opened 2025-09-25
style: update mailbox UI

richburdon:
- Fixed height inbox grid with balanced padding

## Update

Screenshot 2025-09-26 at 5 38 39 PM


packages/plugins/plugin-inbox/src/components/Calendar/EventtList.tsx
```diff
@@ -23,7 +21,8 @@ export const EventList = ({ events = [], selected, onSelect }: EventListProps) =
         {events.map((event) => (
           <ListItem.Root
             key={event.id}
-            classNames={mx('flex flex-col cursor-pointer', ghostHover, selected === event.id && styles.selected)}
+            // TODO(burdon): Change bg-activeSurface to use data-active
```
thure, line 24: We can just `data-[active="true"]:bg-activeSurface`, conditionally applying tokens on `data-` and `aria-` attributes is already common throughout the repo.

If there are multiple places we can already identify should need this, we could also mint a semantic token which has `data-active` as a condition, but that means minting a new name to remember and use.

packages/plugins/plugin-inbox/src/components/MessageContainer/MessageHeader.tsx
```diff
@@ -71,7 +72,7 @@ export const MessageHeader = ({ message, viewMode, contactDxn }: MessageHeaderPr
         </div>
       </Avatar.Root>
 
-      <div className='grid w-[8rem] _gap-1 justify-items-end'>
+      <div className='grid w-[12rem] justify-items-end'>
```
thure: Would prefer to avoid arbitrary lengths like this, it seems overly magical, and wherever possible it should use extant tokens, e.g. 12rem = `is-48` here.

https://v3.tailwindcss.com/docs/width

packages/plugins/plugin-navtree/src/experimental/Tree.tsx
```diff
@@ -159,13 +150,9 @@ const ItemIcon = ({
...
 }: Pick<TreeNodeProps, 'node'>) => {
   return (
-    (iconName && (
-      <IconButton
-        iconName={iconName}
-        classNames={color ?? 'text-neutral-700 dark:text-neutral-300'}
-        weight={color ? 'duotone' : 'regular'}
-      />
-    )) || <div />
+    (iconName && <IconButton iconName={iconName} classNames={color ?? 'text-neutral-700 dark:text-neutral-300'} />) || (
```
thure: We have a semantic token for this, `text-subdued`.
```suggestion
    (iconName && <IconButton iconName={iconName} classNames={color ?? 'text-subdued'} />) || (
```

packages/plugins/plugin-inbox/src/components/Mailbox/mailbox.css
```diff
@@ -10,7 +10,7 @@
   & > .dx-grid__cell__content,
   &.message__card {
     /** TODO(burdon): Reconcile with styles.ts (mailboxGrid) */
-    @apply grid grid-cols-[3rem_1fr] gap-1 border-0;
+    @apply grid grid-cols-[3rem_1fr] gap-1 p-1 border-0;
```
thure: Please be sure you’ve confirmed this is the desired padding in all the places it applies, including both regular `Mailbox` as well as a `ProjectContainer` with messages.

thure:
> how can we make the grid row and header use common defs
I would still assert it is a coincidence these look similar, and instead of cause them to look the same we should cause them to have better differentiation so that the header can’t be confused with the list items since they serve radically different functions.

packages/plugins/plugin-navtree/src/experimental/Tree.tsx
richburdon: right. some of this is really old.

packages/plugins/plugin-inbox/src/components/Mailbox/mailbox.css
richburdon: now removed

packages/plugins/plugin-inbox/src/components/Mailbox/Mailbox.tsx
```diff
@@ -59,42 +59,48 @@ const renderMessageCell = (message: DataType.Message, now: Date, _isCurrent?: bo
...
-        ? `<div class="message__tag-row">
-            ${message.properties.tags.map((tag: Tag) => `<div class="dx-tag message__tag-row__item" data-label="${tag.label}" data-hue=${tag.hue}>${tag?.label}</div>`).join('')}
-          </div>`
-        : ''
-    }
-    </button>`;
+      <div class="message__abstract__heading">
+        <div class="message__abstract__from">${from}</div>
+        <div class="message__abstract__date">${date}</div>
+      </div>
```
thure, line 65: Why change these from semantic tags to nonsemantic `div`s? Doing so reduces the accessibility of these components.

packages/plugins/plugin-inbox/src/components/Mailbox/Mailbox.tsx
```diff
@@ -140,6 +137,33 @@ export const Mailbox = ({ messages, id, currentMessageId, onAction, ignoreAttent
     [onAction],
   );
 
+  const handleKeyUp = useCallback<NonNullable<GridContentProps['onKeyUp']>>(
```
thure, line 140: Why bind to `keyUp` and not `keyDown`? Usually `keyDown` is preferred – please leave a note comment if this is a workaround for something.

packages/plugins/plugin-inbox/src/components/MessageContainer/MessageHeader.tsx
```diff
@@ -41,16 +39,17 @@ export const MessageHeader = ({ message, viewMode, contactDxn }: MessageHeaderPr
...
     ? { variant: 'ghost', classNames: 'pli-2 gap-2 text-start', onClick: handleSenderClick }
     : { className: 'p-0 hover:bg-transparent', 'data-variant': 'ghost' };
 
   return (
-    <div className='flex border-be border-separator'>
+    <div className='flex border-be border-subduedSeparator'>
       <Avatar.Root>
-        <div className={mx(mailboxGrid, 'p-[1px] is-full')}>
-          <SenderRoot {...(senderProps as any)}>
+        <div className={mx('is-full p-2')}>
```
thure:
```suggestion
        <div className='is-full p-2'>
```

packages/plugins/plugin-inbox/src/components/MessageContainer/MessageHeader.tsx
```diff
@@ -60,30 +59,33 @@ export const MessageHeader = ({ message, viewMode, contactDxn }: MessageHeaderPr
                 size={10}
               />
             </div>
-          </SenderRoot>
+          </SenderRoot> */}
           <div role='none'>
             <Avatar.Label classNames='flex items-center gap-1'>
-              <h3 className='truncate'>{message.sender.name || 'Unknown'}</h3>
+              <h3 className='text-lg truncate'>{message.sender.name || 'Unknown'}</h3>
```
thure: Lovely ✨

packages/plugins/plugin-inbox/src/components/Mailbox/Mailbox.tsx
richburdon: i was copying a precedent from half a dozen other places

PR #9905, opened 2025-09-29
experiment: Use regular instead of `-visible` focus selectors

thure:


richburdon:
thanks -- this is good.
i'd like to merge this and propose variance to this as the base/default case.

PR #9915, opened 2025-10-01
feat!: LMM memoization & changing tool call actor

dmaretskyi:


packages/core/assistant-testing/src/functions/research/research.test.ts
```diff
@@ -63,11 +63,11 @@ const TestLayer = Layer.mergeAll(
   ),
 );
 
-describe('Research', { timeout: 600_000 }, () => {
+describe.skip('Research', { timeout: 600_000 }, () => {
```
wittjosiah: timeouts conditional on whether its generating a new conversation or not would be good I think, that way if this is running in CI it doesn't have a 5min timeout

dmaretskyi: 👍

PR #9918, opened 2025-10-01
feat: Query DSL

richburdon:
https://github.com/user-attachments/assets/3763a384-7129-41d9-ad87-52c8892defd8


packages/plugins/plugin-assistant/src/components/QueryEditor/query.test.ts
```diff
@@ -0,0 +1,254 @@
...
+          'PropertyKey',
+          'Identifier',
+          ':',
+          'Value',
+          'String',
+          '}',
+        ],
+      },
+      {
+        query: '(type:dxos.org/echo/Contact) => (type:dxos.org/echo/Organization)',
```
wittjosiah: not sure how I should interpret this, which relation does this traverse? what about traversing reference fields?

richburdon: i haven't figured out a representation for refs or constrained relations yet

packages/core/echo/echo-query/moon.yml
```diff
@@ -9,3 +9,10 @@ tasks:
     args:
       - '--entryPoint=src/index.ts'
       - '--injectGlobals'
+  prebuild:
+    script: pnpm run build:parser
```
dmaretskyi: I'd just put the command into moon to not spread out the config

```suggestion
    script: lezer-generator --typeScript ./src/parser/query.grammar -o ./src/parser/gen/query.ts
```

packages/core/echo/echo-query/src/parser/query.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    type Test = { input: string; expected: Filter.Any };
+    const tests: Test[] = [
+      {
+        input: 'type:dxos.org/type/Contact',
+        expected: Filter.typename('dxos.org/type/Contact'),
+      },
+      {
+        input: '{ name: "DXOS" }',
+        expected: Filter._props({ name: 'DXOS' }),
```
dmaretskyi: todo to make `Filter.props` not private

packages/core/echo/echo-query/src/parser/query.test.ts
```diff
@@ -0,0 +1,270 @@
...
+          'PropertyKey',
+          'Identifier',
+          ':',
+          'Value',
+          'String',
+          '}',
+        ],
+      },
+      {
+        query: 'type:dxos.org/type/Contact => type:dxos.org/type/Organization',
```
dmaretskyi: do you wanna discuss how to make relations translate into the query AST?

dmaretskyi: How would we express the nature of the traversal -- currently we have 6 kinds:

- follow a reference: `obj -(.ref) -> another_obj`
- another object references us: `obj <-(.ref)- another_obj`
- go for object to relation (via source): `obj =>[Relation]`
- go from obj to relation (via target): `obj <=[Relation]`
- go from relation to its source: `[Relation]<= obj`
- go from relation to its target: `[Relation] => obj`

As you see, traversing the relation is done in two steps: first you go to the relation, and then to one of the objects it references. This is done primarly so that you can build queries that return relation objects. If this is not needed for this DSL, we can simplify and alway do the full hop to the other side. In this case we'd only be left with 2 kinds (times 2 directions):

- reference forwards: `obj -(.ref)-> another_obj`
- reference backwards: `obj <-(.ref)- another_obj`
- relation forwards: `obj -(Relation)-> another_obj`
- relation backwards: `obj <-(Relation)- another_obj`

dmaretskyi (changes requested):

packages/plugins/plugin-assistant/src/components/QueryEditor/query-extension.ts
```diff
@@ -0,0 +1,335 @@
...
+ */
+const queryHighlightStyle = HighlightStyle.define([
+  { tag: t.keyword, class: 'text-blue-500' },
+  { tag: t.string, class: 'text-orange-500' },
+  { tag: t.number, class: 'text-green-500' },
+  { tag: t.bool, class: 'text-green-500' },
+  { tag: t.null, class: 'text-neutral-500' },
+  { tag: t.attributeName, class: 'text-blue-500' },
+  { tag: t.variableName, class: 'text-teal-500' },
+  { tag: t.propertyName, class: 'text-teal-500' },
```
thure: We have semantic tokens for these:
```suggestion
  { tag: t.keyword, class: 'text-blueText' },
  { tag: t.string, class: 'text-orangeText' },
  { tag: t.number, class: 'text-greenText' },
  { tag: t.bool, class: 'text-greenText' },
  { tag: t.null, class: 'text-neutralText' },
  { tag: t.attributeName, class: 'text-blueText' },
  { tag: t.variableName, class: 'text-tealText' },
  { tag: t.propertyName, class: 'text-tealText' },
```

packages/plugins/plugin-assistant/src/components/QueryEditor/query-extension.ts
```diff
@@ -0,0 +1,335 @@
...
+  { tag: t.definitionOperator, class: 'text-subdued' },
+  { tag: t.separator, class: 'text-subdued' },
+  { tag: t.derefOperator, class: 'text-subdued' },
+  { tag: t.brace, class: 'text-subdued' },
+  { tag: t.squareBracket, class: 'text-subdued' },
+  { tag: t.operator, class: 'text-subdued' },
+  { tag: t.paren, class: 'text-yellow-500' },
```
thure: Ditto, though might I suggest `amber` which will look more attractive when dark:
```suggestion
  { tag: t.paren, class: 'text-amberText' },
```

packages/plugins/plugin-project/src/components/Project.stories.tsx
```diff
@@ -152,7 +152,7 @@ const meta: Meta<typeof Project> = {
       types: [DataType.Project, DataType.View, DataType.Collection, DataType.Person],
       createIdentity: true,
       createSpace: true,
-      onSpaceCreated: async ({ space }) => {
+      onCreateSpace: async ({ space }) => {
```
thure, line 155: The norm from Radix is `on{Noun}{Event}` e.g. `onValueChange`, fwiw, though I don’t have a strong opinion in this context, there are just a lot of places in the repo that adopt Radix’s naming patterns.

packages/ui/react-ui-components/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -31,9 +31,16 @@ export type QueryEditorProps = {
...
+<<<<<<< Updated upstream
+/**
+ * @param param0 @deprecated
+ */
+||||||| Stash base
 // TODO(wittjosiah): Factor out.
 // TODO(thure): Reconcile with `react-ui-query-editor`.
 
+=======
+>>>>>>> Stashed changes
```
thure: Merge conflict?

packages/ui/react-ui-editor/src/styles/theme.ts
```diff
@@ -156,7 +156,7 @@ export const defaultTheme: ThemeStyles = {
    * Tooltip.
    */
   '.cm-tooltip': {
-    background: 'var(--dx-modalSurface)',
+    background: 'var(--dx-baseSurface)',
```
thure, line 159: Why do this? Our menus and tooltips use the `modalSurface` token, shouldn’t this match?

thure (commented):
Just some physical tokens to switch to semantic and what looks like a merge conflict, then a couple questions, otherwise looks good to me.

packages/core/echo/echo-query/src/parser/query.test.ts
```diff
@@ -0,0 +1,268 @@
...
+      } while (cursor.next());
+      expect(result).toEqual(expected);
+    }
+  });
+
+  it('should build a query', ({ expect }) => {
+    const queryBuilder = new QueryBuilder();
+
+    type Test = { input: string; expected: Filter.Any };
+    const tests: Test[] = [
```
dmaretskyi, line 237: Use [`test.each`](https://vitest.dev/api/#test-each)

packages/ui/react-ui-editor/src/styles/theme.ts
richburdon: ok i'm still confused by the surfaces (my bad). and i tried modal and it just looked washed out. but let's discuss -- i need your help to make this all consistent.

thure: Maybe CM's tooltip should diverge anyway since the design of it is quite different, I would be amenable to that argument too.

packages/plugins/plugin-project/src/components/Project.stories.tsx
richburdon: +1; I was looking for a logic to follow and I like the Radix approach. I'll revert.

packages/core/echo/echo-query/src/parser/query.test.ts
richburdon: +1

packages/plugins/plugin-assistant/src/components/QueryEditor/query-extension.ts
richburdon: +1

packages/core/echo/echo-query/src/parser/query.test.ts
richburdon: OK Great -- thanks

PR #9948, opened 2025-10-06
feat: Support text and tag filters

richburdon:
- Currently assumes tags are stored at `properties.tags`


packages/ui/react-ui-editor/src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -20,7 +20,7 @@ import { type MaybePromise } from '@dxos/util';
 
 import { commandRangeEffect } from '../../extensions';
 
-// TODO(burdon): Rename and factor out to separate package.
+// TODO(burdon): Rename ComboBox (and factor out to separate package).
```
thure, line 23: fwiw, the experience might be functionally identical to a combobox, but there isn’t a good drop-in replacement for any of this component except for the list of items which could come from `Listbox` – everything else is reasonably specific to anchoring the menu to the command character in the CM instance and calling the search callback using document updates after the character.

PR #9951, opened 2025-10-07
feat: Support ECHO queries for inbox

richburdon:
- [x] Uses QueryEditor (with DSL/grammar) and ECHO filters for inbox.

Screenshot 2025-10-06 at 8 20 47 PM


packages/core/echo/echo-pipeline/src/filter/filter-match.ts
```diff
@@ -148,6 +154,18 @@ export const filterMatchObjectJSON = (filter: QueryAST.Filter, obj: ObjectJSON):
       return true;
     }
 
+    case 'tag': {
```
dmaretskyi, line 156: I missed the original change:

```ts
/**
 * Tag.
 */
const FilterTag_ = Schema.Struct({
  type: Schema.Literal('tag'),
  tag: Schema.String, // TODO(burdon): Make OR-collection?
});
export interface FilterTag extends Schema.Schema.Type<typeof FilterTag_> {}
export const FilterTag: Schema.Schema<FilterTag> = FilterTag_;
```

Are you going back on what we've discussed before of doing tags via relations?

packages/core/echo/echo-db/src/query/api.ts
```diff
@@ -21,6 +21,13 @@ export interface QueryFn {
   <F extends Filter.Any>(filter: F, options?: QueryOptions | undefined): QueryResult<Live<Filter.Type<F>>>;
 }
 
+/**
+ * Common interface for Echo and Queue.
+ */
+export interface Queryable {
```
dmaretskyi, line 27: :+1:

dmaretskyi (dismissed):

packages/core/echo/echo-pipeline/src/filter/filter-match.ts
thure: @dmaretskyi I think this is meant as a stopgap to be refactored to make tags first-class entities with relations in a follow-up.

PR #9954, opened 2025-10-07
feat: Add dynamic blueprints

mykola-vrmchk:
- **Add live object inside registry**
- **Make script registry respect database**
- **Minor improvements**

### DIFF
--- a/packages/core/blueprints/src/blueprint/registry.ts
+++ b/packages/core/blueprints/src/blueprint/registry.ts
@@ -31,7 +31,7 @@ export class Registry extends Resource {
   protected override async _open(): Promise<void> {
     this._reconcileBlueprints = new DeferredTask(this._ctx, async () => {
       const seen = new Set<string>();
-      const newBlueprints = [];
+      const newBlueprints: Blueprint[] = [];
       for (const blueprint of [...this._staticBlueprints, ...this._databaseBlueprints]) {
         if (seen.has(blueprint.key)) {
           continue;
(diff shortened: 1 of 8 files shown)


packages/core/blueprints/src/blueprint/registry.ts
```diff
@@ -2,35 +2,72 @@
...
-export class Registry {
-  private readonly _blueprints: Blueprint[] = [];
-
-  constructor(blueprints: Blueprint[]) {
-    const seen = new Set<string>();
-    blueprints.forEach((blueprint) => {
-      if (seen.has(blueprint.key)) {
-        log.warn('duplicate blueprint', { key: blueprint.key });
-      } else {
+export class Registry extends Resource {
```
dmaretskyi: keep registry as is -- an API for a **remote** blueprint registry.

It would be easiest to handle local BPs in the component that renders the selection. Perhaps even extract it to a hook:

```ts
const useBlueprints = (registry?) => {
const registryBps = registry?.blueprints ?? []; 
const spaceBps = useQuery(space, Filter.type(Blueprint));

const allBlueprints = Array.unionWith(registryBps, spaceBps, (a, b) => a.key === b.key)
return allBlueprints
}
```

dmaretskyi (changes requested):
see comments

packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatOptions.tsx
```diff
@@ -98,7 +99,11 @@ const BlueprintsPanel = ({
 }: Pick<ChatOptionsProps, 'blueprintRegistry' | 'space' | 'context'>) => {
   const { t } = useTranslation(meta.id);
 
-  const blueprints = useBlueprints({ blueprintRegistry });
+  const staticBlueprints = useBlueprints({ blueprintRegistry });
+  const spaceBlueprints = useQuery(space, Filter.type(Blueprint.Blueprint));
+  const blueprints = distinctBy([...staticBlueprints, ...spaceBlueprints], (b) => b.key);
+  blueprints.sort(({ name: a }, { name: b }) => a.localeCompare(b));
```
dmaretskyi:
```suggestion
  const blueprints = useBlueprints({ blueprintRegistry, space });
```

dmaretskyi (approved):

PR #9957, opened 2025-10-08
refactor: Merge `CreateRefPanel` into `RefField`

thure:
This PR:
- begins a resolution of DX-543
- promotes `CreateRefPanel` into an ephemeral state of `RefField` so that this can be used outside of Table
- writes a `play` function for verifying this works as designed

https://github.com/user-attachments/assets/19e67dad-5e06-4e87-9f93-ac642e986039


packages/ui/react-ui-form/src/components/Form/FormContent.tsx
```diff
@@ -33,7 +33,13 @@ export type FormFieldProps = {
   Custom?: Partial<Record<string, InputComponent>>;
 } & Pick<
   RefFieldProps,
-  'readonly' | 'onQueryRefOptions' | 'createOptionLabel' | 'createOptionIcon' | 'onCreateFromQuery'
+  | 'readonly'
+  | 'onQueryRefOptions'
+  | 'createOptionLabel'
+  | 'createOptionIcon'
+  | 'onCreate'
```
richburdon, line 40: nit: group callbacks at end?

packages/ui/react-ui-form/src/components/Form/RefField.tsx
```diff
@@ -66,6 +72,54 @@ export const RefField = ({
   );
   const { options: availableOptions, loading: _loading } = useQueryRefOptions({ refTypeInfo, onQueryRefOptions });
 
+  const [showForm, setShowForm] = useState(false);
```
richburdon, line 75: perhaps time to rename this component, and promoted it to its own folder?
it's not quite a general purpose ComboBox, but "RefField" isn't an intuitive name to me.

ObjectSelector?
ObjectPicker?

e.g., it's not necessarily just for setting a "field" since we want to use it to select objects to insert into other components (board). 

i think perhaps we can further dry this out and move it into a new package along with QueryEditor (which should also be renamed). OR we move QueryEditor here? if we think these are all just "form" components?

richburdon (approved):

packages/plugins/plugin-table/src/components/TableContainer.tsx
```diff
@@ -111,6 +112,14 @@ export const TableContainer = ({ role, view }: TableContainerProps) => {
     tableRef.current?.update?.();
   }, []);
 
+  const handleCreate = useCallback(
+    (schema: Schema.Schema.AnyNoContext, values: any) => {
+      invariant(space);
+      return space.db.add(live(schema, values));
```
dmaretskyi: prefer `Obj.make` instead of `live`.

I'm working on clearly separting database object from plain reactive objects, I think how they are currently entangled is confusing.

`live` - object where every field has reactivity via signals. nothing more.
`Obj.make` database object with reactivity, schema, type, metadata, etc.

dmaretskyi (changes requested):

packages/ui/react-ui-form/src/components/Form/RefField.tsx
thure: Right, that is my intention in the follow-up. It was a massive lift to do away with CreateRefPanel, I wanted to be sure this one move could be accomplished without breaking important features before making further changes.

packages/ui/react-ui-form/src/components/Form/FormContent.tsx
thure: I am not going to be able to consistently stick to such a rule without a tool, I recommend we use `jsx-sort-props`: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md#rule-options

packages/plugins/plugin-table/src/components/TableContainer.tsx
thure: Right, I’d initially used `Obj.make` for an earlier version of this, but ran into an issue, so reverted to `live` which this component was using before this PR, but I don’t encounter that issue with the latest changes.

PR #9964, opened 2025-10-08
fix: improve filter UI with overflow handling and auto-show on tag se…

richburdon:
…lection

### DIFF
--- a/packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
+++ b/packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
@@ -71,7 +71,7 @@ export const MailboxContainer = ({ mailbox, role }: MailboxContainerProps) => {
         }
         case 'select-tag': {
           // TODO(burdon): Check if tag already exists.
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
```diff
@@ -109,15 +109,16 @@ export const MailboxContainer = ({ mailbox, role }: MailboxContainerProps) => {
...
-          <QueryEditor
-            ref={queryEditorRef}
-            autoFocus
-            classNames='grow'
-            space={getSpace(mailbox)}
-            value={queryText}
-            onChange={setQueryText}
-          />
+        <div role='none' className='flex is-full overflow-hidden items-center p-1 gap-1 border-be border-separator'>
+          <div className='grow overflow-hidden'>
```
thure: This many interceding `div`s seems like a smell — `QueryEditor` should behave like `TextInput` wrt layout so it doesn’t need so many interventions for common use-cases like this, I’d imagine.

PR #9965, opened 2025-10-08
feat: use `ObjectPicker` in `BoardContainer`

thure:
This PR:
- resolves DX-543

https://github.com/user-attachments/assets/50f5441f-5c3f-436c-b991-355a15a88e27

### DIFF
--- a/packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.stories.tsx
+++ b/packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.stories.tsx
@@ -3,6 +3,7 @@
 //
 
 import { type Meta, type StoryObj } from '@storybook/react-vite';
+import { Schema } from 'effect';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { expect, fn, userEvent, within } from 'storybook/test';
 
@@ -25,6 +26,9 @@ const createPerson = () =>
   Obj.make(DataType.Person, {
     fullName: faker.person.fullName(),
   });
+
+const omitId = Schema.omit<any, any, ['id']>('id');
+const personSchema = omitId(DataType.Person);
 
 // Mock functions for testing
 const mockHandleSelect = fn();
@@ -61,6 +65,7 @@ const DefaultStory = () => {
 
   const handleCreateCallback = useCallback(
     (values: any) => {
+      console.log('[on create]', values);
       if (!space) return;
       const newPerson = space.db.add(Obj.make(DataType.Person, values));
       mockHandleCreate(values);
@@ -80,7 +85,7 @@ const DefaultStory = () => {
         <ObjectPicker.Content
           options={options}
           onSelect={mockHandleSelect}
-          createSchema={DataType.Person}
+          createSchema={personSchema}
           createOptionLabel={['create new person label', { ns: 'os' }]}
           createOptionIcon='ph--user-plus--regular'
           createInitialValuePath='fullName'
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-board/src/components/BoardContainer.tsx
```diff
@@ -21,29 +21,61 @@ export type BoardContainerProps = {
...
+      });
+    },
+    [board],
+  );
+
+  const handleSelect = useCallback(
+    (id: string) => {
+      if (!pickerState) return;
+
+      // Find the selected object by id from the space
```
richburdon: nit: our standard is to end comments with a period (to guard against comments getting clipped)

packages/plugins/plugin-board/src/components/BoardContainer.tsx
```diff
@@ -21,29 +21,61 @@ export type BoardContainerProps = {
...
+      const selectedObject = allObjects.find((obj) => obj.id === id);
+      if (!selectedObject) return;
```
richburdon, line 66: did we relax the need for braces? i didn't notice that.

packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.tsx
```diff
@@ -0,0 +1,161 @@
...
+  onSelect: (id: string) => void;
+  selectedIds?: string[];
+  createSchema?: Schema.Schema.AnyNoContext;
+  createOptionLabel?: [string, { ns: string }];
+  createOptionIcon?: string;
+  createInitialValuePath?: string;
+  onCreate?: (values: any) => void;
+}>;
+
+const ObjectPickerContent = React.forwardRef<HTMLDivElement, ObjectPickerContentProps>(
```
richburdon, line 26: we really ought to have a story for this coponent

packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { Icon, Popover, type ThemedClassName, useTranslation } from '@dxos/react-ui';
+import { PopoverCombobox } from '@dxos/react-ui-searchlist';
+
+import { type QueryTag } from '../../hooks';
+import { translationKey } from '../../translations';
+import { Form } from '../Form';
+
+export type ObjectPickerContentProps = ThemedClassName<{
+  options: QueryTag[];
+  onSelect: (id: string) => void;
```
richburdon, line 17: group callbacks
(i have a PR pending with the lint change)

packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.tsx
```diff
@@ -0,0 +1,161 @@
...
+  selectedIds?: string[];
+  createSchema?: Schema.Schema.AnyNoContext;
```
richburdon, line 19: perhaps group create options?

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: Interesting, I actually didn’t think so, but I don’t see any lint errors…

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: Right, Junie keeps ignoring that guideline. Will fix.

packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.tsx
thure: Yes, currently this component is demonstrated & tested by `Relations.stories.tsx`, but now that it’s on its own it can have its own story, will look into that.

packages/plugins/plugin-board/src/components/BoardContainer.tsx
thure: fwiw, I find for early returns this is more legible since declaring a block to me sets some intention that the block will scale.

packages/ui/react-ui-form/src/components/ObjectPicker/ObjectPicker.tsx
thure: They are grouped here.

PR #9967, opened 2025-10-09
refactor: store Gmail labels as tags with IDs in mailbox object

richburdon:
- [x] Sync/store tags in meta.tags
- [x] Sync/store mailbox tags on mailbox object (mailbox.properties)
- [x] Autocomplete tag names from inbox tags
- [x] Covert tag name to tag ID
  
## Demo

Screenshot 2025-10-09 at 8 09 38 PM

### DIFF
--- a/packages/core/echo/echo-query/src/parser/query.test.ts
+++ b/packages/core/echo/echo-query/src/parser/query.test.ts
@@ -5,7 +5,7 @@
 import { type Tree } from '@lezer/common';
 import { describe, it } from 'vitest';
 
-import { Filter } from '@dxos/echo';
+import { Filter, Tag } from '@dxos/echo';
 
 import { QueryDSL } from './gen';
 import { QueryBuilder } from './query-builder';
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-inbox/src/testing/data.ts
```diff
@@ -38,7 +41,10 @@ export const createMessages = (count = 10) => {
         properties: {
           subject: faker.helpers.arrayElement(['', 'Re: ']) + faker.lorem.sentence(8),
           snippet: text,
-          tags: faker.helpers.uniqueArray(TAGS, faker.number.int(3)).sort(sortTags),
+          tags: faker.helpers.uniqueArray(
```
dmaretskyi, line 49: move to meta after https://github.com/dxos/dxos/pull/9968 lands

packages/plugins/plugin-inbox/src/testing/data.ts
```diff
@@ -3,22 +3,25 @@
...
 
-const TAGS: Tag[] = [
-  { label: 'important', hue: 'green' },
-  { label: 'investor', hue: 'purple' },
-  { label: 'team', hue: 'green' },
-  { label: 'eng', hue: 'blue' },
-  { label: 'work', hue: 'emerald' },
-  { label: 'personal', hue: 'pink' },
+export const TAGS: (Tag & WithId)[] = [
+  { id: 'tag_1', label: 'important', hue: 'green' },
```
dmaretskyi, line 15: separate id and label will have complications related to queries: 

you would have to write `Filter.tag("tag_1")` (using ID instead of the label)

richburdon: yes, that's right

PR #9972, opened 2025-10-10
refactor: tag colors

richburdon:


packages/ui/react-ui-components/src/components/QueryEditor/query-extension.ts
```diff
@@ -284,7 +286,10 @@ class TypeWidget extends WidgetType {
  * Tag
  */
 class TagWidget extends WidgetType {
-  constructor(private readonly _str: string) {
+  constructor(
+    private readonly _str: string,
+    private readonly _color?: Color,
```
thure: Replace with `hue`? Maybe leave a todo if that’s too much for this PR.

PR #9973, opened 2025-10-10
fix: handle undefined filters and tags in mailbox sync and UI

richburdon:


packages/plugins/plugin-inbox/src/functions/email/sync.ts
```diff
@@ -43,7 +43,7 @@ export default defineFunction({
       // Sync labels.
       const { labels } = yield* listLabels(userId);
       labels.forEach((label) => {
-        mailbox.tags[label.id] = { label: label.name };
+        (mailbox.tags ??= {})[label.id] = { label: label.name };
```
richburdon, line 46: @wittjosiah perhaps we should require a "repair" function for each top-level plugin data-type that is called on each action invocation?

PR #9974, opened 2025-10-10
chore: Remove unused deps with knip

dmaretskyi:


packages/common/graph/package.json
```diff
@@ -27,25 +27,14 @@
...
-    "lodash.defaultsdeep": "^4.6.1"
-  },
-  "devDependencies": {
-    "@antv/graphlib": "^2.0.4",
-    "@antv/layout": "^1.2.13",
-    "@dxos/echo-db": "workspace:*",
-    "@dxos/echo-schema": "workspace:*",
-    "@dxos/random": "workspace:*",
-    "@types/lodash.defaultsdeep": "^4.6.6"
+    "effect": "3.18.3"
```
wittjosiah, line 35: we should make sure this is in dev deps everywhere rather than deps

PR #9976, opened 2025-10-10
refactor: Use FunctionInvocationService inside sheet compute graph

mykola-vrmchk:
- **Add type**
- **Add runtime to context**
- **Add runtime propagation**
- **Refactor function invocation**
- **Make compute graph require function service runtime**
- **Add test**
- **Fix default plugin**
- **Fix**

### DIFF
--- a/packages/plugins/plugin-automation/src/events.ts
+++ b/packages/plugins/plugin-automation/src/events.ts
@@ -9,5 +9,3 @@ import { meta } from './meta';
 export namespace AutomationEvents {
   export const ComputeRuntimeReady = defineEvent(`${meta.id}/event/compute-runtime-ready`);
 }
-
-


packages/core/compute/src/testing/test-runtime.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Minimal compute runtime provider for tests and Storybook.
+ * Will cause errors if used for actual runtime.
+ */
+export const createMockedComputeRuntimeProvider = ({
+  functions,
+}: { functions?: FunctionDefinition<any, any>[] } = {}): AutomationCapabilities.ComputeRuntimeProvider => {
+  return {
+    getRuntime: (_spaceId: SpaceId) => ManagedRuntime.make(FunctionInvocationService.layerTest({ functions }) as any),
```
dmaretskyi: no `as any`

PR #9979, opened 2025-10-10
test: Add production function test

mykola-vrmchk:


packages/plugins/plugin-script/src/e2e/deploy.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    await client.initialize();
+    await client.halo.createIdentity();
+
+    const space = await client.spaces.create();
+    await space.waitUntilReady();
+
+    // Load the echo template source.
+    const template = templates.find((t) => t.id === 'dxos.org/script/echo');
+    invariant(template?.source, 'scriptSource is required');
+    const source = Obj.make(DataType.Text, { content: template.source });
```
dmaretskyi: Can be one line with db.add

packages/plugins/plugin-script/src/e2e/deploy.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(success, error?.message).toBe(true);
+    expect(functionId).toBeDefined();
+
+    // Invoke deployed function via EDGE directly.
+    const edgeClient = client.edge;
+    invariant(edgeClient, 'edgeClient is required');
+    edgeClient.setIdentity(createEdgeIdentity(client));
+
+    // functionId may include a leading '/', strip it for invoke endpoint.
+    const cleanedId = (functionId ?? '').replace(/^\//, '');
```
dmaretskyi: No reason to default to empty string since test would fail if functionId was undefined

packages/core/functions/moon.yml
```diff
@@ -11,3 +11,11 @@ tasks:
       - '--entryPoint=src/edge/index.ts'
       - '--entryPoint=src/index.ts'
       - '--entryPoint=src/testing/index.ts'
+  e2e:
+    command: pnpm vitest run -c vitest.config.ts --reporter=verbose src/e2e/deploy.test.ts
```
dmaretskyi:
```suggestion
    command: vitest run -c vitest.config.ts --reporter=verbose src/e2e/deploy.test.ts
```

packages/core/functions/src/e2e/deploy.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    // Invoke deployed function via EDGE directly.
+    const edgeClient = client.edge;
+    invariant(edgeClient, 'edgeClient is required');
+    edgeClient.setIdentity(createEdgeIdentity(client));
+
+    const input = { from: 'USD', to: 'EUR' };
+    const result = await edgeClient.invokeFunction({ functionId }, input);
+    log.info('>>> result', { result, functionId });
+    // const resultNumber = Number(result);
```
dmaretskyi: add assertions back in?

dmaretskyi (approved):

packages/plugins/plugin-script/src/util/deploy.ts
```diff
@@ -48,7 +49,7 @@ export const deployScript = async ({
   }
 
   try {
-    const bundler = new Bundler({ platform: 'browser', sandboxedModules: [], remoteModules: {} });
+    const bundler = new Bundler({ platform: isNode() ? 'node' : 'browser', sandboxedModules: [], remoteModules: {} });
```
wittjosiah, line 47: isn't this bundling for workers runtime? I thought you weren't supposed to use node platform for that?

mykola-vrmchk: I want to unify APIs for bunding, and this path is now used in node tests. We need to make it default for every function bundle path

PR #9982, opened 2025-10-10
feat: reverse messages order and add sort button to mailbox toolbar

richburdon:


packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
```diff
@@ -41,23 +41,57 @@ export const MailboxContainer = ({ mailbox, role, attendableId, filter: filterPa
 
   const filterEditorRef = useRef<EditorController>(null);
   const saveFilterButtonRef = useRef<HTMLButtonElement>(null);
+  const [ascending, setAscending] = useState(true);
   const [filterVisible, setFilterVisible] = useState(false);
 
   const [filterText, setFilterText] = useState<string>(filterParam ?? '');
   const [filter, setFilter] = useState<Filter.Any | null>(null);
   const messages: DataType.Message[] = useQuery(mailbox.queue.target, filter ?? Filter.everything());
+  const sortedMessages = useMemo(() => (ascending ? messages : [...messages.reverse()]), [messages, ascending]);
```
wittjosiah, line 50: could you try using `Order` from @dxos/echo? if it doesn't work, we should fix it

packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
```diff
@@ -41,23 +41,57 @@ export const MailboxContainer = ({ mailbox, role, attendableId, filter: filterPa
...
+
   const parser = useMemo(() => new QueryBuilder(mailbox.tags), []);
   useEffect(() => {
     setFilter(parser.build(filterText));
   }, [filterText]);
 
-  const actions = useActions(setFilterVisible);
+  // TODO(burdon): Reactivity?
```
wittjosiah, line 57: what's the issue w/ reactivity? why did this get brought back into the component?

richburdon: we're combining RX and react and yet the hook would require a third kind of reactivity in order to pass in state and setters for the menu state? (e.g., sort indicator)

packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
richburdon: i don't see a test/story for Order with useQuery or Filter -- it seems to only work with Query objects?

PR #9987, opened 2025-10-10
fix: Sort messages

richburdon:


packages/plugins/plugin-inbox/src/components/Mailbox/MailboxContainer.tsx
```diff
@@ -216,11 +225,11 @@ const useMailboxActions = ({
               icon: 'ph--magnifying-glass--regular',
               label: ['mailbox toolbar filter', { ns: meta.id }],
             },
-            () => onFilterToggle(true),
+            () => context.set(filterVisible, !context.get(filterVisible)),
           )
           .build(),
       ),
-    [ascending],
+    [],
```
wittjosiah: the rx shouldn't change, but should still be in the dependency array

PR #9993, opened 2025-10-13
feat: implement basic code evaluation in TypescriptEditor with error handling

richburdon:


packages/core/echo/echo-query/src/sandbox/query-sandbox.ts
```diff
@@ -16,15 +17,24 @@ import { unwrapResult } from './quickjs';
  * `Query`, `Filter` and `Order` are provided as globals.
  */
 export class QuerySandbox extends Resource {
+  // Caching the wasm module.
+  static quickJS: Promise<QuickJSWASMModule> | null = null;
```
dmaretskyi: make private

packages/plugins/plugin-script/src/components/TypescriptEditor/TypescriptEditor.stories.tsx
```diff
@@ -36,12 +49,27 @@ const DefaultStory = (props: TypescriptEditorProps) => {
...
-      timestamp: Date.now(),
-    });
+  const handleRun = useCallback(async () => {
+    try {
+      const definitions: [string, any][] = [['x', 100]];
+      const quickJS = await getQuickJS();
+
+      // TODO(burdon): Eval.
+      // eslint-disable-next-line @typescript-eslint/no-implied-eval
+      const result = new Function(...definitions.map(([name]) => name), `return ${object.content.trim()}`)(
```
dmaretskyi, line 59: use quickjs?

PR #10001, opened 2025-10-14
feat: Interpreter compute graph

richburdon:
## POC Notebook

https://github.com/user-attachments/assets/145d998d-36e9-4dab-92ba-23af283ba4cf

### DIFF
--- a/packages/plugins/plugin-script/src/notebook/notebook.test.ts
+++ b/packages/plugins/plugin-script/src/notebook/notebook.test.ts
@@ -14,7 +14,6 @@ import { ComputeGraph } from './notebook';
 describe('notebook', () => {
   test('parse dependency graph', () => {
     const notebook = createNotebook();
-
     const computer = new ComputeGraph(notebook);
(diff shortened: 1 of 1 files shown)


packages/core/echo/echo/src/Obj.ts
```diff
@@ -75,12 +75,14 @@ export const make = <S extends Type.Obj.Any>(
...
 
   if (props[EchoSchema.MetaId] != null) {
-    // Set default fields on meta on creation
+    // Set default fields on meta on creation.
     meta = { ...structuredClone(DEFAULT_META), ...props[EchoSchema.MetaId] };
     delete props[EchoSchema.MetaId];
   }
 
-  return live<Schema.Schema.Type<S>>(schema, props as any, { keys: [], ...meta });
+  const filteredProps = Object.fromEntries(Object.entries(props).filter(([_, v]) => v != null));
```
dmaretskyi, line 83: why?

PR #10016, opened 2025-10-15
chore: Ensure lockfile and node version is hashed by moon

dmaretskyi:


.moon/toolchain.yml
```diff
@@ -1,5 +1,6 @@
 # https://moonrepo.dev/docs/config/toolchain#node
 node:
+  version: 20.19.5 # keep in-sync with .prototools
```
dmaretskyi: add comment that this is a workarpund for moonpreo bug

PR #10024, opened 2025-10-16
fix: companion blueprint initialization

wittjosiah:
Resolves DX-573

### DIFF
--- a/packages/plugins/plugin-assistant/src/stories/Chat.stories.tsx
+++ b/packages/plugins/plugin-assistant/src/stories/Chat.stories.tsx
@@ -688,6 +688,7 @@ export const WithProject: Story = {
     config: config.remote,
     accessTokens: [Obj.make(DataType.AccessToken, { source: 'exa.ai', token: EXA_API_KEY })],
     types: [
+      Tag.Tag,
       DataType.Employer,
       DataType.HasConnection,
       DataType.Message,
--- a/packages/plugins/plugin-assistant/src/stories/testing/testing.tsx
+++ b/packages/plugins/plugin-assistant/src/stories/testing/testing.tsx
@@ -278,9 +278,9 @@ const StoryPlugin = definePlugin<StoryPluginOptions>(
               space.db.add(chat);
               await space.db.flush({ indexes: true });
 
-              binder.open();
+              await binder.open();
               await onChatCreated?.({ space, chat, binder });
-              binder.close();
+              await binder.close();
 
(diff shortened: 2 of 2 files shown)


packages/core/assistant-testing/src/blueprints/design/design.test.ts
```diff
@@ -45,6 +45,8 @@ describe('Design Blueprint', { timeout: 120_000 }, () => {
         const conversation = new AiConversation({
           queue: yield* QueueService.createQueue<DataType.Message | ContextBinding>(),
         });
+        conversation.open();
```
dmaretskyi: i had a wrapper to use a resource in effect - it opens and closes it for you (search in @dxos/effect package)

packages/core/assistant/src/conversation/context.ts
```diff
@@ -64,17 +65,47 @@ export class AiContextBinder {
...
+    invariant(this._blueprints, 'AiContextBinder not open');
+    return this._blueprints;
+  }
+
+  get objects() {
+    invariant(this._objects, 'AiContextBinder not open');
+    return this._objects;
+  }
+
+  open() {
```
dmaretskyi: I think its better to follow the `Resource` pattern here for uniformity

packages/plugins/plugin-assistant/src/hooks/useContextBinder.ts
```diff
@@ -2,17 +2,26 @@
...
 import { type Assistant } from '../types';
 
 export const useContextBinder = (chat: Assistant.Chat | undefined): AiContextBinder | undefined => {
-  const binder = useMemo(() => {
-    const queue = chat?.queue.target;
-    return queue && new AiContextBinder(queue);
-  }, [chat]);
+  const [binder, setBinder] = useState<AiContextBinder>();
+
+  useEffect(() => {
```
dmaretskyi:
```ts
const useResource = <T extends Lifecycle>(get: () => T, deps?: ReactDependencyArray): T => {
   const resource = useMemo(get, deps);
    useEffect(() => {
      return () => resource.close?.()
   }, [resource]);
   use(resource.open?.()); // Let's try out suspense!
   return resource
}
```

wittjosiah: I don't think we should do that until we have more suspense boundaries in place at the app level, which I think should be out of scope for this PR

PR #10025, opened 2025-10-17
feat: Sync repro

dmaretskyi:
- [x] Add message counter logging to test
- [x] Repro sync bug in composer
- [ ] Find out why collection sync defaults to the poll timeout
- [ ] Introduce etags to collection sync so that the poll-interval could be reduced
- [ ] Collection state pushes

### DIFF
--- a/packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
+++ b/packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
@@ -15,10 +15,6 @@ const MIN_QUERY_INTERVAL = 5_000;
 
 const POLL_INTERVAL = 30_000;
 
-// const MIN_QUERY_INTERVAL = 500;
-
-// const POLL_INTERVAL = 1_000;
-
 export type CollectionSynchronizerParams = {
   sendCollectionState: (collectionId: string, peerId: PeerId, state: CollectionState) => void;
   queryCollectionState: (collectionId: string, peerId: PeerId) => void;


packages/apps/composer-app/index.html
```diff
@@ -55,6 +55,15 @@
...
 
+    <script strategy="beforeInteractive">
+      // Suppress lit dev mode warning.
+      // https://github.com/lit/lit/issues/4877
+      globalThis.litIssuedWarnings ??= new Set();
+      globalThis.litIssuedWarnings.add(
+        'Lit is in dev mode. Not recommended for production! See https://lit.dev/msg/dev-mode for more information.',
+      );
+    </script>
+
```
mykola-vrmchk, line 66: Do we need this?

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
```diff
@@ -15,6 +15,10 @@ const MIN_QUERY_INTERVAL = 5_000;
 
 const POLL_INTERVAL = 30_000;
 
+// const MIN_QUERY_INTERVAL = 500;
+
+// const POLL_INTERVAL = 1_000;
+
```
mykola-vrmchk: Does this help in any way?

packages/apps/composer-app/index.html
dmaretskyi: Yes, it prevents lit from writing warnings to console

packages/core/echo/echo-pipeline/src/automerge/collection-synchronizer.ts
dmaretskyi: Remove this

PR #10067, opened 2025-10-24
fix: sheet fn binding reactivity

wittjosiah:


packages/core/compute/src/compute-graph.ts
```diff
@@ -242,14 +243,25 @@ export class ComputeGraph extends Resource {
 
   protected override async _open(): Promise<void> {
     if (this._space) {
+      // TODO(wittjosiah): Use effect-atom instead of signal.
+      const allFunctions = signal<FunctionType[]>([]);
+
```
dmaretskyi: I think even simpler would be to use the reactivity of `query.objects`

```ts
const query = space.db.query(...);

  const dispose = effect(() => {
      this._remoteFunctions = query.objects.filter(({ binding }) => binding);
    });
const unsubscribe = query.subscribe()

    this._ctx.onDispose(() => {
      unsubscribe();
      dispose();
    });
```

PR #10075, opened 2025-10-25
feat: replace form in 1-up view

wittjosiah:


packages/plugins/plugin-preview/src/capabilities/react-surface.tsx
```diff
@@ -130,4 +130,20 @@ export default () =>
...
+
+    createSurface({
+      id: `${meta.id}/section`,
+      role: ['section'],
+      position: 'fallback',
+      filter: (data): data is { subject: Obj.Any } => Obj.isObject(data.subject),
+      component: ({ data }) => {
+        return (
+          <div className='flex justify-center'>
+            <div className='w-96'>
```
richburdon: role=none is-[20-rem]

PR #10085, opened 2025-10-28
feat: Mobile layouts

thure:
This PR implements more mobile-native designs for MobileLayoutPlugin’s Home view and Space one-up view.

To do:
- [ ] System dialogs

https://github.com/user-attachments/assets/93038643-3f24-468e-9dbd-99454e7d7642

### DIFF
--- a/packages/plugins/plugin-mobile-layout/src/components/Home.tsx
+++ b/packages/plugins/plugin-mobile-layout/src/components/Home.tsx
@@ -15,8 +15,6 @@ import { meta } from '../meta';
 
 import { navHeaderButton, navHeaderHeading, navHeaderRoot } from './NavHeader';
 
-const subheading = 'container-max-width pli-cardSpacingInline mlb-cardSpacingBlock text-lg';
-
 export const Home = () => {
   const { t } = useTranslation(meta.id);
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-mobile-layout/src/components/Main.tsx
```diff
@@ -41,14 +41,18 @@ export const Main = () => {
...
   return (
-    <NaturalMain.Root>
-      <Activity mode={id === 'default' ? 'visible' : 'hidden'}>
-        <Home />
-      </Activity>
-      <Activity mode={id !== 'default' ? 'visible' : 'hidden'}>
-        <NavHeader node={node} />
-        <Surface key={id} role='article' data={data} limit={1} fallback={ContentError} placeholder={placeholder} />
-      </Activity>
+    <NaturalMain.Root complementarySidebarState='closed' navigationSidebarState='closed'>
```
wittjosiah, line 44: are these not closed by default?

thure: This is a controllable state property, so setting them to `closed` here prevents anything from unintentionally setting a different state. I did this just in case.

PR #10086, opened 2025-10-28
refactor: rename object functions and add development notes

richburdon:
Partly resolves: https://linear.app/dxos/issue/DX-402/generating-crud-tools-for-a-given-type


packages/ui/react-ui-editor/src/stories/components/EditorStory.tsx
```diff
@@ -5,11 +5,11 @@
 import { type EditorView } from '@codemirror/view';
 import React, { type ReactNode, forwardRef, useEffect, useImperativeHandle, useMemo, useState } from 'react';
 
+import { createDocAccessor, createObject } from '@dxos/client/echo';
```
wittjosiah, line 8: why client and not react-client? generally I try to consistently import from react-client when in react-land

packages/ui/react-ui-editor/src/extensions/automerge/cursor.ts
```diff
@@ -2,8 +2,8 @@
 // Copyright 2024 DXOS.org
 //
 
+import { type DocAccessor, fromCursor, toCursor } from '@dxos/client/echo';
```
wittjosiah, line 5: if we're cleaning this up, the ui package really shouldn't re-export from client at all... ideally it would be much lower level echo/AM interface. Maybe leave a todo?

packages/plugins/plugin-assistant/src/toolkits/AssistantToolkit.ts
```diff
@@ -0,0 +1,55 @@
...
+        const ref = Ref.fromDXN(ArtifactId.toDXN(id, db.spaceId));
+        yield* Effect.promise(() =>
+          binder.bind({
+            blueprints: [],
+            objects: [ref],
+          }),
+        );
+      }),
+    });
+}
```
dmaretskyi, line 55:
```suggestion
export class AssistantToolkit extends Toolkit.make(
  Tool.make('add-to-context', {
    description: trim`
      Adds the object to the chat context.
    `,
    parameters: {
      id: ArtifactId.annotations({
        description: 'The ID of the document to read.',
      }),
    },
    success: Schema.Void,
    failure: Schema.Never,
    dependencies: [AiContextService, DatabaseService],
  }),
) {
  static const tools = Record.keys(this.tools);

  export const layer = (
    _context: PluginContext,
  ) =>
    this.toLayer({
      'add-to-context': Effect.fnUntraced(function* ({ id }) {
        const { binder } = yield* AiContextService;
        const { db } = yield* DatabaseService;
        const ref = Ref.fromDXN(ArtifactId.toDXN(id, db.spaceId));
        yield* Effect.promise(() =>
          binder.bind({
            blueprints: [],
            objects: [ref],
          }),
        );
      }),
    });
}
```

PR #10132, opened 2025-11-04
feat: Function registry

dmaretskyi:


packages/core/functions/src/services/database.ts
```diff
@@ -145,26 +147,65 @@ export class DatabaseService extends Context.Tag('@dxos/functions/DatabaseServic
...
+              ),
+            );
+            return query.runSync();
+          }),
+        } satisfies QueryResultEx<any>;
+      }),
       Effect.withSpan('DatabaseService.query'),
     );
 
+    return Object.create(effect, {
```
wittjosiah: why is `Object.create` needed here?

packages/core/functions/src/services/database.ts
```diff
@@ -145,26 +147,65 @@ export class DatabaseService extends Context.Tag('@dxos/functions/DatabaseServic
...
-    DatabaseService.pipe(
-      Effect.map(({ db }) => db.query(queryOrFilter as any)),
+    <Q extends Query.Any>(query: Q): QueryResultEffect<Live<Query.Type<Q>>>;
+    <F extends Filter.Any>(filter: F): QueryResultEffect<Live<Filter.Type<F>>>;
+  } = (queryOrFilter: Query.Any | Filter.Any) => {
+    const effect = DatabaseService.pipe(
+      Effect.map(({ db }) => {
+        const query = db.query(queryOrFilter as any);
+
+        return {
```
wittjosiah, line 171: what if instead of `{ objects: Effect, rx: Rx }` the return was `{ run: Effect, objects: Rx }`? I don't like `rx` as a part of the api

PR #10150, opened 2025-11-06
feat: create, edit & invoke prompts

wittjosiah:


packages/plugins/plugin-assistant/src/components/PromptArticle.tsx
```diff
@@ -0,0 +1,89 @@
...
+        },
+      });
+
+      // Invoke the function.
+      const result = yield* FunctionInvocationService.invokeFunction(Agent.prompt, inputData).pipe(
+        Effect.provide(
+          ComputeEventLogger.layerFromTracing.pipe(
+            Layer.provideMerge(TracingService.layerQueue(trace.invocationTraceQueue)),
+          ),
+        ),
```
dmaretskyi, line 61: `effect.pipe(InvocationTracer.traceCall({ ... }))`

PR #10175, opened 2025-11-12
fix: correct schema version filtering in query planner

dmaretskyi:
- Added explicit filter step after schema index lookup to handle schema version disambiguation
- Updated test expectation to reflect correct filtering behavior for ContactV1 type
- Added documentation noting that schema index does not separate versions

### DIFF
--- a/packages/core/echo/echo-pipeline/src/query/query-planner.test.ts
+++ b/packages/core/echo/echo-pipeline/src/query/query-planner.test.ts
@@ -40,6 +40,15 @@ describe('QueryPlanner', () => {
             "mode": "only-non-deleted",
           },
           {
+            "_tag": "FilterStep",
+            "filter": {
+              "id": undefined,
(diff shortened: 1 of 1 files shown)


packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -403,7 +403,7 @@ describe('Query', () => {
       await db.flush({ indexes: true });
 
       const assertQueries = async (db: EchoDatabase) => {
-        await assertQuery(db, Filter.type(ContactV1), [contactV1, contactV2]);
+        await assertQuery(db, Filter.type(ContactV1), [contactV1]);
```
wittjosiah, line 406:
```suggestion
        await assertQuery(db, Filter.typename(ContactV1.typename), [contactV1, contactV2]);
```
I'd just revert this to what it was before

PR #10208, opened 2025-11-19
feat: space snapshot via query

wittjosiah:
part of DX-640

### DIFF
--- a/packages/core/echo/echo-db/src/serializer.ts
+++ b/packages/core/echo/echo-db/src/serializer.ts
@@ -2,7 +2,7 @@
 // Copyright 2023 DXOS.org
 //
 
-import { Filter } from '@dxos/echo';
+import { Filter, type Query } from '@dxos/echo';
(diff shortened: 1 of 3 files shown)


packages/core/echo/echo-db/src/serializer.ts
```diff
@@ -30,23 +31,26 @@ export type ImportOptions = {
 export class Serializer {
   static version = 1;
 
-  async export(database: EchoDatabase): Promise<SerializedSpace> {
-    const ids = database.coreDatabase.getAllObjectIds();
-
+  async export(target: EchoDatabase | QueryResult<any>): Promise<SerializedSpace> {
```
dmaretskyi: I've been avoiding touching the exporter API for a while. If you need this change, I think its better to pass Query as an option, and let the serializer run it

dmaretskyi: I'm also thinking about removing this API from ECHO and having the downstream consumers use Obj.toJSON / Obj.fromJSON

PR #10217, opened 2025-11-20
feat: add HTTP plugin and magic import for native functions bundler

mykola-vrmchk:


packages/core/functions-runtime/src/http-plugin-esbuild.ts
```diff
@@ -0,0 +1,60 @@
...
+      namespace: 'http-url',
+    }));
+
+    // When a URL is loaded, we want to actually download the content from the internet.
+    // This has just enough logic to be able to handle the example import from unpkg.com but in reality this would probably need to be more complex.
+    build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
+      return Effect.gen(function* () {
+        const response = yield* HttpClient.get(args.path);
+        if (response.status !== 200) {
+          throw new Error(`failed to fetch: ${response.status}`);
```
wittjosiah: consider handling http errors natively w/ effect https://github.com/Effect-TS/effect/blob/main/packages/platform/README.md#adding-errors

PR #10221, opened 2025-11-20
fix: event back sync

wittjosiah:
also:
- normalize description text
- trim actor list
- event click opens companion


packages/plugins/plugin-inbox/src/components/CalendarArticle/CalendarArticle.tsx
```diff
@@ -36,8 +39,14 @@ export const CalendarArticle = ({ subject: calendar }: SurfaceComponentProps<Cal
   const handleSelect = useCallback(
     (event: Event.Event) => {
       singleSelect(event.id);
+      void dispatch(
+        createIntent(DeckAction.ChangeCompanion, {
+          primary: id,
+          companion: `${id}${ATTENDABLE_PATH_SEPARATOR}event`,
```
richburdon, line 45: there's a function for this

PR #10230, opened 2025-11-22
chore: Refactor dxos/echo APIs 

richburdon:


packages/core/echo/echo/src/internal/index.ts
```diff
@@ -2,21 +2,60 @@
...
+//  - [x] Remove @deprecated from internal methods and mark @internal (e.g., getSchemaDXN).
+//    - NOTE: Internal methods should not use the import * from Obj/Type APIs.
+//  - [x] import ObjectId => @dxos/keys
+//  - [x] Entity type for Obj | Relation
+//  - [x] import LabelAnnotation => Annotation.LabelAnnotation
+//  - [x] import Expando => Type.Expando
+//  - [x] live => Obj.make
+//  - [x] Rename live => makeObject
+//  - [x] Rename {EchoObject, EchoRelation} => {EchoObjecSchema, EchoRelationSchema}
+//  - [x] Rename BaseObject => AnyProperties
```
dmaretskyi, line 26: I we can remove BaseObject entirely and replace with Obj.Any

packages/core/echo/echo/src/internal/entities/entity.ts
```diff
@@ -0,0 +1,109 @@
...
+    //     : Simplify<Schema.TypeLiteral.Constructor<Fields, []>>,
+    //   options?: MakeOptions,
+    // ): Simplify<Schema.TypeLiteral.Type<Fields, []>> {
+    //   const propsWithDefaults: any = _lazilyMergeDefaults(fields, { ...(props as any) });
+    //   return _getDisableValidationMakeOption(options)
+    //     ? propsWithDefaults
+    //     : ParseResult.validateSync(this)(propsWithDefaults);
+    // }
+
+    static instanceOf(value: unknown): boolean {
```
dmaretskyi, line 105: we should remove this and use standalone Obj.instanceOf

packages/core/echo/echo/src/internal/entities/expando.ts
```diff
@@ -4,12 +4,15 @@
...
 import * as Schema from 'effect/Schema';
 
-import { EchoObject } from './entity';
+import { EchoObjectSchema } from './object';
 
 export const EXPANDO_TYPENAME = 'dxos.org/type/Expando';
 
 const ExpandoSchema = Schema.Struct({}, { key: Schema.String, value: Schema.Any }).pipe(
-  EchoObject({ typename: EXPANDO_TYPENAME, version: '0.1.0' }),
+  EchoObjectSchema({
```
dmaretskyi, line 12: replace with Type.Obj eventually

dmaretskyi: We should collapse the defintions (EchoObjectSchema -> Type.Obj). Lets prefer using Type.Obj unless there're circular deps forcing to do otherwise. This stands for all other cases of internal vs api defintions.

packages/core/echo/echo/src/internal/entities/model.ts
```diff
@@ -9,115 +9,71 @@ import { invariant } from '@dxos/invariant';
 import { DXN, ObjectId } from '@dxos/keys';
 import { assumeType } from '@dxos/util';
 
-import { EntityKind } from '../ast';
-
-import { type ObjectMeta } from './meta';
-import type { Version } from './version';
+import {
```
dmaretskyi, line 12: I liked how property symbols and interfaces were're defined in one place here. It makes reasoning about the shape of the data simpler

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface HasId {
+  readonly id: ObjectId;
+}
+
+/**
+ * Object that has an associated typename.
+ * The typename is retrievable using {@link getTypename}.
+ * The object can be used with {@link isInstanceOf} to check if it is an instance of a schema.
+ */
+export type HasTypename = {};
```
dmaretskyi, line 24: is this still useful?

We don't need aspects anymore after simplifying object ontology -- all objects have:

- id
- kind: object, relation
- typename
- schema
- meta
- deletion marker

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Canonical type for all ECHO objects.
+ */
+// TODO(burdon): Rename to AnyEntity? (rename all "object" to "entity")?
+export interface AnyEchoObject extends HasId, HasTypename {}
```
dmaretskyi, line 30: remove

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,60 @@
...
+ * It is stricter than `T extends {}` or `T extends object`.
+ *
+ * @internal
+ */
+// TODO(dmaretskyi): Rename AnyProperties.
+// TODO(dmaretskyi): Prefer `Record<string, unknown>` over `any`.
+export type AnyProperties = Record<string, any>;
+
+// TODO(dmaretskyi): Remove; this type effectively disables type safety due to `any`.
+export type WithId<T extends AnyProperties = AnyProperties> = T & HasId;
```
dmaretskyi, line 44: remove

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type ExcludeId<T extends AnyProperties> = Omit<T, 'id'>;
+
+export type PropertyKey<T extends AnyProperties> = Extract<keyof ExcludeId<T>, string>;
+
+// TODO(dmaretskyi): Remove. This should be using the symbol type.
+export type WithMeta = { [ATTR_META]?: ObjectMeta };
```
dmaretskyi, line 51: remove?

packages/core/echo/echo/src/internal/types/typename.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Property name for typename when object is serialized to JSON.
+ */
+export const ATTR_TYPE = '@type';
+
+/**
+ * DXN to the object type.
+ */
+export const TypeId = Symbol.for('@dxos/echo/Type');
```
dmaretskyi, line 22: I think it was better when the symbols were together with the object model definition in the `model.ts`

packages/core/echo/echo/src/internal/index.ts
```diff
@@ -2,21 +2,60 @@
...
+//  - [ ] Remove WithMeta => AnyEchoObject
+//  - [ ] HasId
+//  - [ ] BaseSchema
+//  - [ ] JsonPath, JsonProp, getValue, setValue => Json.Path?
+//
+//  3. Audit usage from @dxos/echo-db
+//  - [ ] QueryFn, QueryOptions => Database
+//  - [ ] Datatbase.query() options?
+//  - [ ] Expando type.
+//  - [ ] Standardize '@automerge/automerge' imports (A vs. next).
```
dmaretskyi, line 47: ask AI to one-shot a lint rule?

> Create a lint rule (similar to existing rules in the monorepo)" use `pnpm eslint <filename>` to test (try with different files. The rule should standartize namespace imports. The rule should have config: `{ "importNamepaces": { "@automerge/automerge": "Automerge" }`

packages/core/echo/echo/src/Entity.ts
```diff
@@ -0,0 +1,351 @@
...
+import { type DXN } from '@dxos/keys';
+
+import {
+  MetaId,
+  type ObjectJSON,
+  type ObjectMeta,
+  ObjectVersionId,
+  type VersionType,
+  VersionTypeId,
+  getDescription as getDescription$,
```
dmaretskyi, line 19: a better style then importing name$ would be a namespace import:

```ts
import * as internal from './internal';

// ...


internal.getDescription(...)

```

packages/core/echo/echo/src/Annotation.ts
```diff
@@ -0,0 +1,17 @@
+//
```
dmaretskyi, line 1: Could you update the README.md with a list of exported APIs please?

packages/core/echo/echo/src/index.ts
```diff
@@ -2,16 +2,21 @@
 // Copyright 2025 DXOS.org
 //
 
+export * from '@dxos/keys';
```
dmaretskyi, line 5: can we kill this?

packages/core/echo/echo/src/index.ts
```diff
@@ -2,16 +2,21 @@
...
+
 export * from './errors';
+
+// TODO(burdon): Separate namesapces for Query/Filter/Order?
 export * from './query';
```
dmaretskyi, line 10: I don't mind bringing them to top level directory (same as Obj, Relation)

packages/ui/react-ui-table/src/components/TableCellEditor/FormCellEditor.tsx
```diff
@@ -5,8 +5,8 @@
 import * as Schema from 'effect/Schema';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
-import { Filter, Obj } from '@dxos/echo';
-import { FormatEnum, Ref, type TypeAnnotation, getValue } from '@dxos/echo/internal';
+import { Filter, Obj, Ref } from '@dxos/echo';
+import { Format, type TypeAnnotation, getValue } from '@dxos/echo/internal';
```
dmaretskyi, line 9: There's ./scripts/rename.mjs and .windsurf/workflows/rename.md to help with bulk renaming.

Also you could use the vibe & replace feature in windsurf (it's integrated with the search & replace panel):

Image

packages/core/echo/echo/src/Database.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Query objects.
+   */
+  query: QueryFn;
+
+  /**
+   * Adds object to the database.
+   */
+  // TODO(burdon): Restrict to Entity.Any (Obj.Any | Relation.Any).
```
dmaretskyi, line 180: 👍

packages/core/echo/echo/src/Database.ts
```diff
@@ -0,0 +1,189 @@
...
+  add<T extends AnyProperties>(obj: Live<T>, opts?: AddOptions): Live<T & HasId>;
```
dmaretskyi, line 181:
```suggestion
  add<T extends Obj.Any | Relation.Any>(obj: T, opts?: AddOptions): T;
```

Enforcing that what you pass to .add is already a well-formed object (made by Obj.make) instead of adding extra props to the return type

packages/core/echo/echo/src/Database.ts
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * Creates a reference to an existing object in the database.
+   *
+   * NOTE: The reference may be dangling if the object is not present in the database.
+   * NOTE: Difference from `Ref.fromDXN`
+   * `Ref.fromDXN(dxn)` returns an unhydrated reference. The `.load` and `.target` APIs will not work.
+   * `db.ref(dxn)` is preferable in cases with access to the database.
+   */
+  // TODO(burdon): Rename makeRef.
+  ref<T extends Obj.Any = Obj.Any>(dxn: DXN): Ref.Ref<T>;
```
dmaretskyi, line 170:
```suggestion
  ref<T extends Obj.Any | Relation.Any = Obj.Any | Relation.Any>(dxn: DXN): Ref.Ref<T>;
```

packages/core/echo/echo/src/Database.ts
```diff
@@ -0,0 +1,189 @@
...
+   * @default 'linked-doc'
+   */
+  placeIn?: ObjectPlacement;
+};
+
+/**
+ * ECHO Database interface.
+ */
+export interface Database extends Queryable {
+  get spaceId(): SpaceId;
```
dmaretskyi, line 156:
```suggestion
  readonly spaceId: SpaceId;
```

packages/core/echo/echo/src/Database.ts
```diff
@@ -0,0 +1,189 @@
...
+   */
+  fire?: boolean;
+};
+
+export interface QueryResult<T extends Entity.Any = Entity.Any> {
+  readonly query: Query<T>;
+  readonly results: QueryResultEntry<T>[];
+  readonly objects: T[];
+
+  run(opts?: QueryRunOptions): Promise<OneShotQueryResult<T>>;
```
dmaretskyi, line 71: grab jsdoc pls from the implementation class

packages/core/echo/echo/src/internal/types/entity.ts
```diff
@@ -0,0 +1,36 @@
...
+import * as Schema from 'effect/Schema';
+
+import { type ObjectId } from '@dxos/keys';
+
+import { type HasId } from './base';
+
+/**
+ * Entity kind.
+ */
+export const EntityKindId = Symbol.for('@dxos/echo/EntityKind');
```
dmaretskyi, line 14: Effect does an odd but necessary pattern with the unique symbol definitions:

1. An API re-export with a typedef: https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Effect.ts#L81
2. An internal definition type-cast to the API type.

Effectively:

- There's a single type for both symbols -- `unique symbol` are the only nominal type in TS, they are tied to the source code declaration so you must only have one place where you define `unique symbol`.
- No non-type imports from API modules required, other code can import the symbol from `./internal`
- When you click "go-to-definition" on the symbol you get taken to the API module, that's where `unique symbol` definition is.

PR #10233, opened 2025-11-25
feat: ECHO API update

richburdon:


packages/core/echo/echo/src/Obj.ts
```diff
@@ -5,64 +5,86 @@
...
+  objectFromJSON,
+  objectToJSON,
+  setDescription as setDescription$,
+  setLabel as setLabel$,
+} from './internal';
+import * as Ref from './Ref';
 import type * as Relation from './Relation';
 import * as Type from './Type';
 
+interface BaseObj extends AnyEchoObject, Type.OfKind<EntityKind.Object> {}
```
dmaretskyi: I'd wanna kill AnyEchoObject -- those are the old types, the new definitions are good as is.

dmaretskyi: Kill HasTypename too

packages/core/echo/echo/src/Obj.ts
```diff
@@ -5,64 +5,86 @@
...
-import { type DXN } from '@dxos/keys';
-import * as LiveObject from '@dxos/live-object';
+import { type DXN, ObjectId } from '@dxos/keys';
+import { type Live, getSnapshot as getSnapshot$ } from '@dxos/live-object';
 import { assumeType, deepMapValues } from '@dxos/util';
 
-import { live } from './internal';
-import * as EchoSchema from './internal';
-import type * as Ref from './Ref';
+import {
```
dmaretskyi, line 15: the way effect does it, is they keep all type defs in the API module, this makes it much easier to reason about the API shape. For example its hard for me to follow `AnyProps` extending `AnyProperties` thats defined in a separate module

dmaretskyi: In general we should aim to eliminate types in the './interal' dir (most of them are deprecated!) and use the types in API modules which are for the most part self-sufficient. we shouldn't try to blend the types together

packages/core/echo/echo/src/Obj.ts
```diff
@@ -5,64 +5,86 @@
...
 /**
  * Object with arbitrary properties.
  *
  * NOTE: Due to how typescript works, this type is not assignable to a specific schema type.
  * In that case, use `Obj.instanceOf` to check if an object is of a specific type.
  */
-export interface AnyProps extends BaseObj {
-  [key: string]: any;
-}
+export interface AnyProps extends BaseObj, AnyProperties {}
```
dmaretskyi, line 72: I dont think extracting AnyProperties is useful here -- too much granularity -- just have the type defined inline

packages/core/echo/echo/src/Obj.ts
```diff
@@ -5,64 +5,86 @@
...
-export const Meta: unique symbol = EchoSchema.MetaId as any;
+type Props<T = any> = {
+  id?: ObjectId;
+  [Meta]?: Partial<ObjectMeta>;
+} & Type.Properties<T>;
 
-const DEFAULT_META: EchoSchema.ObjectMeta = {
-  keys: [],
-};
+// TODO(burdon): Should we allow the caller to set the id?
```
dmaretskyi, line 83: yes, its useful in some niche circumstances

packages/core/echo/echo/src/Obj.ts
```diff
@@ -77,30 +99,95 @@ const DEFAULT_META: EchoSchema.ObjectMeta = {
...
-    EchoSchema.getTypeAnnotation(schema)?.kind === EchoSchema.EntityKind.Object,
-    'schema',
-    'Expected an object schema',
-  );
-
-  if (props[EchoSchema.MetaId] != null) {
-    // Set default fields on meta on creation.
-    meta = { ...structuredClone(DEFAULT_META), ...props[EchoSchema.MetaId] };
-    delete props[EchoSchema.MetaId];
+  meta?: Partial<ObjectMeta>,
```
dmaretskyi, line 105: we should kill the last parameter and use the `[Meta]` symol in props instead

packages/core/echo/echo/src/Obj.ts
```diff
@@ -77,30 +99,95 @@ const DEFAULT_META: EchoSchema.ObjectMeta = {
...
+
+  return make(schema, props);
+};
+
+//
+// NOTE: The definitions below are common to Obj and Relation.
+//
+
+type AnyEntity = Any | Relation.Any;
+type AnyEntityType = Type.Obj.Any | Type.Relation.Any;
```
dmaretskyi: Move AnyEntityType to `Type.ts`

packages/core/echo/echo/src/Relation.ts
```diff
@@ -6,50 +6,63 @@ import * as Schema from 'effect/Schema';
...
 import * as Type from './Type';
 
-/**
- * NOTE: Don't export: Relation.Relation and Relation.Any form the public API.
- */
 interface BaseRelation<Source, Target>
-  extends EchoSchema.HasId,
-    Type.Relation.Endpoints<Source, Target>,
-    Type.OfKind<EchoSchema.EntityKind.Relation> {}
+  extends AnyEchoObject,
```
dmaretskyi, line 38: inlining AnyEchoObject here is obfuscating the typedef, HasId + Relation.Endpoints + Type.OfKind was more readable to me

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface HasId {
+  readonly id: ObjectId;
+}
+
+/**
+ * Object that has an associated typename.
+ * The typename is retrievable using {@link getTypename}.
+ * The object can be used with {@link isInstanceOf} to check if it is an instance of a schema.
+ */
+export type HasTypename = {};
```
dmaretskyi, line 24: kill this

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Base type for all data objects (reactive, ECHO, and other raw objects).
+ * NOTE: This describes the base type for all database objects.
+ * It is stricter than `T extends {}` or `T extends object`.
+ */
+// TODO(burdon): Make internal.
+// TODO(burdon): Prefer Record<string, unknown>.
+// TODO(burdon): Exclude "id", etc. from keys.
+export type AnyProperties = Record<string, any>;
```
dmaretskyi, line 34: kill this

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Canonical type for all ECHO objects.
+ */
+export interface AnyEchoObject extends HasId, HasTypename {}
```
dmaretskyi, line 39: kill this

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// TODO(dmaretskyi): Remove; this type effectively disables type safety due to `any`.
+export type WithId<T extends AnyProperties = AnyProperties> = T & HasId;
+
+export type ExcludeId<T extends AnyProperties> = Omit<T, 'id'>;
+
+export type PropertyKey<T extends AnyProperties> = Extract<keyof ExcludeId<T>, string>;
+
+// TODO(dmaretskyi): Remove. This should be using the symbol type.
+export type WithMeta = { [ATTR_META]?: ObjectMeta };
```
dmaretskyi, line 49: kill

packages/core/echo/echo/src/internal/types/base.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * The raw object should not include the ECHO id, but may include metadata.
+ */
+export const RawObject = <S extends Schema.Schema.AnyNoContext>(
```
dmaretskyi, line 54: Rename `Type.PropertiesOf`

packages/core/echo/echo/src/internal/ref/ref.ts
```diff
@@ -315,12 +317,12 @@ export interface RefResolver {
    * @param load If true the resolver should attempt to load the object from disk.
    * @param onLoad Callback to call when the object is loaded.
    */
-  resolveSync(dxn: DXN, load: boolean, onLoad?: () => void): BaseObject | undefined;
+  resolveSync(dxn: DXN, load: boolean, onLoad?: () => void): AnyProperties | undefined;
```
dmaretskyi, line 320: Anything perventing you from using `Obj.AnyEntity` as the base type for all of this. 
AnyProperties isn't a great type -- its just `Record<string, any>`!

Obj.Any and Relation.Any have `[EntityKind]: Obj | Relation` forced which makes TypeScript differentiate them from any other JavasScript object. On the other had, `Record<string, any>` doesn't enforce a particular type.

packages/core/echo/echo/src/internal/types/base.ts
dmaretskyi:
```ts
// Obj.ts

export interface AnyProperties extends Obj.Any {
   [key: string]: any;
}
```

dmaretskyi: cc @wittjosiah

PR #10236, opened 2025-11-26
refactor: add QueryResult namespace

wittjosiah:


packages/core/echo/echo/src/QueryResult.ts
```diff
@@ -0,0 +1,75 @@
...
+    source: 'remote' | 'local' | 'index';
+
+    /**
+     * Query resolution time in milliseconds.
+     */
+    time: number;
+  };
+};
+
+export type OneShotResult<T extends Entity.Unknown = Entity.Unknown> = {
```
richburdon: Just OneShot

packages/core/echo/echo/src/QueryResult.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export type SubscriptionOptions = {
+  /**
+   * Fire the callback immediately.
+   */
+  fire?: boolean;
+};
+
+// TODO(burdon): Narrow types.
```
richburdon: remove comment?

packages/core/echo/echo/src/QueryResult.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface QueryResult<T extends Entity.Unknown = Entity.Unknown> {
```
richburdon, line 60: I think we should promote query to Query.ts (and Order, Filter) so that the code is more normalized since these are so related. It feels to me that Database, QueryResult and the three above are qualitatively different than all of the others and we might think about eventually considering them as a different entrypoint/subsystem. but let's keep them here for now. (e.g., we might lift the query defs and impl from echo-db) into "@dxos/echo/Query"

dmaretskyi (approved):

packages/core/echo/echo/src/Database.ts
dmaretskyi, line 94: I'd challenge if we actually even need it in this API

PR #10240, opened 2025-11-27
fix: Resurect composer scripts

dmaretskyi:
- [x] Forex-effect deploys and runs
- [x] Runtime .d.ts types are preloaded in the editor.
- [x] Loading CDN assets shouldn't block composer startup
- [x] Asset load task can fail (there's about 400 .d.ts files - each causes a network request) -- better error handling + caching on R2 side
- [ ] .d.ts files come with externalized packages (had to do it cause rollup was pulled-in)
- [x] Separate envs for bundles (dev, main, production)
- [ ] Create a workflow to upload vendored bundles on CI

### DIFF
--- a/packages/ui/react-ui-editor/src/extensions/tags/extended-markdown.test.ts
+++ b/packages/ui/react-ui-editor/src/extensions/tags/extended-markdown.test.ts
@@ -20,7 +20,8 @@ describe('extended-markdown', () => {
     });
   };
 
-  test('tree', async ({ expect }) => {
+  // Flaky.
+  test.skip('tree', async ({ expect }) => {
     const doc = trim`
       <prompt>
         Hello


packages/plugins/plugin-script/src/capabilities/compiler.ts
```diff
@@ -7,28 +7,86 @@ import wasmUrl from 'esbuild-wasm/esbuild.wasm?url';
...
   compiler.setFile(
-    '/src/runtime.ts',
-    `
-        export const Filter: any = {};
-        export type FunctionHandler = ({ event, context }: { event: any; context: any }) => Promise<Response>;
-        export const functionHandler = (handler: FunctionHandler) => handler;
-      `,
+    '/src/typings.d.ts',
+    trim`
+      declare module 'https://*';
```
wittjosiah: pull my changes from main, it includes a globals file for the compiler which this can go in

packages/apps/composer-app/vite.config.ts
```diff
@@ -64,6 +64,9 @@ export default defineConfig((env) => ({
   },
   esbuild: {
     keepNames: true,
+    supported: {
+      'using': false,
+    }
```
wittjosiah: this didn't help

packages/core/echo/echo/moon.yml
```diff
@@ -10,6 +10,11 @@ tasks:
       - '--entryPoint=src/index.ts'
       - '--entryPoint=src/internal/index.ts'
       - '--entryPoint=src/testing/index.ts'
+      - '--entryPoint=src/Entity.ts'
```
wittjosiah, line 13: does this make the other entrypoints work?

packages/core/functions-runtime/scripts/vendor-packages.mjs
wittjosiah, line 1: is this for building the code which is uploaded to R2? is it superseded by the tsdown config?s

packages/core/functions-runtime/src/bundler/plugins/http-plugin-esbuild.ts
```diff
@@ -40,6 +48,8 @@ export const httpPlugin: Plugin = {
     // This has just enough logic to be able to handle the example import from unpkg.com but in reality this would probably need to be more complex.
     build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
       return Effect.gen(function* () {
+        console.log('download', args.path);
```
wittjosiah: 👀

packages/core/functions-runtime/tsdown.config.ts
```diff
@@ -0,0 +1,119 @@
...
+for (const [key, value] of Object.entries(entryPoints)) {
+  await fs.mkdir(path.dirname(`dist/entrypoints/${key}`), { recursive: true });
+  await fs.writeFile(`dist/entrypoints/${key}.ts`, `export * from "${value}";`);
+  processedEntryPoints[key] = `dist/entrypoints/${key}.ts`;
+}
+
+export default defineConfig({
+  entry: processedEntryPoints,
+  outDir: 'dist/vendor',
+  noExternal: () => true,
```
wittjosiah: you should verify if this is still needed after pulling in the fixes from main to `@dxos/effect` exports

.prototools
```diff
@@ -1,6 +1,6 @@
 # https://moonrepo.dev/docs/proto/config
 
-node = "24.4.1"
+node = "24.11.1"
```
wittjosiah, line 3: bear in mind that there's also an engine entry in the package.json

packages/core/functions-runtime/package.json
```diff
@@ -52,7 +52,8 @@
     }
   },
   "files": [
-    "dist",
+    "dist/lib",
+    "dist/types",
```
wittjosiah, line 56: probably safer to build the upload modules to another directory like `out` so that it doesn't get accidentally published in the future

packages/core/functions-runtime/src/bundler/plugins/embedded-vendored-packages.ts
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import * as Function from 'effect/Function';
+import * as Record from 'effect/Record';
+import type { Plugin } from 'esbuild-wasm';
+
+// TODO(dmaretskyi): Haven't managed to get this working.
```
wittjosiah, line 9: what is this trying to do?

packages/core/functions-runtime/src/native/bundler.ts
```diff
@@ -16,7 +16,7 @@ import { BaseError } from '@dxos/errors';
 import { PublicKey } from '@dxos/keys';
 import { Unit, trim } from '@dxos/util';
 
-import { httpPlugin } from '../http-plugin-esbuild';
+import { httpPlugin } from '../bundler/plugins/http-plugin-esbuild';
```
wittjosiah, line 20: why direct import?

packages/plugins/plugin-script/src/testing/deploy.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      version: '0.0.1',
+      entryPoint: buildResult.entryPoint,
+      assets: buildResult.assets,
+      runtime: Runtime.WORKER_LOADER,
+    });
+    console.log(newFunction);
+
+    expect(newFunction.id).toBeDefined();
+
+    // // Invoke deployed function via EDGE directly.
```
wittjosiah, line 51: this isn't working?

packages/plugins/plugin-script/src/capabilities/compiler.ts
dmaretskyi: not sure I see it

wittjosiah: [https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-script/src/compiler/compiler.ts#L28](https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-script/src/compiler/compiler.ts#L28)

PR #10243, opened 2025-11-27
refactor: SchemaRegistry

wittjosiah:


packages/core/echo/echo/src/SchemaRegistry.ts
```diff
@@ -0,0 +1,92 @@
...
+//
+
+import type * as Types from 'effect/Types';
+
+import type * as JsonSchema from './JsonSchema';
+import type * as Key from './Key';
+import type * as QueryResult from './QueryResult';
+import type * as Type from './Type';
+
+/** @deprecated Replace w/ Query.Query. */
```
dmaretskyi: What's the plan here?

wittjosiah: Leaving for a followup, not planning to address immediately.

dmaretskyi: then I'd say this should be a TODO rather then @deprecated. @deprecated should come with a migration path to a different API.

wittjosiah: fair enough, agreed

PR #10244, opened 2025-11-28
refactor: query api

wittjosiah:


packages/core/echo/echo/src/Filter.ts
```diff
@@ -0,0 +1,382 @@
...
+    type: 'not',
+    filter: {
+      type: 'object',
+      typename: null,
+      props: {},
+    },
+  });
+};
+
+export const relation = () => {
```
richburdon: consistently define return type (as above)

richburdon: typeDXN below returns Any

packages/core/echo/echo/src/Filter.ts
```diff
@@ -0,0 +1,382 @@
...
+    operator: 'lte',
+    value,
+  });
+};
+
+/**
+ * Predicate for property to be in the provided array.
+ * @param values - Values to check against.
+ */
+const in_ = <T>(...values: T[]): Filter<T | undefined> => {
```
richburdon: in$?

packages/core/echo/echo/src/index.ts
```diff
@@ -2,20 +2,21 @@
...
 
-export * from './errors';
-
-// TODO(burdon): Separate namesapces for Query (Query.Query, Query.Filter, etc.)
-export * from './query';
-
 export * as Annotation from './Annotation';
 export * as Database from './Database';
 export * as Entity from './Entity';
+export * as Err from './Err';
```
dmaretskyi, line 12: not sure about this change stylistically; I haven't seen this pattern before. But I won't fight you on this

wittjosiah: not sure about it either but giving it a shot

PR #10271, opened 2025-12-06
feat: CLI with assistant APIs

richburdon:
# Demo

https://github.com/user-attachments/assets/56de8a8b-1356-4810-bfff-0118353ab33d

### DIFF
--- a/packages/core/ai/src/resolvers/ChatCompletionsAdapter.ts
+++ b/packages/core/ai/src/resolvers/ChatCompletionsAdapter.ts
@@ -13,30 +13,12 @@ import * as Effect from 'effect/Effect';
 import * as Layer from 'effect/Layer';
 import * as Stream from 'effect/Stream';
 
-//
-// Types
-//
-
 /**
  * Chat message format (OpenAI-compatible).
  */
 type ChatMessage = {
   role: 'system' | 'user' | 'assistant';
   content: string;
-};
-
-/**
- * API format for the chat completions endpoint.
- */
-export type ApiFormat = 'ollama' | 'openai';
-
-/**
- * Chat completions client configuration.
- */
-export type ChatCompletionsClientConfig = {
-  readonly baseUrl: string;
-  readonly apiFormat: ApiFormat;
-  readonly transformClient?: (client: HttpClient.HttpClient) => HttpClient.HttpClient;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/ai/src/resolvers/anthropic/AnthropicResolver.ts
```diff
@@ -0,0 +1,28 @@
...
+import { type ModelName } from '../../defs';
+import { type AiModelNotAvailableError } from '../../errors';
+
+export const make = () =>
+  AiModelResolver.AiModelResolver.fromModelMap(
+    'Anthropic',
+    Effect.gen(function* () {
+      return {
+        '@anthropic/claude-3-5-haiku-latest': yield* AnthropicLanguageModel.model('claude-3-5-haiku-latest'),
+        '@anthropic/claude-3-5-haiku-20241022': yield* AnthropicLanguageModel.model('claude-3-5-haiku-20241022'),
```
dmaretskyi, line 20: lets kill the timstamped models and just use latest

packages/core/ai/src/resolvers/ChatCompletionsAdapter.ts
```diff
@@ -0,0 +1,439 @@
...
+    }
+  } catch {
+    return null;
+  }
+};
+
+/**
+ * Create a chat completions language model service.
+ */
+export const make = (model: string): Effect.Effect<LanguageModel.Service, never, ChatCompletionsClient> =>
```
dmaretskyi, line 327: Did the existing OpenAI client not work? Its intended to be compatible with lmstudio

packages/devtools/cli-tui/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { AiModelResolver, type ModelName } from '@dxos/ai';
+import { runAndForwardErrors } from '@dxos/effect';
+
+import { App } from './app';
+import { Core } from './core';
+
+type Provider = 'edge' | 'lmstudio' | 'ollama';
+
+const argv = yargs(hideBin(process.argv))
```
dmaretskyi, line 18: why not effect-cli?

packages/devtools/cli-tui/src/core/services.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+export const BaseLayer = Layer.mergeAll(
+  makeToolResolverFromFunctions([fn], toolkit),
+  makeToolExecutionServiceFromFunctions(toolkit, toolkit.toLayer({}) as any),
+  CredentialsService.layerFromDatabase(),
+  TracingService.layerNoop,
+).pipe(
+  Layer.provideMerge(
+    TestDatabaseLayer({
```
dmaretskyi, line 81: pass-in the real database layer

packages/devtools/cli-tui/src/core/services.ts
```diff
@@ -0,0 +1,132 @@
...
+      types: [],
+    }),
+  ),
+  Layer.provideMerge(
+    FunctionInvocationServiceLayerTestMocked({
```
dmaretskyi, line 86: Use the real layer - see https://github.com/dxos/dxos/blob/64c22d9ed5eb6d6f102d323f5bafa3f8b0e8ce19/packages/plugins/plugin-automation/src/capabilities/compute-runtime.ts

packages/core/ai/src/resolvers/ChatCompletionsAdapter.ts
richburdon: no -- they changed the streaming protocol and it's no longer compatible with LM studio or ollama

packages/devtools/cli-tui/src/main.ts
richburdon: i think we'll integrate with cli-next and make this tui just the REPL mode

packages/devtools/cli-tui/src/core/services.ts
richburdon: i tried that but couldn't get it to work

packages/core/ai/src/resolvers/anthropic/AnthropicResolver.ts
richburdon: i think we should remove the map -- and instead the provider should advertise what models are available.

PR #10272, opened 2025-12-08
feat: Don't announce documents that don't need syncing

dmaretskyi:


packages/core/echo/echo-pipeline/src/automerge/automerge-repo.test.ts
```diff
@@ -510,6 +512,37 @@ describe('AutomergeRepo', () => {
       await asyncTimeout(handleB.whenReady(), 1000);
       expect(handleB.doc()!.text).to.equal(text);
     });
+
+    test.only('retry share config', async () => {
```
wittjosiah: .only 👀

packages/core/echo/echo-pipeline/src/edge/echo-edge-replicator.ts
```diff
@@ -107,6 +108,7 @@ export class EchoEdgeReplicator implements EchoReplicator {
   }
 
   async connectToSpace(spaceId: SpaceId): Promise<void> {
+    log.info('connectToSpace', { spaceId });
```
wittjosiah: .info 👀

packages/core/echo/echo-pipeline/src/space/space-protocol.ts
```diff
@@ -294,7 +294,9 @@ export class SpaceProtocolSession implements WireProtocol {
       }),
     );
 
+    console.log('disableP2pReplication', this._disableP2pReplication);
```
wittjosiah: console.log 👀

wittjosiah: I noticed the lint rule for this one disappeared at some point, was that intentional?

dmaretskyi: unintentional

PR #10277, opened 2025-12-09
fix: add restricted mode to gmail sync

wittjosiah:


packages/plugins/plugin-inbox/src/functions/google/gmail/sync.ts
```diff
@@ -240,6 +269,8 @@ const streamGmailMessagesToQueue = Effect.fn(function* (
       }
       return !isDuplicate;
     }),
+    // In restricted mode, limit to N oldest messages.
+    restricted ? Stream.take(STREAMING_CONFIG.restrictedMax) : Function.identity,
```
dmaretskyi, line 273: are the message ids returned in the old to new order?

PR #10282, opened 2025-12-10
ci: split e2e tests into browser and node variants

mykola-vrmchk:


packages/plugins/plugin-script/.storybook/main.mts
```diff
@@ -2,7 +2,7 @@
 // Copyright 2025 DXOS.org
 //
 
-import { createConfig } from '../../../../tools/storybook/.storybook/main';
+import { createConfig } from '../../../../tools/storybook/.storybook/main.ts';
```
wittjosiah: why's this needed?

PR #10283, opened 2025-12-10
feat: CLI App bootstrap

richburdon:


packages/sdk/client/src/echo/util.ts
```diff
@@ -50,7 +50,7 @@ export const SPACE_ID_LENGTH = 33;
 export const OBJECT_ID_LENGTH = 26;
 export const FQ_ID_LENGTH = SPACE_ID_LENGTH + OBJECT_ID_LENGTH + 1;
 
-/** @deprecated Use Key.SpaceId schema. */
+/** @deprecated Use SpaceId.parse or SpaceId.tryParse. */
```
wittjosiah: revert this one, I ended up not adding those methods

PR #10288, opened 2025-12-11
feat: add Obj.getDatabase api

wittjosiah:


packages/core/echo/echo/src/Obj.ts
```diff
@@ -289,6 +291,17 @@ export const getTypename = (entity: Entity.Unknown): string | undefined => {
...
 
+//
+// Database
+//
+
+/**
+ * Get the database the object belongs to.
+ */
+export const getDatabase = (entity: Entity.Unknown): Database.Database | undefined => {
+  return (entity as any)[ObjectDatabaseId];
```
dmaretskyi: there was a helper function to cast things to InternalObjectProps

dmaretskyi (approved):

packages/core/echo/echo/src/Database.ts
```diff
@@ -120,10 +120,14 @@ export type FlushOptions = {
   updates?: boolean;
 };
 
+export const __DatabaseId = Symbol.for('@dxos/echo/Database');
```
dmaretskyi:
```suggestion
export const TypeId = Symbol.for('@dxos/echo/Database');
export type TypeId = typeof TypeId;
```

so its `Database.TypeId`

packages/core/echo/echo/src/Database.ts
```diff
@@ -120,10 +120,14 @@ export type FlushOptions = {
...
+
 /**
  * ECHO Database interface.
  */
 export interface Database extends Queryable {
+  readonly [__DatabaseId]: true;
```
dmaretskyi:
```suggestion
  readonly [__DatabaseId]: __DatabaseId;
```

PR #10289, opened 2025-12-11
feat: disallow node globals (DX-692)

dmaretskyi:
Implements DX-692 by configuring ESLint to disallow Node.js globals (Buffer, process) and enforcing the node: protocol for imports.


dmaretskyi:
@cursor-bot please fix the CI failures

PR #10291, opened 2025-12-11
feat(cli-next): add recover command & fix tool availability in chat

wittjosiah:


packages/devtools/cli-next/src/commands/chat/types.ts
```diff
@@ -2,6 +2,38 @@
 // Copyright 2025 DXOS.org
 //
 
+import { Blueprint } from '@dxos/blueprints';
```
richburdon, line 5: probably time to have a folder for cross-cutting components like this and the processor?

packages/devtools/cli-next/src/commands/halo/join/join.ts
```diff
@@ -23,8 +24,8 @@ export const join = Command.make(
     invariant(!client.halo.identity.get(), 'Identity already exists');
 
     const identity = yield* sendInvitation({ client, invitationCode });
-    yield* Effect.log('Identity key:', identity?.identityKey?.toHex());
-    yield* Effect.log('Display name:', identity?.profile?.displayName);
+    yield* Console.log('Identity key:', identity?.identityKey?.toHex());
```
richburdon, line 27: what's our strategy for Console vs dxos/log?

packages/devtools/cli-next/src/commands/dx.ts
```diff
@@ -23,7 +23,7 @@ import { object } from './object';
 import { profile } from './profile';
 import { queue } from './queue';
 import { repl } from './repl';
```
richburdon, line 25: does repl eventually get rewritten using App.tsx, etc?

packages/devtools/cli-next/src/commands/halo/join/join.ts
wittjosiah: log should be used for debugging type information, console should be used when printing things for the user of the cli as it doesn't include extra debug info cruft

PR #10295, opened 2025-12-12
feat: Enable Functions to use Edge Ai service

mykola-vrmchk:
- **feat(plugin-script): migrate ping template to use defineFunction**
- **feat(functions): add FunctionsAiHttpClient for AI service proxy**
- **feat(protocols): add AI service to function context and edge environment**
- **feat(functions): integrate AI service into function context with Anthropic resolver**
- **feat(plugin-script): add @effect/ai-anthropic dependency**
- **feat(functions): rename aiService to functionsService in function context**
- **feat(functions): update FunctionsAiHttpClient to use aiFetch and improve error handling**
- **feat(plugin-script): add Anthropic AI template and e2e tests**
- **feat(functions): rename AiService to FunctionService and update AI fetch interface**

part of DX-688


packages/core/functions/src/protocol/protocol.ts
```diff
@@ -146,10 +149,31 @@ class FunctionContext extends Resource {
...
 
-    return Layer.mergeAll(dbLayer, queuesLayer, credentials, functionInvocationService, aiService, tracing);
+    const aiLayer = this.context.services.functionsService
+      ? AiModelResolver.AiModelResolver.buildAiService.pipe(
+          Layer.provide(
+            AnthropicResolver.make().pipe(
+              Layer.provide(
+                AnthropicClient.layer({
+                  // TODO(dmaretskyi): Read endpoint from config/settings.
+                  apiUrl: 'https://ai-service.dxos.workers.dev/provider/anthropic',
```
dmaretskyi: does it matter what endpoint it is since you're passing the request through the service binding?
I'd say its better to put it as `http://internal/....` if the hostname doesn't count anyway

packages/core/protocols/src/edge/EdgeFunctionEnv.ts
```diff
@@ -88,6 +89,16 @@ export interface QueueService {
...
 }
 
+/**
+ * FunctionService API for other CF services like functions.
+ */
+export interface FunctionService {
+  /**
+   * Enables proxying HTTP requests to the AI service from other workers.
+   */
+  aiFetch(request: Request): Promise<Response>;
```
dmaretskyi: I think entrypoints in CF have a standard `fetch` method for handling requests. Can you use that?

packages/plugins/plugin-script/src/templates/anthropic.ts
```diff
@@ -0,0 +1,27 @@
...
+  name: 'Anthropic Chat',
+  description: 'Chat with Anthropic',
+  inputSchema: Schema.Struct({
+    message: Schema.String,
+  }),
+  services: [AiService.AiService],
+  handler: ({ data }) => {
+    return Effect.gen(function* () {
+      const model = (yield* AiService.AiService).model('@anthropic/claude-sonnet-4-5');
+      const { text } = yield* LanguageModel.generateText({ prompt: data.message }).pipe(Effect.provide(model));
```
dmaretskyi, line 28: test streaming too pls

dmaretskyi (changes requested):
.

PR #10323, opened 2025-12-17
feat(cli-next): interactive command fallbacks (+ cleanup)

wittjosiah:
- interactive trigger commands
- interactive function import
- move trigger/function out of edge sub-command
- flush and sync for commands which update the database
- combine the multiple function list commands & trigger list commands
- use json flag more consistently across commands
- default space id more consistently across commands
- client.spaces.waitUntilReady throws if there's no identity

### DIFF
--- a/packages/devtools/cli-next/src/commands/function/import.ts
+++ b/packages/devtools/cli-next/src/commands/function/import.ts
@@ -24,7 +24,7 @@ export const importCommand = Command.make(
   'import',
   {
     spaceId: Common.spaceId.pipe(Options.optional),
-    key: Args.text({ name: 'key' }).pipe(Args.withDescription('The key of the function to import.'), Args.optional),
+    key: Args.text({ name: 'key' }).pipe(Args.withDescription('Function key'), Args.optional),
   },
   ({ key }) =>
     Effect.gen(function* () {
--- a/packages/devtools/cli-next/src/commands/function/list.ts
+++ b/packages/devtools/cli-next/src/commands/function/list.ts
@@ -23,7 +23,7 @@ export const list = Command.make(
   {
     spaceId: Common.spaceId.pipe(Options.optional),
     remote: Options.boolean('remote').pipe(
-      Options.withDescription('Query functions from remote EDGE service instead of local database.'),
+      Options.withDescription('Query EDGE service (defaults to local)'),
       Options.withDefault(false),
     ),
   },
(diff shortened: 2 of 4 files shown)


packages/devtools/cli-next/src/commands/function/import.ts
```diff
@@ -0,0 +1,86 @@
...
+import { flushAndSync, printList, spaceLayer } from '../../util';
+import { Common } from '../options';
+
+import { getFunctionStatus, printFunction, selectDeployedFunction } from './util';
+
+export const importCommand = Command.make(
+  'import',
+  {
+    spaceId: Common.spaceId.pipe(Options.optional),
+    key: Args.text({ name: 'key' }).pipe(Args.withDescription('The key of the function to import.'), Args.optional),
```
richburdon: I think we could be briefer in general. And I think the general pattern is no period for attributes.

"Function key"

packages/devtools/cli-next/src/commands/function/list.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Common } from '../options';
+
+import { getFunctionStatus, printFunction } from './util';
+
+export const list = Command.make(
+  'list',
+  {
+    spaceId: Common.spaceId.pipe(Options.optional),
+    remote: Options.boolean('remote').pipe(
+      Options.withDescription('Query functions from remote EDGE service instead of local database.'),
```
richburdon: Query EDGE service (defaults to local).

packages/devtools/cli-next/src/commands/trigger/update/subscription.ts
```diff
@@ -0,0 +1,269 @@
...
+  trigger: Trigger.Trigger,
+  fn: Function.Function,
+  inputOption: Option.Option<Record<string, any>>,
+) {
+  const currentInput = trigger.input as Record<string, any> | undefined;
+  const currentInputStr = currentInput ? JSON.stringify(currentInput) : 'none';
+  const shouldChangeInput = yield* Option.match(inputOption, {
+    onNone: () =>
+      Prompt.confirm({
+        message: `Change the input (current: ${currentInputStr})?`,
```
richburdon: Change input?\n
{}

or

Change input? [xxx]

i.e., keep text completely separate from data

packages/devtools/cli-next/src/commands/trigger/update/timer.ts
```diff
@@ -0,0 +1,183 @@
...
+    const foundFn = functions.find((fn) => fn.id === functionId);
+    if (!foundFn || !Obj.instanceOf(Function.Function, foundFn)) {
+      return yield* Effect.fail(new Error(`Function not found: ${functionId}`));
+    }
+    currentFn = foundFn;
+    trigger.function = Ref.make(currentFn);
+  }
+
+  if (!currentFn) {
+    return yield* Effect.fail(new Error('Trigger has no function reference'));
```
richburdon: Invalid reference for ${id}

PR #10376, opened 2026-01-01
feat!: operations

wittjosiah:


packages/plugins/plugin-automation/src/capabilities/operation-resolver/operation-resolver.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { AutomationOperation } from '../../types';
+
+export default Capability.makeModule((context) =>
+  Effect.succeed(
+    Capability.contributes(Common.Capability.OperationResolver, [
+      OperationResolver.make({
+        operation: AutomationOperation.CreateTriggerFromTemplate,
+        handler: ({ db, template, enabled = false, scriptName, input }) =>
+          Effect.gen(function* () {
```
dmaretskyi, line 23: Stylisticly, I'd urge to use `Effect.fnUntraced(function* (args) { ... })` instead of `(args) => Effect.gen(function* () { ... })`, but its not a huge deal, perhaps we can make a linter plugin

packages/plugins/plugin-client/src/capabilities/operation-resolver/operation-resolver.ts
```diff
@@ -0,0 +1,241 @@
...
+      OperationResolver.make({
+        operation: ClientOperation.CreateIdentity,
+        handler: (profile) =>
+          Effect.gen(function* () {
+            const manager = context.getCapability(Common.Capability.PluginManager);
+            const client = context.getCapability(ClientCapabilities.Client);
+            const scheduler = yield* FollowupScheduler.Service;
+            const data = yield* Effect.promise(() => client.halo.createIdentity(profile));
+            yield* Effect.promise(() => runAndForwardErrors(manager.activate(ClientEvents.IdentityCreated)));
+            yield* scheduler.schedule(ObservabilityOperation.SendEvent, { name: 'identity.create' });
```
dmaretskyi, line 40: do i understand correctly that this runs another operation after this one?

another stylistic thing would be adding a shorthand: `yield* FollowupScheduler.schedule(...)`

packages/plugins/plugin-automation/src/capabilities/operation-resolver/operation-resolver.ts
```diff
@@ -0,0 +1,74 @@
...
+                  queue: (template.queueDXN as DXN).toString(),
+                };
+                break;
+              }
+              default: {
+                break;
+              }
+            }
+
+            yield* invoke(SpaceOperation.AddObject, {
```
dmaretskyi, line 61: hmm, how is this different from FolloupScheduler? can we combine them into one thing:

Invoke operation: now/later, wait/don't wait

packages/plugins/plugin-automation/src/capabilities/operation-resolver/operation-resolver.ts
```diff
@@ -0,0 +1,74 @@
...
+import { AutomationOperation } from '../../types';
+
+export default Capability.makeModule((context) =>
+  Effect.succeed(
+    Capability.contributes(Common.Capability.OperationResolver, [
+      OperationResolver.make({
+        operation: AutomationOperation.CreateTriggerFromTemplate,
+        handler: ({ db, template, enabled = false, scriptName, input }) =>
+          Effect.gen(function* () {
+            const { invoke } = context.getCapability(Common.Capability.OperationInvoker);
```
dmaretskyi, line 24: is the eventual goal, to not have to pass-in context?

PR #10380, opened 2026-01-05
feat: Mosaic DND

richburdon:


package.json
```diff
@@ -113,7 +113,7 @@
     "eslint-plugin-prettier": "catalog:",
     "eslint-plugin-react": "catalog:",
     "eslint-plugin-sort-exports": "catalog:",
-    "eslint-plugin-storybook": "10.1.10",
+    "eslint-plugin-storybook": "10.1.11",
```
wittjosiah: all these version bumps should happen in the catalog, not here

scripts/check-package-cycles.mjs
wittjosiah, line 1: I'm pretty sure that moon compile task will fail if there's a package cycle already, or is this checking for something more specific?

packages/ui/react-ui-mosaic/src/components/Focus/Focus.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const FOCUS_STATE_ATTR = 'focus-state';
+
+type ContextValue = {
+  setFocus?: (state: FocusState | undefined) => void;
+};
+
+const FocusContext = createContext<ContextValue>({});
+
+const useFocus = () => useContext(FocusContext);
```
wittjosiah, line 35: we've generally made the types of the context required and had an invariant in the hook. One exception I can think of is attention where it serves as a progressive enhancement if its available, is that the intention here as well?

packages/ui/react-ui-mosaic/src/components/Focus/Focus.tsx
```diff
@@ -0,0 +1,91 @@
...
+> & { className?: string };
+
+const Group = forwardRef<HTMLDivElement, GroupProps>(
+  ({ classNames, className, children, asChild, axis = 'vertical', ...props }: GroupProps, forwardedRef) => {
+    const rootRef = useRef<HTMLDivElement>(null);
+    const composedRef = useComposedRefs<HTMLDivElement>(rootRef, forwardedRef);
+    const Root = asChild ? Slot : Primitive.div;
+
+    const focusableGroupAttrs = useFocusableGroup({ tabBehavior: 'limited-trap-focus' });
+    const arrowNavigationAttrs = useArrowNavigationGroup({ axis, memorizeCurrent: true });
```
wittjosiah: might be useful for coming back to this later to note why `limit-trap-focus` and `memoizeCurrent` are specified as such

packages/ui/react-ui-mosaic/src/components/Focus/Focus.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Root
+          role='none'
+          tabIndex={0}
+          className={mx(styles.container.root, className, classNames)}
+          {...tabsterAttrs}
+          {...(state && {
+            [`data-${FOCUS_STATE_ATTR}`]: state,
+          })}
+          {...props}
+          ref={composedRef}
```
wittjosiah, line 76: I've seen elsewhere and followed inconsistently the pattern of always putting ref first, is there a reason needs to be last here?

packages/ui/react-ui-mosaic/moon.yml
```diff
@@ -9,9 +9,7 @@ tasks:
   compile:
     args:
       - '--entryPoint=src/index.ts'
+      - '--entryPoint=src/testing/index.ts'
       - '--platform=browser'
       - '--platform=node'
       - '--preactSignalTracking'
-  test:
-    inputs:
-      - .storybook/*
```
wittjosiah, line 17: why remove?

packages/ui/react-ui-mosaic/src/components/Mosaic/Mosaic.tsx
```diff
@@ -45,166 +96,687 @@ const [MosaicContextProvider, useMosaicContext] = createContext<MosaicContextVal
...
 //
 // Mosaic
 //
 
-export const Mosaic = { Root, Container };
+export const Mosaic = {
+  Root,
+  Container,
+  ContainerInfo,
+  Cell,
```
wittjosiah: This should probably be `Tile` as a piece of the mosaic?

scripts/check-package-cycles.mjs
richburdon: this is dima's?

packages/ui/react-ui-mosaic/src/components/Focus/Focus.tsx
richburdon: it IS progressive here, so undefined is valid

packages/ui/react-ui-mosaic/src/components/Focus/Focus.tsx
richburdon: my pattern was always ref first, and will's was ref last (which seems to be preferred by other libs).

PR #10902, opened 2026-04-03
feat(tracing): W3C trace context propagation across RPC, lifecycle spans, remove custom span layer

mykola-vrmchk:
## Summary

Adds connected distributed tracing across RPC boundaries and replaces the custom span system with direct OTEL backend calls.

### W3C trace context propagation across RPC

`ContextRpcCodec` (in `@dxos/context`) encodes/decodes W3C `traceparent`/`tracestate` on every RPC call. Hardcoded in `RpcPeer` — no configuration needed. Incoming handlers receive `options.ctx` with the caller's trace identity already decoded, so `@trace.span()` methods on the service side become children of the caller's span.

### TracingBackend interface replaces custom spans

Deleted `TracingSpan`, `RemoteTracing` buffering layer, `TraceSender` RPC service, and the devtools `TracingPanel`. The `@trace.span()` decorator now calls `TRACE_PROCESSOR.tracingBackend.startSpan()` directly, which returns a `RemoteSpan` (`end()` / `setError()` / `spanContext`). The OTEL backend registers itself at startup via `TRACE_PROCESSOR.tracingBackend`.

Key design: `TRACE_SPAN_ATTRIBUTE` on DXOS `Context` stores `TraceContextData` — plain W3C strings, not live OTEL objects. They remain valid after the originating span ends, cross RPC/HTTP boundaries without inject/extract, and need no retention cache.

### Lifecycle spans

`@trace.resource({ lifecycle: true })` wraps `open()`/`close()` with a long-lived span. `this._ctx` carries the lifecycle span's trace context, so background work (subscriptions, timers) parents correctly under the lifecycle span rather than appearing as orphaned roots.

### Context propagation plumbing

- Service methods across `client-services`, `echo-pipeline`, `edge-client`, and `client` updated to accept and forward `ctx: Context`.
- RPC handlers read `options?.ctx ?? Context.default()` instead of creating bare `Context.default()`.
- Several `Resource` subclasses now use `lifecycle: true` to prevent orphaned internal spans.

### What's preserved

- `showInBrowserTimeline` → still calls `performance.measure()`.
- `showInRemoteTracing = false` → skips `startSpan()`, grandchild spans reconnect to grandparent.
- Resource tracking (`@trace.resource`, `@trace.info`, `@trace.metricsCounter`) unchanged.
- Diagnostics system unchanged.

### Deleted

- `packages/common/tracing/src/remote/tracing.ts` (RemoteTracing)
- `packages/common/tracing/src/trace-sender.ts` (TraceSender)
- `packages/devtools/devtools/src/panels/client/TracingPanel/` (8 files)
- `packages/devtools/devtools/src/components/performance/panels/SpansPanel.tsx`
- `TracingService`, `StreamTraceEvent`, `Span` protobuf messages from `tracing.proto`


mykola-vrmchk:
@coderabbit review

mykola-vrmchk:
@coderabbit review

PR #11008, opened 2026-04-15
test(AiPreprocessor): enchance failure recovery in prompt preprocessor

dmaretskyi:
- Updated existing tests to handle multiple messages and tool call scenarios, ensuring accurate processing of assistant messages with tool calls.
- Added a new test to verify that tool calls without results are dropped correctly, maintaining the integrity of the assistant's message content.
- Adjusted the logic in the preprocessPrompt function to remove unsatisfied server tool calls and improve message filtering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deduplicate duplicate tool results, keeping the first entry.
  * Drop unsatisfied server-executed tool calls and remove orphan tool-result messages appearing before their calls.
  * Ensure process failure handling sets status to FAILED before running cleanup.

* **Tests**
  * Added tests for tool call/result deduplication and omission behaviors.
  * Added tests for process completion: success, failure (with updated error text), and interruption.


dmaretskyi:
@cursoragent fix failures so I can merge this

PR #11018, opened 2026-04-17
feat(plugin-trello): Trello board and card sync integration

chad:
Bidirectional sync of Trello boards, lists, and cards into ECHO. Includes card article UI, poll sync, and Trello REST API bridge.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added Trello plugin for integrating Trello boards into your workspace
  * Sync Trello boards and cards to your local database
  * View and edit Trello card details including descriptions, labels, and due dates
  * Push card updates back to Trello
  * Optional Kanban view of board contents
  * Trello API credential configuration support


packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
```diff
@@ -0,0 +1,309 @@
...
+  }, [board.kanbanId, db, invokePromise]);
+
+  const cardItems = useMemo(
+    () => boardCards
+      .sort((cardA, cardB) => (cardA.position ?? 0) - (cardB.position ?? 0))
+      .map((card) => ({ card })),
+    [boardCards],
+  );
+
+  const handleSync = useCallback(async () => {
```
richburdon: sync logic should be moved to an operation -- independent of UI.
see plugin-inbox for reference: there we have "generic" Google API requesters (which are provided capabilities form a generic OAuth workflow) and "mappers" that conform the results form Google APIs onto Composer types. The sync ops can then be triggered by the agent and the UI menus.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
```diff
@@ -0,0 +1,157 @@
...
+    } catch (error) {
+      log.catch(error);
+      setPushStatus('Failed to push');
+    } finally {
+      setPushing(false);
+    }
+  }, [board, card]);
+
+  return (
+    <div className='flex flex-col gap-4 p-4'>
```
richburdon: container components should try to reuse existing Composer UI patterns (e.g., Panel, Column, Form, etc.) most containers should not have custom DOM elements or styles/classNames.

packages/plugins/plugin-trello/src/operations/sync-board.ts
```diff
@@ -0,0 +1,137 @@
...
+        const members = remoteCard.members?.map((member) => ({
+          trelloId: member.id,
+          fullName: member.fullName,
+          username: member.username,
+          avatarUrl: member.avatarUrl ?? undefined,
+        }));
+
+        const existing = existingCardsByTrelloId.get(remoteCard.id);
+        if (existing) {
+          Obj.change(existing, (mutable) => {
```
richburdon: this seems duplicative with the sync logic above (in the component)

packages/plugins/plugin-trello/src/operations/trello-api.ts
```diff
@@ -0,0 +1,121 @@
...
+  members: Schema.optional(Schema.Array(TrelloMemberResponse)),
+  url: Schema.optional(Schema.String),
+  closed: Schema.optional(Schema.Boolean),
+  dateLastActivity: Schema.optional(Schema.String),
+});
+
+export type TrelloCardResponse = Schema.Schema.Type<typeof TrelloCardResponse>;
+
+/** Fetches board metadata. */
+export const getBoard = (boardId: string, auth: TrelloAuth) =>
```
richburdon, line 78: ideally we would embrace and extend the existing Kanban types and make this generic, where the sync/mapper conforms to Composer types to the Trello objects (again see plugin-inbox for mapping Google APIs to Composer Message and Event types).

packages/plugins/plugin-trello/src/types/Trello.ts
```diff
@@ -0,0 +1,158 @@
...
+
+import * as Schema from 'effect/Schema';
+
+import { Annotation, Obj, Type } from '@dxos/echo';
+import { FormInputAnnotation, LabelAnnotation } from '@dxos/echo/internal';
+
+// @import-as-namespace
+
+/** Trello label attached to a card. */
+export const Label = Schema.Struct({
```
richburdon, line 18: "Labels" are implemented as "Tags" in Composer; this is a cross cutting feature across all objects. Although there is an open issue re how to map (2-way) third-party tags/labels onto a shared Composer ontology.

packages/plugins/plugin-trello/src/types/Trello.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * TrelloCard schema representing a card synced from Trello.
+ */
+export const TrelloCard = Schema.Struct({
+  /** Card name/title. */
+  name: Schema.String,
+  /** Card description (markdown). */
+  description: Schema.optional(Schema.String),
+  /** Trello card ID for sync. */
+  trelloCardId: Schema.String.pipe(FormInputAnnotation.set(false)),
```
richburdon: Each ECHO object has a meta object that contains foreignKeys which are intended to track external identifiers for 2-way sync

chad:
@richburdon take a look when you get a moment.  it seems like it understood the assignment

packages/plugins/plugin-trello/src/capabilities/app-graph-builder.ts
```diff
@@ -0,0 +1,77 @@
...
+                    return Operation.invoke(LayoutOperation.AddToast, {
+                      id: 'sync-board-error',
+                      icon: 'ph--warning--regular',
+                      duration: 5_000,
+                      title: ['sync-board-error.title', { ns: meta.id }],
+                      closeLabel: ['close.label', { ns: meta.id }],
+                    });
+                  }),
+                );
+              }),
```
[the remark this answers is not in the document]
chad: Fixed in 82a670ab0a — capability lookup now runs inside the Effect.gen so setup failures go through the same catchAll that raises the error toast.

packages/plugins/plugin-trello/src/capabilities/poll-sync.ts
```diff
@@ -0,0 +1,18 @@
...
+import { log } from '@dxos/log';
+
+/**
+ * Placeholder startup module for Trello polling sync.
+ * TODO(trello): Implement reactive subscription to TrelloBoard objects to auto-start polling.
+ */
+export default Capability.makeModule(
+  Effect.fnUntraced(function* () {
+    log('trello poll-sync module loaded');
+  }),
```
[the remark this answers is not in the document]
chad: Fixed in 82a670ab0a — removed the stub module rather than landing an empty polling loop. Will re-add once there's a real reactive/backoff implementation to ship.

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
```diff
@@ -0,0 +1,309 @@
...
+  const [viewport, setViewport] = useState<HTMLElement | null>(null);
+  const { invokePromise } = useOperationInvoker();
+
+  const handleOpenKanban = useCallback(() => {
+    if (!board.kanbanId || !db) {
+      return;
+    }
+    const kanbanPath = getTypePath(db.spaceId, 'org.dxos.type.trelloCard', board.kanbanId);
+    void invokePromise(LayoutOperation.Open, { subject: [kanbanPath] });
+  }, [board.kanbanId, db, invokePromise]);
```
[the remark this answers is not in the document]
chad: Fixed in b1797b7b8e — handleOpenKanban now uses `Kanban.Kanban.typename` so the deep-link path matches the object it points at.

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
```diff
@@ -0,0 +1,309 @@
...
+        fetch(`${TRELLO_API_BASE}/boards/${board.trelloBoardId}?${auth}`),
+        fetch(`${TRELLO_API_BASE}/boards/${board.trelloBoardId}/lists?filter=open&${auth}`),
+        fetch(`${TRELLO_API_BASE}/boards/${board.trelloBoardId}/cards?members=true&filter=open&${auth}`),
+      ]);
+
+      const [boardInfo, remoteLists, remoteCards] = await Promise.all([
+        boardRes.json(),
+        listsRes.json(),
+        cardsRes.json(),
+      ]);
```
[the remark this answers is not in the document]
chad: Outdated after rebase — the component now invokes the SyncBoard operation instead of calling fetch() itself, and `trello-api.ts` (the operation's client) already checks status via `ensureOk` before parsing JSON.

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
```diff
@@ -0,0 +1,309 @@
...
+          }));
+        }
+      }
+
+      const remoteIds = new Set(remoteCards.map((rc: any) => rc.id));
+      for (const [trelloId, existing] of existingByTrelloId) {
+        if (!remoteIds.has(trelloId)) {
+          Obj.change(existing, (mutable) => { mutable.closed = true; });
+        }
+      }
```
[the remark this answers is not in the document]
chad: Outdated after rebase — the component no longer reconciles cards; the operation handles that and is now scoped to the board (see 23f6f5d5db).

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+      if (!board.kanbanId && db) {
+        try {
+          const { view } = await ViewModel.makeFromDatabase({
+            db,
+            typename: 'org.dxos.type.trelloCard',
+            pivotFieldName: 'listName',
+            fields: ['name', 'description', 'listName', 'dueDate'],
+            createInitial: 0,
+          });
```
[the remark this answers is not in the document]
chad: Fixed in 23f6f5d5db — moved Kanban view creation into the SyncBoard operation, and the underlying `ViewModel.makeFromDatabase` still creates a view over all TrelloCards. Added a TODO noting multi-board spaces need a board-scoped filter; flagging as a follow-up since the ViewModel API doesn't yet expose per-ref filtering.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  // Find the board that owns this card (to get API credentials).
+  const db = Obj.getDatabase(card);
+  const boards: Trello.TrelloBoard[] = useQuery(db, Filter.type(Trello.TrelloBoard));
+  const board = boards[0];
+
+  const handlePush = useCallback(async () => {
+    if (!board?.apiKey || !board?.apiToken) {
+      setPushStatus('No API credentials found');
+      return;
```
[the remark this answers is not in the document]
chad: Fixed in b1797b7b8e — TrelloCardArticle now prefers `card.board?.target` over `boards[0]`. Added a new optional `board: Ref<TrelloBoard>` field to TrelloCard (populated on every sync). Fallback heuristic for older cards never picks a no-credentials board if a credentialed one exists.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
```diff
@@ -0,0 +1,157 @@
...
+                </span>
+              ))}
+            </div>
+          )}
+        </div>
+        <div className='flex items-center gap-2 text-sm'>
+          <button
+            onClick={handlePush}
+            disabled={pushing || !board?.apiKey}
+            className='rounded bg-blue-600 px-3 py-1 text-white hover:bg-blue-700 disabled:opacity-50'
```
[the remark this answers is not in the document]
chad: Already addressed — the button's `disabled` is bound to `!board?.accessToken`, i.e. the single credential reference. The old dual-field `apiKey/apiToken` state was removed when the schema switched to AccessToken refs.

packages/plugins/plugin-trello/src/operations/sync-board.ts
```diff
@@ -0,0 +1,137 @@
...
+      const db = Obj.getDatabase(board);
+      if (!db) {
+        return yield* Effect.fail(new Error('Board has no database'));
+      }
+      const allCards: Trello.TrelloCard[] = yield* Effect.promise(() =>
+        db.query(Filter.type(Trello.TrelloCard)).run(),
+      );
+      const existingCardsByTrelloId = new Map(
+        allCards.map((card) => [card.trelloCardId, card]),
+      );
```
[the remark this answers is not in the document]
chad: Fixed in 23f6f5d5db — `existingCardsByTrelloId` now starts from `boardOwnedCards` (filtered by `card.board` ref or fallback to trelloListId ∈ this board's lists). Closing a missing card is also guarded to skip cards already closed, so no cross-board writes.

packages/plugins/plugin-trello/src/operations/trello-api.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/** Creates a new card on Trello. */
+export const createCard = (listId: string, name: string, auth: TrelloAuth) =>
+  Effect.gen(function* () {
+    const client = yield* HttpClient.HttpClient;
+    const request = HttpClientRequest.post(`${TRELLO_API_BASE}/cards`);
+    const response = yield* client.execute(withAuth(HttpClientRequest.setUrlParams(request, { idList: listId, name }), auth));
+    const json = yield* response.json;
+    return yield* Schema.decodeUnknown(TrelloCardResponse)(json);
+  });
```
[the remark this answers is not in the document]
chad: Addressed differently — `trello-api.ts` has an `ensureOk` helper (line 19-26) invoked after every `client.execute` and before `response.json`. Equivalent effect to `HttpClient.filterStatusOk`. Happy to switch to the HttpClient pipe-style if preferred.

packages/plugins/plugin-trello/src/types/Trello.ts
```diff
@@ -0,0 +1,158 @@
...
+  /** Board name. */
+  name: Schema.optional(Schema.String),
+  /** Trello board ID for sync. */
+  trelloBoardId: Schema.String.pipe(FormInputAnnotation.set(false)),
+  /** Board URL on Trello. */
+  url: Schema.optional(Schema.String.pipe(FormInputAnnotation.set(false))),
+  /** Trello API key. */
+  apiKey: Schema.optional(Schema.String.pipe(FormInputAnnotation.set(false))),
+  /** Trello API token. */
+  apiToken: Schema.optional(Schema.String.pipe(FormInputAnnotation.set(false))),
```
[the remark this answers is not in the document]
chad: Outdated — TrelloBoard already stores credentials as `accessToken: Ref<AccessToken>` (see Trello.ts:88-93). No `apiToken` field persisted on the schema.

packages/plugins/plugin-trello/src/types/Trello.ts
```diff
@@ -0,0 +1,180 @@
...
+  description?: string;
+  trelloListId: string;
+  listName?: string;
+  position?: number;
+  url?: string;
+  closed?: boolean;
+  lastActivityAt?: string;
+  labels?: Label[];
+  members?: Member[];
+}): TrelloCard => {
```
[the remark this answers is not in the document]
chad: Fixed in 23f6f5d5db — added `dueDate` and `dueComplete` to makeCard props.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
```diff
@@ -0,0 +1,149 @@
...
+            icon='ph--cloud-arrow-up--regular'
+            iconOnly
+            disabled={pushing || !board?.accessToken}
+            onClick={handlePush}
+          />
+          {pushStatus && <Toolbar.Text>{pushStatus}</Toolbar.Text>}
+          {card.url && (
+            <a href={card.url} target='_blank' rel='noreferrer'>
+              <Toolbar.IconButton label='Open in Trello' icon='ph--arrow-square-out--regular' iconOnly />
+            </a>
```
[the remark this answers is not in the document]
chad: Fixed in b1797b7b8e — replaced `<a><Toolbar.IconButton></a>` with `<Toolbar.Link>` wrapping an `<Icon>` in both containers.

packages/plugins/plugin-trello/src/operations/sync-board.ts
```diff
@@ -0,0 +1,168 @@
...
+  Operation.withHandler(({ board: boardRef }) =>
+    Effect.gen(function* () {
+      const board = yield* Database.load(boardRef);
+      const { trelloBoardId, apiKey, apiToken } = board as Trello.TrelloBoard;
+
+      if (!apiKey || !apiToken) {
+        return yield* Effect.fail(new Error('Trello API key and token are required'));
+      }
+
+      const auth = { apiKey, apiToken };
```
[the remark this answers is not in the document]
chad: Fixed in 23f6f5d5db — rewrote the entire handler to use `Trello.getTrelloBoardId(board)`, `Trello.getTrelloCardId(card)`, and `TrelloCredentials.fromBoard(boardRef)`. No more destructured fields that don't exist on the schema.

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
chad: Fixed — the sync operation now does all the work (23f6f5d5db). Component just dispatches `SyncBoard`. `TrelloCredentials.fromBoard` is the Trello equivalent of plugin-inbox's `GoogleCredentials.fromMailbox` pattern.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
chad: TrelloCardArticle uses only Composer primitives now (Panel, Toolbar, Input, Tag, Icon) — no custom DOM or classNames remain on the article (b1797b7b8e).

packages/plugins/plugin-trello/src/operations/sync-board.ts
chad: Fixed — Kanban creation moved into the sync operation; component-side duplicate removed in b1797b7b8e. Sync-board is now the single source of truth.

packages/plugins/plugin-trello/src/operations/trello-api.ts
chad: Partial — the Kanban view is built off the TrelloCard type, so cards + Kanban are already linked (board.kanbanId, view over trelloCard typename). Making the *sync mapper* fully generic (TrelloCard→Kanban-schema fields) is deferred: keeping TrelloCard as its own type is what gives us trelloListId / trelloListName / labels / members for round-trip sync. Happy to revisit if you have a concrete generalization in mind.

packages/plugins/plugin-trello/src/types/Trello.ts
chad: Tracked — TODO at Trello.ts:13-15 calls out the Labels→Tags mapping + 2-way color mapping as a follow-up. Out of scope for this PR since it's cross-cutting across plugins; want to sequence it after the Composer Tag ontology settles.

packages/plugins/plugin-trello/src/types/Trello.ts
chad: Already addressed — TrelloBoard and TrelloCard both use `Obj.Meta.keys` with `{source: 'trello.com', id}`. See the `getTrelloBoardId`/`getTrelloCardId` helpers at Trello.ts:161-175. The sync operation was rewritten in 23f6f5d5db to consume them.

chad:
## Review response (2026-04-24)

Pass through all 17 comments. Details inline; summary:

**Fixed:**
- **23f6f5d5db** — sync-board operation was reading fields that don't exist on the schema (trelloBoardId, apiKey, apiToken). Rewrote it to use `getTrelloBoardId`, `TrelloCredentials.fromBoard`, and added a `board: Ref<TrelloBoard>` back-ref on TrelloCard so reconciliation is scoped per-board. Also added `dueDate`/`dueComplete` to `makeCard` props and moved Kanban-view creation into the operation.
- **b1797b7b8e** — Kanban typename in `handleOpenKanban`, removed duplicate Kanban creation in TrelloArticle, `card.board?.target` preferred over `boards[0]` in TrelloCardArticle, replaced `<a><Toolbar.IconButton>` with `Toolbar.Link` + `Icon` (invalid HTML → valid).
- **82a670ab0a** — removed stub `poll-sync` capability; wrapped the sync action's capability/runtime lookup inside the catchAll so setup failures raise the same error toast.

**Already addressed in earlier commits:**
- TrelloBoard already uses `accessToken: Ref<AccessToken>` (no field-level `apiToken` leak).
- External IDs already in `Obj.Meta.keys` via the `getTrello*Id` helpers.
- `trello-api.ts` already gates response.json() on `ensureOk` (equivalent to `filterStatusOk`).

**Deferred to follow-ups:**
- Labels → Composer Tags (TODO in-file; cross-cutting across plugins, want to sequence after Tag ontology settles).
- Generic TrelloCard → Kanban field mapper.
- Multi-board-scoped Kanban view (ViewModel doesn't yet expose a ref filter).

packages/plugins/plugin-trello/src/meta.ts
```diff
@@ -0,0 +1,17 @@
+//
+// Copyright 2025 DXOS.org
+//
+
+import { type Plugin } from '@dxos/app-framework';
+import { trim } from '@dxos/util';
+
+export const meta: Plugin.Meta = {
```
richburdon, line 8: plugins should be registered with composer-app

richburdon (commented):
Think deeply about the following instructions:
Consider how we may use the existing plugin-kanban containers and components and make this a headless plugin that just has synchronization operations that interface with Trello.

The plugin may define a Sync object that references the Trello endpoint and has a reference to a Kanban object. The sync mechanism should use the existing Credential mechanism that is used by plugin-mailbox.

We should consider Trello to be the first of many sync mechanisms. Therefore, the Sync object should be generic and something that we may later factor out into a core plugin.

First create a concise design document. 
Create unit tests that sync with a mocked Trello endpoint and creates the Kanban object.

packages/plugins/plugin-trello/src/capabilities/react-surface.tsx
wittjosiah, line 1: note that Kanban is going to expect a card renderer, it looks like one is defined in TrelloArticle but its not exported

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
wittjosiah, line 1: It seems like having both a renderer for the trello board as a whole as well as a kanban view of the cards would fulfill the same purpose, is there anything which this is accomplishing which the kanban view does not?

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
```diff
@@ -0,0 +1,159 @@
...
+          />
+          {pushStatus && <Toolbar.Text>{pushStatus}</Toolbar.Text>}
+          {card.url && (
+            <Toolbar.Link href={card.url} target='_blank' rel='noreferrer' aria-label='Open in Trello'>
+              <Icon icon='ph--arrow-square-out--regular' size={4} />
+            </Toolbar.Link>
+          )}
+        </Toolbar.Root>
+      </Panel.Toolbar>
+      <Panel.Content classNames='p-4 space-y-4'>
```
wittjosiah, line 94: I'd suggest using the `Form` component from `@dxos/react-ui-form` since this is already using react-ui components.

packages/plugins/plugin-trello/src/meta.ts
chad: Done in 3d5aa40 — registered `TrelloPlugin` with `composer-app` (Labs default) via the new dynamic-import + ID-constant pattern in `plugin-defs.tsx`. Also added the workspace dep + composer-app tsconfig reference.

packages/plugins/plugin-trello/src/capabilities/react-surface.tsx
chad: Done in 3d5aa40 — added a `TrelloCard` `card--content` surface (`AppSurface.object(AppSurface.Card, Trello.TrelloCard)`) so plugin-kanban auto-discovers it. The component lives in `src/components/TrelloCard/` (own dir + `index.ts` + `TrelloCard.stories.tsx` per the plugin skill). It renders title + listName + open-in-Trello link — same shape MessageCard/EventCard use in plugin-inbox.

packages/plugins/plugin-trello/src/containers/TrelloArticle/TrelloArticle.tsx
chad: You're right — that custom renderer was a placeholder and the only thing it did beyond Kanban was the manual sync button. Removed in 3d5aa40 (deleted the entire `containers/TrelloArticle/` directory). The sync action now lives on the `TrelloCardArticle` toolbar / operation invoker rather than the board view. Boards render as native Kanban exclusively.

packages/plugins/plugin-trello/src/containers/TrelloCardArticle/TrelloCardArticle.tsx
chad: Agreed in principle. I've sequenced this as a follow-up (tracked as F-6 in PLUGIN.mdl): `TrelloCardArticle` mixes editable fields (name, description) with read-only computed fields (listName, dueDate, members) plus a "Push to Trello" toolbar action. Doing it cleanly with `Form` needs a read-only field annotation that doesn't exist yet — would prefer to land that in a focused PR rather than smuggle it into this one. Happy to bump priority if you'd like it sooner.

chad:
@richburdon — pushed 3d5aa40 to address your latest review:

**Concrete code fixes**
- Registered \`TrelloPlugin\` with composer-app (Labs).
- Added the \`card--content\` surface (@wittjosiah's renderer comment): \`src/components/TrelloCard/TrelloCard.tsx\` with its own dir + storybook, mirroring plugin-inbox's \`MessageCard\` shape.
- Deleted the dead \`TrelloArticle\` (@wittjosiah's "duplicate purpose" point): native Kanban view is the only board renderer.
- \`AppSurface.object(AppSurface.Article|Section|Card, Trello.TrelloCard)\` — adapted to the new typed-Surface API.

**Design doc for the headless / generic-Sync direction**

You asked for a concise design document before the bigger refactor. It's here: \`packages/plugins/plugin-trello/DESIGN.md\` and the spec is in \`packages/plugins/plugin-trello/PLUGIN.mdl\` (per the Claude plugin skill). Short version:

1. **plugin-trello becomes a thin Trello-API → Composer-types mapper.** No board-shaped containers, no scheduler. Just \`mapBoard(...)\` / \`mapCard(...)\`. Most of that is already where it needs to be (\`operations/trello-api.ts\` is the API caller; \`operations/sync-board.ts\` is the mapper).
2. **A generic \`Sync\` ECHO type** — \`{ endpoint, kanban: ref<Kanban>, accessToken: ref<AccessToken>, pollIntervalMs, lastSyncAt }\` — lives in a future \`plugin-sync\` (or core extension). Trello is the first consumer; Linear (#11027) and GitHub (#11024) fit the same shape.
3. **\`TrelloCredentials.fromBoard\`** is already the analogue of plugin-inbox's \`GoogleCredentials.fromMailbox\`. The future refactor lifts it onto \`Sync\` instead of \`TrelloBoard\`.
4. **Tests with a mocked Trello endpoint** belong in the next slice (alongside the \`plugin-sync\` skeleton). Doing them in this PR would mean adding the mock harness here and then ripping it out when the mapper moves; happier landing them with the refactor.

DESIGN.md spells out what's in this PR (mapper isolation already done, headless container removed) vs. what's deferred (Sync extraction, two-way labels↔tags, Form migration). Roadmap entries are tracked as F-4/F-5/F-6 in PLUGIN.mdl so they don't get lost.

Branch was 95 commits behind main; this push also merges \`origin/main\` and adapts to the new \`AppSurface\` API + the TS 6 \`Effect.Effect<T>\` strictness around the credentials service.

PR #11025, opened 2026-04-18
feat(plugin-granola): Granola meeting note sync integration

chad:
## Summary

New plugin for syncing Granola meeting notes into Composer.

- **GranolaSyncRecord** — links a Granola note to a Composer Document via foreignKey (`source: 'granola.ai'`)
- **GranolaAccount** — credentials via `AccessToken` ref (not inline API key)
- Meeting metadata: attendees, calendar event, owner
- Container with sync UI and note browser
- TODO: move sync logic to operation handler, replace custom DOM with Composer UI

## Test plan
- [x] Build passes
- [x] Verified in live Composer session with real Granola API

## Summary by CodeRabbit

* **New Features**
  * Introduces the Granola plugin to sync meeting notes into Composer with automatic initial sync and configurable polling.
  * Adds a dedicated Granola article view with a virtualized list, toolbar controls, and document-opening flows.
  * AI-powered matching between Granola notes and Trello cards, plus generation/saving of match reports.
* **Localization**
  * Adds English locale labels and placeholders for Granola entities.

### DIFF
--- a/packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
+++ b/packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
@@ -100,7 +100,7 @@ const aiMatch = async (
     const doc = record.document?.target;
     const summary = doc?.content?.target?.content?.slice(0, 500) ?? '';
     return {
-      id: record.granolaId,
+      id: Granola.getGranolaId(record),
       noteTitle: doc?.name ?? '',
       meetingTitle: record.calendarEvent?.title ?? '',
       attendees: (record.attendees ?? []).map((attendee) => attendee.name ?? attendee.email ?? '').join(', '),
@@ -176,15 +176,15 @@ const NoteTile = forwardRef<HTMLDivElement, NoteTileProps>(({ data, location, cu
   const doc = record.document?.target;
 
   const handleClick = useCallback(() => {
-    setCurrentId(record.granolaId);
+    setCurrentId(Granola.getGranolaId(record));
     onOpen(record);
   }, [record, onOpen, setCurrentId]);
 
   const published = record.createdAt ? new Date(record.createdAt).toLocaleDateString() : undefined;
 
   return (
-    <Mosaic.Tile asChild classNames='dx-hover dx-current' id={record.granolaId} data={data} location={location}>
-      <Focus.Item asChild current={current} onCurrentChange={() => setCurrentId(record.granolaId)}>
+    <Mosaic.Tile asChild classNames='dx-hover dx-current' id={Granola.getGranolaId(record)} data={data} location={location}>
+      <Focus.Item asChild current={current} onCurrentChange={() => setCurrentId(Granola.getGranolaId(record))}>
         <Card.Root ref={forwardedRef} onClick={handleClick}>
           <Card.Toolbar>
             <Card.IconBlock>
@@ -293,7 +293,7 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
   const handleMatchToTrello = useCallback(async () => {
     setMatching(true);
     try {
-      const uniqueByGranola = [...new Map(syncRecords.map((record) => [record.granolaId, record])).values()];
+      const uniqueByGranola = [...new Map(syncRecords.map((record) => [Granola.getGranolaId(record), record])).values()];
       const activeCards = trelloCards.filter((card) => !card.closed);
       const results = await aiMatch(uniqueByGranola, activeCards);
       setMatches(results);
@@ -340,14 +340,14 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
   }, [db, matches, invokePromise]);
 
   const handleSync = useCallback(async () => {
-    if (!db || !account.apiKey) {
+    if (!db || !account.accessToken?.target?.token) {
       return;
     }
 
     setSyncing(true);
     try {
-      const headers = { Authorization: `Bearer ${account.apiKey}` };
-      const existingByGranolaId = new Map(syncRecords.map((record) => [record.granolaId, record]));
+      const headers = { Authorization: `Bearer ${account.accessToken?.target?.token ?? ''}` };
+      const existingByGranolaId = new Map(syncRecords.map((record) => [Granola.getGranolaId(record), record]));
 
       let cursor: string | undefined;
       let hasMore = true;
@@ -365,6 +365,10 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
         const data = await response.json();
         remoteNotes.push(...data.notes);
         hasMore = data.hasMore;
+        if (hasMore && !data.cursor) {
+          log.warn("granola: pagination stuck — hasMore=true but no cursor");
+          break;
+        }
         cursor = data.cursor ?? undefined;
       }
 
@@ -440,13 +444,13 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
 
   const didAutoSync = useRef(false);
   useEffect(() => {
-    if (!didAutoSync.current && account.apiKey && !account.lastSyncedAt) {
+    if (!didAutoSync.current && account.accessToken?.target?.token && !account.lastSyncedAt) {
       didAutoSync.current = true;
       void handleSync();
     }
-  }, [account.apiKey, account.lastSyncedAt, handleSync]);
+  }, [account.accessToken?.target?.token, account.lastSyncedAt, handleSync]);
 
-  const uniqueRecords = [...new Map(syncRecords.map((record) => [record.granolaId, record])).values()];
+  const uniqueRecords = [...new Map(syncRecords.map((record) => [Granola.getGranolaId(record), record])).values()];
   const sortedRecords = [...uniqueRecords].sort((recordA, recordB) => {
     const dateA = recordA.createdAt ? new Date(recordA.createdAt).getTime() : 0;
     const dateB = recordB.createdAt ? new Date(recordB.createdAt).getTime() : 0;
@@ -473,7 +477,7 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
             label={syncing ? 'Syncing...' : 'Sync notes'}
             icon='ph--arrows-clockwise--regular'
             iconOnly
-            disabled={syncing || !account.apiKey}
+            disabled={syncing || !account.accessToken?.target?.token}
             onClick={handleSync}
           />
           {trelloCards.length > 0 && (
@@ -532,7 +536,7 @@ export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) 
                   gap={8}
                   items={noteItems}
                   draggable={false}
-                  getId={(item) => item.record.granolaId}
+                  getId={(item) => Granola.getGranolaId(item.record) ?? item.record.id}
                   getScrollElement={() => noteViewport}
                   estimateSize={() => 120}
                 />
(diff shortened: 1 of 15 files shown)


chad:
@richburdon would appreciate your review. Following the same patterns from the Trello PR feedback.

dmaretskyi:
this one I'm most excited about, I can have an agent run pre-programmed prompts after every meeting

packages/plugins/plugin-granola/src/types/Granola.ts
```diff
@@ -0,0 +1,148 @@
...
+import { FormInputAnnotation, LabelAnnotation } from '@dxos/echo/internal';
+import { Markdown } from '@dxos/plugin-markdown/types';
+import { AccessToken } from '@dxos/types';
+
+// @import-as-namespace
+
+/** Attendee of a Granola meeting. */
+export const Attendee = Schema.Struct({
+  name: Schema.optional(Schema.String),
+  email: Schema.optional(Schema.String),
```
dmaretskyi, line 17: add optional ref to standard Person type

packages/plugins/plugin-granola/src/types/Granola.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export type Attendee = Schema.Schema.Type<typeof Attendee>;
+
+/** Calendar event metadata. */
+export const CalendarEvent = Schema.Struct({
+  title: Schema.optional(Schema.String),
+  eventId: Schema.optional(Schema.String),
+  organizerEmail: Schema.optional(Schema.String),
+  startTime: Schema.optional(Schema.String),
+  endTime: Schema.optional(Schema.String),
```
dmaretskyi, line 28: I think we have a standard event type too

packages/plugins/plugin-granola/src/types/Granola.ts
```diff
@@ -0,0 +1,148 @@
...
+  });
+};
+
+// ── Foreign key helpers ─────────────────────────────────────────────────────
+
+const GRANOLA_SOURCE = 'granola.ai';
+
+/** Extract the Granola ID from an object's foreignKeys. */
+export const getGranolaId = (obj: Obj.Any): string | undefined => {
+  const meta = Obj.getMeta(obj);
```
dmaretskyi:
```suggestion
  return Obj.getKeys(obj, GRANOLA_SOURCE)[0]?.id;
```

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,555 @@
...
+  noteTitle: string;
+  meetingTitle: string;
+  cardName: string;
+  cardListName: string;
+  confidence: string;
+  reasoning: string;
+};
+
+/** Use Claude to find matches between Granola notes and Trello cards. */
+const aiMatch = async (
```
dmaretskyi, line 93: change to operation that requests AiService, this way it wouldn't require custom token and will go through the same API as all other code that requires a language model

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,551 @@
...
+    labels: (card.labels ?? []).map((label) => label.name).join(', '),
+  }));
+
+  const response = await fetch('/api/anthropic/v1/messages', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-api-key': apiKey,
+      'anthropic-version': '2023-06-01',
+    },
```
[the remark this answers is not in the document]
chad: Partially addressed — the `ANTHROPIC_API_KEY` read in `aiMatch` is a known gap. Switching to the `AiService` operation is called out in the cross-cutting PR-feedback plan and will land in a follow-up (same refactor as dmaretskyi's comment at line 87). This PR didn't include the AiService work to keep the diff reviewable.

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,551 @@
...
+
+  if (!response.ok) {
+    const errorText = await response.text();
+    throw new Error(`Anthropic API ${response.status}: ${errorText}`);
+  }
+
+  const result = await response.json();
+  const text = result.content?.[0]?.text ?? '[]';
+  const jsonText = text.replace(/^```json?\n?/, '').replace(/\n?```$/, '').trim();
+  return JSON.parse(jsonText) as MatchResult[];
```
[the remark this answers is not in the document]
chad: Fixed in 872adc955e — AI response now decoded through an Effect schema (`MatchResultArraySchema`) with a separate try/catch on the JSON parse step, so malformed or partial model output surfaces a clear error instead of casting past the type system.

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,551 @@
...
+MatchTile.displayName = 'MatchTile';
+
+//
+// Main article component.
+//
+
+export const GranolaArticle = ({ role, subject: account }: GranolaArticleProps) => {
+  const db = Obj.getDatabase(account);
+  const syncRecords: Granola.GranolaSyncRecord[] = useQuery(db, Filter.type(Granola.GranolaSyncRecord));
+  const trelloCards: Trello.TrelloCard[] = useQuery(db, Filter.type(Trello.TrelloCard));
```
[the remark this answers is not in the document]
chad: Fixed in 872adc955e — added an optional `account: Ref<GranolaAccount>` field, populated on every sync, and the article filters `syncRecords` by it so one account surface never sees or mutates another account's notes. Legacy records without the ref are back-filled during update.

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,551 @@
...
+  }, [db, matches, invokePromise]);
+
+  const handleSync = useCallback(async () => {
+    if (!db || !account.apiKey) {
+      return;
+    }
+
+    setSyncing(true);
+    try {
+      const headers = { Authorization: `Bearer ${account.apiKey}` };
```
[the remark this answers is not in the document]
chad: Outdated — the current code already resolves the token via `account.accessToken?.target?.token`, not the old direct `apiKey` field. In React context the Ref target auto-resolves; when this sync moves to an Effect operation handler (follow-up) it will use `Database.load` per the inbox pattern.

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
```diff
@@ -0,0 +1,551 @@
...
+        }
+        const response = await fetch(`${GRANOLA_API_BASE}/notes?${params}`, { headers });
+        if (!response.ok) {
+          throw new Error(`Granola API: ${response.status}`);
+        }
+        const data = await response.json();
+        remoteNotes.push(...data.notes);
+        hasMore = data.hasMore;
+        cursor = data.cursor ?? undefined;
+      }
```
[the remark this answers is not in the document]
chad: Fixed in 872adc955e — escalated from `log.warn` + break to a thrown `Error`. A stuck cursor now halts sync immediately instead of silently returning partial data.

packages/plugins/plugin-granola/src/types/Granola.ts
```diff
@@ -0,0 +1,147 @@
...
+  accessToken: Schema.optional(
+    Ref.Ref(AccessToken.AccessToken).annotations({
+      title: 'Granola credentials',
+      description: 'API key for syncing meeting notes.',
+    }),
+  ),
+  /** Last sync timestamp. */
+  lastSyncedAt: Schema.optional(Schema.String.pipe(FormInputAnnotation.set(false))),
+  /** Polling interval in milliseconds. */
+  pollIntervalMs: Schema.optional(Schema.Number),
```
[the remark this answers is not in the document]
chad: Fixed in e034babf33 — constrained `pollIntervalMs` to `Schema.Number.pipe(Schema.greaterThan(0))`.

packages/plugins/plugin-granola/src/types/Granola.ts
```diff
@@ -0,0 +1,147 @@
...
+/** Creates a GranolaSyncRecord with foreignKey for the Granola note ID. */
+export const makeSyncRecord = (props: {
+  granolaId: string;
+  document: Ref.Ref<Markdown.Document>;
+  attendees?: Attendee[];
+  calendarEvent?: CalendarEvent;
+  ownerName?: string;
+  ownerEmail?: string;
+  createdAt?: string;
+}): GranolaSyncRecord => {
```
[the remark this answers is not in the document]
chad: Already addressed — `makeSyncRecord` accepts `updatedAt` in its props type (see `Granola.ts:129`). `872adc955e` also updates the container to use `Granola.makeSyncRecord` instead of raw `Obj.make`, which was silently dropping `granolaId` (it's a foreign key, not a schema field).

packages/plugins/plugin-granola/src/types/Granola.ts
chad: Tracked as a follow-up — mapping Granola attendees to `Person` (with `Ref<Person>`) is a deeper refactor because it needs resolver logic (email → existing Person vs. new Person) that mirrors plugin-inbox's `InboxResolver.Live`. Keeping `Attendee` as a simple struct for this PR so the review stays scoped; will open a dedicated follow-up PR for the standard-type migration.

packages/plugins/plugin-granola/src/types/Granola.ts
chad: Same follow-up — `CalendarEvent` currently mirrors the Granola payload shape; migrating to the standard `Event` type from `@dxos/types` needs the Attendee→Person and Actor→Person plumbing first. Tracked in the cross-plugin AccessToken/standard-type plan.

packages/plugins/plugin-granola/src/types/Granola.ts
chad: Fixed in 872adc955e — `getGranolaId` now uses `Obj.getKeys(obj, GRANOLA_SOURCE)[0]?.id`.

packages/plugins/plugin-granola/src/containers/GranolaArticle/GranolaArticle.tsx
chad: Noted — migrating `aiMatch` from the direct Anthropic call to an `AiService` operation is planned as a follow-up PR. The refactor touches multiple plugins (Granola, Slack both call Anthropic directly from the browser, which is also a credential-leak issue), so I'm sequencing it as its own change rather than bundling it here.

chad:
## Review response (2026-04-24)

Pass through all 11 comments. Details inline; summary:

**Fixed:**
- **872adc955e** — multi-account scoping (new `account: Ref<GranolaAccount>` on `GranolaSyncRecord`, populated on sync, filtered in the article so one account surface never touches another's notes), `Obj.getKeys` one-liner, AI-response schema decode, fail-fast on stuck pagination, and fixed a latent bug where the container was calling `Obj.make` directly which silently dropped the `granolaId` foreign key.
- **e034babf33** — `pollIntervalMs` constrained to `> 0`.

**Deferred to dedicated follow-up PRs:**
- `Attendee → Ref<Person>` and `CalendarEvent → Event` standard-type migrations (needs a resolver layer like plugin-inbox's `InboxResolver.Live`).
- Switching `aiMatch` from direct Anthropic fetch to the `AiService` operation. This is shared with Slack which has the same issue — sequencing as one cross-plugin refactor so the pattern lands consistently.

Both follow-ups are tracked in `plans/pr-feedback-cleanup.md`.

PR #11058, opened 2026-04-21
refactor(client): move observability/singleClient fields to Config and simplify service setup

dmaretskyi:
## Summary

Part of [DX-930](https://linear.app/dxos/issue/DX-930). Refactors client services setup:

- Moves `observabilityGroup`, `signalTelemetryEnabled`, and `singleClientMode` from ad-hoc options passed through `createClientServices` to `Runtime.Client` in the config proto. Factory, local services, and worker services now read them from a single source.
- Removes these fields from `StartRequest` in `iframe.proto` and drops the RPC threading through `SharedWorkerConnection.open` → `WorkerSession` → `WorkerRuntime`.
- The shared worker now receives the client's `config.values` via the initial port postMessage and overlays it on top of its `Storage/Envs/Local/Defaults`, aligning the shared-worker path with the already-existing dedicated-worker config handoff.
- Simplifies `fromHost` (drops positional observability args) and the `CreateClientServicesOptions` shape.

## What's in scope (from DX-930)

- [x] Move `singleClientMode`, `observabilityGroup`, and `signalTelemetryEnabled` to Config
- [x] Client services always read config from client (shared-worker now overlays client-supplied config)
- [x] Simplify service setup (removes parameter threading across factory → services → RPC → session → runtime)

## Out of scope (follow-up)

- [x] \`create*Worker\` callbacks should not influence how storage is configured — the OPFS vs file vs memory SQLite choice still flows from `createOpfsWorker` / `sqlitePath` presence. That's a structural change worth its own PR.

## Test plan

- [x] `moon run client:compile` and `client-services:compile` clean
- [x] `moon run composer-app:compile` clean
- [x] `moon run client:test -- shared-worker.test.ts` — 3/3 pass (validates shared-worker config handoff)
- [x] `moon run client:test -- dedicated-worker-client-services.test.ts` — 5/5 pass
- [x] `moon run client:test -- client.test.ts spaces-invitations.test.ts spaces.test.ts` — all pass in isolation
- [x] `moon run client-services:test -- space-invitation-protocol.test.ts service-context.test.ts` — all pass in isolation
- [x] `moon run client:lint client-services:lint composer-app:lint -- --fix` — clean
- [ ] Observability group/telemetry metadata still reaches signal server in a live browser (worth a smoke test before merge)

Note: full-concurrent `client:test` / `client-services:test` runs produced a handful of timeout-related flakes in invitation/service-context tests with 1s test timeouts, all unrelated to touched APIs and all passing when rerun in isolation.

part of DX-930

## Summary by CodeRabbit

* **New Features**
  * Added a services mode setting (HOST / SHARED_WORKER / DEDICATED_WORKER) to control how client services are deployed; runtime now auto-selects a safe mode for iOS.

* **Refactor**
  * Observability and telemetry controls moved into centralized runtime config (no longer sent in startup requests).
  * Workers now receive a config overlay on connect and can update signal metadata at runtime.
  * Coordinator selection respects the runtime single-client mode.

### DIFF
--- a/packages/apps/composer-app/src/main.tsx
+++ b/packages/apps/composer-app/src/main.tsx
@@ -175,7 +175,8 @@ const main = async () => {
 
   // Decide the deployment mode for client services. The factory is a dumb switch on
   // `runtime.client.services_mode` — the app is responsible for picking the right mode from its
-  // env / platform constraints and then supplying the matching worker factory below.
+  // env / platform constraints. Worker factories are passed unconditionally; the factory only
+  // invokes the one required by the configured mode.
   const useLocalServices = isTrue(config.values.runtime?.app?.env?.DX_HOST);
   const useSharedWorker = isTrue(config.values.runtime?.app?.env?.DX_SHARED_WORKER);
   // iOS has a SharedWorker crash bug (Apple FB11723920); if a caller asks for SharedWorker there,
@@ -190,6 +191,13 @@ const main = async () => {
         : defs.Runtime.Client.ServicesMode.HOST
       : defs.Runtime.Client.ServicesMode.DEDICATED_WORKER;
 
+  // Host mode uses OPFS SQLite in a dedicated worker; worker modes run their own in-memory SQLite
+  // (OPFS does not yet work from inside a SharedWorker per the TODO in `worker-runtime.ts`).
+  const sqliteMode =
+    servicesMode === defs.Runtime.Client.ServicesMode.HOST
+      ? defs.Runtime.Client.Storage.SqliteMode.OPFS
+      : defs.Runtime.Client.Storage.SqliteMode.MEMORY;
+
   config = new Config(
     {
       runtime: {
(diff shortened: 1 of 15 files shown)


dmaretskyi:
Thanks @coderabbitai — fair catch.

Pushed [2e2893f](https://github.com/dxos/dxos/pull/11058/commits/2e2893f2ed) which preserves the pre-DX-930 per-session (last-writer-wins) semantics for `observabilityGroup` and `signalTelemetryEnabled`:

- New `WorkerRuntime.updateSignalMetadata(config)` refreshes the telemetry fields on the live runtime.
- `onconnect` calls it on every client init message, so a second tab with a different `observabilityGroup` overwrites the signal tags just like the pre-refactor `WorkerService.start` RPC did.
- The first client still seeds the worker's core config (storage / signaling / edge features) — those can't change after the runtime is up.
- Documented the last-writer-wins scope on `Runtime.Client.observability_group` and `signal_telemetry_enabled` in the proto.

dmaretskyi:
Fixed in [f37236d](https://github.com/dxos/dxos/pull/11058/commits/f37236dd4f): `updateSignalMetadata` now `delete`s the stale `group` tag when the incoming config has no `observabilityGroup`, so last-writer-wins holds when a later client clears it.

dmaretskyi:
Fixed both in [f027153](https://github.com/dxos/dxos/pull/11058/commits/f027153ff4):

1. **CI (TS2367)**: the `!servicesMode || servicesMode === UNSPECIFIED_SERVICES_MODE` guard was redundant (UNSPECIFIED is 0, so `!` already covers it) and stricter `build` tsc flagged it as always-false. Dropped the second branch.
2. **CodeRabbit**: added `.catch(log.catch)` to the fire-and-forget `updateSignalMetadata` call in `onconnect.ts`.

Verified `moon run client:build client-services:build` passes locally now.

packages/apps/composer-app/src/main.tsx
```diff
@@ -173,38 +173,63 @@ const main = async () => {
...
+      },
+    },
+    config.values,
+  );
   const services = await createClientServices(config, {
     createWorker:
-      useLocalServices || !useSharedWorker
-        ? undefined
-        : () =>
+      servicesMode === defs.Runtime.Client.ServicesMode.SHARED_WORKER
```
dmaretskyi: you can provide them unconditionally and just not use in createClientServices

dmaretskyi: Done in [a1d129a](https://github.com/dxos/dxos/pull/11058/commits/a1d129ab22) — worker factories are now passed unconditionally; the factory only invokes the one required by the selected `services_mode`.

PR #11084, opened 2026-04-23
refactor(assistant-e2e): rebuild on composer plugin testing framework

dmaretskyi:
## Summary

- Rewrite the assistant-e2e harness to bootstrap real composer plugins via `createComposerTestApp` and harvest their blueprints, schemas, and operation handlers into `AssistantTestLayer`. Tests now pass `plugins: [SomePlugin()]` instead of manually re-importing blueprints and handler sets.
- Default the harness to the `edge-remote` inference preset (no API key required) and keep per-test memoization togglable via `disableLlmMemoization`.
- Node-safe activation: `autoStart: false` + fires only `SetupSchema`, `SetupArtifactDefinition`, and `SetupOperationHandler` — avoids pulling browser-only React surfaces.
- Fixes latent bugs exposed along the way:
  - `plugin-inbox` `#capabilities/node.ts` now exports all lazy capabilities (not just `OperationHandler`) so node consumers see `BlueprintDefinition` etc.
  - `InboxPlugin`'s settings module now has an explicit `id`.
- Regenerated `*.conversations.json` memoized caches via `edge-remote`.

## Test plan

- [x] `moon run assistant-e2e:test` (replays from cache, all 5 tests pass)
- [x] `ALLOW_LLM_GENERATION=1 moon run assistant-e2e:test` (regenerates cleanly)
- [x] `moon run assistant-e2e:lint` and `plugin-inbox:lint`

## Summary by CodeRabbit

* **New Features**
  * Test harness can load real plugins and accepts extra blueprints/types/handlers for more realistic, customizable testing.

* **Improvements**
  * Inbox settings registration made more explicit for clearer behavior.
  * Plugin capabilities now expose additional lazy-loaded entry points for improved extensibility and startup performance.

* **Tests**
  * E2E tests updated to explicitly inject a fixture plugin and provide explicit blueprints for deterministic behavior.

* **Documentation**
  * Memoized-LLM regeneration docs simplified and clarified; added package to regeneration list.

* **Chores**
  * TypeScript project references and test package dependencies updated.


packages/core/assistant-e2e/src/harness.ts
```diff
@@ -22,20 +24,39 @@ import {
...
+import { createComposerTestApp } from '@dxos/plugin-testing/harness';
 import { trim } from '@dxos/util';
 
 export const DEFAULT_TEST_TIMEOUT = 120_000;
 
+/**
+ * Built-in blueprints shipped by `@dxos/assistant-toolkit` that are not
+ * contributed via composer plugins. Always included in the test layer.
+ */
+const DEFAULT_TOOLKIT_BLUEPRINTS = (): Blueprint.Blueprint[] => [
```
dmaretskyi: no, EVERYTHGING must come from plugins, no default blueprints, operation types, only plugins.

packages/core/assistant-e2e/src/harness.ts
```diff
@@ -74,64 +151,51 @@ export const agentTest: {
...
-        },
-        { conversation: Obj.getDXN(conversationFeed).toString() },
-      ).pipe(Effect.exit);
-
-      if (options.expect === 'failure') {
-        if (!Exit.isFailure(exit)) {
-          throw new Error('Expected the agent to fail, but it succeeded');
+      const contributions = yield* harvestFromPlugins(options.plugins ?? []);
+
+      const TestLayer = AssistantTestLayer({
```
dmaretskyi: do not use assystant test layer. instead fire the PromptAgentWorker (what's the name?) operation inside the test app harness

packages/core/assistant-e2e/src/harness.ts
dmaretskyi: Addressed in 8180c8cfd0 and 87628aa34c. Harness is now a thin wrapper that fires activation events and invokes `AgentPrompt` via `Operation.invoke` through the managed runtime. All contributions (blueprints, operation handlers, schemas, toolkits) now flow from plugins — tests pass `AssistantTestFixturePlugin()` (plus any extra) into `agentTest({ plugins: [...] })`. A `TestServicesPlugin` gathers plugin contributions at `SetupLayer` and contributes the runtime service layer.

packages/core/assistant-e2e/src/harness.ts
dmaretskyi: Addressed in 8180c8cfd0 and 87628aa34c. The harness now runs `Operation.invoke(AgentPrompt, ...)` inside the test app `ManagedRuntime` (via `harness.get(Capabilities.ManagedRuntime)`). `AssistantTestLayer` is only used to assemble the underlying Effect service layer, which is contributed to the runtime by `TestServicesPlugin` at `SetupLayer`. Using the runtime routes through `ProcessOperationInvoker`, which correctly resolves per-invocation services like `AiContextService` via `ServiceResolver`.

PR #11088, opened 2026-04-23
feat(echo): enforce schema registration for feed.append + query typenames

dmaretskyi:
## Summary

Verifies and enforces the schema registration semantics requested in the task.

1. `db.add` continues to fail when the object's schema is not in the runtime registry or the persistent schema registry (pre-existing behavior).
2. `feed.append` (queue.append) now fails with `Schema not registered` unless the item's schema is in the runtime registry or the persistent schema registry.
3. Queries throw when any typename referenced by their filters cannot be resolved in either registry.
4. Query results whose objects have an unresolvable schema are filtered out.
5. A new `skipSchemaValidation` option on `QueryOptions` disables both (3) and (4). Default is `false`.
6. `Entity.getValidationErrors(obj)` returns a readonly array of `ParseResult.ArrayFormatterIssue` describing validation errors against the object's schema.

## Implementation notes

- Query validation is implemented as a `SchemaValidatingQueryResult` decorator applied at the `EchoDatabaseImpl._query` and `QueueImpl._query` entry points.
- `DatabaseSchemaRegistry.getSchemaByDXN` resolves both TYPE and ECHO DXNs, falling back to a live query for persistent schemas not yet cached.
- New AST helpers: `QueryAST.collectTypenames`, `QueryAST.getEffectiveOptions`.

## Test coverage

`packages/core/echo/echo-db/src/testing/schema-registration.test.ts` covers every matrix variant (runtime vs. persistent schema × feed vs. database automerge object), the `skipSchemaValidation` escape hatch, and `Entity.getValidationErrors`.

## Test plan
- [x] `moon run echo-db:test` (571 passing, 19 skipped, 1 todo)
- [x] `moon run echo:test` (284 passing, 9 skipped)
- [x] `moon run echo-db:build echo:build echo-protocol:build` (clean)
- [x] `moon run echo-db:lint echo:lint echo-protocol:lint` (clean)
- [x] `moon run client:build composer-app:build react-client:build` (clean; downstream type-checks)

https://claude.ai/code/session_01QEacfXpBbPd6NeKtVp8ELr

---
_Generated by [Claude Code](https://claude.ai/code/session_01QEacfXpBbPd6NeKtVp8ELr)_

## Summary by CodeRabbit

* **New Features**
  * Per-query schema validation with a bypass option (skipSchemaValidation).
  * New entity validation API returning schema decode issues.

* **Improvements**
  * Query and queue results now filter out objects whose types cannot be resolved (unless skipped).
  * Clearer errors for unknown typenames; export/restart flows can opt out to include unregistered objects.

* **Tests**
  * Expanded integration/unit tests for registration, validation, queries, queues, and edge cases.


packages/core/echo/echo-db/src/proxy-db/database-schema-registry.ts
```diff
@@ -102,6 +102,35 @@ export class DatabaseSchemaRegistry extends Resource implements SchemaRegistry.S
     return schemaId != null && this.getSchemaById(schemaId) != null;
   }
 
+  /**
+   * Resolves a schema by its type or echo DXN.
+   * Checks for a matching PersistentSchema in the space.
+   */
+  public getSchemaByDXN(dxn: DXN): Type.RuntimeType | undefined {
```
dmaretskyi, line 109: can't query method already filter schema by DXN?

dmaretskyi: Good catch — simplified in 85877c5. Delegates to `this.query({ typename, version })` for TYPE DXNs and `this.query({ backingObjectId })` for ECHO DXNs.

---

packages/core/echo/echo-db/src/query/schema-validating-query-result.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Decorator over a {@link QueryResult} that enforces schema validation semantics:
+ * - Throws when the query references typenames that cannot be resolved in either the runtime or
+ *   persistent schema registry.
+ * - Filters out results whose schema cannot be resolved.
+ *
+ * Both behaviors are disabled when the query's options clause sets `skipSchemaValidation: true`.
+ */
+export class SchemaValidatingQueryResult<
```
dmaretskyi: instead of a mixing, put this logic into existing query source classes

dmaretskyi: Done in 8f787b5. Moved validation into `GraphQueryContext` and `QueueQueryContext` via an optional `SchemaResolvers` field. Deleted `SchemaValidatingQueryResult`. Added `HypergraphImpl._createQueryContext` so `EchoDatabaseImpl` can inject its own database-scoped resolvers.

One subtlety: the assertion runs on `run()` (user-initiated) but only *filters* (no throw) in `getResults()`, because `getResults` is called reactively from `QueryResultImpl`'s `changed` listener — throwing there surfaces as an unhandled rejection rather than reaching the caller.

---

PR #11165, opened 2026-04-30
feat(composer-devtools-protocol): schema-driven devtools panels over CustomEvent + Comlink

dmaretskyi:
## Summary

- New `@dxos/composer-devtools-protocol` package — declarative spec (`stack` / `action` / `debug` / `input` / `select`), builders, and an app-side host that exposes a `ComposerDevtools.showPanel({...})` API. Panels are registered on the page; only ids and serialized trees cross the bridge.
- Transport is **Comlink** over a **CustomEvent** bridge between the page and the content script, then a long-lived `chrome.runtime.Port` hop content ↔ background ↔ panel. The page-side host announces with a `hello` / `probe` handshake so the page-loaded-first vs. extension-loaded-first race is harmless.
- Wires `composer-crx`: bridge installed in the content script, router in the background, new `devtools.html` / `src/devtools.ts` entrypoint that registers the panel via `chrome.devtools.panels.create`, and `panel.tsx` rewritten to render the spec tree and dispatch UI events back to the page.

## Architecture

```
page (host)  <- CustomEvent ->  content script  <- Port ->  background  <- Port ->  devtools panel
   ^                                                                                       |
   └─────────────────────────  Comlink RPC (DevtoolsHostApi) ───────────────────────────────┘
```

The `subscribe` / `dispatch` API uses subscription-id tokens (rather than returning teardown closures over Comlink) so cleanup doesn't depend on round-tripping a proxy. Builders register handlers in a per-render context keyed by id; the host clears and rebuilds the dispatch table on every render.

## Public API

```ts
import { ComposerDevtools } from '@dxos/composer-devtools-protocol';

ComposerDevtools.showPanel({
  id: 'my-panel',
  name: 'My Panel',
  onMount(handle) { /* handle.update() to re-render */ },
  onUnmount() { /* ... */ },
  onRender() {
    return ComposerDevtools.stack({ direction: 'vertical' }, [
      ComposerDevtools.action({ name: 'Refresh' }, () => { /* ... */ }),
      ComposerDevtools.debug({ value: state, label: 'state' }),
      ComposerDevtools.input({ name: 'filter', value: filter }, (v) => { filter = v; }),
      ComposerDevtools.select({ name: 'mode', value: mode, options }, (v) => { mode = v; }),
    ]);
  },
});
```

## Test plan

- [ ] `moon run composer-devtools-protocol:build` — passes (verified locally).
- [ ] `moon run composer-crx:bundle` — passes (verified locally).
- [ ] `moon run composer-devtools-protocol:lint composer-crx:lint` — clean.
- [ ] Load the unpacked extension, open Chrome DevTools on a Composer page → "Composer" panel appears, lists registered panels, renders the tree, action / input / select round-trip back to the page.
- [ ] Verify the page-loaded-first race: reload the inspected tab with devtools open and confirm the panel re-binds via the `probe` handshake.

---
_Generated by [Claude Code](https://claude.ai/code/session_015H9mJ87eJ5i4sFwrxargTg)_

## Summary by CodeRabbit

* **New Features**
  * Chrome DevTools panel added for the Composer extension.
  * Register and manage Composer panels directly within DevTools.
  * Interactive UI to inspect and interact with Composer specs (actions, inputs, selects).
  * DevTools panel auto-detects and shows active panels from inspected pages.


packages/apps/composer-crx/src/devtools/bridge.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Content-script side of the bridge. Relays messages between the page
+ * (CustomEvents on `window`) and the panel (a long-lived
+ * `chrome.runtime.Port` to the background worker).
+ *
+ * Call once from a content script that runs at `document_start` on pages
+ * the devtools panel should be able to inspect. Idempotent per channel.
+ */
+export const installDevtoolsBridge = (channel = DEFAULT_CHANNEL): void => {
```
dmaretskyi, line 20: keep protocol as typedefs and client API only, put everything else in the CRX package

PR #11198, opened 2026-05-03
fix(plugin-wnfs): hide image decorations in source view mode

richburdon:
## Summary

- The plugin-wnfs image extension replaced `wnfs://` URLs with image widgets even when the markdown editor was in `source` (plain-text) view mode, so users saw images instead of the raw URL.
- Threaded `viewMode` through `MarkdownExtensionProvider` (extending the props the callback receives) so other plugins can opt out per view mode.
- Updated `plugin-wnfs`'s contribution to skip the image extension when `viewMode === 'source'`, mirroring the existing gate on plugin-markdown's core decorations in `useExtensions.tsx`.

## Test plan

- [ ] Open a markdown document containing a `wnfs://...` image link.
- [ ] Verify the image renders inline in `preview` and `readonly` modes.
- [ ] Toggle to `source` mode and verify the raw `` markdown is shown (no image).
- [ ] Confirm other plugins contributing markdown extensions (mermaid, sheet, thread, files, etc.) continue to work — `viewMode` was only added as an optional prop on the provider callback.

## Summary by CodeRabbit

* **New Features**
  * Markdown editor is view-mode aware; extension providers now receive the current mode.
  * Extensions regenerate when editor mode or document changes, so the UI updates when switching modes.
  * Certain extensions (e.g., media/inline) are automatically disabled in the raw/source view to prevent inappropriate behavior.

### DIFF
--- a/packages/plugins/plugin-markdown/src/containers/MarkdownContainer/MarkdownContainer.tsx
+++ b/packages/plugins/plugin-markdown/src/containers/MarkdownContainer/MarkdownContainer.tsx
@@ -54,11 +54,11 @@ export const MarkdownContainer = forwardRef<HTMLDivElement, MarkdownContainerPro
         return [];
       }
 
+      const document = Obj.instanceOf(Markdown.Document, object) ? object : undefined;
       return [...(otherExtensionProviders ?? []), ...(extensionProviders ?? [])]
         .flat()
         .reduce((acc: Extension[], provider) => {
-          const extension =
-            typeof provider === 'function' ? provider({ document: object as Markdown.Document, viewMode }) : provider;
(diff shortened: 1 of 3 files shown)


richburdon:
Addressed the CodeRabbit nit in 02cecc392b — replaced the inline `{document?: any; viewMode?: EditorViewMode}` parameter type with the canonical `MarkdownExtensionProvider` from `@dxos/plugin-markdown/types`.

richburdon:
Addressed the CodeRabbit comment in e8adca0f26 — only pass `document` to providers when the editor's subject is a `Markdown.Document`; pass `undefined` for `Text.Text` so providers don't receive a mistyped object.

PR #11212, opened 2026-05-04
chore(build): migrate ts-build packages to tsdown/rolldown JS bundling with tsgo DTS

dmaretskyi:
## Summary

Migrates all `ts-build` packages (~120) from `dx-compile` (esbuild) to **tsdown/rolldown** for JavaScript bundling, while keeping `dx-build` (tsgo CLI) for TypeScript declaration generation. Removes the need for rolldown-plugin-dts and its associated TS2883 workarounds.

### Key changes

- **`tsdown.base.config.ts`** at workspace root — standalone config factory with all plugins and `defineConfig()`. Imported by relative path from each package's `tsdown.config.ts`; no compile step needed.
- **Per-package `tsdown.config.ts`**: ~120 files, each imports from the base config and calls `defineConfig()`.
- **`.moon/tasks/tag-ts-build.yml`** — `compile` task now runs `pnpm exec tsdown` with `tsdown.config.ts` input.
- **Package exports** — all `types` fields updated from `dist/types/index.d.ts` → `dist/types/src/index.d.ts` (tsgo emits per-file declarations in `src/` subdirectory).
- **Reverted all TS2883 workarounds** from rolldown-plugin-dts era — no longer needed with tsgo CLI per-file DTS: `Capability.LazyCapability`, `Capability.ModuleReturn`, `Operation.Definition.Any`, `Schema.Schema.AnyNoContext`, `Type.AnyObj`, `Meta<any>`, and explicit function return types.

### Plugins in tsdown.base.config.ts

| Plugin | Purpose |
|---|---|
| `workspaceExternalPlugin` | Marks `@dxos/*` imports external before rolldown follows pnpm symlinks to real paths |
| `nodeStdPlugin` | Remaps `node:*` / bare built-ins → `@dxos/node-std/*` externals in browser builds |
| `rawImportPlugin` | Handles `?raw` suffix imports; uses `.mjs` virtual ID suffix to avoid `@tsdown/css` conflicts |
| `urlImportPlugin` | Handles `?url` suffix imports |
| `rolldownLogMetaPlugin` | Injects log-meta transforms for `@dxos/log` call sites |

### Packages requiring dx-compile (special cases)

Four packages override the `compile` task to use dx-compile (esbuild) instead of tsdown, since they need special bundling:
- `@dxos/sql-sqlite` — native sqlite bindings
- `@dxos/node-std` — node polyfill bundler
- `@dxos/random-access-storage` — native FS bindings  
- `@dxos/teleport` — native network bindings

These override only `compile` (not `build`) so the inherited `build` task still runs tsgo for DTS.

### DTS architecture

```
compile task → tsdown (JS → dist/lib/browser/ + dist/lib/node-esm/)
build task   → dx-build/tsgo (DTS → dist/types/src/**/*.d.ts)
```

tsgo generates per-file `.d.ts` files — no bundling, no TS2883 "cannot reference internal chunk" errors.

https://claude.ai/code/session_01RWGgdV1Sxwrgzm6VkUGWNH


tsdown.base.config.ts
```diff
@@ -0,0 +1,428 @@
...
+const TSC_FORMAT_HOST: ts.FormatDiagnosticsHost = {
+  getCurrentDirectory: () => process.cwd(),
+  getCanonicalFileName: (f) => f,
+  getNewLine: () => '\n',
+};
+
+/**
+ * Runs `tsc --emitDeclarationOnly` using the TypeScript compiler API.
+ */
+const runTscDts = async (tmpTsconfig: string, outDir: string): Promise<void> => {
```
dmaretskyi: tsdown has this built-in; why wasnt that working?

tsdown.base.config.ts
```diff
@@ -0,0 +1,428 @@
...
+      plugins: [nodeStdNeutral, rawPlugin, urlPlugin, wsExternalPlugin, logPlugin],
+    });
+  }
+
+  // Run tsc DTS generation on success of the first config.
+  if (configs.length > 0) {
+    const prev = configs[0].onSuccess;
+    configs[0] = {
+      ...configs[0],
+      onSuccess: async (resolvedConfig, signal) => {
```
dmaretskyi: I for sure wanna use built-in dts gen with TSGO! instead of custom thing

tsdown.base.config.ts
dmaretskyi: Tried both built-in paths:

1. **rolldown-plugin-dts (standard tsc path)** — uses `ts.createLanguageService` internally, which hits a TypeScript compiler bug ("Debug Failure. False expression: Lexical environment is suspended") on several packages in this repo (e.g. `errors`). The custom `ts.createProgram` call below avoids this code path entirely.

2. **tsgo path** — `@typescript/native-preview` v7.0.0-dev.20260421.2 doesn't yet resolve `#`-prefixed package.json `imports` fields (`resolvePackageJsonImports`). Packages like `crypto` (uses `#subtle`), `log`, and many plugins fail with "did not generate dts file" silently.

Once tsgo ships support for package.json `imports` field resolution, we can drop all of this and switch. Added a comment in the source explaining the two failure modes.

---

tsdown.base.config.ts
dmaretskyi: Same as above — tsgo (`@typescript/native-preview` v7.0.0-dev.20260421.2) doesn't yet support `#`-prefixed package.json `imports` field resolution. Packages like `crypto` (uses `#subtle`), `log`, and many plugins with subpath imports silently produce no `.d.ts` output. Once that's fixed upstream we can switch to the tsgo path and delete all of this.

---

PR #11235, opened 2026-05-05
fix(echo): chunk querySnapshotsJSON IN clauses to avoid SQLite variable limit

dmaretskyi:
## Summary

- **Root cause**: `FtsIndex.querySnapshotsJSON` issued `SELECT rowid, snapshot FROM ftsIndex WHERE rowid IN (?, ?, …)` with one bound parameter per recordId. SQLite's `SQLITE_LIMIT_VARIABLE_NUMBER` is 999 in most wasm builds. Trace queues (and other high-volume queues) routinely contain >999 items, so every call to `_loadQueueSnapshotMap` failed with `SqlError: Failed to execute statement`.
- **Impact**: Because the error was swallowed by a bare `log.catch(err)`, `query.dirty` was never reset to `false`, causing queue-backed live queries to re-run on every invalidation forever — the hot-loop saturating the dedicated worker described in the issue.

### Changes

**`packages/core/echo/index-core/src/indexes/fts-index.ts`**
- Chunk `recordIds` into batches of 500 (safe margin under the 999 limit) and issue one `SELECT … WHERE rowid IN (…)` per chunk, then flatten results. RecordIds absent from the FTS index (stale rows) are silently omitted — the existing `_loadFromQueue` null-check already handles missing snapshots gracefully.

**`packages/core/echo/echo-pipeline/src/query/query-executor.ts`**
- Add a `queryId` getter so callers outside the class can read the query ID without reflection.

**`packages/core/echo/echo-pipeline/src/db-host/query-service.ts`**
- Replace `log.catch(err)` in `_executeQueries` with `log.warn(…, { queryId, query, err })` so failures are visible with enough context to diagnose (which query, what its AST looks like, what the SQLite error says).

**`packages/core/echo/index-core/src/indexes/fts-index.test.ts`**
- Three new tests under `describe('querySnapshotsJSON')`:
  1. All snapshots present → returns full result set
  2. Stale/missing recordIds → omits them without error
  3. >999 recordIds (1 100 items) → chunking path, all results returned (regression test for the variable-limit bug)

## Test plan

- [x] `moon run index-core:test` — all 12 FtsIndex tests pass, including the new ones
- [x] `moon run echo-pipeline:compile` — no TypeScript errors in modified files
- [x] `moon run echo-pipeline:test` — 123 tests pass

https://claude.ai/code/session_016XvGPHE7YEZHxMPDuMxW22

---
_Generated by [Claude Code](https://claude.ai/code/session_016XvGPHE7YEZHxMPDuMxW22)_

## Summary by CodeRabbit

* **New Features**
  * Exposed a public query identifier for retrieving a query's ID.

* **Bug Fixes**
  * Improved error logging for query execution failures with structured diagnostics and ensured active queries receive error callbacks.
  * Avoided database bound-variable limits by chunking large snapshot queries and silently omitting missing records.

* **Tests**
  * Added tests covering snapshot query behavior for present, mixed, and large input sets.

### DIFF
--- a/packages/core/echo/echo-pipeline/src/db-host/query-service.ts
+++ b/packages/core/echo/echo-pipeline/src/db-host/query-service.ts
@@ -183,7 +183,11 @@ export class QueryServiceImpl extends Resource implements QueryService {
             query.sendResults(query.executor.getResults());
           }
         } catch (err) {
-          log.catch(err);
+          log.warn('query execution failed', {
+            queryId: query.executor.queryId,
+            query: JSON.stringify(query.executor.query),
+            err,
+          });
         }
       }),
     );
--- a/packages/core/echo/echo-pipeline/src/db-host/query-service.ts
+++ b/packages/core/echo/echo-pipeline/src/db-host/query-service.ts
@@ -183,11 +183,11 @@ export class QueryServiceImpl extends Resource implements QueryService {
             query.sendResults(query.executor.getResults());
           }
         } catch (err) {
-          log.warn('query execution failed', {
+          log.catch(err, {
             queryId: query.executor.queryId,
             query: JSON.stringify(query.executor.query),
-            err,
           });
+          query.onError(err as Error);
         }
       }),
     );
--- a/packages/core/echo/echo-pipeline/src/query/query-executor.ts
+++ b/packages/core/echo/echo-pipeline/src/query/query-executor.ts
@@ -274,6 +274,10 @@ export class QueryExecutor extends Resource {
     this._plan = queryPlanner.createPlan(this._query);
   }
 
+  get queryId(): string {
+    return this._id;
+  }
+
   get query(): QueryAST.Query {
     return this._query;
   }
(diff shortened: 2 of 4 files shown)


packages/core/echo/echo-pipeline/src/db-host/query-service.ts
```diff
@@ -183,7 +183,11 @@ export class QueryServiceImpl extends Resource implements QueryService {
             query.sendResults(query.executor.getResults());
           }
         } catch (err) {
-          log.catch(err);
+          log.warn('query execution failed', {
```
dmaretskyi: both of them are visible. and both dont throw error - I'd prefer log.catch, can you hook it up so a failure here actually results in a visible client-side error? (but keep the log)

dmaretskyi: Done — switched to `log.catch(err, { queryId, query })` and added `query.onError(err as Error)` so failures now propagate through the stream's `close(err)` path to the client-side subscriber.

---

PR #11259, opened 2026-05-07
fix(functions-runtime): keep TriggerDispatcher state atom alive

dmaretskyi:
## Summary

The trigger status indicator (`SpaceStatus` / `TriggerStatus.tsx`) showed a stale `enabled: false` even when the underlying `TriggerDispatcher` was running. Root cause: the `_state` atom was a non-`keepAlive`, lazy `@effect-atom/atom` writable.

`compute-runtime.ts` subscribes to `space.properties` with `{ immediate: true }` and synchronously dispatches `triggerDispatcher.start()`/`stop()` via the `transition()` helper. On first load this fires **before** any React subscriber has mounted `useAtomValue(dispatcher.state)`. Without `keepAlive`, the registry disposes the atom node as soon as the write completes, so the resulting `enabled: true` value is dropped. The next subscriber then reads the atom fresh and gets the default `{ enabled: false, ... }` again.

`Atom.keepAlive` keeps the node retained across the no-subscriber gap, matching the pattern already used in `plugin-space/src/capabilities/state.ts`.

## Test plan

- [x] `moon run functions-runtime:test -- packages/core/functions-runtime/src/triggers/trigger-dispatcher.test.ts` passes (18 tests, 1 skipped).
- [ ] Manual: open Composer, observe trigger status indicator reflects current `enabled` immediately on first paint and after `computeEnvironment` toggles.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where trigger dispatcher state could be lost during early application startup, particularly when triggers started and stopped before the user interface fully initialized, improving overall reliability during application boot.


dmaretskyi:
### Merge activity

* **May 7, 6:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/dxos/dxos/11259) couldn't merge this PR because it failed for an unknown reason (This repository has GitHub's merge queue enabled. Please configure the GitHub merge queue integration in Graphite settings.).

PR #11308, opened 2026-05-11
feat(echo): first-class registry key/version support

dmaretskyi:
## Summary

Adds optional `key` (FQN, e.g. `org.example.type.foo`) and `version` (semver) to ECHO object meta so that registry-backed entities (operations, blueprints, etc.) can be identified and queried without leaking those fields into the data payload.

- **ECHO meta** gains optional `key: string` and `version: string` fields alongside `keys` / `tags` (`ObjectMeta` in `echo-protocol` and the Effect `ObjectMetaSchema` in `@dxos/echo`).
- **New filter clause** `Filter.key(key, { version?: '^1.2.3' })`:
  - New `'key'` node in `QueryAST.Filter`.
  - In-memory matching uses `semver.satisfies` (with `includePrerelease: true`) for both `MatchedObject` and JSON objects.
  - Query planner handles AND-combinations of simple filters so things like `Filter.and(Filter.type(Foo), Filter.key('org.example.foo', { version: '^1.0.0' }))` plan cleanly.
- **New `@dxos/echo-registry` package** (sits between `@dxos/echo` and `@dxos/echo-db`):
  - Composable in-memory registry of keyed ECHO objects.
  - CRUD: `add`, `addMany`, `remove`, `clear`, `get`, `list`.
  - `query(query: Query.Query<T>): QueryResult<T>` executes a subset of the ECHO Query API locally (`select` / `filter` / `limit`, with all `filterMatchObjectJSON` filter nodes including `Filter.key`).
  - Optional `upstream` Registry — local results take precedence; upstream fills in missing entries by id.
  - Effect Layer surface: `Registry.Service` tag, `Registry.layer({ ... })`, and `Registry.layerWithUpstream({ ... })` that picks up the upstream from the Effect environment.
- **Migrations for common entities that exercise keys:**
  - `Blueprint` schema bumped to `v0.2.0`, dropping `key` from data (now in meta). Adds `LegacyBlueprint` (`v0.1.0`) and `Blueprint.migration`.
  - `PersistentOperation` schema bumped to `v0.2.0`, dropping `key` / `version` from data (now in meta). Adds `LegacyPersistentOperation` and `Operation.migration`.
  - `Blueprint.getKey` / `Operation.getKey` / `getVersion` convenience accessors over `Obj.getMeta`.
  - All call sites (assistant, doctor, automation, script, sheet, inbox, devtools, CLI, etc.) updated to read `key` / `version` via `Obj.getMeta(x)` and to use `Filter.and(Filter.type(...), Filter.key(...))` where appropriate.

## Verification

- `Filter.key` tests in `packages/core/echo/echo-db/src/query/query.test.ts`:
  - exact-key match
  - any-version match (with and without meta `version` set)
  - `^1.2.3` caret range
  - `~1.2.3` tilde range
  - version range excludes objects with no `version`
- Migration test in `packages/core/echo/echo-db/src/proxy-db/object-migration.test.ts` exercises a `data.key/version` → `meta.key/version` migration end-to-end and queries the migrated objects via `Filter.key(..., { version: '^1.0.0' })`.
- 7 tests in `packages/core/echo/echo-registry/src/Registry.test.ts` covering CRUD, query by key (with and without version range), upstream delegation, local override of upstream, `limit`, and both `Registry.layer` / `Registry.layerWithUpstream`.

## Test plan

- [x] `moon run echo-db:test -- src/query/query.test.ts`
- [x] `moon run echo-db:test -- src/proxy-db/object-migration.test.ts`
- [x] `moon run echo-registry:test`
- [x] `moon run compute:test`
- [x] `moon exec --on-failure continue --quiet :build` — clean across the workspace.
- [ ] Smoke-test Blueprint / Operation flows in Composer.

## Notes

- The object-migration framework currently aliases `before` and `object` in the `onMigration` callback (both point to the post-replace proxy), so the new Blueprint / Operation migrations snapshot the legacy `key` / `version` during `transform` and read them back in `onMigration` keyed by object id. Worth fixing in the migration framework as a follow-up.
- `metaNotEmpty` in `echo-handler.ts` now also considers `key` / `version`, so meta-only edits round-trip through the proxy.

https://claude.ai/code/session_01QAZsSbjjaoZCfQjDRqVVcC

---
_Generated by [Claude Code](https://claude.ai/code/session_01QAZsSbjjaoZCfQjDRqVVcC)_

## Summary by CodeRabbit

* **New Features**
  * Query objects by fully-qualified registry key with semver version range support.
  * Added an in-memory registry with optional upstream delegation and Effect integration.

* **Improvements**
  * Canonicalized identity: keys/versions moved into object metadata and adopted across CLI, UI, runtimes, and pipelines.
  * Query planner and in-memory matching now understand meta key/version semantics.

* **Tests**
  * New and updated tests covering registry behavior, migrations, and meta key/version query semantics.


packages/core/echo/echo-protocol/src/query/ast.ts
```diff
@@ -95,6 +95,27 @@ const FilterTag_ = Schema.Struct({
 export interface FilterTag extends Schema.Schema.Type<typeof FilterTag_> {}
 export const FilterTag: Schema.Schema<FilterTag> = FilterTag_;
 
+/**
+ * Filter objects by registry key (and optional semver range) stored in object meta.
+ */
+const FilterKey_ = Schema.Struct({
```
dmaretskyi: Roll those into FilterObject

packages/core/compute/compute/src/Blueprint.ts
```diff
@@ -18,21 +18,50 @@ import * as Operation from './Operation';
...
 /**
- * Blueprint schema defines the structure for AI assistant blueprints.
- * Blueprints contain instructions, tools, and artifacts that guide the AI's behavior.
- * Blueprints may use tools to create and read artifacts, which are managed by the assistant.
+ * Legacy Blueprint schema (v0.1.0) — `key` is stored as a data property.
+ * @deprecated Use {@link Blueprint} (v0.2.0) instead; the `key` and `version` now live in the object meta.
  */
-// TODO(burdon): Rename Skill?
-export const Blueprint = Schema.Struct({
+export const LegacyBlueprint = Schema.Struct({
```
dmaretskyi: instead name Blueprint_v0_1_0, placeat the end of the file, place migration in this module too

dmaretskyi: do the same for all other updated types

packages/core/compute/compute/src/Operation.ts
```diff
@@ -511,3 +520,73 @@ export const withInvocationOptions = (options: InvokeOptions): Layer.Layer<Servi
...
+    version: '0.1.0',
+  }),
+);
+export interface PersistentOperation_v0_1_0 extends Schema$.Schema.Type<typeof PersistentOperation_v0_1_0> {}
+
+/**
+ * Migration from {@link PersistentOperation_v0_1_0} (v0.1.0) to {@link PersistentOperation} (v0.2.0).
+ * Moves `key` and `version` from the data section into the object meta.
+ */
+// The migration framework currently aliases `before` and `object` in `onMigration`, so we snapshot
```
dmaretskyi: do I understand correctly that you are unable to set meta on the migration result? if so, pls fix upstream

packages/core/compute/compute/src/Operation.ts
```diff
@@ -511,3 +520,73 @@ export const withInvocationOptions = (options: InvokeOptions): Layer.Layer<Servi
...
+// the legacy `key` and `version` during `transform` and read them back keyed by object id.
+const _operationMigrationSnapshots = new Map<string, { key?: string; version: string }>();
+export const migration = Migration.define({
```
dmaretskyi: export const migrations = [....[ -- export array even if its just one -- for extensibility

dmaretskyi (commented):
good to merge after comments are adressed

packages/core/echo/echo-registry/src/Registry.ts
```diff
@@ -0,0 +1,281 @@
...
+  /**
+   * Add or replace an object in the local registry.
+   * If an object with the same id already exists, it is replaced.
+   */
+  add(object: Obj.Unknown): void;
+
+  /**
+   * Add or replace many objects in the local registry.
+   */
+  addMany(objects: readonly Obj.Unknown[]): void;
```
richburdon: why support 2 verbs -- just pass array.

packages/core/echo/echo-registry/src/Registry.ts
```diff
@@ -0,0 +1,281 @@
...
+ * Only AST nodes that can be evaluated locally against an in-memory object collection are supported:
+ * - `select` clauses applied to plain {@link QueryAST.Filter} nodes (object, key, tag, props, etc.).
+ * - Boolean combinators (`and`, `or`, `not`).
+ * - `limit` clauses.
+ *
+ * Server-side concerns such as `from`, `order`, traversal, and text/timestamp filters are not supported.
+ */
+class RegistryQueryResult<T> implements QueryResult.QueryResult<T> {
+  constructor(
+    private readonly _registry: Registry,
```
richburdon: when do you use _ vs #

packages/core/echo/echo-registry/src/Registry.ts
```diff
@@ -0,0 +1,281 @@
...
+    case 'filter': {
+      const selection = executeQuery(registry, ast.selection);
+      return selection.filter((object) => matchFilter(ast.filter, object));
+    }
+    case 'limit': {
+      const inner = executeQuery(registry, ast.query);
+      return inner.slice(0, ast.limit);
+    }
+    default:
+      throw new Error(`Query clause '${ast.type}' is not supported by the in-memory registry.`);
```
richburdon: "Query clause not supported: ast.type"

packages/core/echo/echo-registry/src/Registry.ts
```diff
@@ -0,0 +1,281 @@
...
+import * as Effect from 'effect/Effect';
+import * as Layer from 'effect/Layer';
+
+import { Entity, type Obj, type Query, type QueryResult } from '@dxos/echo';
+import { filterMatchObjectJSON } from '@dxos/echo-pipeline/filter';
+import { type QueryAST } from '@dxos/echo-protocol';
+import { invariant } from '@dxos/invariant';
+
+/**
+ * Composable, in-memory registry of keyed ECHO objects.
```
richburdon, line 17: per-space or hypergraph global?

PR #11309, opened 2026-05-11
refactor(plugins): migrate all plugins to neutral platform build with /plugin entrypoint

wittjosiah:
## Summary

Migrates every DXOS plugin to a single, canonical build shape and unblocks the cli and Solid plugins along the way.

### The canonical plugin shape

Every plugin now follows the same structure:

- **`src/index.ts`** — types, schemas, util, meta; **does NOT export the Plugin constructor**.
- **`src/plugin.ts`** — thin `Plugin.lazy(meta, () => import('#plugin'))` wrapper, plus `XOperationHandlerSet` re-export (for plugins that have one). Reached via the `./plugin` subpath export.
- **`src/XPlugin.tsx`** — the eager plugin (`Plugin.define` + `AppPlugin.add*`). Reached via the `#plugin` import condition.
- **`src/XPlugin.node.ts`** — optional simplified node-only variant (no ReactSurface) reached when `#plugin` resolves under the `node` condition.
- **`src/capabilities/index.ts`** — full capabilities, all `Capability.lazy(...)`.
- **`src/capabilities/node.ts`** — optional, slimmed for node consumers (no ReactSurface chain) when reached via `#capabilities` `node` condition.
- **moon.yml** — `--platform=neutral` (plus `--platform=node` only for plugins whose `.node.ts` actually uses `node:*` builtins missing from `@dxos/node-std`).
- **package.json** — all subpath imports use the conditional `source/types/default` pattern; `#plugin` (and `#capabilities` when needed) carries the `node` condition.

Consumers must import the Plugin constructor via the `/plugin` subpath, never from main entry. `composer-app/plugin-defs.tsx` and `cli/util/blueprints.ts` updated accordingly.

### What this PR contains

- **Waves 1–5**: each batch of plugins switched from `--platform=browser,node` to `--platform=neutral` with the conditional subpath pattern.
- Move `capabilities`/`events`/`operations` definitions into `types/` with `@import-as-namespace` so they participate in cleaner barrel imports.
- Require the `/plugin` subpath for plugin constructor imports; drop the constructor re-export from main entry for every plugin.
- Re-export `XOperationHandlerSet` from `/plugin` (alongside the Plugin constructor) so cli/blueprints consumers don't reach into `/types` or the main entry.
- Restore the 8 `XPlugin.node.ts` files that earlier wave deletions removed (chess/deck/files/map/navtree/observability/thread/wnfs). Vitest test envs need them to avoid loading React/CSS-heavy capability modules in node.
- Wire `.node.ts` forks into `#plugin` map with the `node` condition so cli bundling (bun) picks up the slim definition.
- Add `--platform=node` only where actually needed (only `plugin-integration` uses `node:child_process`/`node:os`, both missing from `@dxos/node-std`).
- Special handling for plugin-script: factor `capabilities/node.ts` so bun's static import tracer doesn't walk into `Capability.lazy('ReactSurface', () => import('./react-surface'))` which transitively reaches `@dxos/echo-query/api.d.ts?raw` (a package-path raw import bun can't resolve).
- New plugin-bluesky migrated to canonical shape (merged in from main mid-PR).
- Fix circular import in plugin-feed: `types/FeedOperation.ts` was importing from `./index` (the barrel that re-exports it), causing TDZ on namespace re-exports.

### dx-compile: Solid JSX support

SWC's automatic React JSX runtime (driven by `jsxImportSource: 'solid-js'`) emits `_jsx(...)` calls + `import { jsx } from 'solid-js/jsx-runtime'`, but solid-js's `./jsx-runtime` subpath maps to `dist/solid.js` which has no `jsx`/`jsxs` exports — Solid JSX needs `babel-plugin-jsx-dom-expressions` (via `babel-preset-solid`) to compile into reactive primitives, not function calls.

New `BabelSolidTransformPlugin` runs Babel + `babel-preset-solid` + `@babel/preset-typescript` on `.tsx` files when the package's tsconfig sets `jsxImportSource: 'solid-js'`. Registered before `SwcTransformPlugin` so esbuild's first-match-wins picks Babel for `.tsx` and SWC for `.ts` (no JSX, ~5x faster). Affects all 7 solid packages (`plugin-map-solid`, `app-solid`, `solid-ui`, `solid-ui-geo`, `web-context-solid`, `effect-atom-solid`, `echo-solid`) automatically.

Verified plugin-map-solid's dist now contains proper Solid reactive primitives (`template`, `insert`, `effect`, `memo`, ... from `solid-js/web`) — zero `_jsx` calls, zero `solid-js/jsx-runtime` imports — and `composer-app:bundle` succeeds.

## Test plan

- [ ] `pnpm build` (full workspace) — passes
- [ ] `moon run cli:compile` — bun bundles cli with the new node-condition routing; no `@dxos/echo-query/api.d.ts?raw` resolution error
- [ ] `moon run composer-app:bundle` — production rolldown bundle succeeds; Solid chunks contain reactive primitives, not `_jsx`
- [ ] `moon run :lint -- --fix` — passes
- [ ] `pnpm format` — passes
- [ ] All plugin tests pass under `MOON_CONCURRENCY=4 moon run :test -- --no-file-parallelism` (`plugin-assistant`'s `trace-timeline.test.ts` snapshot failures are pre-existing on this branch and orthogonal — see CI)
- [ ] Manually load `composer-app` and confirm Solid Map plugin renders


packages/plugins/plugin-assistant/package.json
```diff
@@ -13,70 +13,86 @@
...
+    },
     "#plugin": {
-      "browser": "./src/AssistantPlugin.tsx",
-      "node": "./src/AssistantPlugin.node.ts",
-      "default": "./src/AssistantPlugin.tsx"
+      "source": {
+        "browser": "./src/AssistantPlugin.tsx",
+        "node": "./src/AssistantPlugin.node.ts"
+      },
+      "types": "./dist/types/src/plugin.d.ts",
```
dmaretskyi: you have to mirror "source" and dist output.

The valid structure would be

```

types: dist/src/AssistantPlugin.d.ts // pick one of them

source:  
  browser: "./src/AssistantPlugin.tsx",
  node: "./src/AssistantPlugin.node.ts"
browser: "./dist/lib/AssistantPlugin.js", // same as source
node: "./dist/lib/AssistantPlugin.node.js"    
```

dmaretskyi (dismissed):

PR #11320, opened 2026-05-12
fix(ci): close TanStack-style cache-poisoning supply chain attack vector

wittjosiah:
## Summary

Closes the TanStack-style npm supply chain attack vector identified during a GitHub Actions security audit.

**The attack chain:** a fork PR writes attacker-controlled binaries into the shared pnpm store cache via a `pull_request_target` workflow → the `publish-all` workflow restores that same cache → malicious binaries run with `id-token: write` in scope and extract the npm OIDC token.

### Fixes

**Cache isolation** (`cache-scope` input, default `ci`): `publish-all` now uses `cache-scope: publish`, so its pnpm store cache key lives in a completely separate namespace from PR caches. A cache entry written by any PR run can never be restored during a release.

**Integrity verification** (`verify-integrity` input, default `false`): `publish-all` sets `verify-integrity: true`, enabling pnpm's built-in checksum verification before any build that holds OIDC credentials.

**`pull_request_target` removed entirely**: `check.yml` now uses the standard `pull_request` event. For fork PRs, GitHub withholds secrets at the platform level — no gating, no elevated trigger needed. Token-presence conditions (`env.TRUNK_TOKEN != ''`) gracefully degrade Trunk analytics and Codecov uploads when secrets are unavailable.

**Permissions least-privilege**: `check` and `test` jobs are fully read-only. `pull-requests: write` is isolated to `check-report`, which runs no user code — it only downloads artifacts and posts the moon run report.

**Inlined `run-tests.yml`**: the reusable workflow was originally extracted to filter secrets in a separate environment; with `pull_request_target` gone that indirection has no purpose, so the steps are inlined back into `check.yml`.

**`persist-credentials: false`** on all checkout steps that run user build/test code.

## Reference

- [TanStack npm supply chain compromise postmortem](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem)

## Test plan

- [ ] Verify `check` / `test` workflows still populate and restore the pnpm cache normally (cache key prefix `ci-`)
- [ ] Verify `publish-all` uses a separate cache key (`publish-`) and runs pnpm install with integrity checks enabled
- [ ] Confirm fork PRs run tests without Trunk analytics (no token) and without failing
- [ ] Confirm `check-report` posts moon run timing for both check and test jobs on same-repo PRs


wittjosiah:
Re: CodeRabbit suggestion to move `moonrepo/run-report-action` into a dedicated reporting job —

Partially addressed in 77fa08c: `pull-requests: write` is now scoped to the `check` job only (removed from the top-level), so `e2e` and the reusable `test` job are genuinely read-only.

A fully separate reporting job would be cleaner, but `moonrepo/run-report-action` reads `.moon/cache/runReport.json` produced during the `check` run. That file would need to be uploaded as an artifact and restored in the reporting job, and it's unclear whether the action works with a pre-placed report vs. a live moon context. Deferring that restructure until we can verify the action's behaviour.

For the security scenario that matters most (fork PRs), GitHub enforces read-only tokens at the platform level regardless of what permissions are declared in the workflow — so the remaining exposure (lint/build steps in the `check` job running with `pull-requests: write`) only affects same-repo runs where the code is already trusted.

---

PR #11332, opened 2026-05-13
refactor: eager plugin stories + root-cause fixes for test stability

wittjosiah:
## Summary

Two threads, woven together because the second was exposed by the first.

1. **Eager plugin imports in stories.** Stories imported plugins from `/plugin` (lazy) entrypoints, which fail under storybook (vite-dev) because dynamic-import promises settle with partially-evaluated namespaces (see `@dxos/plugin-testing/src/core.ts`). Stories now import from `/testing`, which re-exports the eager plugin.
2. **Real teardown races, exposed by removing `dangerouslyIgnoreUnhandledErrors`.** The vitest base config had a project-wide flag that swallowed unhandled rejections from every test surface — node, browser, storybook. With the flag gone, a chain of latent races and config issues turned the test suite red. Each one is fixed at its root cause; nothing is re-suppressed.

After the merge, **no `dangerouslyIgnoreUnhandledErrors` anywhere** — node, browser, and storybook tests all observe unhandled rejections natively, so future regressions of this shape surface immediately instead of accumulating behind a flag.

## Eager plugin imports

- Added/updated `testing.ts` entrypoints to eagerly re-export the plugin for `assistant`, `automation`, `chess`, `explorer`, `inbox`, `map`, `markdown`, `search`, `table` (`client`, `space`, `preview`, `attention`, `settings`, `graph`, `theme` already had them).
- Added `./testing` exports in `package.json` and `src/testing.ts` entrypoints in `moon.yml` where missing.
- For plugins that already had a `src/testing/` directory of fixtures (`assistant`, `automation`, `inbox`), added a sibling `src/testing.ts` containing only the eager plugin re-export — internal fixtures untouched, public surface unchanged.
- Updated all `*.stories.tsx` across `packages/plugins/` and `packages/stories/` to import from `@dxos/plugin-X/testing`.

## Root-cause fixes for test stability

Affects **node, browser, and storybook test projects**; each fix is at the layer where the race or config error originates.

### Test runner / config

- **`fix(testing): resolve storybook subpath imports from source`** — added `#*` to the storybook project's `importSource` include and dropped the `dangerouslyIgnoreUnhandledErrors` safety net so future regressions surface. Without `#*`, `#translations`/`#meta` imports in a package's own stories fell through to `default` (`./dist/...`) and required the package's own `compile` task to have run first — flaky depending on moon scheduling order.
- **`fix(vite-plugin-import-source): narrow dist-importer skip to subpath imports`** — when the storybook fix above was first applied broadly, it caused `#storage` re-exports inside `@dxos/random-access-storage`'s compiled `dist/` to jump back to source and pull in node-only `node:path` in the browser. Skip is now scoped to `#*` imports originating from `/dist/`, leaving cross-package `@dxos/*` resolution unchanged.
- **`fix: patch @vitest/browser to guard birpc proxy against toJSON and symbols`** — phantom RPC calls from `toJSON` / Symbol-keyed introspection serialized to `m: undefined` and surfaced as `[birpc] function "undefined" not found`. Patch in `patches/@vitest__browser@4.1.5.patch` matches the maintainer-suggested fix in upstream vitest issue #10214.

### Cooperative cancellation

- **`fix(async,echo-pipeline): make DeferredTask.runBlocking cooperative on dispose`** — `runBlocking` used to throw `ContextDisposedError` on a disposed context. That throw escaped fire-and-forget callers as an unhandled rejection — observed as `Context disposed.` from `DataServiceImpl.updateIndexes` during `shell:test-browser` teardown. Now resolves cleanly on disposal (entry or mid-wait); `EchoHost.updateIndexes` gates its loop on `_ctx.disposed` so it doesn't busy-spin against the silent resolution. Used by node, browser, and storybook tests via every database flush.
- **`fix(echo-pipeline): handle leveldb iterator-after-close in loadRange`** — when a `SublevelDB` closes mid-iteration (external owner doesn't await the adapter's close first), level rejects `.next()` with `LEVEL_ITERATOR_NOT_OPEN`. Treat that specific error code as a graceful cancellation signal — same shape as the `DeferredTask` change; rethrow anything else.

### Subscription / cleanup wiring

- **`fix(plugin-space): bind zen-observable unsubscribe in spaces-ready cleanup`** — `SubscriptionList` stores raw cleanup functions and invokes them as plain calls, losing `this`. Storing `subscription.unsubscribe` directly meant zen-observable's `unsubscribe` ran with `this === undefined` and crashed on `this._state` during capability teardown. Capture the subscription locally and wrap unsubscribe in an arrow.

### Per-package config

- **`fix(plugin-script): override @typescript/vfs moduleResolution default`** — `createVirtualTypeScriptEnvironment` seeds `moduleResolution: NodeJs` (== deprecated `node10`), which TypeScript 6 throws as a compiler-options diagnostic. Override to `Bundler` at both `Compiler` callsites that didn't already set it.
- **`fix(react-ui-editor): await initSubduction before new Repo in Automerge story`** — the subduction-fork `Repo` constructor synchronously calls `set_subduction_logger` on a wasm reference that's still `undefined` if `initSubduction()` hasn't been awaited.

## Pulled in from main

- **#11335 (`fix(echo-pipeline): clean up per-collection state on connection close`)** — root-cause fix for `CollectionSynchronizer` leaking disconnected peers in `interestedPeers`/`lastQueried`/`lastBroadcast`. Replaces an earlier symptom-level fix on this branch (reverted in `89304784`).

## Pre-existing main failure addressed

- **`test(assistant-toolkit): regenerate memoized LLM conversations for blueprint.test`** — `Agent > planning` prompt had drifted from the cached conversation (different agent name, new `update-tasks` planning tool). Regenerated via the `regenerate-memoized-llm` skill. Same failure exists on `origin/main` (run [25816372214](https://github.com/dxos/dxos/actions/runs/25816372214) line 38919); this PR is just the first to get CI past the earlier storybook failures.

## Test plan

- [x] `moon run plugin-{chess,explorer,search,table,markdown,assistant,automation,inbox,map}:build` succeeds.
- [x] `moon run plugin-{...}:lint` succeeds for all touched plugins.
- [x] `moon run stories-assistant:build stories-ui:build` succeeds.
- [x] No remaining `@dxos/plugin-*/plugin` imports in story files.
- [x] `moon run plugin-space:test-storybook` — 8/8 over 3 reruns on a fresh tree.
- [x] Node tests: `client:test`, `client-services:test`, `echo-db:test`, `echo-pipeline:test`, `async:test` all green locally.
- [x] Browser tests: `shell:test-browser` green locally (was failing on `Context disposed.` before this branch).
- [x] Storybook tests: `plugin-space`, `plugin-kanban`, `plugin-script`, `plugin-markdown`, `plugin-transformer`, `plugin-client`, `plugin-deck`, `react-ui-editor` all green locally.
- [x] CI run [25830849930](https://github.com/dxos/dxos/actions/runs/25830849930) green end-to-end.

---
_Generated by [Claude Code](https://claude.ai/code/session_0122saaXFeUM2QUR9dHdKoWa)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown and resource cleanup to avoid stuck operations, hanging tasks, and teardown-related errors; prevents background work from continuing after resources close.
  * Suppressed phantom/unintended RPC/event calls and hardened test-run logging to avoid spurious failures.

* **New Features**
  * Added dedicated testing entrypoints/exports for numerous plugins to simplify Storybook and local testing.

* **Infrastructure**
  * Updated Storybook and build tooling to use testing variants and improved module resolution for dev builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dxos/dxos/pull/11332)


packages/common/async/src/task-scheduling.ts
```diff
@@ -57,15 +57,25 @@ export class DeferredTask {
...
   async runBlocking(): Promise<void> {
     if (this._ctx.disposed) {
-      throw new ContextDisposedError();
+      return;
     }
 
     this.schedule();
-    await this._nextTask.wait();
+
+    let removeDisposeHandler: (() => void) | undefined;
```
dmaretskyi: I believe this is incorrect - the caller should wait for the task to finish regradless of the context disposing.

Context dispose will cause the caller to exit early if it cooperates, but this is best effort, and may hapen with a delay or not happen at all.

Having the task run while the caller of `runBlocking` believes it has finished may lead to more unhandled rejections.

wittjosiah: You're right — my change broke the contract. The `Promise.race` resolves `runBlocking` on disposal while the task callback may still be in flight (`scheduleTask` only `clearTimeout`s the *pre-run* timer; if the callback was already executing, it keeps running). The caller proceeds thinking the work finished, and any later rejection from the still-running task becomes unhandled — exactly the regression I was trying to fix.

Reverted in 1f82a75d05. `runBlocking` once again throws `ContextDisposedError` at entry and waits unconditionally on `_nextTask` (with the original deadlock-window for *post-schedule, pre-run* disposal that I shouldn't have papered over here).

The original symptom on this branch was a `Context disposed.` unhandled rejection from `shell:test-browser` teardown of "multiple concurrent invitations" — `_createSpaceInternal → flush → updateIndexes → runBlocking` fires after the EchoHost's ctx has been disposed by the test fixture. That's a real teardown-ordering bug but the right place to fix it is at the layer that's calling `flush` against an already-shut-down host (or by making client-side teardown await in-flight flushes), not by changing `DeferredTask`'s contract. I'm going to leave that one to surface in CI and address at the call-site level — please let me know if you want it kept on this PR or split out.

packages/core/echo/echo-pipeline/src/automerge/leveldb-storage-adapter.ts
```diff
@@ -134,12 +144,18 @@ export class LevelDBStorageAdapter extends Resource implements StorageAdapterInt
...
+    try {
+      for await (const [key] of this._params.db.iterator<StorageKey, Uint8Array>({
+        gte: keyPrefix,
+        lte: [...keyPrefix, '\uffff'],
+        ...encodingOptions,
+      })) {
+        batch.del<StorageKey>(key, { ...encodingOptions });
+      }
+    } catch (err: any) {
+      if (!isLevelDbIteratorNotOpenError(err)) {
```
dmaretskyi: I think swallowing errors here is the wrong way to tackle those -- the consumer should not be calling methods on closed storage. 
Throwing NOT_OPEN errors indicates race codition in the code above

tools/vite-plugin-import-source/src/index.ts
```diff
@@ -76,6 +76,17 @@ const PluginImportSource = ({
...
         }
 
+        // Don't re-route `#*` subpath imports to source when the importer
+        // is already in a compiled `dist/` tree. Compiled packages expect
+        // their own subpath imports to stay on the dist→dist chain;
+        // jumping back to source would pull in TypeScript that may not be
+        // browser-safe (e.g. raw `node:path` in `random-access-storage`'s
+        // src). Non-subpath `@dxos/*` imports from dist are unaffected —
+        // they still benefit from source resolution.
+        if (source.startsWith('#') && importer.includes('/dist/')) {
```
dmaretskyi, line 86: this is good, but also when do we ever go into dist for packages that have source set? this hints a misconfigured package

packages/common/async/src/task-scheduling.ts
wittjosiah: Re-addressed at the lifecycle layer in c80fc3fa1d — `DeferredTask.runBlocking` is untouched. Two changes in `EchoHost`:

1. `_runUpdateIndexes` now sets `_indexesUpToDate = true` at each cooperative exit (the three `_ctx.disposed || !isOpen` short-circuits). The `DataServiceImpl.updateIndexes` RPC handler's `do { await runBlocking() } while (!_indexesUpToDate)` loop sees the flag and exits cleanly instead of spinning into the next `runBlocking()` against a disposed ctx.
2. `_close` `await`s `_updateIndexes.join()` before the `Resource` base disposes `_ctx`, so any in-flight indexer pass finishes (cooperatively setting the flag) while the ctx is still alive — implementing the TODO at task-scheduling.ts:17.

This covers mid-task and post-task disposal without changing `runBlocking`'s contract that the caller always waits for the task to finish. CI verifying.

packages/core/echo/echo-pipeline/src/automerge/leveldb-storage-adapter.ts
wittjosiah: Right — that error code was the symptom; the race lives upstream. Walking the stack of the actual failure (`client:test` → `lazy-space-loading.test.ts` → "can access closed space information"):

```
Iterator.next                  abstract-iterator.js:96
LevelDBStorageAdapter.loadRange leveldb-storage-adapter.ts:116
StorageSubsystem.loadDocData    @automerge/automerge-repo/.../StorageSubsystem.js:83
StorageSubsystem.loadDoc        @automerge/automerge-repo/.../StorageSubsystem.js:130
```

The caller is `@automerge/automerge-repo`'s internal `StorageSubsystem`, kicked off by `_repo.find(...)`. `AutomergeHost._close` awaits `_repo.shutdown()` and `_storage.close?.()`, but `_repo.shutdown()` doesn't actually wait for in-flight `loadDocData → loadRange` iterations to drain, and the base `Resource` close on the adapter doesn't either. The kv owner upstream then closes the sublevel while a `for await` is still pending.

Fixed in dd718f6563 by tracking in-flight `loadRange`/`removeRange` iterations in the adapter and `await`ing them in `_close`. New calls after close already short-circuit on `!isOpen` (the `Resource` base flips state before invoking `_close`), so this is sufficient — by the time `adapter.close()` resolves, there are no in-flight iterations for the upstream kv close to race with. The catch is gone.

PR #11393, opened 2026-05-19
react-ui: runtime icon resolver for plugin-reachable icons

wittjosiah:
## Summary

- Adds a runtime icon resolver in `@dxos/react-ui` so any Phosphor icon resolves on demand, even when the host app's build-time sprite scanner never saw the reference. This unblocks plugins loaded dynamically from the external registry (e.g. the YouTube plugin) — previously their icons would silently fail because the scanner only walks `{packages,tools}/**`, never `node_modules` or runtime plugin code.
- Switches `Icon` (React) and `<dx-icon>` (Lit) from cross-document `<use href="/icons.svg#name">` to same-document `<use href="#name">`, backed by a hidden in-DOM defs container managed by a new `IconRegistry` wired into `ThemeProvider`. Stories inherit for free via `withTheme`.
- The registry is a **refcounted document-level singleton**: `ThemeProvider` mounts many times per document (editor tooltip roots, shell, dialogs), and all mounts share one registry — one `/icons.svg` fetch, no duplicate symbol ids. A `globalThis.__dxIconRegistry` bridge lets `@dxos/lit-ui`'s `<dx-icon>` use the same registry without a package dependency.
- `IconsPlugin` gains an opt-in, icon-set-agnostic `assets` option (`{route, dir}` pairs) that serves raw catalogs in dev and copies them into the build output; composer-app passes `{ route: '/phosphor', dir: <@phosphor-icons/core assets> }`. The catalog (~9,000 SVGs) is **excluded from the PWA precache** — precaching it would add one install-time request per file — and is instead cached at runtime by the service worker (workbox `CacheFirst` + `ExpirationPlugin`), so any icon the app has rendered once stays available offline.

## How it works

`useIconHref` subscribes to the registry via `useSyncExternalStore` and returns `undefined` until the symbol is present, firing `requestIcon(name)` on a miss; `<dx-icon>` follows the same contract with its own registry subscription, omitting the `<use>` until the symbol lands rather than leaving a transient broken ref. The resolver waits for the static `/icons.svg` sprite to finish ingesting, then — if the icon still isn't present — fetches `/phosphor/{weight}/{name}.svg`, constructs a `<symbol>`, appends it to the in-DOM defs, and notifies subscribers to re-render. In-flight fetches are deduped, and permanent misses (unsupported prefix, 404, malformed source) are cached so a bad icon name doesn't refetch on every render; network errors stay retryable.

## Verification

- Composer-app preview build: `ph--youtube-logo--regular` (confirmed absent from both `public/icons.svg` and `out/composer/icons.svg`) renders correctly in the Registry view, sourced from a single fetch to `/phosphor/regular/youtube-logo.svg`.
- 1,512 regular-weight Phosphor SVGs (plus the other 5 weights) are copied into `out/composer/phosphor/` by the `closeBundle` hook of the assets option.
- Existing static-sprite icons still render identically — same fetch of `/icons.svg`, now ingested into the in-DOM defs node instead of referenced cross-document.
- Network audit showed real-world icons already living outside the static scan (`lightning-slash`, `circle-duotone`) being resolved correctly with one fetch each.
- CI green: build, test, lint across affected packages; react-ui unit tests exercise the provider path.

## Out of scope (follow-ups)

- Adopt the `assets` option in the other hosts (tasks-app, testbench-app, storybook) — one line each; until then the runtime resolver falls back gracefully (404) outside composer-app. composer-crx deliberately omits it (copying a full catalog would bloat the packaged extension).
- Runtime resolution for non-Phosphor icon sets (`dx--*`, plugin namespaces) — the vite-side `assets` option is already set-agnostic; the react-ui resolver currently only auto-fetches `ph--*`.
- SSR support for the in-DOM sprite (none of our hosts render `Icon` server-side today).
- Replacing `@ch-ui/icons` with `phosphor-icons-tailwindcss` (existing TODO at `tools/vite-plugin-icons/src/index.ts:7`).

## Test plan

- [ ] Build composer-app in production mode and load the YouTube plugin from the external registry; confirm the icon renders and exactly one `/phosphor/regular/youtube-logo.svg` fetch happens.
- [ ] Hard-refresh the app and confirm existing sidebar/navigation icons still render (static-sprite path unchanged).
- [ ] Render an `<Icon icon="ph--airplane-tilt--duotone" />` (not in scan) and confirm `/phosphor/duotone/airplane-tilt-duotone.svg` is fetched once.
- [ ] Offline check: after rendering a runtime-resolved icon once, go offline and reload — the icon should render from the service-worker cache.
- [ ] Verify `<dx-icon>` (Lit) still works in non-React contexts (e.g. rich-text inline icons), including icons that resolve after first render.

## Summary by CodeRabbit

* **New Features**
  * Phosphor icons now load on demand, reducing initial icon resources.
  * Dynamically loaded icons are cached for reliable offline availability after first use.
  * Icon references now resolve consistently across Composer, React, and Lit components.

* **Refactor**
  * Icon resolution now uses a shared registry, improving reliability for dynamically loaded icons.
  * Icons that are not yet available now render automatically when loading completes.


packages/apps/composer-app/vite.config.ts
```diff
@@ -410,6 +417,58 @@ export default defineConfig((env) => ({
...
+/**
+ * Serves the full Phosphor icon catalog as individual SVGs under `/phosphor/{weight}/{name}.svg`,
+ * so the runtime icon resolver in @dxos/react-ui can fetch glyphs that weren't statically
+ * referenced (e.g. icons used only by runtime-loaded plugins).
+ *
+ * In dev: middleware streams from node_modules. In build: assets are copied into the output dir.
+ *
+ * TODO(wittjosiah): Roll this into @dxos/vite-plugin-icons so other apps (tasks-app,
+ *   testbench-app, composer-crx) get the same behavior without duplicating the plugin.
+ */
```
richburdon: or integrate with our theme plugin?

packages/ui/react-ui/src/hooks/useIconHref.ts
```diff
@@ -2,12 +2,33 @@
 // Copyright 2025 DXOS.org
 //
 
-import { useThemeContext } from './useThemeContext';
+import { useCallback, useSyncExternalStore } from 'react';
 
-const ICONS_URL = '/icons.svg';
+import { useIconRegistry } from '../components/ThemeProvider/IconRegistry';
```
richburdon: just `../components`

packages/ui/react-ui/src/components/ThemeProvider/IconRegistry.tsx
```diff
@@ -0,0 +1,230 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+import React, { type PropsWithChildren, createContext, useContext, useEffect, useMemo, useState } from 'react';
```
richburdon: use radix context

packages/ui/react-ui/src/components/ThemeProvider/IconRegistry.tsx
```diff
@@ -0,0 +1,230 @@
...
+  return {
+    registry,
+    dispose: () => {
+      listeners.clear();
+      defs.remove();
+    },
+  };
+};
+
+export const IconRegistryProvider = ({ children }: PropsWithChildren) => {
```
richburdon: add comment describing purpose

packages/ui/react-ui/src/components/ThemeProvider/IconRegistry.tsx
```diff
@@ -0,0 +1,230 @@
...
+    setRegistry(handle.registry);
+    return () => {
+      if (getIconRegistry() === handle.registry) {
+        setActiveRegistry(undefined);
+      }
+      handle.dispose();
+    };
+  }, []);
+
+  const value = useMemo(() => registry, [registry]);
```
richburdon: this isn't required.

packages/apps/composer-app/vite.config.ts
wittjosiah: Extracted in 742730a18 — the plugin now lives in `@dxos/vite-plugin-icons` as `PhosphorAssetsPlugin` rather than the theme plugin, since that package already owns the Phosphor assets path (`IconsPlugin`'s `assetPath` resolves into the same catalog). Other hosts (tasks-app, testbench-app, composer-crx, storybook) can now adopt it with one line; composer-app is the only consumer in this PR.

---

wittjosiah: Revised in 991d47888: now an icon-set-agnostic, opt-in `assets` option on `IconsPlugin` (`{route, dir}` pairs) instead of a separate phosphor-specific plugin — hosts that only need the static sprite omit it (e.g. composer-crx, where copying a catalog would bloat the packaged extension).

---

PR #11410, opened 2026-05-20
fix(views): use Filter.typename for RuntimeType schemas in Table/Kanban/Map/Masonry queries

wittjosiah:
## Summary

- `Filter.type(runtimeType)` emits an object-DXN filter (`dxn:echo:@:{id}`) because database-registered schemas carry an object-based `TypeIdentifierAnnotation`
- Objects created from the original Effect schema are tagged with a type-DXN (`dxn:type:{typename}:{version}`), so `compareTypename` falls back to strict `DXN.equals` — which never matches
- Fix: detect `RuntimeType` and use `Filter.typename(schema.typename)` instead, which produces a type-DXN filter that `compareTypename` handles with optional version matching

Affects `TableArticle`, `KanbanArticle` (`ViewKanbanArticle`), `MapArticle`, and `MasonryContainer`.

The root cause is documented in a TODO in `TableArticle`:
> This should use the full query AST directly. That currently doesn't work for dynamic schema objects because their indexed typename is the schema object DXN.

This fix handles the inverse case: objects indexed by type-DXN (created from an unregistered Effect schema) are invisible to `Filter.type(runtimeType)`. The `instanceof Type.RuntimeType` guard selects the correct filter kind at the call site.

## Test plan

- [ ] Open a space with a custom ECHO schema registered via `schemaRegistry.register()` (or loaded from a snapshot with a `PersistentSchema` object)
- [ ] Open a Table view bound to that schema — rows should appear
- [ ] Open a Kanban view bound to that schema — cards should appear in columns
- [ ] Verify Map and Masonry views likewise show objects for database-registered schemas
- [ ] Verify existing compiled schemas (Task, Person, etc.) continue to work normally in all four view types

## Summary by CodeRabbit

* **New Features**
  * Added a schema-based filter hook to standardize item queries across views.

* **Bug Fixes**
  * Ensured Kanban, Map, Masonry, and Table views use the standardized filter, improving correctness for runtime vs. standard schema types and fixing query dependency issues.

* **Chores**
  * Exposed the new filter hook in the UI toolkit for reuse.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dxos/dxos/pull/11410?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/plugins/plugin-kanban/src/containers/KanbanArticle/KanbanArticle.tsx
+++ b/packages/plugins/plugin-kanban/src/containers/KanbanArticle/KanbanArticle.tsx
@@ -7,7 +7,7 @@ import React, { useCallback, useContext, useMemo } from 'react';
 
 import { useCapabilities, useOperationInvoker } from '@dxos/app-framework/ui';
 import { AppCapabilities } from '@dxos/app-toolkit';
-import { type AppSurface } from '@dxos/app-toolkit/ui';
+import { useSchemaFilter, type AppSurface } from '@dxos/app-toolkit/ui';
 import { Filter, Obj, Query, type Ref, Type } from '@dxos/echo';
 import { AtomObj, AtomQuery } from '@dxos/echo-atom';
 import { useObject, useSchema } from '@dxos/react-client/echo';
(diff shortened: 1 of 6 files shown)


wittjosiah:
The TODO is intentional — it documents the known limitation that `useSchemaFilter` works around (using the full query AST doesn't work yet for `RuntimeType` schemas because their indexed typename is the schema object DXN, not a type-DXN). Leaving the TODO in place as a marker for a future fix to the underlying filter matching.

wittjosiah:
Fixed — changed `[ast, schema]` to `[ast, baseFilter]` in the `useQueryWorkaround` memo dep array (commit e59771c).

PR #11608, opened 2026-05-29
refactor(plugin-navtree): remove primary action column from navtree items

wittjosiah:
## Summary

There should be no "primary" actions in the navtree. This change removes the dedicated primary-action column from navtree items so item names get more horizontal space. Actions now follow a single rule everywhere:

- **One action** → rendered directly as an inline icon button.
- **Multiple actions** → collapsed into a `…` dropdown menu.

## Changes

**`NavTreeItemColumns.tsx`**
- Removed the primary/secondary disposition sorting and the dedicated primary-action `Treegrid.Cell`.
- All `list-item` / `list-item-primary` actions are flattened into a single `allActions` list (groups expanded via `groupedActions`).
- Renders `NavTreeItemMonolithicAction` for a single action and `NavTreeItemActionDropdownMenu` for multiple.

**`L1Panel.tsx`** (workspace header + tree grid)
- Folded the former primary action into the existing `MenuActions` component, which already renders 1 action inline and >1 as a menu.
- Dropped the extra `min-content` column from the tree grid template (4 → 3 effective action/end columns reduced to 2).
- Header action columns switched from fixed `28px` to `min-content` and `pe-1` removed so the header action buttons align with the tree's action column and the right edge.
- Simplified `useL1MenuActions` (no longer returns `primaryAction`/`groupedActions`).

## Notes for reviewers
- No new type casts introduced.
- `MenuActions` already had the "1 inline / >1 menu" behavior — this change just feeds it the full action set instead of only the secondary actions.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined navigation tree item action rendering so single actions appear inline while multiple actions use a consolidated dropdown.
  * Simplified sidebar L1 header action area to use a unified menu and adjusted header layout for improved alignment.
  * Consolidated action filtering/ordering logic for more consistent action visibility and behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dxos/dxos/pull/11608?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


wittjosiah:
Addressed both nitpicks in be1588503c:
- Extracted the duplicated flatten/filter pipeline into a shared `getListActions` helper in `useActions.ts`, now used by both `NavTreeItemColumns` and `useL1MenuActions`.
- `L1Panel` now imports `NavTreeItemColumns` directly from its defining module instead of through the barrel.

PR #11647, opened 2026-06-01
feat(client): add replicated space.delete() (tombstone)

wittjosiah:
## Summary

Today `space.close()` is a **purely local** action — it persists `SPACE_INACTIVE` in the local `MetadataStore` and stops replication, but it does **not** sync to the user's other devices. This PR adds `space.delete()`: a **replicated tombstone**. Deleting on one device records a `SpaceDeleted` credential in the identity's HALO, which replicates to every other device of the same identity; each device then stops replicating and unloads the space.

It is a **tombstone, not a hard delete** — no data is removed until future garbage collection — and it is terminal (re-join via a fresh invitation is still possible).

This rides an existing mechanism: space membership is already recorded as a `SpaceMember` credential in the HALO, and the `_deviceSpaceSync` credential processor reacts to it on every device to `acceptSpace(...)`. We add a `SpaceDeleted` credential and teach that same processor to **tombstone instead of accept**.

## What changed

**Scope (confirmed during design):** HALO-only v1. Cross-member "delete for everyone" (control-feed tombstone + issuer authority), garbage collection, and undelete are deferred follow-ups.

- **Protos:** new `SpaceState.SPACE_DELETED`; new `dxos.halo.credentials.SpaceDeleted` assertion; new persisted `EchoMetadata.deleted_spaces` tombstone list (needed because a fresh device may see the tombstone before the `SpaceMember` credential).
- **MetadataStore:** `addDeletedSpace` / `deletedSpaces` (survives restart).
- **DataSpace.delete:** mirrors `deactivate`, persists `SPACE_DELETED`, tears down subsystems; `activate` cannot resurrect a deleted space.
- **DataSpaceManager:** `markSpaceDeleted` (signs + records the HALO credential, unloads locally), `handleRemoteSpaceDeleted` (applies a tombstone replicated from another device), `isSpaceDeleted`; `_open` skips deleted spaces; `acceptSpace` refuses them.
- **HALO wiring (`service-context.ts`):** `_deviceSpaceSync` tombstones on `SpaceDeleted` and skips re-accepting deleted spaces (handles out-of-order credential replay across device feeds).
- **API:** `Space.delete()` on the client-protocol interface; `SpaceProxy._deleteInternal`; a `SPACE_DELETED` case in the existing `updateSpace` RPC (no new RPC).
- **`space-list.ts`:** now removes proxies that disappear from the `querySpaces` snapshot (previously it only ever added — so a deleted space never left `client.spaces`).
- **Composer UI:** adds **Delete Space** to the space-settings Danger Zone with a two-step confirmation dialog. Removes the **Archive Space** control, and removes the **Reset Space** action and **Space contents** panel (including the underlying `SpaceOperation.Reset` operation, handler, and test) — these were stopgaps for the missing delete capability.

## Testing

- **Cross-device acceptance test** (`packages/sdk/client/src/tests/space-delete.test.ts`): two clients on one identity (joined via device invitation) → create space on device 1, observe it on device 2 → `space.delete()` on device 1 → it vanishes from both. This is the definition of done.
- **Unit tests:** `MetadataStore` tombstone round-trip + restart; `DataSpaceManager` delete/restart/out-of-order-guard/remote-tombstone.
- Full suites green after merging `origin/main`: client (186), client-services (117), echo-pipeline (233), plugin-space (18). No casts introduced.
- **Manually verified in Composer:** app boots clean; Danger Zone shows only **Delete Space**; the "Delete this space? … cannot be undone" confirmation dialog appears on click.

## Notes for reviewers

- The `space-list.ts` removal-reconciliation change affects all space-list updates (not just deletes); `querySpaces` always emits a full snapshot, so absence ⇒ removed.
- `SpaceOperation.Reset` had no consumers outside the settings UI, so it was removed entirely rather than left as dead code.

## Summary by CodeRabbit

* **New Features**
  * Space deletion: users can delete non-personal spaces from settings; deletion replicates across devices, is tombstoned, prevents reopening/rejoining, and is exposed via SDK/client APIs.

* **Removed Features**
  * Removed "Reset Space" functionality.
  * Archive/unarchive controls and related UI removed.

* **UX**
  * Danger zone updated with delete confirmations and messages.

* **Tests**
  * Added tests for deletion, idempotence, persistence, and cross-device propagation.


packages/plugins/plugin-space/src/containers/SpaceSettingsContainer/SpaceSettingsContainer.tsx
```diff
@@ -194,20 +149,20 @@ export const SpaceSettingsContainer = ({ space }: AppSurface.SpaceArticleProps)
...
     } catch (err) {
-      log.catch(err, { stage: 'reset: invocation rejected', spaceId: space.id });
+      log.catch(err, { stage: 'delete: invocation rejected', spaceId: space.id });
       throw err;
     }
-    setResetConfirmOpen(false);
-  }, [space, invokePromise]);
-  const handleResetCancel = useCallback(() => setResetConfirmOpen(false), []);
+    setDeleteConfirmOpen(false);
+  }, [space, client, invokePromise]);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in 8979e9969f: `handleDelete` no longer rethrows (so the onClick handler can't produce an unhandled rejection), always closes the dialog, and shows a failure toast via `LayoutOperation.AddToast` wiring `delete-space-failed.message`.

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -393,6 +399,67 @@ export class DataSpaceManager extends Resource {
     return space;
   }
 
+  /**
+   * Whether the space has been tombstoned (soft-deleted). Deleted spaces are never opened or replicated.
+   */
+  isSpaceDeleted(spaceKey: PublicKey): boolean {
+    return this._metadataStore.deletedSpaces.some((key) => key.equals(spaceKey));
+  }
```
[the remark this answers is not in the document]
wittjosiah: Fixed in 8979e9969f: `isSpaceDeleted()` now also returns true when the persisted `SpaceMetadata.state === SPACE_DELETED`, matching the predicate in `_open`.

packages/plugins/plugin-space/src/operations/delete.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { SpaceOperation } from './definitions';
+
+const handler: Operation.WithHandler<typeof SpaceOperation.Delete> = SpaceOperation.Delete.pipe(
+  Operation.withHandler((input) =>
+    Effect.promise(async () => {
+      await input.space.delete();
+    }),
+  ),
+);
```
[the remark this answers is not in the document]
wittjosiah: Declining for consistency: every sibling lifecycle handler in this folder (close, open, snapshot, create, migrate, …) uses `Effect.promise`. The operation invoker surfaces a rejection to the caller's promise, and the only caller (`handleDelete`) now catches it and toasts on failure — so switching just `delete` to `tryPromise` would diverge from the established pattern without changing observable behavior.

packages/sdk/client-services/src/packlets/spaces/data-space.ts
```diff
@@ -622,6 +622,25 @@ export class DataSpace {
     this.stateUpdate.emit();
   }
 
+  /**
+   * Tombstones (soft-deletes) the space. Persists the deleted state and tears down all subsystems.
+   * Unlike {@link deactivate}, this is terminal: {@link activate} will not re-open a deleted space.
+   * Data is not removed until garbage collection (future work).
+   */
+  @synchronized
+  async delete(ctx: Context): Promise<void> {
```
dmaretskyi: separate deletion from teardown

you have a race here -- open() and close() use the resource mutext
while `@synchronized` uses a different mutex

packages/sdk/client-services/src/packlets/spaces/data-space.ts
```diff
@@ -622,6 +622,25 @@ export class DataSpace {
...
+    if (this._state === SpaceState.SPACE_DELETED) {
```
dmaretskyi, line 634: have a `delete()` method just put it into a terminal state and a separte space.close() to close it

packages/sdk/client-services/src/packlets/spaces/data-space.ts
wittjosiah: Done in c1e8fe3722: `delete()` no longer calls `_close()`. It now only does the terminal state transition; teardown is owned by the canonical `close()` path, which `DataSpaceManager` invokes separately when tombstoning — so `delete()` no longer drives inner-space (resource-lifecycle) teardown under the `@synchronized` mutex.

packages/sdk/client-services/src/packlets/spaces/data-space.ts
wittjosiah: Done in c1e8fe3722: `DataSpace.delete()` just moves to the terminal SPACE_DELETED state (persist + runtime). `DataSpaceManager._tombstoneSpace` now does `if (space.isOpen) await space.close(ctx)` then `await space.delete()` as separate steps.

PR #11678, opened 2026-06-04
fix(ui-theme): restore popover hover, calendar striping, and toast dismiss

richburdon:
## Summary

The monotonic-elevation retune ([#11662](https://github.com/dxos/dxos/pull/11662) / [#11643](https://github.com/dxos/dxos/pull/11643)) collapsed several previously-distinct surface tokens to the same elevation level, surfacing three regressions. Root cause for all three: tokens that used to differ now resolve to identical colors.

### 1. Hover invisible inside popover / dialog
`--color-hover-surface` (neutral-750 dark) equalled `--color-popover-surface` (elevation-7, also neutral-750) → the hover highlight vanished. Sidebars already re-derive hover off their surface (`.dx-main-sidebar`); popover/modal did not. Now `.dx-modal-surface` / `.dx-popover-surface` re-derive `hover` / `current` / `current-hover` off the actual `--surface-bg` via relative oklch, so highlights stay visible at any elevation.

### 2. Calendar month striping invisible
`Calendar` zebra-stripes months with `bg-group-surface` vs `bg-group-alt-surface`, but both mapped to `--dx-elevation-4` (identical). Gave `--color-group-alt-surface` a ±0.03 oklch offset so the banding tracks the ramp and is visible again.

### 3. Toast close took ~1s to dismiss
The close button called `onOpenChange(false)` directly without flipping Radix's `open`, so the 100ms exit animation never played and the toast sat fully visible until a hardcoded `setTimeout(…, 1000)` removed it. Now the toast controls its `open` state (exit animation plays), and the removal timeout is shortened to 150ms to match.

### Also included
Two in-flight fixes folded in at the user's request:
- boot-loader RSS panel: `text-wrap: balance`
- `MosaicVirtualStack` scroll-into-view: offset by the inter-item `gap` so the tile lands one gap below the viewport edge.

## Verification
Verified in storybook (dark mode) via Playwright:
- Menu item hover now L=0.37 vs popover surface L=0.32 (was identical).
- Calendar: June cells L=0.283 vs July L=0.253 — banding restored.
- `ui-theme`, `plugin-deck`, `react-ui-mosaic`, `app-framework` all build + lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Faster, more reliable toast dismissal and unified close handling

* **New Features**
  * Added toolbar regions to draft and message editing screens
  * Enhanced modal/popover surface visuals for clearer hover/current states

* **Style**
  * Balanced text wrapping in the loading screen
  * Improved scrolling behavior for stacked lists
  * Adjusted theme tokens for alternate surface (zebra striping) consistency

### DIFF
--- a/packages/plugins/plugin-deck/src/capabilities/react-root.tsx
+++ b/packages/plugins/plugin-deck/src/capabilities/react-root.tsx
@@ -20,22 +20,27 @@ export default Capability.makeModule(() =>
 
         const handleDismissToast = useCallback(
           (id: string) => {
-            const index = state.toasts.findIndex((toast) => toast.id === id);
-            if (index !== -1) {
-              // Allow time for the toast exit animation (animate-toast-hide, 100ms) before unmounting.
-              // TODO(burdon): Factor out and unregister timeout.
-              setTimeout(() => {
-                updateEphemeral((s) => {
-                  const toastToRemove = s.toasts[index];
-                  const newCurrentUndoId = toastToRemove?.id === s.currentUndoId ? undefined : s.currentUndoId;
-                  return {
-                    ...s,
-                    currentUndoId: newCurrentUndoId,
-                    toasts: s.toasts.filter((_, i) => i !== index),
-                  };
-                });
-              }, 150);
+            if (!state.toasts.some((toast) => toast.id === id)) {
+              return;
             }
+            // Allow time for the toast exit animation (animate-toast-hide, 100ms) before unmounting.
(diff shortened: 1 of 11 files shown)


richburdon:
Good catch — fixed in 8e2de88c0d. The delayed removal now re-resolves the toast by `id` inside `updateEphemeral` (and no-ops if it's already gone) instead of using the captured pre-delay `index`, so a toast added/removed during the 150ms exit-animation window can't cause the wrong one to be removed.

PR #11685, opened 2026-06-04
fix(functions): rename FunctionWrappingOptions.types to entities (Entity.Unknown[])

wittjosiah:
## Summary

- Renames `FunctionWrappingOptions.types?: Type.AnyEntity[]` → `entities?: Entity.Unknown[]`
- `Type.AnyEntity` was too narrow — it only covers ECHO schema type definitions, which prevented passing blueprint instances (and other runtime ECHO objects) without a cast
- `Entity.Unknown[]` is the correct type: it covers the full entity universe (schema types + runtime objects)
- `func.types` (from `Operation.Definition`) still requires a DB — those are schema types that need ECHO to decode objects, unchanged
- Caller-supplied entities work without a DB too: the no-DB path uses `makeRegistry({ initial: opts.entities })` so e.g. blueprints are resolvable via `Blueprint.resolve()` even without a space context

## Test plan

- [ ] `packages/core/compute/functions` type-checks
- [ ] Existing `wrapFunctionHandler` tests still pass
- [ ] Edge `compute-intrinsics` can drop its `as Entity.Unknown[] as Type.AnyEntity[]` cast once this lands on npm

## Summary by CodeRabbit

* **Refactor**
  * Updated function wrapping options API: replaced `types` parameter with `entities` for specifying function schema requirements
  * Improved registry initialization for both database-backed and in-memory configurations to support entity registration
  * Updated logging to track entity count instead of type count

### DIFF
--- a/packages/core/compute/functions/src/protocol/protocol.ts
+++ b/packages/core/compute/functions/src/protocol/protocol.ts
@@ -20,7 +20,7 @@ import {
   Trace,
 } from '@dxos/compute';
 import { LifecycleState, Resource } from '@dxos/context';
-import { Database, Feed, JsonSchema, Ref, Registry, type Type } from '@dxos/echo';
+import { Database, Entity, Feed, JsonSchema, Ref, Registry } from '@dxos/echo';
 import {
   createFeedServiceLayer,
   EchoClient,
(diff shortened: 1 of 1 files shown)


wittjosiah:
Closing — the cast cleanup is better handled locally in the edge repo rather than as an upstream protocol change.

PR #11719, opened 2026-06-08
feat(ui-theme): sidebar text tokens + elevated-surface state color tuning

richburdon:
## What

Theme/design-system color tuning, all in `packages/ui/ui-theme`:

- **Sidebar text tokens.** Adds `--color-sidebar-fg` and `--color-sidebar-current-fg` and applies them on `.dx-main-sidebar` (layout/main.css). Previously the L1 navtree had *no* text-color rule — it inherited the critical `<html>` pre-paint fallback, which is why devtools showed an inherited color with no class. Now the sidebar has an explicit, semantic foreground, and current/selected items (`aria-current` / `aria-selected`) read brighter than default body text.
- **Critical fallbacks synced.** `plugins/main.css` (the styles the Vite `ThemePlugin` injects into `<head>` before stylesheets load) were stale: zero-chroma `neutral-50`/`neutral-950`, out of sync with the theme. They now mirror `--color-base-fg` / `--color-base-surface` and carry the ramp tint (chroma 0.001, hue 190), so first paint matches the loaded theme.
- **Elevated-surface state colors.** Tunes the modal/popover hover & current re-derivations in surface.css and lowers `--color-popover-surface` to elevation-6; adjusts the dark `--dx-hover-surface-base`.

## Why

The L1 sidebar text rendered brighter than intended (and with the wrong tint) because it was riding an inherited, out-of-sync `<html>` fallback rather than a real token. Giving the sidebar its own semantic fg tokens fixes that and lets current/selected items be emphasized independently of base body text. The fallback sync removes a first-paint color flash. The elevated-surface tweaks keep hover/current highlights visible on popover/modal surfaces.

## Reviewer notes

- All changes are CSS tokens / declarations — no TS, no casts.
- `@theme` token changes are **not reliably HMR'd** — a dev-server restart is required to regenerate the `:root` custom properties (see the existing note near the grid tokens in semantic.css).
- The cool ramp tint (hue 190) means exact warm target hexes can't be matched by level selection alone; lightness is matched to the nearest ramp stop.

## Summary by CodeRabbit

* **Style**
  * Increased contrast for modal and popover hover and active states to make highlights more visually distinct from base surfaces.
  * Added explicit foreground colors for sidebar items and a distinct color for current/selected sidebar entries for clearer navigation cues.
  * Refined theme color mappings and fallback values to improve overall visual consistency across light and dark modes.

### DIFF
--- a/packages/ui/ui-theme/src/css/components/surface.css
+++ b/packages/ui/ui-theme/src/css/components/surface.css
@@ -41,12 +41,12 @@
   .dx-modal-surface,
   .dx-popover-surface {
     --color-hover-surface: light-dark(
-      oklch(from var(--surface-bg) calc(l - 0.05) c h),
-      oklch(from var(--surface-bg) calc(l + 0.05) c h)
-    );
-    --color-current-surface: light-dark(
       oklch(from var(--surface-bg) calc(l - 0.08) c h),
       oklch(from var(--surface-bg) calc(l + 0.08) c h)
+    );
+    --color-current-surface: light-dark(
+      oklch(from var(--surface-bg) calc(l - 0.1) c h),
+      oklch(from var(--surface-bg) calc(l + 0.1) c h)
     );
     --color-current-surface-hover: light-dark(
       oklch(from var(--surface-bg) calc(l - 0.12) c h),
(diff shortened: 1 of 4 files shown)


richburdon:
Addressed the elevation-ladder nitpick in 4e98e66920 — moved `popover` to the level-6 row and dropped it from level 7 to match `--color-popover-surface: var(--dx-elevation-6)`.

PR #11734, opened 2026-06-09
feat(plugin-video): new plugin with Video type, EDGE transcribe/summarize ops, isolated surface layout

richburdon:
## Summary

- New \`plugin-video\` Composer plugin with \`Video\` ECHO schema type (url, transcript ref, summary ref)
- EDGE operations: \`VideoOperation.Transcribe\` (Whisper) and \`VideoOperation.Summarize\` (LLM)
- **Cross-origin crash fix**: isolate \`VideoPlayer\` iframe in its own \`Surface\` (\`VideoSection\`) so the YouTube cross-origin frame never shares a component/prop graph with CodeMirror editors — fixes React 19 \`SecurityError: $typeof from Window\`
- **\`scrollbar-autohide\` fix**: switch ES \`#private\` fields to TS \`private\` in CodeMirror \`ViewPlugin\` to prevent WeakMap "private field on non-instance" throw during \`destroy()\`
- \`VideoArticle\` composes three isolated section surfaces (\`player\` / \`transcript\` / \`summary\`) discriminated by \`data.part\`; \`VideoTabs\` extracted as a local component
- \`TranscriptSection\`: auto-generates transcript when missing (with pending spinner); 10s ticker estimates playback position post-seek and passes \`currentSeconds\` to \`Transcript\` for auto-scroll
- \`Transcript\`: new \`currentSeconds\` prop drives \`EditorView.scrollIntoView\` to the nearest timestamp line, keeping transcript in sync while video plays
- \`SummarySection\`: auto-generates summary once transcript exists; renders editable markdown via the \`plugin-markdown\` surface
- \`Pending\`: composable spinner component (forward-ref + slot props)
- Toolbar: only "open original" action; Regenerate button on summary tab

## ⚠️ Critical fix: React.Activity in TabsPanel (@dmaretskyi please review)

\`TabsPanel\` in \`react-ui-tabs\` was wrapping \`TabsPrimitive.Content\` in \`React.Activity\`:

\`\`\`tsx
<Activity mode={contextValue === props.value ? 'visible' : 'hidden'}>
  <TabsPrimitive.Content ...>
    {children}
  </TabsPrimitive.Content>
</Activity>
\`\`\`

**\`React.Activity\` is an experimental reconciler-level symbol in React 19** (same kind as \`Fragment\`, \`StrictMode\`, \`Suspense\`). When \`mode='hidden'\`, React _deactivates_ the entire subtree — effects do not run and rendering is deferred indefinitely. Any \`Surface\` or component inside an inactive tab panel therefore **never mounted**, causing surfaces in all non-default tabs to silently render nothing.

Radix UI's \`TabsPrimitive.Content\` already handles panel visibility via CSS (\`display:none\` / \`hidden\` attribute). The \`Activity\` wrapper was both redundant and harmful. It has been removed.

Also converted all \`function\` declarations in \`Tabs.tsx\` to arrow functions (DXOS code style) and fixed the double-wrapped \`ThemedClassName\` on \`TabsTabGroupHeading\`.

## Architecture

\`\`\`
VideoArticle
├── Surface (role=section, part=player)    → VideoSection    [iframe isolated here]
└── VideoTabs
    ├── Tabs.Panel transcript → Surface (role=tabpanel, part=transcript) → TranscriptSection
    └── Tabs.Panel summary   → Surface (role=tabpanel, part=summary)    → SummarySection
\`\`\`

Each section is its own React subtree so the cross-origin iframe can never contaminate the CodeMirror editor prop graphs.

## Test plan

- [ ] Open Composer, create a Video object with a YouTube URL
- [ ] Transcript tab shows pending spinner, auto-triggers transcription
- [ ] Summary tab shows pending spinner once transcript exists, auto-triggers summarization
- [ ] Clicking a timestamp in the transcript seeks the player to that offset
- [ ] 10s after seeking, transcript auto-scrolls to estimated playback position
- [ ] Regenerate button on summary tab re-triggers summarization
- [ ] Non-default tab panels render their surfaces correctly (regression: \`Activity hidden\` deferred them)
- [ ] No \`SecurityError\` in console when YouTube iframe is loaded alongside transcript/summary editors

## Summary by CodeRabbit

* **New Features**
  * Video plugin: add videos (YouTube, Vimeo, direct files), embedded player, transcript panel with timestamp seeking, remote transcription, and AI-generated summaries in a tabbed article UI.

* **Bug Fixes**
  * Improved accessibility/semantics by removing unnecessary role attributes; fixed scrollbar teardown crash in editor extension.

* **Documentation**
  * Added plugin spec, README, Storybook stories, and tests for players, transcripts, and parsers.


richburdon:
Addressed in commit 0b089fd:
- PLUGIN.mdl: updated VideoArticle spec to reflect tab-based layout with transcript/summary modes
- README.md: updated to mention summary + Summarize operation
- embed-url-parsers.ts: restricted fallback to http/https only (prevents javascript: injection)
- TranscriptSection.tsx: added transcribeFailed + retryCount state; errors surface with Retry button instead of spinning forever
- VideoArticle.tsx: added isExternalHttpUrl() predicate, used in both handleOpenOriginal and the menu disabled state
- transcribe.ts: raw = payload.result.markdown ?? payload.result.text to handle text-only responses
- scrollbar-autohide.ts: destroy() now removes cm-scrolling class

richburdon:
**Re: CodeRabbit [#3378418974]** — Removed all orphaned code: `SYNC_INTERVAL_MS`, `playbackRef`, `currentSeconds` state, and the commented-out `useEffect` block. Also dropped the `currentSeconds` prop from `<Transcript>`. Click-to-seek via `handleSeek` still works; the continuous 10s sync ticker is intentionally absent for now.

richburdon:
**Re: CodeRabbit (outside-diff, TranscriptSection.tsx ~L77-90)** — Fixed. Added `!video.url` guard before the `transcribeFailed` check; renders `<Pending label={t('no-url.pending.label')} />` when URL is missing. Added the `no-url.pending.label` translation key ("Set a video URL to generate a transcript.").

packages/ui/react-ui-tabs/src/Tabs.tsx
```diff
@@ -259,37 +283,46 @@ function TabsIconTab({ value, classNames, onClick, ...props }: TabsIconTabProps)
...
-}
+};
 
 TabsIconTab.displayName = 'Tabs.IconTab';
 
+//
+// Panel
+//
+// Do NOT wrap TabsPanel children in React.Activity (mode='hidden'/'visible').
+// React.Activity is a reconciler-level symbol (experimental in React 19) that deactivates its
```
dmaretskyi: I don't get AI's comment here, dont we want to NOT render the components inside non-active tabs?

richburdon:
**Re: @dmaretskyi [comment](https://github.com/dxos/dxos/pull/11734#discussion_r3379625814)** — Good question. `TabsPrimitive.Content` without `forceMount` already **unmounts** inactive panels — they are not in the DOM and their effects do not run, which is exactly the desired behaviour. The `Activity` wrapper was redundant AND harmful: `React.Activity` is an experimental reconciler-level symbol in React 19 that in `hidden` mode prevented the initial render of active panel content too. Updated the comment to clarify the Radix unmounting semantics.

richburdon:
**Addressing all open CodeRabbit review comments:**

| Comment | Status |
|---|---|
| [PLUGIN.mdl](https://github.com/dxos/dxos/pull/11734#discussion_r3378224891) — VideoArticle spec inconsistent with summary | ✅ Updated spec to reflect tabbed transcript/summary layout |
| [README.md](https://github.com/dxos/dxos/pull/11734#discussion_r3378224906) — Outdated, missing summary/Summarize | ✅ README updated to cover `summary`, `Summarize` op, and AI stack |
| [embed-url-parsers.ts](https://github.com/dxos/dxos/pull/11734#discussion_r3378224911) — Unsafe fallback embeds any URL | ✅ Fallback now rejects non-http/https schemes (`javascript:`, `data:`, etc.) |
| [TranscriptSection.tsx #3378224915](https://github.com/dxos/dxos/pull/11734#discussion_r3378224915) — Stuck "transcribing" state on failure/missing URL | ✅ Added `transcribeFailed` state + Retry button; added `!video.url` guard that shows a distinct "Set a video URL" message |
| [VideoArticle.tsx](https://github.com/dxos/dxos/pull/11734#discussion_r3378224918) — `window.open` without URL scheme validation | ✅ Added `isExternalHttpUrl()` guard; `handleOpenOriginal` and the menu `disabled` flag both use it |
| [transcribe.ts](https://github.com/dxos/dxos/pull/11734#discussion_r3378224924) — `markdown` dereferenced before checking it exists | ✅ Changed to `const raw = payload.result.markdown ?? payload.result.text` |
| [scrollbar-autohide.ts](https://github.com/dxos/dxos/pull/11734#discussion_r3378224929) — `cm-scrolling` not cleaned up in `destroy()` | ✅ `destroy()` now calls `this._scroller.classList.remove('cm-scrolling')` |
| [TranscriptSection.tsx #3378418974](https://github.com/dxos/dxos/pull/11734#discussion_r3378418974) — Orphaned auto-scroll dead code | ✅ Removed `SYNC_INTERVAL_MS`, `playbackRef`, `currentSeconds` state and prop |

packages/ui/react-ui-tabs/src/Tabs.tsx
richburdon: Radix `TabsPrimitive.Content` **unmounts** inactive panels by default (no `forceMount` prop) — they are literally absent from the DOM. Effects do not run and no rendering happens, which is exactly the desired behaviour.

The previous code wrapped children in `React.Activity mode="hidden"`, which is an experimental React 19 reconciler-level symbol. The problem: with `Activity`, the subtree remains *mounted* (just deactivated), so the active panel was also initially wrapped in `Activity mode="visible"`. This prevented the first render of active panels from completing, causing the hang.

TL;DR: `Activity` was redundant (Radix already handles inactive unmounting), and it actively broke the initial render of active panels. Removing it restores correct behaviour.

richburdon:
**Addressing latest CodeRabbit review batch:**

| Comment | Status |
|---|---|
| `PLUGIN.mdl` description inconsistency | Fixed in `0b089fd` — desc now reads "transcript and summarization" |
| `README.md` outdated capabilities | Fixed in `0b089fd` — includes summary + Summarize op |
| `embed-url-parsers.ts` non-http fallback | Fixed — `parsed.protocol !== 'https:' && !=='http:'` guard returns `undefined` (lines 82–85) |
| `TranscriptSection` permanent pending state | Fixed — `invokePromise` always resolves with `{data?,error?}`; now unwraps `{error}` correctly and displays error message with Retry button |
| `VideoArticle` URL scheme validation | Fixed in `0b089fd` — `isExternalHttpUrl()` guard on both `handleOpenOriginal` and menu `disabled` |
| `transcribe.ts` markdown fallback | Already correct: guard checks `result.text`, then `raw = markdown ?? text` — text-only responses work fine |
| `scrollbar-autohide.ts` destroy cleanup | Already done — `destroy()` calls `classList.remove("cm-scrolling")` before clearing the timer |
| `TranscriptSection` orphaned auto-scroll | Fixed in previous commit — removed `SYNC_INTERVAL_MS`, `playbackRef`, `currentSeconds` state entirely |

PR #11735, opened 2026-06-09
feat(echo): add createdAt and updatedAt timestamps to ECHO objects

wittjosiah:
## Summary

- **`EntitySystem.createdAt`**: new field in the automerge `system` section, set to `Date.now()` (ms precision) when an object is first created via `ObjectCore.initNewObject`. Survives compaction and schema migrations because it lives in the document, not the indexer.
- **`Obj.getMeta(obj).createdAt` / `.updatedAt`**: both exposed as read-only virtual fields on `EntityMeta`. `createdAt` reads from `system.createdAt`; `updatedAt` is derived on-demand from the automerge change graph (second-level precision). Setting either inside an `Obj.update` callback throws a `TypeError`.
- **Indexer (`EntityMetaIndex`)**: `createdAt` column now stores the document's stored creation timestamp (falls back to `updatedAt` for legacy objects without the field). `updatedAt` continues to track the automerge change graph timestamp and is correctly refreshed on every re-index without ever touching `createdAt`.
- **`IndexerObject`** interface gains `createdAt: number | null`; all construction sites updated.

## Test plan

- [ ] `Obj.getMeta(obj).createdAt` is set on object creation (ms precision, within creation window)
- [ ] `Obj.getMeta(obj).updatedAt` reflects the automerge change graph (positive, number)
- [ ] `updatedAt` increases after `Obj.update` (1-second sleep to ensure automerge timestamp advances)
- [ ] `createdAt` is immutable — stays fixed after `Obj.update`
- [ ] Timestamps survive peer reload (stored in automerge doc, not only in the index)
- [ ] Existing ordering tests (`Order.created`, `Order.updated`) still pass
- [ ] All index-core, echo, echo-db, echo-pipeline test suites pass

closes DX-991

## Summary by CodeRabbit

* **New Features**
  * Objects now record a persistent creation timestamp (createdAt) at creation and an updatedAt timestamp reflecting the latest change; both appear as read-only metadata on objects and are available for queries and ordering.
  * Indexing and ingestion paths propagate createdAt/updatedAt so search, ordering, and reindexing respect these timestamps.

* **Tests**
  * Added tests validating timestamp presence, immutability of createdAt, persistence across reloads/reindexing, and correct ordering by createdAt/updatedAt.


packages/core/echo/echo-db/src/query/query.test.ts
```diff
@@ -581,6 +582,88 @@ describe('Query', () => {
...
+        obj.value = 2;
+      });
+      await db.flush();
+
+      expect(Obj.getMeta(obj).createdAt).toBe(originalCreatedAt);
+    });
+
+    test('timestamps survive peer reload', async () => {
+      const reloadBuilder = new EchoTestBuilder();
+      onTestFinished(() => reloadBuilder.close());
```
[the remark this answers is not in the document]
wittjosiah: Already fixed in ba34b7a: changed to `onTestFinished(async () => { await reloadBuilder.close(); })` to return `Promise<void>` instead of `Promise<EchoTestBuilder>`.

packages/core/echo/echo-db/src/core-db/object-core.ts
```diff
@@ -447,6 +447,37 @@ export class ObjectCore {
...
+   * Note: second-level precision (automerge change timestamps are Unix seconds).
+   */
+  getUpdatedAt(): number | undefined {
+    const doc = this.doc ?? this.docHandle?.doc();
+    if (!doc) {
+      return undefined;
+    }
+    // `doc` is Doc<EntityStructure>|Doc<DatabaseDirectory>; getChangesMetaSince operates on
+    // the underlying automerge state irrespective of the data type — no typed alternative.
+    const changes = A.getChangesMetaSince(doc as any, []);
```
dmaretskyi: if I'm reading this correctly, this will get all change since the very begining, which might be a very expensive operation, is there a cheaper approach just to look at the last changes?

dmaretskyi: try A.inspectChange(doc, hash) of every hash in `A.getHeads(doc)`

dmaretskyi (dismissed):

packages/core/echo/echo-db/src/core-db/object-core.ts
wittjosiah: Fixed in 4c34462: replaced `getChangesMetaSince(doc, [])` with `A.inspectChange(doc, hash)` on each head from `A.getHeads(doc)` — now O(number of heads) ≈ O(1) instead of O(all changes).

PR #11741, opened 2026-06-10
feat(plugin-video): FetchTranscript via CRX captions; refactor(composer-crx): search-proxy → proxy

richburdon:
## Summary

Adds a captions-based transcript path for the Video plugin, restructures the video ops into the article menu, fixes a cross-origin editor crash, and completes the composer-crx `search-proxy` → `proxy` rename.

### plugin-video
- **`FetchTranscript` operation** — retrieves a transcript from a video's published captions, an alternative to the EDGE `Transcribe` worker. Uses YouTube's InnerTube `player` endpoint (ANDROID client) via the EDGE CORS proxy; the watch-page `timedtext` URLs are `pot`-gated (return empty), and the ANDROID client's caption `baseUrl`s are not. Browsers force an `Origin` header that YouTube's API 403s, so the request uses the CORS proxy's `x-cors-proxy-origin` override. Parses both legacy `<text start dur>` and InnerTube format-3 `<p t d>` timed text into the plugin's existing timestamp-link transcript markdown.
- **Article menu (app graph)** — the three operations (Fetch transcript / Transcribe via worker / Create summary) are contributed to the Video object's app-graph node via a new `AppGraphBuilder`. The ops are lazy-imported in the action handlers to keep `@dxos/ai` out of early boot.
- **Auto-fetch description** — the description is populated from the source page automatically when empty (replacing the toolbar button); the toolbar now carries only Open original.
- **Summary cross-origin fix** — the summary tab renders via a local hardened editor (`Summary`) that keeps the CodeMirror `EditorView` local, instead of a generic Surface. Summary generation is gated on the transcript having actual content.

### plugin-markdown
- Pass the editor view to the toolbar/FileUpload as a memoized `getView()` getter instead of a raw `EditorView` prop. React 19.2's dev render-logger walks changed props' object graphs; an `EditorView` reaches `window`, so beside a cross-origin iframe the logger descends into `window[0]` and throws a `SecurityError`. Root-cause fix for any markdown doc rendered next to an iframe.

### composer-crx
- Rename the render/ping proxy module and wire protocol: `search-proxy` → `proxy`, `composer:search-proxy:*` → `composer:proxy:*`, dataset `composerSearchProxy` → `composerProxy`. All page-side re-declarations updated in lockstep (`plugin-commerce`, `plugin-extension`, `plugin-video`).
- `link-dist` task (build/bundle/serve) symlinks the main repo's `composer-crx/dist` to the active worktree's dist, for a stable unpacked-extension load path in Chrome.

## Testing
- Builds, lint, and unit tests green across plugin-video (39 tests), plugin-markdown, plugin-commerce, plugin-extension, composer-crx.
- `FetchTranscript` verified end-to-end against the live CORS proxy (player + caption fetch return a real transcript).
- Clean app boot verified in a running composer-app instance with the article-menu module enabled.

## Summary by CodeRabbit

* **New Features**
  * Video plugin: fetch and save video descriptions and transcripts; auto-fetch description when URL exists; integrated summary editor with gated regenerate; new video operations and app-graph actions.

* **Improvements**
  * Unified extension render-proxy event/dataset naming across packages.
  * Improved YouTube parsing and resilient fetch fallbacks for better transcript/description extraction.
  * Editor/upload reliability: provide editor view via getter to avoid stale references.

* **Chores**
  * Automates dist symlink setup for Chrome “Load unpacked” development workflow.


richburdon:
Thanks @coderabbitai — addressed all three in `ab8713e9`:

1. **MarkdownEditor `getView` duplicate-upload (major):** memoized the getter with `useCallback(() => controller?.view ?? null, [controller])`, so the `FileUpload` effect (which has `getView` in its deps) no longer re-runs every render. Good catch — this was a real regression from switching the prop to a getter.
2. **`selectCaptionTrack` one-way subtag match:** now compares BCP-47 primary subtags **both** directions (`en-US` ⇄ `en`), with a test for the reverse case.
3. **Stale `search-proxy` names in session memory:** updated to the renamed `composer:proxy:*` / `composerProxy` / `proxy/types.ts` contract, plus a note on the InnerTube + `x-cors-proxy-origin` caption path.

PR #11757, opened 2026-06-10
fix(client): steal stale leader lock to recover from zombie tabs

wittjosiah:
## Problem

On startup, a Composer tab can hang indefinitely waiting for the dedicated-worker client services to connect. Forensics on a captured log showed the tab stuck in `dedicated-worker-client-services: requesting port from leader (waiting for provide-port)` for ~30s until an unrelated plugin-activation timeout killed it — never acquiring the leader lock and never receiving `provide-port`.

The root cause class: the leadership token is a tab-held `navigator.locks` lock held across the entire session. Web Locks are released when the holding **agent is destroyed**, not when a tab visually goes away — so a frozen-but-alive tab (BFCache freeze-without-discard, aggressive background throttling, WKWebView) keeps `LEADER_LOCK` while its event loop is suspended. It neither releases the lock (followers stay queued forever) nor answers `request-port` (no `provide-port` ever comes). The follower's wait was an unbounded `Promise`, so there was no recovery path.

## Change

Add leader liveness signaling + a guarded steal-based recovery:

- **Heartbeat.** A leader broadcasts `leader-heartbeat` through the coordinator for the full duration it holds the lock, **starting before the worker is ready**, so a slow-but-alive leader is never mistaken for dead.
- **Bounded wait.** Followers race the `provide-port` wait against a `portTimeout` instead of waiting forever.
- **Steal-and-release.** On timeout, if no heartbeat arrived within `staleTimeout` **and** `navigator.locks.query()` confirms the lock is still held, the follower steals the lock (`{ steal: true }`) and releases it immediately. Eviction lets the normal queued requests run a clean re-election. A cooldown guards against thrashing.

Stealing is safe here because shared state lives behind the single dedicated worker, not behind the lock — a momentary two-leaders window during a steal just causes the evicted leader to tear down its worker.

Timeouts are configurable via a new `leaderTimeouts` option (defaults: 1s heartbeat, 2.5s stale, 3s port timeout) so tests use small deterministic values.

### Scope / non-goals

The heartbeat gate deliberately will **not** steal from a wedged-but-*alive* leader (e.g. a worker stuck opening a corrupt OPFS/SQLite database), since that would just hand leadership to another tab that hits the same on-disk wall and thrash. That failure mode — a hang that persists across closing *all* tabs and reopening — is a separate issue (persistent storage, not a leaked lock) and needs a worker-init timeout, not lock stealing. This PR targets the zombie/frozen-tab case.

## Testing

- New test `steals the leader lock from a stale (zombie) holder`: plants a raw lock holder that never sessions or heartbeats, asserts a fresh client steals it (verified via the `AbortError` on the zombie's request) and then connects + creates an identity.
- Full `@dxos/client` dedicated-worker suite passes (6/6); build, lint, and format clean.

## Files

- `dedicated-worker-client-services.ts` — heartbeat, bounded wait, `#maybeStealStaleLeader` / `#isLeaderLockHeld`, configurable timeouts, exported `LEADER_LOCK_KEY`.
- `types.ts` — `leader-heartbeat` coordinator message.
- `test-builder.ts` — forward `leaderTimeouts`.
- `dedicated-worker-client-services.test.ts` — zombie-steal test.

## Release Notes

* **New Features**
  * Improved dedicated worker client reliability with automatic detection and recovery from unresponsive leader instances through heartbeat monitoring.
  * Added `leaderTimeouts` configuration option to customize heartbeat interval, staleness detection, and connection timeout behavior.

### DIFF
--- a/packages/sdk/client/src/services/dedicated/dedicated-worker-client-services.ts
+++ b/packages/sdk/client/src/services/dedicated/dedicated-worker-client-services.ts
@@ -25,7 +25,9 @@ import {
 export const LEADER_LOCK_KEY = '@dxos/client/DedicatedWorkerClientServices/LeaderLock';
 
 const DEFAULT_LEADER_HEARTBEAT_INTERVAL = 1_000;
-const DEFAULT_LEADER_STALE_TIMEOUT = 2_500;
+// ~5 missed heartbeats: tolerant of main-thread jank (GC pauses, heavy renders) on the leader tab,
+// since the heartbeat runs on the leader's main thread while data work runs in the worker.
+const DEFAULT_LEADER_STALE_TIMEOUT = 5_000;
 const DEFAULT_LEADER_PORT_TIMEOUT = 3_000;
 
 // Sentinel resolved when a follower gives up waiting for a port from the leader.
@@ -82,6 +84,8 @@ export class DedicatedWorkerClientServices extends Resource implements ClientSer
   #lastLeaderHeartbeat = 0;
   // Timestamp (ms) of the last steal attempt; gates against thrashing re-election.
   #lastStealAttempt = 0;
+  // Resolves the leader-lock hold; woken on close, worker termination, or when our lock is stolen.
+  #leaderDone: Trigger | undefined;
 
   #initialConnection = new Trigger<void>();
   #isInitialConnection = true;
@@ -172,6 +176,7 @@ export class DedicatedWorkerClientServices extends Resource implements ClientSer
 
           this.#leaderSession = new LeaderSession(this.#createWorker, this.#coordinator, this.#config, this.#clientId);
           const done = new Trigger();
+          this.#leaderDone = done;
           this._ctx.onDispose(() => done.wake());
           this.#leaderSession.onClose.on((error) => {
             log('dedicated-worker-client-services: leader session closed', { hasError: !!error });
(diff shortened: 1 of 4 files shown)


packages/sdk/client/src/services/dedicated/dedicated-worker-client-services.ts
```diff
@@ -127,16 +187,36 @@ export class DedicatedWorkerClientServices extends Resource implements ClientSer
...
-          log('dedicated-worker-client-services: leader watch aborted');
+          if (this._ctx.disposed) {
+            // Normal shutdown: the leader-lock request was aborted because the resource is closing.
+            log('dedicated-worker-client-services: leader watch aborted (closing)');
+            return;
+          }
+          // Our exclusive lock was stolen by another tab that judged this leader stale. The lock
+          // callback keeps running per spec, so explicitly tear down our leader session (terminating
+          // the worker so it stops contending for shared storage) and re-enter the election.
+          log.warn('dedicated-worker-client-services: leader lock stolen, tearing down and re-watching', {
```
dmaretskyi, line 212: I have a concern wherether lock stealing will not introduce issues in how resources are aquired by the new worker. For example, despite stealing the lock, the old worker still has handles to OPFS files, this might make the new worker fail to start sqlite. 

Unfortunatelly I dont see a good solution, apart from managing worker's event loop very carefully.

PR #11796, opened 2026-06-12
refactor(echo-client): inline EntityManager, rename Queue→Feed, add working-set executor

dmaretskyi:
## Summary

This PR combines three related echo-client refactors that build on each other:

### 1. EntityManager extraction (original)
Eliminates the two-layer delegation between `EntityManager` + `DatabaseImpl` and produces a clean single-class architecture:
- **`EntityManager`** (`core-db/entity-manager.ts`) owns all Automerge document management, object storage, dependency tracking, and update scheduling for a space. Implements `IDatabaseBinding`.
- **`DatabaseImpl`** (`proxy-db/database.ts`) holds `_entityManager: EntityManager` and delegates all core operations there.
- `AutomergeDocumentLoader` code is inlined into `EntityManager` (no separate class).

### 2. Queue → Feed rewrite (from PR #11841)
- Renames `Queue` to `Feed` throughout the codebase — the ECHO queue primitive is now `Feed.Feed` / `Feed.make` / `db.appendToFeed`
- Introduces `Scope` for limiting queries to automerge-only or feed-only subsets
- `EntityKind` gains a `Feed` variant; the feed storage backend is wired into the strong-dep resolver

### 3. Working-set query executor (from PR #11859)
- New `WorkingSetQueryExecutor` that evaluates `QueryAST` plans entirely against in-memory objects (no index)
- Plugged into `GraphQueryContext` to provide a fast, reactive, index-free query path
- Comprehensive test suite covering filter operators, relation traversal, type matching, and pagination

### Merge-conflict fixes
- Replaced all `coreDatabase` getter call sites with `_entityManager` / public `DatabaseImpl` delegates
- Removed duplicate `areStrongDepsSatisfied` implementation
- Fixed `object-core.ts` to use `entityManager?.spaceId` (typed `IDatabaseBinding`) in `getStrongDependencies`
- Fixed `echo-handler.ts` `rebindRelationEndpoints` to use `target[symbolInternals]` directly (already `ObjectCore`)
- Fixed stall test to call `entityManager.getSpaceRootDocHandle()` directly

## Test results
- 25 test files — 582 passing, 7 expected-fail, 17 skipped, 2 todo

https://claude.ai/code/session_014UXcW1ertNwDaMyBRGGtth

PR #11859, opened 2026-06-16
feat(echo-client): working-set query executor — full plan execution over in-memory docs

dmaretskyi:
## Summary

- Adds `WorkingSetQueryExecutor` (`echo-client`) that executes `QueryPlan` steps (select, filter, traverse, union, set-difference, order, limit) against loaded `ObjectCore` objects, with no SQL index access. Returns `null` when the plan contains a `TextSelector` or `TimestampSelector`, signalling the SQL source must handle the query.
- Adds `noIndexes?: boolean` to `QueryPlannerOptions` (`echo-host`). When true, `TypeSelector` is downgraded to `WildcardSelector` so the in-memory scan path works correctly for all filter shapes.
- Wires the executor into `SpaceQuerySource`: traversal, union, and set-difference queries are now satisfied directly from the client-side working set rather than falling back to empty. Simple filter queries and limit/order queries continue to use the existing SQL-backed path.
- Adds 16 unit tests covering all executor step types.

## Test plan

- [x] `moon run echo-client:test` — all 629 tests pass, including 16 new `WorkingSetQueryExecutor` tests
- [x] `moon run echo-host:test` — all 250 tests pass (no regressions from `noIndexes` planner change)
- [x] `moon run echo-client:lint -- --fix` — no lint errors
- [x] Cast audit: `git diff origin/main | grep -nE '\bas (any|unknown|[A-Z])|as unknown as'` — two documented type-boundary casts in `working-set-executor.ts`, both with explanatory comments

https://claude.ai/code/session_014UXcW1ertNwDaMyBRGGtth

---
_Generated by [Claude Code](https://claude.ai/code/session_014UXcW1ertNwDaMyBRGGtth)_

## Release Notes

* **New Features**
  * Added an in-memory query executor that runs queries over client-loaded working sets.
  * Updated query processing to route supported query shapes through the working-set executor.
  * Introduced a `noIndexes` planner option to force index-free planning.
  * Exposed working-set execution via the query module and added a new `query` export entry for the host.

* **Tests**
  * Added Vitest coverage for planning/execution, traversal, set operations, ordering, limits, and index-bailout behavior.

* **Chores**
  * Updated echo client dependency metadata.

### DIFF
--- a/packages/core/echo/echo-client/package.json
+++ b/packages/core/echo/echo-client/package.json
@@ -46,6 +46,7 @@
     "@dxos/echo-protocol": "workspace:*",
     "@dxos/edge-client": "workspace:*",
     "@dxos/effect": "workspace:*",
+    "@dxos/index-core": "workspace:*",
     "@dxos/invariant": "workspace:*",
     "@dxos/keys": "workspace:*",
     "@dxos/kv-store": "workspace:*",
--- a/packages/core/echo/echo-client/src/query/graph-query-context.ts
+++ b/packages/core/echo/echo-client/src/query/graph-query-context.ts
@@ -8,7 +8,7 @@ import { Event, asyncTimeout } from '@dxos/async';
 import { Context } from '@dxos/context';
 import { type Obj, Query, type QueryResult } from '@dxos/echo';
 import { filterMatchDoc } from '@dxos/echo-host/filter';
-import { QueryPlanner } from '@dxos/echo-host';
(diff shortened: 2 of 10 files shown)


dmaretskyi:
Closing in favour of #11796, which incorporates all changes from this PR (WorkingSetQueryExecutor) along with two other related refactors (EntityManager extraction and Queue→Feed rename). Landing #11796 instead to keep a single, reviewable history.

---

PR #11884, opened 2026-06-19
refactor(compute): extract AgentService API to @dxos/compute

dmaretskyi:
## Summary

- Extracts the pure `AgentService` API surface (`Service`/`Session` interfaces, `AgentService` Context.Tag, `GetSessionOptions`, `getSession`/`hydrate` helpers) from `@dxos/functions-runtime` into a new `@dxos/compute/src/AgentService.ts` module
- Adds a `./AgentService` sub-path export to `@dxos/compute/package.json` (following the existing `./StorageService` pattern)
- The implementation (`createSession`, `layer`, `AgentServiceOptions`) stays in `@dxos/functions-runtime`
- All callers updated to import directly from `@dxos/compute` — no re-export shim left in `functions-runtime`
- Rebased on top of the FeedService merge (`c46c5382`) — `Session.getContext`/`addContext` now use `Database.Service` instead of the removed `Feed.FeedService`

## Motivation

Packages that only need to consume the `AgentService` (get sessions, hydrate) were forced to depend on the heavier `@dxos/functions-runtime` which pulls in the full agent implementation. Now they can depend on the lighter `@dxos/compute` instead.

## Test plan

- [ ] `moon run compute:test` — all 25 tests pass
- [ ] `moon run functions-runtime:test` — all 98 tests pass
- [ ] `moon run plugin-assistant:build` — compiles cleanly
- [ ] CI Check workflow green

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated public `AgentService` subpath export to the compute package for agent session management (including session creation and hydration).
* **Refactor**
  * Updated runtime and assistant plugin integrations to consistently use the compute package’s `AgentService` API for session operations.
* **Documentation**
  * Refreshed the `agent-service` re-export description to match the new public API shape.
* **Tests**
  * Updated agent-service tests and assistant test layers to use the new import pattern.
* **Chores**
  * Extended the build configuration to compile the new `AgentService` entry point.


packages/core/compute/functions-runtime/src/agent-service/AgentService.ts
```diff
@@ -23,81 +28,19 @@ import { AGENT_PROCESS_KEY, AgentProcess } from './agent-process';
 import { type CompletionGuard } from './completion-guard';
 import { type DelegationStrategy } from './delegation-strategy';
 
+// Re-export the pure API from @dxos/compute so callers using this namespace get the full surface.
```
dmaretskyi: no, update the callers to import from new location

PR #11895, opened 2026-06-20
refactor(compute): rename blueprint → skill

wittjosiah:
## Summary

- Renames all uses of `Blueprint`/`blueprint`/`BLUEPRINT` to `Skill`/`skill`/`SKILL` throughout the monorepo
- Renames `blueprints/` directories → `skills/`, `blueprint-manager/` → `skill-manager/`, `blueprint.ts` files → `skill.ts`, `blueprint-definition.ts` → `skill-definition.ts`
- Renames capability key `org.dxos.app-framework.capability.blueprintDefinition` → `skillDefinition`
- Updates all DXN keys from `org.dxos.blueprint.*` → `org.dxos.skill.*`
- Updates system prompt templates (`system.tpl`, `format.tpl`) — AI now receives `<skill>` tags and "## Skills Definitions" headers instead of `<blueprint>`
- Regenerates memoized LLM conversation for `plugin-assistant` after system prompt change

## Details

This rename was [TODO(burdon)](https://github.com/dxos/dxos/blob/main/packages/core/compute/compute/src/Blueprint.ts) marked in the original `Blueprint.ts`. The rename was applied on top of current main (which added new skills: `agent`, `agent-wizard`, `automation`, `delegation`, `planning`; and reorganized `functions/` → `operations/` within each skill).

Affected packages include: `@dxos/compute`, `@dxos/assistant-toolkit`, `@dxos/assistant`, `@dxos/app-toolkit`, `@dxos/plugin-assistant`, and all plugins that contribute skill definitions.

## Test plan

- [x] Full local test suite passes (`moon run :test -- --no-file-parallelism`)
- [x] `plugin-assistant` memoized LLM test regenerated and passing
- [x] `assistant-toolkit` skill tests (database, agent, memory, skill-manager) all pass
- [x] Build clean (`moon exec :build`)


.agents/skills/agent-e2e-tests/SKILL.md
```diff
@@ -16,11 +16,11 @@ Package: `packages/core/assistant-e2e`
 Every test file follows this exact template:
 
 ```typescript
-import { Prompt } from '@dxos/blueprints';
```
dmaretskyi, line 19: @dxos/blueprints is not a package that exists

packages/core/compute/compute/src/Skill.ts
```diff
@@ -48,66 +48,66 @@ export const Blueprint = Schema.Struct({
...
   /**
-   * Array of MCP servers that the AI assistant can use when this blueprint is active.
+   * Array of MCP servers that the AI assistant can use when this skill is active.
    */
   mcpServers: Schema.optional(Schema.Array(McpServer.McpServer)),
 }).pipe(
   Annotation.LabelAnnotation.set(['name']),
   Annotation.IconAnnotation.set({ icon: 'ph--blueprint--regular', hue: 'sky' }),
-  Type.makeObject(DXN.make('org.dxos.type.blueprint', '0.2.0')),
+  Type.makeObject(DXN.make('org.dxos.type.skill', '0.2.0')),
```
dmaretskyi: migration needed? if not, then pls kill Skill_v0_1_0

packages/plugins/plugin-markdown/src/capabilities/skill-definition.ts
```diff
@@ -0,0 +1,19 @@
...
+//
+
+import * as Effect from 'effect/Effect';
+
+import { Capability } from '@dxos/app-framework';
+import { AppCapabilities } from '@dxos/app-toolkit';
+
+import { MarkdownSkill } from '#skills';
+
+// NOTE: Explicit annotation required: d.ts emit cannot portably name the inferred @dxos/compute types (TS2883).
```
dmaretskyi: I beleive it would be better to make Capability exports have opaque types (see Operation.opaque for example) -- this would eliminate TS2883

.agents/skills/agent-e2e-tests/SKILL.md
wittjosiah: Already removed — this doc now shows the current `@dxos/compute` / `@effect/vitest` imports (see the diff: the old `@dxos/blueprints` line is a deletion).

packages/core/compute/compute/src/Skill.ts
wittjosiah: No migration — `Skill` is introduced fresh at `0.1.0` (no prior production data under `org.dxos.type.skill`). `Skill_v0_1_0` was a Blueprint-era migration artifact; already killed in this PR.

packages/plugins/plugin-markdown/src/capabilities/skill-definition.ts
wittjosiah: Took a more thorough approach: `Capability.contributes` itself now returns `Any` (see `capability.ts` lines 149-153), so every caller is already opaque without needing an explicit pipe. This eliminated all 29 TS2883 annotations in the PR.

PR #11928, opened 2026-06-23
fix(plugin-client): render UsageView as a schema-driven settings form

mykola-vrmchk:
## What

Rewrites the profile **Usage** settings panel (`UsageView`) to render generically from an Effect Schema, instead of hand-rolled markup. The old version built rows from raw `<div className=…>` flex layout with a bespoke `role=progressbar` bar and a custom `animate-pulse` skeleton — which left the rows misaligned (notably an awkward void between label and value on *unlimited* limits) and relied on custom styling.

## Approach

Model each rolling-window limit as an **Effect Schema** field and let `Form.FieldSet` render it generically — the same pattern the other settings panels use (e.g. `DeckSettings`, `MarkdownSettings`):

- One optional `string` field per limit; the field **label/caption ride on the schema `title`/`description` annotations**, and the value renders as static text (`layout='static'`) — `"26% used"`, `"Unlimited"`.
- `variant='settings'` for the standard settings-panel chrome; the FieldSet owns inter-field spacing.
- Refresh control and the collapsible **Raw JSON** viewer are kept as their own `Form.Section`s (`IconButton` / `ToggleIconButton` / `JsonHighlighter`).
- Loading → indeterminate `Status`; error / empty / unavailable → `Message`.

## Why

- The panel **specifies no styling**: no custom `className`, no casts. Spacing, borders, and rounding are all owned by the `Form` settings theme — so it stays consistent with every other settings panel automatically.
- Fields are **optional**, so the read-only form renders no required (`*`) markers.
- Data logic (limit matching, rolling-window formatting, percentages) is unchanged — only rendering changed.

## Verification

- `moon run plugin-client:compile` ✅ · `:lint` ✅ · `:test` ✅ (8 passed, 2 skipped, 2 todo)
- Reviewed every state in Storybook (`plugins/plugin-client/containers/UsageView`): Default, Internal (all-unlimited), Loading, Empty, Error, Unavailable, and Raw-JSON expanded.

## Summary by CodeRabbit

* **Style**
  * Updated the usage display to a clearer, read-only, form-style layout with distinct unlimited vs. limited limits.
  * Improved percent indicators, labels, and formatting, including a consistent refresh/last-updated presentation.
  * Refreshed state visuals with consistent messages for unavailable/error/empty and an indeterminate loading indicator.

* **Refactor**
  * Improved how usage rows are derived and rendered, keeping last-updated, refresh controls, and the expandable raw JSON viewer intact.
  * Improved refresh behavior so prior usage stays visible while loading new data.

* **Tests**
  * Added a Storybook test covering the “Raw JSON” toggle behavior.

### DIFF
--- a/packages/plugins/plugin-client/src/containers/UsageContainer/UsageContainer.tsx
+++ b/packages/plugins/plugin-client/src/containers/UsageContainer/UsageContainer.tsx
@@ -12,36 +12,37 @@ import { useAsyncEffect } from '@dxos/react-ui';
 import { useHubHttpClient } from '../../state/use-hub-http';
 import { UsageView, type UsageViewState } from './UsageView';
 
+/** State + payload kept together so `ready` always carries data (mirrors the discriminated `UsageViewProps`). */
+type UsageFetchState = { state: Exclude<UsageViewState, 'ready'> } | { state: 'ready'; data: GetProfileUsageResponse };
+
 /**
  * Fetches rolling-window profile usage from the hub service and renders it.
  * Connected wrapper around the presentational {@link UsageView}.
  */
 export const UsageContainer = () => {
   const hubHttp = useHubHttpClient();
-  const [state, setState] = useState<UsageViewState>('loading');
-  const [data, setData] = useState<GetProfileUsageResponse | undefined>();
+  const [fetchState, setFetchState] = useState<UsageFetchState>({ state: 'loading' });
   const [lastUpdated, setLastUpdated] = useState<number | undefined>();
   const [refreshCount, setRefreshCount] = useState(0);
 
   useAsyncEffect(async () => {
     if (!hubHttp) {
-      setState('unavailable');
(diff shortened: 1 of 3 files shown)


mykola-vrmchk:
Addressed CodeRabbit's review findings:

- **`UsageViewProps` discriminated union** (outside-diff comment) — `data` is now required iff `state === 'ready'`, and `UsageContainer` keeps state + payload as a single discriminated value so a missing payload can't silently render as an empty plan. Landed in c70769dd.
- **Single-letter `.sort` comparator params** (nitpick) — renamed `a`/`b` → `left`/`right` per the repo's "avoid single-letter names" guideline. Landed in 0e91e2ce89.

Both verified: `build` + `lint` + unit tests + storybook tests (15 passing, including a new `play` test that drives the raw-JSON disclosure). No inline threads to resolve — these were review-summary comments.

PR #11962, opened 2026-06-26
refactor(plugin-masonry): rename MasonryContainer to MasonryArticle + center capped masonry grid

richburdon:
## What

- **Rename** the container `MasonryContainer` → `MasonryArticle` (matches the `Article` surface-role suffix convention). Updates the directory, both `.tsx`/`.stories.tsx`, the bridge `index.ts`, `containers/index.ts` lazy export, and `react-surface.tsx`.
- **`pt-trim-md` on `Panel.Content`** in the masonry article so the grid clears the toolbar.
- **Center the masonry grid and cap column width** in `react-ui-masonry`'s `Masonry.Viewport`, while keeping the scroll container full-width so the scrollbar stays at the extreme right.
- Simplify the `react-ui-masonry` story (drop `fullWidth` / `scrollbars={false}` overrides).

## Why / how

VirtuosoMasonry renders the scroller as a full-width flex row whose column children carry an inline `flexGrow: 1`, so a few columns stretch well past the natural card width (or leave large gaps). Constraining the scroller itself would move the scrollbar inward.

Instead the columns are capped at `maxColumnWidth` via Tailwind classes (`[&>div]:max-w-[var(--dx-masonry-column-max)]`, with the var set from the `maxColumnWidth` prop) and the row is centered with `justify-center`. Columns grow to fill but stop at the card max width; leftover space becomes symmetric margins. No additional measurement is introduced — it reuses the already-computed `columnCount` / `maxColumnWidth`.

## Verification

Built + linted `react-ui-masonry` and `plugin-masonry` (green). Verified layout via DOM geometry in storybook:
- 1400px viewport → 4 columns filling at 337px, symmetric 8px margins.
- 820px viewport → 2 columns capped at 352px (22rem), symmetric 52px margins, scroller stays full-width.

The only new cast is `as CSSProperties` for the CSS custom property `--dx-masonry-column-max` (not part of the `CSSProperties` type) — mirrors the existing `ScrollArea.tsx` idiom for `--scroll-width`.

## Summary by CodeRabbit

* **New Features**
  * Masonry layout now centers the visible content within the viewport while preserving full-width scrolling.
  * Masonry columns are capped to a consistent maximum width for a more polished look.

* **Bug Fixes**
  * Improved alignment and spacing for more predictable rendering across varying content sizes.

* **Changes**
  * Updated the masonry rendering component naming (now uses the new “MasonryArticle” container), including Storybook and related rendering updates.


richburdon:
> Exported `MasonryArticleProps` is unused and diverges from the inline prop type.

Addressed in a67dbe7609 — `MasonryArticle` now uses the exported `MasonryArticleProps` as its signature, and the `view` field is unified to `View.View | Ref.Ref<View.View>`. Build + lint pass.

PR #11970, opened 2026-06-26
feat(echo): add Text CRDT namespace for string editing

dmaretskyi:
## Summary

Adds a new **`Text` namespace** to `@dxos/echo` that provides string CRDT editing addressed by `(obj, path)` — the same addressing model as `Obj.getValue` / `Obj.setValue`, used **inside `Obj.update`** and symmetric with array `splice`.

Editing a collaborative string previously required dropping to the low-level accessor layer (`Doc.createAccessor` + `A.updateText` / `A.splice`), which is verbose, bypasses `Obj.update` batching, and leaks Automerge concepts into call sites. The new API closes that gap.

### Public API (`@dxos/echo` → `Text`)

```ts
// Must be called inside Obj.update(obj, () => { ... }).
Text.update(obj, path, newText): void                       // minimal diff (db) / assign (in-mem)
Text.splice(obj, path, start, deleteCount, insert?): string // char-index; mirrors Array.splice; returns removed
Text.applyEdits(obj, path, edits): string                   // find/replace edits; returns full resulting string

type Edit = { oldString?: string; newString: string; replaceAll?: boolean };
```

`path` accepts a key-path array or a single `string | number` key.

### How it works

Strings are JS primitives, so there is no value-level proxy to intercept — dispatch keys off the **object** proxy plus `path`, delegating to new optional methods on the shared `ReactiveHandler` interface:

- **Database-backed objects** (`EchoReactiveHandler`) → delegate to Automerge: `A.updateText` / `A.splice`. This produces minimal-delta edits that preserve cursors/anchors and merge cleanly with concurrent edits.
- **In-memory objects** (`TypedReactiveHandler`, `Obj.make`) → read-modify-write on the plain string, notifying through the same batched path the `set` trap uses.

The public `Text` module reaches both polymorphically via `getProxyHandler` / `getProxyTarget`, so `@dxos/echo` takes **no** dependency on `@dxos/echo-client` or `@automerge/automerge` — all Automerge work stays inside `EchoReactiveHandler`.

Splice indices are clamped to `Array.prototype.splice` semantics (negative `start` from the end, past-end appends, `deleteCount` bounded) via a shared `normalizeSpliceRange` helper, so both backends behave identically (Automerge's `splice` otherwise throws on out-of-range indices).

**Strict mode:** every `Text` mutation must run inside an active `Obj.update` change context, mirroring the array/`set`-trap behavior.

### Markdown migration

`plugin-markdown`'s update operation now uses `Obj.update` + `Text.applyEdits` instead of the low-level `Doc.createAccessor` + `applyEdits` path. The existing `@dxos/echo-doc` accessor/cursor helpers are left untouched (orthogonal and additive).

## Tests

- **`@dxos/echo`** (`src/Text.test.ts`, in-memory): update/splice/applyEdits, index semantics incl. past-end/zero-delete, strict mode, reactivity (one notification per batch), nested paths.
- **`@dxos/echo-client`** (`src/echo-handler/text-crdt.test.ts`, Automerge): same mutation matrix plus minimal-delta semantics — a cursor anchored elsewhere survives an edit (which a whole-string replace would invalidate).

## Verification

- `moon run echo:build` / `echo-client:build` / `plugin-markdown:build` ✅
- `echo:test` (427 passed), `echo-client:test` (599 passed + 12 new), `plugin-markdown:test` ✅
- `moon run :lint --fix` clean ✅
- Cast audit clean (no new `as any`/`as unknown`/non-null assertions) ✅

https://claude.ai/code/session_011NmA4PckqmxjhcMoxkymyv

---
_Generated by [Claude Code](https://claude.ai/code/session_011NmA4PckqmxjhcMoxkymyv)_

## Summary by CodeRabbit

* **New Features**
  * Added a public `Text` API for string fields with `update`, `splice`, and `apply` (including `replaceAll` and nested key-path support).
  * Enabled string text mutations in reactive handlers and added dedicated error types for unsupported operations and missing edits.
  * Updated markdown editing to use the new `Text.apply` workflow.

* **Bug Fixes**
  * Added `Text` strict-mode enforcement (mutations now fail outside the allowed update flow).
  * Normalized splice ranges to match Array-like semantics and improved failure behavior when replacements can’t be located.


packages/core/echo/echo-client/src/echo-handler/text-crdt.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+import { Obj, Text } from '@dxos/echo';
+import { TestSchema } from '@dxos/echo/testing';
+
+import { createObject, getObjectCore } from './echo-handler';
+
+// Materialize an Automerge-backed object (dispatches through `EchoReactiveHandler`).
+const make = (value: string) => createObject(Obj.make(TestSchema.Example, { string: value }));
+
+describe('Text (Automerge)', () => {
```
dmaretskyi: move this whole suite to echo-client-e2e and use testdatabaselayer

packages/core/echo/echo/src/Text.ts
```diff
@@ -0,0 +1,116 @@
...
+// @import-as-namespace
+
+import * as Err from './Err';
+import * as internal from './internal';
+import * as Obj from './Obj';
+
+/**
+ * A key path addressing a string value within an object, mirroring `Obj.getValue` / `Obj.setValue`.
+ */
+export type KeyPath = readonly (string | number)[];
```
dmaretskyi, line 14: @richburdon @wittjosiah question to you: should this be its own module or rolled up into Obj.updateText / Obj.spliceText?

note that Obj.getValue already exsists so there's a precedent for data-level helpers in Obj module

packages/core/echo/echo-client/src/echo-handler/text-crdt.test.ts
dmaretskyi: Done — moved the whole suite to `packages/core/echo/echo-client-e2e/src/text-crdt.test.ts` and rewrote it on `TestDatabaseLayer` + `Database.add` (real DB-backed objects instead of `createObject` materialization). Deleted the old `echo-client/src/echo-handler/text-crdt.test.ts`. Added `@automerge/automerge` to echo-client-e2e for the cursor-survival (minimal-delta) assertions. All 12 cases pass.

---

packages/core/echo/echo/src/Text.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * A single find/replace edit applied to a string field. When `oldString` is missing or empty the
+ * `newString` is appended to the end of the current value.
+ */
+export type Edit = {
```
richburdon, line 20: I don't see this referenced in this PR, but your intention is to replace the new packages/core/echo/echo-doc/src/edits.ts correct? I needed to factor out this stuff and create echo-doc, and I still think it makes sense to have a low-level AM package, but I think what you have here is a more principled implemention. Just consider moving it to core/echo/echo-doc.

NOTE: I called it doc, because it goes beyond just text.

packages/core/echo/echo/src/Text.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Apply a sequence of find/replace {@link Edit}s to the string value at `path`, returning the resulting
+ * full string. Implemented on top of {@link splice} so it works for both backends.
+ *
+ * An edit with a missing or empty `oldString` appends its `newString`. A non-`replaceAll` edit whose
+ * `oldString` is not found throws {@link Err.TextEditNotFoundError}.
+ *
+ * Must be called inside `Obj.update(obj, () => { ... })`.
+ */
+export const applyEdits = (obj: Obj.Unknown, path: KeyPath | string | number, edits: readonly Edit[]): string => {
```
richburdon: Maybe just apply?

packages/core/echo/echo/src/Text.ts
dmaretskyi: Two thoughts: 

1. having an AM-specific API is a good idea -- that's what echo-doc solves
2. text-editing is a core functionality that must be in an echo package that's AM agnostic.

packages/core/echo/echo/src/Text.ts
dmaretskyi: claude, do this

dmaretskyi: Done — renamed `Text.applyEdits` → `Text.apply` across the module, tests, and the markdown call site.

---

PR #11974, opened 2026-06-26
refactor(echo-client): prototype-based ECHO object proxies

dmaretskyi:
## Summary

Lightens the ECHO object proxy modules by giving objects a proper **prototype chain** instead of resolving everything through large `switch`/`case` chains in the proxy traps.

> Rebased onto latest `main` (the package was renamed `echo-db` → `echo-client` and `symbolInternals` now points directly at `ObjectCore`); the refactor is re-applied against that model.

Each ECHO object proxy now has the layering:

```
target (clean)
  └─▶ instanceState   per-object state: [symbolInternals] / [symbolNamespace] / [symbolPath] / [EventId]
        └─▶ Echo{Root,Record}Prototype   shared ECHO system surface (accessors + methods), null-terminated
```

### Why the intermediate `instanceState`

`ProxyHandlerSlot` lets a proxy's handler be reassigned without changing object identity. When handlers write their per-object state as own-properties on the *shared* target, they compete for the target's object shape. Keeping that state on an intermediate `instanceState` object and pointing the target's prototype at it means a handler swap is just a prototype re-point — neither handler perturbs the target's shape.

### What changed (`EchoReactiveHandler`)

- The `get` trap's ~120-line switch collapses to: cross-cutting internal accessors → arrays → `Reflect.has(target, prop)` delegation to the prototype → virtual document-backed data.
- The ECHO system surface (`id`, `[Type]`, `[Schema]`, `[Meta]`, `[Parent]`, relation source/target, `[ChangeId]`, `toJSON`, …) is declared as accessors/methods on `EchoRootPrototype` (root objects) / `EchoRecordPrototype` (nested & meta records). The root/nested distinction is now **structural** rather than re-derived via `isRootDataObject` on every access. Newly-public accessors carry JSDoc.
- Prototypes are **null-terminated** so `Reflect.has` only reports the ECHO surface, preserving today's behaviour for `Object.prototype` members and user data. Meta `createdAt`/`updatedAt` virtual props are preserved.
- A `getPrototypeOf` trap hides the internal prototypes (so `instanceof`/`Object.getPrototypeOf` behave as for a plain object/array).
- The in-place `typed → database` handler swap in `createObject` re-points the prototype to a fresh `instanceState` (preserving the existing reactivity `Event`) and strips own-properties left by the previous handler once migrated into the document — **except** the relation-endpoint proxies, which `EchoDatabase.add` re-reads to re-stamp cross-space/relative refs once the space is known.
- Arrays keep their existing `EchoArray` prototype chain and are exempt from the split.

### Tests

`moon run echo-client:test` → **587 passed, 7 expected-fail, 17 skipped, 2 todo** — identical to the pre-refactor baseline on `main`.

### Follow-up (in this PR)

Converging `TypedReactiveHandler` (in-memory typed objects) onto the same prototype/`instanceState` model — in progress.

https://claude.ai/code/session_01KEbJVxExCkuE4a7KaF6nq7


packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,223 @@
...
+ * Recover the raw proxy target from a receiver.
+ * Accessors on the prototypes below run with `this` bound to the receiver, which
+ * is the proxy when reached through the `get`/`set` traps, but the raw target when
+ * a system property is read directly off the target (e.g. internal code).
+ */
+const targetOf = (receiver: any): ProxyTarget => (isProxy(receiver) ? getProxyTarget(receiver) : receiver);
+
+// The handler singleton is injected by the echo-handler module (which imports this
+// one) to avoid a runtime import cycle; the type-only import above is erased.
+let installedHandler: EchoReactiveHandler | undefined;
```
dmaretskyi: ugly, figure out how to fix

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Behaviour prototype for root ECHO data objects. Inherits from
+ * {@link EchoRecordPrototype}; the extra members here are the reason a root
+ * object structurally differs from a nested record — replacing the former
+ * `isRootDataObject(target)` branching in the `get`/`has` traps.
+ */
+export const EchoRootPrototype: object = Object.create(EchoRecordPrototype);
+
+Object.defineProperties(EchoRootPrototype, {
```
dmaretskyi: move the implemetation from handler over to the prototype, dont use `object` type -- type it correctly so that methods have corect `this` type

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,223 @@
...
+  }),
+  [SchemaKindId]: getter(function () {
+    return handler().getSchemaKind(targetOf(this), this);
+  }),
+  [StaticTypeSchemaSlot]: getter(function () {
+    return handler().getStaticTypeSchemaSlot(targetOf(this), this);
+  }),
+});
+
+/**
```
dmaretskyi, line 628: also migration other proxy-handlers -- typed one too

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,656 @@
...
+    return getDevtoolsFormatter(this);
+  }
+}
+
+/**
+ * Behaviour prototype for root ECHO data objects. The extra members here are the
+ * reason a root object structurally differs from a nested record — replacing the
+ * former `isRootDataObject(target)` branching in the `get`/`has` traps.
+ */
+export class EchoRoot extends EchoRecord {
```
dmaretskyi, line 528: add constructions to those classes (marked private) and make them throw so its extra-clear this is not supposed to be instantiated

packages/core/echo/echo/src/internal/common/proxy/typed-handler.ts
dmaretskyi: those swithh cases is exactly what I wanted to move onto prototypes

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,656 @@
...
+  const metaTarget = createRecordTarget(createInstanceState(target[symbolInternals], META_NAMESPACE, []));
+  return createProxy(metaTarget, getProxyHandler(self)) as any;
+};
+
+/**
+ * Stored schemas surface through the database schema registry so consumers see the registered
+ * `Type.Type` entity rather than the raw persisted object. Only *persisted* (db-backed) stored
+ * schemas need registration; an entity resolved directly from the registry passes through unchanged.
+ */
+export const handleStoredSchema = (target: ProxyTarget, object: any): any => {
```
dmaretskyi, line 365: is this not already handled by RefResolver? perhaps move it there and make it clean. also kill createRefResolver.middleware

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,656 @@
...
+// reached through a trap); both resolve `[symbolInternals]` to the backing
+// `ObjectCore` via the prototype chain. Methods read state off `this` and
+// delegate the larger read paths to the free functions below — which are shared
+// with the handler's data-path traps, so the implementation has a single home.
+
+//
+// Free functions (the implementation, operating on a proxy target).
+//
+
+const getNamespace = (target: ProxyTarget): string => target[symbolNamespace];
```
dmaretskyi, line 125: I wonder if most of them can go as methods onto the prototype, maybe as private (true #private) ?

packages/core/echo/echo-client/src/echo-handler/echo-prototypes.ts
```diff
@@ -0,0 +1,656 @@
...
+  const database = getEchoDatabase(target[symbolInternals]);
+  if (database) {
+    return resolveStrongDepFromWorkingSet(database, parentDXN);
+  } else {
+    invariant(target[symbolInternals].linkCache);
+    const parentEchoUri = EID.tryParse(parentDXN);
+    const echoUri = parentEchoUri ? EID.getEntityId(parentEchoUri) : undefined;
+    invariant(echoUri);
+    return target[symbolInternals].linkCache.get(echoUri);
+  }
```
dmaretskyi: is see this pattern a lot here. maybe extract this to object core

packages/core/echo/echo-client/src/echo-handler/echo-handler.ts
```diff
@@ -494,12 +456,7 @@ export class EchoReactiveHandler implements ReactiveHandler<ProxyTarget> {
       const newTarget = defaultMap(
         target[symbolInternals].targetsMap,
         targetKey,
-        (): ProxyTarget => ({
-          [symbolInternals]: target[symbolInternals],
-          [symbolPath]: dataPath,
-          [symbolNamespace]: namespace,
-          [EventId]: new Event(),
-        }),
+        (): ProxyTarget => createRecordTarget(createInstanceState(target[symbolInternals], namespace, dataPath)),
```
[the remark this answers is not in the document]
dmaretskyi: Good catch — fixed in 4c850722eb. Derived record and meta proxies now reuse the root target's `EventId` (matching the array path), so subscribers attached to a nested/meta proxy are notified by the central `core.updates` emit. Added a regression test in `subscription.test.ts`.

---

PR #12052, opened 2026-07-02
feat(composer-crx): use a Chrome side panel instead of a popup

richburdon:
## Summary

Converts the Composer browser extension (`composer-crx`) from a toolbar **popup** to a Chrome **side panel**, and fixes the MV3 issues that surfaced along the way (CSP, first-install content-script reachability, and messaging errors).

### Side panel
- Registered `panel.html` as the extension's side panel (`side_panel.default_path`), added the `sidePanel` permission, and removed `action.default_popup` — a popup and an action-click side panel are mutually exclusive. Clicking the toolbar icon now opens the panel (`setPanelBehavior({ openPanelOnActionClick: true })`).
- Ported the full chat / page-actions / thumbnail UI from the old popup into `panel.tsx`, adapted for a **persistent** surface: it reacts to tab switches/navigation (`tabs.onActivated`/`onUpdated`) and picks up thumbnails via `storage.onChanged` (the popup read once on open and closed on blur).
- Deleted `popup.html` / `popup.tsx` and dropped the popup link/`openPopup()` fallback.

### MV3 fixes
- **CSP**: `ThemePlugin` injects an inline `<script>` into every page's `<head>`, which the MV3 `extension_pages` CSP (`script-src 'self'`, no hash/nonce allowed) blocks. Added a `transformIndexHtml` post-pass in `vite.config.ts` that strips it; the extension is dark-only (`Container` forces `themeMode='dark'`) so `class="dark"` on the pages covers first paint. Inline `<style>` tags are left intact.
- **First-install reachability**: declarative content scripts only inject on navigations *after* install, so already-open tabs had no receiver ("Receiving end does not exist"). `background.ts` now injects the declared content script into open http/https tabs on `onInstalled` via `browser.scripting.executeScript`.
- **Fire-and-forget messaging**: the `PAGE_ACTIONS_READY` handler returned the long-running `refreshRegistry` promise, holding the message channel open and producing "async response … channel closed" / "receiving end does not exist" when the SW suspended or the tab navigated. It's now fire-and-forget.
- **Connection-driven registry refresh**: removed the proactive startup `refreshRegistry()` probe that messaged tabs whose content script wasn't attached yet. Refresh is now driven by the content script announcing itself (`PAGE_ACTIONS_READY`) — the background never messages a tab it hasn't heard from. `refreshRegistry`'s unreachable-tab path logs at info (a missing content script is a normal state), matching the existing convention in `invoke.ts`.

### Types
- Added a minimal ambient `chrome.sidePanel` declaration in `typings.d.ts` (pinned `@types/chrome@0.0.125` predates the API and is excluded from tsconfig `types`).

## Reviewer notes
- No new `as`/`!` casts. `class="dark"` on the extension pages is intentional (dark-only UI).
- The literal "click toolbar icon → panel docks" and the context-menu `sidePanel.open()` gesture can't be automated from CI/agent tooling; they need a manual check with the unpacked extension. Everything else (SW health, CSP-clean pages, content-script injection, messaging) was verified.

## Verification
- Typecheck (tsgo), lint (0/0), and the full `composer-crx` test suite (104 tests) pass. Extension bundles cleanly; generated `manifest.json` has `side_panel` + `sidePanel` and no `default_popup`.

## Summary by CodeRabbit

* **New Features**
  * Replaced the toolbar popup with a dedicated side panel (popup removed) and set the side panel as the default UI.
  * Side panel now tracks the active tab, responds to saved thumbnail/config state, and drives clip/ping interactions.
  * Enabled dark-mode styling across the extension UI.
* **Bug Fixes**
  * Improved reliability by injecting content scripts into already-open HTTP/HTTPS tabs during install/reload.
  * Right-click and toolbar interactions now open the side panel immediately and more consistently.
  * Page action refresh is more tolerant when content scripts aren’t attached yet.
* **UI**
  * Adjusted image layout to prevent the image container from shrinking in flex layouts.


richburdon:
@coderabbitai Addressed both nitpicks in e3249807e9:

- **`vite.config.ts` — brittle strip regex:** the pattern now matches the `data-dxos-theme` marker attribute regardless of attribute order/extra attributes (`<script\b[^>]*\bdata-dxos-theme\b[^>]*>…`), and warns at build time if nothing was stripped — so a future ThemePlugin markup change surfaces in the build log instead of silently reintroducing the CSP violation.
- **`panel.tsx` — duplicated active-tab query:** extracted a `getActiveTab()` helper resolved at call time (preserving the intentional no-stale-closure behavior) and reused it from `useActiveTab`, `handlePing`, and `handleClip`.

Typecheck (tsgo), lint (0/0), and the 104-test suite pass.

PR #12071, opened 2026-07-03
feat(echo): support changing feed items by appending a new entry with the same id

dmaretskyi:
## Summary

Feeds are append-only logs, so a feed item cannot be edited in place. This adds change support for feed objects **through the existing `Obj.update` API** — no new API surface. Mutating a feed item persists a new entry to the feed with the **same id**, and reads collapse entries by id (keeping the latest) so queries observe the updated value.

Mechanism (all in `FeedHandle`, echo-client):
- Each decoded/appended feed object gets a persist-on-change subscription (`Obj.subscribe`). On mutation via `Obj.update`, a new entry with the same id is appended to the queue. Idempotent per object (`WeakSet` guard); `append` only touches hidden metadata + the local set, so it never re-triggers the subscription (no loop).
- `Obj.update` is synchronous, so the append is fire-and-forget. In-flight appends are tracked in `#pendingWrites`; `fetchObjectsJSON` (query `run` / ref resolution) and `sync` (push) drain them first, so a read or sync after a change deterministically observes it.
- Reads collapse entries by id, keeping the latest (`dedupeByIdKeepLast`), applied to the optimistic append view, the reactive refresh, and `fetchObjectsJSON`. Both the original and updated blocks remain in the append-only log.

Usage:
```ts
const [post] = await Feed.query(feed, Filter.type(Post)).run;
Obj.update(post, (p) => { p.title = 'Updated'; }); // appends a new entry with post's id
```

## Changes

- `packages/core/echo/echo-client/src/feed/feed-handle.ts` — persist-on-change subscription, pending-write tracking + drain in `fetchObjectsJSON`/`sync`, and id-collapsing (`dedupeByIdKeepLast`) across the read paths.
- `packages/core/echo/echo-client/src/feed/feed.test.ts` — tests driving the feature.

(An earlier revision added a dedicated `Feed.change` / `Database.changeFeedItem` API; per review it was removed in favor of the existing `Obj.update`, so `Feed.ts` / `Database.ts` / `proxy-db/database.ts` are unchanged from `main`.)

## Tests

`describe('change')` in `feed.test.ts`:
1. `Obj.update` on a feed item persists a new version with the same id — a re-query returns the single latest version, and the queue retains both blocks (append-only).
2. The change is visible to a **fresh client** reading the queue from scratch (after `sync`).
3. `query.subscribe` reflects a changed feed item.

## Preview

https://pr-12071.composer-app.pages.dev

https://claude.ai/code/session_01PNiYNkVm2v4JiSYn8V4uLk

## Summary by CodeRabbit

* **New Features**
  * Feed items now stay in sync when updated, with the latest value shown for each item ID.
  * Live query subscriptions and refreshes now correctly reflect `Obj.update` changes.

* **Bug Fixes**
  * Duplicate feed entries are collapsed so only the most recent version per ID is shown.
  * Reading feed data now skips invalid entries instead of failing the entire result.

### DIFF
--- a/packages/core/echo/echo/src/Database.ts
+++ b/packages/core/echo/echo/src/Database.ts
@@ -180,6 +180,19 @@ export interface Database extends Queryable {
   deleteFromFeed(feed: Feed.Feed, entities: Entity.Unknown[]): Promise<void>;
 
   /**
+   * Applies a mutation to a feed item and persists the result.
+   *
+   * Feeds are append-only logs, so a change is stored as a new entry with the same id rather than
+   * an in-place edit. Reads collapse entries by id (keeping the latest), so queries observe the
+   * updated value.
+   */
+  changeFeedItem<T extends Obj.Unknown>(
+    feed: Feed.Feed,
+    item: T,
+    callback: (item: Obj.Mutable<T>) => void,
+  ): Promise<void>;
+
+  /**
    * Wait for all pending changes to be saved to disk.
    * Optionaly waits for changes to be propagated to indexes and event handlers.
    */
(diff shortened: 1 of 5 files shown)


packages/core/echo/echo/src/Database.ts
```diff
@@ -179,6 +179,19 @@ export interface Database extends Queryable {
...
   deleteFromFeed(feed: Feed.Feed, entities: Entity.Unknown[]): Promise<void>;
 
+  /**
+   * Applies a mutation to a feed item and persists the result.
+   *
+   * Feeds are append-only logs, so a change is stored as a new entry with the same id rather than
+   * an in-place edit. Reads collapse entries by id (keeping the latest), so queries observe the
+   * updated value.
+   */
+  changeFeedItem<T extends Obj.Unknown>(
```
dmaretskyi: no, make it work with existing api

dmaretskyi: Done — dropped the `Feed.change` / `Database.changeFeedItem` surface. `Obj.update` on a feed item now persists the change: `FeedHandle` subscribes to each feed object and, on mutation, appends a new entry with the same id (reads collapse by id, keeping the latest). The fire-and-forget append is tracked so queries/sync wait for it. Only `feed-handle.ts` + `feed.test.ts` change now.

---

PR #12078, opened 2026-07-04
fix(plugin-onboarding,client): surface real oauth recovery errors, fix stuck leader watcher

wittjosiah:
## Summary

Investigated an `oauth recovery finalize failed` error surfaced in SigNoz (two occurrences this morning, same user, both failing identically), which uncovered two related bugs:

- **`plugin-onboarding`**: [`oauth-recovery-redirect.ts`](packages/plugins/plugin-onboarding/src/capabilities/oauth-recovery-redirect.ts) logged `UnknownException.message` on finalize failure. Since the RPC calls inside `finalizeRedirect` use bare `Effect.tryPromise(() => ...)` with no `catch` mapper, Effect always sets `.message` to the generic boilerplate `"An unknown error occurred in Effect.tryPromise"` — the real cause was sitting in `.error`/`.cause` and being silently discarded, making every occurrence of this error undiagnosable from telemetry. Now unwraps the real cause before logging, and shows a toast on the generic failure path so a failed OAuth finalize is no longer silent to the user (mirrors the existing `params.error` toast for kms-service pre-identity failures).
- **`client`**: [`DedicatedWorkerClientServices#watchLeader`](packages/sdk/client/src/services/dedicated/dedicated-worker-client-services.ts) gave up permanently after any non-abort error from the leader session (e.g. a worker crash/init failure), instead of retrying. Since the exclusive lock is released once the callback throws, this left the tab (and any followers) unable to ever elect a new leader for the lifetime of the page. This matches a real production symptom found in the same SigNoz window: two separate sessions stuck in a 6-minute loop of `waiting for dedicated worker provide-port RPC reply` timeouts, retrying every ~15s with no leader ever re-elected. Now backs off (`retryBackoff`, default 1s, configurable like the other `leaderTimeouts`) and re-enters election.

## Test plan

- [x] Added a regression test (`recovers when the leader session itself fails, backing off and re-electing`) that injects a `createWorker` failure on the first leader-election attempt and asserts the client still connects after the watcher retries.
- [x] Verified the regression test actually fails (hangs, times out) against the pre-fix `watchLeader` code, confirming it exercises the real bug rather than being a trivial happy-path check.
- [x] Verified the test is stable (10+ repeated no-cache `vitest run` invocations) after tuning `leaderTimeouts` — an earlier, overly aggressive shrink of those values caused genuine intermittent hangs from lock/timer races, not just slowness.
- [x] `moon run client:build client:lint client:test` and `moon run plugin-onboarding:build plugin-onboarding:lint plugin-onboarding:test` all pass.
- [x] No `as any`/`as unknown`/`as <Type>` casts introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth recovery failure handling with clearer, consistent error reporting and a reliable fallback notification.
  * Enhanced dedicated worker leader election to automatically recover when the leader session fails, using configurable retry backoff and avoiding follower stalls while waiting to connect.
* **Tests**
  * Added automated coverage to verify recovery behavior when worker leadership initialization fails, including retry/backoff and re-election.

### DIFF
--- a/packages/plugins/plugin-onboarding/src/capabilities/oauth-recovery-redirect.ts
+++ b/packages/plugins/plugin-onboarding/src/capabilities/oauth-recovery-redirect.ts
@@ -2,6 +2,7 @@
 // Copyright 2026 DXOS.org
 //
 
+import * as Cause from 'effect/Cause';
 import * as Effect from 'effect/Effect';
 
 import { Capabilities, Capability } from '@dxos/app-framework';
@@ -84,6 +85,16 @@ const deleteSnapshot = (accessTokenId: string | undefined): void => {
 };
 
 /**
+ * `Effect.tryPromise` wraps a rejected promise in `UnknownException`, whose own `.message` is a
+ * generic boilerplate string ("An unknown error occurred in Effect.tryPromise") — the real cause
+ * lives in its `.error` property, so unwrap it before logging or it is silently lost.
+ */
+const describeError = (error: unknown): string => {
+  const cause = Cause.isUnknownException(error) ? error.error : error;
+  return cause instanceof Error ? cause.message : String(cause);
+};
+
+/**
  * Startup module that finalizes redirect-flow OAuth-recovery callbacks.
  *
  * atproto/bsky nullifies `window.opener`, so the register / recovery flows cannot relay their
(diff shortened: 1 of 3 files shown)


wittjosiah:
Addressed the CodeRabbit nitpick on `dedicated-worker-client-services.ts` (fixed retry backoff with no cap/jitter): the leader-retry backoff is now exponential (capped at 30s) with jitter, tracked via a `#leaderFailureCount` that resets once a leader session opens successfully. This bounds retry frequency under a persistently failing worker and avoids synchronized retries across tabs, per the finding.

(Note: CodeRabbit's comment wasn't posted as an inline/threaded review comment — there's no review thread to reply-in-thread to or resolve via the API, so posting this as a regular PR comment instead.)

PR #12088, opened 2026-07-04
feat(echo): paginated queries for ECHO feeds

wittjosiah:
## Summary

- New `usePagination` hook (`@dxos/echo-react`) wraps a query with page management: `getNext`/`getPrevious`/`jumpToHead`, `hasMore`, `isLoading`, `atHead`. The query's own `.limit(n)` is the page size; the hook rewrites `skip`/`limit` internally and bounds the loaded window at `maxWindowSize` (default 10× page size), evicting the far edge once exceeded. Built directly on `useSyncExternalStore`, mirroring `useQuery`'s own `subscribe`/`getSnapshot` pattern — no `useEffect`/`useState`, no `.run()`.
- New general-purpose query clauses to express this: `.skip(n)` (mirrors `.limit(n)`) and `Order.natural(direction)` (feed insertion order, for when there's no schema field to sort by).
- New `useVirtualizerPagination` hook (`@dxos/react-ui-mosaic`) factors the pagination-trigger + scroll-anchor-compensation logic into a reusable piece any `Mosaic.VirtualStack` consumer can opt into — structurally typed against `{ getNext?, getPrevious? }`, no dependency on echo-react. Demonstrated with a non-ECHO story (`Stack/Pagination`) over a plain in-memory array.
- Wired into the inbox mailbox (`MailboxArticle`/`MessageStack`) as the first real consumer.

### Caveat: this is pagination, not lazy loading

`usePagination` bounds what's **held and rendered** (via `maxWindowSize` + eviction), but does not reduce what's **fetched**: every `getNext`/`getPrevious` re-fetches and decodes the entire underlying feed before slicing out the requested page, regardless of ordering. There is no cursor/lazy-fetch machinery in this PR — `FeedStore`/`FeedProtocol`/the feed service only support plain forward reads (`after`/`limit`), which is what `FeedHandle` uses today.

True lazy loading is left as a `TODO` on `usePagination` and `FeedQueryContext`: the right fix is a general, index-backed keyset-pagination feature covering all orderings uniformly, once the host index (`EntityMetaIndex`/`QueryExecutor`) supports ordered range reads — today it has the same full-scan-then-sort-then-slice limitation for content-based ordering on *any* ECHO object, not just feeds, so this is really a query-engine gap rather than a feed-specific one. (An earlier version of this PR built a `FeedWindow` cursor cache that made only `Order.natural('desc')` genuinely lazy; it was removed since a partial-laziness story for one ordering wasn't worth carrying.)

## Test plan

- [x] `moon run feed:test echo:test echo-host:test echo-client:test echo-react:test` — all green.
- [x] `moon run :lint -- --fix` — clean across all packages.
- [x] `moon run react-ui-mosaic:build plugin-inbox:build echo-react:build` — clean.
- [x] `MOON_CONCURRENCY=4 moon run :test -- --no-file-parallelism` — full monorepo suite, all passing.
- [x] CI (Check workflow) — passing.
- [x] Verified `plugins/plugin-inbox/containers/MailboxArticle` (`Default`) and `ui/react-ui-mosaic/Stack/Pagination` (`Default`) stories in Storybook.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added client-side pagination for feed/query results, including next/previous page controls and scroll-aware loading.
  * Virtualized lists now support loading more items as you scroll, with smoother position retention during page changes.
  * Query sorting now respects ordering direction consistently, including natural ordering.

* **Bug Fixes**
  * Fixed offset-based reads so skipped results are handled correctly across query execution paths.
  * Improved feed loading behavior for older/newer items and deleted entries in paginated results.


packages/core/echo/echo-query/src/query-lite/query-lite.ts
```diff
@@ -30,7 +30,8 @@ class OrderClass implements Order$.Any {
 }
 
 namespace Order1 {
```
dmaretskyi, line 32: I remember @richburdon said he removed quickjs, maybe this module is not needed anymore?

packages/core/echo/echo-react/src/usePagination.ts
```diff
@@ -0,0 +1,271 @@
...
+   * growing further). Defaults to 10x the query's page size.
+   */
+  maxWindowSize?: number;
+};
+
+export type PaginationResult<T> = {
+  /** Loaded window, in query order (newest first for `Order.natural('desc')`). */
+  items: T[];
+  /** Extends toward older items by one page. Single-flight; no-op while loading or exhausted. */
+  getNext: () => void;
```
dmaretskyi, line 23: the api seems good, one small detail: do we need to retain the concept of a page here, or can we say:

```ts
loadNext(count: number): void
```

this way we don't need to think in terms of pages, we're just fetching more items.

PR #11054, opened 2026-04-21
refactor: Use ProcessManager to run composer operations

dmaretskyi:
## Summary

Decomposes the compute-runtime package into focused modules so the process manager is easier to read and extend. No behavior changes — pure mechanical refactor on top of the existing service-mesh work.

Starting point was a single 1234-line `ProcessManager.ts`. After this branch:

| File | Lines | Purpose |
|------|-------|---------|
| `ProcessManager.ts` | 638 | Public types, `ProcessManagerImpl`, scoped layer factory |
| `ProcessHandle.ts` | 457 | `ProcessHandleImpl` — status atom, handler bookkeeping, alarms, ephemeral subscribers, lifecycle transitions |
| `ProcessOperationInvoker.ts` | 246 | Operation-to-process invoker + its layer |
| `process-trace.ts` | 203 | `buildBaseTraceContext`, `prepareProcessTracing`, `createProcessTraceService` |
| `trace-buffer.ts` | 87 | `EphemeralTraceBuffer` + `detachData` (strips ECHO proxies from `Schema.Unknown` payloads) |
| `process-id.ts` | 20 | `UUIDProcessIdGenerator` / `SequentialProcessIdGenerator` |

The branch also includes the earlier work on this session:

- `fix(compute-runtime): detach ECHO proxies from trace payloads` — fixes `Obj.make` recursion through `Schema.Unknown` event payloads that was breaking queue-trigger traces.
- `refactor(app-framework): route OperationInvoker through process manager` and `refactor(app-framework): drive HistoryTracker from trace messages` — part of the service-mesh / LayerSpec effort.
- `feat(functions): emit ephemeral operation input/output trace events` + sink routing work.

Memoized LLM conversation caches and inline trace-graph snapshots were regenerated to reflect the new trace flow.

## Why

- The monolithic `ProcessManager.ts` made the process-spawn flow hard to follow; each extracted helper has a single, documented responsibility.
- Isolating `ProcessHandleImpl` and the ephemeral buffer keeps the handle's internal state (atoms, queues, alarm timer) separate from the manager's registry/tree bookkeeping.
- `process-trace.ts` is now the single place describing how a process wires up into the tracing system (invocation start/end, ephemeral vs persisted routing, ECHO-proxy detachment).

## Test plan

- [x] `moon run compute-runtime:build` passes
- [x] `moon run compute-runtime:test` — 32/32 pass
- [x] `moon run functions-runtime:test` — 39 passed, 7 skipped
[...]

PR #11044, opened 2026-04-21
feat(assistant): add space-level MCP server configuration and chat invocation

richburdon:
## Summary

- Adds `McpServer` ECHO type in `@dxos/assistant-toolkit` for space-level MCP server configuration (name, URL, protocol, optional API key, enabled toggle)
- Adds "MCP Servers" tab in ChatOptions UI for CRUD management of MCP servers in the space
- Extends `McpToolkit.make()` with optional `apiKey` for authenticated MCP server connections
- Extends `AiConversation.createRequest()` with optional `mcpServers` param for space-level servers
- Wires space MCP servers through the agent process path via `AgentService.layer`

## Design

MCP servers are configured as ECHO objects at the space level, shared across all chats. The assistant transparently includes MCP tools alongside built-in tools — users invoke them via natural language.

Design spec: `docs/superpowers/specs/2026-04-21-mcp-chat-invoke-design.md`

## Future work

- `@completion` in prompt editor for explicit tool invocation
- Per-chat server filtering
- Tool grouping/labeling by server origin

## Test plan

- [ ] Build passes for all modified packages
- [ ] Lint passes
- [ ] Existing tests pass
- [ ] Manual: add MCP server via ChatOptions UI, verify it appears in the list
- [ ] Manual: toggle server enabled/disabled, verify tools are available/unavailable
- [ ] Manual: connect to an MCP server with API key auth

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Space-level MCP server management in chat settings: list, toggle, add, remove
  * Per-server protocol selection (SSE/HTTP) and optional API key for authenticated connections
  * Enabled servers are automatically included in conversations and automation flows

* **Documentation**
  * Added design spec describing chat→MCP invocation behavior

* **Style**
  * UI text/translations and minor chat UI refinements for MCP controls

## Demo

https://github.com/user-attachments/assets/a3c1b8dd-ba61-42e2-9da6-54312e9c9e70

PR #12089, opened 2026-07-04
fix(assistant): restore space-level MCP server wiring to agent sessions

Steve:
## Summary

Space-level MCP server configs — `McpServer` ECHO objects added via the chat MCP panel — stopped reaching agent sessions, so a configured MCP server connected to nothing.

**Root cause.** The consumption wiring (a `getMcpServers` callback querying the space's enabled `McpServer` objects) originally lived in `plugin-automation`'s compute-runtime capability (#11044). The ProcessManager refactor (#11054) deleted that file, and the successor application-wide `AgentService` layer never re-provided the callback. The MCP panel and `McpServer` type survived; the wire that consumed them did not.

## Fix

Reconnect the wiring, now parameterized by `spaceId` — a single application-wide `AgentService` serves sessions across all spaces (unlike the original per-space layers), so the resolver must know which space a session belongs to:

- **`functions-runtime/agent-service/AgentService.ts`** — `getMcpServers` becomes `(spaceId) => Promise<McpServer[]>`; `getSession` binds the session's space (already derived from the feed URI) into the spawned process.
- **`functions-runtime/agent-service/agent-process.ts`** — resolves the space's MCP servers per request, so servers added or toggled at runtime take effect without respawning the process.
- **`plugin-assistant/capabilities/agent-service.ts`** — provides the resolver: looks up the space by id and returns its enabled `McpServer` configs. This is the layer permitted to depend on `assistant-toolkit`'s `McpServer` ECHO type, which keeps the dependency direction correct (the reason the resolver can't live down in `functions-runtime`).

Covers both interactive chat and the scheduled `AgentWorker`, since both spawn agent processes.

## Test

`AgentService.test.ts` gains a regression test that runs a turn with a spy resolver and asserts it is invoked with the session's `spaceId` — which never happened before this fix. It reuses an existing memoized conversation fixture, so it runs green without regeneration.

## Verification

- Full workspace build passes.
- `moon run :lint -- --fix`: 0 errors / 0 warnings, no files changed.
- `functions-runtime` agent-service suite passes (incl. the new test).
- No new type casts.

## Summary by CodeRabbit

* **New Features**
  * Agent sessions now resolve MCP server settings per space, allowing different spaces to use different configured servers.
  * MCP server discovery is performed at request time and includes enabled-server filtering.

* **Bug Fixes**
  * Agent requests now re-resolve MCP servers on each turn for consistent behavior when switching spaces.
  * Restored sessions now retain space context so MCP resolution works after restart.
  * If a space isn’t ready or unavailable, the agent continues safely without MCP servers.

### DIFF
--- a/packages/core/compute/functions-runtime/src/agent-service/AgentService.test.ts
+++ b/packages/core/compute/functions-runtime/src/agent-service/AgentService.test.ts
@@ -21,7 +21,7 @@ import { getSession, hydrate } from '@dxos/compute/AgentService';
 import { Annotation, Feed, Filter, Obj, Ref } from '@dxos/echo';
 import { TestHelpers } from '@dxos/effect/testing';
 import { AssistantTestLayer } from '@dxos/functions-runtime/testing';
-import { DXN, EntityId } from '@dxos/keys';
+import { DXN, EID, EntityId, type SpaceId } from '@dxos/keys';
 import { Message, Organization } from '@dxos/types';
 
 import { ProcessManager } from '../index';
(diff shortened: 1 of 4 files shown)


Steve:
Thanks @coderabbitai — all three findings addressed in `ac22363`:

- **`agent-process.ts` — resolver rejection could abort the turn.** Wrapped so a failed resolver falls back to no servers (`Effect.tryPromise` + `tapError` + `orElseSucceed(() => undefined)`) instead of becoming a turn-killing defect. *(inline thread auto-resolved)*
- **`plugin-assistant/agent-service.ts` — unbounded `waitUntilReady`.** Bounded with `asyncTimeout(..., 10s)`; on timeout the turn proceeds with no servers rather than stalling. *(inline thread auto-resolved)*
- **`AgentService.hydrateAgents()` — rehydrated agents ran unscoped.** This one was posted outside the diff, so noting it here: `hydrateAgents()` now builds a per-agent executable scoped to `agent.environment.space` (persisted on the handle, so it survives restart), closing the gap where agents restored after restart resolved no MCP servers.

Build, lint, `oxfmt --check`, and the agent-service test suite all pass.

dmaretskyi (approved):

PR #12090, opened 2026-07-04
feat(echo): add generic Blob API with inline/edge/wnfs backends

wittjosiah:
## Summary

Introduces a generic `Blob` namespace in `@dxos/echo` that abstracts file byte storage behind pluggable backends, replacing the previous inline-only `File.data` field and the WNFS-specific upload/URL-resolver capabilities.

- **`@dxos/echo`**: new `Blob.ts` — schema (`inline`/`external` variants), `Backend` interface (`put`/`get`/`has`/`getUrl?`), `Storage`/`Scheme` constants, and `fromBytes`/`read`/`exists`/`url` ops requiring `Database.Service`.
- **`@dxos/echo-client`**: `BlobManager` dispatches external blobs by URI scheme, registered on `Hypergraph` via `registerBlobBackend`; inline blobs are handled natively with a size cap.
- **`@dxos/edge-client`**: `putBlob`/`getBlob`/`hasBlob`/`deleteBlob` methods; `_callRaw` factored out of the 401-challenge retry loop for raw-body requests.
- **`@dxos/client`**: registers an edge `Blob.Backend` addressed by `sha256:<hex>` content-hash URIs (named for what the identifier is — a digest — not which backend wrote it), with `getUrl` building a direct edge link. Online-only in v1; no local cache (see TODO in `edge-blob-backend.ts` — blocked on GC/refcounting since content-addressed blobs can be shared across objects).
- **`plugin-wnfs`**: adapts existing upload/load helpers to the `Blob.Backend` interface.
- **`@dxos/types`**: `File.data` is now `Ref(Blob.Blob)` (was inline-only bytes); `File.fromBytes` parents the Blob under the File so deleting the File cascades to the Blob. No migration for existing `0.1.0` File objects — they stop appearing once the old schema is deregistered.
- **`plugin-file`**: backend selection reworked around `Blob.Storage` names; settings UI now reads the registry's actual configured default instead of a hardcoded fallback.

### Bug fixes bundled in
- Default storage backend was ignoring the registry's configured default (always fell back to `inline`).
- `FileArticle`/`useImageUrl` re-render loop caused by a `useEffect` keyed on a freshly-constructed `Ref` wrapper on every render.
- `client.ts` race where edge identity setup could throw before the HALO device was ready.

## Test plan
- [x] `moon exec --on-failure continue --quiet :build` — full repo build green
- [x] `MOON_CONCURRENCY=4 moon run :test -- --no-file-parallelism` — full repo test suite green (516 tasks completed)
- [x] `moon run :lint -- --fix` — 0 warnings, 0 errors across 281 packages
- [x] Diff audited for new type casts (`git diff origin/main | grep -nE '\bas (any|unknown|[A-Z])|as unknown as'`) — none found
- [ ] Manual verification in Composer: upload via each backend (inline/edge/wnfs), confirm render in gallery/lightbox

## Summary by CodeRabbit

* **New Features**
  * Added centralized blob management with pluggable storage backends, inline/external blob creation, byte reads, existence checks, and optional renderable URL support.
  * Exposed blob backend registration and default blob storage via Hypergraph and new Database APIs.
  * Added edge blob HTTP support and an edge-backed blob backend; updated WNFS and file handling to use Blob references with `Blob.url`/fallbacks.
* **Bug Fixes**
  * Improved preview URL fallback behavior and safer revocation of generated object URLs.
* **Tests**
  * Expanded Blob, File, Edge, and WNFS test coverage for success, missing/offline, and size-limit scenarios.


packages/core/echo/echo/src/Blob.ts
```diff
@@ -0,0 +1,230 @@
...
+import * as Err from './Err';
+import * as internal from './internal';
+import * as Obj from './Obj';
+import * as Type from './Type';
+
+/**
+ * Inline blob data: bytes stored directly on the ECHO object.
+ */
+export const InlineData = Schema.TaggedStruct('inline', {
+  bytes: Schema.Uint8ArrayFromSelf.annotations({ jsonSchema: { type: 'string', contentEncoding: 'base64' } }),
```
dmaretskyi, line 24: is it worth separating blob data into a separate object so you could query blobs without touching on inline data?

packages/core/echo/echo/src/Blob.ts
```diff
@@ -0,0 +1,230 @@
...
+export interface PutResponse {
+  /** URI locating the stored bytes; must use a scheme the backend resolves. */
+  uri: string;
+}
+
+/**
+ * Implemented by pluggable blob storage backends and registered on the Hypergraph via
+ * `registerBlobBackend`.
+ */
+export interface Backend {
```
dmaretskyi: this feels more like echo-protocol level interface.

I'd omit pluggable blob backends for core @dxos/echo api

packages/core/echo/echo/src/Blob.ts
wittjosiah: we already have this separation with File/Blob

packages/core/echo/echo/src/Blob.ts
wittjosiah: Good call — moved `Backend`/`PutRequest`/`PutResponse` out of `@dxos/echo`'s `Blob.ts` into `@dxos/echo-protocol` as `BlobBackend`/`BlobPutRequest`/`BlobPutResponse` (same pattern already used there for `ForeignKey`/`QueryAST`/`EncodedReference`). Checked `Feed.ts` for precedent — it has no backend-plugin interface at all, it just delegates everything to `Database.Service`, which supports the "this doesn't belong in echo's public surface" read.

`@dxos/echo`'s `Blob` namespace now only exposes the schema (`InlineData`/`ExternalData`/`BlobData`/`Blob`), the `Storage`/`Scheme` constants, and the app-facing `fromBytes`/`read`/`exists`/`url` ops. All backend implementations (echo-client's `BlobManager`, the edge backend in `@dxos/client`, plugin-wnfs's WNFS backend) and `Hypergraph.registerBlobBackend` now import `BlobBackend` directly from `@dxos/echo-protocol`.

3ee4e28d83

PR #12098, opened 2026-07-05
fix(assistant-toolkit): coerce stringified jsonSchema in add-schema tool

dmaretskyi:
## Summary

Fixes the `add-schema` database-skill tool, which failed whenever a model passed the JSON Schema as a JSON-encoded string instead of an object.

## Root cause

`SchemaAdd.input.jsonSchema` was typed `Schema.Any`. That serializes to an unconstrained tool parameter (`{"$id":"/schemas/any"}` — no `type`), so the model is free to emit `jsonSchema` as a **string**. Nothing coerced it: the only decode at the tool-call boundary is `@effect/ai`'s `Toolkit.handle`, and `Schema.Any` decodes as a passthrough. `Type.makeObjectFromJsonSchema` then received a string and produced a corrupted type (the "escaped quotes" symptom).

## Fix

Type the field as a union that accepts an object (preferred) or a JSON string, parsing the string back into an object:

```ts
jsonSchema: Schema.Union(Schema.parseJson(Schema.Object), Schema.Any)
```

- The transform survives the operation→tool projection (`mapSchemaTypeForLLM` passes the union through unchanged) and runs during `@effect/ai`'s parameter decode, so a stringified argument is coerced before it reaches the handler.
- The object path is unchanged (byte-for-byte), and the decoded type stays `any`, so the handler needs no change and no cast is introduced.
- A description now nudges the model to provide an object.

## Why existing tests missed it

Three independent reasons, each sufficient on its own:
1. The assertion only checked that a type with the given `typename` existed (`length >= 1`) — never the created schema's fields. `typename` is a separate top-level arg, so it passes even when the `jsonSchema` payload is malformed.
2. `jsonSchema: Schema.Any` gets no decode/validation at the boundary.
3. It ran against a memoized LLM fixture that froze a clean object payload — the stringified form the bug needs was never exercised.

## Tests

- `schema-add: decodes a stringified jsonSchema at the tool-call boundary` — decodes the operation input with `jsonSchema` as both a JSON string and an object; asserts both yield the object. Deterministic, no LLM.
- `schema-add: creates a schema with the declared fields` — decodes a stringified schema through the tool-call boundary, invokes the handler, and asserts the created type actually carries the declared fields (`name`/`description`/`status`).
- Strengthened the existing LLM-driven `schema-add` test to also assert the created schema's fields.

## Verification

Local build/test is blocked in this environment (the moon TypeScript toolchain plugin download is egress-denied), so the vitest suite runs in CI. Verified here: the coercion behavior via standalone `effect` Schema probes (object passthrough identical, string→object, LLM-facing schema unchanged apart from the description), and a source-resolved `tsc` typecheck of the changed files (677 files loaded, no type errors).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lj3AjcySUTwunJTM5qzfMo)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of database schema input by accepting schema definitions provided as either a JSON object or serialized JSON text.
  * Added stricter validation to ensure only object-shaped JSON schemas are accepted at the tool boundary.
  * Strengthened checks that newly created schemas retain expected fields/properties.
* **Tests**
  * Expanded coverage for schema decoding at the tool boundary, including rejection of invalid (non-object) payloads.
  * Added/strengthened assertions verifying the created schema’s expected properties and contents.


packages/core/compute/assistant-toolkit/src/skills/database/operations/definitions.ts
```diff
@@ -199,7 +199,13 @@ export const SchemaAdd = Operation.make({
...
       description: 'The typename of the schema in the format of "com.example.type.type".',
     }),
-    jsonSchema: Schema.Any,
+    // Unconstrained tool parameters are serialized to the LLM without a `type`, so some models
+    // emit the JSON Schema as a JSON-encoded string rather than an object. Accept either form and
+    // decode a string back into an object at the tool-call boundary, since `makeObjectFromJsonSchema`
+    // requires an object.
+    jsonSchema: Schema.Union(Schema.parseJson(Schema.Object), Schema.Any).annotations({
+      description: 'The JSON Schema (draft-07) object describing the fields of the new type. Provide a JSON object.',
+    }),
```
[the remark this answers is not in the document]
dmaretskyi: Good catch — fixed in 560f01663. Narrowed the fallback to `Schema.Object` so a malformed payload (invalid JSON, or a non-object) now fails at the decode boundary instead of reaching `makeObjectFromJsonSchema`. The decoded type stays assignable without a cast, and I added a malformed-input assertion to the boundary test.

---

packages/core/compute/assistant-toolkit/src/skills/database/operations/definitions.ts
```diff
@@ -199,7 +199,13 @@ export const SchemaAdd = Operation.make({
     typename: Schema.String.annotations({
       description: 'The typename of the schema in the format of "com.example.type.type".',
     }),
-    jsonSchema: Schema.Any,
+    // Unconstrained tool parameters are serialized to the LLM without a `type`, so some models
+    // emit the JSON Schema as a JSON-encoded string rather than an object. Accept either form,
+    // decoding a string back into an object, and reject anything that is not an object at the
+    // tool-call boundary, since `makeObjectFromJsonSchema` requires an object.
+    jsonSchema: Schema.Union(Schema.parseJson(Schema.Object), Schema.Object).annotations({
```
dmaretskyi: do Schema.Record<String, Any> to force model to emit and object

dmaretskyi: Done in c344c0f4f — `jsonSchema: Schema.Record({ key: Schema.String, value: Schema.Any })`, which advertises `type: object` to the model. Decoded value stays assignable to `makeObjectFromJsonSchema` without a cast.

Heads-up: this changes the `add-schema` tool's serialized `inputSchema`, which is embedded in the database skill's memoized conversations. So `src/skills/database/skill.conversations.json` needs regenerating (`ALLOW_LLM_GENERATION=1`) — until then the ~17 memoized LLM tests in `skill.test.ts` fail with "No memoized conversation found" (that's the current red `test` job). I can't run the regen in this sandbox (no edge AI-service egress and the moon TS toolchain download is blocked here). Could you regenerate the cache, or let me know if there's an env where I should?

The new deterministic, non-LLM tests (object-only decode contract + schema-creation assertion) pass independently of the cache.

---

PR #12099, opened 2026-07-05
feat: Chess.com plugin; support feed objects in the graph

dmaretskyi:
## Summary by CodeRabbit

* **New Features**
  * Added Chess.com integration for Composer, including account creation from a username, syncing archived games, browsing synced games as cards, and actions to resync or clear imported games.
  * Expanded chess support with player review and position indexing features to help organize and rebuild game positions.

* **Bug Fixes**
  * Improved game and object loading so synced items and feed-backed content display more reliably in the app.


dmaretskyi:
Addressed the CodeRabbit review feedback in `8776e91c7` (on top of a merge with `main`, which also resolves the conflict):

- **ChessGameArticle** – the game actions button now uses a translated `game-actions.label` instead of a hardcoded string.
- **clear-synced-games** – synced `Game` and their `Chess.State` variants are now removed from the feed via `Feed.remove` before the old feed is dropped, so they are no longer orphaned.
- **chess-com-api** – added a 30s request timeout on `getJson`; timeouts are mapped to `ChessComRequestError` by the existing catch-all.
- **ChessComAccount.applyProfile** – now skips `undefined` fields so a transient API omission no longer wipes previously-synced profile data.
- **position-index** – FEN-only games now load the FEN into the board before replaying, and `loadPgn`/`load` are guarded so a malformed record returns `[]` instead of aborting the rebuild.
- **rebuild-position-index** – only `GameVariantMismatchError` is skipped now; database/ref failures surface instead of silently producing a partial index.

CI (build, check, test, storybook) is green.

---

PR #12130, opened 2026-07-08
feat: Grouped queries 

dmaretskyi:
## Summary by CodeRabbit

* **New Features**
  * Added `groupBy` support to queries, producing grouped results with group keys, per-group `count`, and grouped `values`.
  * `groupBy` is available in the query builder, pretty-printed queries, and React `useQuery`.
  * Grouped queries execute with consistent caching and reactive updates (including stable result identity).
* **Bug Fixes**
  * Improved interaction with `orderBy` (correct group ordering and within-group ordering).
  * Fixed `limit`/`skip` to paginate over whole groups rather than individual items.
  * Fixed group membership updates when records change or groups become empty.
* **Tests**
  * Added end-to-end and planner/executor coverage for `groupBy`.


packages/core/echo/echo-client/src/client/index-query-source-provider.ts
```diff
@@ -463,3 +465,9 @@ let nextQueryId = 1;
  * Keyed by the type DXN.
  */
 const emittedSchemaValidationWarnings = new Set<string>();
+
+/**
+ * Builds the `Entry.group` metadata from a wire record; present iff the query has a `groupBy` clause.
+ */
+const _groupFromRemoteResult = (result: RemoteQueryResult): QueryResult.EntityEntry['group'] =>
+  result.groupKey !== undefined ? { key: JSON.parse(result.groupKey), count: result.groupCount ?? 0 } : undefined;
```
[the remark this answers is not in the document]
wittjosiah: Fixed in be58bf2254 — the fallback now uses `?? 1` to match the working-set source (a present record implies at least one member).

packages/core/echo/echo-host/src/query/query-planner.ts
```diff
@@ -659,6 +662,50 @@ export class QueryPlanner {
...
+        foundElsewhere = true;
+      }
+    });
+    if (foundElsewhere) {
+      throw new QueryError({
+        message: 'groupBy must be the outermost query clause — only from()/options() may follow it',
+        context: { query },
+      });
+    }
+  }
```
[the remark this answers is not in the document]
wittjosiah: Good catch — this was only partially handled. Fixed properly in be58bf2254: `_validateGroupByPlacement` now counts every group-by in the tree (`QueryAST.visit` recurses through `from` sources), so a grouped `.from(subquery)` source — whether or not the outer query is grouped — is rejected before flattening. Added regression tests for both a grouped from() source and a double-grouped (outer + subquery) query.

packages/core/echo/echo-client-e2e/src/query.test.ts
```diff
@@ -368,6 +369,314 @@ describe('Query', () => {
...
+      await db.flush();
+
+      const groups = await db
+        .query(Query.select(Filter.everything()).groupBy(GroupKey.property('category'), GroupKey.property('value')))
+        .run();
+
+      expect(groups).to.have.length(3);
+      expect(
+        groups
+          .map((group: any) => group.count)
```
dmaretskyi: why :any?

packages/core/echo/echo-host/src/query/plan.ts
```diff
@@ -278,4 +279,104 @@ export namespace QueryPlan {
...
+   * Groups become contiguous, ordered by the first occurrence of their key in the incoming
+   * (already-ordered) working set — this lets a preceding `OrderStep` also control group order.
+   */
+  export type GroupByStep = {
+    _tag: 'GroupByStep';
+
+    keys: readonly QueryAST.GroupByKey[];
+  };
+
+  export const GroupByStep = Object.freeze({
```
dmaretskyi: this module is the declarative description of a query plan. implementation should go into the respective executor

packages/core/echo/echo-client-e2e/src/query.test.ts
wittjosiah: Removed — grouped query results are now typed as `Group<K, T>[]`, so these annotations were redundant. Dropped the `: any` on the group/object callbacks and typed the reactive test helper as `<T>(query: QueryResult.QueryResult<T>) => Promise<T[]>`. Fixed in 5d086d173c.

packages/core/echo/echo-host/src/query/plan.ts
wittjosiah: Agreed. Moved the runtime helpers (coerce/serialize/partition/take/drop) into a new execution module `query/group-by.ts` (`GroupBy`), re-exported via the `@dxos/echo-host/query` subpath so the host `QueryExecutor` and client `WorkingSetQueryExecutor` share one implementation. `plan.ts` now keeps only the declarative `GroupByStep` type. Fixed in 5d086d173c.

PR #12131, opened 2026-07-08
feat(echo): add access.spaceId field to automerge documents

dmaretskyi:
## Summary

Automerge `DatabaseDirectory` documents identified their owning space only via `access.spaceKey`. This PR adds a new `access.spaceId` field, deprecates `access.spaceKey`, and coalesces the two on read.

### Schema (`@dxos/echo-protocol`)

- `DatabaseDirectory.access` gains an optional `spaceId` field (multibase base-32 space ID).
- `access.spaceKey` is marked `@deprecated` (and is now optional in the type).
- New coalescing reader `DatabaseDirectory.getSpaceId(doc)`: prefers `access.spaceId`, and for documents that predate it derives the id from the legacy `access.spaceKey` / `experimental_spaceKey` fields via `createIdFromSpaceKey` (async, since the derivation is a SHA-256 hash).
- `DatabaseDirectory.make` accepts `spaceId` (with `spaceKey` kept as a deprecated option).

### Writers — now stamp `access.spaceId`

- `EchoHost.createSpaceRoot` (new space root documents).
- `EntityManager._createDocumentForObject` (new linked object documents) and `_initDocAccess` (stamping on load; the guard now also upgrades documents missing `spaceId`).
- `DataSpace._onNewAutomergeRoot` (legacy root stamping now fills in both identifiers).
- `MigrationBuilder` (epoch-migration documents).

**Note:** the deprecated `spaceKey` is still written alongside `spaceId`. The id is a one-way hash of the key, so older clients — and code paths that need the space *public key*, e.g. share-policy resolution in `AutomergeHost._getContainingSpaceForDocument` — cannot work from `spaceId` alone. Dropping the key write is a follow-up once those paths are id-based.

### Readers

- The indexer data source (`AutomergeDataSource`) now uses the coalescing `DatabaseDirectory.getSpaceId` instead of deriving the id from the key itself.
- `DatabaseDirectory.getSpaceKey` is marked deprecated, pointing consumers to `getSpaceId`.

### Tests

- `automerge-data-source.test.ts`: fixtures now write `access.spaceId` (they were previously stuffing a `SpaceId` into `spaceKey`); new cases cover deriving the id from legacy `spaceKey`-only documents and `spaceId` taking precedence when both are present.
- `entity-manager.test.ts`: asserts the root document is stamped with the database's `spaceId`.
- `@dxos/echo` test fixtures (`test-data.ts`) create directories with `spaceId`.

## Testing

- `moon run echo-protocol:build echo-host:build echo-client:build echo:build client-services:build migrations:build` — green.
- `moon run echo-host:test` (253 passed), `echo-client:test` (466 passed), `echo:test` (437 passed), `client-services:test` (123 passed), `migrations:test` — all green.
- Lint clean on all touched packages.

https://claude.ai/code/session_01A5QMTuxgDzQFHgjQRZaquN

---
_Generated by [Claude Code](https://claude.ai/code/session_01A5QMTuxgDzQFHgjQRZaquN)_

## Summary by CodeRabbit

* **New Features**
  * Added support for a primary space identifier (`spaceId`) in document access metadata, while continuing to support the legacy key-based (`spaceKey`) format.

* **Bug Fixes**
  * Space access metadata is now consistently initialized and backfilled on root and object documents.
  * Document change detection now recognizes space identifiers when either is present, preferring `spaceId` when available.
  * Compaction and migrations now preserve/restamp space identifier info on recreated documents.

* **Tests**
  * Expanded automated coverage for legacy and `spaceId`/`spaceKey` interoperability, including additional derivation and compaction assertions.

* **Documentation**
  * Clarified guidance around stamping space access metadata in developer-facing comments.

### DIFF
--- a/packages/core/echo/echo-protocol/src/document-structure.ts
+++ b/packages/core/echo/echo-protocol/src/document-structure.ts
@@ -31,9 +31,9 @@ export interface DatabaseDirectory {
 
   access?: {
     /**
-     * ID of the space that owns the document (multibase base-32 encoded).
+     * ID of the space that owns the document.
      */
-    spaceId?: string;
+    spaceId?: SpaceId;
 
     /**
      * @deprecated Use {@link spaceId}. Still written alongside `spaceId` so older clients
@@ -71,11 +71,9 @@ export const DatabaseDirectory = Object.freeze({
    * `experimental_spaceKey`), deriving the id from the key for documents that predate `spaceId`.
    */
   getSpaceId: async (doc: DatabaseDirectory): Promise<SpaceId | null> => {
-    // String(...) to handle RawString.
-    const rawSpaceId = doc.access?.spaceId != null ? String(doc.access.spaceId) : null;
-    if (rawSpaceId != null) {
-      invariant(SpaceId.isValid(rawSpaceId), 'Invalid space ID');
-      return rawSpaceId;
+    if (doc.access?.spaceId != null) {
+      invariant(SpaceId.isValid(doc.access.spaceId), 'Invalid space ID');
+      return doc.access.spaceId;
     }
 
     const spaceKeyHex = DatabaseDirectory.getSpaceKey(doc);
--- a/packages/core/echo/echo-protocol/src/document-structure.ts
+++ b/packages/core/echo/echo-protocol/src/document-structure.ts
@@ -3,13 +3,14 @@
 //
 
 import { invariant } from '@dxos/invariant';
-import type { EntityId, URI } from '@dxos/keys';
+import { type EntityId, PublicKey, SpaceId, type URI } from '@dxos/keys';
 import { visitValues } from '@dxos/util';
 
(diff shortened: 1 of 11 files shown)


packages/core/echo/echo-protocol/src/document-structure.ts
```diff
@@ -29,7 +30,19 @@ export interface DatabaseDirectory {
   version?: SpaceDocVersion;
 
   access?: {
-    spaceKey: string;
+    /**
+     * ID of the space that owns the document (multibase base-32 encoded).
+     */
+    spaceId?: string;
```
dmaretskyi: use SpaceId type from keys

dmaretskyi: Fixed in 5db62cda — `access.spaceId` is now typed as `SpaceId`.

---

PR #12143, opened 2026-07-09
refactor(echo): fold groupBy into aggregate, flatten group results

wittjosiah:
## Summary

Cleans up the query aggregate DSL introduced in #12137. Grouping is now folded into `.aggregate()` via a new `Aggregate.group(...)`, and results are flat records instead of a `Group<K, E>` wrapper.

```ts
// before
Query.type(Message)
  .orderBy(Order.property('created', 'desc'))
  .groupBy(GroupKey.property('threadId'))
  .aggregate({ lastMessageAt: Aggregate.max('created'), items: Aggregate.items() })
  .orderBy(Order.property('lastMessageAt', dir));

// after
Query.type(Message)
  .orderBy(Order.property('created', 'desc'))
  .aggregate({
    threadId: Aggregate.group('threadId'),
    lastMessageAt: Aggregate.max('created'),
    items: Aggregate.items({ limit: 20 }),
  })
  .orderBy(Order.property('lastMessageAt', dir));
```

- Removes `GroupKey` (module + AST type), `Query.Group`, and `.groupBy()`.
- Renames the `group-by` AST/plan node to `aggregate` throughout the host (`echo-host`) and client (`echo-client`) executors; zero `group` entries aggregates the whole input into a single row.
- `Aggregate.items({ limit })` caps the exposed member array per group at client result assembly — `count` still reflects the full group size, so callers get both a bounded preview and the true total.
- Adds a `Query.AggregateResult` phantom brand so `useQuery`/`usePagination` return the flat aggregate record un-wrapped instead of coercing it through `Entity.Entity`.
- Mirrors the change in the sandboxed `echo-query` "lite" DSL reimplementation.
- Updates the mailbox's threaded-conversation query to declare `Aggregate.count()` alongside a 4-message `Aggregate.items({ limit })` preview; `MessageStack` now surfaces the full thread size as a "+N more" affordance.

## Test plan

- [x] `moon run echo:test` — DSL/type-level/pretty tests in `Query.test.ts`
- [x] `moon run echo-host:test` — `query-planner.test.ts`, `query-invalidation.test.ts`
- [x] `moon run echo-client:test` — `working-set-executor.test.ts`
- [x] `moon run echo-client-e2e:test -- src/query.test.ts` — end-to-end aggregate behavior incl. items-limit-vs-count, aggregate-everything, renamed group field ordering
- [x] `moon run echo-react:test` — `useQuery.test.tsx`, `usePagination.test.tsx`
- [x] `moon run echo-query:test` — sandbox pretty-printer parity
- [x] `moon run plugin-inbox:test`
- [x] `moon run :lint -- --fix` across all touched packages
- [x] Cast audit (`git diff origin/main | grep -nE '\bas (any|unknown|[A-Z])|as unknown as'`) — no new casts
- [ ] Manual check of the mailbox "+N more" affordance in the Composer app

## Summary by CodeRabbit

* **New Features**
  * Query grouping now uses a streamlined **aggregate** syntax, with grouped key fields returned directly in results (flat aggregate records).
  * Inbox conversation previews now include a capped message list plus a **“+ more”** indicator, along with the conversation’s full total size.

* **Bug Fixes**
  * Improved grouped pagination and ordering: stable group boundaries, correct skip/limit semantics, and consistent ordering across reactive updates.
  * Fixed grouped/aggregated result shapes in query and UI layers so grouped fields (and counts/items) appear reliably.


packages/core/echo/echo/src/Query.ts
```diff
@@ -49,16 +48,14 @@ type ReferenceTraversalTarget<P> = P extends Ref.Unknown
...
+ * Phantom brand on the flat row produced by {@link Query.aggregate}. Present only at the type level
+ * (never at runtime), it lets hooks like `useQuery`/`usePagination` distinguish an aggregate-row
+ * query from an entity query and avoid wrapping the row in `Entity.Entity`. The brand is a required
+ * property so `T extends AggregateResult` discriminates — an optional one would be satisfied by any
+ * type. Consumers never read it.
  */
-export interface Group<K, E> {
-  readonly key: K;
+export interface AggregateResult {
+  readonly '~AggregateResult': true;
```
dmaretskyi:
```suggestion
  readonly '~@dxos/echo/Query.AggregateResult': true;
```

packages/core/echo/echo/src/Query.ts
```diff
@@ -149,60 +146,44 @@ export interface Query<T> {
...
-   * a {@link groupBy}: before, by member properties; after, by the group's fields (its `key` and any
-   * aggregates declared via {@link aggregate} — e.g. `Order.property('lastMessageAt')` reorders the
-   * groups by that aggregate).
+   * an {@link aggregate}: before, by member (row) properties; after, by the flat record's fields
+   * (any group or aggregate field — e.g. `Order.property('lastMessageAt')` reorders the groups by
+   * that aggregate).
    * @param order - Order to sort the results.
    * @returns Query for the ordered results.
    */
   'orderBy'(...order: EffectArray.NonEmptyArray<Order.Order<T>>): Query<T>;
```
dmaretskyi, line 154: make sure Order.aggergate is removed in favor of Order.property as aggregate props are just normal props

packages/core/echo/echo/src/Query.ts
```diff
@@ -149,60 +146,44 @@ export interface Query<T> {
...
    * ```
    *
+   * Must be the last data-selecting clause in the chain — only `from`/`options`/`orderBy`/`limit`/
+   * `skip` may follow.
    * @param aggregates - Record of aggregate declarations keyed by result field name.
-   * @returns Query whose group results carry the named aggregates as fields.
+   * @returns Query whose flat result records carry the named aggregates as fields.
    */
-  'aggregate'<const A extends Record<string, Aggregate.Aggregate<T extends Group<any, infer E> ? E : never, any>>>(
+  'aggregate'<const A extends Record<string, Aggregate.Aggregate<T, any>>>(
```
dmaretskyi, line 183: are the prop names for aggregation typed?

packages/core/echo/echo/src/Query.ts
```diff
@@ -149,60 +146,44 @@ export interface Query<T> {
...
     aggregates: A,
-  ): Query<
-    EffectTypes.Simplify<
-      (T extends Group<infer K, infer E> ? Group<K, E> : Group<unknown, unknown>) & {
-        readonly [N in keyof A]: Aggregate.ValueOf<A[N]>;
-      }
-    >
-  >;
+  ): Query<EffectTypes.Simplify<AggregateResult & { readonly [N in keyof A]: Aggregate.ValueOf<A[N]> }>>;
```
dmaretskyi:
```suggestion
  ): Query<Aggregate.AggregationResult<A>>;
```

packages/core/echo/echo-query/src/query-lite/query-lite.ts
```diff
@@ -69,26 +68,6 @@ namespace Order1 {
 const Order2: typeof Order$ = Order1;
 export { Order2 as Order };
 
-class GroupKeyClass implements GroupKey$.Any {
```
dmaretskyi, line 72: can this module be killed? Rich removed quickjs afaik

packages/core/echo/echo-protocol/src/query/ast.ts
```diff
@@ -420,53 +420,43 @@ export interface QuerySkipClause extends Schema.Schema.Type<typeof QuerySkipClau
...
+ *   Composite keys are formed from multiple `group` entries. A query with no `group` entries
+ *   aggregates its entire input into a single row.
  * - `max`/`min` reduce a scalar member `property`.
- * - `items` collects the group's members; `count` yields the member count. Both are opt-in — a group
- *   carries neither its members nor a count otherwise. `property` is unused for `items`/`count`.
+ * - `items` collects the group's members, optionally capped to `limit`; `count` yields the member
+ *   count. Both are opt-in — a row carries neither its members nor a count otherwise. `property` is
+ *   unused for `items`/`count`; `limit` is used only by `items`.
  */
 const GroupAggregate_ = Schema.Struct({
```
dmaretskyi: convert into tagged union for each kind to avoid optional + unused fields

packages/core/echo/echo/src/Query.ts
wittjosiah: Done — renamed to `'~@dxos/echo/Query.AggregateResult'` in 65b6eaa.

packages/core/echo/echo/src/Query.ts
wittjosiah: Confirmed — `Order.aggregate` never existed and no remnants of the concept remain. Ordering by an aggregate/group field goes through `Order.property(name)` (the field is just a normal prop on the flat result row), which is exercised by the "a group field named differently from its property is orderable by that name" test in echo-client-e2e.

packages/core/echo/echo/src/Query.ts
wittjosiah: Two different things here, both typed:
- Each aggregate's *input* property (e.g. `Aggregate.max('created')`) is checked against the query's element type via `<T, K extends keyof T & string>(property: K)`.
- The *output* field name (the record key you pick, e.g. `{ latest: Aggregate.max('created') }`) is intentionally an arbitrary string — renaming the result field is the whole point of the DSL (see the "a group field named differently from its property is orderable by that name" test). It's typed as `keyof A` in the mapped result type, just not constrained to anything else.

Let me know if you meant something more specific and I'll follow up.

packages/core/echo/echo/src/Query.ts
wittjosiah: Done — moved the computation into `Aggregate.AggregationResult<A>` (in Aggregate.ts) and the signature is now `Query<Aggregate.AggregationResult<A>>` in 65b6eaa.

packages/core/echo/echo-query/src/query-lite/query-lite.ts
wittjosiah: Checked — quickjs is still live: `echo-query/src/sandbox/quickjs.ts` + `quickjs.test.ts`, the `quickjs` dependency in `echo-query/package.json`, and real consumers of this sandbox DSL (`react-ui-components`' QueryEditor, `plugin-script`, `plugin-explorer`, `plugin-inbox`'s match-filter, `stories-assistant`). Kept the module as-is rather than removing it — happy to revisit if I'm missing context on a removal that's in flight elsewhere.

packages/core/echo/echo-protocol/src/query/ast.ts
wittjosiah: Done — split into a tagged union per kind (`group`/`max`/`min` require `property`, `items` has an optional `limit`, `count` has neither) in 65b6eaa. This also let me drop the non-null assertions on `aggregate.property!` that the shared-optional-field version needed at every read site (group-by.ts, both executors, query-result.ts, both pretty-printers) — the discriminated union narrows them for free.

PR #12149, opened 2026-07-09
feat(react-ui-form): reactive companion form + table external-mutation reactivity

wittjosiah:
## Summary

Makes the selected-objects companion form reactive to external/remote mutations while letting it hold in-progress (including invalid) edits without snapping back, and fixes the table so it reflects mutations made outside the grid. ECHO React reactivity is atom-based (`useObject`/`Obj.atom`), so a render-time read of an object does **not** subscribe — several places relied on that and weren't actually reactive.

### Reactive companion form (`react-ui-form`)
- `useFormHandler` reworked to a **sparse-overrides** model: un-edited fields reflect the live `values` source; edits are held as per-path overrides (invalid values included) until the source catches up, then pruned. Replaces the `useControllableState` reset-on-any-change behaviour that caused invalid input to snap back.
- `ObjectForm` subscribes via `useObject` and passes the reactive snapshot as `values`, so external mutations flow into the form.
- New stories + play tests: `Form/ReactiveBuffered` and `ObjectForm/Reactive` (seeded background mutations via `@dxos/random`, live JSON panel, storybook play asserting reactivity).
- Removed the now-unused `@radix-ui/react-use-controllable-state` dependency.

### `Obj.valuesEqual` is no longer public (`echo`)
- Made `Obj.valuesEqual` module-private again (it's an internal detail; still used by `updateFrom`). The form keeps a small local copy for override reconciliation, with a `TODO` to factor a shared util.
- A spike showed Effect's `Schema.equivalence` is **not** a safe substitute — it returns false-positive equality for dynamic/union/ref-array schemas, which silently pruned edits (broke the `ObjectProperties` create flows). `Ref` fields, however, compare correctly through it by URI with no annotation.

### Table reflects external row mutations (`react-ui-table`)
- **Root cause:** the per-row `Obj.subscribe` that bumps the cell-update counter was wired to the model's `_visibleRange` atom, but nothing ever set it — the grid's range was only reported to the presentation's (write-only) copy. So no row was ever subscribed and external mutations never repainted.
- Fix: `TableModel.setVisibleRange` (guarded), driven from `TablePresentation.getCells`; re-subscribe visible rows when the row set changes too (rows load async after the range is first reported); removed the presentation's dead `_visibleRange`.
- New `Table/ExternalMutation` story + play test reproducing the issue (mutate a row outside the grid, assert the cell updates).

## Verification
- `react-ui-form`: unit 64, storybook 60 (incl. `Reactive`).
- `react-ui-table`: unit 21, storybook 14 (incl. `ExternalMutation`).
- `echo`: 451 unit.
- Build + lint clean across affected packages; merged `origin/main` (resolved `ObjectCardStack` + lockfile).

## Summary by CodeRabbit

* **New Features**
  * Forms now stay synchronized with external updates while preserving in-progress edits and invalid intermediate values.
  * Added interactive examples demonstrating reactive forms and externally changing table rows.

* **Bug Fixes**
  * Improved table updates so visible cells refresh reliably after data changes, sorting, or external mutations.
  * Refined card layout, spacing, scrolling, and disabled unintended dragging.

* **Tests**
  * Added coverage for reactive form buffering, reconciliation, and concurrent updates.


packages/ui/react-ui-form/src/components/ObjectForm/ObjectForm.stories.tsx
```diff
@@ -60,3 +71,83 @@ export default meta;
...
+    return () => clearInterval(interval);
+  }, [running, person, mutate]);
+
+  if (!person) {
+    return <></>;
+  }
+
+  return (
+    <TestLayout json={snapshot}>
+      <div className='flex flex-col gap-2'>
```
richburdon, line 118: pass classNames to TestLayout?

richburdon (commented):
looks good -- i appreciate the test

PR #12180, opened 2026-07-13
fix: WebSocket message ordering & PostHog cross-subdomain cookie probe

wittjosiah:
## Summary

Two independent fixes surfaced while running Composer on a `*.pages.dev` preview deploy.

### `fix(edge-client)`: preserve WebSocket message arrival order during reassembly
WebSocket frames arrive in order, but converting frame data to bytes is async on the `Blob` fallback path (`blob.arrayBuffer()`), and concurrent conversions are not guaranteed to complete in arrival order. `WebSocketMuxer` segmented-message reassembly requires chunks to reach `receiveData` in arrival order, so out-of-order completion corrupts multi-chunk messages.

- Chain each message onto a `_receiveChain` promise so bytes reach `receiveData` in arrival order; a single bad message is caught and dropped rather than stalling the queue.
- Set `binaryType = 'arraybuffer'` so bytes are available synchronously, avoiding the async `blob.arrayBuffer()` read entirely on modern paths; `toUint8Array` gains an `ArrayBuffer` branch with the `Blob` path kept as fallback.
- Adds `edge-ws-connection.test.ts` covering both the out-of-order `Blob` race and the `ArrayBuffer` path.

### `fix(observability)`: disable PostHog cross-subdomain cookie probe
The PostHog `dmn_chk_*` cross-subdomain cookie probe fails on public-suffix hosts (e.g. `*.pages.dev` preview deploys) and spams the console. Set `cross_subdomain_cookie: false` to scope cookies to the exact host.

## Testing
- `moon run edge-client:test` — 35 passed (incl. new suite)
- `moon run edge-client:lint observability:lint` — clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebSocket receive handling to preserve ordered, contiguous segmented-message reassembly when frames are multiplexed and delivered asynchronously.
  - Standardized binary payload decoding for WebSocket messages provided as `ArrayBuffer` or `Blob`.
  - Ensured receive-path failures and malformed messages don’t block subsequent message processing.
  - Reduced PostHog console warnings by preventing cross-subdomain cookie probing.
- **Tests**
  - Added Vitest coverage for segmented protobuf message reassembly for both `ArrayBuffer` and `Blob` frame delivery.

### DIFF
--- a/packages/core/mesh/edge-client/src/edge-ws-connection.ts
+++ b/packages/core/mesh/edge-client/src/edge-ws-connection.ts
@@ -184,16 +184,17 @@ export class EdgeWsConnection extends Resource {
         return;
       }
 
-      // Serialize processing so bytes reach `muxer.receiveData` in arrival order. The lock
-      // releases even if `_receiveMessage` throws, so a single bad message is logged and
-      // dropped instead of stalling every message queued after it.
-      void this._receiveMutex
-        .executeSynchronized(() => this._receiveMessage(event.data, muxer))
-        .catch((err) => log.catch(err));
+      // `_receiveMessage` serializes on `_receiveMutex`; `acquire` enqueues synchronously,
+      // so locks are taken in arrival order regardless of async conversion timing.
+      void this._receiveMessage(event.data, muxer).catch((err) => log.catch(err));
     };
   }
 
   private async _receiveMessage(data: WebSocket.Data, muxer: WebSocketMuxer): Promise<void> {
+    // Serialize processing so bytes reach `muxer.receiveData` in arrival order. The guard
+    // releases on scope exit even if processing throws, so a single bad message is logged
+    // and dropped instead of stalling every message queued after it.
+    using _guard = await this._receiveMutex.acquire();
     const bytes = await toUint8Array(data);
     this._recordBytes(0, bytes.byteLength);
     if (!this.isOpen) {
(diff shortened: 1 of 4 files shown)


packages/core/mesh/edge-client/src/edge-ws-connection.ts
```diff
@@ -170,23 +183,34 @@ export class EdgeWsConnection extends Resource {
...
-      if (!this.isOpen) {
-        return;
-      }
 
-      this._messagesReceived++;
+      // Chain each message onto the previous one so bytes reach `muxer.receiveData` in
+      // arrival order. Catching within the chained callback (rather than on
+      // `_receiveChain` itself) keeps `_receiveChain` resolved, so a single bad message
+      // is logged and dropped instead of stalling every message queued after it.
+      this._receiveChain = this._receiveChain.then(() =>
```
mykola-vrmchk: Not a fan of this pattern — we don't use it anywhere else in the codebase. If we could make _receiveMessage synchronous, we wouldn't need this workaround. Otherwise, might be cleaner to rewrite using locks syntax, which fits the codebase.

wittjosiah: Good call — rewrote it to use `Mutex.executeSynchronized` from `@dxos/async` (the codebase-standard lock) instead of the hand-rolled chain. Fully synchronous processing isn't possible while the async `Blob` fallback path in `toUint8Array` remains, so serialization is still needed there. Done in 4cd0c316ec.

wittjosiah: Switched to `using _guard = await this._receiveMutex.acquire()` inside `_receiveMessage` (matches the pattern in echo-host's edge replicators). Ordering still holds since `acquire` enqueues synchronously. 575fbbcaf9.

PR #12191, opened 2026-07-13
apps: convert todomvc & tasks to dedicated worker; fix iframe shell over effect-rpc

dmaretskyi:
## Summary

`todomvc` and `tasks` were broken on `main` — they still used the **shared-worker** services setup, and were completely unable to start after the effect-rpc worker migration (#12127).

This PR:

1. **Converts both apps to the dedicated-worker services mode** (coordinator `SharedWorker` + dedicated `Worker`), matching `composer-app` and the `persistentClientServices` preset. Each app now builds services via `createClientServices(..., { createDedicatedWorker, createCoordinatorWorker })` with `servicesMode: DEDICATED_WORKER`, referencing the prebuilt `@dxos/client/dedicated-worker` and `@dxos/client/coordinator-worker` entrypoints. The obsolete `shared-worker.ts` files are removed.

2. **Fixes the iframe shell (`shell='./shell.html'`) over effect-rpc.** While verifying, both apps hung on startup. Root cause: #12127 migrated the shell's `ClientServicesProxy` consumer to effect-rpc but left the app-side shell serving (`Client._shellClientProxy`) on the removed protobuf peer (`createProtoRpcPeer`). The mismatched channel flooded the protobuf decoder with `invalid wire type` errors and `client.initialize()` never completed, so nothing rendered.
   - `Client` now re-serves its services to the shell iframe over effect-rpc via `ClientRpcServer` + `makeHandlersFromRpc(this._services.rpc)` (mirroring the worker session), and closes it on teardown (the old proxy was never closed).
   - `runShell` provides the parent origin (`window.location.origin`) to the client-services iframe port. The old protobuf peer sent an `open` frame from both sides, which bootstrapped the port's origin; the effect-rpc server is passive, so without an upfront origin the shell's client can never post its first frame and the connection deadlocks (30s `RemoteServiceConnectionTimeout`). The shell is served same-origin with its host, so `window.location.origin` is correct.

   This path is only active for apps that embed the external iframe shell (todomvc, tasks); `composer-app` uses the in-process shell plugin and is unaffected.

## Verification

Driven end-to-end against the pre-installed Chromium (the sandbox has no WebSocket connectivity — see below):

- **Build/typecheck/lint** pass for `client`, `shell`, `todomvc`, `tasks`. Dedicated + coordinator worker chunks emit correctly in the production bundle.
- **todomvc boots fully** — `add-button`, `share-button`, `join-button`, `list`, `new-todo` render, with zero RPC decode errors (was: only `dxos-shell`, continuous `invalid wire type` flood, no `new-todo`).
- **The shell iframe renders** its full space-sharing panel (`spaces-panel.create-invitation`, `copy-invitation`, …) — previously its `#root` stayed empty and its client timed out.
- **tasks boots** with no RPC flood (empty home is expected — it doesn't auto-create a space).

The two-peer invitation e2e tests (`basic.spec.ts`) progress through invitation creation, auth-code generation, and guest acceptance, then stall at the guest↔host peer connection. This is **environmental**: the sandbox browser cannot open any WebSocket (both the edge signaling server and a generic public echo server time out at 12s), and the swarm requires WebSocket signaling. These tests are expected to pass in CI where WebSocket/WebRTC connectivity is available.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FoSE3pDW2EeyUPZrQgq4jF)_

## Summary by CodeRabbit

- **New Features**
  - Improved RPC bridging for apps embedded in the external shell iframe.
  - Updated Tasks and TodoMVC to run client services in a dedicated-worker model with a coordinating shared worker.
- **Bug Fixes**
  - Fixed startup hangs in apps using the external shell iframe.
  - Prevented effect-RPC connection deadlocks by correctly providing parent origin for iframe communication.
  - Enhanced shell service (re)connection so services are re-served during app startup.


dmaretskyi:
Re: the open/close race finding — fixed `serveOverProtocol` (the helper this PR adds) with an `openPromise` guard in `25e9ec7d`: concurrent opens now share one init, and `close()` awaits an in-flight open before disposing.

I left the pre-existing `serve` and `ClientRpcServer` twins untouched: they're worker-framework code outside this PR's scope, and in practice `Client._open`/`destroy` are `@synchronized` so their open/close is serialized (the race isn't reachable through the client). Happy to apply the same guard to them uniformly if you'd prefer it handled here.

---

PR #12193, opened 2026-07-13
client-services: replace ServiceContext class with effect-native lifecycle

dmaretskyi:
## Summary

Now that the client service components are composed as Effect layers, the monolithic `ServiceContext` `Resource` class was redundant: its constructor took a bag of already-constructed components and cached them as fields, and a `serviceContextServiceLayer` existed purely to resolve ~20 tags and pass them in.

This removes that class and reshapes the orchestration to be effect-native:

- **Lifecycle moved out of the class into standalone functions** (`service-lifecycle.ts`). `open`/`close`, `createIdentity`/`_acceptIdentity`, `_setNetworkIdentity`, `_initialize`, storage-version check, profile broadcast, and the readiness gates are now plain functions that **resolve each component from its Effect tag on demand** (`runtime.runPromise(SpaceManagerService)`, `Effect.serviceOption(...)` for optionals) instead of reading cached `this.*` fields.
- **`ClientServicesHost` drives the lifecycle** and provides only the surface the RPC handlers need — `createIdentity`, `broadcastProfileUpdate`, readiness gates — via a slim new `ClientLifecycleService` tag. The `serviceContextServiceLayer` and `ServiceContextService` tag are gone.
- **`ServiceContext` is now an interface** (a handle over the resolved components) assembled by `makeServiceContext`, retained for the legacy `@deprecated` devtools, diagnostics, and test consumers so their surface is unchanged. `instanceof ServiceContext` in the invitation test utils becomes a structural check.
- `service-context.ts` keeps only the layer composition (`ServiceContextLayer`, storage/echo-host/replicator layers) and now exposes the component tags via a `ServiceContextComponents` union.

Behavior is preserved: the open/close ordering, identity flow, and readiness semantics are copied verbatim from the old class — only the source of each service (tag vs. cached field) changed.

## Scope note

This is the first of the requested changes (remove `ServiceContext`, move code up to the host, access tags instead of fields). The follow-up — turning the host's still-imperative construction of the signal manager, transport factory, and edge clients into Effect tags + layers — is intentionally left for a separate PR to keep this one reviewable.

Known minor follow-up: the `@trace.span()` lifecycle spans that decorated the old class methods are not reproduced on the standalone functions (decorators are class-only); tracing of the open/initialize sequence can be re-added separately if desired.

## Test plan

- `moon run client-services:build` — green
- `moon run client:build client-e2e:build` — green (downstream consumers of the `ServiceContext` handle)
- `moon run client-services:test` — 141 passed / 2 skipped, incl. `service-context`, device/space invitation, identity/spaces/devices/network service, and `service-host` tests
- `moon run client-services:lint` — clean; `oxfmt` applied

---
_Generated by [Claude Code](https://claude.ai/code/session_019oNWG13aYDMwR2h4nZBrEM)_

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated client lifecycle surface to drive service open/close and identity initialization, including improved readiness gates for data spaces and edge agents.
  * Added Effect-layer factories for edge connection, edge HTTP client, signal manager, swarm network manager, and WebRTC transport.
  * Enabled exported storage-migration support as part of the client lifecycle flow.
* **Bug Fixes**
  * Improved invitation routing reliability by using a robust service-context discriminator.
* **Documentation / Refactor**
  * Refined devtools, spaces, metadata, and diagnostics APIs to use narrower context surfaces instead of broader service context.


packages/sdk/client-services/src/packlets/services/service-lifecycle.ts
```diff
@@ -0,0 +1,569 @@
...
+  ClientLifecycleService,
+  ClientLifecycle
+>() {}
+
+/**
+ * Handle over the composed component stack, extending {@link ClientLifecycle} with direct access to
+ * the resolved components. Retained for the legacy devtools / diagnostics / test consumers; the
+ * production RPC path depends only on {@link ClientLifecycle}.
+ */
+export interface ServiceContext extends ClientLifecycle {
```
dmaretskyi, line 202: break down legacy consumers to only depend on required components

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -271,45 +311,82 @@ export class ClientServicesHost {
...
+    const transportFactoryLayer = this._transportFactoryOverride
+      ? Layer.succeed(TransportFactoryService, this._transportFactoryOverride)
+      : RtcTransportFactoryLayer({
+          webrtcConfig: { iceServers: config.get('runtime.services.ice') },
+          iceProviders: config.get('runtime.services.iceProviders'),
+        });
 
-    const endpoint = config?.get('runtime.services.edge.url');
     if (endpoint) {
       const clientTag = resolveTelemetryTag(config);
```
dmaretskyi, line 390: use Layer.empty to simplify this code

packages/sdk/client-services/src/packlets/services/service-context.ts
```diff
@@ -7,102 +7,59 @@ import type * as SqlError from '@effect/sql/SqlError';
 import * as EffectContext from 'effect/Context';
```
dmaretskyi: looks like this file got small enough to be merged with service-host

packages/sdk/client-services/src/packlets/services/service-host.ts
dmaretskyi: Tried this, but `Layer.empty` can't collapse the two branches here. The edge composition genuinely *requires* `EdgeConnectionService`/`EdgeHttpClientService` (pulled in by the feed syncer, edge replicator, and edge signal manager), which the non-edge composition neither needs nor provides. `Layer`'s `ROut` is contravariant, so a `endpoint ? realEdgeClients : Layer.empty` ternary widens to a type that isn't seen as *providing* those services, and the requirement leaks — tsc rejects it (`Type 'EdgeConnectionService' is not assignable to type 'never'`). So the two branches are kept, with a comment noting the constraint. Open to a cleaner approach if you have one in mind.

---

PR #12195, opened 2026-07-13
feat(plugin-connector): connection health checks, in-place reauth, and expired-credential toasts

wittjosiah:
## Summary

Adds a credential-lifecycle layer for external-service connections (Gmail/JMAP, Linear, Slack, Trello, Discord, GitHub, Bluesky), so a connection whose token has expired or been revoked is surfaced clearly and can be recovered without recreating it.

### Connection health + in-place reauth
- **`TestConnection` connector hook** + per-provider probes. The connection view shows live credential status and a **Test connection** action. Google's probe retries transient failures the same way sync does and only reports 401/403 as "reauthenticate" (a network blip no longer looks like an expired token).
- **In-place reauthentication** (`useReauthenticate` + `ConnectorCoordinator.reauthenticate`): re-runs OAuth for an existing connection, replacing its stored token **without** recreating the connection or its sync bindings.
- **`ConnectionView`** reordered so **Sync now** leads, then Test connection / Reauthenticate.

### Expired-credential toast
- A connector API call that fails with **HTTP 401** is retagged as `ConnectionAuthExpiredError`, producing a **"Connection expired"** toast with a **"Go to connection"** action — instead of a raw provider stack trace.

### Supporting primitives
- **`Operation.SerializedInvocation`** (`@dxos/compute`): a serializable `{ operation, input }` reference describing a deferred invocation that must cross the process-failure boundary (where a live `Effect` cannot).
- **`Process.Info.failure`** (`@dxos/compute`): the raw fail/die payload of a failed process, so the notify layer can inspect typed errors rather than the `Cause.pretty` string.
- **`LayoutOperation.NotifyOverride`** + `get`/`setNotifyOverride` (`@dxos/app-toolkit`): a failing error may carry `context.notifyOverride` to replace the default failure toast with a curated message and a serialized-invocation action; the deck notification tracker renders it via its own invoker. (Layered on top of main's change to keep raw exceptions out of toasts — the override fields are curated, safe to display.)
- **`makeGoogleApiRequest`** classifies failures by HTTP status, so a 401 with a non-JSON (HTML) error body surfaces as a typed `GoogleApiError(401)` rather than a raw `SyntaxError` defect.
- **`useTargetSync`** routes the mailbox "Sync" button through `ConnectorOperation.SyncConnection`, so every sync entry point shares one failure-handling path.

## Testing
- Unit tests: `LayoutOperation` notify-override round-trip, `isUnauthorizedError` / `ConnectionAuthExpiredError`, plus existing connector/inbox/deck suites (all green).
- Manually verified in Composer: breaking a connection's token produces the "Connection expired" toast with the action button.

## Notes / follow-ups
- A pre-existing, unrelated feed-decode `SyntaxError` (codec-protobuf choking on emoji/surrogate content during `seedDedupSet`) can mask the 401 in mailboxes whose feed contains such content. Tracked separately; not addressed here.

## Summary by CodeRabbit

* **New Features**
  * Added connection testing for Google, Bluesky, Discord, GitHub, Linear, Slack, and Trello with checking/valid/invalid/unsupported states.
  * Added a reauthentication flow for expired/failed credentials that refreshes tokens without recreating connection or sync settings.
  * Failure toasts can now use richer notification overrides, including optional action buttons.
* **Bug Fixes**
  * Improved structured error reporting for failed processes and clearer failure labels in the process tree.
  * Hardened Google API parsing and unauthorized/expired-credential detection.


packages/core/compute/compute/src/Operation.ts
```diff
@@ -473,6 +473,20 @@ export const NotifyOptionsAnnotation = Annotation.make({
...
+/**
+ * A serializable reference to an operation invocation — the target operation's key plus its input.
+ * Describes a deferred/late-bound invocation that must survive a serialization boundary (e.g. riding
+ * on a process-failure error, or persisted) where a live {@link Definition} cannot. Resolve the key
+ * with `OperationHandlerSet.getHandlerByKey` and hand the result to the invoker. Carries no functions,
+ * so `input` must itself be serializable.
+ */
+export interface SerializedInvocation {
+  /** Target operation key as a bare NSID (the `dxn:` prefix omitted), e.g. `org.dxos.plugin.deck.operation.open`. */
+  readonly operation: DXN.Name<string>;
```
dmaretskyi: it would be cleaner for this to be a URI

packages/core/compute/compute/src/Operation.ts
```diff
@@ -473,6 +473,20 @@ export const NotifyOptionsAnnotation = Annotation.make({
...
+  /** Input passed to the operation; must be serializable. */
+  readonly input?: unknown;
+}
+
```
dmaretskyi, line 490: add `Operation.prepare(defintion, input) -> SerializedInvocation`?

packages/core/compute/compute/src/Process.ts
```diff
@@ -469,6 +469,17 @@ export class ProcessMonitorService extends Context.Tag('@dxos/functions/ProcessM
...
 >() {}
 
+/**
+ * How a process failed. `message` is the pretty-printed cause for display and logging; `value` is the
+ * raw fail/die payload (the typed error object) for consumers that need to inspect it rather than the
+ * string — e.g. the notify layer reading a failure's toast override. `value` is not guaranteed to
+ * survive a serialized/persisted monitor.
+ */
+export interface Failure {
+  readonly message: string;
```
dmaretskyi: replace message with SerializedError from protocols

wittjosiah:
Addressed the review comments in `e8e8f5d`:

- **`Process.Info` "replace message with SerializedError from protocols"** — `Process.Info.error` is now a `@dxos/protocols` `SerializedError`. Its `context` carries the notify override, so this also removes the separate `failure`/`{ message, value }` field entirely — one structured, fully-serializable error field.
- **`Operation.ts` "cleaner for this to be a URI"** — `SerializedInvocation.operation` is now a `DXN` (URI) instead of a bare NSID.
- **`Operation.ts` "add `Operation.prepare(definition, input) -> SerializedInvocation`"** — added; `sync-connection` now builds the toast action via `Operation.prepare(LayoutOperation.Open, …)`.
- **`useReauthenticate` missing `coordinator` null check** (CodeRabbit) — guarded; returns early if the coordinator capability is absent so the button can't get stuck.
- **`TestConnection` untyped `Error` in the Effect error channel** (CodeRabbit) — introduced a typed `ConnectionTestError` and threaded it through the `TestConnection` type and all provider probes.

Build, lint, and tests pass across the affected packages.

PR #12210, opened 2026-07-14
worker-framework: complete effect-rpc session migration end-to-end

dmaretskyi:
## Problem

The `worker-framework` browser e2e tests (`counter.browser.test.ts`) were all timing out. The `migrate` commit had `Worker.run` bake effect-rpc protocol layers into the session context (`sessionProtocols`) and require `createSession` to return an effect over `RpcClient.Protocol | RpcServer.Protocol`.

Building the **reverse** worker→client client protocol eagerly awaits a `Ready` handshake from a tab-side runner — `@effect/rpc`'s `makeProtocolWorker` always does `pool.get` at layer-build time. The counter demo never sets up a reverse server, so the merged layer blocked forever, the RPC server never opened, and every session RPC timed out.

## Fix

Per the package's own `SPEC.mdl`, the framework is transport plumbing: it should hand over the **raw session ports** and leave RPC composition to the consumer (symmetric with `Client.onConnect`, which already receives raw ports).

- `Worker.RuntimeHandle.createSession` now receives `clientToWorker` / `workerToClient` `MessagePort`s and returns `Effect<never, never, Scope>`.
- Removed `sessionProtocols` from `Worker.ts` and the internal `serveFromContext` / `serveOverProtocol` helpers, replaced by a port-based `serveScoped` in `internal/rpc.ts`.
- The counter worker now serves on the raw port, matching the (passing) `RpcTiming` test.
- Switched `Effect.runPromise` → `EffectEx.runPromise` per the repo lint rule.

## Propagation to `@dxos/client`

The `migrate` commit changed the `Worker.run` API but left the consumers on the old signature, breaking the `@dxos/client` build. Updated:

- `dedicated-worker.ts` and `test-worker-factory.ts` — `createRuntime`/`createSession` converted to the Effect-based API; the `WorkerRuntime` session (which self-manages its lifecycle via the tab-liveness lock) is opened then the effect blocks for the session's lifetime.
- `dedicated-worker-client-services.ts` — renamed the `onConnect` port fields (`appPort` → `clientToWorker`, `systemPort` → `workerToClient`).

Also updated `SPEC.mdl`'s `RuntimeHandle` API line and the "app + system ports" wording to match the renamed wire protocol, and picked up an oxfmt import-order fix in `Client.test.ts`.

## Verification

- `worker-framework`: build ✓, `test-browser` 16/16 ✓, `test` 12/12 ✓, lint ✓
- `@dxos/client`: build ✓, lint ✓, `dedicated-worker-client-services.test.ts` 7/7 ✓ (full flow — leader election, port exchange, RPC, reconnection, lock stealing)

---
_Generated by [Claude Code](https://claude.ai/code/session_012B4greJXiRTxp2woBwXGtS)_

## Summary by CodeRabbit

* **Enhancements**
  * Migrated worker session/runtime setup to effect-scoped, long-lived sessions.
  * Standardized directed ports across the worker, coordinator, and bridge flows as `clientToWorker` and `workerToClient`.
  * Added protocol-layer support for RPC and bridge client setup (including protocol-based bridge client creation).
* **Bug Fixes**
  * Improved provide-port routing and connection wiring to better avoid race edge cases.
* **Documentation**
  * Updated the worker framework spec and examples to match the new session/channel interface.
* **Tests**
  * Refreshed worker and coordinator tests for the updated directed endpoint naming/behavior.


packages/sdk/client-services/src/packlets/worker/worker-session.ts
```diff
@@ -16,11 +19,14 @@ import { type ClientServicesHost } from '../services';
 export type WorkerSessionProps = {
   serviceHost: ClientServicesHost;
   /**
-   * Reverse-direction (worker→tab) channel serving the tab's {@link BridgeService} (WebRTC transport)
+   * Reverse-direction (worker→tab) protocol serving the tab's {@link BridgeService} (WebRTC transport)
    * over effect-rpc. The worker is the client; the tab is the runner.
    */
-  systemPort: MessagePort;
-  appPort: MessagePort;
+  systemProtocol: Layer.Layer<RpcClient.Protocol>;
```
dmaretskyi: instead of layers make those actually the values that RpcServer/Client.Protocol tags return

PR #12214, opened 2026-07-14
client-services: remove shared-worker path; convert worker runtime & host to Effect services

dmaretskyi:
## Summary

Worker-runtime refactor, in four phases (all on this branch):

1. **Kill shared-worker** — delete the dead SharedWorker client-services path now that `DEDICATED_WORKER` is the only worker mode. Removes the worker entrypoints/exports (`client/src/worker/onconnect.ts`, `react-client/src/worker.ts`, the `./worker` package exports — `./opfs-worker` retained), the composer + testbench `shared-worker.ts`, the legacy docs snippet, the `createWorker` client option and `SHARED_WORKER` factory case, and `WorkerRuntime`'s shared-worker-only lifecycle (`manageLifecycle`, `BroadcastChannel` displacement, `updateSignalMetadata`). The `SHARED_WORKER` proto enum values are deprecated but retained for wire compatibility. Composer defaults to `DEDICATED_WORKER` (`HOST` when `DX_HOST`).

2. **`WorkerRuntime` / `WorkerSession` → Effect services** — `WorkerRuntime` is now a `Context.Tag` + `layerWorkerRuntime` exposing an Effect service surface (`start` / `stop` / `createSession` / `connectWebrtcBridge`, + `host`); `WorkerSession.open` / `close` return Effects; the dedicated-worker and test-worker call sites drop the `Effect.promise` bridges. Lifecycle stays caller-driven (the worker framework wraps `createRuntime` in `Effect.scoped` and resolves it immediately, so a scoped-Layer finalizer would tear the runtime down too early).

3. **Eliminate `ServiceRegistry`** — the `ClientServicesHost` holds its resolved effect-rpc handlers directly (set on open, reset on close) instead of via a mutable registry bag.

4. **Fold `ServiceContext` into `ClientServicesHost`** — delete the `ServiceContext` orchestrator class and the `ServiceContextService` tag. The host now owns the full component stack and its lifecycle (migrate → identity → network → space init) plus the orchestration API (`createIdentity`, `broadcastProfileUpdate`, readiness gates, `_setNetworkIdentity`, `_initialize`). To break the circular dependency (the RPC handler layers are built inside the stack yet need orchestration), the host provides itself into its own stack via `ClientServicesHostService`. `ServiceContextLayer` is retained as pure component composition; `ServiceContext` remains only as a `type` alias for `ClientServicesHost` so type-only consumers are unchanged.

## Testing

- `client-services` build + lint + **154** unit tests (incl. `service-context.test.ts` device-invitation sync through the folded host).
- `client-e2e` **154 → 174** tests (incl. HaloProxy/EchoProxy invitations) — all green.
- `client` / `react-client` / `devtools` / `react-ui-form` / `composer-app` build; `effect-proto` tests.

## Notes

- Phase progress and design notes are tracked in `packages/sdk/client-services/TASKS.md`.
- `.changeset/drop-shared-worker-services.md` covers the consumer-facing removal (breaking → `minor`).

https://claude.ai/code/session_019WZkU8ArgFrmm87kqEreg9


packages/sdk/client-services/src/packlets/worker/worker-session.ts
```diff
@@ -103,38 +107,44 @@ export class WorkerSession {
...
 
-  async open(): Promise<void> {
-    log('opening...');
-    // The tab serves the WebRTC `BridgeService` on the reverse channel; build a client the worker's
-    // network stack can proxy through.
-    const { bridgeService, close } = await makeBridgeServiceClientOverProtocol(this._systemProtocol);
-    this.bridgeService = bridgeService;
-    this.#closeBridge = close;
+  open(): Effect.Effect<void> {
+    return Effect.promise(async () => {
```
dmaretskyi: prefer to have multiple smaller Effect.promise call sites instead of wrapping everything

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -98,7 +97,9 @@ export type InitializeOptions = {
  */
 export class ClientServicesHost {
   private readonly _resourceLock?: ResourceLock;
```
dmaretskyi: convert all other vars to #private then

packages/sdk/client-services/src/packlets/worker/worker-session.ts
```diff
@@ -66,7 +70,7 @@ export class WorkerSession {
       Effect.sync(() => {
```
dmaretskyi: yield* Effect.async(cb => setTimeout(cb))

packages/sdk/client/src/services/dedicated/dedicated-worker.ts
dmaretskyi: prefer wrapping individual calls in Effect.promise instead of whole blocks

dmaretskyi: extract opfs probe to a function

PR #12227, opened 2026-07-15
plugin-inbox: auto-create sync Routine on mailbox/calendar binding

wittjosiah:
Closes [DX-1091](https://linear.app/dxos/issue/DX-1091/inbox-sync-operation-schedule-force-trigger)

## Summary

- Auto-creates a recurring (every 10 min, local) sync Routine whenever a mailbox or calendar is bound to a connection — on a fresh OAuth connection, a multi-target selection (Calendar), and when reusing an already-authenticated connection. Discovered via `connectedRoutinesQuery` (surfaces in the routines companion) rather than an ownership field on the mailbox.
- The Routine's trigger invokes the connector's own `sync` operation — the exact same one `ConnectorOperation.SyncConnection` already calls directly — via a locally-serialized `Operation.PersistentOperation` record. No deployed Edge function is involved; the sync always runs in-process against the already-registered local handler.
- The mailbox toolbar's "Sync" button now force-runs that trigger through `Trigger.TriggerMonitorService` when one exists (creating one first if missing), and disables while the mailbox's own sync progress monitor is running, not just its own in-flight flag.
- `createSyncRoutine` is an Effect resolving `Database.Service` ambiently (not a `db` parameter), matching the codebase's existing pattern for space-scoped Effect functions.
- JMAP Mail now also gets `onCursorCreated` (previously only Gmail/Calendar) — it was excluded for a reason (no deployed edge function) that no longer applies now that the trigger invokes the connector's own local `sync` operation.
- `plugin-connector`'s `connector-coordinator.ts` split into a directory (`connector-coordinator/`), mirroring `plugin-space`'s `app-graph-builder` and `plugin-native-filesystem`'s `state` capability, since it had grown to cover several distinct concerns (OAuth flow, pending-connection snapshots, single/multi-target cursor creation, capability wiring).

## Bug fixes surfaced along the way

- Cancelling a Gmail sync mid-run left its progress monitor frozen at the last "running" snapshot forever: `Pipeline.abortWith` resolves a cancel as a genuine fiber interrupt, which skipped the cleanup that ran after the pipeline's `yield*`. `abortWith` now takes an `onCancel` effect and runs it via `Effect.onInterrupt` itself, so every caller (not just this one) gets correct cancellation cleanup for free.
- Binding a mailbox to an *existing* connection (the "reuse" toolbar action) created its `Cursor` directly, bypassing the new `onCursorCreated` hook — routines were only created for brand-new connections. Fixed so all cursor-creation paths route through the same hook.

## Test plan

- [x] `moon run plugin-inbox:test`, `plugin-connector:test`, `plugin-routine:test`, `echo:test`, `pipeline:test` — all pass
- [x] New tests: `sync-routine.test.ts` (routine creation, idempotency, calendar keying), `connector-auth.test.ts` (reuse fires `onCursorCreated`), `sync.test.ts` (cancel clears the progress monitor), `Pipeline.test.ts` (`abortWith`'s `onCancel` runs on abort)
- [ ] Manual: connect a Gmail mailbox, confirm a "Sync" routine appears in its routines companion and the toolbar sync button force-runs it; repeat for Calendar, JMAP, and for binding an existing connection to a new mailbox


packages/plugins/plugin-inbox/src/util/sync-routine.ts
```diff
@@ -0,0 +1,77 @@
...
+/** How often an auto-created sync routine's timer trigger fires. */
+const SYNC_ROUTINE_CRON = '*/10 * * * *';
+
+/**
+ * Finds an existing local record for `definition`, or persists a fresh one via
+ * {@link Operation.serialize}. Entirely local — `definition` is the connector's already-registered
+ * in-code operation (e.g. `InboxOperation.GoogleMailSync`), the same one `ConnectorOperation.SyncConnection`
+ * invokes directly; nothing is fetched from or deployed to Edge.
+ */
+const ensureOperationRecord = async (
```
dmaretskyi: why, can't we refrence the record in registry?

packages/plugins/plugin-inbox/src/util/sync-routine.ts
```diff
@@ -0,0 +1,77 @@
...
+}): Promise<Trigger.Trigger | undefined> => {
+  const connected = await db.query(connectedRoutinesQuery(target)).run();
+  for (const routine of connected) {
+    const existingTrigger = routine.triggers.find((ref) => ref.target?.spec?.kind === 'timer')?.target;
+    if (existingTrigger) {
+      return existingTrigger;
+    }
+  }
+
+  const operation = await ensureOperationRecord(db, sync);
```
dmaretskyi: reference record from registry?

packages/plugins/plugin-inbox/src/util/sync-routine.ts
```diff
@@ -0,0 +1,77 @@
...
+  const inputKey = Obj.getTypename(target) === Type.getTypename(Calendar.Calendar) ? 'calendar' : 'mailbox';
+  const trigger = Trigger.make({
+    enabled: true,
+    spec: Trigger.specTimer(SYNC_ROUTINE_CRON),
+    input: { binding: Ref.make(cursor), [inputKey]: db.makeRef(Obj.getURI(target)) },
+  });
+
+  const routine = Routine.make({
```
dmaretskyi: do we want inbox sync to be a routine or just a trigger?

PR #12235, opened 2026-07-16
echo: live feed objects with Obj.update semantics + subscription mutation types

wittjosiah:
## Summary

Feed-backed objects are now **live** — the same contract as automerge-backed objects — and subscription triggers gain real mutation semantics over feeds.

- `Obj.update` synchronously mutates a feed object and notifies subscribers (`Obj.subscribe`/atoms/`useObject`).
- Identity is stable across queries, re-appends, and ref resolution: a query returns the *same* instance you appended, not a fresh decode.
- Updates persist as a background whole-object re-append reusing the object's id; the index collapse-by-id (#12226) does the rest.
- Direct property assignment outside `Obj.update` throws instead of silently mutating in memory (matching the automerge-backed contract).
- **No opt-out.** Feed objects are uniformly live; a non-live fallback would be a sharp API edge.
- `db.add(obj, { to: feed })` is a synchronous alternative to `db.appendToFeed`, confirmed by `db.flush()`.
- Subscription triggers report a real `SubscriptionEvent.type` — `'created' | 'updated' | 'deleted'` (previously the placeholder `'unknown'`) — uniformly for both the space database and feed-scoped queries.

## Key design points

- **`@dxos/echo`** — `makeDecodedEntityLive` rewraps a JSON-decoded entity as a live `TypedReactiveHandler` proxy, reusing the existing backend-agnostic mutation/notification machinery rather than adding a new proxy handler. Live materialization is deliberately *not* inside `objectFromJSON`: the feed handle calls `makeDecodedEntityLive(await objectFromJSON(...))` itself. Also fixed `compactMetadataToInstanceState` choking on the non-configurable `SchemaId`/`TypeId` that `setSchema`/`setTypename` stamp on decoded objects.

- **`@dxos/echo-client`** — new `FeedObjectCore` holds a **single latest-state field plus a version baseline** (not a list of outstanding writes), with an explicit `change → dirty → appended → roundtripped` lifecycle: local state wins until our own append roundtrips (matched by content signature), after which newer remote blocks are adopted. `FeedHandle` is reworked around `#cores` with background flush via `UpdateScheduler` (coalescing N synchronous updates into one append) and `upsertFromJSON` as the single materialization entry point for polling, ref resolution, and index-backed query hydration.

  The version axis is `KEY_QUEUE_POSITION` rather than `insertionId`, which is not currently client-exposed (it lives in the block envelope and is stripped before the object JSON reaches the client; it is also absent on the index-served read path). Concurrent writers across tabs/processes are **last-flush-wins at whole-object granularity** — documented as a known limitation, not papered over.

- **Feed tombstone visibility (`@dxos/index-core`)** — feed deletes surface as queryable tombstones rather than an id-set diff. `Feed.remove` appends a bodyless `{ id, '@deleted': true }` block, and the index previously replaced the object's snapshot wholesale, so the tombstone lost its type and body and dropped out of results entirely. Now `FtsIndex.update` merges a partial block onto the prior snapshot and `EntityMetaIndex.update` preserves the prior row's type/kind/relation-endpoint/parent columns. A `deleted: 'include'` query therefore returns the deleted feed object with type + body intact, so it hydrates as deleted (`Obj.isDeleted === true`).

- **`@dxos/compute` / `@dxos/compute-runtime`** — change detection is **content-signature based** (a canonical sorted-key JSON signature), not `Obj.version`, because feed objects are unversioned — no automerge heads. The dispatcher detects deletes uniformly via `deleted: 'include'` + `Obj.isDeleted` for both sources. There is deliberately **no spec-level mutation-type filter**: `FeedSpec.ignoreUpdates` and the per-trigger delivered-ids state were removed to keep triggers light — gate on `event.type === 'created'` in the runnable instead.

## Notes

- `Feed.getFeedUri` is marked in jsdoc as private-ish and on track for removal.
- `evictFeedHandle` (test-support cache eviction) is public rather than `_`-prefixed: the d.ts pipeline strips *all* `_`-prefixed members, so an underscore name is not callable cross-package at `:build` time.
- The `Database.add` Effect-namespace wrapper intentionally omits the method's `opts` — it is applied point-free (`Effect.forEach(Database.add)`), where a second parameter collides with the iteratee index. `{ to }` lives on the method.

## Follow-ups (deliberately out of scope)

- Partial-object update blocks + field-level LWW merge at the index, and compaction/retention of superseded feed blocks (`Feed.RetentionOptions`).
- Wiring `insertionId` through the codec + index path so the version axis is always present (queue position is null unless `assignQueuePositions` is on).
- A bounded subscription-dedup strategy (TTL / high-water cursor) for the edge dispatcher — `processedVersions` is O(objects ever processed), fine for the process-scoped local dispatcher only.
- Optional `SubscriptionSpec.options.mutationTypes` filter.

## Test plan

- [x] `echo:test` — 469 passed, 9 skipped
- [x] `echo-client:test` — 508 passed, 7 expected fail, 12 skipped, 2 todo
- [x] `echo-client-e2e:test` — 282 passed, 1 expected fail, 7 skipped (includes the "Live updates" suite and a two-client cross-tab race)
- [x] `index-core:test` — 34 passed (tombstone partial-block merge)
- [x] `feed:test` — 19 passed
- [x] `compute:test` — 46 passed
- [x] `compute-runtime:test` — 152 passed, 1 skipped (subscription mutation-type dispatcher)
- [x] Build green across touched packages; lint 0 warnings / 0 errors; `oxfmt --check` clean
- [x] Merged latest `main`

## Changesets

- `feed-live-objects` — `@dxos/echo` minor
- `subscription-trigger-mutation-types` — `@dxos/compute` minor
- `feed-tombstone-body-preserved` — `@dxos/index-core` patch

## Summary by CodeRabbit

* **New Features**
  * Feed-backed objects are now live by default: `Obj.update` mutates synchronously, notifies subscribers, preserves identity across queries and re-appends, and persists in the background; `db.flush()` and feed-handle disposal wait for pending writes.
  * Subscription triggers now report real mutation types: `'created'`, `'updated'`, or `'deleted'`.
  * Deleted feed objects retain their type and body when included in queries.
* **Bug Fixes**
  * Improved feed hydration/reconciliation prevents stale remote updates from overwriting local changes.
  * Direct property assignment on live feed-backed objects now throws.
* **Tests**
  * Added/expanded end-to-end coverage for live updates, coalescing, hydration, retry behavior, and dispose/flush ordering.


dmaretskyi:
add `db.add(object, { to: feed })` as a sync alternative to `db.appendToFeed(object)`, confirmed by `db.flush()`

wittjosiah:
Added `db.add(obj, { to: feed })` in 0b938fc954 — a synchronous alternative to `db.appendToFeed`. It registers the object as a live feed object and schedules the write in the background, returning the same instance immediately; `db.flush()` confirms persistence. Implemented via a new `AddOptions.to` option routing to `FeedHandle.appendSync` (shares the register/optimistic-update path with the async `append`). Covered by two new e2e tests in `feed.test.ts` (sync add + live update, and reload durability).

packages/core/echo/echo-client/src/feed/feed-object-core.ts
```diff
@@ -0,0 +1,260 @@
...
+ * silently diverges from the feed until the next genuinely-newer *positioned* write arrives (or the
+ * object is re-touched). Enabling positions closes this gap; it is not yet the default.
+ */
+export class FeedObjectCore {
+  readonly entity: Entity.Unknown;
+
+  #dirty = false;
+  #deleted = false;
+  #applyingRemote = false;
+  #unsubscribe: (() => void) | undefined;
```
dmaretskyi, line 78: i dont think we need to append all intemittent updates, only the combined final state.

so I'd store a single state field -- the latest json state of the feed.

then a set of versions -- that we can compare according to the feed semantics:

- the version of the block that came back from feed store
- whether the current state is dirty and hasn't been appended
- if the current state is dirty and has been appended, what was the version it was appended at, so we know whether to retain local state or remote state when we get another update from the feed
- the version of latest block we got from feed service


lifecycle

Obj.change -> dirty local -> appended -> roundtripped (came back from service) 

after the local state roundtripped, all future blocks should update state, before that we prefer local dirty state

dmaretskyi: insertionId seems like the best choice to tack versions

packages/core/echo/echo/src/internal/Obj/json-serializer.ts
```diff
@@ -86,7 +94,14 @@ export const objectFromJSON = async (
...
     database,
     parent,
-  }: { refResolver?: RefResolver; uri?: URI.URI; database?: Database.Database; parent?: Obj.Unknown } = {},
+    live,
+  }: {
+    refResolver?: RefResolver;
+    uri?: URI.URI;
+    database?: Database.Database;
+    parent?: Obj.Unknown;
+    live?: boolean;
```
dmaretskyi: seems like it doesn't belong in this function. rather pull it out

packages/core/compute/functions-runtime/src/triggers/feed-delivery-state.ts
```diff
@@ -0,0 +1,43 @@
...
+ * independently-committed write, and a crash between the two would either replay a successful
+ * delivery as `isUpdate: true` or drop a genuine update back to `isUpdate: false`.
+ */
+export const KEY_FEED_DELIVERED_IDS = 'org.dxos.key.local-trigger-dispatcher.feed-delivered-ids';
+
+/**
+ * Reads the set of feed-item ids already delivered to this trigger. Read-only — use
+ * {@link setDeliveredFeedIds} inside an `Obj.update` callback to persist additions.
+ */
+export const getDeliveredFeedIds = (trigger: Obj.Unknown): Set<string> => {
```
dmaretskyi: nononono, this will explode the trigger document. lets keep triggers light and remove the skip updates functionality. reserve that for subscription triggers.

packages/core/echo/echo-client/src/feed/feed-object-core.ts
wittjosiah: Done in c41bfbcd06. Reworked `FeedObjectCore` to hold a single latest-state field (`#state`) plus a version baseline instead of a list of outstanding appends, with an explicit change→dirty→appended→roundtripped lifecycle: we prefer local state until our own append roundtrips (matched by content), then adopt newer remote blocks (whole-object LWW). Appends already coalesce to one combined state, so there's only ever one pending write.

wittjosiah: Used `KEY_QUEUE_POSITION` as the version axis for now: `insertionId` isn't currently exposed to the client — it lives only in the block envelope and is stripped before the object JSON reaches the client (and it's absent on the index-served read path). Wiring it through the codec + index path so it's stamped like the queue position would be a good follow-up to make versioning always-present (position is null unless `assignQueuePositions` is on); left a note in the class doc.

packages/core/echo/echo/src/internal/Obj/json-serializer.ts
wittjosiah: Done in c41bfbcd06 — dropped the `live` option from `objectFromJSON`; the feed handle now calls `makeDecodedEntityLive(await objectFromJSON(...))` itself.

packages/core/compute/functions-runtime/src/triggers/feed-delivery-state.ts
wittjosiah: Done in c41bfbcd06 — removed the skip-updates functionality and the per-trigger delivered-ids state entirely (deleted feed-delivery-state.ts), and reverted `FeedEvent.isUpdate` / `FeedSpec.ignoreUpdates` and the dispatcher feed-branch changes. Triggers stay light; new-vs-update is left to subscription triggers.

dmaretskyi:
I've pushed a spec for subscription trigger support for feeds. impl it

wittjosiah:
On it — implementing the subscription-trigger-for-feeds spec. Starting with the feed-deletion-visibility spike (step 1) since it gates the delete path.

wittjosiah:
Implemented the subscription-trigger-for-feeds spec (835160b9b9).

- `SubscriptionEvent.type` is now `'created' | 'updated' | 'deleted'` (SubscriptionMutationType), replacing the `'unknown'` placeholder.
- Added `Trigger.specSubscriptionFromFeed(feed, filter?, options?)` — builds a feed-scoped subscription so create/update/delete semantics apply to feed items.
- Test matrix: database × {created, updated, deleted, no-op, delete→re-create}, feed × {created, updated, deleted, filtered}, plus a `subject` dereference test; un-skipped the no-op idempotency test. All green (105 in trigger-dispatcher.test.ts; assistant-toolkit/compute unaffected).

Two deviations from the spec, both driven by the spike (details in the plan doc's new 'Implementation notes'):

1. **Change detection is content-signature based, not `Obj.version`.** Feed-backed live objects are unversioned (no automerge heads), so version comparison never sees a feed re-append as an update. A canonical JSON signature works uniformly for both sources. This also fixed the previously-skipped DB no-op test.
2. **Feed deletes use the id-set diff.** The spike showed `Feed.remove` leaves no queryable tombstone (`deleted:'include'` returns nothing for a removed feed item), so feed-scoped deletes are detected by diffing seen ids vs the live set. The database source keeps the precise `deleted:'include'` + `Obj.isDeleted` path; an object merely falling out of the filter predicate is not reported as deleted (per the plan's default).

Flag on the content-signature choice since it replaces the `processedVersions`/`Obj.version` mechanism you specced — happy to revisit if you'd prefer version for DB + a separate feed path, but the unified signature is simpler and the only thing that actually detects feed updates.

packages/core/compute/compute/src/Trigger.ts
```diff
@@ -95,6 +96,22 @@ export const specSubscription = (
     : undefined,
 });
 
+/**
+ * Construct a Subscription trigger spec scoped to a feed, so subscription semantics
+ * (created/updated/deleted mutations) apply to feed items rather than the space database.
+ *
+ * The feed must be persisted (added to the database) so its queue URI is resolvable.
+ */
+export const specSubscriptionFromFeed = (
```
dmaretskyi: first of, Feed.getFeedUri is a private-ish api, thats on track to be removed -- mark it jsdoc as such

second, we dont need this new `specSubscriptionFromFeed`, since Trigger.specSubscription(Query.select(...).from(Scope.feed(feed)))) is perfectly usable

delete new helper and use the specSubscription

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
```diff
@@ -686,39 +694,70 @@ class TriggerDispatcherImpl implements Context.Tag.Service<TriggerDispatcher> {
                 ),
               );
               invariant(state.state?._tag === 'subscription');
+              const processedVersions = state.state.processedVersions;
```
dmaretskyi, line 701: its okay for local dispatcher, but edge should use something that doesn't turn into O(number_of_object_ever_processed). -- add a comment

packages/core/compute/functions-runtime/src/triggers/trigger-dispatcher.ts
```diff
@@ -686,39 +694,70 @@ class TriggerDispatcherImpl implements Context.Tag.Service<TriggerDispatcher> {
...
+                invocations.push(yield* fire(type, object.id, Obj.getURI(object)));
+                (processedVersions as Record<string, string>)[object.id] = currentSignature;
                 updated = true;
               }
 
+              // Feed removals leave no visible tombstone (unlike the database source), so detect
+              // deletes by diffing previously-seen ids against the current live result set. Only for
+              // feed-scoped queries: for the database source a real tombstone is authoritative and an
+              // object merely falling out of the filter predicate must not be reported as deleted.
+              if (feedScoped) {
```
dmaretskyi: feed objects absolutely leave tombstones. they are also indexed the same way. if this is a query deficiecy, we should address it upstream

packages/core/compute/compute/src/Trigger.ts
wittjosiah: Done in 1cac62f067. Marked `Feed.getFeedUri` jsdoc as private-ish / on-track-for-removal (with a `TODO(dmaretskyi)`), and removed `specSubscriptionFromFeed` — the tests now build feed-scoped subscriptions inline via `specSubscription(Query.select(...).from(Scope.feed(...)))`.

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
wittjosiah: Done in 1cac62f067 — added a comment noting `processedVersions` is O(objects ever processed), which is fine for the process-scoped local dispatcher but needs a bounded strategy (TTL / high-water-mark cursor) on the edge.

packages/core/compute/functions-runtime/src/triggers/trigger-dispatcher.ts
wittjosiah: You're right — fixed upstream in 1cac62f067. A `Feed.remove` appends a bodyless `{ id, "@deleted": true }` block, and the index was replacing the object's snapshot wholesale, so the tombstone lost its type/body: the client couldn't hydrate it (`Obj.fromJSON` needs `@type`) and it dropped out of results entirely. Now:

- `FtsIndex.update` merges a partial (bodyless) block onto the prior snapshot instead of replacing it, and `EntityMetaIndex.update` preserves the prior row's type/kind/relation/parent columns for such a block. A `deleted: 'include'` query returns the deleted feed object with type + body intact, so it hydrates as a deleted object (`Obj.isDeleted === true`).
- The dispatcher now drops the feed id-set diff entirely and detects deletes uniformly via `deleted: 'include'` + `Obj.isDeleted` for both the database and feed sources.

PR #12248, opened 2026-07-17
plugin-inbox: incremental mail sync (delta tokens + reconcile) + unified system tags

wittjosiah:
## Summary

Builds on the bidirectional/watermark cursor work to make mail sync **incremental**, and unifies provider system state (Gmail labels / JMAP roles + keywords) onto a **shared canonical tag registry**.

### Incremental delta-resume sync
- The sync cursor now carries an opaque provider **delta token** (Gmail `historyId`, JMAP `Email/get` `state`) plus an optional `foreignId → EntityId` map (built only on reconcile runs).
- After the initial window backfill, each tick **fetches the delta itself** (pull, not push) since the stored token — Gmail `history.list`, JMAP `Email/changes` — and applies additions + label/flag reconciliation to already-committed feed messages via **objectless commit units** (no feed append, no `max`/`min` movement).
- A stale token falls back to the window scan and recaptures; the token advances **only after the run's merged stream fully drains**, so a crash/cap re-fetches the delta idempotently (additions dedup-drop, tag ops re-apply).
- **Gmail `history.list` is paginated** — the drain follows `nextPageToken` before adopting the new `historyId`, so a delta spanning >1 page isn't silently dropped (Gmail returns the mailbox's *current* `historyId` on every page). Regression test included.

### Unified system tags
- New canonical, space-general tag registry `types/SystemTags.ts` (`org.dxos.tag`): `starred`, `inbox`, `important`, `sent`, and the Gmail categories. Gmail system labels, JMAP mailbox roles, and the JMAP `$flagged` keyword all resolve to the **same** `Tag` objects instead of provider-scoped ones — a Gmail star, a JMAP flag, and a locally-toggled star are now one tag.
- **Intentionally dropped** (explicit, commented): read-state (`UNREAD`/`$seen`), `drafts`, `trash`, `spam`, and `archive` — archive is derived as *"not in inbox"* (the Gmail model), so an archive view is a query, not a tag.
- `toggleStarred` and `PERSONAL_TAG_KEYS` repointed to canonical keys. JMAP reconcile diffs folders remote-wins (add+remove) and keywords **add-only** (a remote star can be added, but a local star is never auto-clobbered pending write-back).

### Notable decisions
- Starred key moves `org.dxos.org` → `org.dxos.tag`; existing locally-starred items under the old key are **not migrated** (per design discussion — acceptable).
- Categories are canonicalized so future providers map onto them and first-party auto-tagging can reuse them.
- `delete` reconcile is wired but stubbed pending safe in-feed object deletion (TODO in `EmailStage`).

## Test
- `@dxos/link` Cursor: pass. plugin-inbox google sync, jmap sync, jmap mapper, sync harness, Starred: **all green** (incl. new tests for multi-page history drain, delta token capture/advance/crash-recovery, and Gmail `STARRED`/`CATEGORY_*` + JMAP `$flagged` → canonical tags).
- The full suite has a heavily-loaded-machine flake (unrelated app-harness activation timeouts); affected files verified green in isolation. CI is the authority.

## Summary by CodeRabbit

* **New Features**
  * Mail synchronization now updates incrementally, fetching only new or changed messages after the initial sync.
  * Gmail and JMAP changes are resumed reliably, with automatic recovery for stale tokens and interrupted runs.
  * Existing messages are retagged when labels, folders, or starred status change.
  * Gmail and JMAP system labels now use consistent shared tags across providers.
* **Bug Fixes**
  * Prevented duplicate messages during incremental sync recovery.
  * Improved handling of archived, deleted, draft, spam, and read-state messages.


wittjosiah:
Thanks @coderabbitai — triaged all findings. Fixed the still-valid ones (commit 4c3298c), skipped the rest with reasons below.

**Fixed**
- **Bound `history.list` pagination** (`Cursor`-adjacent, google `sync-provider`): the drain loop now caps at `MAX_HISTORY_PAGES`; exceeding it reroutes through the existing stale-token path (clear + full window scan) rather than looping unboundedly.
- **Apply the configured filter to incremental additions** (both providers): when a user filter is set, the delta's account-wide `created` ids no longer bypass it — additions fall back to the filtered forward window scan, while the delta still drives reconcile. (Chose the filtered-window fallback over an id∩query intersection to keep it minimal and correct.)
- **Gmail mock `listHistory` chain validation**: now fails `404` when the traversed chain doesn't reach the current `historyId` (a gap), matching a server past retention.

**Skipped (with reasons)**
- **Rebuilding the full-feed foreignId→EntityId index each reconcile run** (`Cursor.seedForeignIndex`): intentional and documented — built *only* on token (reconcile) runs, and feed/queue members aren't indexable by foreign key via `db.query`, so a targeted lookup isn't available without a durable secondary index. That's a separate, larger optimization, tracked as follow-up rather than done in this PR.
- **Bound JMAP `Email/changes` before advancing / honor `maxChanges` / `hasMoreChanges`**: verified the premise doesn't hold for our flow — the token advances *only* when the run isn't capped (`writeToken` is gated on `!capped`), so an oversized `created` delta truncated by the run cap leaves the token unadvanced and the next run resumes with no loss. `maxChanges`/`hasMoreChanges` is a latency optimization (fewer round-trips), not a correctness fix.
- **Total change/page budget covering retags+deletes** (`mail-sync`): same class as the full-feed item — reconcile deltas are label/flag changes (small in practice); a durable continuation budget across additions+reconcile is a larger design change, deferred.
- **Non-null assertions in tests / test-helper placement / `CommitUnit` discriminated union**: the `!` usage matches the pervasive existing idiom in these test files (the no-`!` rule targets production code); helper placement and the union refactor are style/scope beyond this PR.

Affected suites green locally after the fixes (google + jmap sync, 35/35); Check was green pre-fix and re-running now.

wittjosiah:
Follow-up: on reflection (and per reviewer feedback) I've now **properly addressed** the two scale findings I'd initially deferred, in commit 6075228:

- **Full-feed foreign-index rebuild → targeted lookup.** `Cursor.layer` now takes a `reconcileFilter` (`Filter.foreignKeys(Message, <delta keys>)`) and resolves EntityIds for *only* the delta's reconcile messages instead of `Filter.everything()`. O(delta) once the feed query engine supports filter pushdown; correct today (the filter is applied in-memory — verified by a `link` unit test that a two-key filter returns exactly those two).
- **Bound reconciliation like additions.** Both providers now fetch **one bounded delta chunk per run** (JMAP `Email/changes` `maxChanges`; Gmail one `history.list` page via `maxResults`) and expose `hasMoreDelta`. The token advances to the *chunk boundary* (JMAP `newState`; Gmail the last processed record id), and the harness `runAgain()`s while more remains — so reconciliation rides the same per-run budget + durable re-run as additions, and a large delta drains across runs rather than in one unbounded pass. This replaces the earlier drain-all-pages / `MAX_HISTORY_PAGES` fallback.

Mocks honor the bounds and there are regression tests for both providers (a label-change delta and a created delta each larger than the per-run budget draining across bounded runs). link Cursor 23/23, google+jmap sync 36/36 green locally.

PR #12256, opened 2026-07-18
echo: per-object branching + time travel (core mechanism)

richburdon:
Stage 1 of the branching convergence (#11829 × #12237): the **ECHO-core** mechanism only —
time travel and per-object branching in `echo-protocol` / `echo` / `echo-client` / `echo-host`.
The `@dxos/versioning` rewire and the plugin/UI convergence land as follow-ups on top; this PR
is the self-contained foundation. Split out of #12246 to shrink review surface.

## What's here

- **Time travel** — `ObjectCore.setTimeTravel`/`clearTimeTravel` (memoized `A.view` read-pin;
  writes throw); dual notification channels (`updates`/`displayUpdates`, `LatestEventId`) with a
  `latestOnly` subscription option + ambient `withLatestRead`, so side-effecting subscribers never
  observe scrubbed data; `Entity.isTimeTraveling`/`timeTravelAtom`; `getEditHistoryWithDiffs`.
- **Branching** — the space-root `branches` registry (`SpaceBranchRegistry`/`BranchRecord`);
  `EntityManager.createBranch`/`switchBranch`/`mergeBranch`/`deleteBranch` (history-preserving
  `forkDump` + `RepoProxy.import`, CRDT merge-back, serialized `#branchOpChain`); injected
  `BranchStore` device-local selection; writable per-surface `BranchBinding` (`db.branch(obj,name)`).
- **Lifecycle wiring** — `DatabaseRoot.getAllLinkedDocuments` unions branch-doc urls (replication +
  export); `DatabaseRoot.mapLinks` remaps the registry on import/copy.
- **Design doc** — `packages/core/echo/echo-client/docs/VERSIONING.md` (mechanism + Status/risk).

## Why it's independently landable

Bottom of a strict dependency stack (`versioning → echo-client → echo-host/echo/protocol`): no
echo-core file imports `@dxos/versioning` or any plugin. Additive and backward-compatible — the new
API lands unused until the follow-up; the existing content-copy versioning path is untouched, so the
repo stays green. Self-tested by `time-travel.test.ts`, `edit-history.test.ts`, `branching.test.ts`
(incl. two-peer replication, reload persistence, concurrent-frontier fork), `branch-binding.test.ts`.

## Known limitations (see VERSIONING.md § Status)

`getDocumentHeads`/`reIndexHeads` are not yet branch-aware (deferred; `reIndexHeads` deliberately
excludes branch docs to avoid phantom objects); nested branch-of-branch is unsupported (flat
registry); inline objects can't be branched (clear error).

Original author of the ported ECHO-core commits: @wittjosiah (co-authored per the convergence plan).

## Summary by CodeRabbit

* **New Features**
  * Added per-object time travel with `latestOnly`-aware reactivity and a `timeTravelAtom` indicator.
  * Added subtree branching: create, switch, merge, delete, and read object data on a specific branch.
  * Added `BranchBinding` plus optional device-local persistence of current branch selection.
  * Added per-version edit history diffs with added/removed magnitude reporting.
* **Bug Fixes**
  * Fixed branch-document link remapping when importing/copying spaces.
  * Updated ref atoms to resolve to `undefined` on initial read when the referenced target is already deleted.
* **Documentation**
  * Published a design/spec for time travel and branching behavior.


packages/core/echo/echo-client/docs/VERSIONING.md
```diff
@@ -0,0 +1,303 @@
...
+migration is needed.
+
+Full design, plugin/timeline plumbing, and estimate:
+[`agents/superpowers/specs/2026-07-17-nested-branches-design.md`](../../../../../agents/superpowers/specs/2026-07-17-nested-branches-design.md).
+
+## Design notes & constraints
+
+- **Selection is per-surface/session, not device-global.** The canonical object is never globally
+  switched out from under other surfaces; each plank/agent binds independently. `switchBranch`
+  remains the lower-level device-global capability that `BranchStore` persists.
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: reworded the design note to distinguish the two mechanisms — the per-surface `db.branch()` bindings never globally switch the canonical object, whereas `switchBranch` IS a device-global rebind. No more contradictory 'never globally switched' phrasing.

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -229,6 +246,47 @@ export class EntityManager implements IDatabaseBinding {
     }
 
     this.opened.wake();
+
+    // Restore device-local branch selections (e.g. after a reload) and re-bind each branched subtree
+    // to its selected branch — objects otherwise load on `main`.
+    if (this._branchStore) {
+      await this._hydrateCurrentBranches();
+    }
```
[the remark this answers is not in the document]
richburdon: Valid edge case, deferring with rationale. This only bites on a space-root URL swap (`updateSpaceState` replacing the root), which is rare; the initial-open hydration path is correct. A clean fix clears and re-hydrates `_currentBranches` after the root is replaced — tracked as follow-up hardening rather than risked in this foundational PR.

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -748,6 +806,326 @@ export class EntityManager implements IDatabaseBinding {
...
+    const baseHeads = memberHeads[rootObjectId] ?? getHeads(rootCore.getDoc());
+    const createdAt = Date.now();
+    spaceRoot.change((doc: DatabaseDirectory) => {
+      // Assign through re-read doc proxies (not a chained `??=` result, which returns the orphan
+      // literal under Automerge and silently drops the write).
+      doc.branches ??= {};
+      doc.branches[rootObjectId] ??= {};
+      doc.branches[rootObjectId][name] = { members: memberUrls, baseHeads, createdAt };
+    });
+    this.branchesChanged.emit();
```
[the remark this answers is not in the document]
richburdon: Addressed in f3c3b50: `createBranch` now runs through `#branchOpChain` (so a concurrent `switchBranch` can't rebind members between the awaited imports), validates the whole subtree up front — every member has its own document AND is on `main` (`!_currentBranches.has(id)`) — before importing any doc, so a mid-loop failure can no longer orphan already-imported branch docs, and the fork can't capture a non-main member's content. Serialization also subsumes the separate 'capture then import' phase.

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -748,6 +806,326 @@ export class EntityManager implements IDatabaseBinding {
...
+        memberIds.add(id);
+      }
+    }
+    for (const memberId of memberIds) {
+      await this._rebindMemberToBranch(memberId, name, registry);
+      if (name === 'main') {
+        this._currentBranches.delete(memberId);
+      } else {
+        this._currentBranches.set(memberId, name);
+      }
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: `switchBranch` now records a member's branch only when it is actually in that branch's member set (`registry[name].members[memberId] != null`); members that `_rebindMemberToBranch` fell back to main are cleared from `_currentBranches`, so `getCurrentBranch` matches the real binding.

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -748,6 +806,326 @@ export class EntityManager implements IDatabaseBinding {
...
+    return this.#enqueueBranchOp(async () => {
+      const record = this.getBranchRegistry(rootObjectId)?.[name];
+      invariant(record, `branch not found: ${name}`);
+      for (const [memberId, urlData] of Object.entries(record.members)) {
+        const branchHandle = this._repoProxy.find<DatabaseDirectory>(urlData.toString() as DocumentId);
+        await branchHandle.whenReady();
+        const branchDoc = branchHandle.doc();
+        const mainHandle = await this._mainDocHandle(memberId);
+        mainHandle.update((doc) => A.merge(doc, branchDoc));
+      }
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: `mergeBranch` now preflights every member's branch doc + main handle into a list before applying any `A.merge`, so a member that fails to load aborts the merge cleanly instead of leaving the subtree half-merged.

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -748,6 +806,326 @@ export class EntityManager implements IDatabaseBinding {
...
+      this.#enqueueBranchOp(async () => {
+        for (const memberId of orphaned) {
+          await this._rebindMemberToBranch(memberId, 'main', undefined);
+          this._currentBranches.delete(memberId);
+        }
+        this._persistCurrentBranches();
+        this._scheduleThrottledDbUpdate(orphaned);
+      }).catch((err) => log.warn('failed to fall back to main after branch deletion', { err }));
+    }
+    this.branchesChanged.emit();
```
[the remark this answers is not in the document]
richburdon: Real, deferring the standalone-delete race with rationale. The common delete path — `mergeBranch(..., {deleteAfter:true})` — already runs the whole registry-removal + rebind inside `#branchOpChain`, so it is serialized. Closing the window for a *direct* `deleteBranch()` racing an in-flight `switchBranch` requires making `deleteBranch` async and updating its callers; tracked as follow-up rather than an API change in this PR.

packages/core/echo/echo-client/src/core-db/object-core.ts
```diff
@@ -405,14 +450,84 @@ export class ObjectCore {
...
+    assertArgument(Array.isArray(heads), 'heads', 'expected automerge heads array');
+    const latest = this.getDoc();
+    let view: AutomergeDoc<unknown>;
+    try {
+      view = A.view(latest as AutomergeDoc<any>, heads);
+    } catch (err) {
+      // Heads not reachable in the local doc (e.g. a peer's changes not yet synced); leave live.
+      log.warn('ignoring setTimeTravel with unreachable heads', { id: this.id, heads, err });
+      return;
+    }
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: the unreachable-heads catch path now calls `clearTimeTravel()` before returning, so an already-pinned object returns to live (matching the documented contract) instead of stranding a stale historical view.

packages/core/echo/echo-client/src/echo-handler/edit-history.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+    expect(Doc.getValue<string>(accessor)).toBe('aaabbbb');
+
+    let diffs = getEditHistoryWithDiffs(obj);
+    // One diff per change.
+    expect(diffs.length).toBe(getEditHistory(obj).length);
+    // The newest version is the insertion: additions, no deletions.
+    expect(diffs.at(-1)!.added).toBeGreaterThanOrEqual(4);
+    expect(diffs.at(-1)!.removed).toBe(0);
+    expect(diffs.at(-1)!.time).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: replaced every `.at(-1)!` with an explicit guard — `expect(x).toBeDefined()` + optional chaining for assertions, and `invariant(...)` where a value must exist for the test to proceed.

packages/core/echo/echo-client/src/echo-handler/edit-history.ts
```diff
@@ -28,6 +28,96 @@ export const getEditHistory = (object: Obj.Unknown): State<any>[] => {
...
+    return {
+      heads: after,
+      // Automerge stores change time in epoch seconds.
+      time: change.time * 1000,
+      actor: change.actor,
+      message: change.message ?? undefined,
+      added,
+      removed,
+    };
+  });
```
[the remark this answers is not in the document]
richburdon: Deferring with rationale. In the shipped use case the versioned targets are `Text` objects that live in their own documents (not inline), so `getDoc()` already returns only that object's history — the cross-object bleed only occurs for inline objects sharing the space-root doc. Filtering diffs by `objectCore.mountPath` (+ a two-object shared-document test) is a correct generalization and is tracked as follow-up; it does not affect the branching/versioning path this PR enables.

packages/core/echo/echo-client/src/echo-handler/time-travel.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      registry.subscribe(timeTravelAtom, () => {}),
+    ];
+
+    expect((registry.get(defaultAtom) as any).title).toBe('v2');
+    expect((registry.get(latestAtom) as any).title).toBe('v2');
+    expect(registry.get(timeTravelAtom)).toBe(false);
+
+    setTimeTravel(obj, headsV0);
+    expect((registry.get(defaultAtom) as any).title).toBe('v0');
+    expect((registry.get(latestAtom) as any).title).toBe('v2');
```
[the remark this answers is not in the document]
richburdon: Acknowledged; deferring as a scoped test-typing sweep. These `any` callbacks/`!` in the three core test files are not lint-blocking (CI is green), but they do relax type coverage. Tightening them across `time-travel`/`branching`/`branch-binding` tests is tracked as a follow-up cleanup; the `edit-history` test was already de-`!`-ed in f3c3b50.

packages/core/echo/echo-client/src/proxy-db/database.ts
```diff
@@ -813,6 +832,53 @@ export class DatabaseImpl extends Resource implements EchoDatabase {
...
+  async branch<T extends Obj.Unknown>(obj: T, name: string): Promise<BranchBinding<T>> {
+    assertArgument(isEchoObject(obj), 'obj', 'expected ECHO object stored in the database');
+    // Guard against foreign/unbound objects: 'main' would hand back an unrelated live object as a
+    // valid binding, and other branches resolve by id only (an id collision could bind another
+    // space's data).
+    assertArgument(getObjectCore(obj).database === this, 'obj', 'object is not bound to this database');
+    if (name === 'main') {
+      // The live object IS the main binding; nothing to release.
+      return { object: obj, branch: 'main', dispose: () => {} };
+    }
```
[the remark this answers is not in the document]
richburdon: Real edge case, deferring with rationale. `db.branch()` is the per-surface flow; the versioning UI keeps the device on `main` and binds branches per-surface, so `name === 'main'` returning the canonical live object is correct there. It is only wrong when the device has `switchBranch`'d away — and returning a correct `'main'` binding in that state needs a `bindCoreToBranch` main-doc path (it currently invariants `name !== 'main'`). Tracked as follow-up; the shipped flow is correct.

packages/core/echo/echo/src/Entity.ts
```diff
@@ -249,6 +249,12 @@ export const getKeys = (entity: Unknown | Snapshot, source: string): ForeignKey[
  */
 export const isDeleted = (entity: Unknown | Snapshot): boolean => internal.isDeleted(entity);
 
+/**
+ * Check if an entity is currently in a historical (time-travel) read mode.
+ * Synchronous guard — use before mutating to avoid the time-travel write error.
+ */
+export const isTimeTraveling = (entity: Unknown | Snapshot): boolean => internal.isTimeTraveling(entity);
```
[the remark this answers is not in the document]
richburdon: Fixed in f3c3b50: `Entity.isTimeTraveling` now takes `Unknown` only — a detached `Snapshot` carries no pin state and can't transition, so accepting it only invited a misleading always-`false`.

packages/core/echo/echo/src/internal/common/proxy/reactive.ts
```diff
@@ -24,6 +34,12 @@ export const subscribe = (obj: unknown, callback: () => void): (() => void) => {
     return () => {};
   }
   const target = getProxyTarget(obj as any);
+  if (opts?.latestOnly) {
+    // Non-ECHO objects have no time-travel, so their default channel is already latest-only.
+    if (target && LatestEventId in target) {
+      return (target as any)[LatestEventId].on(callback);
+    }
+  }
```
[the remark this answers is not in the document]
richburdon: Deferring with rationale. The `(target as any)[LatestEventId]` access is one instance of a pre-existing pattern — `change-context.ts` and `event-batch.ts` access `[EventId]`/`[LatestEventId]` on the proxy target the same way, because `ProxyTarget` only declares `[EventId]`. Adding `[LatestEventId]?: Event<void>` to `ProxyTarget` and dropping the casts is a small, correct sweep best done across all sites at once rather than one call site here; tracked as follow-up.

packages/core/echo/echo-client/docs/VERSIONING.md
```diff
@@ -0,0 +1,305 @@
...
+(`DatabaseDirectory`) holds inline objects under `objects` and a `links` map (`objectId → doc url`)
+for objects that live in their own documents. Automerge tracks history as a Merkle-DAG of changes;
+`heads` is the frontier (the set of latest change hashes — an array, because concurrent edits leave
+multiple leaves).
+
+Two facts drive the design:
+
+- `A.view(doc, heads)` reconstructs any historical state, and `A.save`/`A.load` round-trips a
+  document **preserving its change ancestry** — the basis for both features.
+- A branch cannot hide inside one document: to diverge and later merge cleanly it needs its own
```
dmaretskyi, line 64: I dont understand why

packages/core/echo/echo-client/src/core-db/object-core.ts
```diff
@@ -405,14 +450,86 @@ export class ObjectCore {
...
+    return this.#timeTravelView;
+  }
+
+  /**
+   * Pin data/meta reads to the given historical heads. Reads on this object (and its nested
+   * records/arrays, which share this core) resolve the historical value; `displayUpdates` fires so
+   * the editor and label re-render. The full edit history and `system` reads stay on the latest doc.
+   */
+  setTimeTravel(heads: Heads): void {
```
dmaretskyi, line 493: since version-pinned objects are immutable wouldn't the better api be functional:

```ts
Obj.getVersion(obj, version) -> Obj.Snapshot
```

return a separate instance of an object, doesn't need to store state in object-core

packages/core/echo/echo-client/docs/VERSIONING.md
```diff
@@ -0,0 +1,305 @@
...
+- `#getReadDoc(path)` — while pinned, `data` and `meta` reads resolve the historical view; `system`
+  reads, indexing, writes, and edit history always use the latest doc. An ambient
+  `isReadingLatest()` context (below) lets a `latestOnly` recompute bypass the pin.
+- `change`/`changeAt` throw `Cannot mutate a time-traveling object` while pinned.
+
+### Dual notification channels
+
+Reactivity is split so side-effecting subscribers never observe scrubbed (historical) data:
+
+- `ObjectCore.updates` — fires on **real** data changes only (local writes, remote sync).
```
dmaretskyi, line 106: implementing `Obj.getVersion(obj, version) -> snapshot` instead of pining a version would eliminate this complexity

packages/core/echo/echo-client/docs/VERSIONING.md
```diff
@@ -0,0 +1,305 @@
...
+## Branching
+
+### The registry (protocol)
+
+`DatabaseDirectory.branches` (`SpaceBranchRegistry`) is a synced map on the space root:
+
+```ts
+branches[rootObjectId][branchName] = {
+  members: { [objectId]: docUrl },  // one branch doc per subtree member
+  baseHeads?: string[],             // root's heads at fork time (provenance)
```
dmaretskyi: members points to many documents, but baseHeades is likely heads of just one document. Which one is it?

packages/core/echo/echo/src/Database.ts
```diff
@@ -94,6 +94,19 @@ export type FlushOptions = {
   updates?: boolean;
 };
 
+/**
+ * A caller-owned, writable per-surface binding to one branch of one object.
+ * @see Database.branch
+ */
+export type BranchBinding<T extends Obj.Unknown = Obj.Unknown> = {
```
dmaretskyi, line 101: couldn't we collapse concept by having objects have a branch property (via `Obj.getBranch(obj)`)?

packages/core/echo/echo/src/Database.ts
```diff
@@ -185,6 +198,47 @@ export interface Database extends Queryable {
...
   flush(opts?: FlushOptions): Promise<void>;
 
+  //
+  // Branching. A branch is a writable alternate timeline of an object subtree (same object ids,
+  // shared automerge history, true CRDT merge-back). The registry is synced on the space root;
+  // the currently-viewed branch stays device-local.
+  //
+
+  /** The branch name this device currently views the object on (`'main'` by default). */
+  getCurrentBranch(objectId: string): string;
```
dmaretskyi, line 208: should be at lease Obj.ID,, better is to take an object, even better is to put it onto `Obj.getBranch`

packages/core/echo/echo/src/Database.ts
```diff
@@ -185,6 +198,47 @@ export interface Database extends Queryable {
...
+  deleteBranch(rootObjectId: string, name: string): void;
+
+  /**
+   * Create a caller-owned, writable binding to one branch of one object — a live object whose reads
+   * resolve the branch document and whose writes land on the branch document only. Multiple bindings
+   * to different branches of the same object may coexist; the device-global current branch and other
+   * bindings are unaffected. Binding to `'main'` returns the canonical live object. Bindings are
+   * ephemeral and never persisted — the caller must `dispose()`.
+   */
+  branch<T extends Obj.Unknown>(obj: T, name: string): Promise<BranchBinding<T>>;
```
dmaretskyi, line 240: I dont get how this api works with `getCurrentBranch` --- do we use branches by creating a new idnependent copy of an object, or do we checkot branches on a single shared instance of an object. This API implies both which is odd.

packages/core/echo/echo-client/docs/VERSIONING.md
richburdon: A single automerge document is one converged state — any change applied to it is immediately part of that doc's history, so you can't hold two independently-editable, separately-syncing timelines of the same object inside it: they would converge, never diverge, and there's nowhere to keep a draft that is edited in isolation and merged on demand. Giving each branch its own document (a history-preserving `A.save`/`A.load` fork) is what lets the timelines evolve independently and later reconcile as a true `A.merge` 3-way merge over shared ancestry. Added a NOTE to the doc making this explicit.

packages/core/echo/echo-client/docs/VERSIONING.md
richburdon: Agreed — `Obj.getVersion(obj, heads) -> Snapshot` is the right target and the pin should be retired. The pin time-travels the live object everywhere at once (viewing v1 in the editor would also rewind the doc title in the sidebar, tabs, and cross-document backlinks), whereas a snapshot scopes the historical view to the surface that asked — which is the behaviour we want, and what the markdown editor already does (it renders a detached snapshot because binding CodeMirror to the pinned doc broke). The cross-surface hand-off needs no pin: History companion sets the selected version -> shared `VersioningState` selection atom -> the article re-renders `Obj.getVersion(...)`. Retiring the pin also drops the dual channels + `latestOnly`/`withLatestRead`, which exist only to stop the global pin leaking into side-effecting subscribers. Captured as a NOTE; will refactor to the snapshot API after this review pass.

packages/core/echo/echo-client/src/core-db/object-core.ts
richburdon: Agreed — `Obj.getVersion(obj, heads) -> Snapshot` is the right target and the pin should be retired. The pin time-travels the live object everywhere at once (viewing v1 in the editor would also rewind the doc title in the sidebar, tabs, and cross-document backlinks), whereas a snapshot scopes the historical view to the surface that asked — which is the behaviour we want, and what the markdown editor already does (it renders a detached snapshot because binding CodeMirror to the pinned doc broke). The cross-surface hand-off needs no pin: History companion sets the selected version -> shared `VersioningState` selection atom -> the article re-renders `Obj.getVersion(...)`. Retiring the pin also drops the dual channels + `latestOnly`/`withLatestRead`, which exist only to stop the global pin leaking into side-effecting subscribers. Captured as a NOTE; will refactor to the snapshot API after this review pass.

packages/core/echo/echo-client/docs/VERSIONING.md
richburdon: `baseHeads` is the root object's main-doc heads at fork time — not per-member. It's provenance only: currently written but never read anywhere (the merge relies on shared automerge ancestry, not this field). Clarified the comment on both the doc and the `BranchRecord` type.

packages/core/echo/echo/src/Database.ts
richburdon: Agree — `Obj.getBranch(obj) -> string` collapses the `branch` field and works for both the canonical object (returns its device-current branch) and a `db.branch()` instance. The one piece that can't move onto the object is `dispose()` (it releases a doc-handle listener), so `db.branch()` will return `{ object, dispose }` and `Obj.getBranch(object)` reports the name. Also fixing the misleading "per-surface" wording -> "independent instance" (it's an object instance, not a UI surface). Captured as a NOTE; folding into the versioning refactor.

packages/core/echo/echo/src/Database.ts
richburdon: Yes — `getCurrentBranch(objectId: string)` becomes `Obj.getBranch(obj)`, taking the object (not a raw id) and unifying with the per-instance accessor from #5. The string-id signature goes away.

packages/core/echo/echo/src/Database.ts
richburdon: Both are intentional and serve different needs, but you're right it's under-explained. Two models: `switchBranch` is a device-global **checkout** — it rebinds the single shared canonical instance (+ subtree) to the branch, persisted via BranchStore; `db.branch(obj, name)` mints an **independent object instance** bound to the branch doc, ephemeral and disposed, leaving the device selection untouched. Checkout = "I move my whole view to a branch"; independent instance = "an agent edits a draft while I stay on main / two panes show two branches at once." Keeping both, documenting the split explicitly (added a NOTE with a two-model table) and unifying the name accessor on `Obj.getBranch` (see #5/#6).

richburdon:
@dmaretskyi — thanks for the review. Wrote up the target API from your comments before refactoring, to sanity-check the shape:

**A — retire the pin for a functional snapshot** (your #2/#4)
- Add `Obj.getVersion(obj, heads) → Obj.Snapshot` (pure, on top of the existing `checkoutVersion`/`A.view`); returns a detached instance, no core state.
- Delete `ObjectCore.setTimeTravel`/`clearTimeTravel`/`#getReadDoc`, and the whole dual-channel stack it forces (`displayUpdates`, `LatestEventId`, `latestOnly`, `withLatestRead`, `Entity.timeTravelAtom`/`isTimeTraveling`, `TimeTravelingId`).
- `@dxos/versioning` `Version.view`/`restore`/`contentAt` read a snapshot; `plugin-markdown` already renders a detached snapshot in the editor, so it just drops the pin effect. Cross-surface hand-off is unchanged (companion → `VersioningState` selection atom → article re-renders `Obj.getVersion`).

**B — unify the branch accessor** (your #5/#6)
- Add `Obj.getBranch(obj) → string`; change `db.branch()` to return `{ object, dispose }` (name via `Obj.getBranch`); replace `getCurrentBranch(objectId)` with `Obj.getBranch(obj)`. One external caller to update (`plugin-comments/threads.ts`).

**C — terminology** (your #7): "per-surface" → "independent instance" in code JSDoc; the two branch models (device-global checkout vs independent instance) are now documented in VERSIONING.md.

Open question for you: **land #12256 as the minimal core (current pin design) and do this as a follow-up PR, or apply it here so we land the snapshot design once?** Estimate ~4–5 eng-days; Change A (pin removal) is the risky part since it deletes reactivity-layer machinery, so I'd do A and B as separate commits. Full design/estimate in `agents/superpowers/specs/2026-07-18-versioning-snapshot-refactor-design.md`.

PR #12266, opened 2026-07-19
typescript: update to 7.0.2 (major — still blocked, needs review)

dmaretskyi:
## Summary

Periodic dependency sweep — `typescript` group. Refreshed onto current `main` (2026-08-16) and re-confirmed: nothing has changed since the last check (2026-07-19/26). `typescript` is pinned at `^6.0.3` — already the newest 6.x release — with `7.0.2` (major, the Go-native compiler rewrite) the only newer published version.

## Validation result: both previously-found breaks reproduce byte-for-byte

- **Install**: succeeds, resolves `typescript` 6.0.3 → 7.0.2 cleanly.
- **Build**: fails, same error as every prior check: `dx-build:build | Unexpected error: TypeError: Cannot read properties of undefined (reading 'fileExists')`. Root cause (unchanged): `tools/dx-build/src/main.ts` calls `ts.findConfigFile(process.cwd(), ts.sys.fileExists, ...)`, and TypeScript 7 doesn't expose `ts.sys` in the same shape. `dx-build` is a shared prebuild step most packages' build tasks depend on.
- **`pnpm run check-cycles` still crashes independently**, same as before: `@typescript-eslint/typescript-estree@8.39.0` reads `ts.Extension.Cjs` at import time, undefined under TS7.
- **Re-verified today**: `@typescript-eslint/typescript-estree`'s published peer range is still `"typescript": ">=4.8.4 <6.1.0"` — no release supports TS7 yet.

## Why this needs review

- Major compiler version with well-documented, unchanged breaking changes.
- Two concrete, independent, reproducing breaks: two in-repo call sites (`dx-build`, `protobuf-compiler`) built against the classic Compiler API shape, plus one upstream blocker (`@typescript-eslint`) outside this repo's control.
- Nothing has changed in the ~4 weeks since this was last checked — still architecturally significant enough that it shouldn't be a speculative fix from an automated sweep.

Not auto-merging.

## Summary by CodeRabbit

* **Chores**
  * Updated the TypeScript version constraint used by the project.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1760,7 +1760,7 @@ overrides:
   react-dom: ~19.2.7
   tar: '>=7.5.7'
   tar-fs: '>=3.1.1'
-  typescript: ^6.0.3
+  typescript: ^7.0.2
   undici: '>=7.18.2'
   vite: '>=8.1.4'
   web-ext: ^10.5.0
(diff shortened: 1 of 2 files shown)


dmaretskyi:
Confirmed — verified `substitutions-parser.ts:36` has the same `ts.sys.fileExists` pattern. Agreed this shouldn't merge as-is; updated the PR description with the full picture (both in-repo call sites, the `@typescript/typescript6` compat package pointer, and the separate `@typescript-eslint` upstream blocker found by CI). Leaving this for a human to pick an approach rather than guessing — not auto-merging.

---

dmaretskyi:
Closing as superseded: `main` now carries `typescript` catalog pin `^7.0.2`, which is npm's current `latest` dist-tag — the full version this PR was proposing. It looks like this landed independently of this sweep (alongside the `@typescript/typescript6` compat shim added in #12886), not through this PR's branch, so this diff is now a no-op / stale duplicate rather than a live proposal.

No action needed for the `typescript` group in this sweep — it's already at latest.

---
_Generated by [Claude Code](https://claude.ai/code/session_017W2LbzUkEU69kRj1dbz4AA)_

https://claude.ai/code/session_017W2LbzUkEU69kRj1dbz4AA

---

PR #12272, opened 2026-07-19
plugin-theme: add appearance setting (light/dark/system override)

Steve:
## Summary

Adds an **Appearance** setting to Composer (`plugin-theme`) so the web app can be forced to Light or Dark independent of the OS `prefers-color-scheme`.

- New `appearance: 'light' | 'dark' | 'system'` setting, default `system`.
- `system` keeps the existing OS-following behavior; `light`/`dark` override it.
- Persisted per-device via `createKvsStore` (localStorage), like every other plugin setting.
- Rendered as a Light/Dark/System select in the Settings dialog by the existing generic `DefaultSettings` form — no custom settings UI needed.

## How it works

`plugin-theme` now contributes a KVS-backed settings atom as both a plugin-local capability (`ThemeCapabilities.Settings`) and `AppCapabilities.Settings`. `react-context.tsx` reads that atom and recomputes the effective theme (the `dark` class + `themeAtom`) from three live sources, all torn down on deactivation:

1. the OS `matchMedia('(prefers-color-scheme: dark)')` `change` event (only visible while on `system`),
2. an in-tab subscription to the settings atom (the Settings select), and
3. a cross-tab `storage` listener — `Atom.kvs` doesn't observe the `storage` event, so this keeps the theme in sync across all tabs of the same browser.

`ReactContext` now declares `firesBeforeActivation: [SetupTranslations, SetupSettings]` so the settings atom exists when it activates. The persisted value is applied synchronously at activation to avoid a flash on load.

## Scope decisions

- **Per-device only.** Matches all 27 other plugins' settings (all localStorage-backed); cross-device sync would need an ECHO/HALO-backed store and is out of scope.
- **No custom settings UI.** The generic `DefaultSettings` renders the schema, so this adds no container/surface and no new `@dxos/react-ui-form` dependency.
- **No new i18n strings.** Field/option labels come from schema annotations, as in `plugin-deck`.

## Testing

`ThemePlugin.test.ts` (7 passing) covers: setting contributed + defaults to `system`; dark override under a light OS; light override under a dark OS; `system` follows the OS; and the cross-tab `storage` path. Verified end-to-end in the running app: override flips the theme live, the choice persists across reloads, and switching back to `system` re-follows the OS.

## Summary by CodeRabbit

* **New Features**
  * Added an appearance setting with Light, Dark, and System options.
  * Appearance defaults to System and can override the device’s color-scheme preference.
  * Theme selections are saved per device and update live across browser tabs.

* **Tests**
  * Added coverage for appearance overrides, system preference behavior, and cross-tab synchronization.


Steve:
### CI note: the `storybook` failure is pre-existing on `main`, not from this PR

- `check` ✅ and `test` ✅ both pass on this PR (incl. `plugin-theme`'s tests).
- The only red job is `storybook`, which fails on **`plugin-blogger`'s `src/containers/PostArticle/PostArticle.stories.tsx > Default`** — a client-capability init race (`No capability found for org.dxos.plugin.client.capability.client`) plus flaky browser-test infra (`Missing file /__vitest_browser__/orchestrator-*.js`, `Port … in use`).
- This branch does **not** touch `plugin-blogger` (or `plugin-client`). Because `plugin-theme` is a core plugin, editing it pulls nearly every plugin's storybook into the "affected" set, so this PR merely re-runs the same test.
- The **identical** failure is present on upstream `main` — e.g. Check run [29691542339](https://github.com/dxos/dxos/actions/runs/29691542339) shows `check` ✅ / `test` ✅ / `storybook` ❌ on the same `plugin-blogger PostArticle` story (6 of the last 8 `main` Check runs are `failure`).

Not fixing `plugin-blogger`'s storybook here, since it's an unrelated, pre-existing breakage on `main`.

Steve:
Superseded by #12281 — moved onto a dxos/dxos branch so maintainers can push directly.

PR #12285, opened 2026-07-20
ai: add scripted mock AI model for tool-contract tests

wittjosiah:
## Context

Tool-contract tests (that tools/operations execute correctly *within a conversation*) currently record real LLM conversations to `<test-file>.conversations.json` and replay them by **structural deep-equality of the entire prompt** — system prompt + every message + every tool's `name`/`description`/`inputSchema`. Any edit to a tool description, an input schema, or a system-prompt template changes the serialized prompt → the memo no longer matches → `"No memoized conversation found"`. Fixing it means regenerating against a live model (needs `DX_ANTHROPIC_API_KEY`, whole-file runs to keep the shared PRNG aligned, then committing large JSON blobs). This is a major source of development friction.

This PR moves the conversation **into the test** as a statically-specified, type-checked mock model, so tests evolve with the code and a schema change surfaces as a **compile error** rather than a stale fixture. The approach is [the canonical Effect-AI testing pattern](https://effect.website/docs/ai/introduction/) (provide an alternative `LanguageModel`/`AiService` implementation) and mirrors the Vercel AI SDK's `MockLanguageModel`.

This is **Phase 0 + Phase 1** only (build the harness + prove it on 3 representative tests). Migrating the remaining tests and removing the memoization engine is deferred to a follow-up.

## What's here

**Harness (`@dxos/ai/testing`) — `ScriptedAiService`**
- Implements `AiService` with no upstream model, no API key, no `TestContextService`.
- Requests are matched **sequentially within a `(model, stream)` bucket**, so the streaming main loop, a non-streaming summarizer, and a different-model side-call (e.g. the planning plan-reminder on sonnet) never collide — without any full-prompt matching.
- Both stream and non-stream part synthesis. Streaming tool turns emit **both** raw `tool-params-*` and a resolved `tool-call` part, so DXOS's `AiParser` and `@effect/ai`'s built-in tool resolution both execute the call (a real provider emits both).
- Tool-call `input` may be a **lazy function of the turn context**, for refs to objects created at run time.
- Clear diagnostics on script exhaustion / guard rejection (bucket, cursor, available tools, last message).

**Typed builder (`@dxos/agent-runtime/testing`) — `operationToolCall(op, input)`**
- Derives the model-facing tool name from the operation (via the newly-exported `toolNameForOperation` in `@dxos/assistant`) so it can't drift from what the agent registers.
- `input` is type-checked against the operation's input schema — the headline win.

**Converted tests (run offline, no `ALLOW_LLM_GENERATION`, no API key):**
- `operations/run-instructions.test.ts` (2) — text + inline `completeJob` tool + 2-turn loop.
- `skills/planning/skill.test.ts` (4) — exercises the sonnet plan-reminder side-call, validating `(model, stream)` bucketing.
- `skills/database/skill.test.ts` (20) — the richest contract test; `operationToolCall` typing + lazy refs, asserting on real ECHO DB side-effects.
- Their `.conversations.json` fixtures are deleted.

## Verification

- `moon run @dxos/ai:test -- src/testing/scripted/scripted.test.ts` — 9 engine unit tests pass (synthesis, bucketing, cursor, guard, exhaustion).
- All 3 converted files pass via vitest offline (26 tests).
- `@dxos/ai`, `@dxos/assistant`, `@dxos/agent-runtime` build cleanly. The full typecheck of the converted test files runs in CI Check.

## Not in this PR (follow-up)

- Migrating the remaining ~12 tool-contract test files.
- Removing the memoization engine, the `regenerate-memoized-llm` skill, and remaining fixtures.
- A codemod to convert `.conversations.json` → starter scripts.
- `generateObject`-path support (qualifier / proper-noun extraction).

---
_Generated by [Claude Code](https://claude.ai/code/session_01RxSWUKkUGkF9vNYGUc8c2s)_


packages/core/compute/ai/src/AiSummarizer.test.ts
```diff
@@ -8,16 +8,32 @@ import * as Layer from 'effect/Layer';
...
+// `AiSummarizer.summarize` issues a single non-streaming `generateText` call, so the response is
+// scripted under the model's `generate` bucket rather than the (streaming-only) top-level `turns`.
 const TestLanguageModel = AiService.model('com.anthropic.model.claude-sonnet-4-6.default').pipe(
-  Layer.provide(TestAiService()),
+  Layer.provideMerge(
+    ScriptedAiService.layer({
+      models: {
+        sonnet: {
+          generate: [
+            ScriptedAiService.text(
```
richburdon: use @dxos/util trim

richburdon (changes requested):

PR #12369, opened 2026-07-28
composer-app: derive serve-min deps and importSource excludes from package metadata

wittjosiah:
## Problem

The fact "these `@dxos` packages must be consumed from `dist`, not source" was hand-written in three places with no drift detection:

1. `composer-app/moon.yml` — the `serve-min` task's 17 hardcoded `:build` deps
2. `composer-app/vite.config.ts` — the `importSource` `exclude` list
3. `storybook-react/.storybook/main.ts` — the same exclude list duplicated verbatim

A new dist-only package (or a new source-unsafe one) required remembering to edit all three, and a miss surfaced only as a mysteriously broken dev server.

## Change

Each package declares its own resolution, and everything else is derived.

**Packages consumed from dist publish no `source` export condition.** `random-access-storage`, `lock-file`, `network-manager`, `teleport`, `config`, `client-services` and `observability` drop it (14 conditions removed), so node, bun and Vite all resolve them the same way with no Vite-side special case. Both app-local `exclude` lists are deleted, and `vite-plugin-import-source` is unchanged from `main` — this PR no longer touches it.

**Two moon tags replace the dependency lists.** `dist-runtime` marks packages whose dist is consumed at runtime (the seven above plus `shell`, `protocols`, `lit-grid`, `lit-ui`, which never had a `source` condition). `vite-plugin` marks packages that *export a vite plugin `vite.config.ts` loads* — `app-framework`, `ui-theme`, `config`, and the four `vite-plugin-*` tools. `serve-min`'s deps collapse from 17 entries to `'#dist-runtime:build'` + `'#vite-plugin:build'`. A non-plugin import like `@dxos/util` stays untagged: the plugin packages depend on it, so their own `^:build` builds it first.

**`optimizeDeps` plugin scanning is derived too.** The hardcoded brace list in `vite.config.ts` had already drifted from the registries — missing `plugin-tasks` and `plugin-progress`, still naming a removed `outliner`. It now reads the registry sources directly.

**pkg-lint enforces the invariant:** `dist-runtime` tag iff no `source` export condition. It also stops warning `no-source-export` for tagged packages, where omitting `source` is deliberate.

## Verification

- `moon task composer-app:serve-min` resolves both tags to exactly the dependency set the hardcoded list produced.
- `moon run composer-app:serve-min` boots (7.5 s) and serves HTTP 200; a live request confirms `client-services/dist/lib/index.mjs` in the served graph while `@dxos/echo` still resolves to source.
- `pnpm run pkg-lint` passes, and negative-testing a removed tag makes the rule fire and exit 1.
- The derived plugin glob yields 27 plugins, each verified to resolve to a real entry file.

## Notes

Trade-off accepted per review: `--conditions=source` consumers (`DX_SOURCE=1`, `introspect-mcp`'s `tsx --conditions=source`) now get built output for these seven packages, so iterating on their TypeScript requires a build. The `imports` field still carries `source` for `#tcp-transport`, `#locks-platform` and `#storage` — consistent across runtimes, and a wider change if we want those too.

Changeset included: these packages publish `src` in their tarballs, so removing the `source` condition is observable to consumers resolving with `--conditions=source`.

https://claude.ai/code/session_01HyFLgNCYKmkLiHjnssNNBy


wittjosiah:
**CI triage** — the two red storybook runs were pre-existing flakiness in brand-new tests, not caused by this PR's resolution change:

- **`stories-inbox CreateProject > Test`**: fails locally *even with this PR's resolution change fully neutralized* (old exclude lists restored, opt-out disabled). Root cause: the play function waited for the button with `waitFor`'s default **1 s** timeout, but the button only renders after full client init (identity + space + indexed flush), which takes ~6–7 s cold — while the same play already allows 10 s for the next assertion. Fixed in `261ad06c` (button wait now gets the same 10 s budget); verified locally.
- **`plugin-terra TerraArticle > Objects`**: passes locally on this branch (824 ms vs the 15 s test timeout), passed on main's storybook job at 12:42, and has no dependency on any package whose resolution this PR changes — `makeDemoWorld` is pure synchronous grid/pathfinding work, so this reads as runner-load timeout. Watching the fresh run; if it keeps timing out it's a quarantine candidate ([Trunk page](https://app.trunk.io/dxos/flaky-tests/test/f2fe0fac-7d75-5579-b072-702f414c865e?repo=dxos%2Fdxos)).

For confidence in the PR itself: request-time smoke test shows opted-out packages resolving to `dist` and everything else still resolving to source, and `moon task composer-app:serve-min` expands `#dist-runtime:build` to exactly the old hardcoded dep set.

---

wittjosiah:
Per review feedback, the `stories-inbox CreateProject` timeout fix has been split out of this branch into its own PR: **#12372**. My earlier triage comment referenced commit `261ad06c`, which is now reverted here (`932e4409`) — this branch's net diff is once again the build-dependency change only, no `packages/stories/` changes.

The triage conclusion is unchanged: that test failure was independent of this PR (reproduced with this PR's resolution changes fully neutralized).

---

wittjosiah:
**Correctness fix pushed (`689afc9e`)** — the Codecov comment on this PR (−0.14%, −3 files, −86 hits) was a real signal, not noise, and it caught the refactor changing behavior it was meant to preserve.

**The bug.** The `dx.importSource: false` check keys on the *resolved package*, but the exclude lists it replaces were minimatch patterns tested against the import *specifier*. A pattern like `@dxos/config` matches neither `@dxos/config/vite-plugin` nor `#tcp-transport`, so those stayed on source. Keying on the package captured them:

| Specifier | Before this PR | As first pushed | After `689afc9e` |
| --- | --- | --- | --- |
| `@dxos/config` | dist | dist | dist |
| `@dxos/config/vite-plugin` | source | **dist** ✗ | source |
| `@dxos/teleport/testing` | source | **dist** ✗ | source |
| `#tcp-transport` (inside `network-manager`) | source | **dist** ✗ | source |

The `#*` row is the one that matters: sending package-internal subpath imports to dist is the exact failure the exclude-list comments warn about, since `dist/lib/**` is absent until the package is compiled. It was latent here — all checks were green — which is why coverage was the only thing pointing at it. The coverage delta itself was the `/testing` subpaths (28 call sites) being served from uninstrumented dist instead of `src`.

**The fix** restricts the opt-out to bare specifiers, so `#*` and subpaths keep reaching source exactly as before.

**Verification.**
- Decision-table equivalence across 16 representative specifiers — old patterns vs. new logic now agree on every case.
- Probed the **real built plugin** (oxc-resolver included), which is what distinguishes the two: with the fix, `#tcp-transport` → `packages/core/mesh/network-manager/src/transport/tcp/index.ts`; against the previously-committed version the same probe falls through to dist. Bare `@dxos/client-services` and `@dxos/config` still opt out; `@dxos/echo` still resolves to source.
- `pnpm run pkg-lint` green; a storybook run over the client-backed story path is still finishing locally and I'll follow up if it disagrees.

Worth flagging for review: `@dxos/vite-plugin-import-source` has no test task, so these semantics aren't currently locked by CI. Adding one is a reasonable follow-up — I left it out here to avoid bolting new test infrastructure onto this PR.

---

packages/core/mesh/network-manager/package.json
```diff
@@ -98,5 +98,8 @@
         "root": "src/network-manager.ts"
       }
     ]
+  },
+  "dx": {
```
dmaretskyi: why not just omit the "source" condition for these packages instead of a special flag?


I do like the moon tag though -- its the right direction.

dmaretskyi (dismissed):
I wanna move the standard serve + test jobs to this tag where build is required despite import source. but it doesn't have to be this pr

packages/core/mesh/network-manager/package.json
wittjosiah: Good question — and it's already the mechanism for 4 of the 11 tagged packages. `shell`, `protocols`, `lit-grid` and `lit-ui` have no `source` condition at all, so they carry only the tag and no flag; resolution falls through to dist naturally. So the flag exists purely for the 7 packages that *do* publish `source`.

For those 7, the `source` condition is load-bearing on the **Node** side, which is what makes deleting it lossy rather than equivalent:

- `packages/devtools/cli/bin/dx` runs `bun run --conditions=source` under `DX_SOURCE=1` ("useful when iterating on TS sources without a build").
- The CLI depends on `client-services`, `config` and `lock-file` directly, and reaches `network-manager`, `teleport` and `random-access-storage` transitively through `client-services` — 6 of the 7.
- `introspect-mcp` does the same via `tsx --conditions=source` (its `start` script, smoke/inspect scripts, and two tests assert that exact launch path).

And their source is genuinely fine in Node — `random-access-storage`'s raw `node:path` is *why* it's browser-unsafe, not a reason it can't run under bun. So "unsafe" here means "unsafe in Vite's browser pipeline", not "unsafe everywhere", and that's the distinction the flag encodes: it scopes the opt-out to the vite resolver instead of withdrawing source resolution from every Node consumer. Dropping the condition would trade a vite-only problem for a regression in `DX_SOURCE=1` and the MCP server's documented dev launch.

`observability` is the one where your suggestion may well be viable — it's the only flagged package no `--conditions=source` consumer appears to reach. Happy to drop the condition there and remove its flag if you want the list shorter; the test is simply "does any Node source-conditions consumer reach it".

On moving the standard `serve` + `test` jobs onto the tag — agreed that's the right direction, and noted as follow-up rather than this PR.

---

dmaretskyi: just kick the unsafe packages, from the import-source path, by removing the source coditions. random-access-storage is almost never updated, so we dont benefit much from the source condition anyway. I want node/bun and vite to be the same regarding which packages use source

wittjosiah: Done — `ab3fd8eb`. The 7 packages now publish no `source` export condition (14 conditions removed), so node/bun and vite resolve them identically with no vite-side special case.

Knock-on effects, all in the simplifying direction:

- **`dx.importSource` is gone**, and with it the plugin change that read it — `vite-plugin-import-source` is now byte-identical to `main`, so this PR no longer touches it at all.
- **pkg-lint's invariant collapsed** from a three-way flag/tag/condition agreement to: `dist-runtime` tag iff no `source` condition. It also stops emitting the pre-existing `no-source-export` warning for tagged packages, since omitting `source` is deliberate there (untagged ones like `node-std` still warn).
- Net **−98/+21** across 12 files.

Verified: pkg-lint green; the built plugin resolves `@dxos/client-services` and `@dxos/config` to their `dist` while `@dxos/echo` still goes to source; `serve-min` boots (7.5 s) and a live request confirms `client-services/dist/lib/index.mjs` in the served graph.

Two notes:

1. **`imports` still carry `source`** — `#tcp-transport`, `#locks-platform`, `#storage`. Those still resolve to source, and consistently so across node/bun and vite, so the invariant you asked for holds. Kicking those too is a bigger blast radius (it changes what `#*` resolves to everywhere), so I left them; say the word if you want a second pass.
2. **The accepted cost is now live**: `DX_SOURCE=1` and introspect-mcp's `tsx --conditions=source` take dist for these 7, so iterating on their TS requires a build.

---

wittjosiah:
@coderabbitai review

(The previous run aborted because the head moved mid-review — `3cc0d6c8` → `46467b1e` was the changeset commit. The head is stable now.)

---

PR #12373, opened 2026-07-28
observability: send $survey_completed and fix feedback log uploads

wittjosiah:
Fixes two independent reasons Composer feedback wasn't landing as expected in PostHog.

## 1. Notifications never fired — `$survey_completed` was missing

PostHog survey destinations carry an implicit trigger filter on `$survey_completed = true`. Composer's survey is API type (we capture `survey sent` ourselves), and we sent `$survey_id`, `$survey_questions`, and the response — but not `$survey_completed`. Every event was skipped before it reached the "Only notify when" filters, so no Slack/webhook notification ever fired.

Added the property in `posthog/extension.ts`.

## 2. `debug_log_dump_key` was always `'failed'`

SigNoz shows 6 records over 7 days, all from `extension.ts:51` — the `catch` branch of `uploadLogs`, never the `!response.ok` branch. So the `fetch` **rejected**; there was no HTTP status at all.

Ruled out the competing explanation (the SPA fallback swallowing `/api/*`, which would also land in that catch via a non-JSON 200): production sets `DX_OTEL_ENDPOINT=/api/otel`, and those same-origin browser OTel records are arriving in SigNoz, so `run_worker_first: ["/api/*"]` demonstrably works and the POST does reach the Worker.

That points at the request itself, and the Worker was buffering the whole body with `await request.arrayBuffer()` before handing it to R2 — up to 50 MB held in an isolate limited to 128 MB. A Worker torn down that way resets the connection, which is exactly what the client sees.

Changes:

- Hand `request.body` straight to `env.FEEDBACK_LOGS.put()` so the body never lands in isolate memory. R2 accepts it because a request body carries a known length from `Content-Length`.
- Require `Content-Length` (411 without it): it is both the size guard and what makes the unbuffered put legal.
- Return 502 when R2 fails, rather than letting it escape as an unhandled exception.

### An earlier attempt in this PR was wrong

The first version piped the body through a byte-counting `TransformStream` into R2. That **failed every upload with a 502** — R2 will not accept a value whose length it cannot determine, and a `TransformStream` readable has none. A fake-R2 unit check couldn't model that constraint; the PR preview did, with `main.composer.space` on the same bucket returning 200 as a control. The committed version is simpler as a result: no transform, no orphaned pipe.

## 3. Why this was hard to diagnose — logged errors were being dropped

`log.warn('feedback log upload error', { error: err })` reached SigNoz with no `ctx_error` attribute. `stringifyValues` walks `Object.entries(...)` and falls back to `JSON.stringify` — and an `Error` has non-enumerable `message`/`stack`, so both paths reduce it to nothing. Errors are now serialized via their stack at any nesting depth, so the next occurrence names its own cause.

## Upload cap is tied to log retention

The upload limit and the log store's retention cap were both 50 MB, but only coincidentally — composer hardcoded its own constant while the store fell back to the package default, and either could have moved independently. They're now one constant, `LOG_STORE_MAX_BYTES`, used as the store's `maxBytes` (passed explicitly at every `IdbLogStore` site instead of relying on the default), the client's export trim, and the Worker ceiling. A feedback report therefore carries everything the circular buffer retained.

`maxBytes` is a soft cap in the store — eviction drops whole chunks and always keeps the newest — but `export({ maxSize })` hard-trims to an exact UTF-8 byte count, so an upload is always within the cap and the Worker's check never trips on a legitimate one.

## Testing

- `moon run observability:test` — 60 passed, 1 skipped. Includes a new `logs.test.ts` covering error serialization (it caught a real hole in the first attempt: errors nested past the flatten depth still went through `JSON.stringify`).
- `moon run observability:build`, `observability:lint`, `composer-app:lint` — clean.
- `pnpm build:functions` bundles the worker.
- Endpoint exercised end-to-end against the PR preview (dev-tier bucket), which is what caught the R2 stream-length bug above:

  | Body | Result |
  | --- | --- |
  | small NDJSON | 200 + key |
  | 8 MB | 200, 5/5 attempts |
  | 52,428,800 B (exactly the cap) | 200 + key |
  | 52,428,801 B (cap + 1) | 413 Payload too large |
  | `GET` | 405 |

  So an at-cap upload — the realistic worst case now that the client trims to the retention cap — completes, and going over is refused with a status rather than a reset connection.
- The pre-existing `src/functions/tsconfig.json` `rootDir` errors are unchanged from the baseline (verified against a stash); that config can't reach `../util/constants`, and CI only esbuild-bundles the worker.

Note: the root-cause claim for #2 is inferred, not captured — the actual error text was destroyed by the logging bug in #3, so the memory argument is a plausible mechanism rather than a demonstrated one. With #3 fixed, any recurrence will report its own cause.

## Summary by CodeRabbit

* **New Features**
  * Survey feedback submissions now include the required completion indicator so responses reach configured destinations.
  * Observability feedback and error serialization improvements ensure richer event payloads.

* **Bug Fixes**
  * Feedback and recovery logs now enforce a consistent 50 MB storage limit across upload and export flows.
  * Feedback-log uploads validate `Content-Length`, reject empty/invalid/oversized requests, and improve error responses.

* **Tests**
  * Added a test suite covering log value stringification, including correct `Error` handling and flattening behavior.

### DIFF
--- a/.changeset/survey-completed-property.md
+++ b/.changeset/survey-completed-property.md
@@ -1,0 +1,5 @@
+---
+'@dxos/observability': patch
+---
+
+Send `$survey_completed` with captured user feedback so PostHog survey destinations trigger, and serialize errors logged as context to OTel instead of dropping them.
--- a/packages/apps/composer-app/src/functions/_worker.ts
+++ b/packages/apps/composer-app/src/functions/_worker.ts
@@ -47,7 +47,18 @@ const handleFeedbackLogs = async (request: Request, env: Env): Promise<Response>
     return new Response('Feedback logs storage not configured', { status: 503 });
   }
 
-  const contentLength = Number(request.headers.get('content-length') ?? 0);
+  // R2 only accepts a known-length stream, so Content-Length is required rather than advisory: it
+  // is both the size guard and what lets the body go straight to R2 unbuffered.
+  const contentLengthHeader = request.headers.get('content-length');
+  const contentLength = contentLengthHeader === null ? Number.NaN : Number(contentLengthHeader);
+  if (!Number.isInteger(contentLength) || contentLength < 0) {
+    return new Response('Content-Length required', { status: 411 });
+  }
+
+  if (contentLength === 0) {
+    return new Response('Empty body', { status: 400 });
+  }
+
   if (contentLength > FEEDBACK_LOGS_MAX_BODY_SIZE) {
     return new Response('Payload too large', { status: 413 });
   }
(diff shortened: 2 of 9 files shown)


wittjosiah:
Thanks — one applied, one declined.

**`constants.ts:21` (drop `// 50 MB`)** — applied in 152bef75. The JSDoc above it already explains the constant, so the inline comment only restated the arithmetic.

**`_worker.ts:8` (use the `../util` barrel)** — not applied: it would break the Worker build. That import is deliberately narrow, and the three lines directly above it say why — the barrel re-exports `./config`, `./halo`, and `./storage`, which pull Automerge's wasm into the Cloudflare Worker bundle, and esbuild cannot load it.

Confirmed by switching the import to the barrel and running `pnpm build:functions`:

```
✘ [ERROR] Could not resolve "@dxos/app-framework"
    src/util/help.ts:5:53
✘ [ERROR] Could not resolve "@dxos/app-toolkit"
    src/util/help.ts:6:32
```

`_worker.ts` is bundled by esbuild for the Workers runtime rather than built with the app, so it can only import modules that are safe in that environment — hence importing the one constant from `../util/constants` directly. The barrel-import convention still holds for app code; this file is the documented exception.

Worth noting the ESLint tool run timed out in this review, so the guideline check ran without the surrounding context.

---

PR #12374, opened 2026-07-28
plugin-deck: keep L0 reachable when `closed` persists across the lg breakpoint

wittjosiah:
Fixes [DX-1141](https://linear.app/dxos/issue/DX-1141/l0-sidebar-disappears-permanently-after-rapid-horizontal-window-resize) — L0 sidebar disappears permanently after rapid horizontal window resize.

Preview: https://pr-12374-composer-main.dxos.workers.dev

## Problem

`sidebarState` has three values, and `closed` is unrecoverable at `lg` and above:

1. **`closed` hides L0 entirely.** At `lg`+, `packages/ui/ui-theme/src/css/layout/main.css` gives `[data-side='is'][data-state='collapsed']` a `w-(--dx-l0-size) start-0`, but `closed` gets neither — L0 stays at the base `-start-[100vw]`. `Main.tsx` also marks it `inert`.
2. **Every escape hatch is out of reach.** The floating toggles carry `lg:hidden` (`DeckViewport.tsx`), and the L0/complementary toggles live *inside* the sidebar that is off-screen and inert.
3. **It persists.** `sidebarState` is part of `StoredDeckState`, written through `createKvsStore` to localStorage, so a reload restores the broken state.

Both writers of `closed` are gated on a per-component `useMediaQuery('lg')` in `NavTreeContainer` — activating a nav item, or re-clicking the active L0 tab, while below `lg`. Resizing across 1024px mid-interaction persists `closed`, and widening back removes the only control that could undo it.

## Fix

`closed` only ever means "the small-breakpoint drawer is dismissed", so it is resolved to `collapsed` at `lg`+:

- `util/resolve-sidebar-state.ts` — the rule as a pure, unit-tested function.
- `containers/Deck/DeckRoot.tsx` — applies it to the context it provides, the single point feeding both `Main.Root` (via `DeckContent`) and `--main-sidebar-width` (via `DeckViewport`), so the width, the sidebar element and the navtree can't disagree.

This **derives** rather than rewrites the persisted value, deliberately:

- Below `lg`, `closed` is the only state that makes the off-screen drawer `inert` — `collapsed` also sits at `-start-[100vw]` there but stays focusable. Rewriting `closed → collapsed` on the way up to desktop would therefore leave a hidden-but-keyboard-reachable sidebar after a rotation back down.
- A reactive derivation self-heals whenever a stale `useMediaQuery` subscription writes `closed` mid-resize, whereas a one-shot repair effect races it.
- Below `lg`, `closed` stays meaningful and persisted, which is the existing (correct) behaviour.

Every remaining reader of the raw persisted value — `ToggleSidebarButton`, the `app-graph-builder` toggle action, and `layoutAtom` — treats `closed` as "not expanded" and toggles to `expanded`, so none of them get stuck.

### Resulting behaviour

Measured in the app (`data-state` / `inert` / sidebar x-position):

| step | viewport | `data-state` | `inert` | x |
| drawer open below `lg` | 800 | `expanded` | no | +8 (on screen) |
| dismiss it | 800 | `closed` | yes | −800 |
| widen to desktop | 1280 | `collapsed` | no | 0, width 72 (L0 rail) |
| back below `lg` | 800 | `closed` | yes | −800 (still dismissed) |

The round trip is idempotent: desktop always gets the rail back, and the dismissal is preserved below `lg` because the persisted value is never rewritten.

## Testing

Both tiers were verified against a reverted `DeckRoot` and fail there, so neither passes vacuously.

- **Unit** — 3 tests for `resolveSidebarState`. `plugin-deck:test`: 44/44.
- **Storybook** — a `SidebarClosedAtDesktop` story (`tags: ['test']`) whose `play` function asserts the sidebar is `collapsed`, not `inert`, and on screen. Without the fix it fails with `data-state="closed"`. `plugin-deck:test-storybook`: 14/14.
  - `sidebarState` is now a story arg, so the deck stories can render each sidebar presentation. The stories previously hard-coded `closed`, which at desktop meant they rendered with no sidebar at all.
  - Adds a `deck.sidebar` testid for the assertion.
- `plugin-deck:build` and `plugin-deck:lint` clean; `pnpm format` applied.

The change is contained to `plugin-deck` — no composer-app-level e2e, since the invariant is a layout concern the story tier can hold.

Includes a patch changeset for `@dxos/plugin-deck`.

### DIFF
--- a/.changeset/l0-sidebar-closed-at-lg.md
+++ b/.changeset/l0-sidebar-closed-at-lg.md
@@ -1,0 +1,5 @@
+---
+'@dxos/plugin-deck': patch
+---
+
+Fix the navigation sidebar disappearing permanently at desktop widths after a dismissal persisted from below the `lg` breakpoint.
--- a/packages/plugins/plugin-deck/src/containers/Deck/Deck.stories.tsx
+++ b/packages/plugins/plugin-deck/src/containers/Deck/Deck.stories.tsx
@@ -6,6 +6,7 @@ import { Atom } from '@effect-atom/atom-react';
 import { type Meta, type StoryObj } from '@storybook/react-vite';
 import * as Effect from 'effect/Effect';
 import React, { useEffect, useMemo, useRef } from 'react';
+import { expect, waitFor, within } from 'storybook/test';
 
 import { Capabilities, Capability, Plugin } from '@dxos/app-framework';
 import { withPluginManager } from '@dxos/app-framework/testing';
@@ -194,9 +195,11 @@ type DefaultStoryProps = {
   count?: number;
   /** Fold-transition variant to apply (see {@link FOLD_ANIMATIONS}). */
   foldAnimation?: FoldAnimation;
+  /** Navigation sidebar state to seed. `closed` is only reachable below `lg`. */
+  sidebarState?: StoredDeckState['sidebarState'];
 };
 
-const DefaultStory = ({ count = 0, foldAnimation = 'slide' }: DefaultStoryProps) => {
+const DefaultStory = ({ count = 0, foldAnimation = 'slide', sidebarState = 'closed' }: DefaultStoryProps) => {
   const settings = useAtomCapability(DeckCapabilities.Settings);
   const pluginManager = usePluginManager();
   const { graph } = useAppGraph();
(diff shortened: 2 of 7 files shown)


wittjosiah:
CI red on `storybook (24.11.1)`, but the failure is not from this PR — it's `plugin-terra` › `TerraArticle.stories.tsx` › `Objects`, timing out at 15s. I've re-run the failed job.

Why I'm confident it's unrelated:

- `plugin-terra` has no `@dxos/plugin-deck` dependency, and this diff is 7 files all inside `plugin-deck`.
- The rest of the run was green — 303 tests, 302 passed, 1 failed. This PR's own `plugin-deck` storybook tests passed.
- The failing story is a bare render smoke test (no `play`) that builds a Babylon.js WebGL world via `makeDemoWorld`. Its stdout contains only the Babylon banner, so it never got through scene init.

Reproduced locally on this branch, which shows it as a timeout flake rather than a real break: on the first attempt **all three** Terra stories exceeded the 15s vitest timeout while taking 31s / 90s / 68s of wall time; moon's third attempt passed. So these stories render slower than the timeout allows whenever the runner is loaded — several storybook browsers run concurrently at `MOON_CONCURRENCY=4`.

Trunk currently classifies the test as `HEALTHY` rather than flaky, so it isn't being quarantined. Worth a separate look at `plugin-terra` — either a longer `testTimeout` for the Babylon stories or quarantining them — but that's outside this PR's scope, so I haven't touched it here.

---

PR #12378, opened 2026-07-28
plugin-client: add account signup command

wittjosiah:
Adds `dx account signup <code>` alongside `dx account login`, mirroring the sign-up flow in plugin-onboarding's `WelcomeScreen`.

## Flow

```
dx account signup <code> [<input>] [--method email | atmosphere]
```

1. **Access code first.** The 8-character code is a required positional arg. It is format-checked locally (Crockford base32, hyphen optional — same rule as the gate's `validInvitationCode`), then validated against hub-service (`/account/invitation-code/validate`) before anything else happens.
2. **Then the method** — prompted if `--method` is omitted, same as `login`:
   - **Email** — mirrors `handleCreateAccount`: creates the local identity (if this profile has none) and redeems the code with the address the user wants to register.
   - **Atmosphere account** — mirrors `handleCreateAccountWithOAuth` plus the browser redirect finalizer, keeping the OAuth-first ordering: authenticate with the provider, *then* create the local identity, complete OAuth recovery registration against Edge, store the credential, and redeem the code with the provider-verified email.
3. Either way `ClientOperation.CreateAgent` provisions the EDGE agent, as the gate does once the Account exists.

The method input (email address / Atmosphere handle) is an optional second positional arg, prompted when omitted — same shape as `login`.

## Naming

The atproto OAuth path is now called **atmosphere** everywhere the user sees it, matching Composer's `login-atmosphere.label` ("Atmosphere account") and plugin-connector's `atmosphere` connector id. This applies to `login` as well as `signup`, so the two commands agree. `--method atproto` is still accepted as an alias (it resolves to `atmosphere`; it's listed last in help rather than presented as a separate method). Low-level values — `OAuthProvider.ATPROTO`, the requested scopes, the wire protocol — are untouched.

## `Connection`

The gate's `CompleteOAuthRegistration` stores the `AccessToken` **and** wraps it in a `Connection`; the CLI now does the same, so a CLI sign-up leaves a first-class connected-account object rather than an orphaned credential. That required `Connection` to be reachable from plugin-client, which plugin-connector depends on — so `Connection` moved to `@dxos/link`, next to `AccessToken` and `Cursor` (its only dependencies are `@dxos/echo` and `@dxos/link`, and its own doc comment already reasons about `Cursor` living there). The Atmosphere constants moved with it as `Atmosphere.PROVIDER_ID` / `Atmosphere.SOURCE`, which also deletes the two hand-copied `ATMOSPHERE_SOURCE` duplicates (plugin-registry's and the one this PR originally introduced).

That move is a separate commit (`link: move Connection and the Atmosphere constants out of plugin-connector`) so it can be reviewed or dropped on its own. It is mechanical — import-site rewrites across ~40 files in 14 packages — but it does change `@dxos/plugin-connector`'s public API: `Connection`, `ATMOSPHERE_PROVIDER_ID`, and `ATMOSPHERE_SOURCE` are no longer exported from it. No compatibility re-exports were left behind, per the repo convention.

## Changes

- `packages/plugins/plugin-client/src/commands/account/signup/signup.ts` — the new command.
- `packages/plugins/plugin-client/src/commands/account/util.ts` — shared hub-service client (now used by `login` too), the access-code format/normalization helpers, and the `--method` option builder with its alias table. Unit-tested.
- `packages/devtools/cli-util/src/oauth/recovery.ts` — new `performRegisterOAuthFlow` (`purpose: 'register'`, `registerRecovery: true`, returns the opaque `registrationToken`). The initiate + local-callback-server round-trip is now shared with `performRecoveryOAuthFlow` rather than duplicated.
- `packages/core/compute/link/src/{Connection,Atmosphere}.ts` — the moved type and constants.
- `docs/.../publishing-plugins.md` — documents `--method atmosphere` and adds `dx account signup` to the command reference.

## Verification

- Full-repo `moon exec :build` sweep is clean (two sweeps: the first surfaced the import forms my rewrite missed — `import type { … }`, the `/types` subpath, and specifier lists carrying comments — all fixed).
- Lint clean on link, plugin-connector, plugin-client, plugin-onboarding, plugin-registry, plugin-blogger, plugin-bluesky, plugin-github, plugin-inbox; `pnpm format` clean.
- Tests pass: link (23), plugin-connector (23), plugin-inbox (225), plugin-client, plugin-onboarding, plugin-registry. Two plugin activation tests time out when four packages run concurrently in this container and pass individually.
- `dx account signup --help` / `login --help` show the new method lists; `--method atproto` → `atmosphere` and an unknown value is rejected, verified by driving `Options.processCommandLine` under bun.
- Not exercised end-to-end: redeeming a real access code and the atproto OAuth browser round-trip need a live hub and a valid unredeemed code.

## Summary by CodeRabbit

* **New Features**
  * Added `dx account signup <CODE>` for creating accounts with access codes.
  * Added email and Atmosphere-based signup options with interactive and JSON output.
  * Added access-code validation and clearer signup error messages.
* **Changes**
  * Removed invitation-code support from `dx account login`; use signup instead.
  * Renamed the primary authentication method to “Atmosphere,” while retaining `atproto` as an alias.
  * Improved account recovery and OAuth registration flows.
* **Documentation**
  * Updated authentication, signup, and invitation-code guidance.


wittjosiah:
Ejected from the merge queue by a single unrelated test, not by this PR:

`echo-client-e2e` → `src/query.test.ts > Query > aggregate > reactivity > deleting the last object in a group removes the group` ([queue run](https://github.com/dxos/dxos/actions/runs/31496400143/job/93795101301); the queue's `check`, `workerd`, and this branch's own `Check` were all green).

Why it isn't from here: this diff touches no echo/query/GC files at all, while that test lives in the package #12535 (echo GC) just changed on main. On the merged state locally it passes — 162/162 in the full file, and 3/3 when run in isolation — so it reads as a flake or as timing-sensitivity under CI's concurrency + coverage rather than a real regression.

Pushed two commits in the meantime:
- merge of `main` (`cafa2404`), resolving the `pnpm-lock.yaml` conflict by regenerating the lockfile — `.gitattributes` marks it `merge=binary`, so it conflicts on every concurrent lockfile change and cannot be textually merged;
- changeset now names one package per fixed group (`@dxos/app-toolkit` + `@dxos/plugin-client`) instead of enumerating Group A, per `agents/instructions/changesets.md`.

Auto-merge is still armed, so it should re-queue once Check passes on the new head.

---

PR #12382, opened 2026-07-28
connector: declare a connector's sync surface, and sync new bindings automatically

wittjosiah:
## Summary

A connector&#39;s sync behavior is now declared as data on one nested field, and both &#34;how does this sync run?&#34; and &#34;should a new binding sync itself?&#34; follow from it.

```ts
sync?: {
  operation: Operation.Definition;  // reconcile one binding
  getTargets?: …;          // multi-target discovery
  materializeTarget?: …;   // create the local root
  optionsSchema?: …;       // per-binding options
  auto?: boolean;          // sync a binding as soon as it&#39;s created (default false)
  trigger?: Trigger.Spec;  // schedule to keep bindings on
}
```

**`sync.trigger` decides how a sync runs.** Declaring a spec means each binding gets a Routine wrapping a trigger built from it, and every sync of that binding — scheduled or on demand — force-runs that trigger, creating the Routine first if the binding has none. Both paths then go through the dispatcher and get its durable execution (`Operation.runAgain()` continuation, so a capped run finishes its remaining batches). No spec means the sync operation is invoked directly, which is the only path that skips the dispatcher. Previously `SyncConnection` always invoked directly, bypassing the Routine that Gmail&#39;s `onCursorCreated` had set up.

**`sync.auto` decides whether a new binding syncs itself.** A freshly created connection used to sit empty until the user found **Sync now**. Both setup paths now sync when the connector opts in: single-target connectors after `createSingleCursor` binds the target (ordered after navigation, so the user lands on the target while it fills in), multi-target connectors after the sync-targets dialog creates the first cursors. A later *change* of targets still syncs on demand — `reconcileCursors` reports how many cursors existed before the submission, so only the first binding auto-syncs. The sync is forked, or the dialog would sit on &#34;Saving…&#34; for the whole first sync.

Only the two email connectors (Gmail, JMAP Mail) declare `trigger` and `auto`. **Google Calendar declares neither, so it loses the recurring routine it used to get from `onCursorCreated`** — worth a second look if that wasn&#39;t intended. Every other connector was already on-demand-only and is unchanged.

A connection&#39;s bindings sync at concurrency 2 rather than unbounded, so one manual sync doesn&#39;t burst every binding&#39;s requests at the provider.

## Landing state

`MAIL_AUTO_SYNC` in `plugin-inbox/src/capabilities/connector.ts` is **`false`** — this lands with auto-sync off for Gmail and JMAP Mail. It is the single switch for that behavior; flipping it to `true` turns it on for both.

`AUTO_SYNC` in `composer-app/src/playwright/inbox.spec.ts` mirrors it, and selects which of two mutually-exclusive e2e tests runs: with auto-sync on, the mailbox must populate with no sync press; with it off, the mailbox must be empty until the press. Both configurations were run against the app before landing (see Testing).

## What this removes

- **The hard-coded mail cron.** `createSyncRoutine` took a mail-specific `*/10 * * * *`; the schedule now comes from the connector, and the cron literal lives in `plugin-inbox`.
- **The `onCursorCreated` hook.** Its only use was creating the sync routine, which the trigger spec now expresses declaratively.
- **The global `AUTO_SYNC_ON_CONNECTION_SETUP` constant**, replaced by per-connector `sync.auto`.
- **The duplicate trigger lookup.** `syncTarget` and `SyncConnection` share one per-binding dispatch (`syncBinding`), and the target-keyed lookup is gone.
- **The space copy of the sync operation.** `createSyncRoutine` used to serialize the connector&#39;s sync into the space and point the routine&#39;s runnable at that copy. Sync operations are statically defined and already in the hypergraph registry, so the routine now refers to one by key — `Ref.fromURI(operation.meta.key)`, matching the brain and magazine templates.

Placement: `syncTarget`, `createSyncRoutine` and `findBindingForTarget` moved from `plugin-inbox` into `plugin-connector` (which gains a `@dxos/plugin-routine` dep — acyclic, `plugin-routine` doesn&#39;t depend on `plugin-connector`), alongside the new trigger and binding helpers. No compatibility shims left behind.

## Known gaps

- The dispatcher reports failures through the run&#39;s own process, so a 401 on the trigger path surfaces as a generic sync failure rather than the &#34;Connection expired / Go to connection&#34; toast the direct path produces. `fireSyncTrigger` returns when the run is *dispatched*, not when it completes, so there is nothing for `SyncConnection` to classify; the fix is to attach the notify override where the 401 is raised (inside each connector&#39;s sync operation), which would also give scheduled background syncs the reauth prompt they never had. Marked `TODO` in `sync-connection.ts` rather than silently accepted.
- The guard against concurrent Routine creation for one binding is process-local: two peers racing still need the one-routine-per-binding invariant enforced at the data layer.
- A scheduled connector has no direct-invocation fallback: where no trigger monitor exists (CLI, workerd) its sync fails rather than running outside the dispatcher and losing durable execution.
- Tests assert the runnable ref&#39;s *shape*, not that the dispatcher resolves it end to end. That resolution is the path the brain and magazine templates already rely on in production.

## Testing

- `plugin-connector:test` → 36 passed. `operations/sync-connection.test.ts` covers the dispatch matrix: direct invoke without a spec (even when a trigger exists in the space), force-run with a spec, routine created when a scheduled binding has none yet, failure rather than a direct sync when the space has no monitor, no-op when no connector is registered, and both sides of the `sync.auto` opt-in. `sync-routine.test.ts` asserts the created trigger is reachable from its cursor (the reverse-ref a manual sync depends on), that the runnable is the operation&#39;s key, that nothing is persisted into the space, and that two calls racing on one binding create a single routine. `connector-auth.test.ts` asserts reuse sets up the routine for a scheduled connector.
- `plugin-inbox:test` → 222 passed, 14 skipped. Also green: trello, slack, bluesky, linear, github, discord.
- **Inbox e2e (`composer-app`, JMAP against the fixture-backed HTTP mock), both configurations:** with `AUTO_SYNC`/`MAIL_AUTO_SYNC` on → 3 passed, 1 skipped, mailbox populated with nothing pressing sync; with both off (the landing state) → 3 passed, 1 skipped, mailbox empty until the press and populated after. Re-run green after merging `main`. These runs are the evidence for this behavior — the `e2e` job is skipped on this PR.
- Full-workspace `moon exec :build` clean; lint on all touched packages and `pnpm format` clean.

Unrelated flake seen once: `InboxPlugin.test.ts &gt; modules activate on the expected events` timed out at 15s under a parallel run, and passes in 9–13s on its own — it sits close to its limit.

## Summary by CodeRabbit

* **New Features**
  * Connectors can auto-start synchronization on connection creation (via `sync.auto`).
  * Sync can be trigger-driven, including scheduled/force-run behavior.
  * Sync capabilities are standardized under a single nested `sync` block (operation, target discovery, materialization, options).
* **Bug Fixes**
  * Multi-target reconciliation now reports already-bound items and syncs newly added targets immediately.
  * Reduced sync bursts with bounded concurrency.
  * Background sync is automatically ensured for newly created/reused cursor bindings (replacing prior cursor-created hooks).
* **Chores**
  * Improved user-facing sync error reporting and updated sync-related test coverage/e2e behavior.


wittjosiah:
Worked through the review — pushed `cdbe866c`.

**Fixed**

- **Generic `Error` channels.** `syncBinding` and `syncTarget` now fail with a tagged `ConnectionSyncError` (added to `errors.ts` alongside its neighbours), covering the two ways a sync never starts: no handler registered for the connector's sync operation, or no trigger monitor in the space to force-run its schedule. A sync that ran and then failed is still reported by the run's own process.
- **Test helper ordering** in `sync-connection.test.ts` and `sync-target.test.ts`, per `code-style` §Testing ("suite config and main definitions at the top; helpers at the bottom"). Suite state, the stand-in operation, the recording monitor and `makeConnector` stay up top as definitions; the function helpers moved below the cases.
- **A stale comment** the review indirectly surfaced: `ensureSyncTrigger` still claimed callers "fall back to a direct sync" for a broken target ref. That fallback was deliberately removed earlier in this PR, so the comment was wrong.

**Not fixing, with reasons**

- **Reachable fallback for broken target refs** (`sync-routine.ts`) — the suggested change would reintroduce direct-sync fallback for a connector that declares a schedule. That fallback was removed on purpose: the trigger *is* the sync path for such a connector, and running outside the dispatcher silently drops the durable execution (`Operation.runAgain()` continuation) this PR exists to get. `Effect.orDie` on an unresolvable target ref is intended — a binding whose target no longer resolves is broken beyond what routine setup can repair. The comment that implied otherwise is what was wrong, and it's fixed.
- **Reauthentication on the trigger path** (`sync-connection.ts`) — real, and already marked `TODO(wittjosiah)` in the file plus listed under Known gaps in the description. A 401 raised inside a dispatcher-run routine surfaces as a generic sync failure rather than the "Connection expired" prompt. This already matches how the scheduled 10-minute sync behaves, so the PR doesn't regress anything; plumbing 401 classification out of the process boundary is its own change.
- **Concurrent routine creation** (`sync-routine.ts`) — the query-then-add is genuinely non-atomic. It isn't reachable today: the only concurrency is `SyncConnection`'s fan-out at 2, which runs over *distinct* cursors, and `ensureSyncTrigger` dedupes per cursor before creating. The one case where several cursors share a target is a targetless connector (Google Contacts), which declares no trigger spec and so never creates a routine. A cursor-level guard is worth having if that changes; adding locking for an unreachable path now would be speculative.

`plugin-connector` is at 33 passing; full-workspace build, lint and format clean.


---

PR #12385, opened 2026-07-29
ui-editor: route widget updates by id so rehydrated tool state survives a remount

richburdon:
Tool call rows in a chat rendered blank after navigating away and back — the text survived, the rows came back empty. This predates the branch it was found on and affects any chat with tool calls.

## Root cause

Tool rows render from out-of-band widget state, not from the document: `blockToMarkdown` emits a `<toolCall id/>` placeholder and the blocks go out via `updateWidget`, so an empty state renders an empty row. On remount `MessageSyncer.reset` replaces the document — clearing that state via `xmlTagResetEffect` — and rehydrates it. The rehydration never reached the portals.

The update plugin resolved each `xmlTagUpdateEffect` by scanning the decoration set for a `StubWidget` with a non-null `root`. The reset rebuilds decorations into **fresh** widget instances, while `StubWidget.eq` (id equality) makes CodeMirror keep the previously rendered DOM. So the instance holding the mounted root and the instance in the decoration set are different objects, and every candidate the scan found had `root === null`. Instrumenting the story showed exactly that — `1234:false, 4567:false` on each rehydration effect, with the widgets mounted and rendering `blocks=0`.

Updates are now routed through the notifier's id-keyed map of mounted widgets, which is authoritative regardless of which instance the decoration set holds. It also drops an O(decorations) scan per effect.

## Two further defects, fixed along the way

- `widgetStateMapStateField` returned on the first matching effect, so a transaction carrying several `xmlTagUpdateEffect`s kept only the first.
- A widget's props are baked when its decoration is built, so state arriving *before* a widget mounts (a widget outside CodeMirror's viewport) was never picked up. The accumulated state is now merged at mount time.

## Ruled out

Two plausible theories were tested and disproven rather than assumed:

- **Dispatches dropped because the view did not exist yet.** Instrumentation showed `view=true` at every single `updateWidget` call. A buffer-and-replay fix was implemented, observed never to fire, and reverted.
- **The state field losing updates across the reset.** It does not; the test `widget state survives the reset that replaces the document` passes without any fix and pins that down.

## Tests

Four editor-level regression tests in `xml-tags.test.ts` (no browser needed), plus the end-to-end `ChatThread` **Remount** story and the syncer's `sync.test.ts` rehydration case.

Verified non-vacuous: reverting only the routing fix fails exactly `state reaches a mounted widget after a rebuild replaced its decoration instance`, and the other two failed before their respective fixes.

```
ui-editor           346 passed | 1 expected fail
plugin-assistant    41/41 storybook
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed blank XML tag widgets after documents are replaced or chat threads are remounted.
  - Preserved tool-call and tool-result widget state across navigation, streaming updates, and document rebuilds.
  - Ensured widget updates are applied whether they arrive before or after a widget mounts.
- **Tests**
  - Added coverage for widget state restoration, multiple updates, remounting, and rebuilt editor decorations.

### DIFF
--- a/.changeset/xml-widget-state-remount.md
+++ b/.changeset/xml-widget-state-remount.md
@@ -1,0 +1,5 @@
+---
+'@dxos/ui-editor': patch
+---
+
+Fix XML tag widgets rendering blank after the document is replaced — widget state applied around the reset (for example tool call rows when returning to a chat) now reaches the mounted widget.
--- a/packages/plugins/plugin-assistant/src/components/ChatThread/ChatThread.stories.tsx
+++ b/packages/plugins/plugin-assistant/src/components/ChatThread/ChatThread.stories.tsx
@@ -7,6 +7,7 @@ import * as Effect from 'effect/Effect';
 import * as Fiber from 'effect/Fiber';
 import * as Layer from 'effect/Layer';
 import React, { useEffect, useMemo, useState } from 'react';
+import { expect, userEvent, waitFor, within } from 'storybook/test';
 
 import { withPluginManager } from '@dxos/app-framework/testing';
 import { Database, Feed, Filter, Query } from '@dxos/echo';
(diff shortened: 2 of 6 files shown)


richburdon:
Added `.changeset/xml-widget-state-remount.md` (`@dxos/ui-editor`: patch) in a4793562ae — the fix changes observable behavior in a published package, so it warrants a changelog entry.

PR #12397, opened 2026-07-29
plugins: introduce beta/alpha/labs quality tiers

wittjosiah:
Every non-system plugin now carries exactly one quality tier — `beta`, `alpha`, or `labs` — so the registry expresses expected quality directly instead of inferring it from the absence of `labs`.

## Tiers

| Tier | Count | Plugins |
|---|---|---|
| `system` | 21 | unchanged (20 via `dx.config.ts`, plus `onboarding` via `src/meta.ts`) |
| `beta` | 4 | `kanban`, `markdown`, `review`, `table` |
| `alpha` | 18 | `assistant`, `chess`, `crm`, `explorer`, `file`, `game`, `inbox`, `magazine`, `map`, `mermaid`, `outliner`, `presenter`, `sheet`, `sketch`, `stack`, `thread`, `tictactoe`, `transcription` |
| `labs` | 50 | everything else |

The tier leads the `tags` array; secondary tags follow — e.g. `['alpha', 'connector']` for `inbox`, `['labs', 'game', 'connector']` for `chess-com`.

## Registry behaviour

The `recommended` category changes from "everything not tagged `labs`" to an allowlist of `beta` + `alpha`. This fixes the case where an **untagged** plugin was surfaced as recommended by default — that's how the scaffolding plugins (`sample`, `template`, `testing`) ended up there.

Recommended goes from 23 plugins to 22, with 16 unchanged:

- **Leaving:** `blogger`, `brain`, `chess-com`, `projects`, `sample`, `template`, `testing`
- **Entering:** `crm`, `magazine`, `map`, `outliner`, `tictactoe`, `transcription`

`alpha` is added to `RegistryTagType`. `Config2` already types `tags` as `Schema.Array(Schema.String)`, so no schema change was needed.

## Tag hues

Tier hues ramp **green → blue → purple** (beta → alpha → labs) so the ordering reads without knowing the labels. `beta` shares `green` with `popular`, which no in-repo plugin uses — it only arrives on catalog plugins via the registry manifest.

## `integration` → `connector`

The `integration` secondary tag is renamed to `connector` across the 10 plugins carrying it: `bluesky`, `chess-com`, `discord`, `freeq`, `github`, `ibkr`, `inbox`, `linear`, `slack`, `trello`. Nothing keyed off the old string — it was absent from `RegistryTagType` and from every filter — so this is a rename in the configs alone. The name deliberately parallels `plugin-connector`, the same way `game` parallels `plugin-game`.

## Composer defaults

`search` is dropped from the unconditional `getDefaults()` set. It is tagged `labs`, so it would otherwise ship enabled for every new user while being absent from the registry's Recommended list.

## Also

`packages/plugins/plugin-files` is deleted — a single orphaned `FilesPlugin.tsx` with no `package.json` and no references anywhere in the repo.

## Verification

- `plugin-registry` tests — 5 passed, including 4 new cases in `categories.test.ts` covering tier selection, tier-in-any-position, core/remote exclusion, and the untagged-is-excluded regression.
- `plugin-registry` lint + build — clean.
- `composer-app:build` — exit 0 (compiles every plugin config).
- `oxfmt --check` — clean.
- `pnpm install --frozen-lockfile` — clean after merging main.
- Checked the `dependsOn` graph: no `beta`/`alpha` plugin depends on a `labs` plugin, so enabling a Recommended plugin cannot pull in a labs dependency.
- 93 plugin directories, 92 `dx.config.ts` (only `plugin-onboarding` lacks one, tagged via `meta.ts`) — every plugin has exactly one tier.

## Summary by CodeRabbit

- **New Features**
  - Introduced plugin quality tiers (beta, alpha, labs) and updated plugin listing classifications, including refreshed Composer defaults/selections.
  - Added/updated plugin metadata tags and renamed the integration label to **connector**.
- **Bug Fixes**
  - Updated “recommended” plugin selection to include only **beta** and **alpha**, excluding untagged and labs plugins by default.
- **Documentation**
  - Updated internal plugin scaffolding guidance for quality tiers.
- **Removed**
  - Removed the Files plugin.
- **Style**
  - Refined tier color mapping and adjusted several icon hues for visual consistency.


wittjosiah:
CI is red here for a reason that predates this branch — flagging it once rather than per-job.

All jobs fail in the shared setup action's "Log out debug info" step:

```
v24.11.1
10.28.0
bun: command not found
##[error]Process completed with exit code 127
```

Nothing in this PR has run at that point — no build, no tests — which is why no `test-results` artifact is produced.

The cause is on `main`. `.github/Dockerfile` deliberately does not install bun:

> `bun` is deliberately absent — moon installs it, and a copy here would shadow the pinned one.

But `bun --version` was added to `.github/actions/setup/action.yml` as a bare PATH invocation in b7525814 (#12324). Moon installs bun into its proto-managed toolchain rather than onto `PATH`, so that step only succeeds when the restored toolchain cache happens to expose a shim. In the same run, `moon setup` reports `skip SetupToolchain(bun:1.3.11) … All toolchains are already up to date!` immediately before the failure — moon considers bun present, the shell cannot see it.

That cache dependence explains the intermittency: commit c66de9b5 went fully green (cache key `…d025b0b33fce…`), while c66de9b5→d198934f fail with key `…28680e21113a…`.

Merging `main` (d198934f) did not help, since the defect is on `main` itself.

The fix belongs in the setup action, not here — either drop the bare `bun --version`, or invoke it through moon/proto so it resolves the pinned toolchain. Happy to make that change if it should ride along with this PR, but it is shared CI infrastructure and unrelated to plugin tagging, so my default is to leave it for a dedicated PR.

Everything in this PR passes locally: `plugin-registry` 5/5 tests, lint, build, `composer-app:build` exit 0, and `oxfmt --check` clean.

---

PR #12408, opened 2026-07-30
feat(edge-client): fetch the auth nonce from /auth instead of provoking a 401

mykola-vrmchk:
## Problem

Both edge clients obtained their auth challenge by sending a request they expected to be **rejected**.

`BaseHttpClient` prefetched `/auth` and only acted on a failure:

```ts
if (!this._authHeader && args.auth) {
  const response = await fetch(new URL('/auth', this._baseUrl));
  if (response.status === 401) {                    // <- the success case
    this._authHeader = await this._handleUnauthorized(response);
  }
}
```

`EdgeClient._createAuthHeader` was worse — it never touched `/auth` at all. It fired a GET at the `/ws/:identityDid/:peerKey` upgrade path itself, purely to harvest the challenge off that path's 401.

Both produced a red `Failed to load resource: 401` in the browser console on every client boot, and a routine `auth.failure` in edge's audit trail, for an operation that succeeded.

## Changes

Both paths now go through a shared `authenticateViaChallengeEndpoint`, which GETs `/auth` and reads the challenge from **either** a 200 body or a `WWW-Authenticate` header. Edge answers 200 for anonymous callers as of dxos/edge#775; the header path keeps this working against servers that predate it. Both clients keep their lazy 401-and-retry path for stale credentials.

Three real bugs fell out of the shared parser:

1. **Challenge lists were unparseable.** `WWW-Authenticate` carries a comma-separated *list* (RFC 9110 §11.6.1), but the old code asserted the header *started with* `VerifiablePresentation challenge=`. So `Bearer realm="dxos", VerifiablePresentation …` — exactly what edge emits whenever admin-key auth is also allowed — failed the invariant outright. `payments-service` carries a separate VP-only middleware purely to dodge this.

2. **Quotes were never stripped.** `slice()` kept the surrounding quotes and relied on `Buffer.from` silently discarding them. Works today; `atob` would throw.

3. **Dead protocol assignment.** `httpUrl.protocol = getEdgeUrlWithProtocol(...)` assigned a whole URL string to `URL.protocol`, which is an invalid protocol value and silently ignored. `_baseHttpUrl` was already correct, and is now used directly.

The prefetch is also skipped when no identity is set — fetching a challenge you cannot sign is a wasted round trip.

## Compatibility

Works against edge on either side of dxos/edge#775:

| Server | Path taken |
|---|---|
| New (`200` + body) | Challenge read from body, no console error |
| Old (`401` + header) | Challenge read from header, unchanged behaviour |
| Challenge endpoint down | Falls back to lazy 401-and-retry |

`handleAuthChallenge` keeps its signature; the 401 assertion is relaxed to accept either shape, so `plugin-payments` needs no change.

## Test plan

- `edge-client` — 51 passing (10 files). New `edge-identity.test.ts` covers header parsing (quoted, legacy unquoted, VP not first in the list, quote stripping), both response shapes, header-vs-body precedence, that the body is left unconsumed for the caller, network-failure tolerance, and that the signed bytes are the decoded challenge.
- `putBlob sends a raw POST body and pre-fetches /auth` now sets an identity — it previously asserted the wasted no-identity call.
- `moon run edge-client:build`, `edge-client:lint`, `pnpm run format-check` clean.

## Merge order

Land dxos/edge#775 first. It is backwards-compatible on its own, and this PR is compatible with both server versions, so the two can land independently — but the console noise only goes away once both are deployed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication handling for edge HTTP and WebSocket requests.
  * Authentication retries now occur only for valid Verifiable Presentation challenges.
  * Requests without a valid challenge preserve the original authentication error.
  * Added graceful handling for missing, malformed, or unavailable authentication challenges.

* **Tests**
  * Expanded coverage for challenge parsing, authentication responses, retries, and failure scenarios.


mykola-vrmchk:
Addressed in 7d1f43f — thanks.

1. **`edge-client.ts` 401 fallback** — done. The `WWW-Authenticate` guard is now there, matching `BaseHttpClient`. Worth noting this was a live bug, not just a robustness nit: `handleAuthChallenge` invariants on the challenge being present, so an upstream-forwarded 401 would have thrown instead of degrading to the warn-and-return-undefined path.

2. **Test helper placement** — done, moved below the describes per the convention in `edge-ws-muxer.test.ts`.

3. **Untyped fetch mock** — done, and this one was load-bearing. It failed `edge-client:build` with `TS2493`, which cascaded into the `build`, `check`, `test` and `storybook` jobs; all four red checks were this single error.

mykola-vrmchk:
Good catch on the follow-up — fixed in 7e24ab1.

You were right that header presence is the wrong predicate. Both clients now gate on a *parsed* VP challenge, not just a non-null header:

- `EdgeClient._createAuthHeader` reads the challenge once via `readAuthChallenge` and signs with `presentCredentialsForChallenge`, which also removes a redundant second parse of the same header.
- `BaseHttpClient` had the same weakness in both its retry sites, so it gets the same treatment via a shared `hasVpChallenge` helper. Worth flagging since it wasn't in the original comment: this one was pre-existing, and the failure mode is nastier than a throw — the invariant was caught by the surrounding try/catch and rewrapped as a generic `'Error processing request.'`, so the caller never saw the real upstream status.

On the regression test: my first attempt asserted `rejects.toThrow()` and a single resource call, and I checked it against the old guard — it passed either way, because the broken path throws too. It now asserts the surfaced error matches `/HTTP code 401/`, which is the actual behavioural difference. Verified failing against the previous guard with `AssertionError: expected [Function] to throw error matching /HTTP code 401/ but got 'Error processing request.'`

Local: `edge-client` 52 passing, clean build/lint/format-check.

mykola-vrmchk:
Both addressed in 5dcba17.

**Empty VP challenge** — valid, and reachable rather than theoretical, which makes it worth more than the 🟡 it was tagged. `parseChallengeHeader` returned `''` for `challenge=""`, which slipped past `!== undefined` and routed the request into the auth path, failing on the missing challenge and rewrapping as `'Error processing request.'` — the same masking bug as the non-VP case, one value short of it. The trigger is in the paired PR: edge emits exactly `challenge=""` when `DX_HUB_SERVICE_KEYPAIR` is unset (dxos/edge#775), so a misconfigured deployment would hide every 401 behind a generic processing error.

I normalised at the parser (`(match[1] ?? match[2]) || undefined`) rather than at the call site as suggested. Same result for `hasVpChallenge`, but `readAuthChallenge` and any future caller inherit it instead of each needing its own truthiness check — an empty challenge isn't a challenge, so the parser is the right place to say so.

The two non-signable header cases are now a table-driven pair, and I verified both new tests fail against the previous behaviour (`expected '' to be undefined`, and `expected [Function] to throw error matching /HTTP code 401/ but got 'Error processing request.'`).

**Fetch mock typing** — done, via a `requestUrl` helper that narrows all three `RequestInfo | URL` shapes, replacing the `input.url ?? input` dance. Scoped to the test I added; the pre-existing mocks in the file still use `input: any` and I left them alone rather than widening this diff.

Local: 54 passing, clean build/lint/format-check.

PR #12412, opened 2026-07-30
echo: convergence-key object merging

wittjosiah:
Follow-up to #12410, which landed the object-merging feasibility research. This settles the design review and implements convergence-key merging end to end, with the merge — and the reference rewriting — running **in the worker**.

## Design decisions

All §7 questions resolved, with reasoning in `DESIGN.md`. Identity lives on a **new dedicated `EntityMeta` field** rather than `meta.key`/`meta.version` (provenance, not identity — overloading them would make every registry-stamped object a merge candidate) or a `ForeignKey` (no version, and alias sets can *become* overlapping later). Version matching is **exact**. Scope is **every storable entity kind**, phased; objects only for now, enforced.

The field is **`meta.convergenceKey`**, a single opaque string. Callers encode generations in the string (`com.example.seed@2`) and get distinct-entity semantics for free. The name states the guarantee the engine makes — entities sharing the key converge to one, eventually — and was renamed from `naturalKey` during review (`DESIGN.md` §4.4 records the candidates and rationale; `singletonKey` was declined as implying write-time enforcement).

**Where the merge runs went through three homes, deliberately.** On-space-open was implemented and reverted (detection scanned `Filter.everything()` and hydrated the whole space, breaking lazy loading). Inside query evaluation was implemented and superseded after a day (writes during query evaluation, per-query-instance work, blind to id/ref loads). The final home is the worker's indexing stream.

## What ships

**Public API** — none, deliberately: the meta field itself is the API. Identity is declared by assigning it inside the caller's update callback — `Entity.update(x, (x) => { Entity.getMeta(x).convergenceKey = '…' })` — so the write batches with the caller's other changes (dedicated accessors were dropped during review: a setter wrapping `Entity.update` defeats batching, and its validation was advisory — the engine layers each re-verify because the field replicates from untrusted peers). The deterministic core lives in the worker (`echo-host/src/db-host/merge-core.ts`): set-wise `mergeCandidates` (winner = minimum id — a superset never picks a larger winner than a subset, which is why redirect chains terminate; deliberately not a pairwise fold, which isn't associative; deduplicates by id because query results aren't unique). The one piece the client keeps, in `@dxos/echo/internal`, is `resolveMergeRedirect` (transitive; an edge that fails to decrease the id ends the chain, so cycles and forward references terminate).

**`echo-host`** — `ConvergenceKeyMerger` (`db-host/convergence-key-merge.ts`), the worker trigger, with `loadDoc`/`flushDoc`/`queryByConvergenceKeys`/`queryReferrers` injected via the constructor (documents are held as disposable leases and returned per group). `EchoHost._runUpdateIndexes` processes every document change, including replication arrivals — the moment duplicates come into existence on a device. Detection is crash-proof: keys seen while indexing are recorded as durable intents in the same transaction as the index cursors and cleared only once serviced, honoring a durability dual — the winner's folded data is flushed before any tombstone, and every loser document before the intent that would retry it is cleared. Straggler folds diff from the watermark read as the **union of the stored `mergedAtHeads` register and its automerge conflicts**, so concurrent merges never re-fold already-folded edits; the fold advances the watermark only when its write applied. Deletion mid-flight is respected, never merged.

**Reference rewriting is index-driven (DESIGN §4.11)** — `IndexEngine.queryReferrers` joins the reverse-reference index to the object metadata, naming each referrer of a merged-away loser and the exact property holding the ref, so repointing is a point load per referrer: no client, no scan, no `Filter.everything()` anywhere. It runs right after a merge's tombstones — only for losers whose tombstone actually applied, so a loser re-keyed or deleted mid-flush keeps its referrers — and again on every service pass over a re-indexed loser, which is what re-covers referrers written concurrently with the merge. Deliberately best-effort and outside the durability guarantee: every failure surface (the write included) logs and skips, referrer loads never wait on the network, and a ref left behind still resolves through the redirect — so rewriting is an optimization plus the compatibility path for clients too old to follow `mergedInto`.

**`index-core`** — `objectMeta.convergenceKey` (entity-meta migration `0007`) with a `(spaceId, convergenceKey)` point-lookup index and `queryByConvergenceKeys`; `ReverseRefIndex.queryReferrers` (the `reverseRef ⋈ objectMeta` join, discharging that module's standing TODO); chunked `lookupByRecordIds`; the durable `ConvergenceKeyIntentStore` with its own migration history, joining the indexing transaction (an intent means "check this key's group", not "duplicates exist"); and cursor-name versioning (`fts5`→`fts6`, `reverseRef`→`reverseRef2`, retired names purged by tracker migration `0003`) so databases indexed before the column existed re-index once and become detectable.

**`echo-client`** — reference resolution follows `system.mergedInto`, sync and async, so an un-rewritten reference reaches the survivor instead of dangling; transitive deletion follows redirects too, so relations and children anchored at a merge loser stay visible, judged at the survivor. The query path keeps a **read-only** filter dropping tombstoned losers. There is no client merge surface: `db.mergeDuplicates()` and the client executor were removed in favor of the worker's rewriting. The winner records `system.mergedFrom` — the transitively-closed reverse edge of `mergedInto`.

**Semantics**: per-field, smallest-id-defining-candidate wins; per-field writes so a concurrent edit to an untouched field keeps its LWW outcome; losers are tombstoned, never erased. The never-see-two guarantee is **eventual** (~one indexing cycle), and convergence is by *agreement* — every peer independently computes the same winner.

## Review history

Hardened through three adversarial review rounds (`DESIGN.md` decision log, 2026-08-02 / 2026-08-03 ×2: synchronous read-compute-write, applied-gated watermarks, durable intent log, cross-document flush ordering, conflict-union watermark, prototype-safe accumulation, redirect-following deletion) plus two structural rounds from @dmaretskyi: 2026-08-13 (machinery internal, migration-based cursor invalidation, intent store split out of the indexer, merger class, the `convergenceKey` rename) and 2026-08-31 (accessors dropped — the meta field is the API; merge/convergence suites to `echo-client-e2e`; worker success-path logging; typed `QueryAST` walk; and §4.11 — the `Filter.everything()` pass replaced by worker-side index-driven rewriting, the client executor deleted, the merge core moved host-side). A further three-model adversarial review of §4.11 (2026-08-31) produced the declined-tombstone guard, full failure containment on the rewrite path, the index-core join, and the chunked lookup.

## Tests

All green: `echo` 603, `echo-host` 359 (merge-core unit suite + 19 merger tests incl. 9 on reference rewriting), `echo-client` 556, `echo-client-e2e` 355 (24 merge/convergence), `index-core` 57.

- Unit tests on the pure core, including the non-associative case that motivates set-wise merging, prototype-member field names, and `__proto__` as data; `resolveMergeRedirect`'s own suite beside it in `@dxos/echo`.
- Worker tests driving `ConvergenceKeyMerger` against real repo handles with injected mid-load/mid-flush mutations, flush-order recording, forked-and-merged concurrent-watermark states, and rewrite coverage (nested/array paths, qualified-ref spelling, stale index rows, tombstoned and merged-away referrers, lookup faults, late-referrer re-cover, declined-tombstone losers, shared group leases).
- Index tests for the `queryReferrers` join (path grouping and escape round-trip; same-space and document-bearing filters).
- Multi-peer convergence tests over `TestReplicationNetwork`: independent workers agree on the winner; a partial-view merge (staged through cores, as a peer's worker would write it) builds a redirect chain that still resolves to the global minimum; a straggler's late edit is folded. The §4.11 freshness residual is pinned: a referrer added after convergence resolves through the redirect.

## Notes for reviewers

- **`objectStructureToJson` omitted `@meta` entirely** — the indexer could never see a convergence key, making the trigger silently inert. It now emits the meta section, matching the feed path.
- Adding a field to `EntityMeta` used to need three hand-maintained lists; `getSnapshot` and `metaNotEmpty` now enumerate `SCALAR_META_FIELDS`, derived from the schema.
- `waitUntilHeadsReplicated` is not a query barrier — queries read through the index, which settles a tick later. Multi-peer tests wait on the observable result.

## Remaining (tracked in TASKS.md)

`db.ensure`-style creation API (@dmaretskyi's `querySingleton`/`ensureSingleton` sketch is recorded against it), doctor duplicates diagnostic, property-based determinism over randomized schedules, relation endpoint rewriting (`objectMeta.source/target/parent` are indexed columns, so it is the same point-lookup shape as §4.11), the collaborative-text merge policy (min-id-wins discards a loser's entire text — must be decided before documents are in scope), relations/types as merge subjects.

https://claude.ai/code/session_01V6KKq2x1qcgB5akNCAETQs

## Summary by CodeRabbit

* **New Features**
  * Added automatic merging for duplicate objects sharing a `convergenceKey`.
  * Preserved a single survivor, combined fields, and retained merge history.
  * Redirected lookups and references from merged objects to the survivor.
  * Made merges durable across restarts and later edits.
  * Excluded relations and explicitly deleted objects from automatic merging.
  * Preserved convergence metadata through snapshots and serialization.

* **Bug Fixes**
  * Corrected grouped counts after duplicate tombstones are collapsed.

* **Documentation**
  * Updated object-merging design notes and task tracking.

* **Tests**
  * Added comprehensive coverage for convergence, redirects, references, persistence, and edge cases.

---
🚀 Composer preview: https://pr-12412-composer-dev.dxos.workers.dev


packages/core/echo/echo/src/internal/Merge/merge.ts
```diff
@@ -0,0 +1,267 @@
...
+//
+// Candidates
+//
+
+/**
+ * Storage-independent view of one entity participating in a merge.
+ *
+ * The merge core is defined over these rather than over live objects so that winner selection
+ * and the merge function stay pure and testable without a database.
+ */
```
dmaretskyi, line 17: do we really need this api?

packages/core/echo/echo/src/Merge.ts
```diff
@@ -0,0 +1,267 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+// @import-as-namespace
+
```
dmaretskyi: a lot of the apis here seem internal to echo - not something users would use

packages/core/echo/index-core/src/indexes/entity-meta-index.ts
```diff
@@ -112,10 +121,30 @@ const buildSourceCondition = (
...
+    const priorColumns = yield* sql<{ name: string }>`SELECT name FROM pragma_table_info('objectMeta')`;
+    const needsNaturalKeyBackfill = priorColumns.length > 0 && !priorColumns.some(({ name }) => name === 'naturalKey');
+    if (needsNaturalKeyBackfill) {
+      // The tracker migrates before the indexes, so the cursor table exists here. Dropping every
+      // cursor re-presents all documents to the indexing loop, whose per-index upserts are
+      // idempotent — the one-time cost of a full re-index buys correct duplicate detection over
+      // data indexed before the column existed. The wipe runs BEFORE the ALTER below: migration
+      // statements auto-commit individually, and a crash between them must re-run the wipe on
+      // the next startup, not see the column and skip it.
+      yield* sql`DELETE FROM indexCursor`;
```
dmaretskyi: a proper way to do that would be to change the index name, so cursors become invalidated automatically. do index names already have versions appended?

packages/core/echo/index-core/src/indexes/entity-meta-index.ts
```diff
@@ -149,10 +181,111 @@ export class EntityMetaIndex implements Index {
...
+      }),
+  );
+
+  /**
+   * Durably queue natural keys for duplicate detection. Runs inside the same transaction that
+   * commits the index rows and cursors (see `IndexEngine.#update`), so a keyed write can never
+   * be indexed-but-forgotten: until the merge pass services the key and clears the intent, every
+   * later pass re-presents it.
+   */
+  recordNaturalKeyIntents = Effect.fn('EntityMetaIndex.recordNaturalKeyIntents')(
```
dmaretskyi: this intents table seems really out of scope for indexer? can it be safely decoupled into something separate?

packages/core/echo/echo-host/src/db-host/natural-key-merge.ts
```diff
@@ -0,0 +1,537 @@
...
+  }
+
+  // The change callback runs in the same tick as the reads the merge was computed from, so the
+  // re-checks are a backstop against out-of-band mutation, not the concurrency mechanism.
+  let applied = false;
+  winner.handle.change((doc: DatabaseDirectory) => {
+    const entity = doc.objects?.[winner.objectId];
+    if (
+      !entity ||
+      entity.system?.mergedInto !== undefined ||
```
dmaretskyi: the approach with `mergedInto` seems right.

packages/core/echo/echo-host/src/db-host/natural-key-merge.ts
```diff
@@ -0,0 +1,537 @@
...
+import { type DocHandle, type DocumentId } from '@automerge/automerge-repo';
+
+import { type Context } from '@dxos/context';
+import { Merge } from '@dxos/echo';
+import { type DatabaseDirectory, type EntityStructure, PROPERTY_ID } from '@dxos/echo-protocol';
+import { type EntityMeta } from '@dxos/index-core';
+import { type EntityId, type SpaceId } from '@dxos/keys';
+import { log } from '@dxos/log';
+
+export type NaturalKeyGroupContext = {
```
dmaretskyi: I would make this whole module a class, that takes deps in the contructor (can depend on automerge host & indexer or query engine for example)

packages/core/echo/echo/src/internal/Merge/merge.ts
wittjosiah: `Candidate` is the input to the pure merge core, which both write paths share: echo-host builds candidates from raw automerge entities (`natural-key-merge.ts`) and echo-client builds them from live objects via `candidateOf` (`merge-executor.ts`), so winner selection and the field-wise merge exist once and both sides provably agree. So the type is needed — but you're right that it has no user-facing consumers. Per the thread above, the plan would be to move it (with `merge`, `candidateOf`, `findDuplicates`, `resolveRedirect`) behind `@dxos/echo/internal`; `selectWinner` and `groupByNaturalKey` have no consumers outside this module's own tests and can be dropped outright.

---

packages/core/echo/echo/src/Merge.ts
wittjosiah: Agreed. The consumption map: `setNaturalKey`/`getNaturalKey` are the only user-facing pieces (declaring/reading the domain identity); `merge`, `candidateOf`, `findDuplicates`, and `resolveRedirect` are consumed only by echo-host (`natural-key-merge.ts`) and echo-client (`merge-executor.ts`, `object-core.ts`, `hypergraph.ts`); `selectWinner`/`groupByNaturalKey` only by tests.

Proposal: keep a slim public `Merge` namespace with just the natural-key accessors, move the machinery to `@dxos/echo/internal`, and delete the two test-only helpers. Ready to implement once there's agreement on that shape.

---

packages/core/echo/index-core/src/indexes/entity-meta-index.ts
wittjosiah: There's no systematic version scheme, but the pattern exists ad hoc: the FTS cursor is named `fts5` and the retired `fts` name is cleaned up via `DEPRECATED_INDEX_NAMES` in `index-tracker.ts`.

The wrinkle is that `EntityMetaIndex` has no cursor of its own — `objectMeta` rows are upserted as a side effect of both driving passes (`fts5` and `reverseRef` in `IndexEngine.update`) — so invalidating by name means bumping both driving cursor names (e.g. `fts6`, `reverseRef2`) and adding the old ones to the deprecated list. That would replace the `pragma_table_info` probe and the cursor wipe here (including its wipe-before-ALTER ordering subtlety) with the existing declarative mechanism, at the cost of re-indexing both driven indexes on upgrade — which the wipe does anyway. Happy to make that change.

---

packages/core/echo/index-core/src/indexes/entity-meta-index.ts
wittjosiah: The coupling is transactional rather than conceptual: an intent has to commit atomically with the cursor advance it guards — if the cursor advances and the intent write lives in a different store, a crash between the two loses the detection forever, because that keyed write is never re-presented to the indexing loop.

So it can't move to separate storage, but it can safely move out of `EntityMetaIndex` into its own module (e.g. a `NaturalKeyIntentStore`) that joins the same `SqlTransaction` inside `IndexEngine.#update` — same guarantee, cleaner ownership, and the indexer stops carrying merge-workflow tables. Happy to split it out that way.

---

packages/core/echo/echo-host/src/db-host/natural-key-merge.ts
wittjosiah: Makes sense. The per-call context object was chosen so tests can drive `mergeNaturalKeyGroup` with plain fakes (no automerge host or database), but constructor injection keeps that property — a class taking the automerge host + index engine (or the same narrow `loadDoc`/`flushDoc`/`queryByNaturalKeys` seam) is a mechanical refactor. Will fold it into the restructuring pass along with the API-surface changes discussed on `Merge.ts`.

---

.agents/projects/object-merging/DESIGN.md
```diff
@@ -220,39 +392,132 @@ The key correctness argument, for peers acting on different views:
...
+public key across 17 files, so the collision would be actively misleading.
+
+Accepted cost: the name carries no hint that stamping it opts the entity into
+merging, so that consequence belongs in the field's doc comment.
+
+The candidates weighed, for the record (`naturalKey` is the one chosen):
+
+| Candidate      | Reads as                                                                                                     | Against                                                                                     |
+| -------------- | ------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------- |
+| `naturalKey`   | Standard relational term: the natural key beside the surrogate key (`EntityId`). Precise about what it _is_. | Assumes DB vocabulary; understates the merge consequence.                                   |
```
dmaretskyi: I'd go with singletonKey despite OO baggage. "naturalKey" carries no meaning to me.

Also I think it's worth creating new APIs for this:

Database.querySignleton(KEY,SCHEMA): Effect<Option<T>>
Database.querySignleton(SCHEMA): Effect<Option<T>> // can schema have the key as annotation?
Database.ensureSignleton(SCHEMA): Effect<T>

wittjosiah: The table above records why `singletonKey` lost: the invariant is eventual, not write-time — before the merge runs there genuinely are two live entities carrying the key, and "singleton" reads as an enforced guarantee. But the API sketch changes the calculus: if `ensureSingleton`/`querySingleton` become the primary entry points, the name describes the contract those APIs expose (callers always observe one), and the merge becomes the resolution mechanism underneath — the eventual-ness turns into an implementation detail. The rename itself is mechanical while nothing has shipped (meta field, index column, protocol accessors, docs).

The APIs line up with the roadmap: a creation-side `db.ensure` is already a planned Phase-2 item, and `ensureSingleton(SCHEMA)` is that item with the key carried as a schema annotation. One cardinality note: today's key is per `(space, key)` — instance-level identity (e.g. one entity per external org id) — while `querySingleton(SCHEMA)` implies per-type singletons. Both fit if the annotation supplies the key for type-level singletons and the `(KEY, SCHEMA)` overload covers instance-level ones. Treating the rename + new APIs as a decision to confirm before implementing.

---

wittjosiah:
@dmaretskyi — everything from your review is now implemented on the branch; summary of the outcomes:

1. **Public API slimmed** (`Merge.ts` threads): the public `Merge` namespace is gone. Users declare identity with `Entity.setConvergenceKey` / `Entity.getConvergenceKey` (beside the other meta accessors); the deterministic core moved to `@dxos/echo/internal` (`mergeCandidates`, `toMergeCandidate`, `findMergeDuplicates`, `resolveMergeRedirect`). `selectWinner` was deleted outright and `groupByNaturalKey` became a private helper.
2. **Index-name versioning** replaced the cursor wipe: the driving cursor names bumped (`fts5` → `fts6`, `reverseRef` → `reverseRef2`) with the old names purged via `DEPRECATED_INDEX_NAMES` — after the branch merged main's numbered-migration system, the purge landed as tracker migration `0002` and the new column as entity-meta migration `0004`, so the pragma probe and wipe-before-ALTER ordering are gone entirely.
3. **Intents decoupled from the indexer**: `ConvergenceKeyIntentStore` owns the table with its own migration history, joining the same `SqlTransaction` in `IndexEngine` — the transactional guarantee is unchanged.
4. **Module → class**: `ConvergenceKeyMerger` takes the narrow `loadDoc`/`flushDoc`/`queryByConvergenceKeys` seam as constructor deps; `EchoHost` constructs it once; tests drive it with plain fakes as before.
5. **Naming**: `naturalKey` is renamed **`convergenceKey`**. We didn't take `singletonKey` — it reads as a write-time uniqueness guarantee the engine deliberately doesn't make — but the complaint stood, and `convergenceKey` states the actual guarantee (entities sharing the key converge to one, eventually) in vocabulary this codebase already owns. Your `querySingleton`/`ensureSingleton` sketch is recorded against the planned Phase-2 `db.ensure` API. Rationale in `DESIGN.md` §4.4 + decision log.

All behavior-preserving; suites green (echo 612, echo-host 288, echo-client 552, index-core 49) and CI green on the tip. I've resolved the implemented threads to keep the review view tidy.

---

.agents/projects/object-merging/DESIGN.md
wittjosiah: Resolved with a different name: the field is now **`convergenceKey`** (`Entity.setConvergenceKey`). `singletonKey` still read as write-time uniqueness enforcement, but the underlying complaint was right — "natural key" meant nothing without relational-database background. `convergenceKey` states the actual guarantee (entities sharing the key converge to one, eventually) in local-first vocabulary. Your `querySingleton`/`ensureSingleton` API sketch is recorded against the planned Phase-2 `db.ensure` item — likely as `ensure`/`queryByKey` shapes over the new name. Full rationale in `DESIGN.md` §4.4.

---

packages/core/echo/echo/src/Entity.ts
```diff
@@ -242,6 +242,37 @@ export function getMeta(entity: Unknown | Snapshot | Mutable<Unknown>): internal
...
+ * @throws On a non-object entity, or an empty-string key (which would group unrelated entities).
+ */
+export const setConvergenceKey = (entity: Unknown, convergenceKey: string | undefined): void => {
+  if (internal.getEntityKindBrand(entity) !== internal.EntityKind.Object) {
+    throw new TypeError('Convergence keys are limited to objects; relations and types are not merge subjects.');
+  }
+  if (convergenceKey !== undefined && convergenceKey.length === 0) {
+    throw new TypeError('Convergence key must be a non-empty string.');
+  }
+  update(entity, (entity) => {
```
dmaretskyi: I'm generally against APIs that call `Entity.update` inside of them -- this makes batching multiple updates impossible. We should instead have the consumer call `Entity.update`.

```ts
Entity.update(x, (x) => {
   Entity.setConvergenceKey(x, 'foo')
})
```

...but then do we even need a separate function? why not:

```ts
Entity.update(x, (x) => {
   Entity.getMeta(x).convergenceKey = 'foo'
})
```

packages/core/echo/echo/src/Entity.ts
```diff
@@ -242,6 +242,37 @@ export function getMeta(entity: Unknown | Snapshot | Mutable<Unknown>): internal
...
+ * carrying the same convergence key are the same entity and eventually converge to one. It sits
+ * alongside the entity's id, which is a surrogate — system-minted, random, and meaningless
+ * outside the database.
+ */
+export const getConvergenceKey = (entity: Unknown | Snapshot): string | undefined => getMeta(entity).convergenceKey;
+
+/**
+ * Declare an entity's convergence key, or clear it when passed `undefined`.
+ *
+ * Objects only: relations and types are not merge subjects yet — merging a relation would
```
dmaretskyi: if so, doesn't this whole set of APIs belong in the `Obj` namespace?

packages/core/echo/echo-client/src/merge/merge-executor.ts
```diff
@@ -0,0 +1,317 @@
...
+ * worker off the indexing stream (`echo-host`'s convergence-key merge). Being a pure function of the
+ * candidate set, it is safe to run redundantly — a client pass racing the worker converges, and a
+ * second pass over the same entities is a no-op. Groups of one are skipped, so a space with no
+ * duplicates costs one grouping pass and no writes.
+ *
+ * Writes only the fields whose values differ from the merged result, rather than replacing the
+ * whole object — a whole-object replace would rewrite every field, so a concurrent edit to a
+ * property the merge never touched would lose the last-write-wins race.
+ */
+export const mergeDuplicates = (entities: readonly Obj.Unknown[]): MergePassResult => {
```
dmaretskyi: does this run in the client? shouldn't we do merging in the worker?

packages/core/echo/echo-client/src/proxy-db/database.ts
```diff
@@ -404,6 +405,35 @@ export class DatabaseImpl extends Resource implements EchoDatabase {
...
+   *
+   * Being a pure function of the entities it sees, every peer computes the same result, so a
+   * pass racing the worker converges rather than fights, and a pass with nothing to merge
+   * writes nothing. Detection here scans the whole space (hydrating it) — acceptable for an
+   * explicit call; the worker's index point-lookup is what keeps the automatic path cheap.
+   */
+  async mergeDuplicates(): Promise<MergePassResult> {
+    // Includes tombstones: the worker usually merges before this pass ever sees the duplicates,
+    // so what is left is repointing references at the survivors — which needs the losers.
+    const entities = await this.query(Query.select(Filter.everything()).options({ deleted: 'include' })).run();
```
dmaretskyi: oh no, "query everything" is a huge perf pitfall

packages/core/echo/echo/src/Entity.ts
wittjosiah: The batching point is fair — wrapping `Entity.update` was inherited from the original shape, not a requirement. And your simpler form works today: the field is an ordinary meta property, and none of the setter's validation is load-bearing — detection skips relations and empty keys (`convergenceKeyOf` in the index engine), the worker re-verifies `system.kind` and the key against the document, and the client executor checks `Obj.isObject` — all because the field is raw-meta-writable and replicated data can't be trusted anyway. A keyed relation or empty string set through `getMeta` is inert, not harmful.

So dropping both accessors in favor of `Entity.getMeta(x).convergenceKey = 'foo'` is viable; the only loss is fail-fast (a mistaken key on a relation sits silently instead of throwing). The alternative is keeping a validation-only setter that must be called inside the caller's `update` (no internal wrapping). Flagging for @wittjosiah to pick a direction before I change it.

---

packages/core/echo/echo/src/Entity.ts
wittjosiah: If the accessors survive the other thread, agreed — the setter is objects-only (relations/types are not merge subjects yet), so `Obj` is the more honest home; they landed on `Entity` to sit beside `getMeta`, but `getMeta` is kind-agnostic and these aren't. If the other thread lands on dropping the functions entirely, this resolves itself.

---

packages/core/echo/echo-client/src/merge/merge-executor.ts
wittjosiah: Merging is in the worker — `ConvergenceKeyMerger` runs off the indexing stream (`echo-host/db-host/convergence-key-merge.ts`), triggered by the durable intent log, with an index point-lookup so the automatic path never scans. This client executor backs only the explicit `db.mergeDuplicates()` call, which exists for the one thing the worker deliberately never does: rewriting references to name the survivor directly (refs to losers already resolve through `mergedInto`; rewriting is an optimization and the old-client compatibility path). A client pass racing the worker converges rather than fights — same pure function of the candidate set.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
wittjosiah: Deliberate for this one call, and only here: the automatic path (the worker) detects via an index point-lookup on the keys seen in each indexing batch and never scans. This explicit pass hydrates the space because reference rewriting has to visit every referrer, and it's documented as the cost of an explicit maintenance call.

The footgun concern is fair though — an innocuous-looking `db.mergeDuplicates()` that hydrates the whole space invites routine use. Two options: keep it documented as-is, or demote it out of the everyday `Database` surface into the planned plugin-doctor "merge now" repair action (its only intended consumer). @wittjosiah's call.

---

packages/core/echo/index-core/src/migrations/convergence-key-intents/0001_init.sql
```diff
@@ -0,0 +1,15 @@
+--
+-- Write-ahead intents for convergence-key merging. Rows are inserted in the same transaction that
+-- commits index rows and cursors, and deleted only after the merge pass services the key — so a
+-- crash or a faulted pass can never leave a detected duplicate unserviced.
+--
+-- `IF NOT EXISTS` is required: databases created by releases that carried the table under the
+-- entity-meta store already hold it.
+--
+-- Immutable: recorded in `convergence_key_intent_migrations` and never re-run.
+--
```
dmaretskyi, line 17: the way I understand this, is whenever any object with convergence key enters the indexer, the intent is pushed. intent being here does not mean there are 2+ objects with this key, only that we want to check if the group named by this key needs to be merged. when the merger processes the group, the entry is vacated.

if this is right, please document this in the comment

packages/core/echo/echo/src/internal/Merge/index.ts
```diff
@@ -0,0 +1,5 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+export * from './merge';
```
dmaretskyi: the internal/Merge directory implies a `Merge` API module, which was since removed, move this to common and keep lowercase

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
```diff
@@ -0,0 +1,538 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+import { next as A } from '@automerge/automerge';
+import { type DocHandle, type DocumentId } from '@automerge/automerge-repo';
+
+import { type Context } from '@dxos/context';
+import { type DatabaseDirectory, type EntityStructure, PROPERTY_ID } from '@dxos/echo-protocol';
+import { mergeCandidates, resolveMergeRedirect } from '@dxos/echo/internal';
```
dmaretskyi: i dont like this hooking into echo/internal -- is it possible to move all merge code into echo-host?

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
```diff
@@ -0,0 +1,538 @@
...
+/**
+ * Merges convergence-key duplicates surfaced by the indexing intent log — the worker-side trigger.
+ *
+ * The merge operates on the raw document structures via the storage-independent core in
+ * `@dxos/echo/internal`; the writes replicate to clients like any other change, and the
+ * `documentsSaved` event re-indexes the tombstones, which is what removes the losers from query
+ * results everywhere.
+ */
+export class ConvergenceKeyMerger {
+  readonly #deps: ConvergenceKeyMergerDeps;
```
dmaretskyi, line 87: this needs more logging in the success case: managing intents, logging ids of objects merged and candidates removed, etc..

packages/core/echo/echo-client/src/query/query-result.ts
```diff
@@ -14,8 +14,29 @@ import { log } from '@dxos/log';
...
+import { getObjectCore, isEchoObject } from '../echo-handler';
 import { type QueryContext, type SourceEntry } from './query-context';
 
+/**
+ * True when any part of the query asks for deleted entities.
+ *
+ * `.options({ deleted })` produces a node that scoping wraps further, so the flag can sit at any
+ * depth rather than on the root.
+ */
+const _queryIncludesDeleted = (node: unknown): boolean => {
```
dmaretskyi: make type-safe

packages/core/echo/echo-client-e2e/src/merge.test.ts
```diff
@@ -0,0 +1,570 @@
...
+import { waitForCondition } from '@dxos/async';
+import { Database, Entity, Filter, Obj, Query, Ref, Relation } from '@dxos/echo';
+import { findMergeDuplicates, mergeCandidates, toMergeCandidate } from '@dxos/echo/internal';
+import { TestSchema } from '@dxos/echo/testing';
+import { EffectEx } from '@dxos/effect';
+
+import { EchoTestBuilder } from '../testing';
+import { getMergedFrom, mergeDuplicates, resolveMerged, rewriteReferences } from './merge-executor';
+
+describe('convergence-key merging', () => {
```
dmaretskyi, line 15: this suite belongs in echo-client-e2e. also I want integration.test.ts-like tests with 2 peers replicating -- demostrate the case of two peers creating duplicates, and then them getting merged -- the whole system should settle eventually

packages/core/echo/echo-client/src/proxy-db/database.ts
```diff
@@ -404,6 +405,35 @@ export class DatabaseImpl extends Resource implements EchoDatabase {
...
+   * path deliberately leaves alone: rewriting references to name the survivor directly. Refs to
+   * losers already resolve through the redirect; rewriting them is an optimization and the
+   * compatibility path for clients too old to follow `mergedInto`.
+   *
+   * Being a pure function of the entities it sees, every peer computes the same result, so a
+   * pass racing the worker converges rather than fights, and a pass with nothing to merge
+   * writes nothing. Detection here scans the whole space (hydrating it) — acceptable for an
+   * explicit call; the worker's index point-lookup is what keeps the automatic path cheap.
+   */
+  async mergeDuplicates(): Promise<MergePassResult> {
```
dmaretskyi: is this even needed since the worker is doing merging automatically?

packages/core/echo/index-core/src/migrations/convergence-key-intents/0001_init.sql
wittjosiah: Exactly right. Documented in this migration's header and in `ConvergenceKeyIntentStore`'s class doc: an intent means "check whether the group named by this key needs merging", not "duplicates exist" — one is recorded for every keyed object the indexer processes, a group of one is serviced as a no-op, and the vacate is bounded by the id captured at read time so a concurrent indexing pass's intents survive.

---

packages/core/echo/echo/src/internal/Merge/index.ts
wittjosiah: Done — the module now lives at `internal/common/merge.ts` (lowercase, no directory), exported through the `common` barrel; the `internal/Merge/` directory is gone. It deep-imports `../Obj/snapshot` and `./api/meta` rather than the barrels to keep the module graph acyclic (`Obj` imports `common`).

---

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
wittjosiah: Not all of it can move: the pure core is load-bearing on the client side too. `resolveMergeRedirect` backs the resolver's redirect-following in echo-client (`hypergraph.ts`, `object-core.ts`) — that's what makes an un-rewritten ref to a merged-away loser reach the survivor — and `mergeCandidates`/`toMergeCandidate`/`findMergeDuplicates` back the explicit `db.mergeDuplicates()` executor. Since echo-client can't depend on echo-host, the shared core has to sit below both, and `@dxos/echo`'s internal entrypoint is that layer (same pattern as index-core importing `ATTR_META` etc. from `@dxos/echo/internal`).

What IS host-only already lives in echo-host: `ConvergenceKeyMerger` and everything that touches raw documents. If the client-side `db.mergeDuplicates()` executor were dropped or demoted (per the query-everything thread), the client would keep only `resolveMergeRedirect` — a ~10-line pure function — and the rest of the core could then move host-side. Happy to go that way if that thread lands on removing the client executor.

---

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
wittjosiah: Added: each merged group logs its key, winner, and loser ids; each serviced redirect logs the folded fields and whether the tombstone was re-asserted; and the orchestrator logs the intent take (spaces, key count, high-water id) and the clear (count, bound). All at default level via `@dxos/log`.

---

packages/core/echo/echo-client/src/query/query-result.ts
wittjosiah: Done — `_queryIncludesDeleted` now takes `QueryAST.Query` and walks it with `QueryAST.visit`, reading `node.options.deleted` off the typed `options` node; no casts or record sniffing.

---

packages/core/echo/echo-client-e2e/src/merge.test.ts
wittjosiah: Moved — `merge.test.ts` and `convergence.test.ts` now live in `echo-client-e2e` beside `integration.test.ts` (the merge executor is exported from `@dxos/echo-client` for them).

The 2-peer replicating scenarios you describe already exist in `convergence.test.ts`, over `TestReplicationNetwork` with a worker per peer and no explicit merge call anywhere: "two peers seeding the same state converge on one object" (both workers merge independently and agree), "a partial view merges into a chain that still resolves to the global minimum" (peer 2 merges a subset, workers collapse the rest, straggler edit survives the chain), and a deleted-duplicate case. Each asserts eventual settling by polling the observable query result on both peers. If there's a scenario shape missing from that set, name it and I'll add it.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
wittjosiah: For merging itself, no — the worker covers it. It exists for reference rewriting, which the worker deliberately never does (refs to losers resolve through `mergedInto`, so rewriting is an optimization plus the compat path for clients too old to follow the redirect). Same question is open on the thread above: if we're fine leaving refs redirect-resolved (or moving "merge now + rewrite" into the plugin-doctor repair action), this method can go — and dropping it would also let most of the shared merge core move host-side, which answers your `echo/internal` layering concern on `convergence-key-merge.ts`. Josiah's call; happy to remove it.

---

packages/core/echo/echo/src/Entity.ts
wittjosiah: Done, your second form is the API now: the accessors are dropped entirely. `meta.convergenceKey` is assigned inside the caller's `update` callback, so it batches with their other writes; the schema field's doc comment carries the contract (objects only, empty string is not a key — enforced by the engine layers, not at write time). Ratified by Josiah.

---

packages/core/echo/echo/src/Entity.ts
wittjosiah: Moot now — the accessors were dropped per the other thread, so nothing remains to place in a namespace.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
wittjosiah: Gone — `db.mergeDuplicates()` is deleted, and with it the last `Filter.everything()` in the branch. Reference rewriting moved into the worker, index-driven: the reverse-ref index names each referrer of a merged-away loser and the exact property holding the ref, so repointing is a point load per referrer, run right after the merge's tombstones (and again whenever a re-indexed loser is serviced, which re-covers referrers written concurrently with the merge). Design + accepted freshness residual in `DESIGN.md` §4.11.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
wittjosiah: Removed. The one job it had left — reference rewriting — now runs in the worker off the reverse-ref index (§4.11), so the method, the client merge executor behind it, and the whole `Filter.everything()` pass are deleted. Ratified by Josiah.

---

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
wittjosiah: Done — dropping the client executor (per the `db.mergeDuplicates()` threads) unblocked this. The merge core (`MergeCandidate`, `mergeCandidates`, `toMergeCandidate`, `findMergeDuplicates`) now lives in `echo-host/src/db-host/merge-core.ts` with its unit suite alongside; `@dxos/echo/internal` keeps only `resolveMergeRedirect`, the ~10-line pure resolver the client's redirect-following and transitive-deletion walks genuinely need.

---

packages/core/echo/echo-client/src/merge/merge-executor.ts
wittjosiah: Now moot in the strongest way: this file is deleted. Merging was already worker-side; with §4.11 the reference rewriting is too (index-driven off the reverse-ref index), so the client keeps no merge code beyond `resolveMergeRedirect`.

---

packages/core/echo/echo-client-e2e/src/merge.test.ts
wittjosiah: Since the move, both suites now target the worker end-to-end: the client staging calls are gone (state is staged through `getObjectCore` where a pre-existing redirect is needed, exactly as a peer's worker would write it), rewrite assertions observe the worker's index-driven reference rewriting, and merge records are read at the storage layer. The 2-peer replication scenarios stand as described.

---

.agents/projects/registry.yml
```diff
@@ -549,8 +549,9 @@ projects:
     summary: ECHO-level deterministic object merging keyed on identity key (meta key + version) — duplicates from uncoordinated per-peer space initialization merge into a min-id winner with redirects and reference rewriting.
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d6848594ce — the summary now describes identity as `meta.convergenceKey`.

---

.changeset/convergence-key-merge-core.md
```diff
@@ -0,0 +1,23 @@
...
+
+Reference rewriting is index-driven and also runs in the worker: the reverse-reference index names each referrer of a merged-away loser and the exact property holding the ref, so repointing is a point load per referrer — no client, no scan, and deliberately no `Filter.everything()` anywhere. It runs right after a merge's tombstones and again whenever a re-indexed loser is serviced, which is what catches referrers written concurrently with the merge. Rewriting is an optimization plus the compatibility path for clients too old to follow `mergedInto` — losers keep replicating rather than being erased, so a reference that was never rewritten still resolves through the redirect.
```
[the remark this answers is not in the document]
wittjosiah: Outdated — the changeset was rewritten in cf05386c3e and no longer makes any claim about cross-space references; it describes redirect resolution without qualifying by space. Nothing further to change here.

---

packages/core/echo/echo-client/src/hypergraph.ts
```diff
@@ -555,6 +557,10 @@ export class HypergraphImpl implements Hypergraph.Hypergraph {
       if (obj) {
         return obj;
       }
+      const survivor = this.#followMergeRedirect(db, objectId);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d6848594ce. `_onUpdate` now handles the miss: when `getObjectById` finds nothing for a pending id, it first tries `#followMergeRedirect` (survivor already loaded → emit immediately), and otherwise resolves the redirect chain over the working set including tombstones (`#resolveMergeSurvivorId`). If the id turns out to be a merged-away loser, the pending listeners are re-registered under the survivor's id, so the survivor's own load fires them. Callbacks no longer stay pending when the tombstone lands before the survivor.

---

packages/core/echo/echo-client/src/query/query-result.ts
```diff
@@ -222,6 +239,8 @@ export class QueryResultImpl<T extends Entity.Unknown = Entity.Unknown> implemen
     entries: QueryResult.EntityEntry<T>[];
     grouped: boolean;
   } {
+    entries = this._collapseDuplicates(entries);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d6848594ce. `_collapseDuplicates` now returns `{ kept, removed }`, and `_assembleGroups` subtracts each distinct removed member from its group's server-side `count` before computing the aggregates, so `count` agrees with `items`/`min`/`max`. Covered by a new unit test in `query-result.test.ts` (three same-key members, one tombstoned → `count: 2`, two items); the test fails without the fix.

---

packages/core/echo/echo-host/src/db-host/convergence-key-merge.ts
```diff
@@ -0,0 +1,751 @@
...
+      }
+      if (entity.meta === undefined) {
+        entity.meta = { keys: [] };
+      }
+      const keys = entity.meta.keys;
+      for (const key of result.keys) {
+        if (!keys.some((existing) => existing.source === key.source && existing.id === key.id)) {
+          keys.push(_clone(key));
+        }
+      }
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d6848594ce, as proposed: the unreachable `entity.meta === undefined` branch is replaced by initializing `entity.meta.keys` when absent. Added a unit test where the winner's stored meta carries a `convergenceKey` but no `keys` array and a loser contributes a foreign key — the merge now succeeds and the key lands on the winner.

---

packages/core/echo/index-core/src/indexes/entity-meta-index.ts
```diff
@@ -442,6 +487,10 @@ export class EntityMetaIndex implements Index {
                   : null;
               // Parent (nullable).
               const parent = preserveBody ? priorRow.parent : (castData[ATTR_PARENT] ?? null);
+              // Convergence key (nullable) — from the meta section of the serialized object.
+              const convergenceKey = preserveBody
+                ? priorRow.convergenceKey
+                : ((castData[ATTR_META] as { convergenceKey?: string } | undefined)?.convergenceKey ?? null);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d6848594ce. The raw `@meta.convergenceKey` is now read as `unknown` and normalized with a `typeof === 'string'` check — anything else indexes as `null`. A full Effect Schema decode is disproportionate for a single primitive field at this hot path (the surrounding meta fields are read the same way), so the guard is the narrow typed boundary instead. Added a unit test: a string key is queryable by key, a numeric key indexes as no key.

---

packages/core/echo/echo-client/src/query/query-result.test.ts
```diff
@@ -45,10 +47,46 @@ describe('QueryResultImpl', () => {
...
+      const people = [
+        db.add(Obj.make(TestSchema.Person, { name: 'Alice', age: 30 })),
+        db.add(Obj.make(TestSchema.Person, { name: 'Bob', age: 30 })),
+        db.add(Obj.make(TestSchema.Person, { name: 'Carol', age: 30 })),
+      ];
+      await db.flush();
+      db.remove(people[2]);
+      await db.flush();
+
+      // The source counts every member it grouped, tombstones included; the client drops the tombstone
```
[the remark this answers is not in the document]
wittjosiah: Not applicable — this is a single comment wrapped across two lines; the sentence continues on line 67 and ends with a period there ("…has to follow."). No change needed.

---

.changeset/convergence-key-merge-core.md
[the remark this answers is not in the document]
wittjosiah: The cross-space rejection in `_resolveAsync` ("Cross-space references are not yet supported") is pre-existing on `main` and independent of this PR: a cross-space `Ref.load()` fails before and after a merge alike, whether or not the target was ever merged. The changeset's "references still resolve through" is deliberately relative — every reference that resolves today keeps resolving when its target becomes a redirect — and that statement holds exactly, since the redirect lookup sits behind the same gate. Documenting the general cross-space limitation belongs to the resolver, not to this changeset, so leaving the wording as is.

---

PR #12415, opened 2026-07-31
composer-app: startup-latency (demand-driven activation, per-plugin start events, streaming start, suspenseful client)

wittjosiah:
Startup-latency work for Composer (`.agents/projects/startup-latency/`), based on the capability-activation branch this PR targets. Built as a sequence of measured waves — every lever below landed with its own A/B, and several were reverted when the measurement disagreed with the theory.

## Headline

**Timings — real hardware, warm-cold (persisted identity, hard reload), branch N=5, measured 2026-08-05:**

| metric | main | branch | delta |
| --- | ---: | ---: | ---: |
| **navToReady** (identity chrome visible) | 14,369 ms | **5,085 ms** | **−65%** |
| profilerTotal (main:start → ready) | 8,335 ms | 2,202 ms | −74% |

Branch samples (median reported): navToReady `[5424, 5085, 5440, 4996, 5057]`, profilerTotal `[2115, 2228, 2202, 2167, 2209]`.

**Boot graph — current bundle:**

| metric | main | branch | delta |
| --- | ---: | ---: | ---: |
| boot requests | 520 | **23** | −96% |
| boot bytes (statically loaded) | 8.8 MB | **5.31 MB** | −40% |

**Modules at ready:** median **322** over 5 warm-cold repeats `[329, 322, 329, 321, 322]`, budget 350. This rose from 291 as several modules moved deliberately *onto* the startup pass — app-shell state (deck/simple-layout/spotlight), the theme and storybook React contexts, the client and attention modules — because the shell cannot paint without them. See "Shell-paint and URL-restore fixes" below.

One caveat on `profilerTotal` specifically: it is only comparable at constant boot-request count. Its window opens at `main:start`, which on main fires late because ~520 serialized requests must resolve first, so work that sat *outside* the window on main is *inside* it here. **`navToReady` is the honest end-to-end number.**

---

## 1. Demand-driven activation

Modules no longer activate on the startup pass by default; they declare what they are *for*.

- **Demand gates as the module-spec default** — makers declare them: `operationHandler` on own-handlers-requested, `skillDefinition` on skills-requested, `createObject` on `CreateObjectRequested`.
- **Role-gated surfaces** — `SurfacesRequested(role)`; `SurfaceComponent` fires on mount with per-manager dedup, `useIsAvailable` fires on miss (self-healing). Roles were generated from a live bundle probe, then codemodded into 49 plugins.
- **Per-operation lazy handler sets** — definitions enumerate without loading bodies; resolving an operation imports only that operation&#39;s module. Flipped back to eager once wave 2 thinned definitions (measured twice by ablation: eager cost +2.2 s / +1.6 MB *before* thinning, ~0 after).
- **Scheduler throughput** — bounded wave concurrency with singleton-providers-first ordering, host-yield between activations, scoped demand pulls, bounded-concurrent bootstrap enables.

## 2. Definition weight (imports/exports)

Boot loads ~100 plugin definitions, so every byte a definition drags in is boot cost.

- Definitions import capability **tags, not implementations**; cross-plugin type references go through `/types`.
- **Per-API subpath entrypoints** for `@dxos/compute`, `@dxos/app-toolkit`, `@dxos/app-framework`, enforced by a new **`dxos-subpath-imports` lint rule** with autofix.
- Dropped 97 `plugin.ts` stub operation re-exports (the node CLI, their only external consumer, got its own entry).
- `pipeline-rdf` FactStore tag/implementation split (1.5 MB of SPARQL off the definition closure).
- Boot fetch 31.1 → 28.8 MB in the first batch alone.

## 3. Per-plugin start events, demand-gated

- Every off-critical-path module rides its plugin&#39;s `Events.Start` (`ActivationEvent.pluginStart` convention), swept across ~67 plugins.
- **Cross-plugin contributions ride the *consuming* plugin&#39;s event** — skills → assistant, editor extensions → markdown, connectors → connector, so a contribution costs nothing until its host is in use.
- **A plugin&#39;s own start is now a demand signal, not a trickle.** The module loader fires it when one of the plugin&#39;s modules contributes a `ReactSurface` — the feature being rendered *is* the demand — so an unvisited feature&#39;s contributions never load. Forked rather than awaited: contributions are read through reactive capability atoms, so consumers re-render when they land.
- Contributions that no surface can gate ride the feature they belong to instead. App-graph builders default to the host&#39;s `Idle` wave — gating a builder on its own plugin&#39;s start is a deadlock, since the builder is what puts the item in the navtree, so the item could never be opened. Skills default to the assistant&#39;s start.
- **Headless state is not gated on its plugin&#39;s UI.** Comment sync, compute graphs and native-filesystem state work in a markdown document with no review/sheet/filesystem surface ever rendered, so gating them on that plugin&#39;s start conflated &#34;the UI is on screen&#34; with &#34;the state exists&#34;. They are ungated and pulled by the consumers that need them.
- **All 71 React surfaces are role-gated.** 19 had been gated on their own plugin&#39;s start (unreachable once the blanket fire was gone) and 3 declared no roles at all.
- **The framework fires the idle wave, not the host.** `Idle` is an `ActivationEvents` member, so the scheduler owns the wait; composer&#39;s bespoke `AfterStartupPlugin`, which re-implemented it, is deleted.
- Settings stay eager, deliberately: values are read app-wide through the strict `useAtomCapability` hook, including by components that mount with the shell at ready.
- `DeferredStartup` and `SkillsRequested` deleted.

## 4. Streaming start

Plugin definitions activate as they register rather than after the full enable chain completes; ready waits for true quiescence. Controlled A/B: pre-client delay −22%.

## 5. Suspenseful client

`client.initialize()` forks off the startup pass so the shell renders while the client boots.

- `Client.waitUntilInitialized()` (Trigger-backed); `useClient` throws the init promise to the per-Surface Suspense boundaries that already existed.
- `ClientProvider` gains a `suspend` mode that provides context immediately.
- HALO adapters are construction-safe pre-init — streams open at initialization, so a pre-init read is silence, never a false &#34;no identity&#34; that would flash onboarding.
- Modules needing an initialized client ride the new `ClientEvents.Initialized`.
- `EffectEx.runDetached` absorbs teardown interruption of fire-and-forget effect runs while still surfacing real failures.

## 6. Static boot-graph slimming

A static-import BFS with sourcemap byte attribution found leaks that no amount of activation deferral would fix — code reachable from the entry&#39;s *import graph*:

| eviction | bytes |
| --- | ---: |
| plugin-sheet definitions → HyperFormula engine (new engine-free `@dxos/compute-hyperformula/types`) | 548 KB |
| `main.tsx` → `ResetDialog` → FeedbackForm → form/editor/picker stack (now `React.lazy`) | ~1.5 MB |
| `@dxos/client` barrel → worker runtime (new `@dxos/client/worker` subpath — **breaking**, changeset included) | ~1 MB |
| `date-fns/locale` all-locales barrel → per-locale entry | 966 KB |
| `@dxos/edge-client` barrel → auth + ws-muxer (new `/http` subpath) | 809 KB |
| `services/index.ts` re-export (hypercore, wa-sqlite, network-manager, teleport) | 810 KB |
| `@dxos/crypto` sodium (new `/random` webcrypto + `/subtle` subpaths) | ~355 KB |
| `@dxos/credentials` barrel → seedphrase → bip39 (new `/assertions` subpath) | ~150 KB |
| yaml, deferred into the node-only `ConfigService.load` | ~90 KB |

## 7. Per-namespace plugin subpath exports

- **80 plugins** gained namespace entrypoints (**351** total) via codemod; `dxos-subpath-imports` now covers every `@dxos/plugin-*`, driven by the exports map rather than an allowlist.
- **258 consumer files** autofixed; cross-plugin barrel imports 707 → 519.
- Effect is on lazy islands as much as boot: the pilot island&#39;s closure went 8.41 → 7.70 MB with the sheet engine fully decoupled.

## 8. Cycle-safe boot chunking

The boot path was 520 default-split chunks whose per-request overhead dominated startup. It is now a handful of chunks, with the partition computed from the module graph **during chunking** (no checked-in artifact):

- Entry closure over static imports → SCC condensation (Tarjan) → contiguous buckets of the dependency-first topological order, so every cross-chunk edge points backwards and the chunk graph is a **DAG by construction**. ~150 ms per build.
- This matters because rolldown&#39;s own `maxSize` cuts by accumulated size with no regard for dependency order, manufacturing chunk cycles — which have no correct ESM evaluation order (`Tag is not a function` at boot). Its documented remedy, `strictExecutionOrder`, wraps every module body and measured **+1.8 MB** of inhibited treeshaking.
- Also: `vite preview` now serves HTTP/2 under `HTTPS=true`, removing the 6-connection serialization that dominated local preview.

## 9. CI budgets

Nothing watched the eager boot graph, which is how it reached 749 chunks / 8.8 MB unnoticed. Two moon tasks now do, split by what a shared runner can actually decide:

- **`composer-app:check-boot-budget`** (static) — the built `index.html`&#39;s entry script plus its modulepreload closure. **30 entries / 6.00 MB**, today **23 / 5.31 MB**. Bytes catch leaks; count catches the chunk partition silently ceasing to apply — it is *not* a bytes proxy, since bucket boundaries follow the SCC condensation. It gets its own task rather than a step inside `bundle`: a perf budget must not fail the local build you are running in order to measure the thing it budgets.
- **`composer-app:check-startup-budget`** (runtime) — median **modules-at-ready** over 5 warm-cold repeats, budget **350** (today **322**). This is the only metric worth failing a build on here: it is a count, so it barely moves with machine speed (same-commit repeats vary ≤10) while the activation regression this branch shipped and reverted moved it +17…+27. It is also the only gate that can see that class at all — moving a module onto the startup pass changes nothing statically, so the boot budget is blind to it.

Timings are recorded and trended but **never gated**: repeats of one unchanged commit spanned 3,828–7,330 ms `profilerTotal` purely from container contention, against ±1.7% for the same branch on real hardware. The noise is environmental, so a shared runner can never gate them — tracked as a TODO pending a fixed runner.

## Shell-paint and URL-restore fixes

Making the idle default *real* (previously an enable/start race dragged idle-default modules into boot, so strict capability readers found their providers present by luck) surfaced a family of latent bugs. All were found by measurement rather than inspection:

- **Shell state providers declare the startup pass.** Deck, simple-layout and spotlight state build the atoms their shell reads on its first render. The gate belongs on the *provider* — declaring it as the reader&#39;s `requires` demotes the reader into the provider&#39;s later wave instead, which measured as a **6.3 s blank shell** (deck root active at 10,593 ms vs the boot loader dismissing at 5,517 ms). Fixed: ~350 ms.
- **React contexts go through `AppCapability.reactContext`.** A context wraps the tree on its first render by definition, so it cannot be deferred; theme and storybook built their specs directly and inherited the idle default, leaving roots mounted outside `ThemeProvider`/`Tooltip.Provider` (`Tooltip.Trigger must be used within Tooltip`).
- **Capability hooks suspend rather than throw.** `useCapability` could not distinguish &#34;not yet&#34; from &#34;never&#34;; readers now park at their nearest boundary, with one Suspense boundary per root so a late root cannot withhold every other root&#39;s content.
- **URL restore waits on the right thing.** The parse builds its key table from the registered builder extensions, so a URL whose keys had not arrived took an immediate not-found return, never reaching the per-pair node wait. It now re-parses as extensions register — keyed off `builder.extensions`, the atom the key table actually reads, rather than the capability that feeds a *different* subscriber which then registers them.
- **An explicit navigation supersedes the restore**, checked before every write rather than only before the deck `Set`; and the outbound URL sync subscribes at activation with only its unconditional baseline write deferred, so a navigation during the restore window is neither lost nor overwritten.

## Typed lazy handler pairings

`OperationHandlerSet.keyed` paired a definition with a handler module through two `Definition.Any` slots, so a definition could be wired to another operation&#39;s handler and fail only at dispatch. `Operation.lazyHandler` relates them through a shared type parameter and `OperationHandlerSet.lazy` replaces `keyed` outright — **367 entries across 79 packages**, converted mechanically after the existing `keyed` suite was moved over wholesale and passed unchanged. A deliberate mispairing is now rejected at compile time (TS2684).

## Correctness fixes found along the way

- Event waves now await modules loading in concurrent waves (demand-pull race that left the create-space dialog dead; regression-tested).
- `OperationHandlerSet.make` answers per-key lookups so contribution order wins — a later lazy set no longer shadows an earlier override.
- Post-boot `enable()` fires the plugin&#39;s own start event; thread/meeting no longer hard-require the calls manager; the eager calls audio root tolerates the deferred call manager.
- **plugin-calls built its placeholder media tracks at activation.** `MediaManager._open()` created a black-canvas `captureStream()` plus an `AudioContext` on the client-initialized wave — app startup — for a call that may never happen. Now built lazily and memoized on first use.
- **`boot:html-parsed` had three readers and no writer**, so the first segment of the boot waterfall was silently missing. It is now emitted by the injected loader IIFE at `body-prepend`, and immediately shows the loader painting at ~180 ms against `main:start` at ~1,081 ms — a ~900 ms segment that was previously unattributed.
- **The startup profiler was silently off in dev.** `isTrue(str, strict = true)` takes a strictness flag, and the default was being passed through it, so the strict branch demanded the very parameter that was absent.
- The process manager logs an error naming any `LayerSpec` contributed after its one-shot runtime snapshot — a boot-time one-shot consumer of a `multi` capability fails silently at the contribution site otherwise.

## Instrumentation retained

Per-module wait/run/import timing with chunk-import attribution, boot-waterfall milestone marks (including inside `Client.initialize`), the startup profiler (`?profiler=1`), and the warm-cold golden scenario with auto-recording to `composer-app/BENCHMARKS.md`.

## Measured dead ends (recorded so they aren&#39;t retried)

- **Lazy-chunk consolidation makes things worse.** Two implementations took the markdown island from 44 req / 1.13 MB to 68–90 req / 4.9–5.4 MB, and *raised* total chunk count. Only 191 of 3,430 lazy chunks (5%) are mergeable at zero cost.
- **Coarse boot grouping fails**: per-package groups weld each package&#39;s eager and lazy halves (boot 4.03 → 10.07 MB); `$initial` tags span all five HTML entries plus recursive deps (→ 19 MB).
- **Deferring settings fails.** Gating settings modules on a post-ready wave trips the strict `useAtomCapability` invariant in components that mount with the shell. Attempted twice; reverted both times.
- **`requires` does not pull a provider forward across waves.** Pointing a Startup reader&#39;s `requires` at an idle-gated provider demotes the *reader* into the idle wave. Measured, reverted; the gate belongs on the provider.

## Known gaps / follow-ups

- **Parse-vs-treeshake gap (~1 MB)**: the graph available at chunking time follows barrel re-exports that treeshaking later drops, so some lazy-only modules land in boot.
- **react-aria** is the only remaining forbidden package in boot, via react-ui&#39;s core `Input` → `SegmentedInput` → `DatePicker` → `Calendar`. A genuine static dependency, so it needs a component change rather than an import fix.
- **PWA install precaches 96.8 MB / 4,560 files** (64.2 MB js + 30.5 MB wasm) for a ~5 MB boot shell. Shell-only precache with `CacheFirst` for hashed code would cut it to ~2 MB; it changes offline coverage from &#34;everything&#34; to &#34;what you&#39;ve opened&#34;, so it&#39;s a product call. Measured and written up, not implemented.
- `plugin-space` / `plugin-client` / `plugin-graph` export `namespace` *declarations* rather than modules, so they can&#39;t take subpaths until refactored — the remaining 519 barrel imports are concentrated there.
- **Cold `Client.initialize()` is the residual long pole.** Measured in-container — construct 3 ms, first `initialize()` 4,688 ms, second (warm) 123 ms. Composer forks `initialize()` so its paint is unaffected, but node consumers await it directly.
- **fast-check reaches the boot graph and we cannot fix it here**: `effect/Schema.js` statically imports `effect/FastCheck`, and `effect/Schema` is used throughout ECHO, operations and config. It needs an upstream change or a pnpm patch, not a lint rule.
- **`activatesOn` is still folded into a concrete `Idle` when omitted**, so the scheduler reconstructs "the author expressed no opinion" by matching event names (`#isBaselineWave`, `#declaresStartup`). Modelling the absence directly would make the scheduler event-agnostic; attempted and reverted here as too broad for this PR.
- **`e2e` is gated to main/changeset-release/dispatch**, so `check-boot-budget` does not gate this PR — it runs post-merge, or on a manual dispatch with `e2e: true`.
- Timing-based CI gates are blocked on a fixed runner (see §9).

## Verification

Full unit/integration sweep green across the framework and all ~70 migrated plugins; targeted suites re-run after each eviction. Storybook suites green on CI. Every consolidated bundle passes a chunk-level acyclicity check plus a warm-cold e2e boot. Measurement note: benchmark runs verify the served bundle by content marker — `composer-app:bundle` produces it, not `:build`, and a moon cache restoring a stale archive silently invalidated a day of measurements before that check existed.

### Local verification of the CI matrix

Run locally against the same commands as `.github/workflows/check.yml`: `moon run :lint :build` (630 tasks), `moon run :check-module-structure` (257 tasks), `oxfmt --check` (12,314 files), and `moon run :test :test-browser :test-workerd :test-storybook -- --no-file-parallelism`.

https://claude.ai/code/session_01BWf8bjs9M7wyrHCcwqXKzE


packages/plugins/plugin-bluesky/src/operations/index.ts
```diff
@@ -2,12 +2,16 @@
...
-export const BlueskyHandlers = OperationHandlerSet.lazy(
-  () => import('./get-bluesky-targets'),
-  () => import('./materialize-target'),
-  () => import('./sync'),
-);
+import { SyncBlueskyTargets } from './definitions';
+import { MaterializeBlueskyTarget } from './definitions';
+import { GetBlueskyTargets } from './definitions';
+
+export const BlueskyHandlers = OperationHandlerSet.keyed([
```
dmaretskyi: Is this typed: definition and implementaiton are required to match?

perhaps do 
```ts
OperationHandlerSet.lazy(
  GetBlueskyTargets.pipe(Operation.lazyHandler(() => import('./....'))),
....
)
```ts

Comment (dmaretskyi) [reply packages/plugins/plugin-bluesky/src/operations/index.ts]: Operation.lazyHandler: (definition) => getImpl => Operation.LazyHandler

Comment (wittjosiah) [reply packages/plugins/plugin-bluesky/src/operations/index.ts]: Done, in exactly that shape — `OperationHandlerSet.lazy` + `Operation.lazyHandler` now replace `keyed` outright, so there is one form rather than two.

```ts
export const BlueskyHandlers = OperationHandlerSet.lazy([
  GetBlueskyTargets.pipe(Operation.lazyHandler(() => import('./get-bluesky-targets'))),
  MaterializeBlueskyTarget.pipe(Operation.lazyHandler(() => import('./materialize-target'))),
  SyncBlueskyTargets.pipe(Operation.lazyHandler(() => import('./sync'))),
]);
```

To the typing question: it was not typed before — `keyed` held two independent `Definition.Any` slots, so wiring a definition to another operation's handler compiled and only failed at dispatch. `lazyHandler` relates them through a shared type parameter, so a deliberate mispairing is now a compile error (TS2684), covered by a `@ts-expect-error` case in the suite.

Converted mechanically across **367 entries in 79 packages**, after moving the existing `keyed` test suite onto `lazy` wholesale and confirming it passed unchanged — so equivalence was established before the sweep rather than asserted after it.

---

PR #12420, opened 2026-08-01
echo-panproto: object lensing — drive one object through a second declared type

wittjosiah:
## Summary

A **lens** binds two declared ECHO types and lets you drive a live object through the second one. There is only ever one object in the database — the lens is a view of it, not a copy — and target properties that nothing maps store themselves in the object's annotation dictionary rather than breaking.

The payoff: a surface written against a type works on any object that lenses to it. The `ObjectLens` story demonstrates this literally — both panels are the *same* `Form` component, differing only in the schema they are handed.

Ships in `@dxos/echo-panproto` (beside the existing `Panproto` wire lens) as a staging ground; promotion into `@dxos/echo` is a later import-path change.

## What's here

**`Lens` namespace** (`packages/core/echo/echo-panproto/src/Lens.ts` + `src/lens/*`, 26 tests)

- `Lens.make(id, Source, Target, mapping)` — both ends are declared types, so `Lens.Target` is free (`Type.InstanceType` of the declared target), even for a persisted lens. `Lens.coded` covers transforms a per-property mapping cannot express (parsing).
- **Partial mapping**: explicit entry → else automatic on matching name + compatible type → else the target property *overlays* itself into the entity annotation dictionary. Unconsumed source properties drop and are restored by `put`.
- **`Lens.coverage`** grades every property `explicit`/`automatic`/`overlaid`/`dropped`/`suspicious`. A name match with incompatible types is `suspicious` — never auto-mapped and never auto-overlaid, since that would record the same fact twice and let the copies drift.
- **Write minimality is a type error, not a convention.** The `Write` vocabulary is `assign | splice | overlay` with no whole-object replace expressible. This is required for live collaboration: a snapshot write would clobber a concurrent peer.
- A lensed object reports the **target's** typename (so existing surfaces render it) while `Obj.getURI` still resolves to the base object. `Obj.update` works through the lens — the callback's assignments are recorded and inverted as one transaction — with no core change.
- `Lens.checkLaws` (GetPut) runs host-side over the object's own values, needing no engine call and no mutation. It caught two real defects while being written, which is the argument for having it.

**Database verification** (`echo-client-e2e/src/lens.test.ts`, 4 tests) — proves it against a real automerge-backed `Task`: live handle, overlay surviving `peer.reload()`, a write touching only named properties, and **two peers concurrently editing one object** — one through the canonical type, one through the lens — with both edits surviving. That last test is what fails if a lens write is not minimal.

**`@dxos/stories-lens`** (private) — 6 passing stories across two demos:

- **ObjectLens** — `Task` ↔ `GtdTask`. Same `Form`, two schemas, plus a raw inspector showing every lensed edit landing as an ordinary property and the target-only fields landing in `meta.annotations`.
- **RichTextLens** — `Text` ↔ a block tree, via a *coded* lens over `@lezer/markdown` source offsets with inline marks. The core markdown editor on one side, a basic ProseMirror editor on the other, neither aware of the other. Every block carries its source range, so a block edit splices that range alone and the markdown editor's character-level automerge edits survive alongside it.
- Each demo has `Default` (renders only), `Spec` (assertions), and `Collaboration` (two peers over a real invitation).

## Notes on decisions

**Why both ends are declared types.** Deriving the target from the lens was the original shape; binding a declared target dissolves the hardest problem in it (inferring a static type through a chain of combinators) and holds for persisted lenses too. No combinator generics to maintain.

**Why the overlay.** Target fields with no counterpart must not break, and they must not become stray properties on the base object. They are validated against the target's declaration for that property. The cost is that they are invisible to queries and indexes until promoted — recorded as a risk, with `Lens.promote` in the backlog.

**Two test lessons worth carrying.** Both were caught by the work, not by review, and both are recorded in the ledger:

- `userEvent.click` reports no pointer coordinates, and CodeMirror derives the caret from them — so clicking a specific `.cm-line` still types at document position 0. A document-wide `toContain` cannot tell a working caret from a broken one; the specs now pin the exact line each edit produced.
- The theme sets `font-synthesis: none` and pins `font-variation-settings: 'wght' 400` at `:root`, so a bare `font-weight: 700` computes as 700 and renders as regular. Weight assertions go through `fontVariationSettings`. Both fixes are mutation-checked.

## Planning docs (no code)

`.agents/projects/lenses/{DESIGN,TASKS}.md` carry the design and the ledger, including two long-term plans that are **specified but not built**: migration (phased M0 research → M1 lens-backed single-object → M2 lossless/cross-object, against a "no unconflicted change may be lost" bar) and cross-object lenses (composition vs projection). The migration plan takes an explicit dependency on the object-merging engine from #12410, whose research falsified the derived-object-id mechanism this plan originally assumed; the feedback loop in the other direction is recorded there too and has landed.

## Test plan

- `echo-panproto`: 30 tests (26 lens + 4 existing wire lens), lint clean
- `stories-lens`: 11 unit tests, 6 storybook stories, build + lint clean
- `echo-client-e2e`: 286 passed incl. `lens.test.ts` (4)
- `pnpm format` run; merged with `origin/main` (conflicts were generated-file list appends — `.changeset/config.json` and `tsconfig.all.json` each gained `stories-lens` here and `stories-projects` on main; both kept)

---
_Generated by [Claude Code](https://claude.ai/code/session_01J9SQKrd1t6B2VnnAWt33iB)_

## Summary by CodeRabbit

* **New Features**
  * Added bidirectional ECHO lenses for alternate typed object views.
  * Supports automatic and custom mappings, conversions, read-only fields, overlays, persistence, coverage checks, and live updates.
  * Added React hooks for subscribing to and editing lensed views.
  * Added task and Markdown rich-text lens examples with collaborative editing.
* **Documentation**
  * Added API documentation, design specifications, and implementation status notes.
* **Tests**
  * Added unit, end-to-end, and Storybook coverage for synchronization, persistence, editing, and collaboration.


wittjosiah:
@coderabbitai review

The previous run failed because the head commit moved during the review (`876ce5b2` → `74f6f3d2`) — that push merged `main` and consolidated four changesets into one, so the earlier file list is out of date.

The "resolve merge conflicts" suggestion is stale from the same cause: the branch is now level with `main` (base `e7f0d9e0`), and `git merge-tree` against it is clean.

---

PR #12455, opened 2026-08-03
Add browser-based terminal emulator for Effect CLI applications

wittjosiah:
## Summary

Introduces `@dxos/react-ui-terminal`, a new React component that hosts an `@effect/cli` command tree entirely in the browser using xterm.js. This enables interactive CLI applications to run client-side with full support for inline editing, history recall, ANSI styling, and interactive prompts.

## Key Changes

- **New package `@dxos/react-ui-terminal`**: A React component wrapping xterm.js that provides a complete terminal environment for Effect CLI commands
  - `Terminal` component: Main React component that mounts an xterm instance and runs a command shell
  - `XtermBridge`: Adapter between xterm and the Effect `Terminal` service contract
  - `readLineResult`/`readLine`: Line editor with support for editing, history, and readline control chords (Ctrl-A/E/U/K/L, arrow keys, etc.)
  - `decodeInput`: Converts raw xterm data stream into Effect `UserInput` events, handling escape sequences, control characters, and alt-modified keys
  - `runShell`: REPL loop that tokenizes input and dispatches commands through `Command.run`
  - Theme integration: Dynamically applies design system color tokens to xterm palette, supporting light/dark mode

- **Platform abstraction in `@dxos/cli-util`**: 
  - Moved platform-specific code (`openBrowser`, `copyToClipboard`) to conditional imports via `#platform` alias
  - Separated Node.js implementation (`platform.node.ts`) from browser implementation (`platform.browser.ts`)
  - Removed OAuth server re-export from root entry (moved to separate export path)

- **CLI command exports**: Added `./commands` export to `@dxos/plugin-space` to enable importing command definitions in browser bundles

- **Storybook stories**: Demonstrates both a self-contained demo command tree and integration with live DXOS services (space, database, queue commands)

## Implementation Details

- **Input handling**: Decodes xterm's raw `onData` stream rather than DOM key events, enabling paste support (pasted text arrives as one chunk and expands to one event per character)
- **Line editor**: Maintains buffer, cursor position, and history state; redraws on every keystroke with ANSI escape sequences for cursor positioning
- **Service layer**: Provides Effect `Terminal`, `Console`, and `FileSystem` implementations backed by xterm, allowing unmodified Effect CLI code to run in the browser
- **Theme system**: Uses a probe element and canvas rasterization to resolve design system CSS custom properties to hex colors that xterm accepts
- **Responsive layout**: Uses ResizeObserver and FitAddon to adapt terminal dimensions to container size

https://claude.ai/code/session_018kVZKeU3py5wzcpEMtg7DT

## Summary by CodeRabbit

- **New Features**
  - Added an interactive CLI terminal to Devtools for running supported commands in the browser.
  - Added browser terminal support, including command history, editing, help aliases, colored output, error handling, and clipboard/browser actions.
  - Added terminal UI components and testing utilities.
  - CLI commands now load on demand, improving initial application startup performance.

- **Documentation**
  - Documented the updated command activation sequence and browser Devtools behavior.


wittjosiah:
Worked through the review in `080600f4`. Four of the findings were real defects, with regression tests for the two that were reachable from a keyboard:

- **Unrecognized escape sequences leaked their bytes.** The table matched only unmodified navigation keys, so ctrl-right (`\x1b[1;5C`) fell through to the alt-modified branch and typed `1;5C` into the line. CSI and SS3 sequences are now measured and discarded.
- **Alt chords typed their character.** Nothing filtered `key.meta`, so alt-f inserted an `f`.
- **`readTerminal` replaced ordinary spaces with ordinary spaces** — the non-breaking padding its own comment describes was never normalized.
- **Disposal raced the shell interrupt**, so finalizers could write through a disposed bridge. Disposal is now chained onto the interrupt.

Two I'm not taking:

1. **`CliPanel/index.ts` default export.** The default alias exists only to satisfy `React.lazy()`, and all four sibling containers do exactly the same — `RegistryPanel/index.ts` is character-for-character identical. Moving the mapping into `containers/index.ts` for this one panel would make it the odd one out without removing a default export from the package.

2. **Typed Effect errors for the platform utilities.** `platform.node.ts` is a 100% pure rename in this PR — `git diff -M` reports zero changed lines. Changing `Effect<void, Error>` to a tagged error is a public API change to functions this PR doesn't otherwise touch, so it belongs in its own change rather than riding along here.

I did fix three pre-existing bugs in that renamed file, since the split makes it the node half of a utility this PR is specifically making platform-aware: `openBrowser` spawned the cmd.exe builtin `start` directly (ENOENT on Windows), a missing `xclip` rejected from `'error'` before the xsel fallback could run, and two comments were missing terminating periods.

Separately, the `layer` prop on `Terminal` no longer widens its failure channel to `any`, which the repo's no-cast rule disallows. A contributed layer that can fail to build is now discharged by the host via `Layer.orDie`, since the host is the only place that can decide what a shell with no services should do.

Changesets consolidated to one file naming one package per lockstep group, per `agents/instructions/changesets.md`.

Verification: 51 unit tests (was 48), both storybook specs, full workspace build and lint all green.

---

PR #12462, opened 2026-08-04
ui, plugins: make component modules react-refresh boundaries

richburdon:
## Why

`vite-plugin-react` only makes a module a refresh boundary when **every** export is a component. A module that exports a hook, context, constant or helper beside its component cannot self-accept: the update propagates to its importers and, when none accept, Vite falls back to a full page reload. In Composer that reload costs a ~30s restart, so editing any UI touched by such a module throws away app state.

Type-only exports are erased and do not count, so `export type` is left alone throughout.

## What

Pure move plus barrel re-export across **131 modules** — 102 in `packages/ui`, 29 in `packages/plugins`. Each non-component export moved to a sibling module (`FooContext.ts` for a context plus its hook, a kebab-case `.ts` for helpers and constants), the component module imports from it and no longer re-exports it, and the previously public names are re-exported from the directory barrel so **package public APIs are unchanged**.

Every new module carries a one-line comment saying why it is separate, matching the five reference splits already on `main`.

Packages touched: `react-ui`, `react-ui-components`, `react-ui-list`, `react-ui-canvas`, `react-ui-canvas-editor`, `react-ui-canvas-compute`, `react-ui-mosaic`, `react-ui-form`, `react-ui-experimental`, `react-ui-editor`, `react-primitives` (`react-input`, `react-list`), `react-ui-menu`, `react-ui-geo`, `react-ui-grid`, `react-ui-dnd`, `react-ui-chat`, `react-ui-attention`, `react-ui-table`, `react-ui-syntax-highlighter`, `react-ui-search`, `react-ui-markdown`, `react-ui-gameboard`, `react-ui-card`, `react-ui-board`, `brand`, and 19 plugins.

### Notable decisions

- **`react-ui-canvas-compute`** — every shape module exported its Effect schema, factory and `ShapeDef` beside its component, so moving only the lowercase exports would have gained nothing. All three moved into sibling `*-def` modules. `trigger-spec.ts` exists because both the component and the def module need those helpers, and a value import back from the def would close a runtime cycle.
- **Radix scope-props types** — `Popover`'s and `DropdownMenu`'s previously file-local `ScopedProps` are now `PopoverScopedProps` / `DropdownMenuScopedProps` in the context modules, and re-exported as types from the component modules. Without that, downstream inferred types (`react-ui-menu`, `react-ui-list`) stop being nameable across package boundaries.
- **Def modules import the component; component modules import only types back.** Type-only imports are erased, so no runtime cycle is introduced.

## Measured, not assumed

Verified on a storybook dev server (`storybook-9014`), watching the browser console:

| Edit | Result |
| --- | --- |
| Comment appended to the split `Carousel.tsx` (component-only exports) | `[vite] hot updated: …/Carousel/Carousel.tsx` and **nothing else** — the update stops at the module |
| Same file with one non-component export added back (`export const hmrProbeHelper = () => 1`) | update escapes the module: also `hot updated` `storybook-stories.js via vite-app.js` and `testing/Loading.tsx` |

The probe edits were reverted.

**This does not mean the app stops reloading.** Other boundaries remain — notably capability modules with non-component default exports, and import cycles. What is measured above is that these specific modules now self-accept.

## Intentionally left alone

- `react-ui-canvas-compute/src/shapes/Boolean.tsx` — exports no components at all, and its gate symbols come from a `createSymbol` factory so react-refresh cannot identify them.
- `chat-def.tsx` / `constant-def.tsx` — def modules that supply props to their component; they are not boundaries by design.
- `solid-ui-geo` — Solid, not React; `vite-plugin-react` does not apply.
- `react-ui-components/HtmlViewer/testing.tsx` and `plugin-terra/testing/withAttention.tsx` — story fixtures/decorators, not app modules.

## Verification

Per-package `build` + `lint` (autofixed to zero warnings) + `format` after each commit; `test` where the package defines one (`react-ui`, `react-ui-canvas-compute`, `react-ui-mosaic`, `react-ui-form`, `plugin-magazine`, `plugin-brain` all green). Whole-graph `moon exec :build` was deliberately avoided — it exhausted memory on a previous attempt — so CI is the first full-graph check.


richburdon:
No code change for the coverage report — the flagged lines are relocations, not new logic.

Every file in the "missing lines" table is one of the sibling modules this PR extracts. Codecov has no baseline for a new file path, so pre-existing untested code is re-counted as an uncovered patch. Measured against the base commit (`5b3db1c`), the three largest entries are verbatim moves:

| New module | Code lines | Not present verbatim in the base component module |
| --- | --- | --- |
| `Ghost/ghost-renderer.tsx` | 1151 | 2 |
| `CellEditor/editor-keys.ts` | 93 | 2 |
| `MarkdownStream/create-controller.ts` | 135 | 4 |

The residue in each case is the added import line plus an `export` keyword. Project coverage moved **48.31% → 52.70% (+4.39%)**, and the base/head reports cover different package sets (1921 → 1240 files, 60010 → 41363 lines), so the patch percentage is not a like-for-like comparison here.

Adding tests for ~389 lines of relocated renderer, CodeMirror-keymap and controller code is real work, but it is a different change from a mechanical module split and would make an already large PR unreviewable. Happy to open a follow-up issue for coverage on `ghost-renderer.tsx` and `create-controller.ts` if that is wanted.

PR #12466, opened 2026-08-04
plugin-inbox: drop a deleted Gmail message instead of wedging sync

mykola-vrmchk:
## What

A `404` from Gmail's `messages.get` escaped `fetchMessages`' stream unhandled, failing the entire sync run. Since the delta token is written only after a clean run ([mail-sync.ts:560-569](https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-inbox/src/operations/mail/mail-sync.ts#L560-L569)), a message deleted between its id being enumerated — listed, or logged in `history.list` — and the full fetch **wedged the mailbox permanently**: the cursor never advanced, so every following run re-read the same history page and died on the same id. `Cursor.skipCommitted` can't filter it either, because it was never committed.

This catches the 404 and drops that one message. Every other error still fails the run, so the durable retry re-fetches it rather than stranding it once the cursor advances.

## Why this shape

Not an invented policy — it's the rule the JMAP provider already documents at [`jmap/sync/sync-provider.ts:430`](https://github.com/dxos/dxos/blob/main/packages/plugins/plugin-inbox/src/operations/mail/jmap/sync/sync-provider.ts#L430):

> Drop an id deleted between query and `emailGet` (returns nothing) by filtering out the null. Do NOT recover the error channel: a real `JmapApiError` must propagate and fail the run…

JMAP got that behavior for free because `emailGet` returns an empty list for a missing id. Gmail signals the same condition with a `404`, and that path was never handled — so Gmail was the only provider where a deleted message was fatal. The same asymmetry exists locally: a 404 is already tolerated for `listHistory` (stale token → fresh capture) and for attachment fetches (logged and dropped); the message fetch was the one hop with no tolerance.

## Production evidence

From SigNoz (edge `main`, `operation-service`), one mailbox on `gmail`:

- **71 consecutive** `mail sync failed` runs from 12:20:16 UTC onward, **zero** successes — retrying every 5 min with no self-recovery.
- All 71 hit the **same** call: `GET .../users/me/messages/19fc8a34d65488b4` → `404 Requested entity was not found.`
- Preceded by a ~62h sync outage for that space, which built the history backlog that came to contain a since-deleted message.

Once this deploys the wedged mailbox recovers on its own — the next run drops that id, the backlog drains, and the token advances past it.

## Tests

Two tests, both written first and confirmed to fail with the exact production signature (`MailSyncError` ← `GoogleApiError 404` at `fetch.ts:88`):

1. `a delta naming a deleted message drops it, syncs the rest, and advances the token` — the incident path. Asserts the live arrival syncs, the deleted id doesn't, and **the token advances `1000` → `1001`**; that last assertion is what proves the wedge is gone.
2. `a window-scan id deleted between the list and the fetch is dropped, not fatal` — the non-incremental path into the same line, for the list→get race during backfill.

Plus a `withDeletedMessages` layer wrapper, alongside the existing `withFaultAfterMessages` / `countingGmailApi` helpers.

```
Test Files  28 passed | 7 skipped (35)
     Tests  234 passed | 14 skipped (248)
```

## Reviewer notes

- One wasted `getMessage` per dropped id per run — a never-committed id can't be filtered by `skipCommitted`. Bounded by `maxItemsPerRun`, same trade the attachment-drop path already makes.
- The dropped message is `log.warn`'d with its id, so a mailbox silently losing messages this way stays visible in SigNoz.
- A separate `401` cluster on the same mailbox (07-30 → 07-31, 366 errors) is **not** addressed here — different cause, self-resolved.

## Summary by CodeRabbit

- **Bug Fixes**
  - Gmail synchronization now skips messages deleted before they can be fetched.
  - Sync runs continue successfully when Gmail returns a “not found” response.
  - Remaining messages are processed normally, and synchronization history continues to advance.
  - Prevents deleted messages from interrupting both initial scans and incremental updates.
- **Release**
  - Included as a patch update for the inbox plugin.

### DIFF
--- a/.changeset/gmail-sync-deleted-message.md
+++ b/.changeset/gmail-sync-deleted-message.md
@@ -1,0 +1,6 @@
+---
+'@dxos/plugin-inbox': patch
+---
+
+Fix Gmail sync wedging permanently when a message is deleted before it is fetched: the 404 now drops that
+message instead of failing the run and stranding the history token.
--- a/packages/plugins/plugin-inbox/src/operations/mail/google/sync/fetch.ts
+++ b/packages/plugins/plugin-inbox/src/operations/mail/google/sync/fetch.ts
@@ -91,11 +91,9 @@ export const fetchMessages = (
           config.onRetrieved?.();
           return Option.some(message);
         }).pipe(
-          // A message deleted between its id being enumerated (listed, or logged in `history.list`) and
-          // this fetch 404s permanently, so failing the run wedges the mailbox: the delta token only
-          // advances on a clean run, leaving every following run to re-read the same history page and die
-          // on the same id. Drop just that message; any other error still fails the run so the durable
-          // retry re-fetches it rather than stranding it once the cursor advances.
+          // A message deleted before this fetch 404s forever and the token only advances on a clean run,
+          // so failing here wedges the mailbox on that one id. Any other error must still fail, so the
+          // durable retry re-fetches it rather than stranding it once the cursor advances.
           Effect.catchIf(
             (error) => error instanceof GoogleApiError && error.code === 404,
             (error) => {
--- a/packages/plugins/plugin-inbox/src/operations/mail/google/sync/fetch.ts
+++ b/packages/plugins/plugin-inbox/src/operations/mail/google/sync/fetch.ts
@@ -14,6 +14,7 @@ import { log } from '@dxos/log';
 import { EmailStage } from '@dxos/pipeline-email';
 
 import { GoogleMail } from '../../../../apis';
+import { GoogleApiError } from '../../../../errors';
 import { GoogleMailApi, type GoogleMailApiError, type GoogleMailApiService } from '../../../../services';
 import { type SyncStreamConfig } from '../../../../types';
 import { type AttachmentMetadata } from '../mapper';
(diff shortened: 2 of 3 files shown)


mykola-vrmchk:
Thanks — three of the four applied in 3042626f96, one declined.

**Applied**

- `fetch.ts` — compressed to the two load-bearing clauses (why a 404 is dropped, why nothing else is). The full mechanism belongs in the PR description, not the source. Agreed this was over the repo's one-clause rule.
- `sync.test.ts` window-scan comment — removed; the test name does state the scenario.
- `sync.test.ts` delta comment — kept but halved. Not removed: the assertion is `toBe('1001')`, which shows *that* the token advances but not why that is the assertion that matters. The remaining clause is the transient-vs-permanent contrast with the crash test directly above it, which is the actual bug and isn't recoverable from the test name.

**Declined — `withDeletedMessages` JSDoc + placement**

The file has two sibling injection helpers, `withFaultAfterMessages` and `countingGmailApi`, both defined before the `describe` block and both carrying JSDoc that explains what they inject. Moving this one after the suite would separate it from its siblings, and stripping its JSDoc while the neighbors keep theirs would make the group inconsistent. The docstring also carries non-obvious domain info the name can't — that Gmail answers a deleted message with a 404, which is the entire premise of the fix.

Happy to revisit if the convention is meant to change, but that's a change to all three helpers, not this one.

PR #12473, opened 2026-08-05
compute: add incomplete outcome for RunAgainError in trace graph

dmaretskyi:
## Problem

`RunAgainError` ("Run again") is raised when an operation yields to the scheduler via `Operation.runAgain()` — it has more work to do and will be re-invoked with the same input. It is **not** a failure. Yet the operation process caught it like any other defect and emitted an `operation.end` trace event with `outcome: "failure"`, so periodic syncs (e.g. Google Mail sync) showed up as **hard errors** in the trace graph on every partial run:

```json
{ "type": "operation.end", "data": { "name": "Sync Google Mail", "outcome": "failure", "error": "Run again" } }
```

## Change

Introduce a third operation outcome, `incomplete` — neither success nor failure, an unfinished execution that will be run again.

- **`@dxos/compute` (`Trace.ts`)** — add `incomplete` to the `OperationOutcome` type and to the `OperationEnd` schema literal.
- **`@dxos/compute` (`Process.ts`)** — in the operation's `catchAllDefect`, detect `RunAgainError.is(defect)` and emit `outcome: "incomplete"` instead of `"failure"`. Genuine failures are unchanged.
- **`@dxos/plugin-assistant` (execution graph)** — present `incomplete` as a distinct **warn-level** state (`ph--arrows-clockwise--regular`, ` - Incomplete` suffix) rather than folding it into the error path. SPEC table updated.
- **`@dxos/plugin-routine` (routine trace companion)** — add an `incomplete` `RunStatus` (precedence: failure > incomplete > success) with its own icon/color.

## Tests

- `ProcessManager.test.ts` — a `runAgain` operation now records a single `operation.end` with `outcome: "incomplete"` (verified via a capturing trace sink).
- `execution-graph.test.ts` — an `incomplete` outcome renders as a warn-level ` - Incomplete` commit, not an error.
- `runs.test.ts` — a run with an `incomplete` `OperationEnd` reports status `incomplete`; `failure` still beats `incomplete`.

All three suites pass locally (compute-runtime 158, plugin-assistant execution-graph 49, plugin-routine runs 10).

## Changeset

`.changeset/operation-incomplete-outcome.md` — `@dxos/compute: minor`.

https://claude.ai/code/session_01BNWqiQ9ujmSXSCG6FEyDzT

---
_Generated by [Claude Code](https://claude.ai/code/session_01BNWqiQ9ujmSXSCG6FEyDzT)_

## Summary by CodeRabbit

* **Improvements**
  * Operations that pause and require another run are now shown as **Incomplete** instead of failed.
  * Execution graphs and routine traces display incomplete runs with warning styling and refresh indicators.
  * Failures continue to take precedence when a run includes both incomplete and failed operations.
  * Operation details now include stable error codes to improve failure identification.

* **Bug Fixes**
  * Improved status reporting for scheduler-triggered reruns, including legacy events, preventing them from being incorrectly presented as errors.


packages/core/compute/compute/src/Trace.ts
```diff
@@ -458,8 +458,12 @@ export const testTraceService = (
 
 /**
  * Outcome of an operation invocation.
+ *
+ * `incomplete` is neither success nor failure: the operation yielded to the scheduler
+ * (via {@link Operation.runAgain}) with work still to do and will be re-invoked. See
+ * {@link RunAgainError}.
  */
-export type OperationOutcome = 'success' | 'failure';
+export type OperationOutcome = 'success' | 'failure' | 'incomplete';
```
dmaretskyi: instead of this just do detection in ui, so the machinery stays the same

dmaretskyi: Done in e2eb597 — reverted the `Trace.ts`/`Process.ts` machinery (outcome stays `success`/`failure`, `RunAgainError` is still recorded as a failure carrying its message). The `incomplete` state is now derived purely in the UI: both the execution graph and the routine run list detect `outcome === 'failure' && error === RUN_AGAIN_MESSAGE` and present it as a warn-level "Incomplete". I exported `RUN_AGAIN_MESSAGE` from `@dxos/compute` (alongside `RunAgainError`) so the detection string stays in sync with the error definition rather than being hard-coded in the UI.

---

packages/core/compute/compute/src/errors.ts
```diff
@@ -4,13 +4,20 @@
 
 import { BaseError, type BaseErrorOptions } from '@dxos/errors';
 
+/**
+ * Message carried by {@link RunAgainError}. Exported so trace consumers can recognize a run-again
+ * yield from a persisted `operation.end` event (which serializes only the error message, not the
```
dmaretskyi: can you update the operation.env event to include error code and match over that?

dmaretskyi: Done in e7e43063. Added an `errorCode` field to the `operation.end` event, populated from the failing error's stable name (`defect.name`). The execution graph and routine run list now match `outcome === 'failure' && errorCode === RUN_AGAIN_ERROR_CODE` (exported from `@dxos/compute`) instead of the message string. `RUN_AGAIN_ERROR_CODE` is the same constant fed into `RunAgainError`'s definition, so the code and the matcher can't drift.

---

packages/core/compute/compute/src/Trace.ts
```diff
@@ -491,6 +491,11 @@ export const OperationEnd = EventType('operation.end', {
     outcome: Schema.Literal('success', 'failure'),
     /** Error message if the operation failed. */
     error: Schema.optional(Schema.String),
+    /**
+     * Stable code (error name) of the failure, when the operation failed with a typed error.
+     * Lets consumers match on the error kind (e.g. a run-again yield) without parsing the message.
+     */
+    errorCode: Schema.optional(Schema.String),
```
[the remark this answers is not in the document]
dmaretskyi: Added the fallback in 6e8e429a. Both consumers (execution graph and routine run list) now treat a failure as run-again when `errorCode === RUN_AGAIN_ERROR_CODE` **or**, for events persisted before `errorCode` existed, when `errorCode === undefined && error === RUN_AGAIN_MESSAGE`. `RUN_AGAIN_MESSAGE` is re-exported from `@dxos/compute` and is the same constant fed into `RunAgainError`'s definition. Added tests covering the legacy message-only path in both `execution-graph.test.ts` and `runs.test.ts`.

---

PR #12482, opened 2026-08-05
e2e: bundle once, then shard by browser x moon --job

wittjosiah:
Closes [DX-1116](https://linear.app/dxos/issue/DX-1116/speed-up-e2e-test-suite-in-ci-via-sharding)

Stacked on #12494 (base `claude/depot-vs-self-hosted-cache-3fbd62`).

## What

The `e2e` job in **Check** was one runner doing everything: it bundled the apps itself, then ran `moon exec :e2e` with `PLAYWRIGHT_BROWSER: all` — all three browsers, all eight suites, serially. It measured 43m+ and brushed its step timeout.

It becomes two jobs:

- **`e2e-bundle`** builds the web bundles once and publishes them to the moon remote cache.
- **`e2e`** is 9 cells: browser (`chromium` / `firefox` / `webkit`, via `PLAYWRIGHT_BROWSER`) × three cells of `moon exec --on-failure continue :e2e-ci plugin-script:e2e --job N --job-total 3`. Each cell hydrates the bundle rather than building it.

moon packs whichever targets a cell is assigned, so there is no target list to maintain — a new `e2e`-tagged suite joins the pool by existing.

## Bundling once

Every runner previously bundled the whole graph itself, and concurrent starts meant none could hit the others' cache writes. The bundle is now built once and shared, but only after two hash mismatches were fixed:

- `bundle-e2e` exists as its own target on composer-app and todomvc rather than the e2e env being overridden on `e2e`'s `bundle` dependency. A dependency-level `env` map is part of that dependency's hash, so `composer-app:bundle` and the bundle the e2e closure actually waited on were two different hashes — the warming job populated an entry no test runner ever read, and each rebuilt the bundle itself (measured 19s each).
- `e2e-bundle` builds only composer-app and todomvc. `:bundle` built all eleven apps, so every runner also hydrated the docs, cli, storybook, testbench, crx, devtools and examples closures — docs' typedoc artifacts cost more to download than composer's bundle costs to build.

Steady state is now `Bundle All` at 66–90s, with cells reporting `bundle-e2e (cached from remote, 7s 931ms)` on an identical hash across browsers.

## Sharding shape

The browser is a GitHub matrix axis rather than a moon task dimension, so it cannot multiply with the shard dimension in the task namespace: a suite that outgrows one cell needs two task definitions, not six.

`--job` partitions targets and can never split one, which made composer-app the floor of the whole phase — 3m28s on chromium against ~160s of fixed per-cell overhead, with no other target above 1m9s. It opts out of the inherited `e2e-ci` (`workspace.inheritedTasks.exclude`) and supplies `e2e-ci-1of2` / `e2e-ci-2of2` using Playwright's `--shard`. `composer-app:e2e` is untouched, so the local command and the docs naming it still work.

Two halves, not more: another cell costs ~160s of setup plus an independent chance of drawing a stalled checkout (2 of 27 cells lost >18m to one while measuring this), so a half at ~104s already sits under the overhead splitting it again would buy.

Three cells per browser for the same reason — ~160s overhead against ~450–590s of test work per browser.

## Signal quality

Changes so a green cell means something:

- **`quarantine: false`** on both uploader steps. With quarantine on, cells reported success while a test failed (`All test failures were quarantined, overriding exit code to be exit_success`), which makes a green cell unfalsifiable. Analytics upload is kept; masking is not. Unstable tests get `test.fixme` in the spec, where the skip is visible in review.
- **`--on-failure continue`** on both steps, not just `Test All`. moon's default bail dropped a cell's two heaviest targets after an earlier failure and made that cell look 3× cheaper than it was.
- **`retries: 0`** in `e2ePreset` (was 2 on CI). Retrying hid flakes behind a 3× time cost — three `inbox.spec.ts` tests alone spent ~9 minutes of a run failing three times each — which made timings useless for sizing the suite.
- `workers: 2` on CI (was 1), overridable with `PLAYWRIGHT_WORKERS`. 4 was measured 2.6× faster than 1 but starved shared setup on suites with enough live tests to contend; 2 keeps the parallelism without the false failures.

## Storybook-backed suites

Six suites (`lit-grid`, `plugin-kanban`, `plugin-sheet`, `react-ui-mosaic`, `react-ui-table`, and their kin) serve stories from `storybook dev`, which resolves packages from source but still needs generated artifacts. Their `e2e` tasks now declare `#vendor:build` and `#storybook:prebuild` — without vendor dist, Vite's dep-scan aborts and takes the web server with it.

`lit-grid` additionally runs one worker. `webServer.port` only waits for the port to listen while `storybook dev` compiles the story on first request, so two workers both requested it mid-`optimizeDeps` and one hung past the 30s action timeout in `grid.ready()` — one of three tests died that way in 5 of 6 firefox/webkit cells while the other two passed in ~3s. Its three tests total ~7s, so the compile is cheaper paid once.

## Deferred tests

`test.fixme` with a TODO naming the run that caught each: `react-ui-mosaic` `rearrange columns` / `rearrange within column`; `composer-app` `collections › delete a collection`, four `comments` tests, `startup › warm-cold start`; `todomvc` `edit a task`, `toggle all tasks & clear completed`.

Deliberately **not** deferred: five composer chromium tests that failed while measuring (`basic › create space`, two `collaboration`, two `collections`, `comments › create`). Their evidence is confounded with the concurrency this PR changes — four came from the single most contended cell — and skipping them would drop composer's chromium coverage from 14 running tests to 8. They get verdicts from the verification run.

## Alternatives measured and rejected

Two other sharding strategies were built out on this branch and run head-to-head against the shipped one in a single 27-cell run (9 cells each, same commit). Neither survives in the diff; recorded here so the choice isn't re-litigated from scratch.

| | critical path | runner-time | targets covered | test failures |
| :-- | --: | --: | --: | --: |
| Knapsack Pro queue mode + per-browser split | 297s | 1618s | 27/27 | 5 (only 2 surfaced) |
| Per-browser moon task variants, 9 cells | ~335s corrected | ~1470s | 23/25 | 2 |
| **Browser × `--job`, 9 cells (shipped)** | 364s | 1500s | 27/27 | 6 |

All three landed within ~10% on runner cost and 297–364s on critical path, so speed did not decide it:

1. **Knapsack Pro** ([`@knapsack-pro/playwright`](https://knapsackpro.com), a file-level queue ordered by recorded duration) needs an external service and a token, and its measured advantage came entirely from offloading composer — the part `--shard` now handles in-repo. Two of its cells also reported success while a test failed, via the quarantine override described above.
2. **Per-browser moon task variants** (`e2e-chromium`/`-firefox`/`-webkit`) put all 24 browser targets in one flat pool, which can lend work across browsers — a real advantage, since chromium's pool is ~30% heavier than the others. But the browser then multiplies with the shard dimension in the task namespace, and splitting composer fixes the same imbalance more directly. Its numbers above are corrected because moon's default bail silently dropped `composer-app:e2e-chromium` and `plugin-sheet:e2e-chromium` from a failing cell, making the arm look cheapest when it had simply skipped the two most expensive targets — which is what motivated `--on-failure continue`.

## Verified

- Playwright 1.57 shards per **test**, not per file, wherever `fullyParallel` is set and no `beforeAll` welds a group together (`testGroups.js:createTestGroups`). The only two composer specs using `beforeAll` are `testIgnore`d by its config.
- composer's halves are an exact disjoint partition: 18 + 18 = 36, union identical to the unsharded list, zero overlap.
- `args` appends under `extends` and is part of the task hash (three distinct hashes for base / `--a` / `--b`), so the halves cannot replay each other's cache entry.
- The pool resolves to exactly 9 targets per browser, with composer's base `e2e` correctly absent.

Not yet verified: cell timings. The 18/18 split is by *listed* test count and Playwright shards by count rather than duration, so the halves may be lopsided in runtime — the verification run will show it. That run is pending the GitHub Actions outage.

## Note for landing

Cell names change from `e2e (chromium)` to `e2e (chromium, 0)`. Any branch-protection required check naming the old form will never appear again.

https://claude.ai/code/session_01WwD57XoixmFug2wWT2ANbx

## Summary by CodeRabbit

- **New Features**
  - First-run onboarding now opens directly to the default space’s Home.
  - Editors remain read-only while referenced content loads.
  - Comment threads offer more reliable selection, editing, deletion, and navigation.
- **Bug Fixes**
  - Improved drag-and-drop behavior in Kanban and Mosaic views, including scrolling.
  - Space creation succeeds even when optional replication settings fail.
  - Invitation acceptance and authentication handle identity-reset states more reliably.
  - Improved RPC error handling and state update consistency.
- **Documentation**
  - Expanded CI, end-to-end testing, and cloud-sandbox guidance.


wittjosiah:
CI note: `Basic tests › create identity, space is created by default` is red on this PR's e2e cells (all three browsers) after syncing with main — and it is failing identically on main itself: run [31261684165](https://github.com/dxos/dxos/actions/runs/31261684165) on `da37a13e` fails it on chromium, firefox, and webkit, 3 retries each.

Cause (from a local trace on the merged head): first run now lands on the space **Home** plank; the README exists but only as a card in Home, and `SpaceHomeWelcome` — which should supply the `Welcome to Composer` heading there — renders nothing. Verified not caused by this branch's echo change (reproduces with the reactivity change reverted, and on main which doesn't carry it).

Will re-sync and re-run e2e when main recovers. The storybook job failure is likewise on main (`plugin-search` story flakes, one unquarantined).

---

wittjosiah:
Update on the note above: rather than waiting on main, `9c9bac2d` fixes the first-run regression on this branch — first run now lands on the README it seeds (navigation moved into plugin-onboarding; plugin-support's on-create-space handler removed, since the Home welcome panel can't resolve the default space during identity bootstrap). `create identity` verified locally at 5/5 repeats plus a full `basic.spec.ts` pass on chromium; e2e dispatch on the new head is running. The fix reaches main when this PR lands.

---

PR #12490, opened 2026-08-06
feed: add optional at-rest encryption via pluggable Cypher

dmaretskyi:
## Summary

Adds optional at-rest encryption for feed blocks (the non-hypercore `@dxos/feed` store), as the dxos-side half of the KMS data-feed-encryption project. `FeedStore` gains a pluggable `Cypher` seam; **without one, blocks are stored as plaintext (no encryption by default)**.

The `Cypher` decides *per feed* whether to encrypt (from feed metadata), and provides encrypt/decrypt. Encryption happens purely at the SQLite boundary — `append` seals before INSERT, `query` opens after SELECT — so everything above the store sees plaintext. The export path (`getAllFeedsForSpace`) stays **byte-faithful**: it preserves ciphertext + envelope so a re-import reconstitutes the exact rows and no plaintext leaks into an archive.

## Changes

- **`@dxos/protocols`** — `Block` gains optional `dekId` + `iv` envelope fields (`data` becomes ciphertext when sealed).
- **`@dxos/feed`**
  - `cypher.ts` — `Cypher` interface (`shouldEncrypt` / `encrypt` / `decrypt`), `EncryptedPayload`, `FeedMetadata`, `CypherContext`, and a typed `CypherError`. Promise-based so an implementation stays a self-contained WebCrypto module.
  - `web-crypto-cypher.ts` — reference `WebCryptoCypher` over AES-256-GCM, binding the block's natural key as AAD. Parameterized by a `CypherKeyProvider`, so the same crypto body runs in the browser, Node, and Cloudflare Workers — the edge reuses it by supplying a KMS-DO-backed provider. Ships `createInMemoryKeyProvider` for tests/single-process.
  - `migrations/0002_block_encryption.sql` — adds nullable `dekId` / `iv` columns to `blocks` (existing rows and plaintext stores unaffected).
  - `FeedStore` — optional `cypher` option; seal-on-append, open-on-query, byte-faithful export.

## Reusability in edge

The `Cypher`/`WebCryptoCypher` is deliberately framework-free and WebCrypto-based, so the edge KMS work can back it with the KMS DO's DEKs (via a `CypherKeyProvider`) rather than reimplementing crypto. That edge-side key registry is a separate change.

## Testing

- `moon run feed:test` → **36 passed** (8 `WebCryptoCypher` unit tests, 4 `FeedStore` encryption tests, 24 pre-existing).
  - Round-trip, random-IV distinctness, tamper rejection, AAD/block-id rejection, unknown-DEK rejection, `shouldEncrypt` predicate.
  - White-box: encrypted feeds store ciphertext + envelope; plaintext feeds store plaintext with null envelope; per-feed selection.
- `moon run feed:build` + downstream `echo-host:build` green (verified the `CypherError` error-channel widening ripples cleanly).
- `pnpm format` + `feed:lint` / `protocols:lint` clean.

Changeset: `@dxos/feed` minor, `@dxos/protocols` minor.

---
_Generated by [Claude Code](https://claude.ai/code/session_0131BaLeAnXi7fBdrJn6SFGg)_

---
<a href="https://app.blacksmith.sh/dxos/codesmith/dxos/pr/12490"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788596362&installation_model_id=431117&pr_number=12490&repository=dxos%2Fdxos&return_to=https%3A%2F%2Fgithub.com%2Fdxos%2Fdxos%2Fpull%2F12490&signature=49984626740ebe10ebfba3510af2df9e94d709becccf6f8c2d795f16a72c4dec"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added optional AES-256-GCM encryption for feed blocks at rest.
  * Added configurable per-feed encryption and key management.
  * Added support for browser, Node.js, and Cloudflare Workers environments.
  * Preserved plaintext storage when encryption is not configured.
  * Preserved encryption metadata during feed archive export.
* **Bug Fixes**
  * Added tamper detection and block identity validation during decryption.
  * Added clear error reporting for encryption and decryption failures.
* **Documentation**
  * Documented encrypted block payloads and associated metadata in feed protocols.


packages/core/echo/feed/src/migrations/0002_block_encryption.sql
```diff
@@ -0,0 +1,13 @@
+--
+-- Adds the at-rest encryption envelope to blocks: `dekId` names the data-encryption key generation
+-- that sealed `data`, and `iv` is the GCM nonce used. Both are NULL on plaintext blocks, so existing
+-- rows and unencrypted stores are unaffected.
+--
+-- Not idempotent, by design: `feed_migrations` guarantees it runs exactly once. Immutable once
+-- shipped — change the schema by adding the next numbered migration.
+--
+-- AlterTable
+ALTER TABLE "blocks" ADD COLUMN "dekId" TEXT;
```
dmaretskyi: rename to "encryptionKeyId"

dmaretskyi:
Addressed the review in `47ed0c2f`:

- **Rename `dekId` → `encryptionKeyId`** (@dmaretskyi) — applied across the Block schema/field, the `blocks` column, the `Cypher` envelope, and the key-provider API, so the feed/protocol layer stays free of KMS "DEK" jargon.
- **Require both envelope fields together** — `#openBlock` now **fails closed** when exactly one of `encryptionKeyId`/`iv` is present, rather than treating a partial envelope as plaintext. I did **not** add a SQLite `CHECK ((encryptionKeyId IS NULL) = (iv IS NULL))` constraint: `ALTER TABLE … ADD COLUMN` can't attach a table-level cross-column CHECK without a full table rebuild, which is disproportionate here since the write path only ever sets both or neither and the read guard is the effective enforcement point. Happy to add the rebuild migration if you'd prefer the DB-level guarantee.
- **Changeset** — reduced to one representative of the fixed group (`@dxos/feed`).
- **`keyProvider` doc** — corrected (it's required; there is no in-memory default).
- **Tests** — merged the encryption cases into `feed-store.test.ts` and switched the promise-based `WebCryptoCypher` cases to `test()`.

The earlier `test (24.11.1)` red was an unrelated pre-existing flake (`echo-client-e2e › feed.test.ts › "last flush observed wins wholesale"`, a two-client flush-ordering race already in Trunk's flaky registry); this change is byte-identical on the no-cypher path, so it can't affect it. The new push re-runs CI fresh.

---

packages/core/echo/feed/src/web-crypto-cypher.ts
```diff
@@ -0,0 +1,92 @@
...
+    },
+  };
+};
+
+/**
+ * A one-generation in-memory {@link CypherKeyProvider}, for tests and single-process use.
+ *
+ * The DEK never leaves the process; production callers back the provider with the KMS DO instead.
+ */
+export const createInMemoryKeyProvider = async (dekId = 'dek-in-memory'): Promise<CypherKeyProvider> => {
```
dmaretskyi: dekId = public key

dmaretskyi: Kept the field as `string` per your follow-up, with the value being the public key rather than the type.

`encryptionKeyId` is now documented as the hex-encoded public key naming the sealing key — on the `Cypher` envelope, the `Block` schema, the `blocks.encryptionKeyId` column, and both `CypherKeyProvider` methods. `createInMemoryKeyProvider` defaults to 32 random bytes as hex instead of the `'in-memory-key'` label, so tests exercise the same identifier format the KMS hands out; the id stays overridable. Generated inline from `crypto.getRandomValues` rather than `PublicKey.random()` to keep the module dependency-free for the Workers build.

`moon run feed:test`: 37 passed. Lint clean on `feed` and `protocols`.

dmaretskyi:
Correction on the CI red — it was **this PR**, not a pre-existing flake. I was wrong earlier.

Root cause: sealing on append and opening on query went through `Effect.forEach` **even when no cypher is configured**, adding async turns to the storage hot path. Those extra scheduling turns reliably flipped the `echo-client-e2e` "last flush observed wins wholesale" race.

Confirmed by counterfactual, locally:
- Pre-fix (`04265b26`): the test **failed 2/2**.
- With the fix (`f70c5c47`): **11/11 pass**; full `feed.test.ts` e2e suite 286 pass / 1 expected-fail / 7 skipped.

Fix (`f70c5c47`): take the original synchronous `map` when no cypher is set, so "no encryption by default" is zero-overhead and byte-for-byte timing-identical to the pre-encryption path. Encryption paths are unchanged. `feed` unit tests still 37/37; format + lint clean.

So no quarantine needed — the test isn't the problem, my earlier commit was.

---

PR #12493, opened 2026-08-06
model-fixture: enable id sanitization by default + invariance fuzz

dmaretskyi:
## Why

The model-fixture matcher has two layers of defence against run-varying ids: (1) deterministic id generation (`EntityId.dangerouslyDisableRandomness()`) and (2) fixture-level **sanitization** (canonicalize ids for matching). Layer 2 was **opt-in** — only the `assistant-e2e` harness passed patterns, so the skill/harness fixtures ran on layer 1 alone.

Layer 1 is only deterministic while the surrounding **allocation order** is unchanged. When #12415 (demand-driven activation) reordered plugin/operation activation, the seeded ids drifted (`…CNV5A` → `…CNTT6`), and with no sanitization fallback every multi-turn database fixture missed (the prompts feed prior tool-result DXNs back in). #12468's `card-content` → `cardContent` annotation rename drifted them further. Result: the `model-fixture` workflow went red on `main` for the database suite.

## What

1. **Sanitization on by default** — `make()` falls back to `DEFAULT_DYNAMIC_VALUE_PATTERNS` (space id, entity id, UUID, ISO timestamp) when none are given; pass `dynamicValuePatterns: []` to opt out. Normalization — not id determinism — is now the load-bearing defence. Removes the `assistant-e2e` harness's now-redundant hardcoded pattern list.
2. **Regenerated every model-fixture suite** (`ai`, `agent-runtime`, `assistant-toolkit`) against current `main` with sanitization on — all replay **offline with zero misses** (ai 5 ✓, agent-runtime 18 ✓, assistant-toolkit 23 ✓).
3. **Match-key invariance fuzz** (`match-invariance.test.ts`, `manual`-tagged so it never runs in CI): asserts the normalized key is invariant to same-shape id churn and that distinct ids never collapse — validating the sanitizer independently of deterministic id generation. Covers the exact DXN form that broke, plus a real-corpus pass over every committed fixture. It passes today (the field failure was sanitization being *off*, not the patterns being wrong).

## Testing

- `DX_RUN_MODEL_FIXTURE_TESTS=1` replay: green across all 7 suites, 0 misses.
- `DX_RUN_MANUAL_TESTS=1` fuzz: `match-invariance.test.ts` 4/4 passing (synthetic + real corpus).
- `ai` unit tests + affected-package builds green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01H3ATd4wAFWUhaqyX9q1f6Z)_

---
<a href="https://app.blacksmith.sh/dxos/codesmith/dxos/pr/12493"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788609913&installation_model_id=431117&pr_number=12493&repository=dxos%2Fdxos&return_to=https%3A%2F%2Fgithub.com%2Fdxos%2Fdxos%2Fpull%2F12493&signature=27334a7dab018dc04c9a91447664d5932ae41c9812c81f4356462d5313e1baf7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Fixture matching now automatically handles changing space IDs, entity IDs, UUIDs, and timestamps.
  * Custom matching patterns remain supported, with an option to disable default normalization.
* **Bug Fixes**
  * Improved consistency when replaying streamed AI interactions whose generated identifiers or timestamps differ.
* **Tests**
  * Added broader validation for dynamic identifier replacement, repeated values, boundary cases, and real stored conversation data.
  * Refreshed conversation scenarios covering research, database, memory, calculator, and other assistant interactions.


dmaretskyi:
Addressed the review in `cb22a5376a`:

**Applied**
- **Descriptive names** — renamed the mulberry32 mix variable `t`→`mixed`; the `a`/`b` locals are gone with the test rewrite below.
- **Negative-boundary test** — replaced the positive space-id case with your suggested embedded-token assertion (`A${\`B${'A'.repeat(32)}\`}A` normalizes unchanged), verifying the `SPACE_ID_PATTERN` boundary guard rejects a space-shaped run inside a longer token. Passes.

**Skipped, with reasons**
- **Import order (builtin-first)** — oxfmt (the CI formatter) sorts `@effect/vitest` before `node:*`; forcing builtin-first fails `format-check`. The formatter governs this.
- **Rename to `LanguageModelFixture.test.ts`** — that filename already exists in the same directory (the engine's own suite); two files can't share it. `match-invariance.test.ts` is a deliberately separate concern.
- **`24a3b86…` "6 schemas"→"7"** and **`7940dcd…` `in`-shape / finish event** — these are faithful recordings of the model's actual streamed output, not hand-authored data; editing them would falsify the fixtures. The "6 schemas" is the model's own miscount, and the `{"/": …}` `in` value is the model copying the `query` tool description's example — that example is a real bug in `assistant-toolkit/.../database/operations/definitions.ts`, tracked as a separate fix (correcting the description changes the hashed tool parameters and must land with a coordinated regen). Both suites replay green as recorded, so no missing finish event blocks replay.

---

PR #12503, opened 2026-08-07
space: move app config to a dedicated settings space

wittjosiah:
App configuration moves out of the personal space and into a hidden, membership-locked **settings space**. The personal space becomes an ordinary space — the **default space** — designated by a setting rather than by a tag.

## Audit — what depended on the personal space

| Category | Sites |
| --- | --- |
| **Definition** | `AppSpace.ts`: `PERSONAL_SPACE_TAG`, `isPersonalSpace`, `getPersonalSpace`, `setPersonalSpace` (legacy `__DEFAULT__` property), `resolvePersonalSpace` (legacy `DefaultSpace` credential) |
| **Creation** | `plugin-space/capabilities/identity-created.ts`, `plugin-client` `halo create` + `initializeIdentity`, `stories-assistant` decorators |
| **Data stored in it** | the `shared-spaces` order Expando (written by `spaces-ready`, read by `plugin-space` and `plugin-native-filesystem` graph builders); `WelcomeDismissedAnnotation` on space properties (4 sites in `plugin-support`) |
| **Default-target fallback** | 18 `getPersonalSpace()` call sites — export/share/settings actions, create-object dialog, assistant chat + graph, preview popover, transcription entity lookup, tasks quick entry, native-filesystem, onboarding OAuth registration, debug diagnostics, CLI |
| **UI special-casing** | translated "Personal Space" label, navtree visibility allowlist, hidden Members panel, hidden name/icon fields, hidden Danger Zone, welcome carousel gating |
| **Shareability** | `MembershipPolicy` is fixed at genesis and exposed as `space.membershipPolicy`, but every UI gate keyed off the *tag* instead |

## Changes

**Settings space** — `AppSpace.SETTINGS_SPACE_TAG` (`org.dxos.space.settings`), `isSettingsSpace`, `getSettingsSpace`. Created locked at genesis and EDGE-replicated. New `SpaceCapabilities.SettingsSpace`. It hides itself: `AppSpace.isVisibleSpace` treats any tagged space as internal, except the exemplar space and the legacy personal tag.

**Default space** — `AppAnnotation.DefaultSpaceAnnotation` stores a space id on the settings space properties; `getDefaultSpace` resolves it, falling back to the legacy tag → `__DEFAULT__` → credential chain. `readDefaultSpaceId` / `setDefaultSpaceId` are the paired accessors and require an open settings space. Settings → Your spaces gains a picker.

**One bootstrap** — `AppSpace.setupIdentitySpaces` is the single definition of a new profile (settings space + default space, both locked and EDGE-replicated). Used by `identity-created`, the `halo create` CLI command and the story/test harnesses. It lives in app-toolkit because plugin-client cannot depend on plugin-space.

**Migration** — `plugin-space/src/migrations/settings-space.ts`, run from `spaces-ready`: creates the settings space, transfers the ordering, designates the legacy space, and stamps `properties.name = 'Personal'` when unnamed. Idempotent, and re-runs when a legacy space is discovered *after* the settings space (the space list fills incrementally, and the `__DEFAULT__` marker is only readable once the space opens).

**Deprecation** — `PERSONAL_SPACE_TAG` is `@deprecated` and no longer written. `isPersonalSpace` / `resolvePersonalSpace` → `isLegacyDefaultSpace` / `resolveLegacyDefaultSpace`; `setPersonalSpace` is deleted, since the designation supersedes the marker.

**De-specialization** — the `personal-space.label` translation and the `personal` option on `getSpaceDisplayName` are gone; name and icon are editable and the space is deletable, except that the space currently designated as default hides its Danger Zone. The dead `defaultSpaceId` prop on `CreateObjectPanel` is removed. `plugin-space` root modules collect into `util/`.

**Private toggle** — `SpaceForm.private` sits ahead of `edgeReplication`, defaults to `false`, and `SpaceOperation.Create` passes `membershipPolicy: LOCKED | INVITE`. The Members panel keys off `space.membershipPolicy === LOCKED` rather than any tag; space settings shows the flag read-only, since genesis fixes it.

## Notes for review

- **Only private spaces may be designated as the default**, because the default space holds integration credentials (the Atmosphere `AccessToken` written by OAuth registration). A `TODO` on the filter removes the restriction once that constraint is gone. The space *already* designated stays listed whatever its policy, so a profile migrated from an unlocked personal space still sees its own choice.
- `HelpOperation.HideWelcome` lost its `space` input — the dismissed flag is app-wide and lives on the settings space. It is **not** migrated, so a dismissed welcome carousel reappears once. It only ever rendered on the default space and still does.
- `identity-created` deliberately does **not** create the ordering Expando; `spaces-ready` owns it so the two cannot race.
- The first-boot workspace switch waits for the designation (`awaitDefaultSpace`). `setupIdentitySpaces` writes it *after* `client.spaces.create` has already published the settings space to the space list, so a synchronous read finds nothing — that race left the navtree on the `default` sentinel showing "You don't have this workspace" while the deck rendered the default space's Home.
- `ensureSpacesOrder` recovers only a closing space; any other write failure propagates, so the migration cannot record a designation while the ordering silently failed to persist.

## Test plan

- `moon exec :build` — 0 TypeScript errors across the workspace.
- `moon run app-toolkit:test` — 136 passed, including `AppSpace` tag/visibility predicates.
- `moon run plugin-space:test` — 46 passed, including 4 real-client migration tests: designation + ordering transfer, idempotence, recovery of a late legacy space, and the empty case.
- `plugin-support`, `plugin-client`, `cli-util` suites pass; `moon run …:lint` and `pnpm knip` clean; `pnpm format` applied.

**Worth verifying by hand on the preview**, since the first-boot race was invisible to the whole suite: create a fresh identity and confirm it lands on the Personal workspace with the sidebar populated. Also worth eyeballing the private toggle in the create-space dialog and the default-space picker under Settings → Your spaces.

**Known rough edge, not fixed here:** form descriptions don't render in the create-space dialog, so the private toggle ships there with a bare label. Its description is written to read correctly in space settings, which is the only place it renders and where it is read-only. Fixing the dialog is a design-system change affecting every dialog form.

**Deliberately out of scope** (pre-existing code this PR only *moved* into `util/`, so it shows in the diff): `Effect.orElseSucceed` swallowing `Database.load` failures in `collection-path.ts`, `random as any` in the pipeline/search stories, and helper placement in `collection-path.test.ts`.

https://claude.ai/code/session_017v13MmQbEn97Agg7GZustn

---
_Generated by [Claude Code](https://claude.ai/code/session_017v13MmQbEn97Agg7GZustn)_

## Summary by CodeRabbit

* **New Features**
  * Added a private, replicated settings space for app-wide preferences and workspace ordering.
  * Added configurable default-space selection, with visible locked spaces available as choices.
  * Added private-space creation and membership controls.
  * Personal/default spaces can now be renamed and re-iconed.
  * Welcome-message dismissal now applies across the app.
* **Bug Fixes**
  * Prevented deletion of the designated default space until another space is selected.
  * Preserved existing space ordering and default-space settings during migration.


wittjosiah:
The review failed because the head moved from `6e5d82f` to `b7435b7d` mid-pass, so the refactor commit (107 files) hasn't actually been reviewed. Re-running.

Two notes for that pass, since the earlier findings were auto-marked resolved:

- The **late-legacy-space** finding was *not* fixed by `6e5d82f` — that commit removed the retry entirely while simplifying `spaces-ready`. `b7435b7d` restores it, and also fixes the half the retry alone didn't cover: `ensureSpacesOrder` short-circuited on the empty ordering object created by a first pass that ran before the legacy space resolved, so a recovered legacy space would still have lost its ordering. Ownership of the ordering *content* now sits with the migration, covered by a regression test running the two passes in that order.
- `readDefaultSpaceId` no longer guards against a closed space — callers pass an open settings space, so `getDefaultSpace` and the migration retry check `SpaceState.SPACE_READY` themselves.

@coderabbitai review

---

wittjosiah:
Addressed in `be62a058`. Three findings deliberately skipped, with reasons:

**`util/collection-path.ts` — `Effect.orElseSucceed` swallowing `Database.load` failures.** A real issue, but pre-existing code this PR only *moved* from `src/collection-path.ts` into `src/util/`. Narrowing it to `catchTag('EntityNotFoundError')` changes runtime behaviour for collection-path resolution, which is unrelated to the settings space. Better as its own change where it can be reasoned about and tested on its merits.

**`random as any` in the pipeline/search stories.** Also pre-existing — the only lines this PR touched in those files are the `initializeIdentity` destructures. The fix belongs with `@dxos/random`'s `ValueGenerator` typing.

**Helper placement in `util/collection-path.test.ts`.** Same moved-file situation; the helpers were already positioned that way. Reordering them here is churn in a diff that's already 100+ files.

One finding is **not** a code issue but a design decision worth a human call — flagged to @wittjosiah separately: `complete-oauth-registration` writing the Atmosphere `AccessToken` to the *default* space. That placement was an explicit instruction, but the review is right that the default space is now user-repointable and may be shareable, so a repoint desynchronises the write from `plugin-registry`'s lookup, and a shared default would widen the token's audience. Not changing it unilaterally.

---

PR #12506, opened 2026-08-07
react-ui-form: surface field descriptions as a tooltip, split out PlaceholderAnnotation

wittjosiah:
Two related changes to how a schema's `description` reaches the UI.

## Before / after

`ui/react-ui-form/Form` → **DynamicFields**. Before, a field's `description` *was* the input's placeholder, so documentation sat in the ghost text (and got truncated). After, the placeholder is a real input hint and the description moves to an info affordance beside the label.

| Before | After |
| --- | --- |
|  |  |
| `Query` reads "Type to load the choices below."; `Tag` truncates to "Combobox: type to filter; your text stays sel…" | `Query` reads "Search term", `URL` reads `https://example.com/feed.xml`; each described field carries an ⓘ, hovered here on `Name` |

## 1. Description → a tooltip beside the label

- `FormFieldLabel` takes a `description` and, when present, renders the label plus a `ph--info--regular` tooltip trigger in the label row's first column. The affordance is a sibling of the label (never a child), so the label's `textContent` stays exactly the label and `getByLabelText('Name')` keeps working.
- `FormRow` forwards it in both field and action mode, but only for variants that do **not** already render the description as visible text — the `settings` variant keeps its inline description and gets no icon.
- `FormFieldHeader` → `ArrayField` and `FormFieldSetContainer`/`FormFieldSet` gained the same pass-through, so array and nested-group headers surface their own descriptions. Every field renderer reaches this via `FormRow` (`AsyncSelectField`/`AutofillField` delegate to `SelectField`/`TextField`).
- New translation key `field-description.label` names the trigger; `type='button'` is set explicitly because `Tooltip.Trigger` deliberately omits it, and the click is stopped from bubbling so the affordance can't toggle a collapsible group header.
- The tooltip opens after 200ms rather than the provider's 700ms. `Tooltip.Trigger` already accepted `delayDuration` but discarded it (a standing TODO); it now threads through `onTriggerEnter` into the open timer, so only this trigger's timing changes.

**Generated text is not documentation.** Effect writes a description for the base scalars (`Schema.String` → "a string") and for every built-in refinement (`minLength` → "a string at least 1 character(s) long"), so presenting a description as documentation surfaced validation text. `getFieldDescription` ignores exactly that set; an authored description always wins, since it replaces the generated text on the same AST node. The filter list is derived from live Effect output and the test asserts against live schemas for all 24 built-ins plus the three scalars, so a wording change in Effect fails the test instead of leaking text into the UI.

## 2. Placeholder gets its own annotation

`description` previously did double duty — field documentation *and* the input's ghost text (`placeholder = examples ?? description ?? label`) — so a field could not have both. Input hints now live on `Annotation.FormPlaceholderAnnotation` in **`@dxos/echo`**, beside the rest of the `Form*` family, so schemas below the UI layer can set them. Placeholder precedence becomes:

```
FormPlaceholderAnnotation → examples ("Example: …") → the field's title
```

The annotation is registered in `EchoAnnotations`, so it maps into the JSON-schema `annotations` namespace (`annotations.formPlaceholder`) and survives a serialize/deserialize round-trip — a stored ECHO type keeps its hint.

### Call sites migrated

**Standard formats** (`@dxos/echo` `Format/*`) — date, time, date-time, currency, integer, percent, timestamp, geo-point, email — regain the ghost text they lost when `description` stopped feeding the placeholder. Their descriptions **stay**: removing them lets Effect's generated refinement text into the serialized schema, which for `Email`/`URL` is a raw regex.

**Regexes stay out of descriptions.** `toEffectSchema` rebuilds any stored string schema with a `pattern` via `Schema.pattern(...)`, and Effect derives "a string matching the pattern ^…$" as its description — which then showed in the form and reached agents as the field's documentation. That derived text is now suppressed (the constraint still travels as `pattern`); a stored `description` is applied afterwards and still wins. `Format.URL`/`Email` descriptions became prose ("Web address.", "Email address.") rather than restatements of their titles, and `plugin-magazine`'s bare `HandleSchema` pattern gained "An atproto handle or DID."

**`@dxos/schema`** `BaseProperty` / `formatToSchema` (the view + field editor): dropped `'Property name'`, `'Property label'`, `'Property description'`, `'Name of the record type'`, `'Available choices'` — each only restated its own title, so those fields fall back to the title and no longer sprout redundant tooltips.

**Form-facing plugin schemas**, by a consistent rule — an explicit example splits (the sentence stays as documentation, the example becomes ghost text); a bare noun phrase or an instruction moves wholesale; a real sentence stays a description:

- `plugin-routine` schedule spec — cron range hints (`'0–23'`, `'e.g. 0 9 * * MON-FRI'`) → placeholder.
- `plugin-tasks` quick-entry dialog — `'Journal entry'` on an untitled field, so the hint was the input's only affordance.
- `plugin-magazine` create-subscription — handle split to `dxos.org`; publication and URL moved.
- `plugin-freeq` channel backend — server URL split to `wss://irc.freeq.at/irc`, channel to `#general`.
- `plugin-chess-com` account + create form, `plugin-studio` variant URL, `plugin-trip` operator, and the `react-ui-form` test fixture's state code — moved.

Left alone deliberately: `plugin-*/types/Settings.ts` descriptions (the `settings` variant renders them as visible text and never shows an icon), `plugin-bluesky`/`plugin-atproto` connector handles and `plugin-connector`'s token form (already pair a real description with `examples`), and ~70 descriptions on operation/tool input schemas, where the text is the model's parameter spec and no form renders it.

## Tests

- `FormRow.test.tsx` (5): affordance appears with a description and its tooltip resolves on hover; label text unchanged; no affordance without a description; a click on the affordance doesn't fire the label row's `onClick`; `default` vs `settings` variant placement.
- `FormField.test.tsx` (4): the annotation drives the input hint while the description stays on the label; fallback is the title, not the description; an explicit placeholder beats `examples`; `examples` still applies when no placeholder is set.
- `description.test.ts` (4): every generated scalar/refinement description is ignored; authored ones survive.
- `json-schema.test.ts`: a stored pattern never becomes the description, and a stored description survives alongside one.
- `@dxos/echo` 490 pass, `@dxos/schema` 48, `react-ui-form` 75, `react-ui` 9, `react-ui-table` 21, `plugin-routine` 65, `plugin-tasks` 22, `plugin-magazine` 58. The two `FormLayout.stories.test.tsx` failures are environmental — they fail identically on a clean checkout in this sandbox (the story's client provider can't reach `127.0.0.1:3000`).
- Builds, lint, `tsc --noEmit`, and `pnpm format` clean across every touched package.

`model-fixture` is red for a pre-existing reason unrelated to this diff — see the comment below.


wittjosiah:
`model-fixture` is red on this branch, but the failure predates it and I can't regenerate the cache from this environment.

**Evidence it isn't this diff:** all 15 failures are `RuntimeException: No memoized conversation found for the given prompt` in `src/skills/database/skill.test.ts`. Replaying with this branch's changes stashed — a clean checkout of the base commit — produces the same 15 failures with the same error:

```
DX_RUN_MODEL_FIXTURE_TESTS=1 moon run assistant-toolkit:test -- src/skills/database/skill.test.ts
→ Test Files 1 failed | Tests 15 failed   (identical with and without this branch)
```

The job doesn't run on `main` pushes, so this staleness wasn't caught there.

**Why I'm not fixing it here:** regeneration makes live Anthropic calls and needs `DX_ANTHROPIC_API_KEY`; this sandbox has neither the key nor the 1Password CLI to fetch it. The command, for whoever picks it up:

```
DX_UPDATE_MODEL_FIXTURES=1 moon run '#model-fixture:test'
```

**One thing to fold in when it is regenerated:** this branch changes agent-visible schema text, so the new fixtures should be recorded on top of it rather than before it — `@dxos/schema`'s `BaseProperty` descriptions were removed, `Format.Email`/`URL` descriptions were reworded, and `Format` schemas now serialize an `annotations.formPlaceholder` field. Those all reach the model through the serialized tool definitions.

---

PR #12515, opened 2026-08-08
client, plugin-space, plugin-routine: fix the boot hang from a property-less space

richburdon:
Composer hung on boot for a profile that had acquired a second, property-less settings space. It rendered the space but could not change the active nav item, with `Finding properties for a space` taking >5s, `Automerge root doc load timeout`, repeated `subduction: session lost`, and a 2s `Waiting for the edge replication to be enabled` timeout.

## Root cause

Space creation is two-phase: the host performs genesis, then the **client** writes the properties object (`SpaceList._createSpaceInternal`). An interruption in between — a reload, a closed tab, a throw — leaves a space that can never open, because `SpaceProxy._initializeDb` waited for those properties **unbounded**. That wait runs inside `_processSpaceUpdate`, which is `@synchronized`, so the space's mutex was held for the lifetime of the session. Nothing could emit `_anySpaceUpdate` again, which is why `setEdgeReplicationPreference` (which awaits it) timed out. `ensureSettingsSpace` then did an unbounded `waitUntilReady()` on that space, so `resolveDefaultSpace` never returned and `SwitchWorkspace` never fired.

The duplicate itself came from #12503 putting a space creation on the boot path of every existing profile: two clients (a second tab, a reload racing the first) can both observe no settings space and both create one, and nothing enforced "exactly one".

## Changes

**`@dxos/client`** — `_initializeDb`'s properties wait is bounded and self-heals: a READY space with no properties gets one written, which is what the interrupted creation would have done. The properties query no longer requires exactly one result; it sorts by id and takes the lowest, and since object ids are time-ordered the original always beats a healed duplicate. A space that somehow acquires two is no longer permanently uninitialized.

**`@dxos/app-toolkit`** — `getSettingsSpace` resolves duplicates deterministically (the space carrying the default-space designation, then id order) so every device and every boot agrees rather than picking whatever `.find()` returns first.

**`@dxos/plugin-space`** — `ensureSettingsSpace` re-checks after creating and deletes the space *it* created when it loses the tie-break. Each client only ever deletes its own.

**`@dxos/plugin-routine`** — separate bug found while chasing a recurring `Trigger has no runnable reference` in the same profile. Deriving a `runnable` from an absent `spec` yields `undefined`, so `wireTriggers` left an actionless routine's triggers enabled and pointing at nothing; the runtime treats that as a defect, so a timer trigger failed on its whole schedule. `makeRoutine` guarded its own call — the guard now lives in `wireTriggers`, since one unguarded caller is enough to strand a trigger.

## Testing

New `space-proxy.test.ts` genesis-es a space through the raw `SpacesService`, skipping the client-side properties write, and asserts it still reaches READY. It hangs on the old code; ~18s on the new (the 15s heal timeout plus setup).

The new `wire.test.ts` case was verified by deleting the guard and re-running — it fails, and so does the pre-existing \"leaves pre-set trigger bindings alone\" test, confirming `wireTriggers` had no protection of its own.

\`\`\`
client         Test Files 5 passed | 2 skipped | Tests 20 passed
app-toolkit    Test Files 13 passed (13)      | Tests 139 passed
plugin-space   Test Files 14 passed | 1 skip  | Tests 46 passed | 3 todo
plugin-routine Test Files 9 passed (9)        | Tests 66 passed
:lint          309 tasks completed
\`\`\`

## Reviewer notes

- The 15s heal timeout is the main judgement call. Erring early costs an orphaned properties object (harmless — the id sort makes the original win); erring late holds the space's mutex for that long on every boot. I weighted it toward early.
- The duplicate-delete in `ensureSettingsSpace` still has a residual window: if a client has not yet seen the other client's settings space, it sees itself as canonical and keeps its own. That is inherent to check-then-create across clients and is much narrower than today's behavior.
- Not addressed: `trigger-dispatcher.ts:459` dies on an unrunnable trigger rather than skipping it. That is a data condition, not a defect, but changing error semantics in compute-runtime felt like someone else's call.
- Profiles that already carry a wedged duplicate recover on next boot via the heal, without console surgery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed actionless routines so their triggers remain safely configured without referencing an unavailable action.
  * Improved settings-space creation when multiple clients create spaces concurrently.
  * Settings-space selection is now deterministic, prioritizing the designated default space and then a stable identifier.
  * Duplicate spaces are cleaned up when appropriate, while the canonical space is preserved.

* **Tests**
  * Added coverage for actionless routine triggers and duplicate settings-space resolution.


packages/sdk/client/src/echo/space-proxy.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const space = await asyncTimeout(
+      new Promise<Space>((resolve) => {
+        const subscription = client.spaces.subscribe((spaces) => {
+          const match = spaces.find(({ key }) => key.equals(spaceKey));
+          if (match) {
+            subscription.unsubscribe();
+            resolve(match);
+          }
+        });
+      }),
```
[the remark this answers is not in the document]
richburdon: Valid — the space list replays on subscribe, so a match already present would read `subscription` in its TDZ; the handle is now stored before use with a synchronous-match guard after assignment.

packages/plugins/plugin-routine/src/util/wire.test.ts
```diff
@@ -71,6 +71,21 @@ describe('wire', () => {
     expect(trigger.runnable?.uri).toBe(preset.uri);
   });
 
+  test('wireTriggers leaves an actionless routine alone rather than stranding its triggers', ({ expect }) => {
+    // Nulling `runnable` here leaves an enabled trigger dispatching nothing, which the runtime treats as a
+    // defect on every firing — a timer trigger then fails for its whole schedule.
```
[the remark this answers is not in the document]
richburdon: Reworded to describe the hypothetical clearing rather than claiming the test nulls the field.

packages/plugins/plugin-space/src/util/settings-space.ts
```diff
@@ -22,10 +23,22 @@ export const ensureSettingsSpace = Effect.fnUntraced(function* (client: Client)
...
+  const canonical = AppSpace.getSettingsSpace(client) ?? created;
+  if (canonical.id !== created.id) {
+    log.warn('discarding duplicate settings space', { created: created.id, canonical: canonical.id });
+    yield* Effect.promise(() => created.delete());
+    yield* Effect.promise(() => canonical.waitUntilReady());
+    return canonical;
+  }
+
+  yield* Effect.promise(() => created.internal.setEdgeReplicationPreference(EdgeReplicationSetting.ENABLED));
+  return created;
```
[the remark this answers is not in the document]
richburdon: Added `discards the settings space it created when another client already made one` — the space list is now taken structurally so the test can hide the canonical space until this client creates its own; verified it fails when the delete branch is disabled.

packages/sdk/app-toolkit/src/echo/AppSpace.test.ts
```diff
@@ -50,8 +53,51 @@ describe('space visibility', () => {
...
+  const properties = Obj.make(Expando.Expando, {});
+  if (defaultSpaceId) {
+    Obj.update(properties, (properties) => {
+      Annotation.set(properties, AppAnnotation.DefaultSpaceAnnotation, defaultSpaceId);
+    });
+  }
+  return properties;
+};
+
+const makeClient = (spaces: Space[]) => ({ spaces: { get: () => spaces } });
```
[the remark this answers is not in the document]
richburdon: Not applied: `makeClient` is declared once, at line 103 — there is no duplicate. `app-toolkit:test` (139 passed) and `app-toolkit:lint` both pass on this file.

richburdon:
Addressed the nitpick from the review body (no inline thread to reply to): `_awaitProperties` is now `#awaitProperties` in `space-proxy.ts`, per the repo preference for ES private members in new code. `client:build`, `client:test` and `client:lint` all pass.

On the Codecov report — it was computed against `60ea966`, which predates the coverage added in `460bf067` (the settings-space duplicate-cleanup test). The next run should reflect it.

richburdon:
@wittjosiah — context for review, since this starts in #12503 but the underlying defect predates it.

**What happened.** My profile acquired a *second* settings space, created without a properties object. Composer rendered the space but the nav was dead, with `Finding properties for a space` >5s, `Automerge root doc load timeout`, repeated `subduction: session lost`, and a 2s `Waiting for the edge replication to be enabled`.

**The landmine (pre-existing).** Space creation is two-phase — the host does genesis, then the *client* writes `SpaceProperties` in `SpaceList._createSpaceInternal`. Interrupt in between (reload, closed tab, a throw) and the space can never open, because `_initializeDb` waited on those properties **unbounded** — inside `_processSpaceUpdate`, which is `@synchronized`. That holds the space's mutex for the whole session, so `_anySpaceUpdate` never fires again, which is exactly why `setEdgeReplicationPreference` (which awaits it) timed out. Then `ensureSettingsSpace`'s unbounded `waitUntilReady()` meant `resolveDefaultSpace` never returned and `SwitchWorkspace` never fired.

**What #12503 changed.** It put a space creation on the boot path of *every existing profile*, which is what made that window reachable in practice. Two clients (second tab, or a reload racing the first) can both see no settings space and both create one, and nothing enforced 'exactly one'.

**Two judgement calls worth your eye:**

1. The 15s heal timeout in `space-proxy.ts`. Erring early costs an orphaned properties object — harmless, because the id sort is time-ordered so the original always wins; erring late holds the mutex that long on every boot. I weighted it early. Push back if you'd rather it were longer.
2. `ensureSettingsSpace` only ever deletes the space *it* created. When the pre-existing one is populated it carries the default-space designation, and `getSettingsSpace` prefers that over id order — so a populated settings space can't lose to a fresh empty one. There is still a residual window: a client that hasn't yet seen the other's space sees itself as canonical and keeps it.

**Not addressed:** `trigger-dispatcher.ts:459` dies on an unrunnable trigger rather than skipping it. That's a data condition, not a defect, but changing error semantics in compute-runtime seemed like Dmytro's call.

**Worth knowing regardless of this PR:** anyone who upgraded may already have a wedged duplicate. The heal recovers them on next boot; without it they need console surgery, which isn't something a non-developer can do.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

richburdon:
Moved the `ensureSettingsSpace` race test to `packages/plugins/plugin-space/src/util/settings-space.test.ts` — it calls the util directly rather than a migration entry point, so it belongs beside its module. `plugin-space:test` 47 passed, `plugin-space:lint` clean.

packages/sdk/client/src/echo/space-proxy.ts
```diff
@@ -465,9 +480,38 @@ export class SpaceProxy implements Space, CustomInspectable {
...
       );
     }
-    await warnAfterTimeout(5_000, 'Finding properties for a space', () =>
-      cancelWithContext(this._ctx, propertiesAvailable.wait()),
-    );
+    await cancelWithContext(this._ctx, this.#awaitProperties(propertiesAvailable));
+  }
+
+  /**
+   * Wait for the space's properties object, writing one if the space has none.
```
dmaretskyi: this shouldn't ever happen, properties are created atomically with the space

richburdon: Not pushing a change for this yet — I think the premise doesn't hold today, and I'd rather agree on which fix you want than quietly drop the recovery.

Creation is two-phase, not atomic. The host's `SpacesService.createSpace` takes only `{ tags, membershipPolicy }` and never writes `SpaceProperties` — the only write is client-side, in `SpaceList._createSpaceInternal` (`space-list.ts:287-290`), *after* genesis has already completed:

```ts
spaceProxy.db.add(Obj.make(SpaceProperties, meta ?? {}), { placeIn: 'root-doc' });
await spaceProxy.db.flush();
```

The only `SpaceProperties` reference anywhere in `client-services` is a read in a diagnostics fetch (`data-space-manager.ts:246`). So anything that interrupts between genesis and that write — reload, closed tab, a throw — leaves a space with none.

Three independent confirmations it does happen:

1. **Observed in the wild.** @richburdon's profile had a space tagged `org.dxos.space.settings` in `SPACE_INITIALIZING` with no properties object and `edgeReplication: undefined`, alongside a healthy one. That space is what wedged his boot — it held `_processSpaceUpdate`'s mutex, which is why `setEdgeReplicationPreference` was timing out at 2s.
2. **Reproduced.** `space-proxy.test.ts` genesis-es via the raw `SpacesService` and skips the client-side write. It hangs on the old code and passes on this branch.
3. **The 30s-plus `Finding properties for a space` warnings** in his console, from a wait that had no bound.

That said — if your point is that creation *should* be atomic and healing treats the symptom, I agree that's the better fix and I'd rather do that. I skipped it here because it means moving the properties write into host-side genesis, which is a protocol change and felt beyond a hang fix.

1. Make genesis atomic (host writes `SpaceProperties`) and drop the heal.
2. Make genesis atomic **and** keep the heal, since profiles already carrying a property-less space need something to recover them — @richburdon's did, and it took console surgery.
3. Keep just the heal as-is and open a separate issue for atomic genesis.

I'd argue for 2: atomicity fixes the future, the heal fixes the profiles already broken. Your call on which, and happy to take the atomic-genesis work in this PR if you want it here.

packages/sdk/client/src/echo/space-proxy.ts
wittjosiah, line 1: this is not the appropriate fix for the space properties issue. If it happens again we will get a profile export to debug the root cause and address there.

wittjosiah (changes requested):

packages/sdk/client/src/echo/space-proxy.ts
richburdon: Done — reverted. `space-proxy.ts` is back to its prior behaviour and its test is removed (`17863810c4`). No client changes remain in this PR.

What stays is the settings-space deduplication, which addresses the upstream cause rather than the symptom: two clients that both saw no settings space both created one, which is what put a property-less space on the boot path. `getSettingsSpace` now resolves duplicates deterministically and `ensureSettingsSpace` discards the space it created when it loses the tie-break. Plus the unrelated `wireTriggers` guard.

On the profile export — no need to wait for a recurrence. @richburdon's affected space is still in his profile: `BR3Q2LZ76FMTGVWMW6DOQCMARHABO7VNU`, tagged `org.dxos.space.settings`, no properties object, `edgeReplication: undefined`. We only `close()`d it, never deleted it, so an export taken now should still contain it alongside the healthy settings space `BDU3GY73PNXKJMSUCH33R3AUZAFC6LQXX`. Say the word and we'll get you one before he cleans it up.

One thing worth recording wherever the root cause gets tracked: the wait in `_initializeDb` is unbounded and runs inside `_processSpaceUpdate`'s `@synchronized` mutex, so a space in that state holds the mutex for the whole session — that is why the visible symptom was `setEdgeReplicationPreference` timing out at 2s on unrelated spaces, which is a confusing place to start debugging from.

packages/sdk/client/src/echo/space-proxy.ts
richburdon: Closing this out — @wittjosiah has ruled on it and the heal is reverted in `17863810c4`, so the question of atomic genesis vs healing is moot for this PR. The root cause will be diagnosed from a profile export instead; I've noted on his thread that @richburdon's affected space is still present and exportable now.

packages/plugins/plugin-space/src/util/settings-space.ts
```diff
@@ -6,26 +6,51 @@ import * as Effect from 'effect/Effect';
...
+import { type Space } from '@dxos/client/echo';
+import { log } from '@dxos/log';
 import { EdgeReplicationSetting } from '@dxos/protocols/proto/dxos/echo/metadata';
 import { MembershipPolicy } from '@dxos/protocols/proto/dxos/halo/credentials';
 
+/**
+ * The space list and factory this needs from a `Client`. Structural so a test can present a list
+ * that changes between calls — the concurrent-creation race this function exists to resolve.
+ */
+type SpaceHost = {
```
wittjosiah: I don't think that we should be introducing custom types for client-shaped things in plugins

wittjosiah (changes requested):

packages/plugins/plugin-space/src/util/settings-space.ts
richburdon: Reverted in `05a06e1b1f` — `ensureSettingsSpace` takes `Client` again and `SpaceHost` is gone.

Context on why it appeared, since it wasn't gratuitous: CodeRabbit asked for coverage of the discard branch, and the only way to reach it is a space list that returns different results across two calls — which a real `Client` can't do. The structural type was how the test hid the canonical space until this client created its own. With it gone that test is deleted too, so the discard branch is now uncovered. The tie-break rule it depended on is still covered by the `getSettingsSpace` unit tests in `app-toolkit`.

If you want that branch covered, the options I can see are:

1. Leave it uncovered — the branch is three lines and the tie-break is tested separately.
2. Cover it at the client level instead, where a fake space list is not out of place.
3. Take the dedupe out of `ensureSettingsSpace` entirely and let `getSettingsSpace`'s deterministic pick be the whole fix — a duplicate would persist but would be harmless and consistently ignored.

Happy with 1 unless you'd prefer otherwise.

richburdon:
On the outside-the-diff finding about global race-safety — split response, since the two halves differ in merit.

**Applied** (`cdae7f4bbf`): the changeset overclaimed. It said every device and boot would agree; the tie-break is only deterministic for a *given* space list. Reworded, and it now states the residual explicitly — a client that has not yet seen another's settings space keeps what it created, so a duplicate can survive until the lists converge, at which point it is consistently ignored rather than competing.

**Not applied**: adding coordination or a reconciliation pass. Three reasons.

1. It is a known, documented residual, not a new finding — it is called out in the PR description and in my reply to @wittjosiah above. Check-then-create across independent clients cannot be made race-free without coordination, and this PR is a hang fix.
2. "Another client may already be using a space before its creator deletes it" does not apply as stated: a client only ever deletes the space *it* created, and only when it loses the tie-break. For another client to be using that space, it would have to have picked it as canonical — which the shared tie-break rule prevents, since both clients evaluate the same rule over the same designation-then-id ordering.
3. The discard branch may not survive review at all. @wittjosiah is weighing removing it and relying solely on `getSettingsSpace`'s deterministic pick, which would leave duplicates in place by design. Building coordination for a branch that might be deleted is premature.

If real coordination is wanted, it belongs with atomic space genesis rather than bolted onto a plugin helper — and that is the root-cause work @wittjosiah has said will be driven from a profile export.

PR #12529, opened 2026-08-09
echo: add per-space storage metrics and garbage collection

dmaretskyi:
Closes DX-1151.

## What

Adds two per-space maintenance APIs to the ECHO `Database`, both routed through the `DataService` RPC so they work against local and remote (EDGE) hosts:

- **`db.stats()`** — a read-only snapshot of what a space occupies on the host: live/deleted object counts, automerge document count, feed count, and total feed-block count.
- **`db.runGarbageCollection(options?)`** — reclaims storage held by soft-deleted objects and the documents they orphan, returning a report of what was reclaimed.

## Garbage-collection algorithm

Implemented per the new spec (`packages/core/echo/echo-host/docs/GARBAGE_COLLECTION.md`), which documents all five steps, the safety invariants, and the EDGE re-implementation:

1. **Unlink soft-deleted objects** from the space directory — drop deleted inlined objects from the root document, remove the `links` entry for deleted linked objects.
2. **Wipe unreachable documents owned by the space** — enumerate every stored document, and wipe (chunks + heads) any that are no longer reachable from the space directory **and** carry `access.spaceId` for this space. Attribution is the safety boundary: a document that cannot be positively attributed to the space is never touched.
5. **Clear stale index rows** (`index` option, default on) — delete `objectMeta` rows for the reclaimed documents/objects, cascading into the FTS and reverse-ref indexes (by record id) and dropping the tracker cursors.

Steps 3–4 (feed-block purge) are **specified and deferred** on the local host: this repo's feeds are SQLite-backed rather than reflected as automerge documents, and feed `blocks` rows store opaque/optionally-sealed payloads with no object-id column, so a keyed purge would need either payload decoding (keys) or a schema migration. `runGarbageCollection` accepts a `feeds` option and reports `purgedFeedBlocks` (currently `0`); the positioned-tombstone safety invariant is documented for when it lands.

## Implementation notes

- New host logic lives in `echo-host` (`space-storage.ts`, `garbage-collector.ts`); `EchoHost.getSpaceStats` / `EchoHost.runGarbageCollection` delegate through `DataServiceImpl` handlers (same callback pattern as `updateIndexes`).
- New storage primitives: `SqliteHeadsStore.remove`, `AutomergeHost.removeDocument`, `IndexEngine.deleteObjects` (+ per-index `deleteByRecordIds` / `selectRecordIdsForRemoval` / `IndexTracker.deleteCursors`), all chunked to respect the SQLite bound-variable limit.
- The EDGE (`functions-runtime-cloudflare`) `DataService` handlers for `stats` / `runGarbageCollection` return `NotImplementedError` for now, consistent with the spec.
- Additive and opt-in: no existing read/write path changes; the new code runs only when the APIs are called.

## Testing

- New end-to-end tests in `echo-client-e2e/src/storage-metrics.test.ts`: empty-space stats, alive/deleted + document counts, feed/feed-block counts, GC unlink + document wipe + index-entry cleanup, and idempotency + persistence across reopen.
- Full suites green locally: `index-core` (46), `echo-host` (274), `echo-client` repo-proxy (12), `echo-client-e2e` (291). Lint clean on the changed packages.

---
_Generated by [Claude Code](https://claude.ai/code/session_011ctavB6AdobLPZPXhjmaes)_

## Summary by CodeRabbit

* **New Features**
  * Added per-space database storage statistics for objects, documents, feeds, and feed blocks.
  * Added garbage collection to reclaim deleted and unreachable data.
  * Added optional cleanup controls for indexes and feeds.
  * Garbage collection reports the number of removed items and supports repeatable runs.

* **Documentation**
  * Added guidance covering metrics, cleanup behavior, safety, and idempotency.

* **Tests**
  * Added coverage for metrics, cleanup results, repeated runs, queries, and database reopen persistence.


dmaretskyi:
Thanks — addressed the substantive findings in `bdfbf4f`:

- **`stats()` hang on an unavailable linked doc** → `countSpaceObjects` now loads storage-only (`fetchFromNetwork: false`); unavailable docs aren't counted.
- **Wiped doc re-persisted via a live Repo handle** → `AutomergeHost.removeDocument` now evicts the handle (`repo.removeFromCache`) before deleting chunks/heads.
- **`chunkArray`** now asserts a positive-integer size.
- **EDGE availability wording** corrected in the spec intro, changeset, and the `runGarbageCollection` RPC doc comment (local-only today; feed-block purge deferred).
- **Test** switched to context-scoped `expect`.

Two I'm intentionally skipping:

1. **`toDocumentId` `as AutomergeUrl` (space-storage.ts:28)** — kept. These URLs come from the space directory's own `links`/`branches`, which are always valid Automerge URLs, and this mirrors the established repo idiom (`entity-manager.ts:652,848`). Adding `isValidAutomergeUrl` validation would silently drop a malformed internal link rather than surface the bug.
2. **`#private` for the new handlers (data-service.ts)** — kept as quoted TS-private (`'_getSpaceStats'` etc.) to match this class's existing members, whose quoted names are load-bearing for the `makeInProcessClient` prototype normalization (`rpc-bridge.ts`). Introducing `#private` only for the new fields would be inconsistent here.

---

packages/core/echo/echo-host/src/db-host/garbage-collector.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+/**
+ * Per-space garbage collection, steps 1–2. See `docs/GARBAGE_COLLECTION.md` for the full algorithm
+ * and its safety invariants: unlink soft-deleted objects from the space directory, then wipe every
+ * document owned by the space that is no longer reachable from the (post-unlink) directory.
+ *
+ * Index (step 5) and feed (steps 3–4) reclamation are layered in by the host from this result.
+ */
+export const runSpaceGarbageCollection = async (deps: GarbageCollectorDeps): Promise<SpaceGarbageCollectionResult> => {
```
dmaretskyi: put this as a method on something. db host itself?

packages/core/echo/echo-host/src/db-host/space-storage.ts
```diff
@@ -0,0 +1,91 @@
...
+  branchDocumentIds: DocumentId[];
+};
+
+const toDocumentId = (url: AutomergeUrl | string): DocumentId => interpretAsDocumentId(url as AutomergeUrl);
+
+/**
+ * Enumerate the documents reachable from a space root (root + object links + branch members).
+ * Requires the root document to be loaded.
+ */
+export const collectSpaceDocuments = (root: DatabaseRoot): SpaceDocumentSet => {
```
dmaretskyi: perfer methods over functions

dmaretskyi:
Done in `cd94c0e` — folded both `space-storage.ts` and `garbage-collector.ts` into private methods on `EchoHost` (the db host), alongside the `getSpaceStats`/`runGarbageCollection` methods that already lived there: `_collectSpaceDocuments`, `_allSpaceDocumentIds`, `_countSpaceObjects`, `_unlinkDeletedObjects`, `_wipeUnreachableDocuments`. Both helper files are deleted; behavior is unchanged and the e2e suite still passes.

---

PR #12532, opened 2026-08-10
plugin-client: pin WebAuthn relying party to composer.space

richburdon:
## Problem

Recovery-passkey creation passed `rp: { id: location.hostname }`, so a passkey created at `labs.composer.space` is bound to the rpId `labs.composer.space`. The hub (`packages/services/hub-service` in dxos/edge) verifies assertions with `DX_WEBAUTHN_RP_ID="composer.space"` in every deployed environment, and `https://composer.space/.well-known/webauthn` pins the allowed origins. A labs-created passkey therefore can never satisfy the hub: MCP authorization (`auth.dxos.network/auth/mcp`) and the account profile (`account.composer.space/profile`) both reject it, and credential providers such as 1Password never offer it on those origins.

Confirmed empirically: a passkey created at `labs.composer.space` is not offered at `composer.space`.

## Change

New `getRelyingPartyId()` in `@dxos/app-toolkit/NativePasskey`, next to the existing `APP_DOMAIN` constant the native (Tauri) path already pins:

- host is `composer.space` or any subdomain (`labs.`, `staging.`) → `composer.space`. It is a registrable-domain suffix of both, so one credential works across all three deployments and matches what the hub already expects.
- anything else (`localhost`, netlify previews) → the page host, since `composer.space` is not a valid rp id there and local dev must keep working.

Applied to both sides so they stay in sync:

- `create-passkey.ts` — `rp: { id: getRelyingPartyId(), name: 'Composer' }`
- `redeem-passkey.ts` — `rpId: getRelyingPartyId()` (without this, the assertion at labs would still scope to the labs hostname and never find the new credential)

Unchanged, deliberately: the WebAuthn `user.id` is still the identity's `lookupKey`, and `residentKey: 'required'` remains — the hub resolves the credential by `userHandle` via `AGENTS_SERVICE.getRecoveryInformation`, and discoverable login depends on it.

## ⚠️ Breaking for existing passkeys

Passkeys already created at `labs.composer.space` are **orphaned** by this change and must be re-created. There is no migration path — the rpId is baked into the credential at creation time.

## Testing

- Five table-driven cases for `getRelyingPartyId` in `NativePasskey.test.ts`; `moon run app-toolkit:test` → 9 passed.
- `plugin-client:build` and lint clean; `pnpm format` run.
- **Not yet run:** the end-to-end check needs a deployed build of this branch and a real authenticator — create a recovery passkey from `labs.composer.space`, then sign in at https://account.composer.space/profile. The passkey should be offered directly by the platform or password-manager authenticator, with no QR / cross-device fallback.

Reviewer note: `https://composer.space/.well-known/webauthn` must list the labs and staging origins, or the browser will reject `rp.id: composer.space` from those pages at *creation* time with a `SecurityError`. Worth confirming before the e2e test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebAuthn passkey compatibility across deployed Composer environments by using a consistent relying-party domain.
  * Passkey requests now work correctly across supported Composer subdomains and custom hostnames.
  * Added fallback handling when the current hostname is unavailable.

* **Important Note**
  * Existing passkeys created on `labs.composer.space` must be recreated after upgrading.

### DIFF
--- a/.changeset/lucky-pianos-shake.md
+++ b/.changeset/lucky-pianos-shake.md
@@ -1,0 +1,6 @@
+---
+'@dxos/plugin-client': patch
+'@dxos/app-toolkit': patch
+---
+
+Pin the WebAuthn relying party to `composer.space` for deployed builds so recovery passkeys created at labs/staging are accepted by the hub. Existing passkeys created at `labs.composer.space` are orphaned and must be re-created.
--- a/packages/plugins/plugin-client/src/operations/create-passkey.ts
+++ b/packages/plugins/plugin-client/src/operations/create-passkey.ts
@@ -47,7 +47,7 @@ const handler: Operation.WithHandler<typeof CreatePasskey> = CreatePasskey.pipe(
               navigator.credentials.create({
                 publicKey: {
                   challenge: new Uint8Array(),
-                  rp: { id: location.hostname, name: 'Composer' },
+                  rp: { id: NativePasskey.getRelyingPartyId(), name: 'Composer' },
                   user: {
                     id: lookupKey.asUint8Array() as Uint8Array<ArrayBuffer>,
                     name: identity.did,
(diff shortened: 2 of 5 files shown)


richburdon:
Addressed the CodeRabbit nitpick (`NativePasskey.test.ts`, `getRelyingPartyId` suite): added a case stubbing `location` as `undefined` so the non-browser `APP_DOMAIN` fallback is exercised — 10 tests pass.

PR #12535, opened 2026-08-10
echo: close GC correctness gaps, reclaim on every peer, and surface what a space occupies

wittjosiah:
Stacked on #12529 (`claude/storage-metrics-garbage-collection-ywhkgf`) — review that first. This PR is the delta on top of it.

## Correctness fixes to garbage collection

1. **Transitive deletion** (`echo-host/src/db-host/deletion.ts`). A child of a deleted parent, or a relation with a deleted endpoint, queries as deleted but carries no `deleted` flag of its own. #12529 tested only the flag, so those objects were invisible to every query yet retained forever. `DeletionResolver` applies the same rule `ObjectCore.isDeleted` uses, including its depth cap, resolving across documents because a parent routinely lives in a different one than the object it condemns.
2. **Subduction records** (`sqlite-storage-adapter.ts`). `removeDocument` swept only the classical chunk keys. The six `subduction-*` key families are disjoint from those and, on the default transport, hold most of a document's bytes — measured at 3 of 4 chunks for a fresh document — so collection was reclaiming roughly a quarter of what it reported.
3. **The wipe is now atomic.** It deleted the heads row and then each chunk range in seven separate awaits. Interrupting it left chunks with no heads row, and the orphan scan *is* `_headsStore.iterateAll()` — so those bytes could never be found again. A partial wipe was unrecoverable rather than merely incomplete. One transaction now, which also subsumes the indexer race the heads-first ordering was defending against. Covered by a test that fails the wipe midway and asserts the document survives whole; verified to fail against the sequential implementation with the heads count at 0 and the chunks retained.
4. **Announce sets.** `_documentsToSync` / `_documentsToRequest` / `_documentsRequested` / `_headsUpdates` retained wiped ids, so the host kept advertising documents it no longer had.

## Reclamation propagates

Documents that leave a space directory are wiped locally, so one explicit collection frees disk on every peer as the change replicates — no second invocation anywhere. A retired epoch root is expanded to its full closure rather than trusting the directory listing, which is debounced and can miss a document linked and orphaned in quick succession.

This rests on link removal being atomic and causally delivered, which I verified with three automerge experiments before relying on it: concurrent multi-key unlinks arrive as one change, and a change whose dependencies are missing is held rather than applied, so no peer observes a half-unlinked directory.

It does not make deletion ambient: the soft-to-hard promotion is decided by whoever ran the collection and arrives as replicated data, so a receiving peer applies a decision already taken.

## `Database.retainObjects(keep)`

Clearing a space names what to **keep**. The retained ids are diffed against the space directory's own maps and everything else is dropped in a single root change — no query over the space, no object loaded, no `deleted` flag written per object. `RemoveAllObjects` is `retainObjects` + `runGarbageCollection`, and no longer offers undo.

It runs on the client, and that is the whole finding. A host-side rewrite is storage-correct — measured `removedObjects: 2, removedDocuments: 2, removedIndexEntries: 2` — and still leaves every object answering queries, because `_handleSpaceRootDocumentChange`, the reconciliation that evicts objects missing from the directory, is reachable only from `updateSpaceState`, which early-returns unless the root **url** changed. It fires on an epoch and never on a content change to the same root. Deleting the index rows doesn't help either: the client's working set is consulted first.

So the client now also evicts objects whose directory entry was removed, reading the ids from the change patches rather than diffing the working set (a root change accompanies every object write). Objects mid-creation are skipped — `_createDocumentForObject` binds an object before writing its link. This is what makes a collection visible in the UI at all, and it covers a peer applying a replicated collection too.

## Surfacing it

- **`SpaceOperation.CollectGarbage`**, plus a debug-plugin toolbar built on `ToolbarMenu` — the idiom owns the pending state, so only the running action disables. Both destructive actions confirm; `notify` carries the error and a toast carries the count, since `Label` interpolates `count` alone and cannot take a result-derived value.
- **Stats panel.** Its `objects` row was dead — `DatabaseInfo.objects` was declared and never assigned, so it always rendered `0`. It now carries `db.stats()` summed across open spaces: live and deleted objects, documents on disk, feeds and blocks. The existing `documents` row is renamed **`documents (loaded)`**: it counts resident handles and diverges from what is stored — a collection drops one and barely moves the other.
- **Space devtools panel** renders the space's own `db.stats()` as JSON beside sync state. Polled, not subscribed: `stats()` walks the space's documents on the host and has no change feed.

## Testing

- `echo-host` — 277 passed / 21 files, including `auto-reclaim.test.ts` (wipe on unlink, wipe a retired root and its closure, atomicity under a mid-wipe failure).
- `echo-client` — 519 passed / 30 files.
- `echo-client-e2e` `storage-metrics` — 7 passed: #12529's 5, one for transitive deletion, one asserting `retainObjects` drops the rest and `stats.documents` falls to 2, so the documents are off disk rather than hidden.
- `plugin-space` — 49 passed / 16 files, including a new `CollectGarbage` test and #12500's `RemoveAllObjects` test restored unmodified.

## Known limits

- **Index rows can outlive their documents** if the process dies between the wipe and the index delete. Closing it needs a transaction spanning the automerge and index stores.
- **EDGE is unimplemented** — `stats` and `runGarbageCollection` return `NotImplementedError`. Its prerequisite is `SpaceStateMachine.getAutomergeRoot` ordering by `Epoch.number` rather than DO-KV key order.
- **The file does not shrink.** Deleting rows returns pages to SQLite's freelist; `VACUUM` would reclaim them to the filesystem but cannot run inside a transaction, so it belongs to a maintenance path of its own.
- **Two-peer convergence is untested** — automatic reclamation is covered single-host.
- **Orphan scan has no age guard**, so a document created but not yet linked is theoretically collectable. Narrowed by the storage-only load and the `access.spaceId` check, not closed.
- **Two epoch-root-swap races and `MigrationBuilder._buildNewRoot`** are real bugs found along the way but on no path this PR touches; recorded in the doc rather than fixed here.

All are written up in `echo-host/docs/GARBAGE_COLLECTION.md` alongside the algorithm.

https://claude.ai/code/session_01V4nVDquQe7GDJ3JKnTzMPu

## Summary by CodeRabbit

* **New Features**
  * Added garbage collection to permanently remove deleted objects and stored documents.
  * Added object retention to remove everything outside a selected set.
  * Added automatic cleanup for transitively deleted documents and relationships.
  * Added database storage statistics, including deleted objects, stored documents, feeds, and blocks.
  * Added developer tools actions for resetting space objects and collecting garbage, with confirmations and result notifications.

* **Bug Fixes**
  * Removed document data is now cleaned up atomically and consistently across peers.
  * Failed cleanup operations safely preserve affected data.


wittjosiah:
Triaged the review. Fixed in 9a3f835:

- **`deletion.ts` memo vs. depth** — real bug, and the one the class comment warns about. The memo is now keyed by `(entityId, remainingDepth)`, so a `false` produced by exhausting the budget is never reused for a walk with a deeper budget. Added `deletion.test.ts` covering it (it fails against the old keying).
- **`retainObjects` leaves `branches` behind** — real leak. Branch registry entries are keyed by root object id, and a surviving entry keeps its member documents reachable, so dropped objects' branches now go with them.
- **Reclaim with no live root** — guarded. It now returns instead of treating an empty `reachable` set as ground truth.
- **`_spaceDocumentIds` not cleared in `removeSpace`** — fixed.
- **`GARBAGE_COLLECTION.md` "per-space" contradiction** — reworded; all three APIs are per-space, only the first two are RPC-routed.
- **`as any` in `auto-reclaim.test.ts`** — removed (the project forbids casts). Reaches through new `EchoHost.automergeHost` / `AutomergeHost.storage` accessors, and the patched method is restored in a `finally`. Also switched to `sleep` from `@dxos/async` and gave the e2e `expect.poll` an explicit 5s timeout.

Not changing:

- **Pending document creation restoring a dropped object** — the object is not in `doc.links` when `retainObjects` diffs, so it is not in the drop set and is not evicted; its link lands afterwards and it stays reachable. An object created concurrently with a clear surviving the clear is the intended outcome, not a resurrection.
- **`#collectClosure` on the live root** — the shallow `#collectSpaceDocuments(root)` is root-local, whereas branch members hang off object documents too. Substituting it would under-approximate reachability and wipe live data; under-approximating is the one error this path cannot make. The walk only runs when documents actually departed the directory or the root was retired.
- **`MenuBuilder` not awaiting the promise** — `executeMenuAction` awaits the effect, and the effect's success value is the promise the callback returned; promise resolution flattens thenables, so the `await` does wait for it and the toolbar item's pending state spans the whole operation.
- **Double-quoted `description`** — that is the formatter's output for a string containing an apostrophe. `pnpm format` (oxfmt) would rewrite the single-quoted-with-escape form straight back, and the codebase does the same elsewhere.

---

PR #12541, opened 2026-08-11
edge-client: fetch the auth nonce from /auth instead of provoking a 401

dmaretskyi:
Client half of the edge auth-nonce work — pairs with dxos/edge#854. This is dxos/dxos#12408 rebased onto current `main` (authorship preserved, `-x` cherry-picks) plus a changeset, so **#12408 can be closed in favour of this PR**.

## Problem

Both edge clients obtained their auth challenge by sending a request they expected to be **rejected**.

`BaseHttpClient` prefetched `/auth` and only acted on a failure:

```ts
if (!this._authHeader && args.auth) {
  const response = await fetch(new URL('/auth', this._baseUrl));
  if (response.status === 401) {                    // <- the success case
    this._authHeader = await this._handleUnauthorized(response);
  }
}
```

`EdgeClient._createAuthHeader` was worse — it never touched `/auth` at all. It fired a GET at the `/ws/:identityDid/:peerKey` upgrade path itself, purely to harvest the challenge off that path's 401.

Both produced a red `Failed to load resource: 401` in the browser console on every client boot, and a routine `auth.failure` in edge's audit trail, for an operation that succeeded.

## Changes

Both paths now go through a shared `authenticateViaChallengeEndpoint`, which GETs `/auth` and reads the challenge from **either** a 200 body or a `WWW-Authenticate` header. Edge answers 200 for anonymous callers as of dxos/edge#854; the header path keeps this working against servers that predate it. Both clients keep their lazy 401-and-retry path for stale credentials.

Three real bugs fell out of the shared parser:

1. **Challenge lists were unparseable.** `WWW-Authenticate` carries a comma-separated *list* (RFC 9110 §11.6.1), but the old code asserted the header *started with* `VerifiablePresentation challenge=`. So `Bearer realm="dxos", VerifiablePresentation …` — exactly what edge emits whenever admin-key auth is also allowed — failed the invariant outright. `payments-service` carries a separate VP-only middleware purely to dodge this.
2. **Quotes were never stripped.** `slice()` kept the surrounding quotes and relied on `Buffer.from` silently discarding them. Works today; `atob` would throw.
3. **Dead protocol assignment.** `httpUrl.protocol = getEdgeUrlWithProtocol(...)` assigned a whole URL string to `URL.protocol`, which is an invalid protocol value and silently ignored. `_baseHttpUrl` was already correct, and is now used directly.

The prefetch is skipped when no identity is set — fetching a challenge you cannot sign is a wasted round trip — and (from the rebase onto `main`) when an API key is in play, since that path never populates the VP header.

`EdgeCredentialsHeaderCodec` in `@dxos/protocols` is new: it owns the `Authorization` and `sec-websocket-protocol` encodings that were previously open-coded on both sides.

## Why this matters for the nonce rollout

`EdgeClient._createAuthHeader` runs per connection open, so WebSocket reconnects mint a fresh challenge rather than replaying a cached one — that is what makes a short nonce TTL safe on the `/ws` path once dxos/edge#854 enforcement is flipped.

## Compatibility

| Server | Path taken |
|---|---|
| New (`200` + body) | Challenge read from body, no console error |
| Old (`401` + header) | Challenge read from header, unchanged behaviour |
| Challenge endpoint down | Falls back to lazy 401-and-retry |

`handleAuthChallenge` keeps its signature; the 401 assertion is relaxed to accept either shape, so `plugin-payments` needs no change.

## Merge order

Either order works — dxos/edge#854 is backwards-compatible on its own and this PR is compatible with both server versions — but the console noise only goes away once both are deployed.

## Test plan

- `moon run edge-client:test` — **57 passing** (4 skipped, 10 files). Coverage for header parsing (quoted, legacy unquoted, VP not first in the list, quote stripping), both response shapes, header-vs-body precedence, that the body is left unconsumed for the caller, network-failure tolerance, and that the signed bytes are the decoded challenge.
- `moon run protocols:test` — **26 passing**.
- `moon run edge-client:lint`, `edge-client:build`, `protocols:build` clean; `oxfmt` reports no changes.
- Changeset added (`@dxos/edge-client`, `@dxos/protocols`, minor) — the original PR had none.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LCeyDCFyB9Lxsv7Q96Meqb)_

## Summary by CodeRabbit

* **New Features**
  * Edge authentication now retrieves challenges from the `/auth` endpoint for HTTP and WebSocket connections.
  * Supports challenges delivered in response bodies or `WWW-Authenticate` headers.
  * Added shared credential encoding for HTTP authorization headers and WebSocket authentication protocols.

* **Bug Fixes**
  * Prevented unnecessary authentication retries when challenges are missing, empty, or unrelated.
  * Improved handling of quoted, comma-separated, and legacy challenge formats.
  * Preserved response bodies during authentication processing and improved fallback behavior.


dmaretskyi:
Reviewed all six findings against the code. Two were real defects in this branch's new code and are fixed in 1da2f934; two style nits applied; two skipped with reasons.

**Fixed — `parseChallengeHeader` matched the scheme anywhere in the header.** Confirmed: `Bearer realm="VerifiablePresentation challenge=…"` declares Bearer alone, but the unanchored regex read a challenge out of the quoted realm and made the client retry an unrelated 401 — running a non-idempotent request twice. Now the scheme is only recognised where a challenge starts, via a scan that tracks quoting (a plain comma split is wrong: an auth-param value may itself contain a comma). Also rejects a longer scheme that merely starts with the VP name. Three tests added, including the quoted-comma list case.

**Fixed — `base64ToUint8Array` let `atob` throw.** Confirmed against the file's own contract: both `EdgeCredentialsHeaderCodec` decoders document that anything unrecognised yields `undefined` so a caller can fall through to another auth method, and `atob` throws outside the base64 alphabet. Now returns `undefined`, with coverage for malformed HTTP and WebSocket payloads.

**Applied — `requestUrl` moved below the suites** per the code-style skill. It was above the first `describe` only to avoid reading before definition; the TDZ never applies, since it is called inside test callbacks.

**Applied — context `expect`** on the new `EdgeCredentialsHeaderCodec` malformed-payload test. The pre-existing tests in that file keep the module-level import; migrating them is unrelated churn for this PR.

**Skipped — browser-safe strict Base64 in `presentCredentialsForChallenge`.** The stated impact ("a valid challenge then throws in browsers") does not hold here: `Buffer.from(challenge, 'base64')` is exactly what the pre-existing `handleAuthChallenge` has shipped on the lazy 401 path, which is how Composer authenticates today — so `Buffer` is present in the app bundle and this PR introduces no new exposure. The strictness half is also inert: a garbled challenge just fails verification server-side. Purging `Buffer` from `edge-client` is worth doing, but as its own change across the package rather than one asymmetric call site here.

**Skipped — removing the `encodeAuthHeader` wrapper.** It is not a leftover from the move: `@dxos/edge-client` exports it and `plugin-payments` imports it. What moved into `@dxos/protocols` is the wire format, and `encodeAuthHeader` remains the package-level API over it. Deleting a published export to collapse a one-line alias would break out-of-repo consumers for no functional gain.

`edge-client` 60 passing (was 57), `protocols` 27 (was 26); lint, build and `oxfmt --check` clean.

---

dmaretskyi:
Confirmed and fixed in 08f27b2b — this one was a real gap in my previous fix, not in the original PR.

Anchoring the scheme stopped `Bearer realm="VerifiablePresentation …"` from matching, but the challenge regex still ran to the end of the header, so `VerifiablePresentation realm="dxos", Bearer challenge="…"` returned Bearer's challenge even though the VP challenge carries none — signing something never issued to that scheme.

The header is now split into segments outside quotes, and only the VP scheme's own auth-params are scanned; the first segment that is not an auth-param starts the next challenge and ends the search. Two tests added: the later-scheme case (with and without a preceding param) and a VP `challenge` that follows another VP param, which the segmenting has to keep finding.

`edge-client` 62 passing (was 60); lint, build and `oxfmt --check` clean.

---

PR #12550, opened 2026-08-12
plugins: XPlugin namespace entrypoints and subpath export validation

wittjosiah:
## Summary

Every plugin package gains an `XPlugin` namespace — `make` plus `meta` — and the `./plugin` subpath is gone. `make` is a lazy stub (`Plugin.lazy(meta, () => import('#plugin'))`), so a host can read a plugin's metadata and decide whether to enable it without loading its body.

Boot graph, measured against `main` at `a3b6ef05` (i.e. **after** the Effect 3 → 4 migration, which took main from 5.73 MB to 4.97 MB on its own):

| | entries | bytes |
| --- | --- | --- |
| `main` @ `a3b6ef05` | 24 | 4.97 MB |
| this branch | 20 | 4,249,977 B (4.05 MB) |
| **delta** | **−4** | **≈ −0.92 MB (−18.5%)** |

Both figures come from `composer-app:check-boot-budget`; main's is the value its own CI run reported, which the script prints to two decimals, so the delta is precise to ~±10 KB.

Nearly all of it came from one finding the migration surfaced: five plugins re-exported their operation handler set from `plugin.ts`, the boot-loaded registration entrypoint, dragging every operation definition and its schemas into the boot graph. Those five had no `./operations` subpath, so the re-export was the only public route to them.

## Entrypoint shape

- `src/XPlugin.ts` — the namespace. Lazy `make`, eager `meta`. This is what hosts import.
- `src/plugin.{ts,tsx,node.ts,workerd.ts}` — the body, reached only through `#plugin` behind the dynamic import.
- `src/XPlugin.testing.ts` — an eager `make` for the `/testing` entrypoints. Storybook runs under vite-dev, where webkit cannot reliably settle the lazy stub's dynamic import; eagerness is required there, but the call shape is not, so `.make()` is the only instantiation pattern repo-wide.

`composer-app` now imports `import * as ChessPlugin from '@dxos/plugin-chess/ChessPlugin'` and calls `ChessPlugin.make()`.

## New ESLint rules

**`dxos-subpath-exports`** validates a package's root barrel against its `exports` map. `dxos-subpath-imports` rewrites a consumer's `import { Drawing } from '@dxos/plugin-illustrator'` into `import * as Drawing from '@dxos/plugin-illustrator/Drawing'` purely from the exports map, so that rewrite is sound only while the two agree — and nothing else checked them together. Eight findings, including `ambiguousNamespace`, where two star paths landing on different modules make ES drop the name from the barrel entirely: no error at the barrel, a link failure in the consumer.

It follows bare `export *` into nested barrels, so `export * from './types'` with the namespaces a directory down satisfies the same contract as declaring them at the root. It engages only for packages that already declare a PascalCase subpath — a package with none has not migrated, and flagging it would report the migration itself.

**`dxos-package-imports`** requires a package's own `imports` aliases over relative paths resolving to the same file (`#apis` had 29 relative imports reaching the module it names). It steps aside for conditional aliases: where `#plugin` resolves per condition, it and a relative path to one branch are different modules, so substituting one for the other would change which module loads.

Both are registered in `.oxlintrc.json` and pass repo-wide. The package README now documents all eleven rules, their real severities, and the three whose scope is narrower than their names.

## Also in this PR

- Schema capabilities converted to `AppCapability.schema(() => import('./schema'))` at 99 call sites across 52 packages. **Measured negative** as a startup lever (−354 B) — the arrays already sat behind the lazy `#plugin` import. Kept for consistency with `commands`.
- `plugin-graph` no longer re-exports `@dxos/app-graph`; consumers import it directly and `app-graph` joins the subpath lint list.
- Fixes an import cycle the migration introduced: routing an intra-plugin sibling import through `#types` pulls in every sibling, and where one imports back, the namespace binding is still in TDZ at evaluation. It fails at runtime with nothing wrong at the type level.
- Fixes eight `exports` entries pointing at `./dist/lib/<name>.mjs` in three platform-neutral packages, whose output lands in `./dist/lib/neutral/`.

## Notes for review

- **Breaking**, covered by `.changeset/plugin-namespace-entrypoints.md`: `./plugin` is removed from every plugin package.
- `scripts/check-cycles.mjs` covers only `packages/{common,core}` and uses madge, which does not resolve the `imports` map — so the plugins tree and every `#` edge are invisible to CI. That gap is why the cycle above reached CI. Extending it is tracked as a follow-up; plugins carry 15 pre-existing cycles, so it needs those fixed or an allowlist first.


wittjosiah:
CI on this branch is blocked by an upstream outage, not by the diff. Flagging it since it has now hit 4 jobs across 3 runs and 2 commits.

**Symptom.** Every job that has to fetch `proto 0.58.2` fails, and every step after it is skipped — no code is compiled, linted, or tested. Two different steps are affected:

- `moonrepo/setup-toolchain@v0` — `Error: Command failed with exit code 22: proto.sh 0.58.2` / `curl: (22) The requested URL returned error: 503`
- `moon setup` — `Setup toolchains with 0 passed, 4 skipped, and 1 failed`, `fail SetupProto(0.58.2)`

Affected so far: `check`, `storybook`, `model-fixture` on `6c5a9faa`; `build` and `check` on `ecc40355`.

Worth noting for anyone triaging: the toolchain cache does **not** insulate a job from this. `check` on `ecc40355` restored the cache successfully (`Cache hit for: moonrepo-toolchain-v3-linux-x64-…`) and still failed, because `moon setup` re-installs proto itself afterwards. So a warm cache only skips the first download, not the second.

**Local verification**, since CI has not yet been able to run anything on this branch:

```
plugin-tldraw:test       Test Files  4 passed (4)   Tests  15 passed (15)
plugin-illustrator:test  Test Files  3 passed (3)   Tests  13 passed (13)
plugin-terra:test        Test Files 20 passed (20)  Tests 141 passed (141)
plugin-voxel:test        Test Files  1 passed (1)   Tests   1 passed (1)
```

Lint and build green on the same four packages; `pnpm format` run before the commit.

The one genuine failure CI did catch — `test (24.11.1)` on `6c5a9faa`, an import cycle introduced by routing intra-plugin sibling imports through the `#types` barrel — is fixed in `ecc40355`.

No action available from this branch. I'll keep re-running the failed jobs as they settle.

---

PR #12557, opened 2026-08-13
assistant-toolkit: make the database typename filter version-tolerant and add expandDepth

dmaretskyi:
Addresses the operation-side half of DX-1154 (MCP feedback). The edge-side half is dxos/edge#877, which has **already merged**.

> Landing order note (superseded): edge#877 landed first. That is harmless — it forwards `expandDepth` only when set, `undefined` is dropped crossing the RPC boundary, and Effect Schema ignores excess properties, so the currently-pinned operation-service quietly ignores it rather than failing. `expandDepth` is therefore dormant in production until edge's `catalog:dxos` pin advances to a dxos commit containing this PR. That pin bump is the remaining follow-up once this lands. (`includeQueues`, the other parameter edge#877 exposes, works today — the operation already supported it.)

## Version-tolerant, dynamic-schema-aware typename filter (issue items 2 & 3)

`database.query` resolved a single schema for the requested typename and filtered on it. `Filter.type` compares the *full* type URI, so that filter matches exactly:

- a static schema yields a **versioned** DXN, so objects written before a schema bump never match — this is why the Bramble sample's `org.dxos.type.task@0.1.0` tasks were invisible while the registry held `@0.2.0`;
- a space-local (stored) schema resolves to an **EID**, so an object stamped with the plain typename DXN never matches — which made every user-defined schema (`example.type.roastLog`) unfilterable.

The filter is now the OR of the version-less typename DXN and every schema registered under that typename, from the space and the static registry alike. An unknown typename returns no results instead of failing the call. The lookup is extracted to `type-filter.ts` and shared with `object-create`, which duplicated it.

## `expandDepth` (issue item 7)

`database.query` and `database.load` accept `expandDepth` (default `0`, maximum `1`, larger values clamped). Referenced objects are inlined in place of their `{ "/": "echo:..." }` envelope, so reading a document with its content, or a task with its assignee, is one call rather than two. A reference that fails to resolve is left as its envelope — a dead or unreplicated link must not fail the read that encountered it.

## Not addressed here

The Bramble exemplar archive (`plugin-onboarding/src/content/exemplar-space.dx.json`) is still a schema generation behind: its tasks are `org.dxos.type.task@0.1.0` and its project container is `org.dxos.type.externalProject@0.1.0`, a type that no longer exists in the sources. `queryObjects` now finds those tasks, but `taskList`/`projectList` filter through the tasks/projects plugins and remain empty. Regenerating via `scripts/build-exemplar-space.ts` emits `task@0.2.0` + `taskSet@0.2.0` and would fix it properly; that is a data change with its own review surface, so it is left out of this PR.

## Testing

```
moon run assistant-toolkit:build      # clean
moon run assistant-toolkit:lint       # clean
moon run assistant-toolkit:test -- src/skills/database    # 13 files, all passing
DX_RUN_MODEL_FIXTURE_TESTS=1 moon run '#model-fixture:test'   # 181 tasks, all passing
```

New cases in `query.test.ts` / `load.test.ts`: the typename filter carries an unversioned branch, an unknown typename returns `[]`, and `expandDepth` inlines one level and stops there (mutation-checked — raising `MAX_EXPAND_DEPTH` to 2 fails it).

The database-skill model fixtures are regenerated: adding `expandDepth` changes the tool definitions sent to the model, so the memoized conversations no longer matched. Regenerated per-file (never with `-t`, since the shared ID PRNG makes per-test regeneration corrupt the suite) and verified in replay mode across every model-fixture suite.

https://claude.ai/code/session_01FSVxiLBNLQ9DuYDSaHJRuR

## Summary by CodeRabbit

- **New Features**
  - Database queries now match all versions of a typename, including user-defined types.
  - Query and load results support optional reference expansion up to one level, while unresolved references remain available as URI envelopes.

- **Bug Fixes**
  - Queries for unknown typenames now return no results instead of failing.
  - Duplicate referenced objects are handled efficiently during expansion.

- **Privacy**
  - Stored model interaction fixtures no longer retain provider request or response headers.


dmaretskyi:
Addressed both review findings in 58470f1a — one fixed, one not constructible.

**`load.test.ts` depth clamp — fixed.** The finding was correct: `Organization` has no nested reference, so an implementation that expanded past the clamp produced an identical result and the assertion couldn't fail. The test now builds a three-link chain (`head → middle → tail`) and asserts that `expandDepth: 5` inlines `head.next` while leaving `head.next.next` as its `{ "/": "echo:..." }` envelope. Verified it has teeth: raising `MAX_EXPAND_DEPTH` to 2 fails it.

**`query.test.ts` two-version fixture — not possible in this harness.** I tried exactly the suggested fixture first. The runtime schema registry is keyed by *typename*, not typename+version, so registering two versions of one typename silently keeps one:

```
types: [WidgetV1 /* @0.1.0 */, WidgetV2 /* @0.2.0 */]
→ REGISTERED_WIDGET_SCHEMAS 1 [ '0.1.0' ]
```

and adding an object of the dropped version then fails with `Schema not registered`. So a space holding objects of two versions of one type — which is exactly the production situation this PR fixes, produced by importing an archive written before a schema bump — cannot be constructed through the registry-backed test layer at all.

Given that, the AST assertion is the strongest available proof of the mechanism: it pins that the filter carries an *unversioned* branch, which is precisely what makes matching version-independent (`compareTypenameStrings` short-circuits to name-only equality when either side omits the version).

Worth noting the registry's typename-keying is arguably its own bug — it means a space cannot hold two versions of a type at runtime even though the storage format and `Filter` semantics both support it. Out of scope here.

---

dmaretskyi:
On the fixture-metadata finding: the observation is **correct on the facts**, but it is pre-existing repo-wide behaviour of the fixture generator rather than something this PR introduces, so I'm not blocking on it here. Verified rather than assumed:

**No credential is exposed.** `x-api-key` is stored as the literal `"<redacted>"`, and there is no `authorization` header and no `sk-ant-` string anywhere in the store.

**Real non-credential metadata is present**, as described — in the response headers block:

```
anthropic-organization-id: e9974f42-…
anthropic-workspace-id:    wrkspc_…
request-id:                req_011Cdz…
anthropic-ratelimit-*:     limits, remaining counts, reset timestamps
cf-ray / traceresponse / traceparent / b3
```

**It is already on `main`, in fixtures this PR never touches.** Same organization UUID, in suites outside this change:

```
.store/conversations/packages_core_compute_agent-runtime_src_agent-service_AgentService/…
  anthropic-organization-id PRESENT   (org value on main: e9974f42-…)
.store/conversations/packages_core_compute_assistant-toolkit_src_skills_memory_skill/…
  anthropic-organization-id PRESENT   (org value on main: e9974f42-…)
```

All 101 files in the database-skill directory carry these fields — the 54 that predate this PR as well as the 47 it regenerates. So the exposure is unchanged by merging: the generator has always recorded the raw response headers.

I agree it should be fixed, and the fix is the one suggested — sanitize the writer in `packages/core/compute/ai/src/testing/model-fixture/LanguageModelFixture.ts` so only fixture-required fields are persisted. But that is a shared-test-harness change whose blast radius is *every* model-fixture suite in the repo (`ai`, `agent-runtime`, `assistant`, `assistant-toolkit`, `plugin-markdown`, `plugin-magazine`, `plugin-commerce`, `plugin-assistant`, `assistant-e2e`), and correcting the already-committed values means regenerating all of them against the live API — plus a history rewrite if the org/workspace IDs are considered sensitive, since they are already in `main`'s history. That does not belong in a DX-1154 MCP fix, and scoping it is the repo owners' call.

Two things worth separating for whoever picks it up:

1. **Stop recording it going forward** — sanitize the writer, regenerate fixtures as they naturally churn. Low risk, no history rewrite.
2. **Scrub what is already committed** — regenerate every suite at once, and decide whether `main`'s history needs rewriting. Higher cost, and only warranted if tenant IDs are treated as sensitive rather than merely noisy.

Happy to do (1) as its own PR if wanted. Flagging rather than silently deferring, since it was raised as Major.

CI is green on `58470f1a` (`Check` and `Model Fixture` both pass).

---

dmaretskyi:
1 + 2

dmaretskyi:
Both done in 815745fe.

**1 — stop recording it.** `saveFixtureConversation` now empties the provider's request and response headers before encoding. Headers are emptied rather than removed so each part keeps its shape, and `method`/`url`/`status` stay: they describe the call, not the caller. Three unit tests cover it (redaction, shape preservation, parts without headers untouched).

**2 — scrub what is already committed.** All 124 fixtures that carried headers are rewritten in place, across every suite, not just this PR's. The store hashes over `parameters` + `prompt` only — the response is not part of the match key — so lookups are unaffected by construction, and that is confirmed empirically below.

Residual check over the whole store:

```
anthropic-organization-id -> 0 files      x-api-key    -> 0 files
anthropic-workspace-id    -> 0 files      ratelimit    -> 0 files
request-id                -> 0 files      cf-ray       -> 0 files
traceparent               -> 0 files
files with non-empty headers: 0
```

(`b3` still greps in 3 files — those are substring hits inside Anthropic message ids like `msg_011CdzaqKmwDeguwb3kD7BJH` and base64 content, not the trace header.)

Verification — no regeneration needed, so no live API calls and no fixture churn beyond the header removal:

```
moon run ai:build / ai:lint                                    clean
moon run ai:test -- .../LanguageModelFixture.test.ts           12 passed
DX_RUN_MODEL_FIXTURE_TESTS=1 moon run '#model-fixture:test'    181 tasks, all passing
```

Two notes on scope:

- This rides on this PR rather than a separate one because the harness assigns one branch per session and I am not permitted to create another. It is a self-contained commit (`ai` package + the store) if you would rather it were split out before merge.
- **History is not rewritten.** The organization and workspace ids remain in `main`'s history from earlier commits. Removing them there means a force-push over shared history, which is your call, not mine — and it is only worth doing if those ids are treated as sensitive rather than merely noisy. Nothing in this PR depends on that decision.

---

dmaretskyi:
The fixture-sanitization work is **withdrawn from this PR** — #12576 landed the same thing on main first, with a better rule. Resolved in ffd14aef (revert) → 0c56aeb1 (merge main) → 3f8aa342.

**On the Major finding — it was correct, and it is the reason I took main's version rather than mine.** My `deepMapValues` walked the whole part tree redacting any object-valued `headers`, so a tool result carrying application data under that key would have been rewritten too. Main's `stripTransportMetadata` drops only the top-level `request`/`response` envelopes, which cannot touch nested tool output. Theirs is also stricter on the security goal: it removes the whole envelope where mine merely emptied `headers`, leaving `method`/`url`/`status` behind. So the finding is moot by revert rather than by patch, and the narrower implementation is the one in the tree. The nitpick on comment length goes with it.

Reconciliation, since the two changes overlapped in three places:

- **The sanitizer** — mine reverted; main's `stripTransportMetadata` is the single mechanism on the branch.
- **The fixtures** — 77 files on this branch still carried transport envelopes after the merge (the 47 this PR generated, which main's codemod never saw, plus others). Ran main's own `tools/codemods/strip-model-fixture-transport.mjs` over the store: 154 envelopes across 77 files. Now zero, on every sensitive key.
- **`definitions.ts`** — main fixed the query tool's `in` example while this branch adds `expandDepth`. Git merged both cleanly, but the combination changes the tool definitions sent to the model, so 17 skill tests failed with `No memoized conversation found` against either side's recordings. Regenerated against the merged schema; the new records pass through main's writer, so they carry no envelopes.

Verification on the merged tree:

```
moon run assistant-toolkit:build / lint / ai:build            clean
moon run assistant-toolkit:test -- .../operations             22 passed
DX_RUN_MODEL_FIXTURE_TESTS=1 moon run '#model-fixture:test'   181 tasks, all passing
fixtures carrying transport envelopes                          0
```

The branch diff against main is now just the DX-1154 change — 9 files, `expandDepth` and the version-tolerant typename filter. One residual worth noting for whoever owns #12576: `stripTransportMetadata` has no unit test on main. I have not added one, since retargeting my tests at a function another PR just landed belongs with that change, not this one — happy to send it separately.

---

PR #12561, opened 2026-08-13
compute-runtime: replace TriggerDispatcher polling with a reactive query

dmaretskyi:
## Summary

Part of the `memory-usage` project (`.agents/projects/memory-usage/`), Phase 4 "idle churn" — separate PR from the in-flight #12505 (Phase 3 code residency), covering sites **S1-S4**.

- **S1 (fixed).** `TriggerDispatcher` re-queried the database for the full `Trigger` list on every 1 Hz natural-time tick, even with no triggers defined — the dominant contributor to an idle tab's SQL churn from this loop. `start()` now subscribes once to a live `Filter.type(Trigger)` query; `refreshTriggers`/`_fetchTriggers` read the subscription's cached results instead of re-querying. The tick itself still runs every second for cron due-time checks and feed/subscription trigger kinds — a full "no timer while idle" scheduler is a larger, riskier change left as a follow-up.
- **S2 (fixed).** `FeedHandle.beginPolling` polled at a fixed 1 Hz per open feed. No server-push alternative exists (`FeedService` has no subscribe/watch RPC — real fix deferred to `feed-live-objects`), so this adds no-change backoff instead: the poll delay doubles (capped at 30s) on ticks that find nothing new, and resets to 1s the instant something changes.
- **S3 (fixed).** Space sync-state polling (`DatabaseImpl.subscribeToSyncState`) issued a real per-namespace `FeedService.getSyncState` RPC (SQLite `COUNT(*)` server-side) every 2s per open space, unconditionally. Same no-change backoff treatment, capped at 15s.
- **S4 (investigated, no change).** `QueryService`'s `matchesHint` already filters invalidation hints precisely and is covered by extensive regression tests; tightening it further looked speculative. The actual amplification traced back to S1: a one-shot per-tick query bypasses hint-matching entirely (always dirty on first run). With S1-S3 no longer generating spurious ticks/RPCs at idle, S4's observed SQL fan-out should fall away as a consequence.

Decisions and rationale recorded in `.agents/projects/memory-usage/TASKS.md` (Phase 4 sites section).

## Test plan

- [x] `moon run compute-runtime:build` — full dependency graph builds clean
- [x] `moon run compute-runtime:test -- src/triggers/trigger-dispatcher.test.ts` — 33/33 passing (incl. natural-time-control start/stop)
- [x] `moon run compute-runtime:test -- src/TriggerMonitor.test.ts` — 7/7 passing
- [x] `moon run echo-client:build`
- [x] `moon run echo-client:test` — full suite, 522 passing / 7 expected-fail / 12 skipped / 2 todo
- [x] `moon run compute-runtime:lint echo-client:lint`
- [x] `pnpm format`
- [ ] Manual idle-churn verification with `scripts/memory/soak.mjs` (tracked in TASKS.md as a follow-up once this lands)

---
_Generated by [Claude Code](https://claude.ai/code/session_01PwgNG6Ehcs8rC75PoVZSiW)_

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary background activity by reusing trigger results and monitoring updates reactively.
  * Added adaptive polling for feeds and synchronization, slowing checks when no changes are detected and responding faster when updates occur.
  * Improved resource usage during idle periods while preserving timely updates when data changes.

### DIFF
--- a/.agents/projects/memory-usage/TASKS.md
+++ b/.agents/projects/memory-usage/TASKS.md
@@ -76,17 +76,19 @@ fixed the same way.
       defaults to 1 s (~line 300), `_startNaturalTimeProcessing` repeats
       `invokeScheduledTriggers` on `Schedule.fixed` (~line 921), and each tick
       also runs an ECHO `Filter.type(Trigger)` query. Started by plugin-routine,
-      a core plugin, so every tab pays it. **Fixed (R1, reactive query):**
+      a core plugin, so every tab pays it. **Partially fixed, toward R1:**
       `start()` now subscribes once to a live `Filter.type(Trigger)` query
-      (`_subscribeToTriggers`); `refreshTriggers`/`_fetchTriggers` read the
+      (`#subscribeToTriggers`); `refreshTriggers`/`_fetchTriggers` read the
       subscription's cached `results` instead of re-querying the database, so
-      the 1 Hz tick no longer issues SQL when the trigger set hasn't changed.
-      The tick itself still runs every second (cron due-time checks and
-      feed/subscription trigger kinds are unchanged) — a full "sleep until
-      next due time, no timer while idle" scheduler (closing S1 completely per
-      R1's description) is a larger, riskier change to feed/subscription
-      trigger semantics and is left as a follow-up if further gains are
-      wanted.
+      the trigger-list lookup no longer issues SQL when the trigger set hasn't
+      changed. This is only the "subscription" half of R1 — the tick itself
+      still fires every second (`TriggerDispatcher.invokeScheduledTriggers`
+      still queries feed and subscription trigger data on each tick, and cron
+      due-time checks still run on the fixed schedule). The full R1 — no timer
+      while the working set is empty, sleep to the next due time otherwise —
+      would additionally need to skip firing altogether when idle, a larger,
+      riskier change to feed/subscription trigger semantics, left as a
+      follow-up if further gains are wanted.
 - [x] **S2. Every subscribed feed polls at 1 Hz.**
       `echo-client/src/feed/feed-handle.ts` (`POLLING_INTERVAL`,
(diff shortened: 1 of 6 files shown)


dmaretskyi:
Addressed all actionable findings in `84a084e7`:

- **Subscription/refresh-fiber leak on crash** (major): `stop()` cleared the trigger-query subscription, but the natural-time-processing failure path set `_running = false` directly and never called it — so a later `stop()` would no-op and leak the subscription. Now both paths go through a shared `#teardownTriggerSubscription`.
- **Refresh fiber racing shutdown** (major): the subscription callback's forked `refreshTriggers` fiber is now tracked and interrupted in that same teardown, so one already in flight can't complete after `_triggerQuery` is cleared and repopulate state via the one-shot fallback query.
- **Backoff on failed reads** (major, both S2/S3): a failed poll no longer counts as "no change" — it resets to the fast interval, since a failure can't tell us the feed/backlog is actually quiet.
- **`#private` fields** (nit): converted the new fields in both files.
- **TASKS.md wording**: fixed the R1/R2 label mismatch and the "1 Hz tick" overclaim (the tick itself still queries feed/subscription trigger data each cycle — only the trigger-list lookup stopped doing that).

One deviation from the suggested fix: for the duplicate-poll-loop race in `FeedHandle.beginPolling` (unsubscribe + re-subscribe while a refresh is in flight), I added a lightweight polling-generation counter rather than the suggested `Effect.sleep`/`Effect.gen` rewrite of the whole polling loop. That race pre-dates this PR and isn't specific to the backoff change; a full rewrite of this file's async internals (which has no existing test coverage for timer/polling behavior) felt like a larger, separately-scoped change than this PR's actual goal. The generation guard fixes the concrete race with a minimal diff.

Rebuilt + retested `compute-runtime` (40/40) and `echo-client` (522/543, rest skipped/expected-fail) after these changes; both lint clean.

---

PR #12575, opened 2026-08-14
schema: FeedAnnotation names the property holding the feed reference

richburdon:
## Problem

`FeedAnnotation` was a bare `boolean`. `true` said "this type owns a feed" but not *which property* held it, so every consumer hardcoded `.feed` — carrying a `TODO(burdon): Should reference the property name?` on the annotation id.

## Changes

`FeedAnnotation` now carries `{ property: string }`, plus two helpers in `@dxos/schema`:

- `getFeedRef(obj)` — resolves the feed through the property named by the annotation.
- `isFeedOwnerSchema(schema)` — predicate for the registry sweep.

All three consumers route through them, deleting two hand-rolled local helpers and the hardcoded property access:

| | before | after |
|---|---|---|
| `sync-automation.ts` | local `getFeedRef` reading `obj.feed` + local `hasFeedAnnotation` | `getFeedRef(target)` |
| `plugin-routine` `selectFeed` | `(obj as any).feed` | `getFeedRef(obj)` |
| `AgentProperties.tsx` | inline `annotation.value === true` filter | `isFeedOwnerSchema` |

The three setters (`Mailbox`, `Calendar`, magazine `Subscription`) now pass `{ property: 'feed' }`. No legacy `true` branch — every setter and reader is in-repo and updated here; a stored schema still carrying `true` degrades to "not discovered as a feed owner" rather than crashing.

## Scope change during review

This PR originally also added `Ref.byAnnotation(annotationId)` to `@dxos/echo` — a reference schema constrained by an annotation on the target's schema. @dmaretskyi pushed back, and he was right on both counts:

1. **Type level** — annotations don't participate in the type system, so `Ref.byAnnotation(X)` produced exactly the same TS type as `Ref.Ref(Obj.Unknown)`.
2. **Runtime** — measured against a real database, the synchronous check only fired for a target already resident in the working set. In a cold context — the edge-function operation boundary that motivated it — the reference was accepted, i.e. identical behaviour to `Ref.Ref(Obj.Unknown)`.

A type-level fix was reachable (a required phantom brand added to `Entity.Properties`' omit list does discriminate, verified by prototype), but that is a change to ECHO's core type plumbing. Per his call, **every echo change has been reverted** — this branch now touches zero files under `packages/core/echo`. The `Ref.byAnnotation` discussion is preserved in the comments below.

## Tests

`packages/sdk/schema/src/annotations/feed.test.ts` — `isFeedOwnerSchema` discriminates; `getFeedRef` resolves the *annotated* property, covering a type that holds its feed under `entries` (the case a hardcoded `.feed` cannot serve); returns undefined for an unannotated type.

Green on merged main: `schema` (51), `plugin-inbox` (213), `plugin-assistant` (168), `plugin-routine` (65), `plugin-magazine` (58), `assistant-toolkit` (70); builds + lint + oxfmt clean.

## Summary by CodeRabbit

* **New Features**
  * Feed-linked schemas now identify their feed reference property explicitly.
  * Added support for resolving feed references across custom property names.
* **Bug Fixes**
  * Improved feed detection and reference handling across calendar, mailbox, subscription, automation, assistant, and routine features.
  * Invalid or unavailable feed references are now handled safely.
* **Tests**
  * Added coverage for feed-owner detection, custom feed properties, and missing annotations.


.changeset/ref-by-annotation.md
```diff
@@ -0,0 +1,7 @@
+---
+'@dxos/echo': minor
+---
+
+Add `Ref.byAnnotation(annotationId)`, a reference schema constrained by an annotation on the target's schema rather than by a concrete typename, so a schema can express "a ref to any type carrying X". The constraint is enforced where references are validated (operation input decoding) and survives a JSON schema round trip.
+
+BREAKING: `FeedAnnotation` (`@dxos/schema`) now carries `{ property: string }` naming the property that holds the feed reference, instead of a bare `true`. Use the new `getFeedRef(obj)` and `isFeedOwnerSchema(schema)` helpers rather than reading `.feed` directly.
```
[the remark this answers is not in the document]
richburdon: Fixed in fc8df7c — added `'@dxos/plugin-inbox': patch`. Matches repo precedent (e.g. `capability-dependency-activation.md` names `@dxos/app-framework` + `@dxos/plugin-markdown`), so the Group B feed-consumer migration lands in that group's changelog rather than only rippling as a silent patch bump.

richburdon:
Addressed the CodeRabbit review in fc8df7c:

1. **Group B changeset entry** — added `'@dxos/plugin-inbox': patch`. Valid; matches repo precedent.
2. **Comment verbosity** — valid per the repo's one-load-bearing-clause rule. Condensed all six sites. The public JSDoc on `Ref.byAnnotation` (`src/Ref.ts`) is left fuller deliberately — it documents the best-effort validation contract, and `AGENTS.md` requires JSDoc on public functions.
3. **`Annotation.make` migration for `FeedAnnotation`** — agreed, out of scope here; the sibling `queue.ts` carries the same pre-existing TODO and both should move together.

Also improved the failure message while in there. `Schema.refine` only takes a static string, so a rejection couldn't say *which* type was referenced; switched to `Schema.makeFilter`, which computes the message from the value:

```
Referenced object (com.example.type.contact) does not carry `com.example.annotation.feed`.
  at ["owner"]
```

`echo` 578 passed, `schema` 52 passed, lint + oxfmt clean.

dmaretskyi (commented):
annotations don't participate in the type system 
how's Ref.byAnnotation(X) better then Ref.Ref(Obj.Unknown)
we don't have a way to express "object that holds annotation X" in the type system

richburdon:
@dmaretskyi you're right on the type level, and measuring it showed the runtime claim was weaker than I wrote. Both points, concretely:

**1. Type level — correct, `Ref.byAnnotation(X)` gives exactly the same TS type as `Ref.Ref(Obj.Unknown)`.** It does not narrow anything at compile time. I did not claim otherwise in the code, but the PR description oversold it.

**2. Runtime — it does bite, but only when the target is already in the working set.** I measured this against a real database rather than argue it (`packages/core/echo/echo-client-e2e/src/ref-constraint.test.ts`, pushed in ff5ba58):

| context | ref to `Contact` (unannotated) |
|---|---|
| same db, target in working set | **rejected** — `Referenced object (com.example.type.contact) does not carry ...` |
| reopened peer, target not loaded | **accepted** — check cannot see it |
| reopened peer, after `ref.load()` | **rejected** |

The cold-context row is the important one: it's the edge-function boundary this was motivated by (`protocol.ts:106` runs `decodeUnknownSync(Schema.toType(input))`, and `_resolveSync` only returns objects already in `db.getObjectById`). So in exactly that case it behaves identically to `Ref.Ref(Obj.Unknown)`. That gap is now an assertion, not a comment.

What's actually left over `Ref.Ref(Obj.Unknown)` is narrower than the PR body claims: the constraint is *declared in the schema* and reaches JSON Schema (`reference.annotation`), so tooling can read it; and it is enforced whenever the target happens to be resident. That may not be worth a new API.

Options, in the order I'd rank them:

1. **Split the PR** — land the uncontested half (`FeedAnnotation` widened to `{ property }` + `getFeedRef`/`isFeedOwnerSchema`, which removes the hardcoded `.feed` from three consumers), drop `Ref.byAnnotation`, and let D6 use `Ref.Ref(Obj.Unknown)` + `getFeedRef`. **My recommendation** unless you want 2 or 3.
2. **Make it reliable** — enforce in the decode transformation with a real load, which requires the operation input boundary to become effectful (`decodeEffect` rather than `decodeUnknownSync`). Cross-cutting into `compute-runtime`, and it makes every constrained-ref input do IO at the edge. Your call whether that's wanted; I did not want to smuggle it in here.
3. **Make annotations participate in the type system** — have `FeedAnnotation.set()` intersect a phantom brand so `Ref.byAnnotation` returns a branded `RefSchema` and a wrong `Ref.make(contact)` fails to compile. This is the direct answer to your comment, but the brand has to survive `Type.makeObject<Mailbox>(...)`, whose instance type is the explicit class parameter — so I don't think it works without changing how types are declared. Happy to prototype if you think it's reachable.
4. Keep as-is on the schema-declaration argument alone.

Tell me which and I'll do it — I'm not landing this over an open objection.

richburdon:
@dmaretskyi correction to option 3 in my last comment — I said the brand probably can't survive `Type.makeObject<Mailbox>(...)` because the instance type is the explicit class parameter. **That was wrong.** `ObjClass<Self, T, Fields>` declares `new (_: never): T & OfKind<Object>` where `T = Schema.Schema.Type<_Schema>`, so `Self` only names the static side and the instance type comes from the *schema*. A brand added by the annotation setter does reach the instance type.

I prototyped it. Two real findings:

1. **A required brand discriminates and works.** `setFeed()` returning `S & { readonly Type: S['Type'] & FeedOwner }` gives `Type.InstanceType<typeof Mailbox> extends FeedOwner`, and `Ref.make(contact)` into a `Ref<Obj.Unknown & FeedOwner>` field is a compile error — which is exactly "object that holds annotation X", expressed in the type system.
2. **But it leaks into construction.** With a required brand, `Obj.make(Mailbox, { name: 'inbox' })` fails: `Argument of type '{ name: string; }' is not assignable to parameter of type 'MakeProps<typeof Mailbox>'`. An *optional* brand fixes construction but stops discriminating entirely (structurally satisfied by anything), so it's useless.

The fix is small but it's in core plumbing: `Entity.Properties<T> = Omit<T, 'id' | KindId | Relation.Source | Relation.Target>` is an explicit omit list, so a brand key has to be added there — same mechanism `KindId` already uses.

So option 3 is reachable, and it's the only option that answers your objection rather than working around it. It needs: a phantom brand key omitted by `Entity.Properties`, a branded annotation-helper variant, and `Ref.byAnnotation` taking the helper instead of a bare id so it can recover the brand type.

That's a change to ECHO's core type plumbing on a PR you've already objected to, so I'm not doing it unilaterally. Revised ranking:

1. **Option 3** if you want annotations to participate in the type system — I have the prototype and can turn it into a real change.
2. **Option 1 (split)** — land the `FeedAnnotation` `{ property }` widening + `getFeedRef`/`isFeedOwnerSchema` (removes hardcoded `.feed` from three consumers), drop `Ref.byAnnotation`. Safe default if you don't want core type changes now.

Still not landing until you pick. Everything else is green.

dmaretskyi:
drop Ref.byAnnotation and any core type-system changes, keep the app-level changes about FeedAnnotation

richburdon:
@dmaretskyi done — `Ref.byAnnotation` and every core type change are gone. This branch now touches **zero files under `packages/core/echo`**:

\`\`\`
$ git diff origin/main --stat -- packages/core/echo/
(empty)
\`\`\`

What's left is only the app-level half:

- `FeedAnnotation` carries `{ property: string }` instead of a bare `true`, resolving the existing `TODO(burdon): Should reference the property name?`.
- `getFeedRef(obj)` / `isFeedOwnerSchema(schema)` in `@dxos/schema`.
- Three consumers routed through them — deletes two hand-rolled local helpers and the hardcoded `.feed` / `(obj as any).feed` accesses.

Also merged `main` (it had moved on substantially, including the scan/enrich/process reshuffle in plugin-inbox) and rewrote the PR title and description to match the reduced scope. Changeset is now `@dxos/schema: minor` + `@dxos/plugin-inbox: patch`.

Green on merged main: `schema` 51, `plugin-inbox` 213, `plugin-assistant` 168, `plugin-routine` 65, `plugin-magazine` 58, `assistant-toolkit` 70; builds + lint + oxfmt clean. Waiting on CI, then I'll land it.

packages/sdk/schema/src/annotations/feed.ts
```diff
@@ -2,12 +2,44 @@
...
-// TODO(burdon): Should reference the property name?
 export const FeedAnnotationId = '@dxos/schema/annotation/Feed';
-export const FeedAnnotation = createAnnotationHelper<boolean>(FeedAnnotationId);
+
+export type FeedAnnotationValue = {
+  /** Name of the property holding the canonical `Ref<Feed>`. */
+  property: string;
+};
+
+export const FeedAnnotation = createAnnotationHelper<FeedAnnotationValue>(FeedAnnotationId);
```
dmaretskyi: use new Annotation.make api, add an .mdl rule to prevent sdk and app code from using echo/internal whenever a sutable public api alternative exists

packages/sdk/schema/src/annotations/feed.ts
```diff
@@ -2,12 +2,44 @@
...
 
 /**
- * Identifies a schema as an object with a canonical feed reference
- * (i.e., a reference to a feed in the feed property).
+ * Identifies a schema as an object that owns a canonical feed, naming the property that holds the
+ * `Ref<Feed>`. Carrying the property name (rather than a bare `true`) is what lets a generic host
+ * resolve the feed without hardcoding `.feed` — see {@link getFeedRef}.
  */
-// TODO(burdon): Should reference the property name?
 export const FeedAnnotationId = '@dxos/schema/annotation/Feed';
```
dmaretskyi: does this need to be exported?

dmaretskyi:
do you have DX_ANTHROPIC_API key configured to regenerate model fixutres?

richburdon:
@dmaretskyi no — no `DX_ANTHROPIC_API_KEY` (or `ANTHROPIC_API_KEY`) in this environment; only `ANTHROPIC_BASE_URL` is set, and there's no root `.env`. So I can't regenerate the fixtures from here.

Worth knowing before anyone does regenerate them: **this failure is already on `main`, not introduced by this PR.**

| run | commit | Model Fixture |
|---|---|---|
| [31769221920](https://github.com/dxos/dxos/actions/runs/31769221920) | `4804da07` (#12574, on main) | ❌ failure |
| [31768266961](https://github.com/dxos/dxos/actions/runs/31768266961) | `f0480626` (main) | ✅ success |
| [31763978193](https://github.com/dxos/dxos/actions/runs/31763978193) | `fc8df7c7` (this PR, pre-merge) | ✅ success |

It went red on main at 04:13 with #12574; my run at 04:45 inherited it by merging main. Identical signature in both — `assistant-toolkit/src/skills/database/skill.test.ts`, 15 tests, memoized-dynamic index drift:

```
-  "obj": "echo://<memoized-dynamic-0>/<memoized-dynamic-1>"
+  "obj": "echo://<memoized-dynamic-0>/<memoized-dynamic-2>"
+  "result": "EntityNotFoundError: Entity not found: {\"uri\":\"echo:///<memoized-dynamic-2>\"}"
```

An extra entity is being allocated before the one the fixture recorded, shifting every subsequent dynamic index — consistent with #12574's aggregate/group-key work changing allocation order in the database skill path, not with anything in this branch (which now touches zero files under `packages/core/echo`).

Two ways forward, your call:

1. **Regenerate on main** as a fix for #12574, and I rebase onto it. Cleanest — the fixtures are wrong for main, independent of this PR, and whoever regenerates needs the key anyway.
2. **You (or someone with the key) regenerate on this branch** and I'll carry it. I can drive everything else, just not the model calls.

I'd take 1. Separately, both your review comments are addressed in d71f6ae — `FeedAnnotation` now uses `Annotation.make` (the `@dxos/echo/internal` import and the separately-exported id are gone; the id had no consumer outside the module), plus a `no-echo-internal-in-sdk` rule in `non-negotiables.mdl` flagging sdk/plugin/app imports of `@dxos/echo/internal` where a public equivalent exists.

dmaretskyi:
shouldn't the fixture account only for entities that are visible in the recorded fixture, so an allocation not observed in conversation wouldn't cause drift? can you show exactly what happened and where that extra entity appears?

richburdon:
@dmaretskyi you're right about the design, and my "extra entity allocated before the recorded one" was wrong — correcting it, with the actual mechanism. I reproduced locally (replay needs no key, just `DX_RUN_MODEL_FIXTURE_TESTS=1 VITEST_ENV=node`).

**1. The fixture already does what you describe.** `collectDynamicValues` ([LanguageModelFixture.ts:177](../blob/main/packages/core/compute/ai/src/testing/model-fixture/LanguageModelFixture.ts#L177)) scans only the canonical serialization of *the prompt* and numbers distinct tokens by first appearance:

```ts
const canonical = jsonStableStringify(prompt) ?? '';
for (const match of canonical.matchAll(matcher)) { ...ordered.push(token) }
```

So an allocation never observed in the conversation gets no placeholder and cannot drift anything. Only a change in the **set or order of ids visible in the prompt** shifts indices.

**2. The dominant cause isn't placeholder drift at all — it's a tool-schema shape change.** #12574 changed the aggregate group kind in [`Aggregate.ts:20`](../blob/main/packages/core/echo/echo/src/Aggregate.ts#L20):

```ts
- | { kind: 'group'; property: string }
+ | { kind: 'group'; properties: readonly [string, ...string[]] }
```

That struct is serialized into the Query tool's JSON schema in the prompt, so every fixture carrying that tool now mismatches:

```
-  "property": { "type": "string" }
+  "properties": { "items": {"type":"string"}, "minItems": 1,
+                  "prefixItems": [{"type":"string"}], "type": "array" }
```

For `query: search for objects`, run in isolation, **that is the entire diff** — zero placeholder lines. Across the whole file: 762 diff lines, 128 of them `property`/`properties`.

**3. The placeholder shift is a second-order symptom, not an independent cause.** 74 of the 762 lines, and the shift is uniformly +1 (`<1>`→`<2>`, `<3>`→`<4>`) paired with a *new* failure:

```
-  "obj": "echo://<memoized-dynamic-0>/<memoized-dynamic-1>"
+  "obj": "echo://<memoized-dynamic-0>/<memoized-dynamic-2>"
+  "result": "EntityNotFoundError: Entity not found: {\"uri\":\"echo:///<memoized-dynamic-2>\"}"
```

An id that was visible in the recorded prompt (`"id": "<1>"`, `"@uri": ".../<1>"`) is absent on replay, and a different one takes its place — consistent with the run diverging once the schema mismatch breaks the fixture match, rather than with an unobserved allocation. I'll flag the limit of what I've proven: I can't fully separate cause from effect on this mode without regenerating, and I don't have the key. But `query: search for objects` is a clean case whose entire diff is the schema change, so the fixtures need regenerating for #12574 either way.

Nothing here is reachable from this branch — it touches zero files under `packages/core/echo`, and Model Fixture was green on it at `fc8df7c7` before I merged main.

Repro:
```bash
DX_RUN_MODEL_FIXTURE_TESTS=1 VITEST_ENV=node moon run assistant-toolkit:test -- src/skills/database/skill.test.ts -t "query: search for objects"
```

dmaretskyi:
good to land

PR #12585, opened 2026-08-14
client-services, edge-client: defer edge networking until the worker has booted

dmaretskyi:
Rebased onto `main` now that #12564 has landed — base is `main`, single commit.

## Why

The worker runs wa-sqlite in-process, so the edge dial, its auth-header request, and the replication behind it all contend with the boot RPCs the tab is blocked on. This is an *ordering* problem, not just a volume one: in the original capture the edge socket connected at **+24.2 s, immediately after the worker thread freed at +23.7 s** — the handshake was starved by work queued ahead of it, and the client reported a connect timeout at +17.4 s.

Gating outbound work removes that contention regardless of how slow local hydration is. It composes with, and does not replace, the hydration work still open as Phase 2.

## What

`EdgeClient` gains `deferConnect` and an idempotent `startNetworking()`. With `deferConnect` set, `open()` no longer dials — the owner decides when connecting is safe. The *policy* deliberately does not live in `EdgeClient`: `ClientServicesHost` owns it and schedules the dial a short grace period (300 ms) after boot.

The trigger is anchored to the worker actually booting: `worker-runtime` calls `startNetworking()` once its whole start sequence drains, which is strictly later than the stack open. Hosts with no external boot signal (node, tests) fall back to starting networking themselves — but on a microtask, and only if nothing else signalled first, so the fallback cannot pre-empt the worker's anchor and latch the guard. (That last part came from a CodeRabbit finding on the draft: an unconditional fallback would dial pre-handshake on a boot slower than 300 ms and silently turn the worker's call into a no-op, defeating the entire gate.)

## Only the dial needed gating

I checked each outbound path rather than adding three gates:

- **Subduction** already returns early while the socket is not `CONNECTED` and resumes from `_handleReconnect` — no change needed.
- **Feed sync** re-schedules its poll and push from `onReconnected` — mostly no change needed.
- The one real gap was `FeedSyncer._open`'s **unconditional** initial `pollTask.schedule()`, which would park on the send-ready trigger until a socket appeared. It now runs only when the socket is already up; `onReconnected` schedules exactly the same work otherwise, so nothing is lost.

## Notes

- `EdgeConnection` had **three** implementors, not one — `EdgeClient`, `TestEdgeConnection`, and an inline stub in `feed-syncer.test.ts`. Adding an interface member broke `messaging:build` and `client-services:build` until all three were updated.
- The `feed-syncer` hunk was re-derived against #12580/#12584's streaming rewrite; the original draft no longer applied there.
- Deletes `.agents/projects/client-startup-timeout/defer-networking.patch` — it existed only to carry this diff between PRs, and is now stale. Phase 3 in `TASKS.md` is marked done, with the browser-verification item left open.
- Reconnect behaviour is untouched; this gates the first attempt only.
- The 300 ms constant is a grace period *after* boot completion, not a substitute for it — if boot is slow, the gate still holds until boot finishes.

## Testing

`edge-client` 68/68 (4 skipped), `messaging` 20/20 (1 skipped), `client-services` 165/165 (3 skipped) — all re-run after the rebase. `oxfmt --check` clean, no casts introduced.

**Not verified end-to-end in the browser.** The acceptance check is that in a fresh single-`i` `app.log`, the first `received subduction batch` / edge-connect timestamps strictly after `worker-session opened`. Worth confirming on a document-heavy profile before merge, since that is the scenario this targets.

part of the `client-startup-timeout` project (Phase 3; spec in `.agents/projects/client-startup-timeout/DESIGN.md`)

## Summary by CodeRabbit

* **New Features**
  * Deferred network startup until application and worker initialization completes.
  * Added automatic or manually controlled networking startup options.
  * Added a brief startup grace period before establishing connections.

* **Bug Fixes**
  * Prevented duplicate connection attempts, including after shutdown.
  * Delayed initial feed synchronization until a connection is established.
  * Improved document cleanup to preserve reachable content and reduce unnecessary work.
  * Added cooperative processing to keep cleanup responsive during larger operations.


packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -630,6 +648,32 @@ export class ClientServicesHost {
...
+   */
+  startNetworking(): void {
+    if (this.#networkingStarted || !this.#edgeConnection) {
+      return;
+    }
+    this.#networkingStarted = true;
+    const edgeConnection = this.#edgeConnection;
+    log('scheduling edge networking start', { delay: EDGE_NETWORKING_START_DELAY });
+    scheduleTask(
+      this.#ctx ?? Context.default(),
```
dmaretskyi: the timeout should live in worker-runtime

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -897,6 +941,18 @@ export class ClientServicesHost {
...
 
     log('stack opened');
+
+    // Fallback ONLY for hosts with no external boot signal (node, tests). A worker host calls
+    // `startNetworking()` itself after its start sequence drains, which is strictly later than this
+    // point — firing unconditionally here would dial before the session handshake and latch the
+    // guard, turning the worker's own call into a no-op and defeating the gate. The microtask lets a
+    // worker host, which signals within the same synchronous boot, be observed as already-signalled.
+    queueMicrotask(() => {
+      if (!this.#networkingStarted) {
```
dmaretskyi: can we isntead solve it with a constructor param to auto-connect?

packages/sdk/client-services/src/packlets/services/service-host.ts
dmaretskyi: Moved. `EDGE_NETWORKING_START_DELAY` and its `scheduleTask` now live in `worker-runtime`, which already owns the boot boundary; `ClientServicesHost.startNetworking()` dials immediately and only exposes the capability. The timer hangs off a `Context` disposed by `stop()`, so a worker torn down mid-grace-period never dials.

packages/sdk/client-services/src/packlets/services/service-host.ts
dmaretskyi: Done — `ClientServicesHostProps.autoConnect` (default `true`). Embedders keep dialing on stack open; the worker passes `autoConnect: false` and drives `startNetworking()` after boot. This removes the ordering race the microtask fallback was working around instead of papering over it.

PR #12593, opened 2026-08-14
halo: complete the consumer-migration surface — invitation flows, devices, grant, atom

dmaretskyi:
## Summary

Closes **Missing APIs 4, 5, and 6** of the HALO consumer migration ([`MIGRATION_PLAN.md`](https://github.com/dxos/dxos/blob/main/packages/core/halo/halo/MIGRATION_PLAN.md)). With 1–3 already landed (#12583, #12588), **no plugin reaches for `@dxos/client` to do HALO work any more** — the HALO tier of the migration is complete.

## API 4 — device invitations and the device list

- **`useInvitationFlow(flow)`** in `@dxos/halo-react` renders any `Invitation.Flow`: its latest lifecycle event plus the shareable code, replacing a subscription to the client's `CancellableInvitationObservable`. The code is re-emitted alongside each event so a rendered QR and the flow state cannot tear.
- **`Identity.DeviceInfo` gained `presence`, `os`, `platform`, and a populated `kind`** — precisely what a device list needs for status, name, and icon. (`kind` was declared but never mapped; the adapter now derives it from the client's `DeviceType`.)
- **`@dxos/shell`'s `DeviceListItem` takes the structural `ShellDevice`** that `Identity.DeviceInfo` satisfies, so a HALO-backed caller renders it with no `@dxos/client` dependency — the same narrowing trick as `DataExtensionsIdentity` in #12229. Shell's own client-backed `DeviceList` maps through the newly exported **`toShellDevice`**, so there is one component, not a HALO fork of it.
- `DevicesContainer` reads devices from `@dxos/halo-react` and drives the invitation through `Identity.share`.

## API 5 — grant service access

**`ClientOperation.GrantServiceAccess`** (`{ serverName, capabilities }`) wraps the existing `Identity.grantServiceAccess` verb. plugin-script's settings surface dispatches it instead of hand-building a `ServiceAccess` credential and calling `client.halo.writeCredentials`.

## API 6 — identity atom

**`Identity.atom(service)`** → `Atom<Option<Info>>`, seeded from `getSnapshot()` and updated through `subscribe()` so a reader evaluating before the first stream tick sees the current identity rather than `none`. Keyed by service **reference**, not structurally — the service shape holds streams whose accessors a structural key would read (the same hazard documented on `CreateAtom.fromObservable`). app-graph-builder uses it in place of `CreateAtom.fromObservable(client.halo.identity)`.

## Two things worth a reviewer's eye

1. **`DevicesContainer` still calls `useClient`.** Not for identity or devices — for `useNetworkStatus` (swarm status) and `client.config` to read the `DX_ENVIRONMENT` gate. Both are Missing API 9. The gate is load-bearing: `composer-app/src/playwright/app-manager.ts` parses the logged invitation code off the console to drive device-invitation e2e tests, so I kept the log and its production guard rather than dropping either.
2. **`getAccessCredential` stays in plugin-script's `util/functions.ts`** — `stories-assistant` still uses it against a directly-constructed client. Only the plugin's own call site moved to the operation.

## Test plan

- `halo-e2e:test src/identity.test.ts` — **12 passed**, including two new cases: the local device reports `presence: 'online'` and a renderable `kind`; and `Identity.atom` resolves the current identity from a fresh `AtomRegistry` (proving the snapshot seeding) and returns the same atom for the same service.
- `plugin-client:test` 15 passed / 2 skipped / 2 todo · `plugin-script:test` 9 passed / 6 skipped.
- Builds green: `halo`, `halo-react`, `halo-adapter-client`, `shell`, `plugin-client`, `plugin-script`, `composer-app` (281 tasks).
- `pnpm knip` clean; lint green on every touched package (two import-order warnings caught and fixed); `pnpm format` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDnTBfQRkHtoGpXrxwn9Rc)_

## Summary by CodeRabbit

- **New Features**
  - Added invitation-flow support with shareable codes and lifecycle status.
  - Device lists now show device type, operating system, platform, and online/offline status.
  - Added service-access support for granting selected capabilities to servers.
  - Script authentication now uses the streamlined service-access flow.
  - Added reactive identity updates for more responsive account and device information.

- **Bug Fixes**
  - Improved device presentation, including icons, labels, current-device indicators, and invitation completion feedback.


dmaretskyi:
Addressed in `23fec67b`. Two of the three lifecycle points were regressions against the pre-migration container, so thanks — they were worth catching:

- **Duplicate invitations.** The create button is disabled on `!!event`, but nothing covered the window between `share()` resolving and the flow's first event. A `pending` latch set *before* the share closes it, and the share's rejection is now caught instead of escaping the promise.
- **Cancel left the flow running.** `onInvitationDone` now calls `flow.cancel()` before dropping the handle. The pre-migration code leaked the same way (clearing the observable's local state didn't cancel it), but `Invitation.Flow` makes cancelling a single call, so there's no reason to carry that forward.
- **Terminal error stranded the user.** Only `success` reset to the creation view, so a cancelled or errored invitation parked on the completion icon with no retry. Every terminal event now resets — which restores the old `state >= SUCCESS` behaviour.

Also took the nitpicks: an e2e assertion that `Identity.atom` tracks `subscribe` updates rather than only serving its seed (it now subscribes, renames the profile, and polls for the new value), three comment trims, and the `any more` → `anymore` fix in all three files.

**Declining one point:** the changeset naming only `@dxos/halo` and `@dxos/plugin-client`. Groups A and B are `fixed` lockstep groups, so naming any one member bumps the whole group — [`agents/instructions/changesets.md`](https://github.com/dxos/dxos/blob/main/agents/instructions/changesets.md) says explicitly "never enumerate a group". `@dxos/halo-react`, `@dxos/halo-adapter-client`, and `@dxos/shell` are all Group A alongside `@dxos/halo`; `@dxos/plugin-script` is Group B alongside `@dxos/plugin-client`. One name per group is the documented form, and it's the same correction CodeRabbit itself asked for on #12583.

---

PR #12599, opened 2026-08-14
agents: scaffold outbound-tag-sync project (registry entry + DESIGN/TASKS)

wittjosiah:
Docs-only scaffold for the **outbound-tag-sync** project — bidirectional email tag/label sync (outbound push): applying/removing a provider label on an email in Composer gets pushed back to the provider (Gmail label / JMAP mailbox·keyword) on the next edge sync run.

## Changes

- `.agents/projects/registry.yml` — new active entry for `outbound-tag-sync`.
- `.agents/projects/outbound-tag-sync/docs/DESIGN.md` — architecture + decisions: per-message tag-set snapshot (`Cursor.readSnapshot`/`writeSnapshot`) + set-wise three-way merge in tag-uri space, local-wins conflict policy, echo-proof by construction; pending-op journal explicitly rejected. All cited file anchors re-verified against main @ `1c995c468d` (this branch's base) — no drift.
- `.agents/projects/outbound-tag-sync/docs/TASKS.md` — phased ledger: Phase 1 merge machinery (app-toolkit `mergeSet`, snapshot helpers, mail-sync harness rewiring, Tag applicability split), Phase 2 Gmail (`users.messages.modify`), Phase 3 JMAP (`emailSetUpdate`), Phase 4 Composer tag UI, Phase 5 tests + changesets, plus a follow-ups ledger.

No implementation in this PR — project scaffolding only. Implementation PR sequencing (in TASKS.md): PR 1 = Phases 1+2, PR 2 = Phase 3, PR 3 = Phase 4.

https://claude.ai/code/session_01GEF8ArC6hZ5pQB4NNazGpz

---
_Generated by [Claude Code](https://claude.ai/code/session_01GEF8ArC6hZ5pQB4NNazGpz)_

### DIFF
--- a/.agents/projects/outbound-tag-sync/docs/DESIGN.md
+++ b/.agents/projects/outbound-tag-sync/docs/DESIGN.md
@@ -1,0 +1,119 @@
+# Outbound Tag Sync — Design
+
+Bidirectional email tag/label sync: applying or removing a provider label on an email in
+Composer is pushed back to the provider (Gmail label / JMAP mailbox·keyword) on the next edge
+sync run. Both providers (plugin-google, plugin-jmap).
+
+Designed against main @ `1c995c468d`; all file citations below re-verified at that commit
+(2026-08-14) — no drift.
+
+## Goal & scope
+
+- **Core acceptance:** add a Gmail label to a message in Composer → next sync run → label
+  visible in Gmail.
+- Offline durability is NOT a design constraint — sync is edge-driven; retry across sync
+  ticks is sufficient.
+
+## Current state
+
+- **Inbound label→tag sync is landed.** Gmail labels map to DXOS Tag objects
+  (`packages/plugins/plugin-google/src/operations/mail/tags.ts` — `GMAIL_TAG_SOURCE` foreign
+  key; system labels map to canonical SystemTags via `GMAIL_SYSTEM_TAGS`). The delta path
+  (`history.list`) folds `labelsAdded`/`labelsRemoved` into `retag` reconcile items
+  (`packages/plugins/plugin-google/src/operations/mail/sync/sync-provider.ts`,
+  `collectLabelChanges`) applied by the shared harness
+  (`packages/plugins/plugin-inbox/src/sync/mail-sync.ts`, `reconcileToChanges`).
+- **JMAP mirrors this:** folders/keywords → tags
+  (`packages/plugins/plugin-jmap/src/operations/mail/tags.ts`, `JMAP_DOMAIN`); its reconcile
+  re-fetches `updated` emails and diffs.
+- **NO outbound path exists:** nothing calls Gmail `users.messages.modify` for labels. The
+  only Gmail write scopes in use: `gmail.send`, and `gmail.modify` solely for trash
+  (`packages/plugins/plugin-google/src/capabilities/connector.ts:136`) — so the OAuth scope
+  needed for label writes is ALREADY granted.
+- **JMAP already has a generic write primitive:** `emailSetUpdate`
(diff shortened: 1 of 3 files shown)


richburdon:
Heads-up on a collision: I've implemented this feature independently on #12611 (branch `claude/mailbox-tag-sync-89f351`), which is green and covers Gmail end-to-end. Our designs were written a day apart without knowing about each other, and they converge on most of the architecture — worth reconciling before Phase 1 starts here, since its first task builds machinery #12611 would delete.

**Where we independently agreed** (encouraging, given no contact):

- Three-way merge in tag-uri space, per (message, tag).
- A pending-op journal explicitly rejected, for the same reason — sync writes tags through the same `Tagging.set` the user does, so an intent queue self-echoes.
- Echo-proof by construction rather than suppression bookkeeping.
- First sync takes remote and pushes nothing.
- Optional provider hook, so a provider without a write path degrades to pull-only.
- Push failure leaves the base unadvanced and self-retries; per-run push budget alongside `maxMessages`.
- Unmapped user tags skipped; label creation deferred.

**Three things that changed under testing**, which may save you the same work:

1. **The base is Automerge heads, not per-message snapshots.** `Obj.version` / `Obj.getVersion` recover the tag index as of the last sync, so there is no shadow copy and — critically — **no write points**. This design has three (insert commit, inbound retag, push success), and a snapshot that must land with every tag commit can disagree with the data after a crash. It also removes the open sub-decision about `cursor.spec.snapshots` growth: one heads array per binding instead of an entry per message. Landed as `Cursor.spec.tagHeads` + `Cursor.writeSyncState`, which writes the heads and the delta token in ONE `Obj.update` — advancing the token without the heads lets the next run diff a fresh delta against a stale base, and no conflict rule catches that because the diff sees no conflict.

2. **The conflict case cannot occur, so "local wins" is unreachable.** Membership is a boolean on each side, so `local ≠ base ∧ remote ≠ base ⇒ local = remote` — both flipped to the same value. All eight `(base, local, remote)` triples resolve to push, pull, or nothing. The "full element truth table" test in Phase 1 would have found this. We removed the policy rather than keep dead code, and left an enumeration test asserting no tag is ever pushed *and* pulled, so a future tri-state fails loudly instead of quietly reviving the question.

3. **`CATEGORY_*` is fine; `SENT` is not.** The DESIGN flags "empirically verify which system labels modify accepts (`CATEGORY_*` suspect)" — good instinct, wrong label. Verified live: all five `CATEGORY_*` accept a write (200), as do `STARRED`, `IMPORTANT`, `UNREAD` and `SPAM`; `SENT` and `DRAFT` return `400 Invalid label`. `SENT` is in `GMAIL_SYSTEM_TAGS`, and the send flow applies the canonical `sent` tag on every send — so without an exclusion every sent message produces a permanent 400. That check found a real bug in our shipped code.

**Two things your design covers that ours doesn't**, and I've recorded both as follow-ups rather than pretending they're out of scope:

- **The `isProviderTag` split.** Ours makes provider tags pushable but the UI still forbids toggling them, so our eligibility is half-connected. Yours is right.
- **The Composer tag affordance (Phase 4).** Without it, most of the push path is reachable only by the classifier.

I also took your mock design — `modifyMessage` mutating the dataset *and* appending a historyLog step — which turned the echo round trip into a unit test instead of something only the live suite could prove. Gmail does record a client's own label write in the account history; we confirmed that against a real account.

Happy to go whichever way you prefer: land #12611 and re-point this project at the remaining scope (the tag split, the UI, JMAP), or treat #12611 as a prototype and rebuild here. Flagging now rather than after Phase 1 is underway.

PR #12602, opened 2026-08-14
composer-app: fix WebKit dev-server boot and narrow tauri dev deps

wittjosiah:
## What

Two fixes to composer-app's dev loop, plus a copy fix in the updater surfaced while verifying them.

**1. WebKit dev-server boot failure.** Booting the dev server in Safari or the Tauri WKWebView failed with `Cannot access 'makeIdentityService' before initialization`, leaving the app on the System Error screen. Chromium was unaffected on the same commit.

**2. `tauri-dev` / `tauri-ios` dependency scope.** Both inherited `deps: ^:build` from the `tauri` tag and built the whole workspace graph (100+ targets) before starting.

**3. Updater settings copy.** The updater disabled itself for two unrelated reasons — no OTA channel on the platform, and running against the dev server — but reported both as "Updates are not available on this platform". On macOS in dev that is simply wrong: the same build updates fine once packaged. Split into `unsupported` and `dev` with separate copy; both still disable the check button.

### The WebKit bug

WebKit evaluates a module **before its own dependencies** when the graph is reached by concurrent dynamic imports and contains top-level await, so a binding first reached that way reads as an uninitialized `const`. Marker order in the app confirmed it — `client-module, util, identity, space, barrel`, exactly backwards.

Isolated to a minimal case on the same dev server:

```
WebKit:   shared-dep, shared-barrel, c-root1, slow-tla, c-root2   ← c-root1 runs BEFORE the TLA module it imports
Chromium: shared-dep, shared-barrel, slow-tla, c-root2, c-root1   ← correct
```

Top-level await alone orders correctly in both engines; TLA **plus** concurrent overlapping dynamic imports is what breaks. The plugin manager activates modules concurrently, so the only lever is the top-level await — which is not in automerge's source, but injected by rolldown's optimizeDeps pre-bundle when it inlines `.wasm` as `var {...} = await initWasm({...})`.

`syncWasmInit` resolves `@automerge/automerge` and `@automerge/automerge-subduction` **without the `browser` condition**, selecting the base64 entrypoints that inline the same bytes and call a synchronous `initSync`.

### Rejected alternatives (both measured, not reasoned)

| Approach | Result |
|---|---|
| `optimizeDeps.exclude` alone | 3/3 still failed — excluding just moves the `.wasm` import to `vite-plugin-wasm`, which re-adds TLA |
| Barrel `export…from` → `import`+`export` | 3/3 still failed — the barrel itself evaluates last, so the binding is uninitialized either way |

### Tauri task deps

Both native dev loops run this app's `vite dev` through `beforeDevCommand` (`tauri.ios.conf.json` overrides only `bundle`), so they need exactly `serve`'s narrowed set — extended from it so the three inner loops cannot drift. Goes from 100+ build targets to 20.

## Reviewer notes

- **`mergeArgs: replace` is load-bearing.** Args merge by append; without it the resolved command was `pnpm exec tauri dev exec vite dev exec tauri dev`.
- **Subpath matching is load-bearing.** `/slim` re-exports the same wasm glue as the package root, so redirecting only the root would split them across separate wasm instances and leave the signer `Repo` gets uninitialized.
- **Scoped to `serve`** via `apply: 'serve'`. The base64 entry costs ~33% over the binary, and the app gates on `check-boot-budget`; the failure also needs vite's module-per-file dev graph to bite. Production Tauri ships today. Extending to production would need a bundle-size measurement first — there's a TODO on the plugin.
- `oxc-resolver` is now a declared devDependency rather than a phantom one; the lockfile diff is 3 lines.
- One changeset, for `@dxos/plugin-native` (the updater copy). The composer-app changes need none — it is private and out of Changesets.

## Verification

| Check | Result |
|---|---|
| Playwright WebKit × 5 | 0/5 (was 5/5) |
| Playwright Chromium × 2 | clean, real UI renders |
| Real Safari | 0 occurrences in `app.log` |
| Module eval order | `util, identity, space, barrel, client-module` — deps first |
| `moon run :lint -- --fix` | pass, 310 tasks |
| `composer-app:build` (tsgo --noEmit) | pass, 281 tasks |
| `composer-app:test` | 16 passed, 1 skipped |
| `plugin-native:build` / `:lint` / `:test` | pass (`Match.exhaustive` enforces the new status case) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed local development loading for Automerge WebAssembly modules, improving startup reliability.
  * Improved Composer development and iOS workflows with dedicated Tauri commands.

* **Development Experience**
  * Improved local module resolution for more consistent development builds.
  * Clarified update availability messaging in development mode and disabled unavailable update checks.
  * Improved update status reporting for platforms that do not support over-the-air updates.

### DIFF
--- a/.changeset/native-updates-dev-mode-message.md
+++ b/.changeset/native-updates-dev-mode-message.md
@@ -1,0 +1,5 @@
+---
+'@dxos/plugin-native': patch
+---
+
+Distinguish the two disabled updater states in settings: a platform with no OTA channel still reads "Updates are not available on this platform", while a dev-server build on a platform that does support OTA now reads "Updates are not enabled in dev mode".
--- a/packages/apps/composer-app/moon.yml
+++ b/packages/apps/composer-app/moon.yml
@@ -72,6 +72,21 @@ tasks:
   tauri-ios-build:
     deps:
       - bundle
+  # Both native dev loops run this app's `vite dev` through `beforeDevCommand` (`tauri.ios.conf.json`
+  # overrides only `bundle`), so they need exactly `serve`'s narrowed dependency set rather than the
+  # `tauri` tag's whole-graph `^:build` — extended so the three inner loops cannot drift.
+  tauri-dev:
+    extends: serve
+    command: pnpm exec tauri dev
+    options:
+      # Args merge by append, so both the extended `vite dev` and the tag's own `tauri dev` would
+      # otherwise be concatenated onto this command.
+      mergeArgs: replace
+  tauri-ios:
+    extends: serve
+    command: pnpm exec tauri ios dev
+    options:
+      mergeArgs: replace
   # Downloads the pinned macOS ollama runtime + triple-named sidecars embedded in the desktop app. Cacheable:
   # keyed on the fetch script (which pins the version + SHA-256), with the downloaded sidecar dir as the
   # output — so a cache hit restores it instead of re-downloading the (large) runtime.
--- a/packages/apps/composer-app/package.json
+++ b/packages/apps/composer-app/package.json
@@ -212,6 +212,7 @@
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
     "jsdom": "catalog:",
+    "oxc-resolver": "catalog:",
     "pwa-asset-generator": "catalog:",
     "rollup-plugin-visualizer": "catalog:",
     "vite": "catalog:",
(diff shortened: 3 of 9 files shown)


wittjosiah:
Addressed the review notes:

**Comment verbosity** (`vite.config.ts` L50-52, L70-71, L137-141) — trimmed all three to a single load-bearing clause, per the repo's comment rule. Pushed in `1c016db`.

**`enabled` invariant** (`updater.ts` L71-75) — verified, no change needed. `enabled` is defined one line above as `SUPPORTS_OTA.includes(platform) && !isDevServer`, so it has exactly two disabling conditions. `disabledStatus` selects `dev` only when `SUPPORTS_OTA.includes(platform)` is true, which leaves `isDevServer` as the sole remaining falsifier — the two states are exhaustive and cannot mislabel by construction.

Issue #12608, opened 2026-08-15
Feed triggers scan the whole feed on every dispatcher tick


richburdon:
`TriggerDispatcher` re-reads the entire feed for every feed trigger on every poll tick, then discards everything at or before the cursor in JS. The cursor already says where to resume, so the scan is proportional to the whole feed rather than to what is new.

[`trigger-dispatcher.ts:645`](https://github.com/dxos/dxos/blob/main/packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts#L645) — the `TODO` is yours:

```ts
// TODO(dmaretskyi): Include cursor & limit in the query.
const chunks = yield* Feed.query(feed, Filter.everything()).run.pipe(
  Effect.map((objects) => filterReadyFeedItems(objects, cursor)),
  Effect.map(Array.chunksOf(concurrency)),
);
```

## Measured

Against a live Composer profile with a 538-message Gmail mailbox, via the agent debug port:

| | |
| --- | --- |
| Full-feed query | **1069 ms**, of which **924 ms** was a single unbroken main-thread long task |
| ECHO queries while idle | **4 / sec** (`graph-query-context.ts` `run query` / `run query results`) |
| Trigger state refreshes | 2 / sec — one dispatcher per space, 2 spaces |
| Main thread blocked | **2302 ms per 10 s — ~23%**, sustained |
| Heap | 1.3–1.5 GB |

That was measured on the **settings page**, with no mailbox open: the cost is the dispatcher, not the mailbox surface.

## Partially mitigated

`livePollInterval` now defaults to 1 minute instead of 1 second (was a constructor default nobody passed — the app builds the dispatcher as `TriggerDispatcher.layer({ timeControl: 'natural' })`). That makes the scan 60× rarer. **It does not make it cheaper**, and the cost still grows with feed size.

## Suggested

1. Push the cursor (and a limit) into the query, as the `TODO` says.
2. Consider splitting the kinds off the wall-clock tick — `invokeScheduledTriggers` defaults to `['timer', 'feed', 'subscription']`, but only `timer` genuinely needs polling. Feed triggers could react to appends.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12615, opened 2026-08-15
echo, compute-runtime: resume feed queries from a cursor and wake feed triggers on append

dmaretskyi:
Fixes the scan in #12608: `TriggerDispatcher` re-read every block of every feed on every poll tick and discarded everything at or before its cursor in JS, so idle cost grew with feed size (1069 ms, 924 ms of it a single unbroken main-thread task, for a 538-message mailbox).

## What changed

**A cursor read is now an index seek.** A feed block's global position is denormalized into a new `objectMeta.queuePosition` column (entity-meta migration `0004`, plain SQL, indexed by `(queueId, queuePosition)`), so the scan bounds and pages in SQL instead of in memory.

**Query API.** `Filter.feedCursor({ begin?, end? })` is a filter node in the standard DSL, composing with the existing `limit()`:

```ts
Query.select(Filter.feedCursor({ begin: cursor })).limit(10).from(feed);          // next 10 after a checkpoint
Query.select(Filter.and(Filter.type(Task), Filter.feedCursor({ begin: cursor }))) // narrowed by type
```

The planner lifts the node off the filter tree onto the select step it bounds, so the rest of the query plans exactly as it would without it. Both bounds name an item and exclude it, so a reader resumes with the cursor it last consumed as `begin` without re-reading that item. Three things are refused rather than silently ignored, since a cursor bounds a scan rather than describing an object: composing with anything but AND, two cursor filters in one query, and use where the query also selects a space's documents (an automerge object has no position). A bound that is not a decimal position is unsatisfiable rather than absent — resuming a corrupted checkpoint from the beginning would re-dispatch the whole feed.

**Typed cursor.** `Feed.Cursor` (branded), read off an item with `Feed.getCursor`, starting at the `Feed.START` sentinel, and stored on a reader with `Feed.CursorAnnotation`. This replaces the stubbed `Feed.cursor` / `Feed.next` / `Feed.nextOption` iterator, which died at runtime and had no callers, and replaces the dispatcher's `@meta` foreign-key checkpoint — a key identifies the same entity in another system, which a checkpoint is not. The old key stays readable once so a trigger already checkpointed in the field resumes instead of re-dispatching its feed, and is dropped when the cursor next advances.

**Trigger kinds split off the wall clock.** Feed triggers watch one item past their cursor and subscription triggers watch their own query; each dispatches only the trigger whose source fired, serialized by a lock so overlapping wake-ups cannot invoke a trigger twice for one item. The poll tick now runs `timer` alone. Feed dispatch reads a bounded page per iteration, and `filterReadyFeedItems` orders by position — the dispatcher advances its cursor to the last item it invoked, so an unordered page would skip everything positioned before it.

## Testing

- `index-core:test` — `EntityMetaIndex` case covering `after`, `after + limit`, `queryTypes` windowing, and exclusion of unpositioned blocks (47 passed).
- `echo-client-e2e:test` — cursor cases over a position-assigning peer: resume, bounded page, `begin`/`end` range, the `START` sentinel, a past-the-tail cursor, and rejection over space-only and mixed scopes (297 passed).
- `compute-runtime:test` — a natural-time case asserting a feed trigger fires on append with a one-hour poll interval, and one asserting the checkpoint lands in the annotation while a legacy foreign key is still adopted. Both verified to fail when their fix is reverted (164 passed).
- `echo:test`, `echo-query:test` green; clean full `:build` sweep.

## Not included

Moving `KEY_QUEUE_POSITION` — the position stamped on feed *items* by the storage layer — from a `@meta` foreign key to an annotation. That changes the stored representation across the feed codec, the index write path, and replicated data, so it wants its own migration and review. The reader-side checkpoint (`Feed.CursorAnnotation`) is done here; the item-side stamp is not.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TqVarSpYx1H6NhuJXrHHJD)_


dmaretskyi:
Addressed in ef8a9ddc:

- **Dispatch the trigger that owns the changed source** — `invokeScheduledTriggers` gains a `triggerIds` filter and the reactive callback passes the firing trigger's id, so one append no longer re-queries every other trigger of that kind. Kind-wide dispatch remains the default for the poll tick and manual calls.
- **Reactive-source key** — now `feed:<feed uri>:<cursor>` / `subscription:<query ast>`, so an edited spec replaces the source instead of leaving a stale one subscribed.
- **Serialize reconciliation** — `#refreshReactiveSources` spans an await and `refreshTriggers` runs from both the trigger-query callback and `invokeScheduledTriggers`, so it now holds its own semaphore.
- **Test teardown** — the natural-time test stops the dispatcher in an `Effect.ensuring` finalizer.

Not changing the fourth (out-of-diff, "preserve cursor order when invoking a page"): the concurrent invocation of a page and the contiguous-prefix cursor advance via `Array.takeWhile` are both pre-existing — the previous code chunked by the same `concurrency` and invoked each chunk with `concurrency: 'unbounded'`. This PR changes where the page comes from (a cursor query rather than a slice of a full-feed scan), not how it is invoked, so the replay window is unchanged rather than introduced. Making feed dispatch strictly sequential is a real behavior change for per-trigger `concurrency` and belongs in its own PR against the same code.

`compute-runtime:test` — 162 passed, 1 skipped.

---

packages/core/echo/echo/src/Feed.ts
```diff
@@ -74,11 +73,33 @@ export class Feed extends Type.makeObject<Feed>(DXN.make('org.dxos.type.feed', '
...
+ * Sentinel cursor preceding every item — "read from the beginning".
+ *
+ * A sentinel rather than `undefined` so a stored checkpoint is always a `Cursor`: a reader that has
+ * processed nothing yet holds this, and the same code path resumes it as resumes a live cursor.
+ */
+export const START: Cursor = Cursor.make('');
+
+/**
+ * Options for {@link query}.
+ */
```
dmaretskyi: no, build them into the standard query DSL

packages/core/echo/echo/src/Scope.ts
```diff
@@ -43,8 +43,17 @@ export const registry = (location: 'local' | 'remote' = 'local'): QueryAST.Regis
...
+ * Pass `after` to resume from a cursor — only items positioned strictly after it are selected,
+ * pushed into the index scan rather than filtered in memory.
+ *
+ * @example
+ * ```ts
+ * db.query(Filter.everything().from(Scope.feed(feedUri, { after: cursor })));
+ * ```
  */
-export const feed = (feedUri: string): QueryAST.FeedScope => ({
+export const feed = (feedUri: string, options?: { after?: string }): QueryAST.FeedScope => ({
```
dmaretskyi: this should be part of the Filter

packages/core/echo/echo/src/Feed.ts
dmaretskyi: Done in 1c5e08bc — the options bag is gone and the bound is a filter node:

```ts
Feed.query(feed, Query.select(Filter.feedCursor(cursor)).limit(10))
```

The planner lifts the `feed-cursor` node off the filter tree onto the select step it bounds, so the rest of the query plans exactly as it would without it and `limit()` is the existing clause rather than a second way to say the same thing.

---

packages/core/echo/echo/src/Scope.ts
dmaretskyi: Moved — `Scope.feed` is back to `(feedUri)` and the bound is `Filter.feedCursor(after)`:

```ts
db.query(Query.select(Filter.feedCursor(cursor)).limit(10).from(Scope.feed(feedUri)));
db.query(Query.select(Filter.and(Filter.type(Task), Filter.feedCursor(cursor))).from(feed));
```

Two things the filter form has to rule out, since a cursor bounds a scan rather than describing an object:

1. Composing with anything but AND (negated, OR-ed, or two cursors in one query) is a `QueryError` rather than a silently dropped bound, which would return the whole feed.
2. Using it where the query also selects a space's documents is a `QueryError` too — an automerge object has no position, so there is no answer to give.

---

packages/core/echo/echo/src/Feed.ts
dmaretskyi: Valid — fixed in 64ae47a1. The scope check now happens in the planner, where the filter is lifted, rather than in the executor where the bound is applied, so the start sentinel (which bounds nothing and therefore reaches no window) is refused over a space's documents just the same. Coverage added for both the sentinel and a live cursor, against a space-only scope and a mixed feed-and-space scope, asserting on the error rather than on any rejection.

That commit also merges `main`, which had landed the same `livePollInterval` knob.

---

dmaretskyi:
CI was red on `64ae47a1` — `echo-query`'s `query-lite` re-implements the `Filter` surface structurally and switches exhaustively over filter kinds, so `feed-cursor` had to land there too. Fixed in 226afb4b; `build`, `check`, `test` and `storybook` all failed off that one compile error.

Same commit picks up two review points:

1. **Rejected items in the initial page** (valid): `Feed.START` previously produced no window at all, so a read from the start came back in whatever order the scan met rows and could hand a pager nothing but unpositioned blocks while positioned ones waited behind them. Now every cursor windows the scan, the sentinel included — positioned blocks only, in position order, `after = -1`. Test asserts a limited read from the start returns the first items *by position*.
2. **Fiber tracking race** (valid, trivial): a dispatch that completes without an async boundary runs its finalizer before `runFork` returns, so the fiber is tracked through a cell and the set is not re-populated with a completed fiber.

Not taking the third (watch the feed without a cursor bound so the source survives cursor advances): an unbounded live query over a feed is exactly the full-feed scan this PR exists to remove — on the 538-message mailbox in #12608 that was 1069 ms per re-run. The rebuild it avoids costs one small indexed query per processed batch, so the trade runs the wrong way.

Local: `echo` 572, `echo-query` 13, `index-core` 47, `compute-runtime` 164, `echo-client-e2e` feed pagination 14 — all passing, plus a clean full `:build` sweep. `echo-host` has one flaky failure per full-suite run in the subduction/replication area (a different test each time, none feed-related); it passes in isolation.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
```diff
@@ -661,18 +695,22 @@ class TriggerDispatcherImpl implements Context.Service.Shape<typeof TriggerDispa
...
               const concurrency = Math.min(trigger.concurrency ?? 1, this._maxConcurrency);
 
-              // TODO(dmaretskyi): Include cursor & limit in the query.
-              const chunks = yield* Feed.query(feed, Filter.everything()).run.pipe(
-                Effect.map((objects) => filterReadyFeedItems(objects, cursor)),
-                Effect.map(Array.chunksOf(concurrency)),
-              );
+              // Read only what follows the cursor, one page at a time: the cursor is pushed into the
+              // index scan, so the cost of a tick is the size of the page rather than of the feed.
+              let cursor = Obj.getKeys(trigger, KEY_FEED_CURSOR).at(0)?.id;
```
dmaretskyi: I wanna change KEY_FEED_CURSOR to be a Feed.CursorAnnotation -- represented in api as an echo annotation. Meta.key is not correct for it.

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
```diff
@@ -661,18 +695,22 @@ class TriggerDispatcherImpl implements Context.Service.Shape<typeof TriggerDispa
...
+              for (;;) {
+                const chunk = yield* Feed.query(
+                  feed,
+                  Query.select(Filter.feedCursor(cursor ?? Feed.START)).limit(concurrency),
```
dmaretskyi: `Filter.feedCursor({ begin?, end? })` -- better api, still works with .limit()

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
dmaretskyi: prefer using Feed.getCursor

packages/core/echo/echo/src/Filter.ts
```diff
@@ -388,6 +388,28 @@ export const updated = (range: TimeRange): Any => _timeRangeFilter('updatedAt',
...
+ * has not yet acknowledged has no place in that order, and `Feed.START` selects the same set from
+ * the beginning rather than everything. Only meaningful against a feed scope — an automerge object
+ * carries no position, so a query that also selects a space's documents is rejected.
+ *
+ * @example
+ * ```ts
+ * db.query(Query.select(Filter.feedCursor(cursor)).limit(10).from(feed));
+ * ```
+ */
+export const feedCursor = (after: string): Any =>
```
dmaretskyi: use Feed.Cursor type

packages/core/echo/index-core/src/migrations/entity-meta/index.ts
```diff
@@ -25,15 +25,29 @@ const LATER_COLUMNS: ReadonlyArray<readonly [name: string, ddl: string]> = [
...
 ];
 
-const addMissingColumns = Effect.gen(function* () {
-  const sql = yield* SqlClient.SqlClient;
-  const columns = yield* sql.unsafe<{ name: string }>('PRAGMA table_info("objectMeta")');
-  const present = new Set(columns.map((column) => column.name));
-  for (const [name, ddl] of LATER_COLUMNS) {
-    if (!present.has(name)) {
-      yield* sql.unsafe(`ALTER TABLE objectMeta ADD COLUMN ${ddl}`);
+const addColumns = (columns: ReadonlyArray<readonly [name: string, ddl: string]>) =>
```
dmaretskyi: why can't it be a pure sql migration?

packages/core/echo/index-core/src/migrations/entity-meta/index.ts
```diff
@@ -25,15 +25,29 @@ const LATER_COLUMNS: ReadonlyArray<readonly [name: string, ddl: string]> = [
...
+  });
+
+const addMissingColumns = addColumns(LATER_COLUMNS);
+
+/**
+ * The global position a feed block was assigned, denormalized out of the indexed snapshot so a
+ * cursor read (`queuePosition > ?` ordered and limited) is an index seek rather than a full feed
+ * scan. Null for automerge objects and for local blocks not yet positioned.
+ */
+const addQueuePosition = Effect.gen(function* () {
```
dmaretskyi: can this be a pure sql migration? I'd stick to pure migrations as much as possible

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
dmaretskyi: Done in 67d71b88 — `Feed.CursorAnnotation` (`org.dxos.annotation.feed-cursor`, schema `Feed.Cursor`), read and written through `Annotation.get` / `Annotation.set`.

One thing I added deliberately: `LEGACY_KEY_FEED_CURSOR` keeps the old foreign key readable as a fallback, and the dispatcher drops it the first time the cursor advances. Triggers already checkpointed in live profiles would otherwise resume from nothing and re-dispatch a whole mailbox — an AI invocation per message. Test covers it, and I verified it fails without the fallback.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
dmaretskyi: Done in 67d71b88 — `Filter.feedCursor({ begin?, end? })`, both typed `Feed.Cursor`, still composing with `limit()`:

```ts
Query.select(Filter.feedCursor({ begin: cursor })).limit(10)          // next 10 after a checkpoint
Query.select(Filter.feedCursor({ begin: first, end: last }))          // between two known items
```

Semantics I picked, flagging in case you want them the other way: **both bounds name an item and exclude it**. `begin` has to be exclusive or a reader that stores the cursor it just consumed re-reads that item forever; `end` follows it for symmetry. `end` maps to `queuePosition < ?` in the same index scan, so a range is still one seek.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
dmaretskyi: Done in 67d71b88 — `filterReadyFeedItems` reads each item's cursor with `Feed.getCursor` and returns it as a `Feed.Cursor`, so the dispatcher no longer reaches for the queue-position key itself. Its local `getFeedPosition` accessor is gone.

---

packages/core/echo/echo/src/Filter.ts
dmaretskyi: Done in 67d71b88 — `Filter.FeedCursorRange` is `{ begin?: Feed.Cursor; end?: Feed.Cursor }`. `Filter` takes a type-only import of `Feed`, so the module cycle stays types-only.

---

packages/core/echo/index-core/src/migrations/entity-meta/index.ts
dmaretskyi: It can, and now is — `0004_queue_position.sql` in 67d71b88, applied with `SqlMigrations.apply`. I had generalised 0002's probe out of caution; unnecessary here, since `queuePosition` has never shipped, so no database in the field can already hold it and a plain `ALTER TABLE ... ADD COLUMN` is safe. 0002 is back to its original shape, untouched.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.ts
```diff
@@ -661,18 +695,22 @@ class TriggerDispatcherImpl implements Context.Service.Shape<typeof TriggerDispa
...
+                ).run.pipe(Effect.map((objects) => filterReadyFeedItems(objects, cursor)));
+                if (chunk.length === 0) {
+                  break;
+                }
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 226afb4b: `Feed.START` now windows the scan like any other cursor (`after = -1`), so the first page is positioned items in position order rather than whatever the unordered scan met first. A page of unpositioned blocks can no longer stall the pager. Covered by a test asserting a limited read from the start returns the first items by position.

---

PR #12619, opened 2026-08-15
compute: add operation tags and filter the trace panel by them

wittjosiah:
The trace panel shows every operation the graph produces, which buries the interesting work under navigation, layout, and per-keystroke edits. This adds a coarse classification to operations and a filter over it, defaulting to a selection that is useful without configuration.

## Tag vocabulary

`Operation.Tag` in `@dxos/compute/Operation`. A tag answers "would someone watching the system want to see this?", not "which plugin owns it". Operations may carry several.

| Tag | Meaning | Default |
| --- | --- | --- |
| `ui` | Layout, navigation, dialogs, popovers, toasts, selection, view toggles | hidden |
| `edit` | Direct user edits to objects, fields, rows, cards | hidden |
| `query` | Read-only lookups, listings, remote target discovery | hidden |
| `system` | Internal plumbing, debug affordances, samples | hidden |
| `space` | Space and membership lifecycle | shown |
| `identity` | Identity, devices, credentials, service access | shown |
| `sync` | Synchronization with an external service | shown |
| `agent` | Agentic or model-driven work (research, summarize, classify, extract, generate) | shown |
| `automation` | Triggers, routines, schedules, alarms | shown |
| `tool` | Exposed to the AI as a tool, so an invocation is something the agent chose to do | shown |
| `untagged` | Pseudo-tag, not a real category — matches operations that declare none | shown |

The defaults follow the shape of the request: trigger-driven, batch, and agent-initiated work is almost always worth seeing; direct UI interaction is mostly noise. `untagged` is shown by default because a trace outlives the code that wrote it — events recorded before tagging, or by an EDGE-deployed definition this build has never seen, carry no tags, and dropping them silently would make the panel look broken on existing data.

`tool` is derived rather than guessed: an operation is tagged `tool` when its key is in the `org.dxos.function.*` namespace, when it appears in a skill's `Skill.toolDefinitions({ operations })`, or when it is projected via `Operation.mcpTool`.

## Changes

**`@dxos/compute`**
- `Operation.Tag`, `Operation.DEFAULT_TAGS`, and `Definition['meta'].tags`. The `Tag` type is `(known tags) | (string & {})`, so plugins can define their own without losing completion on the known set.
- `Trace.OperationStart` / `Trace.OperationEnd` carry `tags`, written by `Process.fromOperation`. Denormalizing onto the event is what lets the panel filter a historical trace: the registry may no longer hold the definition, and cannot be consulted synchronously while rendering.

**Tagging** — 358 operation definitions across 87 files.

**`@dxos/plugin-assistant`**
- `buildExecutionGraph({ operationTags })` drops operations outside the selection. Hiding an operation keeps the work it started: `filterSpanTree` splices the span out and hoists its children, so silencing a noisy wrapper does not silence what it kicked off. Non-operation activity (agent requests, tool calls, status) is never filtered.
- `collectOperationTags` derives the offered options from what actually ran, so the list reflects reality rather than a hard-coded set.
- New `TraceToolbar` above the process tree: a primitive toolbar whose bare funnel button opens a searchable checklist of tags. The selection persists in assistant settings (`traceOperationTags`).

**`@dxos/react-ui-list` — multi-select `Combobox`**

The filter needed a multi-select picker and the package had none: `Combobox`, `Picker`, `Listbox`, and every form field were single-select, while `useListSelection` had a tested multi mode no component exposed.

- `Combobox.Root` accepts `multiple`, a discriminated union over the value model — `value: string` + `onValueChange: (value: string) => void` when single, `readonly string[]` when multiple. Items derive `checked` from the selection, toggle their own membership, and default to leaving the popover open, since setting several values in one visit is the whole point.
- `Picker` separates the two states multi-select divides. The keyboard cursor was riding on `aria-selected`, which in a multi-select listbox must mean *chosen*; it now sets `data-highlighted` and is announced through the input's new `aria-activedescendant`, per the WAI-ARIA combobox pattern. `dx-highlighted` already existed in the theme for exactly this. Single-select behaviour is unchanged apart from the (more correct) attributes.
- `Combobox.List` sets `aria-multiselectable` when multiple.
- The `Grid` story's "Multiselect" popover — the demo behind `dx-grid-multiselect-cell` — now actually multi-selects.

The earlier `selectCardinality: 'multiple'` support in `react-ui-menu` is reverted: the combobox replaced its only consumer.

## Test plan

- `react-ui-list:test` — 8 new tests (`Combobox.test.tsx`) over both variants: single commits and closes, multiple toggles and stays open, `aria-multiselectable`, trigger summary, and that the keyboard cursor is a distinct element from the chosen ones. Full suite 38 passed.
- `react-ui-list:test-storybook` — 31 passed, including the new `Multiple` story rendering in chromium.
- `plugin-assistant:test` — 20 tests over the filter (`operation-tag-filter.test.ts`, `trace-filter.test.ts`): tag collection, selection matching, the untagged fallback, child hoisting, and that non-operation activity survives. Full suite 188 passed | 7 skipped.
- `compute:test` — 54 passed.
- `moon exec :build` across the repo — clean. Lint clean on the touched packages; `pnpm format` applied.

## Open questions

1. The taxonomy and the default selection are proposals — both are cheap to revise (the defaults are one array in `Operation.ts`; a retag is a one-line edit per operation).
2. `tool` being on by default means agent database lookups (`org.dxos.function.database.query`) show. That follows "anything coming from the agent is useful", but it is the noisiest of the default-on tags.
3. Tags are not persisted onto `Operation.PersistentOperation` — that record is at schema `0.2.0` and adding a field needs a version bump. EDGE-deployed operations therefore surface as `untagged`.
4. `Combobox`'s trigger summarizes a multi-selection by joining the values; a caller wanting chips or a count passes `displayValue` (the Root knows values, not the labels its items render). Worth revisiting if a consumer wants token chips.

https://claude.ai/code/session_01FTrDNj3jEpCLJ56nji9XgU


wittjosiah:
Closing in favour of #12624 for the time being.

The branch `claude/trace-panel-operation-filter-vawrun` is pushed and intact at `7d4faa2c` if any of this is worth picking up later:

- `tags` on operation metadata, drawn from a closed vocabulary in `@dxos/app-toolkit/OperationTag` (layout, navigation, assistant, connector, database, identity, system), applied across the operation definitions and recorded on `Trace.OperationStart` / `Trace.OperationEnd`.
- A tag filter over the trace panel's process list, classifying processes by the operation each one runs via `OperationHandlerSet.definitions()`. Offers only tags that have actually been observed; processes that are not operation invocations are never filtered.
- `Combobox` gains a `multiple` variant whose items toggle their own membership, and `Picker` separates the keyboard cursor (`data-highlighted`, addressed by the input's `aria-activedescendant`) from the chosen options (`aria-selected`).

One finding worth carrying forward independently of the UI: an operation's trace events only reach the feed when a space is in scope. `FeedTraceSink.makeRoutingSink` drops any message without `meta.space`, and the app-level invoker sets no default `spaceId`, so operations invoked from app chrome — essentially the whole layout/navigation surface — are never recorded.

---

PR #12269, opened 2026-07-19
automerge: move off the fragments prerelease tag onto stable 3.4.1; bump automerge-subduction (needs review)

dmaretskyi:
## Summary

Periodic dependency sweep — `@automerge/*` group. Branch rebuilt fresh on current `main` (2026-09-07). Re-checked npm state since the last pass (2026-08-16): `@automerge/automerge` and `@automerge/automerge-subduction` unchanged at `3.4.1`/`0.16.1` (still current `latest`); `@automerge/automerge-repo`'s `subduction` dist-tag advanced from `.47` to `.48`.

## What was bumped

| Package | Old | New | Notes |
|---|---|---|---|
| `@automerge/automerge` | `3.3.0-fragments.2` | `3.4.1` | Moves off a stale prerelease tag onto the current stable release. |
| `@automerge/automerge-subduction` | `0.16.0` | `0.16.1` | Patch file renamed (`patches/@automerge__automerge-subduction@0.16.1.patch`), reapplies unchanged — verified in `node_modules` post-install (browser export condition present). |

## What was deliberately left alone

| Package | Current pin | Why not bumped |
|---|---|---|
| `@automerge/automerge-repo` (+ `-network-broadcastchannel`, `-network-websocket`, `-storage-indexeddb`) | `2.6.0-subduction.40` | Still tracks the `subduction` dist-tag, now at `.48` (8 releases behind). Still carries the 556-line `patches/@automerge__automerge-repo@2.6.0-subduction.40.patch` keyed to this exact version. Rebasing a dist-level patch across upstream releases of a private fork is real engineering work — unchanged assessment from the prior three checks. |

## Validation

- **Install**: clean, patch verified present in `node_modules` post-install.
- **Tests**: `echo-host` full suite — 22/22 test files, 311 passed, 2 skipped, 1 todo, **zero failures** (an improvement over the last check's one flaky timeout, not reproducing this run). The 23-test `SubductionPolicy` suite that directly exercises `automerge-subduction`'s policy hooks passes in full.

## Why this needs review

- The group is still **incomplete** relative to its spec: `automerge-repo` and siblings intentionally not touched, same open question as before (bump to `.48` and re-verify the patch, or hold for dedicated attention).
- This is DXOS's forked core replication/sync engine — worth a human's eyes even though validation is clean.

Not auto-merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017W2LbzUkEU69kRj1dbz4AA

[...]

PR #12632, opened 2026-08-16
esbuild: bump esbuild to 0.28.2 (patch)

dmaretskyi:
## Summary

Periodic dependency sweep — `esbuild` group (spec: `esbuild "esbuild-plugin-*"`). Branch rebuilt fresh on current `main` (2026-09-07).

## Versions

| Package | Old | New |
|---|---|---|
| `esbuild` | 0.28.1 | 0.28.2 |

`esbuild-plugin-glsl`, `esbuild-plugin-inline-image`, `esbuild-plugin-raw`, `esbuild-plugin-wasm`, `esbuild-plugin-yaml` already at npm latest. (`esbuild-wasm` and `esbuild-style-plugin` don't match the group's `esbuild-plugin-*` glob, left alone.)

## Same gotcha as the last sweep

Bumping only the catalog entry has **no effect** — `pnpm-workspace.yaml`'s `overrides.esbuild: '>=0.28.0'` is a workspace-wide floor that always wins over the catalog specifier, and 0.28.1 already satisfied it. Raised the override to `'>=0.28.2'` too; verified via the lockfile (0 remaining `esbuild@0.28.1` resolutions, all consumers now on 0.28.2).

## Validation

- **Install**: clean.
- **Build**: `composer-app:build --force` — 391/391 tasks green.

## Classification: trivial

Patch bump, build green after re-raising the override floor. Enabling auto-merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017W2LbzUkEU69kRj1dbz4AA

## Summary by CodeRabbit

* **Chores**
  * Updated the build tooling to version 0.28.2.
  * Raised the minimum supported build tooling version to 0.28.2.
  * This keeps workspace tooling requirements aligned with the updated release.

PR #12634, opened 2026-08-16
esbuild: bump to 0.28.2

dmaretskyi:
## Summary

Bumps `esbuild` from `0.28.1` to `0.28.2` (patch release). All `esbuild-plugin-*` packages (`glsl`, `inline-image`, `raw`, `yaml`) were already at npm latest, unchanged. The `esbuild: '>=0.28.0'` override entry already satisfies the new version, no change needed there.

## Validation

- **Build:** ✅ green (`moon exec --on-failure continue --quiet :build`, exit 0). Along the way, found and cleaned up ~26 orphaned `node_modules/.bin/vite` shim scripts across unrelated packages (leftover from a much older, pre-drift dependency state, referencing a `vite@8.1.4`/`esbuild@0.28.1` resolution that no longer exists anywhere in the lockfile) — pre-existing environment cruft, not caused by this bump, but was blocking a clean build validation until removed. A plain `pnpm install` (even with `--force`, even after a full `node_modules` wipe and `pnpm store prune`) did not regenerate/prune these on its own.
- **Lint:** ✅ green (311/311 tasks)
- **Tests:** full-suite run shows only 2 failing packages, well within the established pre-existing baseline-noise range documented across this sweep (see dxos/dxos#11229, #11113, #12269) — none esbuild-related.

## Lockfile correction (caught by CodeRabbit review)

The first commit bumped the `pnpm-workspace.yaml` catalog entry, but a plain `pnpm install --no-frozen-lockfile` didn't propagate that into `esbuild`'s many transitive/deep resolutions (it's depended on internally by `vite`, `storybook`, etc.) — the lockfile still resolved `0.28.1` everywhere despite the catalog saying `0.28.2`. Fixed with `pnpm update esbuild --recursive --latest` to force real re-resolution (verified: 498 references now correctly at `0.28.2`, zero at `0.28.1`), then reverted several `package.json` files that command incorrectly rewrote from `"catalog:"` to a hardcoded `"^0.28.2"` — this repo's convention requires `catalog:` references. Rebuilt and confirmed green after the fix.

Also note: PR was briefly added to the merge queue with the broken (stale-lockfile) commit before this fix could be pushed — closed and reopened to force it out of the queue since GitHub's branch protection blocks pushes to a queued branch. No merge occurred.

## Classification: trivial

Patch bump, build+lint green, no regressions. Enabling auto-merge.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1807,7 +1807,7 @@ importers:
         version: 2.8.1
       vite-plugin-solid:
         specifier: 'catalog:'
-        version: 2.11.14(@testing-library/jest-dom@6.9.1)(solid-js@1.9.11)(vite@8.2.1(@types/node@22.10.2)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.8.2))
+        version: 2.11.14(@testing-library/jest-dom@6.9.1)(solid-js@1.9.11)(vite@8.2.1(@types/node@22.10.2)(esbuild@0.28.2)(jiti@2.7.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.8.2))
     devDependencies:
       '@changesets/changelog-git':
         specifier: 'catalog:'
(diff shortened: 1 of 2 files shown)


dmaretskyi:
Closing as a duplicate of #12632, which was opened ~30 minutes earlier for the identical `esbuild` 0.28.1→0.28.2 bump and independently found/fixed the same underlying lockfile issue (via the cleaner root-cause fix: raising the `overrides.esbuild` floor, rather than my `pnpm update --recursive` workaround — both land on the same correct end state, verified: 498 resolutions at `0.28.2`, zero at `0.28.1`). Consolidating on #12632 to keep one PR per group.

---

PR #12637, opened 2026-08-17
crawler, pipeline-discord: namespace-export the crawl stores and add service accessors

dmaretskyi:
Continues `dm/discord-pipeline` (its three commits are included here) and resolves the code-style TODOs left on that branch plus the `sql(…).stream` one.

## Namespace exports (`@import-as-namespace` markers)

`AgentRegistry`, `StateStore`, and `MessageStore` carried the bare directive. Converted following the `@dxos/ai` `AiService` pattern:

- directive moved above the imports,
- `XApi` → `interface Service`,
- barrels re-export with `export * as X from './X'`.

`QuestionStore` and `ExtractedQuestionStore` are converted alongside so `pipeline-discord/src/stores/` holds one convention, and the five store files are renamed to PascalCase as the lint rule requires (`message-store.ts` → `MessageStore.ts`, etc.).

## Service accessors (double-`yield*` TODO)

Each store now exports module-level accessors built on `Context.Service.use`, mirroring `@dxos/compute`'s `StorageService`:

```ts
export const count = (...args: Parameters<Service['count']>) => MessageStore.use((store) => store.count(...args));
```

Call sites read `yield* MessageStore.count()` instead of `yield* (yield* MessageStore).count()`. Applied through `Crawler`, `Stage`, the crawler and pipeline stages, and the demos/tests — `Crawler.commit` in particular loses its `Effect.flatMap(StateStore, …)` wrapper entirely.

## `sql(…).stream`

Not adopted, and the TODO is replaced by the reason. `sql\`…\`.stream` routes to `connection.executeStream`, which `@effect/sql-sqlite-node` implements as `Stream.die("executeStream not implemented")` — verified with a throwaway probe test against the same client the RDF source binds under node. Secondary blocker: `asynciterator`'s `wrap` accepts sync iterables/iterators, promises and EventEmitters, but not an `AsyncIterable`, so a `BufferedIterator` bridge would be needed even once a streaming client exists.

## Also

Deletes a stray empty `printf` file that landed at the repo root.

## Testing

```
moon run crawler:build pipeline-discord:build plugin-discord:build stories-brain:build pipeline-rdf:build   # 194 tasks, clean
moon run crawler:lint pipeline-discord:lint plugin-discord:lint stories-brain:lint pipeline-rdf:lint        # clean
CI=1 moon run crawler:test pipeline-discord:test pipeline-rdf:test plugin-discord:test stories-brain:test
  crawler          33 passed
  pipeline-discord 34 passed, 1 skipped
  pipeline-rdf     63 passed, 2 skipped
  plugin-discord    7 passed, 5 skipped
  stories-brain    45 passed, 20 skipped
```

`pnpm format` run; no new casts in the diff.

### Pre-existing failure, not from this change

`pipeline-discord` `replay-fixture.test.ts` fails locally with `expected 0 to be greater than 0` — the fixture `discord-crawl.db` committed in `1674f3c7` holds 10 messages, none of which the sentence-level detector reads as a question. Confirmed identical on the branch head before these changes. The test is `skipIf(CI)`, so it does not affect the Check workflow.

### Not addressed

Two TODOs from the base branch are left in place, both outside the code-style scope:

1. `deterministicAiService` → `MockAiServiceLayer` in `@dxos/ai` — a package move with an open design question (unifying with `ScriptedLanguageModel`).
2. `scripts/secrets.mjs` → the `dxos/edge` script + `secrets.jsonc`.

---
_Generated by [Claude Code](https://claude.ai/code/session_012hD8QVZ9zpciwpMakQ5H6U)_

## Summary by CodeRabbit

* **New Features**
  * Added question storage for adding, answering, listing, and tracking question attempts.
  * Added direct access helpers for crawler, message, question, agent, and state operations.
  * Added optional real AI extraction support in the crawl demo when configured.

* **Documentation**
  * Documented SQLite query materialization and planned secrets configuration updates.

* **Refactor**
  * Simplified service access and module exports across crawling, replay, pipelines, plugins, and dashboards without changing behavior.


dmaretskyi:
Triaged the CodeRabbit review. No code changes — details below, one item deferred to @dmaretskyi.

**Deferred, not mine to decide — `topics.ts:147`, logging `summary` at info level.**
The privacy point is fair: `summary` is an LLM paraphrase of real chat content, and it widens what the info log carries beyond the `name` and `participantLabels` that were already there. But this line comes from `e4aba0e9` on the base branch ("Added summary logging for topics") — a deliberate diagnostic, not part of this PR's refactor. Leaving it for the author; the fix direction (drop it, drop to debug level, or keep) is a product call.

**Won't fix — `crawl-demo.test.ts:10-19`, builtin imports before external.**
`oxfmt` owns import ordering here and sorts `node:*` alphabetically inside the external group. I applied the suggested order and re-ran the formatter; it reverted to the current arrangement byte-for-byte. Committing the suggestion would fail CI's `oxfmt --check`.

**Won't fix — `crawl-demo.test.ts:8`, terminate the comment with a period.**
That line is a shell invocation inside the file's usage header, not prose; the three command lines above it end the same way. Also from `e4aba0e9`.

**Not applicable — JSDoc on `Crawler.stream`, `Crawler.commit`, `Stage.tapStage`.**
All three already carry JSDoc blocks (`Crawler.ts:153-157`, `Crawler.ts:188-193`, `Stage.ts:13-17`); this PR only changed their bodies and signatures.

**Won't fix — `MessageStore.test.ts:13`, import via the `stores` barrel.**
A co-located test importing its sibling module directly is the pattern throughout both packages (`AgentRegistry.test.ts`, `StateStore.test.ts`, `QuestionStore.test.ts`), and it was the pre-existing form in this file. The namespace form satisfies the `@import-as-namespace` rule, which is what the directive requires.

**Out of scope — comment length in `agent-profile.ts:15-19` and `crawl-stores.ts:22-24`.**
Both are unchanged pre-existing comments, outside this diff.

---

packages/core/compute/crawler/src/testing/index.ts
```diff
@@ -271,25 +272,24 @@ export const deterministicAiService = (): Layer.Layer<AiService.AiService> =>
  * State + agent registry + semantic store, with NO `Source` and NO `AiService`. Compose with a
  * `Source` layer and an `AiService` (the deterministic stand-in or a real model) to run the crawler.
  */
-export const coreLayer: Layer.Layer<StateStore | AgentRegistry | FactStore> = Layer.mergeAll(
-  StateStore.layerMemory,
-  AgentRegistry.layerMemory,
+export const coreLayer: Layer.Layer<StateStore.StateStore | AgentRegistry.AgentRegistry | FactStore> = Layer.mergeAll(
+  StateStore.StateStore.layerMemory,
```
dmaretskyi: don't double the name, move layer constructurs as module-level export -- create .mdl rule for this too

dmaretskyi:
Done in 978a80be.

**Layer constructors hoisted to module scope.** Every `static layerMemory` / `static layerSql` moves out of the tag class, which is now an empty body:

```ts
export class StateStore extends Context.Service<StateStore, Service>()('@dxos/crawler/StateStore') {}

export const layerMemory: Layer.Layer<StateStore> = Layer.sync(StateStore, () => makeMemory());
export const layerSql: Layer.Layer<StateStore, never, …> = Layer.effect(StateStore, …);
```

Call sites lose the doubling: `StateStore.StateStore.layerSql` → `StateStore.layerSql`, across all five stores and every consumer.

The tag still doubles in *type* position (`Layer.Layer<StateStore.StateStore>`), which I left alone — the class is genuinely the type, matching `AiService.AiService`, and there's nowhere else for it to live. Operations were already reachable without doubling via the module-level accessors from the previous commit.

**Also swept `@dxos/pipeline`'s `Progress`**, which carried the same `static layer` and was the only other place in the repo repeating a namespace this way (`Progress.Progress.layer` in the stories harness). Changeset added — it's a published package, so the move is consumer-visible.

**Rule added** — `.agents/skills/agentic-review/rules/code-style.mdl`, a new `code-style` rules document alongside `non-negotiables.mdl`:

```
rule namespace-service-layers: Layer constructors are module-level exports, never class statics
  … Flag a `static layer…` member on a `Context.Service` subclass, and flag any
  call site that repeats a namespace as `Name.Name.member`. A bare `Name.Name`
  in a TYPE position is not a violation …
  grep: static layer|([A-Z]\w+)\.\1\.
  severity: warn
```

Verified it loads through the harness's `loadRules`, and that a repo-wide grep for the pattern now returns only `Schema.Schema.Type` (Effect's own API).

Build, lint and tests green across `pipeline`, `crawler`, `pipeline-discord`, `plugin-discord`, `stories-brain` — 140 passed. `check-changeset-count` and `check-changeset-bumps` both pass.

---

dmaretskyi:
Both data-integrity findings were real. Fixed in a9d57804.

**`ExtractedQuestionStore.layerMemory.put`** overwrote on a repeated `(messageId, question)`; the SQL layer's `ON CONFLICT … DO NOTHING` keeps the first record. Memory now keeps the first too.

**`QuestionStore.layerMemory.add`** replaced the stored question on a duplicate id; `id TEXT PRIMARY KEY` makes the SQL layer fail the insert. Memory now fails with the same `StoreError`.

Rather than take the suggestions blind, both cases are now pinned by the shared store suites, which run each assertion against `layerMemory` **and** SQLite — so the implementations can't drift apart silently again. `ExtractedQuestionStore` had no suite of its own (the only store in the directory without one), so it gets one in the sibling style.

Confirmed the tests actually pin the behaviour: reverting just the two store fixes and re-running gives

```
× re-putting a (message, question) pair keeps the first record
× a duplicate id fails rather than replacing the stored question
Tests  2 failed | 14 passed (16)
```

Exactly two failures, both on the memory variants — the SQLite variants passed before and after, which is the independent confirmation that SQL was the correct side of each divergence.

`pipeline-discord` now 40 passed / 1 skipped; build and lint clean; `crawler`, `pipeline`, `plugin-discord`, `stories-brain` unchanged and green.

**Not taking** the `Progress.ts` nitpick (`/** A fresh in-memory registry. */` → state the isolation constraint). It's a pre-existing comment this PR only relocated, it's accurate, and every sibling `layerMemory` in the diff carries a similarly-shaped one-liner — rewording this single instance would make the set inconsistent for no reader benefit.

---

PR #12639, opened 2026-08-17
moon: build only what tests cannot import from source

dmaretskyi:
Applies the `composer-app:serve` narrowing to the test tasks: vitest resolves `@dxos/*` through the `source` export condition (`PluginImportSource` in `vite.base.config.ts`), so `^:build` on `:test` / `:test-browser` / `:test-storybook` / `:test-workerd` compiles dist that the test run never reads.

## Approach

A flat copy of `serve`'s original spelling (`#dist-runtime:build` + `#vite-plugin:build`) does not transfer: those tags reach every tagged project in the repo, whose build closure is 92 projects — *more* than the median test project needs today, and enough to make a `client-services` edit invalidate all 270 test tasks.

Instead, `.moon/tasks/all.yml` gains an inherited `noop` aggregator, `source-deps` — "what a source-importing consumer needs from me, and nothing more". It resolves to:

- `~:build` for packages that cannot be imported from source, via three tags collected in `.moon/tasks/tag-dist-consumed.yml` — `dist-runtime` (publishes no `source` condition), `vite-plugin` (exports a bundler plugin a config loads), `dist-config` (anything else a config file imports);
- `~:prebuild` for the three packages whose own `src/` is partly generated and gitignored;
- nothing at all otherwise.

Each test task then depends on `^:source-deps`, so the built set is scoped to the project's own dependency closure rather than the tag set repo-wide. `composer-app:serve` and `gen-optimize-deps` move onto the same aggregator.

## Audit

Membership was derived rather than guessed. Every `@dxos/*` and `#*` specifier the repo imports (28,645 importer+specifier pairs over 11,024 files) was resolved against its package's export map under both node and browser condition sets, and checked for a target under `dist/`. That found four packages consumed as dist but untagged — `sql-sqlite`, `vendor-hypercore`, `vendor-kbn-handlebars`, and `echo-query`'s `#query-lite` — matching an empirical control: with nothing built, `echo-host`'s suite fails on exactly `@dxos/protocols` and `@dxos/sql-sqlite`.

Four hazards, all verified:

1. **Cache correctness.** `^:build` is what folds the upstream source closure into a downstream `:test` hash. Since moon 2.3 a dependency declaring no `outputs` defaults to `cacheStrategy: ignored`, so a plain `noop` chain runs and invalidates nothing — confirmed by observing `util:source-deps` change hash while `echo-host:test` stayed a cache hit. With `cacheStrategy: hash` spelled out: clean `d6d8e8fa` → cache hit on rerun → edit `util/src/index.ts` → `7c7d67af`, suite re-runs.
2. **Generated source.** Six projects generate into `src/`. `deus`, `echo-query`'s lezer output and `plugin-onboarding`'s exemplar are tracked; `protocols` and `observability` are `dist-runtime` so their `build` pulls `prebuild`; `react-ui-gameboard` (`src/gen`) and `react-ui-experimental` (compiled shaders) are gitignored and get explicit `~:prebuild`. Those last two were also a pre-existing hole in `serve`'s narrowing, which this closes.
3. **Config-load-time resolution.** Vite bundles `vite.config.ts` / `.storybook/main.*` in Node before any plugin runs, so `source` is not among the conditions there however the export map reads — and `await import()` does not dodge it, because `externalize-deps` resolves at bundle time. A second scan over config files with `source` excluded found `agent-claude` and `fixtures`; both now carry `dist-config`. `@storybook/addon-vitest` evaluates `.storybook/main.*` during vitest's *project resolution*, before `--project` filtering, so this applies to a consumer's plain `test` task too.
4. **Suites that read built artifacts by design.** `env-tests/src/build.test.ts` asserts what the published bundle contains — it runs its own `esbuild.build()` under workerd/worker/browser conditions, deliberately without `source`. The `cli` suites spawn `./bin/dx`, which resolves the standard way (`--conditions=source` is opt-in behind `DX_SOURCE=1` because third-party packages advertise a `source` field they do not ship). And `app-solid` is the only `ts-test` project whose `vitest.config.ts` is standalone — it takes precedence over its `vite.config.ts`, so `PluginImportSource` never loads. All three keep `^:build`.

`typecheck` deliberately keeps `^:build` too: `tsconfig.base.json` sets no `paths` or `customConditions`, so `tsgo` resolves `@dxos/*` through the `types` condition into `dist/types`.

## Effect

Distinct upstream `build` actions for a whole-repo run, from `moon action-graph`, before → after:

| target | before | after |
| --- | --- | --- |
| `:test` | 273 | 194 |
| `:test-browser` | 268 | 140 |
| `:test-storybook` | 270 | 99 |
| `:test-workerd` | 7 | 5 |
| all four combined | 273 | 194 |

No project builds more than it did before. `composer-app:serve` goes 108 → 110, the two added being the missing prebuilds. The ~271 `noop` actions cost ~1s on a fully cached `echo-host:test`.

## Test

Full local sweep of `:test` + `:test-browser` (`moon exec --on-failure continue`) from a checkout with **every `dist/` and generated-`src` dir deleted**, so only what `^:source-deps` builds exists — 278 targets. Storybook and workerd suites verified separately, plus targeted runs covering each distinct dist-consumed package: vendor bundles, `client-services`/`shell`/`config`, `lit-ui`/`lit-grid`, the `ui-theme` config-load plugin, and `echo-query`'s `api.d.ts?raw`.

Worth noting for reviewers: CI's remote cache can hydrate a stale pass, so a green CI job is not by itself evidence here — `cli:test` failed locally from cleared dist while CI reported it cached and green. The sweep above is the load-bearing check.

Two failures the sweep surfaced that are **not** caused by this change:

- `plugin-inbox` `analyze-mailbox.test.ts` — 3 cascade assertions expecting status `'failed'`, getting `'completed'`. Reproduced identically with the base `^:build` restored, so it is red on main.
- `client-e2e` `post and listen to messages` and `echo-host`'s `authorizeFetch … (empirical)` — both pass in isolation and fail only under heavy concurrency. The former hard-codes a 200ms budget for two pub/sub round-trips while its siblings in the same file use `{ timeout: 3_000, retry: 1 }`; CI passes it in 468ms. Left alone rather than loosened.

No changeset: build tooling only, nothing consumer-observable.

https://claude.ai/code/session_01Q5u7mZZbDLgQrPyHiGcPyx

## Summary by CodeRabbit

* **Improvements**
  * Improved dependency handling for source-based workflows, ensuring required generated files and runtime artifacts are available before use.
  * Streamlined task execution with more targeted dependency tracking and hash-based caching, reducing unnecessary work.
  * Improved reliability of browser, Storybook, CLI, environment, and package tests by ensuring required builds complete first.
  * Enhanced package resolution for runtime and platform-specific imports.


.moon/tasks/tag-ts-test-browser.yml
```diff
@@ -7,7 +7,10 @@ tasks:
       - node-std:build
       - vite-plugin-icons:build
       - vite-plugin-import-source:build
-      - ^:build
+      # Narrowed like `:test` — the browser project resolves `@dxos/*` to source too. See
```
wittjosiah: clean up comments on every usage, centralize documentation

packages/apps/composer-app/moon.yml
```diff
@@ -128,30 +128,24 @@ tasks:
...
+      # `source-deps` (`.moon/tasks/all.yml`) is that narrowed set, shared with every `:test*` task and
+      # scoped to this app's own dependency closure. It has to cover generated-but-gitignored source
+      # (`react-ui-gameboard`'s `src/gen`, `react-ui-experimental`'s compiled shaders) as well as
+      # dist, because vite's dependency scan aborts wholesale — pre-bundling nothing at all — when a
+      # single import in the crawl fails to resolve.
+      - ^:source-deps
+      # Not reachable through `source-deps`: `vite.config.ts` wires the agent sidecar middleware via
+      # `await import('@dxos/agent-claude')`, and vite's `externalize-deps` resolves that while
+      # bundling the config in Node, where the `source` condition does not apply.
+      - agent-claude:build
```
wittjosiah: this being special-cased in a bunch of places is suspicious, should be incorporated somehow into source-deps

.moon/tasks/tag-ts-test-browser.yml
dmaretskyi: Done in 29ee958e. The four `:test*` tasks now just carry the bare dep, and the documentation lives in two places: `.moon/tasks/all.yml` explains the task (why not `^:build`, and why `cacheStrategy: hash` is required), and a new `.moon/tasks/tag-dist-consumed.yml` is the single record of *which* packages can't be imported from source and why. It replaces `tag-dist-runtime.yml` and `tag-vite-plugin.yml`, so there's one file per concept instead of a comment per usage.


---

packages/apps/composer-app/moon.yml
dmaretskyi: Agreed, and fixed in 29ee958e — the special-casing was me hand-writing what a tag already expresses.

`agent-claude` and `fixtures` were listed on five separate tasks (`composer-app:serve`, plus `test` and `test-storybook` on `stories-assistant` and `stories-inbox`). They're both just config-load-time dist consumers: vite bundles `vite.config.ts` / `.storybook/main.*` in Node before any plugin runs, so `source` isn't among the conditions there however the export map reads — and `await import()` doesn't dodge it, because `externalize-deps` resolves the specifier at bundle time. So they now carry a `dist-config` tag and all five task deps are gone. `gen-optimize-deps` also moves onto `^:source-deps`; it was still spelling the narrowing out via `#dist-runtime:build` + `#vite-plugin:build` + `echo-query:prebuild-query-lite`.

Two notes on what I didn't collapse:

1. `vite-plugin` stays a separate tag from `dist-config` even though they behave identically here, because `scripts/pkg-lint.mjs` keys `build-tool-source-export` off it to disambiguate an overloaded `./plugin` export. I did try renaming it and backed that out.
2. Generated-but-gitignored `src/` (`react-ui-gameboard`, `react-ui-experimental`, `echo-query`) still overrides `source-deps` with `~:prebuild` per project, since what a source import needs there is the codegen, not a build. That's the one remaining per-project form, and it's now documented as such in the tag file.

Build-neutral: whole-repo `:test` resolves the same 194 upstream builds before and after, none added or removed. Verified from a cleared `dist/` that `stories-assistant`, `stories-inbox` (both `test` and `test-storybook`), `react-ui-form`, and `storybook-utils` all still pass.


---

.moon/tasks/tag-ts-test-browser.yml
```diff
@@ -7,7 +7,8 @@ tasks:
       - node-std:build
       - vite-plugin-icons:build
       - vite-plugin-import-source:build
-      - ^:build
+      - target: ^:source-deps
```
dmaretskyi, line 10: ca this be expressed as someting like ^#source-deps:build --- run build for all "#source-deps" packages in transitive closure of this package deps?

I'm trying to avoud new task verb defitninio for source-deps

.moon/tasks/tag-dist-consumed.yml
```diff
@@ -0,0 +1,43 @@
...
+# A package whose own `src/` is partly generated is NOT covered here — it overrides `source-deps` with
+# `~:prebuild` in its own `moon.yml`, because the codegen, not a build, is what a source import needs.
+tasks:
+  source-deps:
+    deps:
+      - ~:build
+
+inheritedBy:
+  tags:
+    - dist-runtime
```
dmaretskyi, line 45: do we actually need to distinguish between thore 3?

.moon/tasks/tag-ts-test-browser.yml
dmaretskyi: I wanted that too and went looking first — moon can't express it, and I tested rather than trusting the docs. Three spellings fail to parse outright:

```
^#dist-runtime:build   => Error: config::parse::failed
#dist-runtime^:build   => Error: config::parse::failed
^dist-runtime:build    => Error: config::parse::failed
```

`^:#tag` *does* parse, but it's a different feature and doesn't do the job:

1. It filters **task** tags, not project tags. I gave `protocols`' `build` task a `tags: [needs-dist]` and `^:#needs-dist` from `echo-protocol` correctly resolved to `protocols:build`, while a bogus tag gave `Depends on: —`. So the mechanism is real.
2. But it is **direct-deps only, not transitive**. `echo` → `echo-protocol` → `protocols`: from `echo`, `^:#needs-dist` resolves to nothing. The transitive reach in the current design comes from `source-deps` recursively depending on `^:source-deps`, which is exactly the part a selector can't do.

There's also a second reason it has to be task-shaped, independent of syntax. `source-deps` isn't only picking which builds to run — it's carrying the **upstream source hashes**. `^:build` is what currently folds the whole transitive source closure into a downstream `:test` hash. Any selector that depends only on the *tagged* packages' builds drops every untagged upstream package's source out of the hash, and `:test` then replays a stale pass after an upstream edit. I hit precisely that on the first attempt: a plain `noop` chain ran but invalidated nothing, because since moon 2.3 a dep with no `outputs` defaults to `cacheStrategy: ignored`. Hence the explicit `cacheStrategy: hash`.

So even given a transitive tag selector, we'd still need something task-shaped alongside it to keep the cache honest — at which point the selector is redundant. Happy to be wrong if you know a moon feature I've missed; the `noop` costs ~1s across ~271 actions on a fully cached run, so it's cheap, just an extra name.


---

.moon/tasks/tag-dist-consumed.yml
dmaretskyi: No — collapsed in b66ee551. `dist-config` is gone and `agent-claude` / `fixtures` just carry `dist-runtime`, so no new tag name is introduced at all.

The consequence was identical for all of them (consumers must read the package's dist → `source-deps` is its real `build`), and the reason only matters when deciding whether a fix is *possible*: a missing `source` condition can be added, whereas a config file importing the package can't be fixed from the package side. That's now two numbered notes in the tag file instead of two tags.

`vite-plugin` still appears in `inheritedBy` — a bundler plugin a config loads is the second case by definition, but the tag isn't mine to merge away: `scripts/pkg-lint.mjs` keys `build-tool-source-export` off it to disambiguate an overloaded `./plugin` export. I did try folding it in earlier and reverted when I found that dependency. The alternative would be double-tagging those 7 packages, which seemed worse.

Behaviour-identical: whole-repo `:test` resolves the same 194 upstream builds, and `protocols` / `ui-theme` / `agent-claude` / `fixtures` all still resolve `source-deps` to their own `build`.


---

dmaretskyi: Correcting myself: I answered "no" above and that was wrong. Collapsing them broke `pkg-lint`, and the reason is a good one — **yes, we do need to distinguish**, and the repo already enforces it. Reverted in 1f57302f.

`pkg-lint`'s `dist-runtime-tag` rule makes the tag an *error* on a package that publishes a `source` condition:

> `"dist-runtime" tag requires no "source" export condition — with one, vite resolves this package from source and the tagged build is unused`

`agent-claude` and `fixtures` both publish `source`. They resolve from source perfectly well in general; only the config-load path reads dist. So `dist-runtime` is factually false for them, and the failing `check` job was right to say so.

That is a sharper justification than the one I gave earlier, so the tag file now leads with it: the tags look interchangeable from `source-deps`' point of view because the consequence is identical, but what a package can *do* about being on the list differs — a missing `source` condition can be added, whereas a config importing the package cannot be fixed from the package side — and pkg-lint enforces two of the three.

`pkg-lint` exits 0 locally now, and the graph is unchanged: same 194 upstream builds for `:test`, with `protocols` / `ui-theme` / `agent-claude` / `fixtures` all still resolving `source-deps` to their own `build`. Reopening this thread since the answer inverted.


---

PR #12645, opened 2026-08-17
deck: keep the companion open across a nav-tree navigation

wittjosiah:
## Problem

Clicking an item in the nav tree closed the companion pane.

Companion state is per plank (`DeckState.companionPlanks`) and `computeActiveUpdates` prunes it to the planks that stay open. A nav-tree click is a `solo` `LayoutOperation.Open`: the deck becomes just the subject, so the plank the companion was open on leaves `active` and its entry is pruned. In solo mode that means *every* navigation closes a companion the user never touched — and the lost width makes the browser clamp the deck's scroll into a one-frame snap.

`open.ts` already carried the companion across a **level** swap (`resolveLevelOpen().replacedId`, e.g. mailbox → message). The plain navigate path had no equivalent.

## Change

- New pure helper `carryCompanion` in `util/companion-anchor.ts`: given the pruned companion list, the pre-open list, the replaced plank and its replacement, hand the entry to the replacement. It also de-dupes, which the old inline level-swap expression did not.
- `operations/open.ts` now tracks a single `replacedId` for both replacing paths — `levelOpen.replacedId`, and for a navigate the companion anchor of the outgoing deck (`resolveCompanionAnchor(deck.active, attended)`, the same anchor the companion is rendered and serialized against). A navigate whose subjects are all already open still only scrolls into view and replaces nothing.
- `DESIGN.md` §4 records the carry-over.

Unchanged: `add`/`auto` opens that grow the deck never open a companion, and a companion explicitly closed on the outgoing plank stays closed.

## Test

`carryCompanion` unit tests in `util/companion-anchor.test.ts` cover carry, closed-stays-closed, replaces-nothing, surviving planks keeping their own state, and the de-dupe.

```
✓ |node| src/util/companion-anchor.test.ts (21 tests) 31ms
 Test Files  10 passed (10)
      Tests  97 passed (97)
```

(full `plugin-deck` node suite)

https://claude.ai/code/session_016vEa1TnvYQww38X4hkn78P

---
_Generated by [Claude Code](https://claude.ai/code/session_016vEa1TnvYQww38X4hkn78P)_

### DIFF
--- a/.changeset/companion-survives-navigation.md
+++ b/.changeset/companion-survives-navigation.md
@@ -1,0 +1,5 @@
+---
+'@dxos/plugin-deck': patch
+---
+
+Keep the companion open when navigating from the nav tree: the companion now follows the plank the new one replaces instead of closing on every navigation.
--- a/packages/plugins/plugin-deck/DESIGN.md
+++ b/packages/plugins/plugin-deck/DESIGN.md
@@ -125,6 +125,9 @@ never moves it.
   to its mailbox.
 - **Per plank, not per deck** — `DeckState.companionPlanks` lists the planks showing their companion,
   so moving between planks restores what each was left in.
+- **Carried across a replacing open** — a navigate (or a level swap) closes the plank the companion was
+  open on, which would prune its entry and close a companion the user never touched. `carryCompanion`
+  hands the entry to the plank standing in for it, so the pane survives nav-tree navigation.
 - **Sizing** — the tile is the plank's own width _plus_ the companion beside it, and the companion's
   width is held deck-wide under `COMPANION_SIZE_KEY` (not a valid node id, so it cannot collide).
   Opening or closing the companion therefore never resizes the plank. Dragging the seam commits both
--- a/packages/plugins/plugin-deck/src/operations/open.ts
+++ b/packages/plugins/plugin-deck/src/operations/open.ts
@@ -21,7 +21,13 @@ import * as ObservabilityOperation from '@dxos/plugin-observability/Observabilit
 import { DeckCapabilities } from '#types';
 
 import { addSubjectsToActiveDeck, resolveLevelOpen, resolveSeededPlanks, updatePlankNames } from '../layout';
-import { computeActiveUpdates, openableChildren, resolveDeckSpec } from '../util';
+import {
+  carryCompanion,
+  computeActiveUpdates,
+  openableChildren,
+  resolveCompanionAnchor,
+  resolveDeckSpec,
+} from '../util';
 import { updateActiveDeck } from './helpers';
 
 const handler: Operation.WithHandler<typeof LayoutOperation.Open> = LayoutOperation.Open.pipe(
@@ -130,9 +136,6 @@ const handler: Operation.WithHandler<typeof LayoutOperation.Open> = LayoutOperat
       // Holding shift forces any disposition into an add (callers forward the raw modifier rather than
       // encoding the policy). Only 'auto' falls back to the attended plank; a shift-forced add from the
       // nav-tree (a 'solo' gesture with no pivot) appends at the end.
-      const navigateSolo = (active: readonly string[]): string[] =>
-        input.subject.every((id) => active.includes(id)) ? [...active] : [...input.subject];
-
       let previouslyOpenIds: Set<string>;
       {
         const deck = yield* DeckCapabilities.getDeck();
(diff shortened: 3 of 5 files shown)


wittjosiah:
## CI status — `storybook` red on a pre-existing `plugin-inbox` failure, not this PR

`check` ✅ · `test` ✅ · `build` ✅ · `workerd` ✅. The only red job is `storybook`, and both failures are in `plugin-inbox`:

- `src/containers/MessageArticle/MessageArticle.stories.tsx > Spec` — `TestingLibraryElementError: Unable to find an element by: [data-testid="conversation.summary"]`
- `src/components/Header/Header.stories.tsx > Spec`

### Why it is not this change

1. `plugin-inbox` does not depend on `plugin-deck`, directly or transitively — `moon project plugin-inbox` lists no such dependency and the package's source contains no reference to it. This PR touches only `plugin-deck` (plus `DESIGN.md` and a changeset).
2. `git diff origin/main HEAD -- packages/plugins/plugin-inbox` is **empty** — the package is byte-identical to `main` on this branch.
3. `MessageArticle > Spec` **reproduces locally on this branch**, deterministically, in the same assertion — 3 moon attempts plus vitest's own retry:

```
❯ src/containers/MessageArticle/MessageArticle.stories.tsx (2 tests | 1 failed)
   ✓ Default 71ms
   × Spec 23800ms (retry x1)
```

Since (2) makes the tested content identical to `main`, that reproduction is a reproduction on `main`.

`Header > Spec`, by contrast, **passes** locally (`✓ Spec 3029ms`), so it looks load-sensitive rather than broken — CI runs every package's storybook suite concurrently at `MOON_CONCURRENCY=4`. Both tests are registered Trunk flaky-test entries.

Note that `main`'s own `storybook` job at this PR's base (`df0ab571`) reports green, but its "Test Storybook All" step completed in 23 seconds — a full moon cache replay, not a live execution — so it is not evidence these stories pass.

### Why I am not fixing it here

The failing assertion waits on the summary tile rendered by `ConversationStack.tsx:429`, i.e. the mailbox summary path — unrelated to this PR's subject (deck companion state) and inside another feature's area. Fixing it here would mean an unrelated change to `plugin-inbox` in a deck bugfix PR. Flagging it for whoever owns the inbox summary work instead.

---

wittjosiah:
## Root cause of the `plugin-inbox` storybook failure (still pre-existing, still not from this PR)

Following up my earlier comment with an actual diagnosis, for whoever owns inbox. `MessageArticle.stories.tsx > Spec` fails because **the mailbox's `annotations` ref is never observed by the article**, so `conversationSummary` is always `undefined` and the tile never mounts.

Instrumented `seedSummaries` and `MessageArticle` and ran the story:

```
DXPROBE seed    {"messages":3,"targets":1,"fraction":0.5,"ratio":0.5}
DXPROBE article {"annotationsRef":false,"annotationsFeed":false,"annotations":0,"parents":[],"messages":3,"ids":[...]}
```

So the seed **succeeds** — it finds all 3 messages, selects 1, and therefore reaches `Mailbox.findOrCreateAnnotations`, which does:

```ts
const feed = db.add(Feed.make());
Obj.setParent(feed, mailbox);
Obj.update(mailbox, (mailbox) => { mailbox.annotations = Ref.make(feed); });
```

…yet **every** article render sees `!!mailbox.annotations === false`. Not "ref present but unresolved" — the property is absent. With no ref there is no annotations query, no annotations, no summary, no tile.

The same sequence against an `EchoTestBuilder` database in Node **does** leave `mailbox.annotations` set and `Mailbox.conversationSummary` resolves it, at both `length: 3` and `length: 8`. The divergence is browser/worker-backed database vs. in-process, which points at the lazily-created optional `Ref` not persisting (or not being observed) on the worker path — plausibly related to #12575, which changed `FeedAnnotation` to name the property holding the feed reference and is the most recent change in this area. `mailbox.feed` (created eagerly in `Mailbox.make()`) resolves fine; only the lazily-provisioned `annotations` feed fails, which fits.

Worth noting this is not only a story problem: if the ref genuinely does not stick on the worker path, a real mailbox would never show a derived summary either, since `SummarizeMailbox` provisions the feed the same lazy way.

**Hypothesis I tested and rejected:** that the seed's read-back query raced the append (no `flush` between `initializeMailbox` and `seedSummaries`). Adding `await db.flush({ indexes: true })` before the query changed nothing — `messages: 3, targets: 1` with and without it. Not committed.

No inbox changes are in this PR; all instrumentation was reverted.

---

PR #12647, opened 2026-08-18
build: upgrade to TypeScript 7

dmaretskyi:
Moves the workspace onto released **TypeScript 7.0.2**. The repo was already typechecking and emitting declarations with the native compiler, but via `@typescript/native-preview@7.0.0-dev.20260421.2` — a package whose last publish was `7.0.0-dev.20260707.2`, frozen now that 7.0 shipped. TypeScript 7 *is* that native compiler, so `tsc` replaces `tsgo` and the preview package is dropped.

## The API split

TypeScript 7 ships no JavaScript compiler API — one is planned for 7.1 — so anything calling `ts.createProgram`, `ts.factory`, `ts.transpileModule` and friends moves to Microsoft's [`@typescript/typescript6`](https://www.npmjs.com/package/@typescript/typescript6) compatibility package (bin `tsc6`, so no collision with TypeScript 7's `tsc`):

| Consumer | What it uses the API for |
| --- | --- |
| `@dxos/protobuf-compiler` | protobuf → TS codegen (peer + dev dependency) |
| `@dxos/plugin-code`, `@dxos/plugin-script` | in-browser script compilation, alongside `@typescript/vfs` |
| `dx-build` | reads `outDir` out of `tsconfig.json` before cleaning |
| `scripts/api-diff.mjs`, `tools/codemods/*.mjs` | AST analysis and codemods |

Third-party tooling that calls the API — typescript-eslint, typedoc, knip, madge, react-docgen, rolldown-plugin-dts, `@typescript/vfs` — is held on the 6.x line by a `.pnpmfile.cjs` rule that converts its `typescript` peer into a real dependency. This replaces the blanket `overrides: typescript: 'catalog:'`, which cannot express the split: pnpm overrides rewrite a workspace project's own direct dependencies too, so a single pin forces one version on the compiler and the ecosystem alike (verified — it silently rewrote the root's `^7.0.2` back to `^6.0.3`).

The one latent bug this surfaced: `@valtown/codemirror-ts` ships `import type ts from 'typescript'` in its `.d.ts` but lists `typescript` only as a *devDependency*, so it had been resolving through `@dxos/plugin-script`'s own dependency. Without that it degrades to `any` and `TypescriptEditor.tsx` fails with `TS7031`; `.pnpmfile.cjs` now gives the package the real dependency it should have declared.

## Other changes

- `.moon/tasks/tag-typecheck.yml` runs plain `pnpm exec tsc --noEmit`. The `DX_USE_TSC` switch is gone — nothing in the repo set it, and it existed only to choose between the preview native compiler and the classic one, a choice that stops making sense once `tsc` *is* native. `@typescript/typescript6` stays for the API; only its `tsc6` bin goes unused.
- `DxTsgoPlugin` → `DxDeclarationsPlugin` in `vite.base.config.ts`, since nothing is named `tsgo` any more.
- `tsdown` `^0.16.7` → `^0.22.14`: `rolldown-plugin-dts` only learned to drive the TypeScript 7 compiler in `0.27`, and `@dxos/echo-query` emits its `query-lite` declarations through it.
- `tools/dx-build` gains an explicit `dependencies` block rather than relying on root hoisting for its compiler API.
- `.vscode/settings.json`: drops `typescript.tsdk` (TypeScript 7's language service speaks LSP and ships no `tsserver.js`) and repoints `typescript.native-preview.tsdk` at `node_modules/typescript`. This is the one change not verified locally — it needs an editor.

## Verification

- `pnpm install --frozen-lockfile` resolves clean under `strictPeerDependencies`, with exactly two TypeScripts in the graph: `typescript@7.0.2` for the workspace and `typescript@6.0.3` behind the compat package and the API tooling. `@typescript/typescript6` resolves `@typescript/old` to the same physical `typescript@6.0.3` that `@typescript/vfs` gets, so declaration identity holds across the boundary.
- `moon exec --on-failure continue :build` across all 328 projects — green, zero TypeScript errors.
- `pnpm format-check` clean.
- Confirmed no tsconfig in the repo uses an option TypeScript 7 removed (`target: es5`, `downlevelIteration`, `moduleResolution: node10`/`classic`, `module: amd`/`umd`/`systemjs`, `baseUrl`), and that `experimentalDecorators`, `module: Preserve`, `moduleResolution: Bundler`, `rewriteRelativeImportExtensions` and project references all still work.

## Not included

`packages/apps/composer-app/src/vite/optimize-deps.ts` is generated and still lists a bare `'typescript'` while omitting `@typescript/typescript6`, which the script plugins now import. Nothing breaks — the app just discovers the compat package mid-session and re-optimizes, which is what that file exists to prevent. Regenerating it is `moon run composer-app:gen-optimize-deps`; left out to keep this PR scoped to the toolchain.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KQo6mjku1D2WYjM3P4jnyP)_


dmaretskyi:
Acted on the inline finding: `vite.base.config.ts` no longer names a compiler in its failure message. The selection lives in `dx-build`, which streams the compiler's own diagnostics through, so duplicating the `DX_USE_TSC6` contract in a second file would just create a second thing to keep in sync — the message is now plain `dx-build failed`.

Skipping the two nitpicks, with reasons:

1. **Import order in `tools/dx-build/src/main.ts`** — that is oxfmt's output, not a hand-written order. `oxfmt` sorts the whole leading import run alphabetically, which puts `@typescript/typescript6` ahead of the `node:*` specifiers, and CI gates on `oxfmt --check` — so applying the suggested order would fail the `check` job. Verified: `pnpm format-check` reports "All matched files use the correct format" on the current order.

2. **`substitutions-parser.ts` using `ts-morph`'s compiler instance** — out of scope here. That file already imported `ts-morph` and the compiler API separately before this PR (`import { Project, type Symbol, type TypeChecker } from 'ts-morph'` next to `import * as ts from 'typescript'`); this diff only renames the specifier to `@typescript/typescript6`, which resolves to the same `typescript@6.0.3` that `typescript: catalog:` resolved to. The version relationship with ts-morph's bundled compiler is unchanged, so consolidating onto one instance is a pre-existing cleanup rather than something this upgrade introduces.

---

.moon/tasks/tag-typecheck.yml
```diff
@@ -1,7 +1,8 @@
 $schema: https://moonrepo.dev/schemas/tasks.json
 tasks:
   build:
-    command: sh -c 'if [ "${DX_USE_TSC}" = "1" ]; then pnpm exec tsc --noEmit; else pnpm exec tsgo --noEmit; fi'
+    # `tsc` is the native TypeScript 7 compiler; `tsc6` is the classic JS one, kept as an escape hatch.
+    command: sh -c 'if [ "${DX_USE_TSC6}" = "1" ]; then pnpm exec tsc6 --noEmit; else pnpm exec tsc --noEmit; fi'
```
dmaretskyi: is the env var actually used anywhere in repo? if no then kill and use native tsc (go) as default

dmaretskyi:
Checked, and no — nothing in the repo set it. `git grep DX_USE_TSC` returned only its own definition and the two call sites; no CI job, script, or doc touched it. It was a leftover of the pre-GA switch that chose between the preview native compiler and the classic one, which stops making sense once `tsc` *is* the native compiler.

Removed in `90a02a50`. The typecheck task is now plain `pnpm exec tsc --noEmit` (the `sh -c` wrapper existed only for the conditional), and `dx-build` spawns `tsc` unconditionally.

`@typescript/typescript6` stays — it supplies the compiler API that `dx-build`, `@dxos/protobuf-compiler` and the script plugins call. Only its `tsc6` bin is now unused.

Verified `tasks:build` (a `typecheck`-tagged project) and `dx-build:build` pass on the new command.

---

dmaretskyi:
`storybook (24.11.1)` went red on `90a02a50`. Diagnosed — it is not this PR, and I am not changing the tests involved. Two failures, neither in a file this PR touches:

**1. `react-ui-feed` → `src/stories/assistant.stories.tsx > Interrupted`**

```
expected { stayed: false, resumed: true } to deeply equal { stayed: true, resumed: true }
```

**2. `plugin-terra` → `TerraArticle.stories.tsx > Objects`** — `Test timed out in 60000ms.`

Evidence:

- `git diff cc9b81fc..HEAD --name-only | grep -E 'react-ui-feed|plugin-terra'` is empty. This PR touches neither package.
- `assistant.stories.tsx` has one commit in its history — `cc9b81fc` (#12627), the commit immediately before this branch's tip. The suite is one commit old.
- Main's own Check run for `cc9b81fc` reports **0 failed jobs of 15**, so the story passed on main in the same environment.
- Reproduced locally on this branch: **it passes, but only on retry.**
  ```
  ✓ |storybook (chromium)| src/stories/assistant.stories.tsx (3 tests | 2 skipped) 8271ms
     ✓ Interrupted  8267ms (retry x1)
  ```
  `retry x1` = failed first attempt, passed second — on this branch, locally, with no CI load. The test is flaky on its own terms.

Why the mechanism rules this PR out: the failing assertion is `Math.abs(scroller.scrollTop - away) <= 2`, a 2-pixel scroll tolerance sampled across frame-timed `settle()` calls while a 60-chunk/40ms stream runs. Storybook tests execute story source through vite; this PR changes declaration emit and dependency resolution, not runtime scroll behaviour. `plugin-terra` hitting a hard 60s timeout in the same run is the load signature that also explains a frame-timing miss.

I am not touching `react-ui-feed`'s assertion — loosening a timing tolerance in a package landed an hour ago could mask a real bug in the follow logic it is there to test, and that call belongs to its author. The repo's sanctioned path is Trunk quarantine (this job already reports `Fail: 1  Quarantined: 0`), which I cannot drive: the Trunk MCP server needs interactive auth this session does not have.

Re-running the failed jobs, since the diagnosis is complete and the cause is a flaky test outside this diff. Auto-merge stays armed, so nothing merges while a required check is red.

---

PR #12653, opened 2026-08-18
composer-app: regenerate optimizeDeps for the TypeScript 6 compat package

dmaretskyi:
Follow-up to #12647, which left this generated file stale on purpose to keep that PR scoped to the toolchain.

`@dxos/plugin-code` and `@dxos/plugin-script` now import the compiler API from `@typescript/typescript6`, and the static `optimizeDeps.include` list did not carry it. That is exactly the cost the file exists to avoid — as its own header says, an unlisted dep is discovered mid-session, costing "one re-optimize and full page reload per plugin family".

Produced by `moon run composer-app:gen-optimize-deps`; not hand-edited.

## The diff

```diff
+  '@automerge/automerge-subduction',
-  '@dxos/client-services > @dxos/echo-host > @automerge/automerge-subduction',
+  '@typescript/typescript6',
```

Three things worth a reviewer's eye:

1. **`@typescript/typescript6` lands in bare form**, not nested `@dxos/plugin-code > …`, because it is also a root devDependency and so resolves from the app root.
2. **`@automerge/automerge-subduction` moves from nested to bare** — unrelated drift accumulated since the file was last regenerated, now that it resolves from the app root. Regenerating picks up reality; I did not filter it out to make the diff smaller.
3. **Bare `typescript` stays, and it is not stale.** `@valtown/codemirror-ts` imports it at *runtime* (`from "typescript"` and `require("typescript")` in its dist), not merely as a type — so the entry is real. Worth noting for whoever owns this file: that bare entry resolves from the app root, which is now TypeScript 7 (whose main entry is `lib/version.cjs`, no compiler API), while `@valtown/codemirror-ts`'s own import resolves under pnpm to its nested `typescript@6.0.3` that #12647's `.pnpmfile.cjs` rule gives it. If the bare entry pre-bundles the wrong copy, that is a property of the generator's bare-vs-nested decision rather than something this change introduces, and I did not hand-patch generated output to work around it.

## Verification

- `moon run composer-app:gen-optimize-deps` — exit 0, 113 tasks.
- `pnpm format` clean.

No changeset: `composer-app` deploys rather than publishes and is `ignore`d by Changesets.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KQo6mjku1D2WYjM3P4jnyP)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency optimization in Composer for more reliable application startup, faster build preparation, and smoother development workflows.
  * Updated optimization handling for illustration and library features to improve loading and build consistency.


dmaretskyi:
Merged main (18 commits) and **regenerated**, rather than just merging — this file is generated from the real dependency graph, so a list produced before those commits was stale by definition. That turned out to matter:

```diff
+  '@dxos/plugin-illustrator > @dagrejs/dagre',
+  '@dxos/plugin-illustrator > elkjs/lib/elk.bundled.js',
-  '@dxos/plugin-assistant > @dxos/react-ui-markdown > remark-gfm',
+  '@dxos/plugin-library > @dxos/react-ui-markdown > remark-gfm',
-  '@automerge/automerge-subduction',
```

The first three are explained by main: `plugin-illustrator` (#12662) pulled in `dagre`/`elkjs`, and `remark-gfm`'s owning package moved to `@dxos/plugin-library`.

**One entry needs an owner's eye — the `@automerge/automerge-subduction` drop.** Flagging rather than fixing, because the file is generated and I won't hand-patch it:

- It is not a truncated scan. The generator fails rather than emitting a partial list when a crawl cannot resolve an import, and it exited 0.
- It is not a race. Two consecutive runs at this commit produced **byte-identical** output.
- But the app does reach it: `@dxos/echo-host` declares it, `packages/core/protocols/src/automerge.ts` imports it, and `composer-app/vite.config.ts` names it in `SYNC_WASM_PACKAGES`.
- The asymmetry is what I can't account for: `@automerge/automerge` sits in that same `SYNC_WASM_PACKAGES` special case and **stays** in the list, while subduction drops. `SYNC_WASM_PACKAGES` predates this branch (it came in #12602), so main didn't introduce the difference.

My best guess is that the sync-WASM resolve hook rewrites the specifier to an absolute path before the scan records it as bare, and something about how each package is reached decides whether that happens — but that is a guess, and it is a property of the generator's bare-vs-nested decision rather than anything this PR changes. If the drop is wrong, the fix belongs in `scripts/gen-optimize-deps.mjs`, not in its output.

Verification: `gen-optimize-deps` exit 0 (113 tasks) ×2 with identical output, `pnpm format` clean.

---

PR #12669, opened 2026-08-19
onboarding: fix email login on composer dev and stop mislabelling failures as expired links

wittjosiah:
Email login on the composer `dev` channel (and every PR preview, which deploys to the same worker) fails immediately with "Login link expired", even when the link is used seconds after it arrives.

## Root cause

`DX_HUB_URL` and `DX_EDGE_BASE_URL` were paired across the environment split. The app never talks to the hub to redeem a token: `RedeemToken` → `Identity.recover` → `POST <DX_EDGE_BASE_URL>/identity/recover`, and db-service resolves the token through EDGE's own `HUB_SERVICE` **service binding**. EDGE `main` binds the `hub` (dev) worker, whose `SESSION_KV` is a different namespace from `hub-staging`'s — so a token minted by hub-staging was looked up in the dev hub's namespace and never found.

Confirmed in telemetry: `GET /account/activate/<token>` returns 302 on `hub-staging` (token valid), then `POST https://main.dxos.network/identity/recover` 500s with a single child span, `HubServiceEntrypoint.lookupRecoveryToken` in `deployment.environment: dev`, and no spans after it.

The coherent pairings are (`hub`, EDGE dev/main/labs), (`hub-staging`, EDGE staging) and (`hub-production`, EDGE production); `dev` was the only environment file crossing them. Pointing it at the dev hub keeps the channel on EDGE nightly as `ENV.md` intends, and costs nothing account-wise — both hubs share the `dev-hub` D1, so existing accounts and identity keys are unchanged.

Follow-up worth doing separately: the dev hub's vars are still localhost-shaped (`DX_WEBAUTHN_RP_ID=localhost`, `APP_URL=http://localhost:5173`), so passkey ceremonies from `dev.composer.space` fail closed. Magic-link login is unaffected — the client-supplied `redirectUrl` overrides `APP_URL`.

## Also fixed: the toast lied

`OnboardingManager` mapped *every* redemption failure to `login-link-expired-toast`, so a misconfiguration or a backend fault was indistinguishable from a spent link and the only honest signal was a `log.warn` in the console. Now:

- `@dxos/protocols` gains `InvalidRecoveryTokenError`. EDGE returns it for a token it cannot resolve through the failure envelope's serialized-error channel (`EdgeFailure.error` / `ErrorCodec`, which preserves error identity by name — the same way other typed failures cross the edge API boundary); it is registered so the name also survives the services RPC boundary.
- `EdgeIdentityRecoveryManager.recoverIdentityWithToken` matches the decoded cause with the class's `.is()` and rethrows the registered class.
- Onboarding shows the expired-link toast only for a refused token, and a new "Could not complete login" toast otherwise.

The EDGE side of the contract (404 + typed error, retiring the token on success rather than on lookup, and a 15-minute login TTL to bound the replay window that change opens) is dxos/edge#903.

## Testing

- `moon run plugin-onboarding:test` — 9 passed, including two new cases asserting which toast each failure shape produces.
- `moon run client-services:build protocols:build plugin-onboarding:build` and the matching `:lint` tasks — clean.
- `pnpm format`.

https://claude.ai/code/session_01PE9kHpy6JjrKrnsCDhrtTR

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovery-link redemption now distinguishes invalid, expired, or already-used tokens from other login failures.
  * Backend errors are no longer incorrectly displayed as expired or invalid recovery links.
  * Login failures now show clearer, context-specific notifications and continue guiding you back to the welcome screen.
  * Improved handling of wrapped recovery errors provides more accurate feedback when a login attempt cannot be completed.

### DIFF
--- a/.changeset/tidy-links-report.md
+++ b/.changeset/tidy-links-report.md
@@ -1,0 +1,6 @@
+---
+'@dxos/protocols': patch
+'@dxos/plugin-onboarding': patch
+---
+
+Distinguish a rejected email login token from a recovery that failed for another reason, so a backend failure is no longer reported as an expired link.
--- a/.github/workflows/env/dev
+++ b/.github/workflows/env/dev
@@ -5,7 +5,10 @@ COMPOSER_APP_POSTHOG_API_KEY=${COMPOSER_APP_TEST_POSTHOG_API_KEY}
 COMPOSER_APP_POSTHOG_FEEDBACK_SURVEY_ID=0199be95-be15-0000-56ca-1f0dfde8731f
 COMPOSER_APP_POSTHOG_PROJECT_ID=93152
 POSTHOG_CLI_API_KEY=${POSTHOG_CLI_API_KEY}
-DX_HUB_URL="https://hub-staging.dxos.workers.dev/"
+# Must be the hub that EDGE's `HUB_SERVICE` binding resolves to (`hub` for EDGE `main` below):
+# magic-link tokens live in that hub's SESSION_KV and db-service redeems them through the binding,
+# so a hub from another environment mints links no redemption can resolve.
+DX_HUB_URL="https://hub.dxos.workers.dev/"
 DX_IPDATA_API_KEY=${IPDATA_API_KEY}
 DX_OTEL_ENDPOINT=/api/otel
 DX_POSTHOG_API_HOST=https://o.composer.space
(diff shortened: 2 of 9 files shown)


wittjosiah:
@coderabbitai review

---

PR #12674, opened 2026-08-19
composer: align type hues, hide unrenderable types, gate the registry button

wittjosiah:
Follow-up to #12670 (the curated production plugin set).

## Icon hues

`Skill`, `Agent` and `Sequence` were `sky` while `Session` (`Chat`), `Project` and `Routine` were `amber`. All six are now amber, so the assistant/automation family reads as one group in the create dialog and the tree.

## Unfinished types in the curated set

`AssistantPluginOptions.experimentalTypes` (default `true`) gates the `Agent` and `Sequence` **create entries**; `plugin-defs.production.tsx` passes `false`. It withholds the entry, not the type — an `Agent` or `Sequence` created in a full-catalog build still resolves and renders in a curated one, it just cannot be created there.

Gating the schema registration as well was tried and deliberately reverted (`c864161e` / `c66658d3`): hiding it from the create dialog is sufficient, and keeping the types registered means existing objects still open normally.

## Objects nothing can render

The collection tree was showing rows for objects whose type is not available in the build — the dashed-circle "New item" entries. Both collection connectors now filter them out.

The predicate is the union of **registered schemas and space-stored types** (`TypeOptions.allTypesQuery`, i.e. `Scope.space()` + `Scope.registry()`), not `graph.registry` alone: persisted schemas never reach the shared graph registry, so a registry-only check would have hidden objects of user-defined types too.

A `TODO(wittjosiah)` records the intended end state — name the plugin that would render the object instead of hiding it.

## Registry button in a curated build

Regression from #12670: the create dialog offered "Open plugin registry" in a build that ships no registry. `PluginRegistryButton` now renders nothing when the registry plugin isn't enabled (fixing plugin-trip's banner too), gated via a new `usePluginRegistryAvailable` hook. The dialog gates its `Dialog.ActionBar` on the same hook, because `Dialog.Close asChild` needs an element child and would throw on a null one.

## Docs

`REPOSITORY_GUIDE.md` gains a **Plugin sets** section — the two sets and how `DX_PLUGIN_SET` selects them at build time, `serve-prod`, `check-plugin-set`, which environments ship which set, and the separate `DX_DEV` axis for enabled-by-default plugins. The deploy table now says production ships the curated set, which it previously implied was the same app as nightly.

## Test plan

Re-run after merging `main` (`b65d4fba`, mobile tweaks):

- `moon run composer-app:build plugin-space:build plugin-assistant:build plugin-preview:build app-toolkit:build` — clean (`tsc --noEmit`).
- `moon run plugin-preview:check-module-structure` — `No import paths from "./PreviewPlugin" to "{react,react-dom}"`.
- `moon run plugin-assistant:check-module-structure` — `No import paths from "./AssistantPlugin" to "@dxos/react-ui"`.
- Lint over every touched package, and `pnpm format` — clean.

Not verified in a running app: the tree filter and the hidden button were reasoned from the graph connectors and the plugin manager's `enabled` atom, not observed. The tree filter is visible on the preview URL (built for `dev`, so full catalog); the hidden button needs a local `serve-prod`.

https://claude.ai/code/session_019Lv5DBcj6euyi3x7KRrF57

## Summary by CodeRabbit

* **New Features**
  * Curated builds now hide unavailable object types from collections and creation menus.
  * Plugin Registry controls appear only when the registry is available.
  * Experimental Agent and Sequence creation is disabled in production curated builds.
  * Updated Agent, Skill, and Sequence icons with clearer color distinction.

* **Documentation**
  * Expanded guidance for curated plugin sets, production serving, build selection, and deployment environments.


wittjosiah:
@coderabbitai review


---

wittjosiah:
Both nitpicks addressed in `14424156` — they were right on the guideline.

- `AssistantOptions.ts` — the option's JSDoc is now one sentence carrying the default and the reason: *"Whether creation is offered for the unfinished types (`Agent`, `Sequence`) — default true, off in the curated set, which must not advertise a type it cannot yet do anything useful with."*
- `create-object.ts` — collapsed to *"Withholds the create entry, not the type, so an object made in a full-catalog build still opens."*
- `collections.ts` — `a schema an enabled plugin registered` stacked two relative clauses; rewritten as *"schemas registered by enabled plugins, plus those stored in the space"*, and the deployment-behaviour sentence dropped (it's in the PR description and `REPOSITORY_GUIDE.md`, which is the right home for it). The `TODO(wittjosiah)` stays — it records the intended end state, naming the plugin that would render the object instead of hiding it.

Verified after the edit: `plugin-assistant:build`, `plugin-space:build`, both lints, `pnpm format` — clean.


---

PR #12675, opened 2026-08-19
echo: normalize Chat ownership onto the parent edge (Filter.hasParent, remove CompanionTo)

richburdon:
## Summary

Normalizes Chat ownership onto the ECHO parent edge and removes the `Chat.CompanionTo` relation. A chat is now exactly one of: **standalone** (no parent, listed in the Chats section) or a **companion** (parented to its subject — a Markdown doc, a `Project`, an `Agent`, …), reachable through its subject's companion panel or, for projects, the navtree.

### `Filter.hasParent` (`@dxos/echo`)

New indexed predicate for parent presence, the blocking primitive for this migration:

- AST node `has-parent`, `Filter.hasParent(value = true)`, query-lite parity.
- Matchers in all three paths (live entity, automerge doc, index JSON); planner folds negation into the value and keeps the type-indexed select in `and` compositions; limits are never pushed past `child-of`/`has-parent` post-filters (`_filterContainsPostSelectPrune`).
- Reactive: `Obj.setParent` after the fact ejects/re-admits objects from subscribed `hasParent(false)` queries (e2e-tested).

### Chat parent-edge migration

- Writes: `Agent.makeInitialized` drops the relation (its `[Obj.Parent]: agent` edge is the linkage); `resetChatHistory` drops the relation dance — the new chat is primary by ULID creation order; `fork-chat` and `ChatCompanion` use `Obj.setParent(chat, companionTo)`.
- Reads: `Agent.loadChat` via `children()`; `Agent.loadForChat` is synchronous via `Obj.getParent` (no query); `ensure-companion-chat` / `useChatToolbarActions` via `children()`; `standaloneChatsQuery` collapses from two `Query.without` subtractions to `Filter.and(Filter.type(Chat), Filter.hasParent(false))`.
- `CompanionTo` deleted with every registration; no compatibility re-exports. Existing dev-space relations are dropped (pre-launch decision): their chats revert to standalone and re-parent on next companion use.
- Cascade delete now covers companion chats: deleting the subject deletes its chats and their feeds.

### Headless Project–Chat–assistant test

`assistant-toolkit/src/skills/project/conversation.test.ts` pins the full loop headlessly: Project ← parent — Chat, ProjectSkill + project bound on the feed, one prompt through `AgentService.getSession`, assertion on `project.artifacts`. Scripted (mock model) flavor runs in CI; the live memoized flavor is written and `.skip`-gated until its fixture is recorded (needs credentials; command in TASKS.md).

### Audits

- `packages/core/compute/compute/src/types/AUDIT.md` — Project type graph (tables + mermaid), edge inventory, migration record.
- `packages/core/compute/assistant/src/tool-runtime/AUDIT.md` — repo-wide tool-name audit (116 ops) + the approved key-derived tool-naming decision (implementation spun out as a separate task).

## Testing

- `echo` 574, `echo-host` 285, `echo-query` 13, `echo-client-e2e` query 167 (incl. 4 new `hasParent` e2e), `assistant-toolkit` 72, `plugin-assistant` 163, `plugin-projects` 28, `agent-runtime` 30 — all green.
- Full `moon run :test` sweep + `moon run :lint` green before push; full `composer-app:build` graph compiles.

## Summary by CodeRabbit

* **New Features**
  * Added parent-aware filtering to find objects with or without a parent.
  * Project, agent, and companion chats now use parent-child relationships for ownership and discovery.
  * Standalone chat views now show only chats without a parent.
* **Bug Fixes**
  * Improved selection of the latest companion chat and chat history.
* **Tests**
  * Added coverage for parent filtering, reactive updates, and project conversations.
* **Documentation**
  * Added project design, planning, milestone, and use-case documentation.


packages/core/compute/assistant-toolkit/src/types/Agent.ts
```diff
@@ -83,24 +83,28 @@ export const loadInstructions = (
...
  */
 export const loadChat = (agent: Agent): Effect.Effect<Chat.Chat | undefined, never, Database.Service> =>
   Effect.gen(function* () {
-    const chats = yield* Database.query(Query.select(Filter.id(agent.id)).targetOf(Chat.CompanionTo).source()).run;
-    const [chat] = chats.filter(Obj.instanceOf(Chat.Chat));
-    return chat;
+    const children = yield* Database.query(Query.select(Filter.id(agent.id)).children()).run;
+    return children
+      .filter(Obj.instanceOf(Chat.Chat))
+      .sort((a, b) => a.id.localeCompare(b.id))
```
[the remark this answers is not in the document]
richburdon: Fixed in 57d3a3735e — renamed to `left`/`right` here and in the same pattern in `ensure-companion-chat.ts`.

packages/plugins/plugin-projects/docs/TASKS.md
```diff
@@ -650,6 +650,13 @@ function` when a tool-call `doc` ref decodes without a resolver; the five doc-re
...
+- [ ] **Derive tool names from DXN keys** (approved 2026-08-19; chip spawned) — replace `makeToolName(meta.name)` with key-derived names (`markdown-create`), unify with `Skill.toolDefinitions`, add `Skill.toolName()` helper + bind-time uniqueness invariant, sweep hardcoded names, regenerate fixtures + the tool AUDIT. Decision record: `packages/core/compute/assistant/src/tool-runtime/AUDIT.md`.
+- [ ] **Record the live-model fixture for the Project conversation test** — `packages/core/compute/assistant-toolkit/src/skills/project/conversation.test.ts` gates its live flavor with `.skip`; run `DX_UPDATE_MODEL_FIXTURES=1 moon run assistant-toolkit:test -- src/skills/project/conversation.test.ts` with 1p credentials, commit `.store/conversations/**`, drop the `.skip`.
+- [x] **Normalize Chat ownership onto the ECHO parent edge** — DONE on this branch: `Filter.hasParent`
+      landed in `@dxos/echo`; all `CompanionTo` write/read sites migrated to `Obj.setParent` /
+      `children()` / `Obj.getParent`; `standaloneChatsQuery` is `hasParent(false)`; `CompanionTo`
+      deleted (pure drop, no migration — pre-launch). Audit:
+      `packages/core/compute/compute/src/types/AUDIT.md`.
```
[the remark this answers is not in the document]
richburdon: Fixed in 57d3a3735e — updated the milestone-3 note in TASKS.md (CompanionTo deleted, all chats on the parent edge) and the `resetChatHistory` description in PLAN.md.

packages/plugins/plugin-assistant/src/containers/ChatCompanion/ChatCompanion.tsx
```diff
@@ -37,16 +37,13 @@ export const ChatCompanion = forwardRef<HTMLDivElement, ChatCompanionProps>(
         return;
       }
 
+      // The parent edge is the ownership statement: the chat belongs to its subject (cascades on
+      // delete, keeps it out of the standalone Chats section).
+      Obj.setParent(chat, companionTo);
```
wittjosiah: Using `Obj.setParent` without a ref in the parent to the child is an anti-pattern. The proper way to generalize this without needing to add a "chats" field to every schema is to use an echo Annotation on the object. We should probably introduce an invariant in setParent that enforces there's a ref to the child being set either in the objects data model or an annotation.

wittjosiah (dismissed):

packages/plugins/plugin-assistant/src/containers/ChatCompanion/ChatCompanion.tsx
richburdon: Agreed there is a real gap: with no inbound ref the chat is only discoverable by index query (`children()` / `hasParent`), never by graph traversal, and nothing at the schema level declares that chats may accumulate under a given type.

Two notes on scope:

1. This PR generalizes the pattern `plugin-projects/create-chat.ts` already established for project chats (ownership = parent edge, enumeration = `children()` — deliberately no `chats` field, per the `Project` doc comment: refs are for what the project owns and orders, queries for what accumulates). Companion chats here do keep an inbound ref — `ChatCompanion` still runs `SpaceOperation.AddObject`, so the space root references them; the fork path parents an already-persisted chat.
2. The fix you describe — a type-level annotation (e.g. `ChildrenAnnotation` naming accepted child types) plus a `setParent` invariant requiring either a data-model ref or that annotation — is an `@dxos/echo` API design that would also have to cover today's ref-less parents (project chats, `Agent.makeInitialized`'s chat). Proposing to do it as an immediate follow-up rather than inside this diff; tracked in `plugin-projects/docs/TASKS.md` with this thread linked.

OK to land with the follow-up, or do you want the annotation + invariant in this PR?

wittjosiah: The invariant is a safeguard which is reasonably separable, though this PR is also already touching echo.<br><br>The annotation however is inseparable from removing the relation. We must not add unreferenced children to objects, so if the removal of the companion relation is included the annotation should replace it.

richburdon: Implemented in 2748c142bd (this PR), warn-first:

- `Annotation.ChildrenAnnotation` (parent-side: child typenames this type may parent ref-less, '*' wildcard) and `Annotation.ParentAnnotation` (child-side: parents this type vouches for — the companion-chat shape, `Chat` sets `['*']`).
- `Obj.setParent` now checks `Obj.isDeclaredParentEdge` — parent-held ref, parent-side annotation, or child-side annotation — and logs a warning on undeclared edges. Promoting to a hard `invariant` is deferred until the repo-wide sweep (~60 call sites) is done; the warning already surfaced and fixed real offenders (TaskSet ledger edges → annotated; routine/trigger + scaffold orderings → ref now set before the edge).
- Unit coverage in `Obj.test.ts` for all three declaration paths.

Follow-up (tracked in plugin-projects TASKS.md): sweep remaining warn sites, then promote to invariant.

richburdon: Agreed on both points, and that is how it now stands in this PR: the annotation landed together with the relation removal (2748c142bd — `Chat` carries `ParentAnnotation(['*'])`, so no unreferenced-and-undeclared children are introduced by this change), and the `setParent` enforcement is staged separately: the check (`Obj.isDeclaredParentEdge`) is in and warns; promotion to a hard invariant follows the repo-wide call-site sweep, tracked in plugin-projects TASKS.md.

wittjosiah: No, this is incorrect. You've created schema annotations not echo annotations, the parent one doesn't make sense and the children one isn't enforceable currently. To be clear about how I was saying this should be done:
- it should use `Annotation.make` from echo, e.g. `DefaultSpaceAnnotation`
- we should have a `CompanionChatAnnotation` to replace the `CompanionChat` relation
- this data is stored on the object rather than in another entity and so it deleted with the object
- this data being on the object gives the object refs to the chats allowing them to be children, thus the chats are also deleted with the object

Additionally, `isDeclaredParentEdge` is very much not something which should be declared on the public object api. The invariant should be an internal implementation detail of `Obj.setParent`.

richburdon: Reworked in d7e0c5ede4 exactly along these lines: the schema-level Children/Parent annotations are removed; `Chat.CompanionChatAnnotation` (`Annotation.make`, `Schema.Array(Ref.Ref(Chat))`) replaces the relation, stored on the subject object so it dies with it and its refs make the parented chats referenced children; `Chat.linkCompanion` wires annotation ref + parent edge at every link site (companion, fork, project create-chat, agent init/reset). `isDeclaredParentEdge` is gone from the public API — the check is now internal to `Obj.setParent` (ref-in-parent, data or object annotations) and warns pending the call-site sweep before promotion to an invariant.

PR #12690, opened 2026-08-20
onboarding, connector: run desktop OAuth in the system browser

wittjosiah:
## Problem

Every OAuth flow in the native desktop app fails before the provider's page appears — account login/signup:

```
operation invocation failed
  cause: "Error: Unable to open OAuth recovery window (popup blocked?)."
  opKey: "dxn:org.dxos.plugin.onboarding.operation.redeemOAuthRecovery"
```

and, silently, connecting or reauthenticating an integration. All four call sites hand the authorization URL to `window.open`; WKWebView creates no window and returns `null`.

### How it regressed

This worked once. #10396 ("feat: oauth in tauri app") added the Rust loopback server *and* its client, `plugin-integration/src/oauth/tauri-server.ts`. #10754 stripped the `if (isTauri())` branch out of the call site; #11915 deleted the orphaned file. The Rust commands have sat in `lib.rs`'s `invoke_handler` with no callers ever since. Onboarding's OAuth recovery came later and was written against the same web-only assumption.

## Approach

Restores that flow and extends it to the recovery/register purposes it predates.

- **The auth page opens in the user's real browser.** An embedded webview is not an option: Google's `disallowed_useragent` policy names `WKWebView` on macOS explicitly, and an in-app window starts from an empty cookie jar, so the user re-enters credentials their browser already holds.
- **`/oauth/initiate` is issued from Rust**, so `Origin` names the loopback server. `api.ts` reads `c.req.header('origin')` and echoes it verbatim as the callback target, with no allowlist — and page script cannot set that header, which is why the call cannot stay in the webview.
- **The loopback server relays the callback into the app**, which finalizes against the client already running there.

Extended beyond the restore, since account recovery postdates it: the server also answers `/redirect/oauth-recovery` (`recoveryProof` / `registrationToken` / `error`), and initiate carries `purpose`, `registerRecovery` and `loginHint`.

Both finalizers — `plugin-onboarding`'s `oauth-recovery-redirect` and `plugin-connector`'s `oauth-redirect` — now take params from either the page they land on (browser) or the relayed callback (desktop). Everything downstream is unchanged, including the `localStorage` snapshots both flows stash.

`EdgeHttpClient.getAuthHeader()` is the supported way to hand the presentation to Rust; it also removes the `_authHeader` cast in `commands/connector/oauth-flow.ts`, whose TODO asked for exactly this API.

### The Tauri ACL

The first packaged build still failed, with `Command start_oauth_server not allowed by ACL` — and that turned out to be broader than OAuth. Tauri ACL-checks *non-plugin* commands whenever the webview origin is not local:

```rust
if (plugin_command.is_some() || has_app_acl_manifest || !is_local) && invoke.acl.is_none()
```

`is_local_url` tests the URL against `devUrl` / `frontendDist`. The bundled app is served from `http://localhost:<port>` by the localhost plugin (SharedWorker needs an HTTP origin), matching neither — so **every** app command was refused, including the asset cache, xattr and spotlight commands. Dev is served from `devUrl` itself, which is why this only appears in a packaged build.

`build.rs` now declares the command list so tauri-build autogenerates each `allow-*` permission, granted in the desktop capability. Declaring the manifest also enables ACL checking for local origins, so iOS grants its own commands rather than silently losing them.

### Verified against kms-service

Reading `edge`'s `kms-service/src/oauth/`, rather than assuming:

- `redirect-page.ts` sends recovery/register flows to `${origin}/redirect/oauth-recovery?…` unconditionally, and integration flows to `${requestOrigin}/redirect/oauth?…` whenever there is no opener — which is every browser-hosted flow here.
- `requestOrigin` is the raw `Origin` header, used verbatim, no allowlist. The allowlist in `provider.ts` governs the *provider→Edge* callback, so a loopback client origin does not affect atproto's registered redirect URI or PAR.

## Pending state

The Continue button was disabled only while the initiating call ran — and that call resolves the moment the provider's page opens, so the form re-enabled while the user was still authenticating, and a second submission could start a duplicate flow. `oauthPending` is now held past the call's resolution: the handle field locks so its value still describes what is being waited on, and the button reads "Authorizing…" alongside the existing "Waiting for passkey…". A failure clears it via the `error` the screen already reports; a 5-minute ceiling releases the form for the user who closes the auth page instead of completing it.

| Idle | Authorizing |
| --- | --- |
|  |  |

Both captured from one build, from `Welcome/EmailPrimary` in storybook, `deviceScaleFactor: 2`:

```
idle     inputDisabled=false  buttonDisabled=false  label="Continue"      input w=238
pending  inputDisabled=true   buttonDisabled=true   label="Authorizing…"  input w=218
```

The joined row gives the button its label's width and the field the rest, so the field narrows by 20px while pending — which is why the label is "Authorizing…" rather than something longer.

## Notes

- An earlier revision of this branch hosted the auth page in an in-app Tauri window. That is replaced here — see the Google policy above.
- `nativeAppRedirect` (302 to `composer://oauth/callback`) is a third route the protocol scaffolds but nothing completes; it carries `recoveryProof` but **not** `registrationToken`, so signup would need an Edge change before it could serve every flow.

## Testing

- `moon run …:build / :lint / :test` across `edge-client`, `app-toolkit`, `plugin-connector`, `plugin-onboarding`, plus `app-toolkit:check-module-structure` — 202 tasks, all pass
- `cargo check` — clean apart from the pre-existing macOS-only `hidden_title` call, which cannot compile on the Linux sandbox
- Desktop login verified end-to-end against a live atproto account on the `0.11.1-dev.6` dev build
- Connector flows not yet exercised against a live provider

https://claude.ai/code/session_01FCSHL56yhWoMWquZ3Aod5e

## Summary by CodeRabbit

* **New Features**
  * Added native OAuth sign-in for supported desktop apps through the system browser.
  * Added browser-based fallback authentication for unsupported platforms.
  * Added OAuth recovery and login-hint support.
  * Added callback handling without requiring an additional app reload.
* **Bug Fixes**
  * Improved handling of invalid callbacks, popup blocking, and startup failures.
  * Prevented competing authentication attempts during sign-in.
* **User Experience**
  * Added an “Authorizing…” status while OAuth sign-in is underway.


wittjosiah:
@coderabbitai review

The previous review was rate-limited before it could see `712ddd7`, so the Merge Risk summary is stale — both residual items it lists are now addressed:

- **Identity race on `getAuthHeader`** — fixed in `712ddd7`. The requesting identity is captured before the prefetch is awaited and the header withheld if it changed, with a regression test that fails without the fix (it returns a presentation minted for `did:halo:other`).
- **Competing authentication actions during OAuth** — fixed earlier in `624a432`. `LoginTab` receives `pending={formPending}` (`pending || oauthPending`), so the passkey button and every `submitDisabled` gate are blocked while a flow is out.

---

PR #12699, opened 2026-08-21
echo,assistant: rebuild recursive JSON schemas instead of overflowing the stack

wittjosiah:
Fixes the composer e2e failure in [run 32446159816](https://github.com/dxos/dxos/actions/runs/32446159816): `chat.spec.ts › sends a message and receives a response` failed on chromium, firefox **and** webkit with `request failed: AI service error / An unexpected error occurred.`

## Diagnosis

The toast text is generic, so the cause came from the Playwright trace attached to the chromium job:

```
request failed RangeError: Maximum call stack size exceeded
    at ProxyHandlerSlot.get
    at Proxy.map
    at toEffectSchema$1
    at objectToEffectSchema
    at toEffectSchema$1
    …
    at AiSession.createRequest
```

Not an outage: the chat's own name-generation round trip succeeded in the same session (`chat name updated { newName: 'Clear Sky Blue' }`), so the AI service was reachable. The request died while building the toolkit.

`makeToolResolverFromOperations` falls back to the operation registry for any tool id the in-process toolkit does not carry, and `plugin-routine`'s registry sync persists **every** plugin operation handler there via `Operation.serialize`. A cyclic schema serializes as `$defs` plus a `$ref` into them — which is exactly what `JsonSchemaType` itself emits (its `$defs` include `jsonSchema` and `dependency`, the two names the `strict` path in `toJsonSchema` already strips as a workaround). Three stages of the read-back path could not represent that shape:

1. **`JsonSchema.toEffectSchema`** resolved a `$ref` by recursing into its definition with no cycle guard, so a cyclic definition recursed until the stack blew. This is the `TODO(dmaretskyi): Currently unable to deserialize` on the existing `serialize circular schema (TypeSchema)` test.
2. **`mapSchemaTypeForLLM`** (the LLM-facing tool projection) delegated `Suspend` to `SchemaEx.mapAst`, which forces the thunk — so once (1) was fixed, the same walk recursed forever one layer up.
3. **`toModelJsonSchema`** kept only `Schema.toJsonSchemaDocument(...).schema` and dropped `definitions`, so the model would have been shown a dangling `$ref`.

## Changes

- `toEffectSchema` publishes each `$defs` entry to a per-conversion cache *before* building its body, and closes a re-entrant `$ref` with `Schema.suspend`. The cache also shares one node between every reference to a definition.
- `mapSchemaTypeForLLM` handles `Suspend` itself: it rebuilds the body lazily and memoizes on node identity, the same technique `withEchoRefinements` already uses on the serialize side.
- `toModelJsonSchema` carries the document's definitions over as `$defs`.
- `makeToolResolverFromOperations` reports a record it cannot rebuild as `AiToolNotFoundError` (with a warning) instead of throwing, so one bad operation drops one tool rather than failing the whole request — matching how an unresolvable tool id is already handled.

## Tests

New, and each one fails on `main`:

- `echo`: `serialize circular schema (TypeSchema)` now deserializes (TODO removed); `deserialize a self-referential schema` and `deserialize a pair of mutually-recursive schemas` decode a nested value, which proves the cycle was rebuilt rather than truncated to `Unknown`.
- `assistant`: `an operation with a recursive input schema survives the registry round trip` drives `serialize → deserialize → projectFunctionToTool` and asserts every `$ref` the model is shown names a present definition.

```
echo:test        Tests  575 passed | 9 skipped (584)
assistant:test   Tests   54 passed | 5 skipped (59)
compute:test      56 passed · schema:test 51 passed · mcp-server:test 52 passed
conductor:test    31 passed · react-ui-table:test 21 passed
echo:lint / assistant:lint clean · pnpm format applied
```

## Not addressed

The same run also failed `collaboration.spec.ts › guest joins host's space` on **webkit only**, through all three retries (`space invitation never reached the auth-code step`). That is the tracked production-edge two-peer stall (DX-1152) the suite already documents, retries for, and deliberately keeps enabled as a sensor; the sibling test that performs the same invitation passed on webkit in the same run. No change made there.

I also did not identify the specific registry operation whose schema is cyclic — the trace does not name it, and enumerating the app's operation set needs the whole plugin graph activated. The fix is at the boundary and covers the class, and the last bullet above bounds the blast radius if another unrepresentable record shows up.

https://claude.ai/code/session_01EvDB8qpHsRe76Xgh7YF4QD

---
_Generated by [Claude Code](https://claude.ai/code/session_01EvDB8qpHsRe76Xgh7YF4QD)_

### DIFF
--- a/.changeset/recursive-schema-tool-projection.md
+++ b/.changeset/recursive-schema-tool-projection.md
@@ -1,0 +1,5 @@
+---
+'@dxos/echo': patch
+---
+
+Fix `RangeError: Maximum call stack size exceeded` when a recursive JSON Schema is rebuilt or projected onto an AI tool. `JsonSchema.toEffectSchema` now resolves a cyclic `$ref` through `Schema.suspend` instead of recursing into it, the LLM-facing tool projection rewrites a suspended node lazily, and a projected tool carries the `$defs` its `$ref`s point at. A stored operation whose schema still cannot be rebuilt is now dropped from the toolkit rather than failing the whole request.
--- a/packages/core/compute/assistant/src/tool-runtime/services.test.ts
+++ b/packages/core/compute/assistant/src/tool-runtime/services.test.ts
@@ -156,4 +156,46 @@ describe('projectFunctionToTool', () => {
     });
     expect(decoded.properties).toEqual({ any: 1 });
   });
+
+  // A cyclic input schema serializes as `$defs` plus a `$ref` into them. Rebuilding it used to
+  // recurse until the stack blew, which reached the user as "Maximum call stack size exceeded" from
+  // the toolkit build and failed the whole chat request; and the advertised schema dropped the
+  // definitions, leaving the model a dangling `$ref`.
+  test('an operation with a recursive input schema survives the registry round trip', ({ expect }) => {
+    interface Node {
+      readonly name: string;
+      readonly children?: readonly Node[];
+    }
+
+    const Node: Schema.Codec<Node> = Schema.Struct({
+      name: Schema.String,
+      children: Schema.optional(Schema.Array(Schema.suspend((): Schema.Codec<Node> => Node))),
+    });
+
+    const Recursive = Operation.make({
+      meta: { key: DXN.make('org.dxos.test.function.recursive'), name: 'Recursive' },
+      input: Schema.Struct({ root: Node }),
+      output: Schema.Void,
+    });
+
+    const advertised = Tool.getJsonSchema(projectFunctionToTool(Operation.deserialize(Operation.serialize(Recursive))));
+    const refs = collectRefs(advertised);
+    expect(refs.length).toBeGreaterThan(0);
+    const defs = (advertised.$defs ?? {}) as Record<string, unknown>;
+    // Every `$ref` the model is shown must name a definition that is actually present.
+    expect(refs.filter((ref) => !(ref.replace('#/$defs/', '') in defs))).toEqual([]);
+  });
 });
+
+/** Every `$ref` value anywhere in a JSON Schema node. */
+const collectRefs = (node: unknown): string[] => {
+  if (Array.isArray(node)) {
+    return node.flatMap(collectRefs);
+  }
+  if (typeof node !== 'object' || node === null) {
+    return [];
+  }
+  return Object.entries(node as Record<string, unknown>).flatMap(([key, value]) =>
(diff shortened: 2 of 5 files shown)


wittjosiah:
Closing in favour of #12698, which reached the same diagnosis ~10 minutes earlier and lands the same three fixes (`toEffectSchema` `$ref` → memoized suspend, `mapSchemaTypeForLLM` suspend memoization, `toModelJsonSchema` carrying `$defs`). It is stronger on verification — a model-fixture test that drives the real path (`ToolExecutionServices` → registry → `Operation.deserialize`) rather than only a unit round trip — and carries a CI-correctness fix this PR doesn't have (tag-gated suites keyed on `DX_RUN_*` in the browser/workerd/storybook moon tasks). Its `$ref` handling is also better: it swaps the placeholder for the expanded node after descending, so a later non-cyclic `$ref` inlines directly and the outermost node stays a plain struct.

One change here has no counterpart in #12698, and is independent of the schema fix — `makeToolResolverFromOperations` wrapping the projection so a record it cannot rebuild is reported as `AiToolNotFoundError` instead of throwing:

```ts
return yield* Effect.try({
  try: () => projectFunctionToTool(Operation.deserialize(results[0])),
  catch: (error) => {
    log.warn('Failed to project a stored operation to a tool', { id, error });
    return new AiToolNotFoundError(id);
  },
});
```

A throw there escapes as a defect and fails the entire turn, which is why one unrepresentable operation record took down every chat request rather than dropping one tool. `resolveToolkit` already tolerates `AiToolNotFoundError` per tool, and `Operation.serializable` already drops an operation it cannot serialize, so this is the symmetric case on the read side. Worth porting separately if wanted — the branch `claude/composer-e2e-tests-k56io1` keeps it.

Two notes for whoever lands #12698:

- Neither PR's own Check runs the composer e2e (the `e2e` job is gated on `main` / `changeset-release/*` / an explicit dispatch), so a green PR Check does not exercise `chat.spec.ts`. I dispatched Check with `e2e: true` on this branch ([run 32477181883](https://github.com/dxos/dxos/actions/runs/32477181883)) to get that signal against the fix; I'll post the outcome here.
- The same nightly ([run 32446159816](https://github.com/dxos/dxos/actions/runs/32446159816)) also failed `collaboration.spec.ts › guest joins host's space` on webkit through all three retries. Neither PR touches that; it is the tracked DX-1152 edge stall the suite already documents and retries for, and the sibling test performing the same invitation passed on webkit in the same run.

---

PR #12725, opened 2026-08-24
echo: add Migration.defineRename for renamed named entities

dmaretskyi:
## Summary

Adds `Migration.defineRename({ from: 'org.example.operation.foo', to: 'org.example.operation.bar' })` — a migration for when a named entity is renamed, repointing existing `dxn:` references at the new name.

Getting there needed the reverse-reference index to cover named entities, and the query language to be able to ask about one. Both are useful on their own: `Query.select(Filter.key(dxn)).referencedBy()` now answers "what references this name?" for any caller.

### Type surface

`Migration.TypeId` and a `Migration.Migration` base carrying a `kind` discriminant. Both concrete variants extend it:

- `Migration.ObjectMigration` (`kind: 'object'`) — produced by `Migration.define`.
- `Migration.RenameMigration` (`kind: 'rename'`) — produced by `Migration.defineRename`.

`Migration.isMigration` guards an unknown value: it requires the brand *and* a recognized `kind`, so a branded value carrying an unknown kind is rejected rather than silently skipped. `Migration.isObjectMigration` / `Migration.isRenameMigration` narrow a known migration. `defineRename` validates its NSIDs at compile time the same way `DXN.make` does.

### Reverse-reference index covers named entities

`extractReferences` previously dropped every non-`echo:` reference, so `dxn:` targets were absent from the index entirely. A `dxn:` reference is now keyed by its **unversioned** NSID, so one lookup finds every version of a name; the referencing object still holds the versioned URI, which is what a rewrite reads. Echo references keep their existing local-form normalization.

The SQL column was already `TEXT`, so there is no table change — but databases written by earlier releases hold no `dxn:` rows. A cursor-clearing migration (`indexCursor`, `indexName = 'reverseRef'`) makes the next indexing pass re-scan every document, following the precedent in `0001_init.sql`, which deletes cursors for retired index names.

### `Query.select(Filter.key(dxn)).referencedBy()`

The construct was already expressible in the AST, but it could never resolve: an `incoming-references` clause anchors on a working set, and `QueryItem.objectId` is a required `EntityId`, so `_anchorTargetDxn` can only ever produce an `echo:` EID. A named entity has no entity id — and in a rename it is not in the graph at all.

The planner now collapses that shape to a single `SelectStep` carrying a new `IncomingReferenceSelector`, resolved straight off the reverse-reference index, with no anchor evaluation. Evaluating the anchor would be both a scan (a meta-key filter is matched in memory, not indexed) and empty for the case this serves. The traversal's type and property narrowing still apply, so `.referencedBy(Type, 'prop')` composes as before.

Detection is deliberately narrow — a bare `Filter.key`, with no typename, id, props or foreign keys:

- `Filter.and(Filter.type(X), Filter.key(k))`, the shape every existing caller uses, is an `and` node and falls through untouched.
- A version-constrained `Filter.key(k, { version })` also falls through and keeps its composed meaning, since the index keys a named entity without its version and so cannot honour a semver range.

The client-side working-set executor implements the same selector, so objects added in the current session — which the index has not seen yet — are matched too.

### Application

`EchoDatabase.runMigrations` validates every `kind` up front — a batch containing an unrecognized migration aborts before the first handler runs, so it cannot leave earlier migrations half-applied — then dispatches on `kind`.

A rename asks the index for the objects referencing `from`, loads only those, and rewrites the matching references in place, preserving any `:<version>` suffix. A reference that already reads correctly is not written, which makes the migration idempotent: a re-run, a peer that already replicated the result, and a rename whose target equals its source are all no-ops at the document level.

Reference holders outside the space's object graph — queues and feeds — are not indexed for reverse lookup and are not migrated.

## Test plan

`echo-client-e2e/src/query.test.ts` — new `describe('referencedBy a named entity')`, 6 tests: finds referrers of a DXN absent from the graph; accepts a full `dxn:` as the key; an unversioned name matches a versioned reference; narrows by referencing type and by property; empty when nothing references the name; a version-constrained key keeps the composed meaning.

`echo-client/src/proxy-db/rename-migration.test.ts` — 9 tests: `defineRename` shape and `isMigration` / `kind` for both variants; `isMigration` rejects unbranded and unknown-`kind` values; references repointed across a scalar field, an array element and a nested struct with the version suffix preserved; idempotency across two runs; a same-name rename leaving the document untouched; an unrecognized migration rejecting without applying the valid one beside it; entity and non-matching named references left alone; and one that reopens the space on a **fresh client with an empty working set**, so the index is the only thing that can resolve the query.

`index-core/src/indexes/reverse-ref-index.test.ts` — a named-entity reference and its versioned spelling collapse to one key, and a versioned lookup finds both.

```
✓ echo-client-e2e  311 passed | 1 expected fail | 7 skipped (319)
✓ echo-client      534 passed | 7 expected fail | 12 skipped | 2 todo (555)
✓ echo-host        290 passed | 2 skipped | 1 todo (293)
✓ index-core        49 passed (49)
```

Full-repo `moon exec :build` clean; lint clean on `index-core`, `echo-host`, `echo-client`, `echo-client-e2e`, `protocols`; `pnpm format` run.

Two assertions were checked for vacuity rather than assumed: stubbing the planner's `namedEntityAnchor` to always decline fails exactly the four rename tests that depend on the collapse (confirming the composed anchor really is empty), and removing the `rewritten === uri` guard fails the same-name no-op test. An earlier document-heads assertion on the *idempotency* test was dropped as flaky — 2 failures in 8 local runs, because the preceding rename's writes are still settling when the snapshot is taken; the same-name test covers that guarantee with nothing preceding it.

Not addressed: a DXN-anchored query has no anchor object for targeted invalidation, so it falls back to catch-all. Fine for a migration, worth knowing if the construct is used in a live reactive query.


dmaretskyi:
@coderabbitai review

The previous review aborted because the head moved to `109b87f0` mid-run. That commit addresses the partial-batch finding — `runMigrations` now validates every `kind` before invoking the first handler — so the standing 🟡 Moderate merge-risk assessment (capped at `6a532`) is stale.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
```diff
@@ -89,7 +89,7 @@ export interface EchoDatabase extends Database.Database {
   /**
    * Run migrations.
    */
-  runMigrations(migrations: ObjectMigration[]): Promise<void>;
+  runMigrations(migrations: Migration.Any[]): Promise<void>;
```
dmaretskyi:
```suggestion
  runMigrations(migrations: Migration.Migration[]): Promise<void>;
```

dmaretskyi: nvm

packages/core/echo/echo-client/src/proxy-db/database.ts
```diff
@@ -727,39 +727,108 @@ export class DatabaseImpl extends Resource implements EchoDatabase {
...
+      if (!DXN.isDXN(uri) || DXN.getName(uri) !== fromName) {
+        return undefined;
+      }
+      const rewritten = DXN.make<string>(toName, DXN.getVersion(uri));
+      return rewritten === uri ? undefined : rewritten;
+    };
 
-        if (migration.onMigration) {
-          await migration.onMigration({ before, object, db: this });
+    const objects = await this._hypergraph.query(Query.select(Filter.everything()).from(this)).run();
```
dmaretskyi: use efficient reverse-reference query

dmaretskyi: Done in `d1986116` — the scan is gone. Two things blocked the index path, both fixed here:

1. **The index didn't cover named entities.** `extractReferences` dropped every non-`echo:` reference, so `dxn:` targets were absent entirely. A `dxn:` reference is now keyed by its **unversioned** NSID, so one lookup finds every version of a name; the referencing object still holds the versioned URI, which is what the rewrite reads. Echo refs keep their local-form normalization. No table change (the column was already `TEXT`), but existing databases hold no `dxn:` rows — so a cursor-clearing migration (`indexCursor`, `indexName = 'reverseRef'`) makes the next pass re-scan, following the precedent in `0001_init.sql`.

2. **No client-reachable lookup by DXN.** An `incoming-references` clause can only anchor on an entity in the working set (`_anchorTargetDxn` derives an `EID` from a `QueryItem`), and a named entity never is one. Rather than extend the AST, planner, plan steps and both executors, this adds `QueryService.queryReverseRef`. Its handler awaits `updateIndexes()` first, so a caller that just wrote an object doesn't read a stale set.

`#runRenameMigration` now asks the index for the referencing objects and loads only those.

The 8 rename tests run client → RPC → index, so they double as proof that `dxn:` targets are indexed and that the unversioned key resolves the versioned reference in the fixture; `reverse-ref-index.test.ts` gained a unit test for the key normalization. `index-core` 48/48, `echo-host` 293, `echo-client` 554 all pass.

---

packages/core/echo/echo-client/src/proxy-db/database.ts
dmaretskyi: I'd already applied this before the "nvm" landed, and left it in — `Migration.Migration[]` reads better as the public signature than a `Migration.Any` union alias. Dispatch narrows through new `Migration.isObjectMigration` / `isRenameMigration` predicates instead of a `kind` switch, so it stays cast-free; `Migration.Any` is gone. Happy to revert to the union if you'd rather.

---

packages/core/echo/echo-host/src/db-host/query-service.ts
```diff
@@ -143,6 +144,31 @@ export class QueryServiceImpl extends Resource implements QueryService.Handlers
...
+        .pipe(RuntimeProvider.provide(this._params.runtime));
+      const metas = yield* this._params.indexEngine
+        .lookupByRecordIds(Array.from(new Set(rows.map((row) => row.recordId))))
+        .pipe(RuntimeProvider.provide(this._params.runtime));
+      const objectIds = Array.from(
+        new Set(metas.filter((meta) => meta.spaceId === request.spaceId).map((meta) => meta.objectId)),
+      );
+      return { objectIds };
+    }).pipe(Effect.orDie);
+  }
```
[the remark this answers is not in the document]
dmaretskyi: Done in `7e0690e2`, which landed before this re-raise was posted (the review ran against `3eb9f765`).

Both failure paths now map to `QueryError` — the repo's existing `BaseError.extend('QueryError')` in `echo-host/src/query/errors.ts` — via `QueryError.wrap(...)`:

```ts
yield* Effect.tryPromise({
  try: () => this._params.updateIndexes(),
  catch: QueryError.wrap({ message: 'Failed to bring the index up to date.' }),
});
...
}).pipe(Effect.mapError(QueryError.wrap({ message: 'Reverse-reference lookup failed.' })));
```

`Effect.orDie` is gone, so index failures no longer become defects.

One thing the suggestion can't have: the handler's declared channel stays `Error`. That type comes from `QueryService.Handlers`, derived from `Rpc.make(..., { error: serviceError })` where `serviceError` is `Schema.Codec<Error, Uint8Array>` — shared by `execQuery` and every other RPC in the group. Narrowing the channel to a specific tagged type would mean changing `serviceError` repo-wide, which is out of scope here. The *value* on that channel is a `QueryError`, and `serviceError`'s decoder reconstructs registered error classes, so the tag does survive the boundary — which is what the guideline is protecting.

---

PR #12726, opened 2026-08-24
echo: credentials in an automerge document, behind DX_AUTOMERGE_CREDENTIALS

dmaretskyi:
Adds an Automerge-backed home for the space credential chain, and anchors a space on an immutable **space root document** that points at it. **Opt-in and off by default** — with the flag unset a space keeps its key-derived id and its hypercore control feed, exactly as today. PR A of two; [dxos/edge#945](https://github.com/dxos/edge/pull/945) is the EDGE counterpart.

Nothing is deleted: hypercore removal is a later phase. Design and task ledger in `.agents/projects/hypercore-removal/`.

## The flag

`DX_AUTOMERGE_CREDENTIALS=1` → `runtime.client.automergeCredentials` (config field 17), read by `ClientServicesHost` into `runtimeProps`. It gates three things, all off by default:

1. **New spaces** — `useSpaceRootDocument` defaults to the flag, so a new space is key-derived.
2. **Migration on open** — `_anchorSpaceOnRootDocument` returns early, so no space grows a root and nothing is mirrored.
3. **HALO anchoring** — `_anchorHaloOnRootDocument` returns early.

`_mirrorCredentialsToDocument` — which installs both the write-mirror **and** the document→state-machine subscription — is reachable only from inside the gate, so with the flag off the document is never involved at all.

Even with the flag **on**, the client still reads credentials from the feed: there is no `CredentialSource` yet, so the document is written and mirrored but never authoritative. Flag-on is additive, not a cutover.

## Topology

```text
space id = hash(space genesis key)          # unchanged, both paths
space root doc (immutable, tiny)
   ├─ directory   -> automerge url          (rotatable)
   └─ credentials -> automerge url          (rotatable)
                       └─ credentials keyed by credential id
```

The space id is **always** derived from the space genesis key, and the HALO's from the identity key. The root anchors the credentials document; it does not identify the space, so a peer that never sees the root computes the same id. Recovery depends on this — it reconstructs the HALO from `haloSpaceKey` alone.

## What's in it

- **Space root document** (`echo-protocol`): `SPACE_ROOT_TYPE` as a versioned DXN, holding only rotatable references.
- **Storage** (`echo-host`): migration `0002` adds `space_root_doc_url` and `credentials_doc_url`. `root_doc_url` keeps its existing meaning — the directory — because every reader depends on it.
- **Credentials document** (`credentials`): a map keyed by credential id, so an append is idempotent and concurrent appends converge. `orderCredentials()` computes the total order every peer computes identically, and **every ordering input is read from the encoded credential**, so a peer cannot reorder processing by editing the document. Genesis is ordered first explicitly.
- **Append-only reads** (`AddOnlySet`, `@dxos/echo-doc`): both the client and EDGE read the credentials map from the automerge **change history**, not the materialized state — otherwise a member with write access could revoke another by deleting their credential. Covered end to end on both sides.
- **Admission**: `SpaceMember.space_root_url` (field 7) rides the credential, so an admitted member finds the root from its admission alone. `SpaceMetadata.space_id` (field 13) carries the id.
- **Migration**: a legacy space is anchored transparently when it loads, keeping its id, and its credentials are mirrored into the document. One subscription both backfills the existing chain and dual-writes new credentials.

## Notable fixes found while building this

- A module-scope protobuf codec in `credentials-document.ts` broke **every** workerd import (`Code generation from strings disallowed`), which would have taken out edge's worker tests on merge day. Found via a local link to this branch; now resolved lazily.
- Mesh replication authorization was keyed by a key-derived space id, so a root-anchored space never authorized peers.
- `SpaceInvitationProtocol.admit` never passed the root URL.
- Mirroring materialized credential state for **closed** spaces, defeating lazy loading.
- EDGE cached replication-auth **denials**, so a joiner that dialled before its admission landed was refused for up to 60s (fixed and covered in the edge PR).

## Testing

Tests default to the **hypercore** path — `client-services` 176, `client-e2e` 174, `echo-client` 534, `echo-host` 299, `credentials` 63, `halo-e2e` 22. The automerge path has explicit opt-in tests (five in `client-services`), plus one pinning the shipping default: `without the automergeCredentials flag a space is key-derived and never anchors`.

Covered: a legacy space anchors itself on the next load with nobody calling migrate; data written before the upgrade is still readable after; a credential deleted from the document is still read back; the document replays into the same state the feed reaches, including an admitted member and its role.

## Breaking

`Runtime.Client` pipeline diagnostics `space_root_url` is renamed to `directory_url` (field number 22 unchanged, so the wire is compatible) — it always meant the **directory**, and keeping that name next to `SpaceMember.space_root_url` would collide. All in-repo call sites are updated; an out-of-repo TypeScript consumer reading `pipeline.spaceRootUrl` needs the rename.

## Not in this PR

The **flip** — stopping feed reads for a space whose document is complete — needs a `CredentialSource` split inside `ControlPipeline`. Phase 3 (dual-write, per-space flip) is blocked on choosing who executes migration, and HALO documents have no replication between devices yet. Hypercore deletion is Phase 4.


packages/core/echo/echo-client-e2e/src/integration.test.ts
```diff
@@ -363,8 +364,8 @@ describe('Integration tests', () => {
     teleportConnections[0].teleport.addExtension('replicator', replicator1.createExtension());
     teleportConnections[1].teleport.addExtension('replicator', replicator2.createExtension());
 
-    await replicator1.authorizeDevice(spaceKey, teleportPeer2.peerId);
```
dmaretskyi, line 366: I'd like to see a test of a space being created with hypercore (old style) and then transparently migrated onto the new style

packages/core/echo/echo-protocol/src/space-root.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface SpaceRoot {
+  type: typeof SPACE_ROOT_TYPE;
+
+  spaceId: SpaceId;
+
+  /** Redundant with {@link spaceId} for `rootDoc` spaces, and the only check available for `spaceKey` ones. */
+  idDerivation: SpaceIdDerivation;
+
+  /** Automerge URL (`automerge:<documentId>`) of the space directory ({@link DatabaseDirectory}). */
+  directory: string;
```
dmaretskyi: AutomergeUrl type for link

dmaretskyi:
@coderabbitai review

The standing "Merge Risk: High · up to `6cdbc`" verdict predates the fixes for the things it names. Since then: the empty-document-id case is rejected (`documentIdFromUrl`), credential ordering reads every input from the encoded credential and orders genesis first, the storage model is one keyed map (the plan text that still said "append-only array" was corrected), the selector is `CreateSpaceOptions.useSpaceRootDocument` defaulting on, and root references persist through creation, admission, acceptance and reopen. CI is green on `3a01b39`.

---

packages/core/echo/echo-client-e2e/src/integration.test.ts
dmaretskyi: Added in `796d4834`, but not here — `@dxos/echo-client-e2e` has no `@dxos/client-services` dependency, and `migrateSpaceToRootDocument` lives on `DataSpaceManager`, so adding one would invert the layering. The test is in `packages/sdk/client-services/src/packlets/spaces/data-space-manager.test.ts` next to the other migration coverage:

`data written before migration is still readable through the space root afterwards` — creates the space with `useSpaceRootDocument: false` (hypercore, id derived from the space key), writes an object link into the directory, migrates, then re-opens the space root by space id and asserts the directory url is unchanged and the link still resolves to the same document.

That complements the two migration tests already there: `a legacy hypercore space migrates onto a space root document, keeping its id` (id preserved, `idDerivation: 'spaceKey'`, migration idempotent, control feed still intact) and `a migrated space mirrors its control-feed credentials into the credentials document`.

One thing this does *not* cover, which may be what you were after: migration is currently explicit — `migrateSpaceToRootDocument` is called, not triggered on open — so "transparently" is only true in the sense that nothing the app stored moves. If you want it to fire automatically on opening a legacy space, that's a behaviour change rather than a test, and I'd rather do it deliberately than fold it into this PR.

---

dmaretskyi: Correction to my last reply — I was wrong that migration is explicit. It is already automatic: `_anchorSpaceOnRootDocument` is wired to `stateUpdate` in `data-space-manager.ts:851` and runs on every path that opens a space. The reason the existing tests all call `migrateSpaceToRootDocument` by hand is the suppression at `data-space-manager.ts:509` — a space created legacy stays unanchored for the session that created it, so that there is a pre-migration state to migrate from.

So the transparent path was the untested one. Added in `244a4f75`:

`a legacy space anchors itself on the next load, with nobody calling migrate` — creates the space with `useSpaceRootDocument: false`, asserts it is unanchored, then restarts the manager (a fresh `DataSpaceManager` over the same stores, since the suppression lives on the instance and a plain close/open would still be the creating session), activates the lazily loaded space, and asserts it anchored with `idDerivation: 'spaceKey'` — no migrate call anywhere in the test.

That plus `796d4834` covers both halves of what you asked for: the migration happens on its own, and data written beforehand still resolves afterwards. No behaviour change was needed.

---

PR #12743, opened 2026-08-25
app-framework: serializable plugin entrypoints (dxplugin.jsonc)

dmaretskyi:
A plugin's entrypoint becomes data. Every plugin can ship a `dxplugin.jsonc` next to its `package.json` carrying its metadata **plus its full module list** — id, activation event, `requires`/`provides` capabilities, and a relative URL to each module's file — so a host can read a plugin's modules, activation waves and capability graph without evaluating any of its code.

```jsonc
{
  "key": "org.dxos.plugin.markdown",
  "name": "Markdown",
  "modules": [
    {
      "id": "ReactSurface",
      "src": "./src/capabilities/react-surface.ts",
      "platforms": ["browser"],
      "activatesOn": { "oneOf": [{ "id": "org.dxos.app-framework.event.surfacesRequested", "specifier": "org.dxos.role.article" }] },
      "provides": ["org.dxos.app-framework.capability.reactSurface"]
    }
  ]
}
```

```ts
const MarkdownPlugin = Plugin.fromManifest(await import('@dxos/plugin-markdown/dxplugin.jsonc'));
```

## What's here

- **`@dxos/protocols`** — `Config2.Descriptor` and its `Module` / `CapabilityRef` / `ActivationEventRef` / `Platform` parts.
- **`Capability.fromRef` / `ActivationEvent.fromRef`** — rehydration from the serialized form. This is sound because the plugin manager only ever reads a capability tag's `identifier` and `arity` (`module-loader.ts`, `activation-graph.ts`), never its service type, and a tag is `Context.Service(identifier)` — so a tag rebuilt from `{identifier, arity}` is the *same* context key its owner exported. That is what makes the capability graph serializable at all.
- **`Plugin.fromManifest`** (+ `parseDescriptor`, `parseJsonc`) — accepts a decoded descriptor, a module namespace, or raw JSONC text, with `baseUrl` and `platform` options. Each module activates by importing its `src` and calling that file's default export, so module bodies are unchanged.
- **`dxPluginManifest()`** — the vite loader, modelled on vite's HTML handling: a descriptor is a *document that references modules*, so the loader walks `modules[].src` and hands each file to vite as a first-class module. In dev each `src` becomes a `/@fs/…` URL the dev server already serves (HMR intact, no bundling); in build each is emitted as a chunk via `this.emitFile` and rewritten to `import.meta.ROLLUP_FILE_URL_*`, so `.ts` paths become the shipped `.js` assets and a module reachable only through a descriptor is never tree-shaken away. Wired into composer-app's vite config.
- **`scripts/generate-dxplugin.ts`** — generates a descriptor from an existing TypeScript entrypoint. The two halves of a module declaration live in different places, so it reads activation specs off the constructed plugin and recovers each `src` from the capabilities barrel's `() => import('./x')` loaders, matched positionally against the entrypoint's `addModule` order.
- **plugin-markdown converted** — `dxplugin.jsonc`, the `./dxplugin.jsonc` package export, and `capabilities/translations.ts` (the one module whose body was an inline value rather than a file; it now lazy-loads in its own chunk).

`platforms` replaces the hand-written `plugin.node.ts` / `plugin.workerd.ts` entrypoint variants — one descriptor, filtered at load.

## What's deliberately not here

The old `meta.ts` / `plugin.*` entrypoints are still in place. Deleting them is blocked on one decision: 18 files in plugin-markdown alone import `meta` from `#meta` **statically**, and sourcing it from the descriptor means importing `.jsonc` from TypeScript — which needs an ambient `declare module '*/dxplugin.jsonc'` *and* the loader inside each package's own `ts-build`/rolldown library build, not just the app's vite build. That, and migrating the remaining ~60 plugins (mechanical via the generator), are written up as Phase 2/3 in `.agents/projects/serializable-plugin-entrypoints/TASKS.md`.

## Testing

Fidelity is the property that matters, so the markdown test asserts the descriptor reconstructs the **same activation graph** as the hand-written entrypoints — same modules, same order, same waves, same capability edges — for both the node and workerd variants.

```
app-framework:test    Test Files 22 passed   Tests 259 passed
plugin-markdown:test  Test Files 15 passed   Tests  38 passed
protocols:test        Test Files  8 passed   Tests  34 passed
```

Builds green for `protocols`, `app-framework`, `plugin-markdown`; `moon run :lint` clean on all three; `pnpm format` applied.

https://claude.ai/code/session_01Sf6EwbkmnsWYdtp7pLcgWX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sf6EwbkmnsWYdtp7pLcgWX)_

## Summary by CodeRabbit

* **New Features**
  * Added data-driven `dxplugin.jsonc` manifests for plugin metadata, modules, capabilities, activation events, and platform support.
  * Added automatic plugin loading across browser, Node, and Workerd environments.
  * Added development and production build support for manifest-defined modules.
  * Added JSON Schema validation and VS Code completion for plugin manifests.
  * Migrated the Markdown plugin to the new manifest format.

* **Documentation**
  * Added migration guidance and tooling for generating plugin manifests.


packages/plugins/plugin-markdown/package.json
```diff
@@ -49,16 +49,16 @@
       "types": "./dist/types/src/operations/index.d.ts",
       "default": "./dist/lib/operations.mjs"
     },
-    "#plugin": {
+    "#platform": {
```
dmaretskyi: can we find a way to not have this module? why does plugin need to statically know platform? can app-framework pass that in?

packages/plugins/plugin-markdown/src/dxplugin.d.ts
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+/**
+ * Types the descriptor import. Declared per package because tsc loads an ambient declaration only
+ * from a file in the program, which a dependency's shipped `.d.ts` is not.
+ */
+declare module '*/dxplugin.jsonc' {
```
dmaretskyi: move this to app-framework

packages/plugins/plugin-markdown/src/MarkdownPlugin.ts
```diff
@@ -5,11 +5,23 @@
...
+/** Plugin metadata, available without loading any module body. */
 export const meta = pluginMeta;
 
-/** Constructs the plugin; the body loads on first enable. */
-export const make = Plugin.lazy(meta, () => import('#plugin'));
+/**
+ * The plugin, built from its serialized entrypoint. `platform` resolves through the package's own
+ * export conditions, which is what narrows the descriptor to the modules a host can actually load.
+ */
+export const make = Plugin.fromManifest(descriptor, {
```
dmaretskyi: is this only used for tests? does it need to exist?

packages/plugins/plugin-markdown/src/capabilities/index.ts
```diff
@@ -24,7 +24,7 @@ export const AnchorSort = AppCapability.anchorSort(() => import('./anchor-sort')
 });
 export const CommentConfig = AppCapability.commentConfig(() => import('./comment-config'));
 export const CreateObject = SpaceCapability.createObject(() => import('./create-object'));
-export const Schema = AppCapability.schema(() => import('./schema'));
+export const Schema = Capability.lazyModule('schema', { provides: [AppCapabilities.Schema] }, () => import('./schema'));
```
dmaretskyi: I dont think this entrypoint (+ per-platform variants) are needed anymore now that jsonc hooks directly into specific capability files

packages/plugins/plugin-markdown/vite.config.ts
```diff
@@ -6,30 +6,24 @@ import { defineConfig } from '../../../vite.base.config.ts';
 
 export default defineConfig({
   entry: {
```
dmaretskyi, line 8: can the entrypoint scan be done instead by the same dxPluginManifest plugin? it has an option to take the path to dxplugin.jsonc and then it will add all entrypoints extracted from it to build, also, copying a version of jsonc, converted to json, to dist with correct paths

dmaretskyi: Done in 57c6df3 — `dxPluginManifest({ manifest })`, defaulting to a `dxplugin.jsonc` at the project root.

`buildStart` emits a chunk per `modules[].src`, so the entrypoints are declared rather than inferred from something importing the descriptor. `generateBundle` then writes `dxplugin.json` beside the chunks with each `src` rewritten to the file that actually shipped. `$schema` is dropped on the way out — it points into the workspace's `node_modules`, so it means nothing to a host that fetched the artifact.

`plugin-markdown/dist/lib/dxplugin.json` now names 12 modules and every `src` resolves to an emitted chunk. That closes the build half of the original design: a published plugin carries a descriptor any host can parse as plain JSON, no vite and no comment-tolerant parser needed.

Two notes on scope:

1. The `entry` map in `vite.config.ts` is untouched, and I don't think it should collapse into this. Those entries are the package's public subpath exports (`Markdown`, `Settings`, `skills`, …), which are a different set from the descriptor's module bodies — the descriptor names `src/capabilities/*`, none of which are exported subpaths. The one entry that *was* redundant, `MarkdownPlugin`, is gone with the file it pointed at.
2. I did not add a `./dxplugin.json` export mapping to `dist/lib/`. A types-only subpath already broke `composer-app:bundle` once on this branch (its import map `this.resolve`s every export), and a target that only exists post-build has the same shape of risk. Happy to add it if you want the artifact reachable by specifier — say the word and I'll verify the bundle against it.

---

packages/plugins/plugin-markdown/src/MarkdownPlugin.ts
dmaretskyi: No — deleted in 5b0d341, along with `MarkdownPlugin.testing.ts`.

A tally of what consumers actually use settled the shape: 41 sites want `make`, 2 want the NSID key, and nothing touches anything else. So the loader now synthesizes `make` and `meta` beside the data, and every call site changed only its specifier:

```diff
-import * as MarkdownPlugin from '@dxos/plugin-markdown/MarkdownPlugin';
+import * as MarkdownPlugin from '@dxos/plugin-markdown/dxplugin.jsonc';
   MarkdownPlugin.make(),   // unchanged
```

No new imports in any of the ~20 story files, and `/testing`'s `MarkdownPlugin` re-export collapsed into the same specifier. `#meta` is now a one-line re-export of the loader's `meta`. The CLI is the one host that can't use them — bun hands it the raw JSONC — so it calls `Plugin.fromManifest` with a base resolved through the same specifier.

One thing worth flagging that this surfaced: the ambient declaration reached its types via `import('@dxos/protocols')` from `@dxos/typings`, a package with **no dependencies**. That doesn't error — it degrades to `any`. Every `dxplugin.jsonc` import has been untyped since I introduced it, which is why nothing caught it; a probe (`const a: number = descriptor`) compiling clean is what exposed it. Fixed by moving the declaration to app-framework, where the types resolve. `make` is now `PluginFactory<void>` and the descriptor a real `Config2.Descriptor`.

---

.changeset/serializable-plugin-entrypoints.md
```diff
@@ -0,0 +1,12 @@
+---
+'@dxos/protocols': minor
+'@dxos/app-framework': minor
+'@dxos/plugin-markdown': patch
```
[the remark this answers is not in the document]
dmaretskyi: Stale — the changeset already names only `@dxos/app-framework`, one Group A package. `@dxos/protocols` was dropped from it earlier on this branch.

Added a second changeset in b9c110f rather than extending this one: `plugin-markdown` now has its own consumer-visible change (the `./MarkdownPlugin` export subpath is gone, replaced by the descriptor), which is a `minor` for Group B in its own right — not the automatic patch-bump a Group A change would give it.

---

packages/sdk/app-framework/src/core/plugin.ts
```diff
@@ -638,3 +647,141 @@ export const resolveLazy = (plugin: Plugin): Effect.Effect<Plugin, LazyPluginErr
...
+  // string verbatim and nothing else can contain an unquoted comma before a closing bracket.
+  return JSON.parse(out.replace(/,(\s*[}\]])/g, '$1'));
+};
+
+/**
+ * Shape a descriptor module's file must have: a default export taking the plugin options and
+ * returning the module's activation effect. Unenforceable at the import — the file is named by a
+ * string — so it is checked at runtime and reported as a `missing-default` descriptor error.
+ */
+type ModuleBody<T> = { default?: (options: T) => Effect.Effect<ModuleActivateResult, Error, any> };
```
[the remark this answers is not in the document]
dmaretskyi: Not taking this one — the premise is inverted.

`Error` in `ModuleBody` isn't an untyped channel this code introduced; it's the framework's existing contract for *every* module, declared on `PluginModule.activate` itself (`plugin.ts:142-144`):

```ts
activate: (…) => Effect.Effect<ModuleActivateResult, Error, Capability.Service | Service | Scope.Scope | never>;
```

`ModuleBody` matches that signature deliberately, because a descriptor module's default export has to satisfy the same boundary a hand-written `addModule` body does. Narrowing it to a named error type would make descriptor-loaded modules *stricter* than every other module in the framework and would not typecheck against `PluginModule.activate` — the contract is what it is at the boundary, so the fix would have to change the framework-wide signature, which is well outside this PR.

The errors this code does own are typed: `PluginDescriptorError` with `load-failed`, `missing-default`, `decode-failed` and `relative-src-without-base` reasons, all raised through `Effect.tryPromise<ModuleBody<T>, PluginDescriptorError>` rather than left to escape.

---

.changeset/serializable-plugin-entrypoints.md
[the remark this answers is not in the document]
dmaretskyi: Correcting my previous reply: splitting this into a second changeset was wrong, and `check-changesets` failed on it in ff98a8b's parent.

The repo's rule is one changelog entry per PR, not one per publish group — `agents/instructions/changesets.md:65-69`: work spanning two groups "is still one story and belongs in one file with a line per group." A second file is only for two genuinely unrelated things, and waives the gate with a `# multiple-changesets:` YAML comment. The descriptor replacing the TypeScript entrypoint is one change, so it does not qualify.

Consolidated back into `serializable-plugin-entrypoints.md`, now naming both groups:

```yaml
'@dxos/app-framework': minor
'@dxos/plugin-markdown': minor
```

Which is the shape this finding originally asked for. The one part of my earlier reply that still holds: `@dxos/protocols` had already been dropped, so there is only one Group A package here.

---

PR #12750, opened 2026-08-25
echo: release automerge objects and their documents once nothing holds them

dmaretskyi:
Follow-up to #12745, which landed the retention harness plus the feed-side fix and left `automerge-retention.test.ts` known-red. Both of its tests now pass at the full 2000-object scale, and the CI step gates instead of reporting.

## What pinned an object

`EntityManager._objects` held an `ObjectCore` per object for the life of the database, so a space's client-side footprint tracked everything it had ever loaded rather than what was open. Three independent references pinned each object, and all three had to go for any of them to matter — heap-snapshot retainer paths at each step:

1. **The id-keyed core map** → now `ObjectCoreRegistry`, a `WeakRef` index with a `FinalizationRegistry`. A plain `WeakRef` per id is enough (and better than keying on the root proxy: a nested subproxy also reaches the core, and a tracing GC collects the core/proxy cycle as a group). Cores are pinned strongly for a macrotask after every touch, so a load in flight — id stored now, core read back once the document arrives — cannot be collected mid-read.
2. **`DatabaseImpl._rootProxies`** — a second identity map duplicating `core.rootProxy`, holding every proxy the database ever handed out. Removed.
3. **`LoadOp.result`** in the resolver's op table → now weak. An op outlives the read that created it (its request is kept for the entity's lifetime), so a strong result made the table itself the reason every queried entity stayed resident. A collected result reads as "not loaded", which is what it is; `state` still records that the body was reachable, and `acquire` re-probes such an op rather than answering `undefined` forever.

## What is released with a core

- Its closure-satisfaction request and the listener that request registered on the database context (previously an unbounded per-object dispose list, whose closure captured the core itself). Only once the request has settled: aborting one mid-flight releases its load ops, which cancels the IO a reader is waiting on.
- Its document, but only when the object left the space directory — `retainObjects` / `_evictRemovedObjects`, and only with the last object mounted in that document, never the space root. Releasing the handle unsubscribes the host from the document. A merely collected core leaves its document alone: another object may share it and a load may be in flight.
- Nothing with an unsent write: `RepoProxy.release` refuses a handle the host has not acknowledged (`DocHandleProxy._isAcknowledged`, since the pending-id sets go quiet while a mutation is in flight) and retries once the send settles. Re-acquiring a document cancels a deferred release, and a re-subscribe cancels a queued unsubscribe for the same id, which would otherwise ride the same batch.

## What a re-read does

`getObjectCoreById` rebuilds a core from the document already bound for the object (`_rehydrateCore`) instead of waiting on a load the loader would short-circuit — without it, a read after collection waits for an update nothing will emit. The rebuilt core carries the selected branch, since a fresh `ObjectCore` defaults to `main`.

## Where entity management lives

`DatabaseImpl.getObjectById` / `_loadObjectById` each did their own working-set lookup, deleted-flag filter and proxy minting. That is entity management, so it moved to `EntityManager.getEntityById` / `loadEntityById` and the database class delegates — it is the API surface. Minting is injected as `createEntity` rather than imported, because `echo-handler` is built on `core-db` and importing it there would invert the dependency.

## Tests

`automerge-retention.test.ts` (the `memory` tag, needs `--expose-gc`), 2000 objects:

```
  checkpoint                           heap    delta     wasm      rss  alive  objs  docs.c  docs.h
  A: cold client, data on disk      281.4MB  +66.9MB  975.9MB 2046.6MB      -     0       1    2001
  B: all held by the caller         384.0MB +169.5MB 1015.5MB 2129.5MB   2000  2000    2001    2001
  C: half held by the caller        378.9MB +164.4MB 1015.5MB 2129.6MB   1000  1000    2001    2001
  D: none held by the caller        368.3MB +153.8MB 1015.5MB 2129.6MB      0      0    2001    2001
```

Test 1's heap-budget assertion is replaced by residency (`objs`, `docs.c`), for the reason the harness itself gives for preferring `WeakRef` liveness to bytes: an automerge-backed suite's absolute footprint is dominated by WASM and allocator state that no release returns, so a byte threshold measures the host it runs on. Heap is still printed at every checkpoint. Its removal loop moved into a helper, because a live stack slot holding the last loop variable kept exactly one object alive and read as retention. Its cold read retries a short result: the index source caps each hit at two seconds and a cold read starts 2000 of those clocks at once, so a loaded runner expires the tail of a first pass.

`object-release.test.ts` is new, and runs in the **ordinary** suite — releasing must not cost correctness on a runner with no collector. It drives every release path reachable without one (removal, `retainObjects`, host garbage collection, reopen) and then keeps using the space, since what breaks is not the release but what a later read finds: a sibling of a released document's neighbour, a write issued across a release, one instance per id, a ref resolved twice, a branch binding after a rebuild.

The `memory` job runs the automerge suite once per test. Vitest isolation is per FILE and both tests live in one, so a single invocation left the second test measuring the first test's WASM — a `574.9MB` floor instead of `5.2MB`. `isolate: true` does not fix that (it is file-granular, measured); a fresh process per test does.

`tools/vitest/leak-setup.ts` gives its snapshot hooks a real timeout — `writeHeapSnapshot` of a multi-GB heap took longer than vitest's 10s hook default and failed the file regardless of its assertions.

## Verification

- `echo-client-e2e:test` — 319 passed (8 new); `echo-client:test` — 534 passed; `echo:test` — 577 passed; `echo-host:test` — 292 passed. All re-run after merging `main`.
- `pnpm format`, `echo-client:lint`, `echo-client-e2e:lint` clean.
- CI green on `887fbbfe`, including the `memory` gate.

https://claude.ai/code/session_01GPs1AUJhyPcANfeqfNB3ME

## Summary by CodeRabbit

* **Performance**
  * Reduced client-side memory usage by releasing ECHO objects and associated documents when no longer in use.
  * Improved cleanup while preserving active objects, pending changes, and necessary root documents.

* **Bug Fixes**
  * Improved document release handling during pending writes, reloads, and object recreation.
  * Preserved object identity, references, and branch state across release and reload operations.

* **Testing**
  * Added coverage for object release, reloading, retention, and memory usage.
  * Strengthened retention checks to detect regressions more reliably.


packages/core/echo/echo-client/src/automerge/repo-proxy.ts
```diff
@@ -106,6 +113,43 @@ export class RepoProxy extends Resource {
...
+
+    // Every listener, not just this class's: the entity manager subscribes to each handle too, and a
+    // released handle must not keep either alive.
+    handle.off('change');
+    delete this._handles[documentId];
+    this._pendingAddIds.delete(documentId);
+    this._pendingRemoveIds.add(documentId);
+    this._sendUpdatesJob?.trigger();
+    return true;
+  }
```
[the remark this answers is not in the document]
dmaretskyi: Confirmed and fixed in 27d25288. `find` returning a cached handle left the id in `_deferredReleaseIds`, so the retry after the send settled would call `release` on a handle a live core had since re-acquired — stripping its `change` listeners and dropping it. Re-acquisition now cancels the deferred release on both paths: the cached-handle branch of `_getOrLoadHandle` and `_loadHandle`.

No regression test: exercising it needs a release refused mid-send and a re-acquisition inside that window, which the existing `repo-proxy.test.ts` harness has no seam for (release eligibility is decided from the handle's acknowledged heads, not injectable). The retention suite covers the release path itself end to end.

---

packages/core/echo/echo-client/src/core-db/object-core-registry.ts
```diff
@@ -0,0 +1,154 @@
...
+      return;
+    }
+    this.#drainScheduled = true;
+    // A macrotask, not a microtask: a load resolves across timer/IO turns, and a microtask drain
+    // would close the window while the same operation is still awaiting its document.
+    setTimeout(() => {
+      this.#drainScheduled = false;
+      this.#pinned = new Set();
+    });
+  }
```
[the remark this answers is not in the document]
dmaretskyi: Correct — fixed in 27d25288, essentially as proposed. The code did not match its own comment: the window ended a macrotask after the first touch, so a core pinned just before a pending drain got almost no window. `#pin` now clears and re-arms a `#drainTimer` on every touch, and `clear()` clears it too.

---

packages/core/echo/echo-client/src/core-db/load-op.ts
```diff
@@ -88,11 +97,17 @@ export class LoadOpTable {
...
       uri,
       maxCeiling: source,
       state: 'pending',
-      result: null,
+      get result() {
+        return resultRef?.deref() ?? null;
+      },
+      set result(value: AnyProperties | null) {
+        resultRef = value == null ? null : new WeakRef(value);
+      },
```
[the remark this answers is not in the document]
dmaretskyi: Real gap — fixed in 27d25288. A cached op left `ready` with a collected result answered every later resolution of that URI with `undefined` for as long as the op stayed in the table (state said ready, so `RequestImpl.getResult()` returned the null result). `acquire` now detects `ready` with no result, re-probes the working set, and falls back to `#startLoad` when the probe misses.

Worth noting for the design: this is only reachable because `LoadOp.result` is now weak, and it is the price of the table no longer pinning every entity a query ever touched.

---

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -354,6 +379,39 @@ export class EntityManager implements IDatabaseBinding {
...
+    const root = this._spaceRootDocHandle;
+    if (root == null || !root.isReady()) {
+      return undefined;
+    }
+    if (DatabaseDirectory.getInlineObject(root.doc(), objectId) != null) {
+      return root;
+    }
+    const url = this._getLinkedDocumentUrl(objectId);
+    return url == null ? undefined : this._repoProxy.handles[interpretAsDocumentId(url)];
+  }
```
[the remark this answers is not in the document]
dmaretskyi: Half of this was real and is fixed in 27d25288; the other half no longer applies.

**Valid**: a rehydrated core came from a fresh `ObjectCore`, whose `branch` defaults to `'main'`, so `Obj.getBranch` would report `main` for an object actually bound to a branch document. `_rehydrateCore` now sets `core.branch = this.getCurrentBranch(objectId)`.

**Not applicable**: the document itself was never wrong. Rehydration reads `_objectDocumentHandles`, which `_rebindMemberToBranch` sets to the branch handle at the same moment it records `_currentBranches` — the two cannot disagree. The `_findLoadedDocumentHandle` fallback your comment refers to (deriving the document from the space root's `links`, which would indeed have picked main) was removed in a59dda96, before this review landed, for an unrelated reason: it cost two extra automerge lookups per miss on the cold-read path.

---

packages/core/echo/echo-client-e2e/src/automerge-retention.test.ts
```diff
@@ -46,6 +45,29 @@ const PAYLOAD_BYTES = 1 * 1024;
...
+ * in a cold read starts that clock at once while the documents arrive over the following minute — so
+ * on a loaded runner the tail of the first pass can expire. The retry reads what has since landed;
+ * what is under test is what the space releases afterwards, not how fast a cold read completes.
+ */
+const queryAll = async (db: EchoDatabase, expected: number): Promise<Obj.Unknown[]> => {
+  let objects: Obj.Unknown[] = [];
+  for (let attempt = 0; attempt < 5 && objects.length < expected; attempt++) {
+    objects = await db.query(Query.select(Filter.type(TestSchema.Task))).run();
+  }
+  return objects;
```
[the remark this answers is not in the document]
dmaretskyi: Took the wait, skipped the throw — 18e0a1bd.

The wait is a good catch: an immediate re-query reads the same not-yet-loaded state, so retries could burn out without the documents having had time to land. There is now a 1s pause between attempts.

Not throwing from the helper: both call sites already assert `expect(queried.length).toBe(OBJECT_COUNT)` on the next line, so a short final result fails the test either way — and it fails with vitest's diff (`expected 1998 to be 2000`) rather than a hand-thrown error, which is the more useful failure. A throw inside the helper would only move the same failure earlier and lose that.

Also on the review's nit about moving `removeAll` below the closing `describe`: skipping it here, because this file's other helpers (`addObjects`, `setupColdPeer`, `queryAll`) all sit above the suite, and moving one of four would make the file less consistent, not more. If the guideline should win, that is a whole-file move and belongs in its own change.

---

.github/workflows/check.yml
```diff
@@ -282,8 +280,7 @@ jobs:
           DX_DEBUG_LEAKS_DIR: ${{ runner.temp }}/profiles
         run: moon run echo-client-e2e:test -- src/feed-retention.test.ts
 
-      - name: Automerge retention (known red, reporting only)
-        continue-on-error: true
+      - name: Automerge retention
```
dmaretskyi: any way to enable per-test isolation? prior tests pad wasm memory which doesn't constrict

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -104,7 +106,12 @@ export class EntityManager implements IDatabaseBinding {
   readonly _repoProxy: RepoProxy;
 
   // ── Object storage ──────────────────────────────────────────────────────
-  private readonly _objects = new Map<string, ObjectCore>();
+  /**
+   * Loaded cores, held weakly: a space holds what the caller has open rather than everything it has
+   * ever read. A collected core takes its document handle and satisfaction request with it — see
+   * {@link _releaseObject}.
+   */
+  private readonly _objects = new ObjectCoreRegistry({ onRelease: (id) => this._releaseObject(id) });
```
dmaretskyi, line 121: make this be held my entity manager instead

packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -125,11 +132,17 @@ export class EntityManager implements IDatabaseBinding {
...
+   * context's dispose list proportional to them. Strong-dependency satisfaction is delegated to the
+   * {@link RefResolver}: each surfaced entity holds a disk-bound request whose `ready` state is the
+   * surface gate, spanning same-space db, cross-space db, feed queues, and the registry.
+   *
+   * Neither map holds the entity: a request reaches it only through its load op, whose result is weak
+   * ({@link LoadOp.result}), so an object is released with its core rather than pinned by the gate
+   * that surfaced it.
    */
   private readonly _satisfactionRequests = new Map<EntityId, RefResolverRequest>();
+  private readonly _satisfactionSubscriptions = new Map<EntityId, CleanupFn>();
```
dmaretskyi, line 154: try to move as much of this as possible into entity manager

.github/workflows/check.yml
dmaretskyi: Done in 238823b6, though not the way I first tried — vitest has no per-*test* isolation, only per-file, so the fix had to be one process per test.

What I measured before settling on that:

1. **`isolate: true` in the instrumented block** (`vite.base.config.ts` forces `isolate: false` today). It does work with `--expose-gc` — the flag reaches each per-file fork, and the feed suite passes under it, so the config comment's worry about `execArgv` not reaching the tests only applies to the threads pool, not forks. But it does **not** fix this: both automerge tests live in one file, so the second still started at `wasm 574.9MB` instead of `5.2MB`. Reverted.
2. **One invocation per test** (`-t '<name>'`), which is what landed. A fresh process is the only clean WASM floor: run alone, test 1 starts at `wasm 5.2MB`; run second in a shared process, `574.9MB`.

The `memory` job now has two automerge steps. Cost is one extra cold start (~10s) against a 160s suite.

Two notes:
- The feed suite has the same shape (two tests, one file) and I left it as one invocation — its WASM is ~5MB, so padding there is not yet distorting anything. Say the word and I'll split it for symmetry.
- Padding no longer affects pass/fail, since the assertions are residency (`alive`/`objs`/`docs.c`), not bytes. It did distort the printed floor, which is what your comment is about, and that is now clean.

---

packages/core/echo/echo-client/src/core-db/entity-manager.ts
dmaretskyi: Moved in 238823b6, taking your chat note ("the echo-database class is mainly for api") as the scope.

`DatabaseImpl.getObjectById` and `_loadObjectById` each did their own working-set lookup, deleted-flag filter and proxy minting. Both are now one-line delegations to `EntityManager.getEntityById` / `loadEntityById`, which own that logic beside the working set they read.

The one thing that could not simply move is minting itself: `echo-handler` is built on `core-db`, so importing `initEchoReactiveObjectRootProxy` there would invert the dependency. It is injected instead — `EntityManager` takes a `createEntity` in its props and `DatabaseImpl` passes `(core) => initEchoReactiveObjectRootProxy(core, this)`. The logic moves; the layering doesn't.

What I deliberately left in `DatabaseImpl`, so you can tell me if you want it moved too:

1. `runMigrations` / `#runObjectMigration` / `#runRenameMigration` — entity mutation loops, but driven by `Migration.*` API types and the query API. Worth moving, bigger than this PR.
2. `_addObject` — the object-creation half is schema/API marshalling (`createObject`, registry checks) ahead of a plain `addCore`.
3. `#clientLoadedStats` — the entity half is already `EntityManager.loadedStats()`; what is left is feed and registry residency, which the database owns.

---

packages/core/echo/echo-client/src/core-db/entity-manager.ts
dmaretskyi: I can't tell which field this is anchored to — the line lands on `_currentBranches` / `_branchStore`, and both are already `EntityManager` fields, so I've likely mis-resolved the anchor. Rather than guess, here is what I did do and what I'd need from you:

Moved in 238823b6: entity surfacing (`getEntityById` / `loadEntityById`) out of `DatabaseImpl` and into `EntityManager` — see my reply on the other thread for the detail and for what I left behind.

If this comment meant something else, which did you mean?

1. `DatabaseImpl.#feeds` — the feed-handle map, so the manager owns feed residency too.
2. `ObjectCoreRegistry`'s pin set — currently the registry's own state rather than the manager's.
3. `_currentBranches` / `_branchStore` — already here; possibly you want the *store* injected differently.
4. Something else — point me at it and I'll move it.

Not blocking on this: you asked me to land, so I'm proceeding with the rest and will take this as a follow-up unless you say otherwise.

---

dmaretskyi: Both of the outside-diff findings from the latest review were real and are fixed in 93486651 — recording them here since they have no thread of their own.

**1. Eviction skipped a collected core's document.** `_evictRemovedObjects` filtered removals by `_objects.has(objectId)`, so a core collected *before* its directory entry was removed skipped release entirely: the document stayed bound, and `_rehydrateCore` could resurrect an object the directory no longer lists. That is the worse half — not a leak but a resurrection. The predicate now counts a bound document as much as a live core.

**2. An unsettled request was never dropped.** Release left a still-resolving satisfaction request in `_satisfactionRequests` / `_satisfactionSubscriptions` on the theory that "the object's next release will get it" — wrong, because the finalizer that reported the release fires exactly once. It is now dropped when it settles: deferred by a turn (aborting from inside the request's own state change would cancel the IO of ops whose refcount drops, which is the failure I chased earlier in this PR), and skipped if the object was re-acquired in the meantime.

Also in that commit: the CI break was mine — `object-release.test.ts` asserted identity via `_loadObjectById`, which is internal to `DatabaseImpl` and not on `EchoDatabase`. `moon run …:test` transpiles without typechecking, so only `:build` caught it; it now compares two query passes, which is the public route to the same invariant. And the flagged comments are trimmed to one clause each.

Verified: `echo-client-e2e:test` 319 passed, `echo-client:test` 534 passed, both `:build` targets typecheck, lints clean, and each retention test passes standalone from a clean `wasm 5.2MB` floor.

---

PR #12757, opened 2026-08-26
ci: partial-clone every full-depth checkout and drop the no-op LFS fetch

wittjosiah:
Adds `filter: blob:none` to the eleven `fetch-depth: 0` checkouts that still lacked it, and removes `lfs: true`, which fetches nothing in this tree.

## Why this regressed

#12485 (*"ci: partial-clone the checkouts in Check"*, Aug 6) converted every checkout in `check.yml` — but it was scoped to that one workflow. #12482 (*"e2e: bundle once, then shard by browser × moon --job"*) merged five days later and introduced the `e2e-bundle` job with a hand-written `fetch-depth: 0` + `lfs: true` block copied from the pre-#12485 template. So Check regressed one job, and the ten call sites outside Check never had the filter at all.

## What it saves, measured on CI

Same workflow (`PR Build` → `build`), two runs 40 seconds apart on the same runner class:

| | run | Checkout |
| --- | --- | --- |
| unfiltered | [708](https://github.com/dxos/dxos/actions/runs/32965548500/job/98167117135), a branch without this change | **25 s** |
| filtered | [709](https://github.com/dxos/dxos/actions/runs/32965608699/job/98167300815), this branch | **12 s** |

`model-fixture`, also converted here, checks out in 13 s on this branch. Both figures now sit with Check's already-filtered jobs, which run 10–14 s — against 43 s for `e2e-bundle` on the unfiltered path.

So roughly **13 s per checkout**: ~26 s off every PR (`pr-build` + `model-fixture`), ~30 s off `e2e-bundle` on main and release runs, and the dispatch/schedule workflows when they run.

## Why the filter is what matters

Against this remote, one checkout of `main` measured locally (a sandbox with far less bandwidth than a Depot runner, so treat the absolute numbers as an upper bound — the *ratio* is what carried over to CI above):

| strategy | wall | `.git` |
| --- | --- | --- |
| `fetch-depth: 0`, no filter | 62 s | **1.10 GiB** |
| `fetch-depth: 0` + `filter: blob:none` | 13 s | **198 MiB** |
| `--depth=1` (the floor) | 9 s | 144 MiB |

The 13 s splits into 5 s for the commit graph (28,483 commits, 394,950 commit+tree objects, 58 MiB) and 8 s backfilling the 16,375 working-tree blobs the filter deferred. Since depth-1 is only 4 s cheaper, nearly all the remaining cost is materialising the working tree, which every job needs regardless.

Full depth is load-bearing: `--affected remote` and `merge-base` resolve against the commit graph. What the filter drops is every *historical* version of every file, which nothing in CI reads. The working tree is byte-identical either way.

## Why `lfs: true` goes

`.gitattributes` declares LFS filters for `*.dxprofile`, `*.spline` and `*.splinecode`, but `git ls-files` matches **zero** tracked files against them — the fetch is a round trip for nothing. This is the same finding the comment at `check.yml:77` already records.

It stays on the **EDGE** checkout in `edge-tests.yml`, whose tree is `dxos/edge` and cannot be audited from here.

## Sites changed

`check.yml` (`e2e-bundle`), `pr-build.yml`, `pkg-pr-new.yml`, `todo-tracker.yml`, `model-fixture.yml`, `edge-tests.yml` (×2), `deploy-apps.yml` (×2), `deploy-tauri.yaml` (×2).

Deliberately left unfiltered, having never carried `lfs: true`: the `release` job in `deploy-apps.yml`, `publish-all.yml`, `introspect-cache.yml`, `upload-introspect-cache.yml`, `claude-composer-implement.yml`. Worth a follow-up, but the release-path ones deserve their own change.

## Verification

- All 23 `fetch-depth: 0` sites audited; every one either carries the filter or is on the list above.
- Every workflow file parses (`yaml.safe_load`), with each checkout's `filter`/`lfs` values confirmed intact.
- `pnpm format` / `oxfmt --check` clean.
- No changeset: CI-only, per `agents/instructions/changesets.md` rule 2. The `changeset-reminder` comment naming `@dxos/react-ui-assistant` on this PR is a false positive unrelated to this diff — that job checks out `refs/pull/N/merge`, which tracks main's moving tip, while `CHANGESET_BASE_REF` stays frozen at the event's `base.sha`, so #12756 landing mid-run got attributed here. The gating `check-changesets` pins `ref:` to the head sha and is unaffected.
- **Only 3 of the 8 files are exercised before merge.** `check.yml`, `pr-build.yml` and `model-fixture.yml` run on `pull_request`; `pkg-pr-new` is push-only, and `todo-tracker` / `edge-tests` / `deploy-apps` / `deploy-tauri` are dispatch, schedule or `workflow_call`. `check.yml`'s own change is the `e2e-bundle` job, which is gated off PRs — it first runs on the merge to main.
- **Not verified:** `filter: blob:none` alongside `lfs: true` on the EDGE checkout. The two inputs are independent in `actions/checkout` (one goes to `git fetch`, the other runs `git lfs fetch`), but `git-lfs` isn't installed in this sandbox, so the combination is untested and `edge-tests.yml` won't exercise it without a dispatch.

https://claude.ai/code/session_0111otfs5FgyWGGCtZkch6jS

## Summary by CodeRabbit

- **Chores**
  - Updated automated build, test, deployment, and packaging workflows to use Git’s blobless partial clones.
  - Reduced unnecessary repository content fetched during workflow checkouts.
  - Preserved required large-file support where applicable.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -440,7 +440,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           persist-credentials: false
 
       - name: Extract branch
--- a/.github/workflows/deploy-apps.yml
+++ b/.github/workflows/deploy-apps.yml
@@ -252,7 +252,6 @@ jobs:
     env:
       DEPLOY_ENV: ${{ needs.plan.outputs.env }}
     steps:
-      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - name: Checkout
         uses: actions/checkout@v5
         with:
--- a/.github/workflows/deploy-apps.yml
+++ b/.github/workflows/deploy-apps.yml
@@ -252,11 +252,12 @@ jobs:
     env:
       DEPLOY_ENV: ${{ needs.plan.outputs.env }}
     steps:
+      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - name: Checkout
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           # On production, the bumped release commit; otherwise the triggering commit.
           ref: ${{ needs.release.outputs.sha || github.sha }}
 
@@ -330,7 +331,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           ref: ${{ needs.release.outputs.sha || github.sha }}
 
       # Needed before "Check app targets this environment" — apps.mjs imports json5 from node_modules.
--- a/.github/workflows/deploy-tauri.yaml
+++ b/.github/workflows/deploy-tauri.yaml
@@ -131,7 +131,6 @@ jobs:
     timeout-minutes: 60
 
     steps:
-      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
@@ -353,7 +352,6 @@ jobs:
     continue-on-error: true
 
     steps:
-      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
--- a/.github/workflows/deploy-tauri.yaml
+++ b/.github/workflows/deploy-tauri.yaml
@@ -131,10 +131,11 @@ jobs:
     timeout-minutes: 60
 
     steps:
+      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           ref: ${{ inputs.ref || github.sha }}
 
       - name: Install Linux dependencies
@@ -352,10 +353,11 @@ jobs:
     continue-on-error: true
 
     steps:
+      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           ref: ${{ inputs.ref || github.sha }}
 
       - uses: tecolicom/actions-use-homebrew-tools@v1
--- a/.github/workflows/edge-tests.yml
+++ b/.github/workflows/edge-tests.yml
@@ -33,17 +33,20 @@ jobs:
         with:
           detached: true
 
+      # Blobless: full history without every historical file version. Nothing in the tree is LFS-tracked.
       - name: Checkout DXOS
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          lfs: true
+          filter: blob:none
           path: dxos
 
+      # `lfs: true` stays: dxos/edge may track LFS files, unlike this tree.
       - name: Checkout EDGE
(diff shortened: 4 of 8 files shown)


wittjosiah:
## `boot-budget` is red here, and it is not this PR's

Standing down on this failure rather than fixing it in this PR. Evidence:

**It is red on the base branch too.** `main` at `5755a428` — this PR's merge base — fails the same check: [boot-budget on the main push run](https://github.com/dxos/dxos/actions/runs/32969273731/job/98178935100). Every other job on that main run is green.

**The bundle is byte-identical to main's.** On [this PR's run](https://github.com/dxos/dxos/actions/runs/32970511711/job/98182881413), `composer-app:bundle` was a *remote cache hit* (`4909a71f`), so `check-boot-budget` measured the artifact main already built. This PR's diff is 11 lines of workflow YAML and cannot move a byte of the boot graph.

```
```

Entries are fine (22/25). Bytes are marginally over `MAX_PRELOAD_BYTES = 4.25 * 1024 * 1024` (4,456,448) — both sides round to "4.25 MB" in the message, so the overage is small, likely single-digit KB.

**Where it came from.** `main`'s last green `boot-budget` was `8db69c61` (#12756). Two commits followed:

| commit | PR | boot-budget on its main run |
| --- | --- | --- |
| `e954c0ff` | #12753 — resolve `google.protobuf.Any` in shape-compat; metadata stores to buf | run **cancelled**, never reported |
| `5755a428` | #12754 — validate docs links on PRs | **failure** |

#12754 touches only `.github/workflows/check.yml` and `docs/astro.config.ts`, neither boot-reachable. By elimination the regression entered with **#12753**, whose own main run was superseded before `boot-budget` could report it — which is how it landed unnoticed. `packages/core/protocols/src/buf/shape-compat.ts` (+96) and `service-rpc.ts` are in a boot-reachable package, so buf-generated code or the `@bufbuild/protobuf` runtime reaching the eager graph is the plausible mechanism — but I have not traced the import chain, so treat that last step as a hypothesis, not a finding. `src/vite/trace-boot-leak.ts` exists for exactly that attribution.

**No fix exists to port.** No open PR addresses this; I searched. So there is nothing to bring into this branch that would no-op once main carries it.

**Not spending the re-run.** A re-run cannot inform this: the byte count is deterministic and the bundle came from cache, so it would reproduce identically. The base-branch evidence above already establishes the failure is not this PR's.

This PR stays blocked until `boot-budget` is green on `main`, since the merge queue runs Check too. The fix belongs in its own change — either making the newly-eager import lazy, or a deliberate budget raise with the growth reviewed, per the header in `check-boot-budget.mjs`.

---

wittjosiah:
Correcting my previous comment: I claimed this PR "stays blocked until `boot-budget` is green on `main`, since the merge queue runs Check too." That is wrong.

`boot-budget` is not a required check. Out of draft, this PR reports `mergeable_state: unstable` — non-required checks failing, still mergeable — and both #12753 and #12754 merged into `main` with *failing* merge_group runs ([32967397302](https://github.com/dxos/dxos/actions/runs/32967397302), [32968769932](https://github.com/dxos/dxos/actions/runs/32968769932)) for this same check.

Everything else in that comment stands: the failure is pre-existing on `main`, this diff cannot affect the boot graph, and the overage traces to a new module-scope `import { StructSchema } from '@bufbuild/protobuf/wkt'` in `packages/core/protocols/src/buf/shape-compat.ts` (#12753) — `@bufbuild/protobuf` exposes no finer subpath, and `StructSchema` is used only inside two function bodies, so a deferred import would keep it off the eager graph. That remains a separate change, and the byte attribution is still unmeasured.

---

PR #12758, opened 2026-08-26
config: finish edge main → preview rename left over from #12759

wittjosiah:
## Summary

Follow-up to #12747 (pointed dev-tier clients at the new EDGE preview environment) and #12759 (which merged first and superseded most of this PR's original diff, including its own fix for the docs link-validation failure this branch hit along the way — see the PR comments below for that detour). What's left after rebasing onto `main`:

- `packages/sdk/config/src/preset.ts`: `configPreset`'s default `edge` value (and its `@default` doc comment) now say `preview` instead of `main`. `'main'` stays as a deprecated alias resolving to the same `https://preview.dxos.network` URL, so this is a no-op for existing behavior.
- `packages/plugins/plugin-script/src/e2e/deploy.test.ts` and `packages/stories/stories-inbox/src/stories/MailboxSync.stories.tsx`: explicit `configPreset({ edge: 'main' })` calls updated to `edge: 'preview'`.

No changeset — pure rename, no observable behavior change.

## Test plan

- [x] `pnpm format` (oxfmt) — clean
- [x] CI (`Check` workflow) — green


wittjosiah:
The `check` job's "Check docs links" step failed on this branch's first two pushes (re-run confirmed it's not a flake). Root cause is unrelated to this PR's diff: `content/legal/terms.md` links to `/privacy`, an Astro custom page (`src/pages/privacy.astro`) rather than a content-collection entry, which `starlight-links-validator` can't resolve — a pre-existing gap from #12507 that this PR happened to be the first to hit since the last docs-affecting push. Pushed a fix (175aaf28) adding `/privacy` and `/terms` to the validator's `exclude` list.

---

PR #12760, opened 2026-08-26
edge-client: floor connection uptime to whole seconds

dmaretskyi:
## Problem

Every `EdgeStatus` encode logged:

```
Error: cannot encode field dxos.client.services.EdgeStatus.uptime to binary: invalid int32: 683.241
```

`EdgeStatus.uptime` is declared `int32` in `dxos/client/services.proto`, but `EdgeWsConnection.uptime` returned `(Date.now() - openTimestamp) / 1000` — a fractional value — so the codec rejected it on each status query.

## Change

Floor the getter to whole seconds (`packages/core/mesh/edge-client/src/edge-ws-connection.ts`). The sibling fields (`rtt`, rates, message counts) are already integers.

## Testing

Toolchain (`moon`, `oxfmt`, `node_modules`) is not installed in this cloud sandbox, so no local build/test was run — relying on CI's Check workflow.

---
_Generated by [Claude Code](https://claude.ai/code/session_012awqy1Pm5b7Jt7Wmr8mJmg)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed edge connection uptime reporting to use whole seconds, preventing connection status encoding errors.
  * Improved reliability when reporting edge connection status.

* **Release**
  * Included this fix in a patch release of the edge client.

### DIFF
--- a/.changeset/edge-status-uptime-int.md
+++ b/.changeset/edge-status-uptime-int.md
@@ -1,0 +1,5 @@
+---
+'@dxos/edge-client': patch
+---
+
+Report edge connection uptime as whole seconds, fixing the `invalid int32` error thrown when encoding `EdgeStatus`.
--- a/packages/core/mesh/edge-client/src/edge-ws-connection.ts
+++ b/packages/core/mesh/edge-client/src/edge-ws-connection.ts
@@ -87,7 +87,7 @@ export class EdgeWsConnection extends Resource {
     return this._rtt;
   }
 
-  /** Whole seconds; the wire type (EdgeStatus.uptime) is int32 and rejects fractions. */
+  /** Floor uptime to satisfy the int32 EdgeStatus.uptime wire type. */
   public get uptime(): number {
     return this._openTimestamp ? Math.floor((Date.now() - this._openTimestamp) / 1000) : 0;
   }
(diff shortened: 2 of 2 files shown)


dmaretskyi:
`boot-budget` is failing on this branch and I don't believe it belongs to this PR.

The failure ([job](https://github.com/dxos/dxos/actions/runs/32970839204/job/98183968570)):

```
```

Entries are fine (22/25); the byte ceiling (`MAX_PRELOAD_BYTES = 4.25 * 1024 * 1024` in `packages/apps/composer-app/scripts/check-boot-budget.mjs`) is exceeded by less than the 0.01 MB the report can resolve — the eager graph is sitting right at the line.

This PR's diff is one `Math.floor(...)` inside an existing `@dxos/edge-client` getter plus a `.changeset/*.md`. That adds roughly a dozen bytes of emitted JS and no new import edge, so it cannot account for the overage. It reproduced identically on both heads pushed here (`77f01661` and `c73c0822`), so it is not a flake either — no re-run needed to establish that.

I'm not fixing it inside this PR: the two available remedies are re-baselining `MAX_PRELOAD_BYTES` or evicting something from the eager graph, and the script's own comment makes raising the budget an explicit review decision ("Raise a budget only for growth you have looked at and accepted — that is the review point"). Doing either here would widen an unrelated one-line bug fix, and no fix for it exists elsewhere yet that I could port.

Happy to open a separate PR that either re-baselines the ceiling with the current numbers or traces what recently entered the boot graph, if that's wanted.

---

PR #12765, opened 2026-08-26
compute: remote process control (agent processes on EDGE)

dmaretskyi:
Client-side half of running `Process` instances — `AgentProcess` first — on EDGE, so Composer can start an agent chat that lives in the cloud and keeps running with the client closed. The EDGE host itself is [dxos/edge#971](https://github.com/dxos/edge/pull/971); this PR is what it speaks to.

## How it fits the existing interfaces

`@dxos/compute-runtime` has two manager tags, and they mean different things: `ProcessManager.Service` is **local, in-process** execution; `RemoteProcessManager.Service` **talks to EDGE**. `ProcessMonitor.layer` merges their trees, which is what gives a UI the local/remote split. So the remote surface grows on its own tag rather than impersonating the local one:

- `ProcessProtocol` (`@dxos/protocols`) — wire types for the seven process routes, shared by client and worker rather than structurally re-declared on each side.
- `RemoteProcessManager.Control` — the transport-agnostic control surface (spawn by process key, list, status, submit input, terminate, cursor event reads, RPC client), stated in **domain** types: `Snapshot extends Process.Info`, an `Event` union, `EventPage`, branded `SpaceId`/`Process.ID`. No `ProcessProtocol` type appears in the interface; decoding is the transport's job.
- `RemoteProcessManager.Manager` also carries the manager verbs — `spawn`, `list`, `attach`, `refreshProcessTree` — each taking the space it addresses, so one manager serves every space. `makeControlVerbs` implements them once over a `Control`, and a manager built without one simply lacks them.
- `RemoteProcessHandle` — cursor-polled output and ephemeral trace, so a client that reloads resumes an in-flight remote process. `runToCompletion`/`runUntilSettled` are derived client-side, so the settle predicates cannot drift from `ProcessHandle`'s.
- `EdgeProcessControl` + `EdgeProcessManager` (`@dxos/edge-compute`) — the EDGE transport, plus the wire decoding (`process-snapshot.ts`) that keeps it out of the domain interface.
- `AgentService.getSession(chat, { location: 'edge' })` — the whole request. `AgentService` picks the local or remote verbs itself; unifying the two is a caller's job, not a façade's. `hydrate` walks both runtimes, since an edge agent that kept running while the client was closed is the point.
- `Process.Monitor.list(filter)` — the read-only aggregator over local **and** remote runtimes, so a caller can find its own process without holding a handle. A space-scoped read re-reads the host first, so it stays accurate when the client was not the one that acted.

Two small additions carry their weight:

- `Process.Process` now exposes its `input`/`output` codecs (already on the value, missing from the interface), so values cross the boundary encoded by the definition's own schema instead of being assumed JSON.
- `Process.Handle.alarmDueAt` — a host that suspends a process between turns (a Durable Object) cannot rely on an in-memory timer, so it needs the due-time to mirror onto a platform alarm; it is also the signal that separates `runToCompletion` from `runUntilSettled` remotely.

Design and task ledger: `.agents/projects/agent-process-edge/` (decisions D1a and D1b record the two corrections this took).

## Testing

- `RemoteProcessManagerVerbs.test.ts` — spawn/list/status, input encoding and output streaming, terminate, `runUntilSettled`, annotation JSON-safety, and the monitor tree, against an in-memory stand-in host (12 tests).
- `compute` 76, `compute-runtime` 205, `agent-runtime` 44, `plugin-routine` 61 passed; repo-wide `:lint`, `:test-types`, `knip`, `:check-module-structure` and `oxfmt` clean.
- End-to-end against a real worker is the edge half: 16 tests there — 9 covering the process routes, 6 the compute APIs, and 1 driving `AgentService` to spawn, observe and terminate an agent hosted on EDGE.

Also fixed here: three test budgets that made the `test` shards flaky under load, each aligned with the convention in its own file (`client-e2e` 200 ms → a hang guard, `plugin-thread` and `echo-host` given the timeouts their siblings already name).

https://claude.ai/code/session_01J4JXLi6a9Gxs5DNwTwmZXW

---
🚀 Composer preview: https://pr-12765-composer-dev.dxos.workers.dev


dmaretskyi:
`boot-budget` (`composer-app:check-boot-budget`) is red here, and the residual failure is not this PR's — it reproduces on `main`. Standing down on it with the evidence, since no fix exists to port yet.

**This PR did contribute at first, and that part is fixed.** Bundling Composer locally measured the eager boot graph at **4,458,496 bytes against the 4,456,448 ceiling — 2,048 over**. 1,095 of those were mine: the seven process routes added to `EdgeHttpClient`, a class Composer instantiates during boot, so its methods ship whether or not the app calls them. They moved to an `EdgeProcessHttpClient` subclass behind a new `@dxos/edge-client/process` subpath (65f3227a), which took the graph to **4,457,401** and left the base class byte-identical to `main`.

**The remaining 953 bytes are not from this branch.** Three independent checks:

1. Moving `RemoteProcessHandle`/`RemoteProcessManagerAdapter` off the `@dxos/compute-runtime` barrel to a `./remote-process` subpath (b09e034f) changed the total by **zero bytes** — they were already tree-shaken.
2. Scanning all 22 boot chunks' sourcemaps, **none** of the new modules appear, and neither do the pre-existing files this PR modifies (`ProcessHandle.ts`, `ProcessManager.ts`, `compute/src/Process.ts`, `protocols/src/index.ts`). The only file of mine in the boot graph is `edge-http-client`, now unchanged from `main`. The five chunks without sourcemaps are 4.5 KB of rolldown/preload runtime helpers.
3. **dxos/dxos#12759**, an unrelated change, has the same `boot-budget` failure with every other check green.

`check-boot-budget`'s own docstring puts the current baseline at 4.05 MB against a 4.25 MB ceiling; the graph is now at the ceiling, so any boot-reachable byte fails. #12761 passed at 13:28 today and `main` has moved since (it was force-updated), so something in that advance spent the remaining margin.

Not re-running the job: the check is deterministic over the built bundle, so a re-run reproduces it. Two ways forward, neither of which is this PR's to decide — raise `MAX_PRELOAD_BYTES` for growth that has been looked at and accepted (what the script's docstring invites), or find and undo whatever on `main` consumed the last ~1 KB. Happy to do either on request.

The three subpath exports stay regardless: they're what keeps this feature's transport, control plane and protocol out of the eager graph, and they're the remedy the check's own message recommends.

---

packages/core/compute/compute/src/Process.ts
```diff
@@ -235,6 +235,14 @@ export interface Process<
 
   readonly services: readonly Context.Key<any, any>[];
 
+  /**
+   * Codecs for the process's inputs and outputs, from {@link MakeProcessOpts}. Exposed on the
+   * interface so a caller that moves an input across a boundary (a remote runtime) can encode it
+   * with the definition's own schema rather than assuming the value is already JSON.
+   */
+  readonly input: Schema.Codec<any, any>;
```
dmaretskyi: use type params

packages/core/compute/edge-compute/src/edge-client.ts
```diff
@@ -32,16 +33,26 @@ export type UploadWorkerArgs = {
   assets: Record<string, Uint8Array>;
 };
 
-/** @deprecated Migrate to `client.edge`. */
-export const createEdgeClient = (client: Client): EdgeHttpClient => {
+/** Resolves the configured EDGE url and binds the client to the current identity. */
+const configureEdgeClient = <T extends EdgeHttpClient>(client: Client, make: (edgeUrl: string) => T): T => {
```
dmaretskyi: why?

packages/core/compute/compute/src/Process.ts
dmaretskyi: Done in 79151cc7 — `readonly input: Schema.Codec<_Input, any>` / `readonly output: Schema.Codec<_Output, any>`.

Worth noting what it caught: `AgentService` restated the agent handle's input as `string | ContentBlock.Any[]`, but `AgentProcess`'s schema is `Schema.Union([Schema.String, Schema.Array(ContentBlock.Any)])`, whose Type has a **readonly** array. `Codec<any, any>` erased that, so the hand-written type had silently drifted from the schema; with the parameters in place it stopped compiling. `AgentHandle` is now derived from the definition instead of restated:

```ts
type AgentHandle =
  ReturnType<typeof AgentProcess> extends Process.Process<infer Input, infer Output, any, infer Rpcs>
    ? ProcessManager.Handle<Input, Output, Rpcs>
    : never;
```

so input, output and the RPC group all come from one place and can't drift again. `compute` 73 passed, `compute-runtime` 169 passed / 1 skipped, `agent-runtime` 30 passed / 21 skipped.

The encoded side stays `any` deliberately — nothing here depends on the wire type, and pinning it would mean threading a third parameter through `Process` for no consumer.

---

packages/core/compute/edge-compute/src/edge-client.ts
dmaretskyi: Honest answer: `configureEdgeClient` exists only because the automated review suggested factoring the duplication between `createEdgeClient` and `createEdgeProcessClient` (url lookup + invariant + `setIdentity`), and I took it. It's marginal — a generic taking a constructor callback in exchange for three duplicated lines. Say the word and I'll drop it and let the two factories each spell it out; I'd rather your file read the way you want than carry a bot's refactor.

The part that *isn't* marginal is why there are two factories at all, since that's the adjacent question. `composer-app:check-boot-budget` measured this branch 2,048 bytes over its ceiling, and 1,095 of those were the seven process routes I'd put on `EdgeHttpClient` — Composer instantiates that class during boot, so its methods ship whether or not anything calls them. They moved to `EdgeProcessHttpClient` behind `@dxos/edge-client/process`, which needs its own construction path; hence `createEdgeProcessClient`. That took the graph to 953 over, and the residual 953 turned out to be `main`'s, not this branch's (details in the PR comment above).

If you'd rather not have the subclass either, the alternative is raising `MAX_PRELOAD_BYTES` and putting the routes back on `EdgeHttpClient` where they'd naturally live — happy to do that instead, it's your budget to spend.

---

packages/core/compute/compute-runtime/src/RemoteProcessManager.ts
```diff
@@ -24,21 +29,82 @@ export type CancelTarget = {
...
+ * (the routes are `/processes/:spaceId/...`), but one manager serves all of a client's spaces, so a
+ * stack does not need a manager instance per space. Ids are strings to keep this interface free of
+ * `@dxos/keys` value imports.
+ *
+ * Effects carry no error channel, matching the local manager: transport and host failures surface as
+ * defects.
+ */
+export interface Control {
+  /** Spawn one of the host's built-in processes in `space`. */
+  spawn(space: string, request: ProcessProtocol.SpawnProcessRequest): Effect.Effect<ProcessProtocol.ProcessInfo>;
```
dmaretskyi:
```suggestion
  spawn(opts: { spaceId: SpaceId, .... }): Effect.Effect<ProcessProtocol.ProcessInfo>;
```

dont rely on imports from protocols.

do the rest likewise

packages/core/compute/compute-runtime/src/RemoteProcessManagerAdapter.ts
```diff
@@ -0,0 +1,267 @@
...
+ * This is a façade, NOT a stand-in for `ProcessManager.Service`: that tag means local
+ * execution, and binding a remote manager to it would report edge processes as local in the
+ * aggregate `ProcessMonitor` tree. Remote processes belong to {@link RemoteProcessManager.Service},
+ * whose `control` this adapts.
+ *
+ * A remote host cannot be handed a process definition, so {@link spawn} sends the definition's `key`
+ * and the host resolves it against the processes it hosts; a key the host does not know fails the
+ * spawn.
+ */
+export class RemoteProcessManagerAdapter implements ProcessManager.Manager {
```
dmaretskyi: kill this module -- consuming code should be aware of remote vs local process manager

dmaretskyi: unifying remote vs local is done a layer above -- for example in AgentService, or Operation.Service

dmaretskyi: Killed (`03797fea`), along with the pretence that a remote process can be driven through the local verbs — the same conflation D1a corrected, one level down.

Where its substance went:

- The verbs are `RemoteProcessManager.Manager`'s own — `spawn`, `list`, `attach`, `refreshProcessTree` — each taking the `spaceId` it addresses, so nothing is space-bound and one manager still serves every space. `RemoteProcessManager.makeControlVerbs(control, registry, processTreeAtom)` implements them once over a `Control`; `EdgeProcessManager` spreads it beside `control`, so a manager built without one lacks all of them and a caller that needs to spawn remotely fails where it asks rather than silently.
- Unifying the two is the consumer's job, as you said: `AgentService` has a `processesFor(location, spaceId)` returning either the local manager's verbs or the remote ones, and spawns through what it picked. It no longer holds a manager per space — only the set of spaces an edge session was opened on, which is all `hydrate` needs. `AgentService.layer` also stopped requiring `AtomRegistry`: the tree atom belongs to the remote manager.
- `Operation.Service` is the other layer you named; nothing there dispatches on local vs remote yet, so I left it alone rather than inventing a seam.

`RemoteProcessManagerAdapter.test.ts` → `RemoteProcessManagerVerbs.test.ts`, driving the remote manager instead of a `ProcessManager.Service` injection — that injection was itself the smell, binding a remote manager to the local tag inside the test.

Recorded as D1b in `.agents/projects/agent-process-edge/DESIGN.md`, Phase 7 in TASKS.md.

Verified: compute-runtime 205 passed (verbs suite 12/12), compute 76, agent-runtime 44 passed/22 skipped, plugin-routine 61, repo-wide `:test-types` and knip clean; edge suites 16/16 against the real worker — 9 process routes, 6 compute APIs, 1 agent-on-edge. Edge side `2beea65`.

One thing the edge run caught from the merge, unrelated to this thread: `ProcessContext.setAlarm` returns an `Effect` on current main, and `TestProcess` called it bare, so it armed nothing.

---

packages/core/compute/compute-runtime/src/RemoteProcessManager.ts
dmaretskyi: Done, and applied to the whole interface rather than just `spawn`. `RemoteProcessManager` no longer names a `ProcessProtocol` type anywhere: the wire shapes are the transport's business, so decoding them is the implementation's job and a consumer of this interface never sees them.

Every verb takes an options bag with the space in it:

```ts
spawn(request: SpawnRequest): Effect.Effect<Snapshot>;                 // { spaceId, key, name?, parentPid?, environment?, annotations? }
list(request: ListRequest): Effect.Effect<readonly Snapshot[]>;        // { spaceId, key?, target?, state? }
status(target: ProcessTarget): Effect.Effect<Snapshot>;                // { spaceId, pid }
submitInput(target: ProcessTarget & { input: unknown }): Effect.Effect<void>;
makeRpcClient(target: ProcessTarget & { group: RpcGroup<Rpcs> }): Effect.Effect<RpcClient<Rpcs>, never, Scope>;
terminate(target: ProcessTarget): Effect.Effect<void>;
readEvents(target: ProcessTarget & { cursor: number }): Effect.Effect<EventPage>;
```

The domain types it returns instead: `Snapshot extends Process.Info` with one added field (`alarmDueAt`, the one thing a remote caller cannot otherwise know), an `Event` union, and `EventPage { events, cursor, truncated, snapshot }`. `spaceId` is a branded `SpaceId`, `pid` a `Process.ID`.

The wire decoding moved to `@dxos/edge-compute/process-snapshot.ts` (`decodeSnapshot`, `decodeEvent`, `toSpawnRequest`) — next to the client that produces the bytes. `RemoteProcessHandle` dropped its own decode path with it, so a handle is a view over already-decoded domain values.

Since then this interface also gained the manager verbs the deleted adapter used to hold, in the same shape — `spawn({ spaceId, key, definition?, ... })`.

---

dmaretskyi:
CI status, and one failure that is not this PR's.

**`Check / check` is green** as of `b692e6fc`. It had been dying at ~100 s on `unused-imports`, which is fatal in that job's stage-2 `moon run :lint :build :test-types` — two batches: the `ProcessManager` type in the renamed `RemoteProcessManagerVerbs.test.ts`, then four imports in `EdgeProcessControl.ts` (`Process`, `invariant`, `SpaceId`, `ProcessProtocol`) left behind when the wire decoding moved to `process-snapshot.ts`. A repo-wide `moon exec :lint` sweep confirms those five were the only lint findings in the workspace.

**`Check / test` shard 0** is the one still red. Depot's logs aren't reachable through the GitHub API (the check run carries no output), so I reproduced the shard's exact command locally — `moon exec :test --job 0 --job-total 4 -- --no-file-parallelism` with the job's env — and the only failure in it is:

```
FAIL |node| src/spaces.test.ts > Spaces > post and listen to messages
Error: Timeout [200ms]
 ❯ src/spaces.test.ts:152:11
```

That is `client-e2e`, which depends on none of the packages this PR changes except `@dxos/protocols`, where the change is a types-only `ProcessProtocol` module. The test allowed **200 ms** for a two-peer post/listen round trip, under its own cost on a loaded runner; it passes in isolation locally, and every sibling deadline in that suite is 2 s.

I can't re-run a depot job, so rather than leave it as "probably a flake" I made the test robust within its own terms — `8341ef9d` raises that one deadline to 2 s, matching the suite, and changes nothing else. If you'd rather this not ride along with the feature work, say so and I'll drop it and report the failure instead.

Everything else on the current head: `memory`, `changeset-reminder`, `model-fixture`, `boot-budget` green; repo-wide `:build` and `:test-types` clean; knip clean; compute-runtime 205, compute 76, agent-runtime 44 (+22 skipped), plugin-routine 61, plugin-assistant green. Edge side ([dxos/edge#971](https://github.com/dxos/edge/pull/971)) 16/16 against the real worker.

---

dmaretskyi:
Status on the two checks still red, and where I'm stuck.

**Fixed and verified since the last update** — three under-provisioned test budgets, each aligned with the convention in its own file, all in packages this PR doesn't touch:

| Test | Was | Now |
|---|---|---|
| `client-e2e spaces.test.ts` — two-peer post/listen round trip | 200 ms → 2 s → still failed | 30 s wait / 60 s test (`50a62504`) |
| `plugin-thread plugin.test.ts` — Composer app boot, 12.3 s unloaded | 15 s default | 60 s (`c511bfa1`) |
| `echo-host` two-phase subduction policy — 5 s poll + 10 s wait | 15 s default | 30 s (`6fbbf928`) |

I got the `client-e2e` value wrong the first time: 2 s was still under the round trip's cost on a loaded shard. It is now a hang guard rather than a latency assertion, which is what `vite.base.config.ts` already says about the default ("too tight for harness-based tests … causing flaky timeouts in a different plugin each run").

**`Check / check`** passed on `8341ef9d` and has failed on `6fbbf928` and `50a62504`, both times *past* the lint/build gate it was failing at earlier (that was `unused-imports`, fixed in `80e131af` + `b692e6fc`). The only commits in between are the two test-timeout one-liners above.

**What I've run locally, all clean:** `oxfmt --check` repo-wide; every stage-1 gate (`check-cycles`, `check-publish-config`, `check-changesets`, `sync-versions --check`, `check-cache-wiring`, `pkg-lint`, `check-public-dependencies`, `check-packages-published`, `toolbox --lint-package-exports` with no resulting diff); `moon exec :lint` repo-wide (only findings were the five imports, already fixed); `:test-types` repo-wide; `knip`; `:check-module-structure`; and `echo-host:build plugin-thread:build client-e2e:build`. The one module-structure error I hit (`plugin-space` not finding declarations for `@dxos/plugin-preview/testing`) was a stale local artifact — `plugin-preview/dist/types/` held only `tsbuildinfo`, and a forced rebuild emits `testing.d.ts` correctly.

**Where I'm stuck:** the checks run on depot, and neither the GitHub API's check-run output nor `get_job_logs` returns anything for them (`output.text` is empty; the jobs aren't GitHub Actions jobs), so I can't see which step or test fails. I also have no way to re-run a depot job. I've been identifying the shard failures by reproducing the shard command locally — including the full four-task form (`:test :test-browser :test-storybook :test-workerd --job 0 --job-total 4`), which is how I found that 2 s was still too short — but that can't tell me what `check` itself is failing on.

If you can paste the failing step from either job (or point me at where depot logs are readable), I'll finish it. Absent that I'll keep reproducing shard by shard, which is slow and can't cover `check`.

Unrelated and not chased: the whole-file `client-e2e` run once failed `getEpochs` with `Space does not have an automerge root` — a non-timeout failure that did **not** occur in the shard run, so I've left it rather than confuse two signals.

---

packages/core/mesh/edge-client/src/edge-process-http-client.ts
```diff
@@ -0,0 +1,117 @@
...
+ * Process-control routes, as a subclass rather than methods on {@link EdgeHttpClient}.
+ *
+ * Composer instantiates `EdgeHttpClient` during boot, so every method on it ships in the eager boot
+ * graph whether or not the app calls it — and that graph is budgeted
+ * (`composer-app:check-boot-budget`). Process control is only reached from the EDGE process manager,
+ * so it lives here, off the package barrel behind the `@dxos/edge-client/process` subpath, and costs
+ * boot nothing.
+ */
+export class EdgeProcessHttpClient extends EdgeHttpClient {
+  /**
```
dmaretskyi: no, put it on the original class

dmaretskyi: Done — `805995f7`. The seven routes are on `EdgeHttpClient` itself; `EdgeProcessHttpClient`, the `@dxos/edge-client/process` subpath and the duplicate `createEdgeProcessClient` builder are all gone (that builder had become a copy of `createEdgeClient`, so `configureEdgeClient` folded back into it).

For the record on why they were split out, since it wasn't arbitrary: they cost ~1,095 boot bytes, and when I first added them `composer-app:check-boot-budget` was **already 953 bytes over** its ceiling on `main` — an unrelated PR failed it identically. `Check / boot-budget` has been green on my recent heads, so main's shortfall has since been resolved and there is margin again. I'm letting CI's `boot-budget` job be the arbiter rather than my sandbox: my local run of it kept tripping over stale artifacts in this container (`nlp` couldn't see `@dxos/ai`'s declarations), which measures my environment rather than the change. If that check goes red on this head, the number is the answer and I'll say so rather than quietly re-splitting.

`edge-client` and `edge-compute` build and lint clean. One thing the move needed: an explicit `import { type EdgeHttpClient } from '@dxos/edge-client'` in `EdgeProcessControl` — without it the parameter type silently degraded and six call sites came back as `unknown`.

The boot-budget note in `.agents/projects/agent-process-edge/TASKS.md` used to say "put new client surface on the subclass"; it now records this decision instead, so nobody reinstates the split.

---

dmaretskyi:
Green and enqueued on `33556723`. Correcting my last two comments, which had this wrong.

**The `test` shard failures were a real bug in this PR, not flakes.** Reproducing shard 0 locally surfaced a hard error rather than a timeout:

```
ServiceNotAvailable: Service not available: @dxos/functions-runtime/AgentService
  — application provider spec pruned due to missing deps: @dxos/compute-runtime/RemoteProcessManager
```

I had declared `RemoteProcessManager.Service` a build-time requirement of both `AgentService.layer` and the plugin-assistant `LayerSpec`. `LayerStack` prunes per spec, so on any stack that hosts only local agents — the Composer test app, `plugin-projects` — the entire `AgentService` provider was dropped and every agent call failed. Remote is needed only for `location: 'edge'`.

Fixed in `33556723`: the manager arrives through `AgentServiceOptions.getRemoteManager`, read from the slice with `Effect.serviceOption` rather than required. My first attempt routed it through `ServiceResolver` and merely moved the pruning to `@dxos/functions/ServiceResolver`, so that was dropped. `plugin-projects` `delegate-task-to-chat` went 3 failed → 4 passed; `agent-runtime` 44, `compute-runtime` 207, `edge-compute` 5, `plugin-assistant` 189 green.

Two things I asserted earlier that were wrong, for the record:

1. I reported the `client-e2e` `post and listen to messages` failure as a latency budget, then as a regression this PR introduced, on the strength of a control run. **Both readings were wrong.** The comparison was confounded — my branch's run was competing with a four-way shard sweep, the control run had an idle machine. On an idle machine that test passes on this branch in 1.33 s. The three budget commits stand on their own merits (200 ms was under the round trip's real cost), but they were not the shard fix.
2. A `git bisect` I ran over the 50 commits used that same load-contaminated test as its predicate, so its result is meaningless and I am not reporting the commit it named.

Also on this head: `@dxos/edge-client`'s `./process` export entry was still pointing at the source file deleted when the split client was folded back into `EdgeHttpClient` — removed in `614af044`. `Check / boot-budget` is green for that move, so putting the seven routes back on `EdgeHttpClient` costs nothing against the ceiling.

---

PR #12768, opened 2026-08-26
plugin-claude-agents: authenticate managed agents with Anthropic OAuth

dmaretskyi:
Lets the Claude managed-agents plugin run agents on a token minted by connecting the user's Claude account, instead of requiring a Console API key pasted into the space.

## Changes

- `@dxos/protocols`: new `OAuthProvider.ANTHROPIC` (`'anthropic'`).
- `plugin-claude-agents`:
  - New authentication-only connector (`src/capabilities/connector.ts`) contributing `ConnectorSpec.Connector` with the Anthropic OAuth spec (`user:inference`, `user:profile`) and a `testConnection` probe (`GET /v1/models?limit=1`), so an expired or revoked grant surfaces as a reauthenticate prompt rather than a mid-session failure. No sync targets.
  - `credentials.ts`: `getApiKey` → `getCredential`, returning `{ token, scheme }`. The scheme is derived from the `sk-ant-oat…` prefix, which is the only discriminator the credential store exposes between an OAuth access token and an `sk-ant-api…` key.
  - `api/client.ts`: an OAuth token goes on `Authorization: Bearer` with `anthropic-beta: managed-agents-2026-04-01,oauth-2025-04-20`; an API key keeps `x-api-key`. All control-plane calls take the credential rather than a bare key.
  - `PLUGIN.mdl` / skill instructions updated for the two credential kinds.

The EDGE half of the flow (provider spec, PKCE, token custody) is dxos/edge#972 — it needs this change published before its `@dxos/protocols` pin can carry `OAuthProvider.ANTHROPIC`, so this lands first.

## Configuration still required

No DXOS OAuth client is registered with Anthropic yet. Until `ANTHROPIC_CLIENT_ID` is configured on EDGE, connecting the Anthropic connector fails at the provider; an API-key connection keeps working exactly as before.

## Testing

- `moon run plugin-claude-agents:build` (plus the 171 upstream builds) — clean.
- `moon run plugin-claude-agents:test` — 14 tests pass.
- `moon run plugin-claude-agents:lint protocols:lint`, `pnpm format` — clean.
- The OAuth flow itself is not exercised end to end here: it needs a registered client (see above).

https://claude.ai/code/session_01EYQjzhJnhXfFaHSx38ynAf

### DIFF
--- a/.changeset/anthropic-oauth-provider.md
+++ b/.changeset/anthropic-oauth-provider.md
@@ -1,0 +1,6 @@
+---
+'@dxos/protocols': minor
+---
+
+Add the `anthropic` OAuth provider, so a Claude account can be connected over OAuth and the
+managed-agent operations run against the resulting access token instead of a pasted Console API key.
--- a/packages/core/protocols/src/edge/edge.ts
+++ b/packages/core/protocols/src/edge/edge.ts
@@ -334,6 +334,7 @@ export type EdgeAuthChallenge = {
 };
 
 export enum OAuthProvider {
+  ANTHROPIC = 'anthropic',
   ATLASSIAN = 'atlassian',
   ATPROTO = 'atproto',
   /** @deprecated Use ATPROTO instead. */
--- a/packages/plugins/plugin-claude-agents/PLUGIN.mdl
+++ b/packages/plugins/plugin-claude-agents/PLUGIN.mdl
@@ -162,8 +162,18 @@ feat F-6: Credentials
 
   req F-6.1:
     when: any operation calls the Anthropic API
-    then: the API key is resolved from the credential stored under the
-          "anthropic.com" service and sent as the `x-api-key` header
+    then: |
+      the credential is resolved from the "anthropic.com" service and sent as
+      the header its kind requires — an OAuth access token (`sk-ant-oat…`, minted
+      by connecting a Claude account) on `Authorization: Bearer` alongside the
+      `oauth-2025-04-20` beta, a Console API key on `x-api-key`
+
+  req F-6.4:
+    when: the user connects the "anthropic" connector
+    then: |
+      EDGE runs the Anthropic authorization-code flow (PKCE) and the resulting
+      access token is stored as the space's "anthropic.com" credential, so no key
+      is ever pasted into the conversation
 
   req F-6.3:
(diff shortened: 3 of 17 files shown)


dmaretskyi:
CI status on this head, so it isn't mistaken for a failure in the diff:

`PR Build` [run 746](https://github.com/dxos/dxos/actions/runs/32984721685) concluded `startup_failure` 14 seconds after it was created — it never reached a job, so nothing in this branch was compiled or tested. The same thing happened to [run 748](https://github.com/dxos/dxos/actions/runs/32985667735) on an unrelated branch in the same window, and run 744 (PR #12767) is still stuck queued from it, while runs from 16:13 onward succeed normally. It's a GitHub-side startup condition in that window, not this PR's.

I re-ran it once (16:38); the re-run has sat queued without picking up a runner. That's my one re-run spent, and I won't push an empty commit to kick it — the next real push to this branch will produce a healthy run, as it did for run 749 on that other branch.

Locally on this head: `moon run plugin-claude-agents:build` (with its 171 upstream builds) is clean, `plugin-claude-agents:test` passes 14/14, and lint plus `pnpm format` are clean.

---

dmaretskyi:
Closing unmerged: Anthropic does not permit this.

Per [legal and compliance](https://code.claude.com/docs/en/legal-and-compliance) § *Authentication and credential use*, OAuth "is intended exclusively for purchasers of Claude Free, Pro, Max, Team, and Enterprise subscription plans … to support ordinary use of Claude Code and other native Anthropic applications", and third-party developers may not "offer Claude.ai login into their own applications" or "collect, store, or intermediate Claude.ai credentials or session tokens". This PR does all three: Composer would offer the login, and EDGE would custody the refresh token. The [supported methods](https://platform.claude.com/docs/en/manage-claude/authentication) are API key, Workload Identity Federation, and App Attest — there is no third-party OAuth client to register.

The API-key path stays as it was before this branch (the built-in Custom Token connector under `anthropic.com`), so nothing regresses by closing this.

Worth salvaging separately if wanted: the credential-kind split in `credentials.ts` + `api/client.ts` (bearer vs `x-api-key`) is what a Workload Identity Federation token would need, and is independent of the OAuth flow.

EDGE counterpart closed likewise: dxos/edge#972.

---

PR #12773, opened 2026-08-26
plugin-support: attach screenshot on every feedback submit path

dmaretskyi:
Closes DX-1193.

The feedback panel already had an `image` toggle, but it was hidden unless `enableGitHubIssues` was set and only the GitHub path honoured it. It is now a first-class option next to "Include debug logs" and applies to all three submit paths.

## Changes

- New `useScreenshotAttachment` hook (`containers/FeedbackPanel/`) — collapses the help companion, waits for the deck animation, then captures + uploads via the existing `screenshot.ts` helpers. Best-effort: returns `{ failed: true }` rather than throwing, so a failed attachment never blocks the report.
- `FeedbackSubmitAction` (PostHog) and `DiscordAction` now attach the screenshot; `formatRequestMessage` embeds it as `[Screenshot](url)` above the body so it survives the single-string message channel.
- `GitHubAction` refactored onto the shared hook (behaviour unchanged).
- `FeedbackPanel` no longer filters the `image` field out; schema title/description reworded to drop "GitHub only".
- Added `feedback-toast-no-screenshot.description` for the capture-failed case, mirroring the GitHub toast.
- Unit test for the screenshot embedding in `request.test.ts`; changeset added.

## Verification

```
moon run plugin-support:build   # ok
moon run plugin-support:lint    # ok
moon run plugin-support:test    # 2 files, 9 tests passed
pnpm format
```

https://claude.ai/code/session_01Uyy888pgBGXTnQqseg4Kp4

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uyy888pgBGXTnQqseg4Kp4)_

### DIFF
--- a/.changeset/feedback-screenshot-toggle.md
+++ b/.changeset/feedback-screenshot-toggle.md
@@ -1,0 +1,5 @@
+---
+'@dxos/plugin-support': patch
+---
+
+Feedback panel: the "Attach screenshot" toggle is now always available (alongside "Include debug logs") and applies to every submit path — PostHog feedback and Discord help threads embed the uploaded screenshot in the report, not just GitHub issues.
--- a/packages/plugins/plugin-support/src/containers/FeedbackPanel/DiscordAction.tsx
+++ b/packages/plugins/plugin-support/src/containers/FeedbackPanel/DiscordAction.tsx
@@ -16,6 +16,7 @@ import { meta } from '#meta';
 import { SupportOperation } from '#types';
 
 import { formatRequestMessage } from './request';
+import { useScreenshotAttachment } from './useScreenshotAttachment';
 
 /** Build a direct PostHog event permalink (±15s search window via timestamp). */
 const makePostHogEventUrl = (projectId: string, eventUuid: string): string =>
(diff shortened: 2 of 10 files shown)


dmaretskyi:
`boot-budget` (`composer-app:check-boot-budget`) is red, and it is not this PR's:

```
```

Measured locally on this branch and again with `packages/plugins/plugin-support` reverted to the merge base (`git checkout HEAD~1 -- packages/plugins/plugin-support`, new files removed), rebundling `composer-app` each time:

| tree | preload entries | boot bytes | over budget |
|---|---|---|---|
| this PR (`00819b3`) | 22 | 4,457,401 | +953 B |
| merge base, plugin-support reverted | 22 | 4,457,401 | +953 B |

Byte-identical — plugin-support is not in the eager boot graph, so this change contributes nothing to it. The base is already 953 bytes over the 4.25 MB ceiling; the check's own message rounds both sides to `4.25 MB`, which is why it reads as a self-comparison. Deterministic, so a re-run would not clear it.

No fix for it exists to port, and raising the ceiling is exactly the review point the script asks for ("Raise a budget only for growth you have looked at and accepted"), so I am not widening this PR with it. The minimal patch, if the owner accepts the growth:

```diff
--- a/packages/apps/composer-app/scripts/check-boot-budget.mjs
+++ b/packages/apps/composer-app/scripts/check-boot-budget.mjs
-const MAX_PRELOAD_BYTES = 4.25 * 1024 * 1024;
+const MAX_PRELOAD_BYTES = 4.3 * 1024 * 1024;
```

The better fix is to find the ~1 KB (or the accumulated growth behind it) and cut it, rather than re-baseline for a 953-byte overshoot. Two secondary notes:

1. The error line should print more precision (e.g. KB, or 3 decimals) — `4.25 MB exceeds 4.25 MB` is unreadable at the margin.
2. Every other check on this head is green: `check`, `build`, all four `test` shards, `memory`, `model-fixture`, `changeset-reminder`.


---

PR #12778, opened 2026-08-26
ci: post a PR comment when the boot graph grows against the base

wittjosiah:
## Why

The boot budget only speaks when the ceiling is crossed, so growth underneath it accumulates unseen — the 2026-08-26 re-baseline in `check-boot-budget.mjs` absorbed ~400 KB spread over many changes that each passed on their own.

## What

On a same-repo `pull_request`, the `boot-budget` job now also measures the **base** commit, and a separate `boot-budget-comment` job posts a sticky `<!-- boot-budget -->` comment when the eager boot graph grew. It is advisory — the budget in `check-boot-budget.mjs` remains the only thing that gates.

- **`check-boot-budget.mjs`** — every run now writes its measurement to `out/boot-budget.json` (total bytes, preload count, budget, per-chunk sizes), before the budget is applied, so an over-budget commit still yields a number. It deliberately takes no arguments.
- **`moon.yml`** — `check-boot-budget` declares `out/boot-budget.json` as its `outputs`, which is what makes the base measurement cheap: main's own run of this job leaves the JSON in the remote cache, so measuring the base restores a few KB rather than rebuilding its bundle. Passthrough args would fold into moon's hash and lose that hit, hence the no-arguments rule above.
- **`compare-boot-budget.mjs`** (new) — diffs two reports and emits `status` (`grew` / `unchanged`) plus `body` to `$GITHUB_OUTPUT`. The body is a base-vs-head table (bytes, preload entries, Δ, budget), remaining headroom, and a collapsed table of the chunks that moved. Chunk content hashes are stripped so the two sides pair up.
- **`.depot/workflows/check.yml`** — `boot-budget` measures head and base and compares, at `contents: read` / `packages: read`. `boot-budget-comment` holds the only `pull-requests: write` grant and posts/updates/deletes the sticky comment. "Check boot budget" is `continue-on-error` with a final step that re-fails the job, so a PR that blows the budget still gets the comment saying by how much and where.

## Notes on the shape

1. **The base commit is measured, not trended from a stored baseline** — the comparison is exact for this pair of commits and depends on no artifact retention or cache eviction, while costing a cache restore rather than a build.
2. **The write token never sits in a job that runs branch code.** A job-level `permissions` block applies to every step, so leaving `pull-requests: write` beside `moon run` and `./.depot/actions/setup` would put it within reach of the code being measured — `persist-credentials: false` keeps the token out of `.git/config` but not out of a local action's expression context. `boot-budget-comment` has no checkout, no local action and no build; it receives two strings as job outputs.
3. **The report is deleted before the base is measured.** If moon ever stops restoring it, the copy fails loudly instead of measuring the head's leftover file as though it were the base — the one failure mode that would silently report "no change" on exactly the PRs that move the graph.
4. **A PR that edits `check-boot-budget.mjs` gets no comparison**, since the base's number necessarily came from the base's own copy of it. `status` defaults to `skipped` at the job boundary so "no comparison ran" is never read as "no growth".
5. **Failures are contained** — every advisory step is `continue-on-error`, and the gate reports under `!cancelled()` so no advisory failure can swallow a real budget failure. Fork PRs are excluded because their token cannot comment.
6. **Byte deltas under 1 KB are not reported** (minifier jitter); any change in the preload entry count is reported regardless, since that axis only moves when the chunk partition reshapes.

## What is *not* demonstrated here, and why it can't be

The base cache-restore path does not run on this PR, for two independent structural reasons:

1. This PR edits `check-boot-budget.mjs`, so note 4 above skips the comparison by design.
2. Even without that, `main` predates the `outputs` declaration — the base's `check-boot-budget` writes no report, so the copy would fail and the comparison would skip anyway.

It first exercises on the next PR opened after this lands. What that looks like: the "Measure the base commit's boot graph" step completing in seconds. For reference, an earlier revision of this branch that rebuilt the base bundle outright took 2m53s in that step (commit `5dacf79a`), which is the cost this design removes.

## Testing

`compare-boot-budget.mjs` exercised against synthetic reports for the grew / unchanged / `$GITHUB_OUTPUT` paths; `check-boot-budget.mjs` against a fabricated `out/composer` for under-budget (exit 0) and over-budget (exit 1, unchanged message), with the report written to `out/` and complete in both. `oxfmt --check` clean; `check.yml` and `moon.yml` parse; `node --check` on both scripts. `moon` is unavailable in the authoring sandbox, so everything moon-side rests on CI.

Also verified end to end on an earlier revision of this branch, before the caching change: head measured, base measured, compare run, and the comment step correctly posting nothing when the graph had not grown.

## Known follow-up, not in scope

`changeset-reminder` matches its sticky comment on the marker alone and holds `pull-requests: write` in a job that runs repo scripts — the same two shapes fixed here. Left alone rather than widening this PR.

No changeset — CI tooling, nothing consumer-relevant ships.


wittjosiah:
Dispositioning the two risks CodeRabbit's walkthrough flagged, since it asked for explicit maintainer awareness.

**1. Marker-only comment ownership — fixed** in `f6162168`. The lookup matched any comment containing `<!-- boot-budget -->`, so a person quoting the marker in the thread could have had their comment edited or deleted. It now requires `comment.user?.type === 'Bot'` as well.

Note the `changeset-reminder` job has the same marker-only shape and the same latent issue. Left alone here rather than widening this PR — worth a follow-up.

**2. `pull-requests: write` alongside branch code — accepted, with the mitigations already in place.** Two things bound it:

- The job runs on `pull_request` and is gated on `head.repo.full_name == github.repository`, so only same-repo branches reach it. Anyone who can push a branch here already has write access and can comment directly, so the grant adds no capability an attacker in that position lacks. Fork PRs never run the comparison steps.
- The token is not reachable from the build steps. `Checkout` sets `persist-credentials: false`, so no credential lands in `.git/config`, and no `run:` step in the job puts `GITHUB_TOKEN` in its environment — it is handed only to the `actions/github-script` step, which executes no branch code.

The structurally tighter option is to split measurement (`contents: read`) from commenting (`pull-requests: write`), passing the two reports between jobs as artifacts. Happy to do that if a reviewer would rather have the hard separation than the two bounds above.

---

wittjosiah:
Both findings from the last two reviews are now fixed.

**Advisory failures could redden or silence the gate** (`2c69cabf`). `Compare with the base` and `Post or remove boot graph comment` could fail ahead of `Report boot budget failure`, which both failed jobs whose budget passed and skipped the step that reports a budget failure when one had occurred. Both are `continue-on-error` now, and the gate runs under `!cancelled()` so nothing above it can swallow the result. Correct finding — it broke the containment this PR's description claims.

**Write token reachable from PR-controlled code** (`e9bef91a`). I was wrong to close this out with `persist-credentials: false` earlier. That keeps the token out of `.git/config`, but a job-level `permissions` block applies to every step, and `./.depot/actions/setup` is a local action the branch under test controls — an expression there can reference `github.token` regardless. The git-credential vector was covered; the action-context one was not.

Split accordingly:

- `boot-budget` drops to `contents: read` / `packages: read` and exposes `increased` / `body` as job outputs.
- `boot-budget-comment` is a new job holding the only `pull-requests: write` grant. It has no checkout, no local action and no build — it sees the two strings the comparison produced and nothing else.
- It runs under `!cancelled()` so a failed budget still gets its comment, and is gated on `increased != ''` so a comparison that never ran (fork PR, unmeasurable base, measurement script edited in the PR) isn't mistaken for one that found no growth.

Still outstanding and not mine to fix here: `changeset-reminder` has the same marker-only comment ownership shape that was fixed in `f6162168`, and the same `pull-requests: write` question, though it runs far less branch code. Worth a follow-up.

---

wittjosiah:
Fixed in `caf18a3a` — correct, and a regression I introduced. The comment step carried `continue-on-error: true` before `e9bef91a`; moving it into `boot-budget-comment` dropped it, so a GitHub API failure while posting an advisory comment would have failed the check. Restored at the step, not the job, so the failure stays visible in the logs without the job's check run going red and blocking a merge.

On the standing **Merge Risk: High** note about `pull-requests: write` — that was addressed in `e9bef91a`, which the risk summary predates. The measuring job is `contents: read` / `packages: read` now, and the write grant lives in `boot-budget-comment`, which has no checkout, no local action and no build.

---

PR #12790, opened 2026-08-27
chore: dedupe the setup composite action

wittjosiah:
## Summary

`.github/actions/setup` and `.depot/actions/setup` were byte-identical duplicates. There's no reason for that — `uses: ./path` resolves any local action from the checkout regardless of which orchestrator (Depot CI or GitHub Actions) runs the job.

Verified live: `check.yml`'s `Setup` step ran clean on a real Depot CI run pointed at `./.github/actions/setup` before this change ([run](https://depot.dev/orgs/t8fblrl00n/workflows/8jv8vq0606)).

Keeps `.depot/actions/setup` as the single canonical copy — deletes `.github/actions/setup`, repoints `publish-all.yml` and `deploy-apps.yml` (the two GitHub-Actions-only workflows that used it) at `.depot/actions/setup`. Every `.depot/workflows/*.yml` file already pointed there.

Note: `pkg-pr-new.yml` and `pr-build.yml` are left untouched here — they're mid-revert back to GitHub Actions on #12788 (not yet merged), so touching their `setup` reference here would just be redone there.


wittjosiah:
Folded into #12788 — it deleted `.github/actions/setup` while #12788 restores workflows that reference it, so landing them separately would break `main` in whichever order came second. One PR for the whole cleanup.

PR #12792, opened 2026-08-27
echo: add Annotation.SetParent for owned ref fields

dmaretskyi:
## Summary

New `Annotation.SetParent` marks a `Ref` field as **owning** its targets. Writing a ref into such a field — or creating the holder with one — sets the target's ECHO parent to the holder, so the child cascade-deletes and deep-clones with it. Ownership becomes a property of the schema instead of a rule every call site has to remember.

```ts
Schema.Struct({
  content: Ref.Ref(Text.Text).pipe(Annotation.SetParent.set(true)),
  sections: Schema.Array(Ref.Ref(Section)).pipe(Annotation.SetParent.set(true)),
})
```

## Implementation

- `SetParentAnnotation` (`org.dxos.annotation.setParent`) — a `PropertyMeta`-backed boolean annotation, exported as `Annotation.SetParent`.
- `internal/Obj/parent-annotation.ts` — resolves a schema's owning-field paths and writes `[ParentId]` on each resolved ref target. Handles single refs, arrays of refs, fields nested in plain structs (`Channel.backend.config`), and members of a discriminated union field (`Routine.spec.instructions`). Paths are memoized per AST node, so a recursive (`Schema.suspend`) or repeated nested type stays linear rather than exploding combinatorially.
- Hooked into `Obj.make` (after construction) and `Obj.update` (after the change transaction); idempotent, and skips a ref with no resolvable target.

Two deliberate limits, documented on the annotation: removing a ref does **not** clear the target's parent (a two-step move between holders would lose the edge), and only object-kind targets are parented.

## Call-site sweep

Types that previously paired each ref write with `Obj.setParent` now declare it on the field: `Instructions.text`, `Outline.content`, `Project.{instructions,outline,taskSet,routines}`, `Chat.feed`, `Agent.instructions`, `Routine.{spec.instructions,triggers}`, `File.data`, `Channel.backend.config`, `Document.content`, `Spec.content`, `SourceFile.content`, `Mailbox.{feed,annotations,tags}`, `Calendar.{feed,tags}`, `Search.{feed,tags}`, `Subscription.{feed,contentFeed,postState,tags}`, `Magazine.{instructions,postState}`, `Scene.objects`, `Terra.objects`, `Artifact.variants`, `Portfolio.feed`, `ChessComAccount.games`, `PlayerReview.positionIndex`, `Blog.{outline,content,instructions}`. This also fixes sites that wrote the ref *without* setting a parent (`Chat.ensureTaskSet`, `ensureTaskSetSync`).

Left as explicit `Obj.setParent` calls, because the field ref is not what implies the parent:

- Edges whose child may be parented elsewhere — `TaskSet.tasks` (a sub-task's parent is its parent task), trip segments/bookings.
- Reverse edges (the child holds the ref) — `Chat.linkCompanion`, CRM profile relations.
- Ref-in-annotation edges — `Chat.CompanionChatAnnotation`.
- Generic containers (`Collection.objects`), where parenting every member would change semantics repo-wide.
- Ref-less parent edges (IBKR triggers/operations).

## Test plan

- New `Annotation.SetParent` cases in `packages/core/echo/echo/src/Annotation.test.ts` (creation, write, re-parent, array append, un-annotated field stays unparented) — `echo:test` 577 passed.
- New `Annotation.SetParent` describe block in `packages/core/echo/echo-client-e2e/src/parent.test.ts`, including a database write and a cascade-delete case.
- **Verification suite `echo-client-e2e`: 324 passed, 1 expected fail, 11 skipped.**
- Also green: `schema`, `types`, `compute`, `assistant-toolkit`, and the touched plugin suites (`plugin-inbox`, `-magazine`, `-routine`, `-studio`, `-projects`, `-blogger`, `-chess`, `-terra`, `-commerce`, `-spacetime`, `-assistant`, `-code`, `-markdown`, `-tasks`, `-connector`, `-space`, `-trip`).
- Full-repo `:build` clean; `oxfmt` applied; lint clean on every touched package.

Changeset: `.changeset/annotation-set-parent.md` (`@dxos/echo` + `@dxos/plugin-markdown`, minor).

---
_Generated by [Claude Code](https://claude.ai/code/session_01MTLVDNoJ1HN3bQn3w42zJy)_

## Summary by CodeRabbit

* **New Features**
  * Added declarative ownership for referenced content, enabling automatic parent assignment, deep cloning, and cascade deletion across supported records.
  * Ownership now works for nested references, arrays, and union-based fields.
* **Documentation**
  * Added guidance and examples for configuring and using owned references, including limitations and unresolved references.
* **Tests**
  * Added coverage for ownership during creation, updates, replacement, array changes, and deletion.
* **Style**
  * Added a review rule discouraging trivial wrappers around official APIs.

### DIFF
--- a/.agents/skills/echo/SKILL.md
+++ b/.agents/skills/echo/SKILL.md
@@ -161,6 +161,26 @@ export const make = (props: Obj.MakeProps<typeof Person>): Person => Obj.make(Pe
 export const make = (props: Obj.MakeProps<typeof Person>): Type.InstanceType<typeof Person> => Obj.make(Person, props);
 ```
 
+## Owned children — `Annotation.SetParent`
+
+Declare ownership on the ref field rather than calling `Obj.setParent` next to every write. Writing a
+ref into an annotated field (or creating the holder with one) sets the target's parent, so the child
+cascade-deletes and deep-clones with its holder.
+
+```ts
+Schema.Struct({
+  // Single ref, array of refs, a field nested in a struct, and a member of a union field all work.
+  content: Ref.Ref(Text.Text).pipe(Annotation.SetParent.set(true)),
+  sections: Schema.Array(Ref.Ref(Section)).pipe(Annotation.SetParent.set(true)),
+});
+```
+
+Do NOT annotate a field whose targets may be parented elsewhere (e.g. `TaskSet.tasks`, where a
+sub-task's parent is its parent task) — every write to the holder would re-parent them to it.
+Removing a ref does not clear the target's parent; call `Obj.setParent(child, undefined)` for that.
+Reverse edges (child holds the ref) and ref-in-annotation edges (e.g. `Chat.CompanionChatAnnotation`)
+still need `Obj.setParent`.
+
 ## Related docs in-repo
 
 - Effect runtime patterns: [.cursor/skills/effect/SKILL.md](../effect/SKILL.md).
--- a/.agents/skills/echo/SKILL.md
+++ b/.agents/skills/echo/SKILL.md
@@ -164,8 +164,9 @@ export const make = (props: Obj.MakeProps<typeof Person>): Type.InstanceType<typ
 ## Owned children — `Annotation.SetParent`
 
 Declare ownership on the ref field rather than calling `Obj.setParent` next to every write. Writing a
-ref into an annotated field (or creating the holder with one) sets the target's parent, so the child
-cascade-deletes and deep-clones with its holder.
+ref into an annotated field (or creating the holder with one) sets an object-kind target's parent, so
+that child cascade-deletes and deep-clones with its holder. A ref to anything else (a relation) is
+left alone — only objects can have a parent.
 
 ```ts
 Schema.Struct({
--- a/.agents/skills/echo/SKILL.md
+++ b/.agents/skills/echo/SKILL.md
@@ -177,7 +177,10 @@ Schema.Struct({
 
 Do NOT annotate a field whose targets may be parented elsewhere (e.g. `TaskSet.tasks`, where a
 sub-task's parent is its parent task) — every write to the holder would re-parent them to it.
-Removing a ref does not clear the target's parent; call `Obj.setParent(child, undefined)` for that.
+The annotation updates the parent on write; it is not an invariant that the target's parent IS the
+holder — `Obj.setParent` can re-parent it afterwards, and an unresolved ref is skipped. Read the
+parent with `Obj.getParent`, never from the field. Removing a ref does not clear the target's
+parent; call `Obj.setParent(child, undefined)` for that.
 Reverse edges (child holds the ref) and ref-in-annotation edges (e.g. `Chat.CompanionChatAnnotation`)
 still need `Obj.setParent`.
 
--- a/packages/core/echo/echo-client-e2e/src/parent.test.ts
+++ b/packages/core/echo/echo-client-e2e/src/parent.test.ts
@@ -210,9 +210,9 @@ describe('Annotation.SetParent', () => {
       linked: Ref.make(linked),
     });
 
-    expect(Obj.getParent(body)).toBe(container);
-    expect(Obj.getParent(section)).toBe(container);
-    expect(Obj.getParent(config)).toBe(container);
+    expect(Obj.getParent(body)?.id).toBe(container.id);
+    expect(Obj.getParent(section)?.id).toBe(container.id);
+    expect(Obj.getParent(config)?.id).toBe(container.id);
     expect(Obj.getParent(linked)).toBeUndefined();
   });
 
@@ -227,8 +227,8 @@ describe('Annotation.SetParent', () => {
       container.sections = [Ref.make(section)];
     });
 
-    expect(Obj.getParent(body)).toBe(container);
-    expect(Obj.getParent(section)).toBe(container);
+    expect(Obj.getParent(body)?.id).toBe(container.id);
+    expect(Obj.getParent(section)?.id).toBe(container.id);
   });
 
   test('parent is set on write to a database object', async () => {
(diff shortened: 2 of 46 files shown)


packages/core/echo/echo-client-e2e/src/parent.test.ts
```diff
@@ -168,3 +169,108 @@ describe('Parent Hierarchy', () => {
...
+      /** Owned, ordered. */
+      sections: Schema.Array(Ref.Ref(Body)).pipe(Annotation.SetParent.set(true)),
+      /** Owned, nested inside a plain struct. */
+      backend: Schema.optional(Schema.Struct({ config: Ref.Ref(Body).pipe(Annotation.SetParent.set(true)) })),
+      /** NOT owned — a plain reference. */
+      linked: Schema.optional(Ref.Ref(Body)),
+    }),
+  ) {}
+
+  const makeBody = (text: string) => Obj.make(Body, { text });
```
dmaretskyi: make an .mdl counterrule to not extract functions for trivial snippets that hide official APIS -- those tests would look better if this was inlined. those in-test custom dsls make it much more difficult to read tests

packages/core/echo/echo/src/internal/Annotation/annotations.ts
```diff
@@ -556,6 +556,25 @@ export const IconFromRefAnnotation = makeUserAnnotation<string>({
...
+ * lose the edge depending on write order; call `Obj.setParent(child, undefined)` explicitly.
+ *
+ * @example
+ * ```ts
+ * Schema.Struct({
+ *   body: Ref.Ref(Text.Text).pipe(Annotation.SetParent.set(true)),
+ * })
+ * ```
+ */
+export const SetParentAnnotation = makeUserAnnotation<boolean>({
```
dmaretskyi, line 577: Add to jsdoc that this does NOT guarantee that object's parrent is always set to this. it's just automatically updating parent on write.

packages/core/echo/echo-client-e2e/src/parent.test.ts
dmaretskyi: Done in 08a1a260.

Inlined `makeBody` at every call site in both suites (`parent.test.ts` and `Annotation.test.ts`) — the calls now read `Obj.make(Body, { text: 'body' })`, so the API under test is visible in the test body.

Counter-rule added as `no-trivial-wrappers-over-official-apis` in `.agents/skills/agentic-review/rules/code-style.mdl`: flags a module- or describe-local helper whose body is a single wrapped call adding no branching, error handling, derived values, or non-trivial defaults. It explicitly exempts helpers that compose several calls, encode a non-obvious setup sequence, are exported for reuse, or are a package's own published factory (`File.make` and friends). Also stated the convention directly in the `code-style` skill's Testing section, so it's guidance at authoring time and not only a review finding.

---

packages/core/echo/echo/src/internal/Annotation/annotations.ts
dmaretskyi: Done in 83ca7115. The JSDoc now leads with the caveat:

> This is NOT an invariant: it does not guarantee that a target held here has this object as its parent, only that a write through this field updates the parent. Nothing stops `Obj.setParent` from re-parenting the target afterwards, a ref whose target is not resolved is left alone, and removing a ref does NOT clear the target's parent […] Read the parent with `Obj.getParent`; never infer it from the field.

Same wording added to the `echo` skill's `SetParent` section, since that's where someone reaching for the annotation is likely to read about it.

---

packages/core/echo/echo/src/internal/Obj/parent-annotation.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const setParent = (value: unknown, parent: unknown): void => {
+  if (!Ref.isRef(value)) {
+    return;
+  }
+  // Only an already-resolved target can be re-parented synchronously; an unresolved ref names an
+  // object whose parent edge is either already written or not this holder's to write. `.target`
+  // throws on a ref with neither an inlined target nor a resolver, so gate on `isAvailable`.
+  const target: any = value.isAvailable ? value.target : undefined;
+  if (target == null || target[KindId] !== EntityKind.Object || target[ParentId] === parent) {
```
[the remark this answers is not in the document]
dmaretskyi: Valid — fixed in 0c1c1d7a.

The guard now compares ids: `target[ParentId]?.id === parent?.id`. Worth being precise about the failure mode, since it isn't a wrong parent. `target[ParentId]` on a database object resolves the edge through the working set, which can hand back a different proxy for the same entity; the identity compare then missed the short-circuit and re-wrote an unchanged parent ref on every `Obj.update` of the holder — a wasted Automerge op, not a bad edge. Identity can never produce the opposite error (same proxy implies same entity), so nothing was mis-parented.

Test assertions in both suites now compare `.id` too. Green: echo 580, echo-client-e2e 324 (+1 expected fail), compute 73, assistant-toolkit 66.

---

.agents/skills/echo/SKILL.md
```diff
@@ -161,6 +161,29 @@ export const make = (props: Obj.MakeProps<typeof Person>): Person => Obj.make(Pe
 export const make = (props: Obj.MakeProps<typeof Person>): Type.InstanceType<typeof Person> => Obj.make(Person, props);
 ```
 
+## Owned children — `Annotation.SetParent`
+
+Declare ownership on the ref field rather than calling `Obj.setParent` next to every write. Writing a
+ref into an annotated field (or creating the holder with one) sets the target's parent, so the child
+cascade-deletes and deep-clones with its holder.
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 0c1c1d7a — the section now reads "sets an object-kind target's parent", and adds that a ref to anything else (a relation) is left alone, since only objects can have a parent.

The SkillSpector AS3 warnings on this file are false positives: `.agents/skills/echo/SKILL.md` links to sibling skill docs for humans and agents to read — that cross-referencing is the documented layout in `AGENTS.md`, not a skill reading peer skills at runtime.

---

PR #12807, opened 2026-08-27
plugin-assistant: recover running state when the chat remounts

dmaretskyi:
Fixes DX-1198 ("Chat prompt doesn't recover running state").

## Root cause

The chat's `active` / `streaming` state lives on `AiChatProcessor`, and a processor is created per mount (`useChatProcessor`). The agent, by contrast, runs in a process that outlives the mount — `AgentService.getSession` explicitly reuses a still-running process "after the UI remounted". Nothing reconnected the new processor to that process, so `active` started `false`: the status pill disappeared, streamed blocks stopped landing, and the send button re-enabled while the agent kept working.

### Evidence (debug bundle `composer-feedback-logs-dev/logs/2026-08-27/5dddaa06-…ndjson`)

```
12:15:14.745 I init agent session   {feed: echo://B5YFIU…/01M0YTPX1AW5CRD4JCFMTDM2FC}  processor.ts
12:15:14.762 D lifecycle: spawn     {pid: f90fcae8-…, key: org.dxos.testing.process.agent, name: Agent}
12:15:14.843 D chat processor submitPrompt returned, waiting for agent
12:15:16.541 D creating processor   …                                                   useChatProcessor.ts
12:15:24.053 D creating processor   …                                                   useChatProcessor.ts
12:15:29.935 D lifecycle: spawn     {key: org.dxos.operation.sandbox.exec, parentPid: f90fcae8-…}
12:15:34.611 D lifecycle: spawn     {key: org.dxos.operation.sandbox.create, parentPid: f90fcae8-…}
```

Agent process `f90fcae8` keeps spawning tool calls well past the two remounts at 12:15:16 and 12:15:24, and no `session complete` / `setting active to false` ever follows — the turn was still in flight when the user filed the report a minute later. The remounted processors were simply blind to it.

## Change

- `AgentService.Session.running` — a derived `Atom`, `true` while the process is `RUNNING` or `HYBERNATING` (working, or parked on a tool call/alarm), `false` when `IDLE` (awaiting input) or terminal. Derived from the process's own `statusAtom`, which the process manager writes on the app-wide registry the UI reads, so it updates in place.
- `AiChatProcessor.adopt()` — resolves the session via `getSession` (which returns the process a previous mount left running) and subscribes to `running`. On a turn this processor did not issue it flips `active`, attaches the ephemeral-trace collector (streamed blocks + MCP errors) and follows the turn to completion; a turn issued here stays owned by `request()`, which reports its own errors. Returns a disposer.
- The observing fiber is tracked as `#observeFiber`, apart from `#requestFiber`, and interrupted by that disposer — otherwise each remount left a live collector on the session's stream, and `cancel()` would conflate "stop the request this chat issued" with "stop observing someone else's turn".
- `useChatProcessor` calls `adopt()` per created processor and runs the disposer on cleanup.
- The `getSession` call and the ephemeral-collector fork are factored into `#getSession()` / `#forkEphemeralCollector()`, shared by both paths.

Notes:
- `getSession` spawns when no process exists, so a mounted chat now spawns its agent process up front rather than on first prompt; it parks in `IDLE` with `running === false`. Open on the review thread — a `spawn: false` option is the alternative if eager spawn is unwanted.
- The elapsed timer restarts when a remounted chat picks up a turn, since the true turn start is not recoverable from the process (`useRequestTiming` derives it from `active`).
- A turn that settles into `HYBERNATING` with only background sub-agents left clears the indicator, because `waitForCompletion` is `runUntilSettled` — the same semantics `request()` has always had, so an adopted turn behaves like a locally-issued one.

## Testing

- `moon run plugin-assistant:test -- src/processor/streaming.node.test.ts` — 3 passed:
  - a running session is adopted (indicator goes active, blocks land, settles inactive) and a live-but-idle one is not — verified it fails without `adopt()` (`expected [ false ] to include true`);
  - disposing the observer mid-turn stops the collector — a gated stub stream pauses between batches, the test disposes, then releases the gate; the released block never reaches the thread while the session still reports `running`. Verified it fails without the interrupt (`expected [ 'Still working…' ] to deeply equal [ 'Still' ]`).
- `moon run agent-runtime:test -- src/agent-service/AgentService.test.ts` — 2 passed; the new case (ungated control-plane suite, no LLM turn) asserts `running` is `false` for a spawned-but-unprompted agent and that the atom tracks the live process rather than the value at resolve time.
- `moon run plugin-assistant:build agent-runtime:build`, `:lint` for the three packages, `pnpm format`, and `moon run composer-app:check-boot-budget` (391 tasks) — `22 preload entries, 4.17 MB (budget: 25 entries, 4.45 MB)`.

Not verified in the running app — the fix is in the mount/adoption path, which the suites cover at the processor and service boundaries.

## Summary by CodeRabbit

- **New Features**
  - Chat now reconnects to ongoing agent work when the view is remounted, preserving the running indicator and streamed content.
  - Ongoing sessions are recognized while actively processing or awaiting completion.

- **Bug Fixes**
  - Navigating away and returning during a chat turn no longer loses in-progress output.
  - Stopping a chat observer now correctly halts further streamed content without affecting request cancellation.


packages/core/compute/compute/src/AgentService.ts
```diff
@@ -25,6 +26,13 @@ export interface Service {
    */
   getSession: (feed: Feed.Feed, options?: GetSessionOptions) => Effect.Effect<Session>;
 
+  /**
+   * Resolves the agent session for a feed if a non-terminal process already exists, without
+   * spawning one. Lets a UI that remounted mid-turn (e.g. the user navigated to another page)
+   * re-attach to the running agent instead of rendering the conversation as idle.
+   */
+  findSession: (feed: Feed.Feed) => Effect.Effect<Option.Option<Session>>;
```
dmaretskyi: I think this is wrong api -- getSession is supoosed to return existing session if it exists, we done need a separate api

packages/core/compute/compute/src/AgentService.ts
```diff
@@ -58,6 +66,12 @@ export interface Session {
    */
   submitPrompt: (prompt: string | ContentBlock.Any[]) => Effect.Effect<void>;
 
+  /**
+   * True while the agent is working on a turn (running, or waiting on a tool call or alarm);
+   * false when it is idle awaiting input, or terminal.
+   */
+  isRunning: () => Effect.Effect<boolean>;
```
dmaretskyi: should be an atom for reactivity

packages/core/compute/compute/src/AgentService.ts
dmaretskyi: Done in 9fede9f — `findSession` is gone; the processor calls `getSession` on both paths through one `#getSession()` helper.

One consequence worth naming: `getSession` spawns when there is no process, so a mounted chat now spawns its agent process up front instead of on first prompt (it parks in `IDLE`, and `running` stays `false`, so nothing else changes). That was the reason for the separate non-spawning API. If eager spawn per mounted chat is unwanted, the alternative is a `spawn: false` option on `getSession` — say the word and I'll add it.

---

packages/core/compute/compute/src/AgentService.ts
dmaretskyi: Done in 9fede9f — `isRunning()` is now `readonly running: Atom.Atom<boolean>`, derived from the process's own `statusAtom` (`RUNNING || HYBERNATING`). Since the process manager and the UI share the app-wide registry, it updates reactively in place.

The processor now subscribes to it (with `immediate`) instead of sampling once, so a chat also picks up a turn that starts after it mounts, not just one already in flight; `adopt()` returns a disposer that `useChatProcessor` runs on cleanup. The agent-runtime test asserts the atom is derived from the live process rather than fixed at resolve time.

---

dmaretskyi:
On CodeRabbit's merge-risk note — one of the two risks was real and is fixed in 2783b9d; the other is pre-existing semantics.

**Fixed: observers alive after unmount.** The disposer unsubscribed the atom but left the observing fiber (and its ephemeral-trace collector) running, one per remount, feeding atoms nothing reads any more. The fiber is now tracked as `#observeFiber`, apart from `#requestFiber`, and interrupted from the disposer. Keeping them apart also stops `cancel()` conflating "stop the request this chat issued" with "stop observing someone else's turn".

**Not changed: activity clears while the agent hibernates with background work.** `waitForCompletion` is `runUntilSettled`, which by design settles on the foreground reply and does not wait on background sub-agents (a supervisor's delegated work reports back out of band) — the same call and the same semantics `request()` has always used, so an adopted turn now behaves exactly like a locally-issued one. Making the indicator span background work would be a deliberate product change to both paths, not a fix here.

Its "overlapping input during an active turn" concern does not apply: the composer's send button is gated on `active` (`ChatPrompt`), which adoption sets, so a remounted chat blocks submission while the agent works — that is part of what this PR restores. `#observe` is additionally guarded on `#requestFiber`/`#observeFiber`, so a turn is never observed twice.

Verified after the change: `plugin-assistant:test src/processor/streaming.node.test.ts` 2 passed, `plugin-assistant:build`, lint, and `pnpm format` clean.

---

packages/plugins/plugin-assistant/src/processor/processor.ts
```diff
@@ -394,6 +376,113 @@ export class AiChatProcessor {
...
+          // value matters as much as subsequent transitions.
+          { immediate: true },
+        );
+      })
+      .catch((err) => log.warn('failed to attach to agent session', { error: err }));
+
+    return () => {
+      disposed = true;
+      unsubscribe?.();
+    };
```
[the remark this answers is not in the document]
dmaretskyi: Regression test added in 93033b2, alongside the interrupt from 2783b9d: a gated stub stream pauses between two batches, the test disposes the observer mid-turn, then releases the gate — the released block never reaches the thread and `active` is cleared. Verified it fails without the interrupt (`expected [ 'Still working…' ] to deeply equal [ 'Still' ]`).

The comment nitpick from the same review is applied in that commit too — the added comments are down to one load-bearing clause each, with the regression-history narration and assertion summaries removed.

---

dmaretskyi:
`Check / boot-budget` failed on `93033b2` without producing a verdict, and it is not this PR's overage. Evidence and what I did:

**The job produced no measurement.** It failed 1m42s in, its check output is empty, and `Check / boot-budget-comment` was skipped — so the `boot graph:` line never printed. `93033b2` itself changed only tests and comments, which cannot move the eager boot graph.

**Measured locally, twice, both under budget** (`moon run composer-app:check-boot-budget`):

| tree | result |
| --- | --- |
| branch at `93033b2` | `boot graph: 22 preload entries, 4.23 MB (budget: 25 entries, 4.45 MB)` ✅ |
| branch merged with `main` (`a09e18e`) | `boot graph: 22 preload entries, 4.17 MB (budget: 25 entries, 4.45 MB)` ✅ |

Same numbers the advisory comment reported for this PR (4.23 MB) and for the base (4.17 MB), so nothing here is near the ceiling — note the merged tree measures *smaller* than the branch alone, main's chunk boundaries having shifted.

**No fix to port.** The one open PR touching this area, dxos/dxos#12769, is docs-only (the per-commit timeline) and explicitly leaves fixing or re-baselining the budget as a separate call. Raising the budget for growth this PR did not cause would be widening it, so I have not.

I have no way to re-run the Depot job from this environment, so instead of a bare re-run: `main` was ahead, so I merged it (`780fe16`) and pushed the strengthened disposal assertion (`c5eb955`) — CI re-runs against the fixed-up head, and the boot-budget job gets a fresh attempt on content I have measured under budget. Validated before pushing: 3 processor tests, 2 agent-runtime tests, lint on the three packages, `pnpm format`, and the full composer bundle build (391 tasks).

If `boot-budget` fails again on `c5eb955` with an actual `boot graph:` line over 4.45 MB, that is a real overage on `main`'s current content and belongs in its own change, not this one.

---

PR #12812, opened 2026-08-27
plugins: make every QA flow resolvable, and add a linter that proves it

dmaretskyi:
Follow-up to dxos/dxos#12809. That PR added a `## QA` section to all 82 plugin specs; its
post-merge review found ~41 real defects. This fixes them, adds a linter so the largest class
of them cannot recur silently, and attaches the recorded demos to their plugins.

## The systemic defect

Most findings were one thing wearing different clothes: **a flow referencing a value it
never binds.** `$given.card` where `given:` declares no `card:`. `$created` where no step
declares `capture:`. Those steps cannot resolve their inputs — the spec reads perfectly and
cannot run.

#12809's verification checked that every *operation key* existed. It never checked that
every *variable* was bound. `tools/qa-lint/qa-lint.mjs` closes that gap: it parses each
flow's `given:` entries and `capture:` declarations and reports every unbound reference.

```
before:  29 flows with unbound references; 38 references total
after:    0 flows with unbound references;  0 references total
```

It reproduces the review's binding findings and catches eight more it did not report:
`plugin-deepseek`, `plugin-markdown` QA-2 and QA-3, `plugin-review`, `plugin-space` QA-2,
`plugin-thread`, `plugin-commerce`, `plugin-debug`.

Each is fixed one of three ways, chosen per flow rather than uniformly:

| Fix | When |
| --- | --- |
| named `given:` entry | the value is a precondition the runner must establish first |
| `capture:` on the producing step | an earlier step actually returns it |
| human-only teardown | the subject is created by a step no agent can perform, so nothing *can* bind it |

## A teardown that could delete a user's data

The most serious finding. `plugin-ibkr`'s cleanup resolved its victim as
`Filter.type(Portfolio)` then took `.objects[0]` — no name filter at all. Run against a space
holding a real portfolio, that deletes **the user's**, not the flow's, and cascade-deletes its
feed and reports. No operation returns the Portfolio, so nothing can bind it: teardown is now
human-only, and `given` requires the space to hold no Portfolio.

Four more teardowns resolved by name without a guard, so a miss passed `undefined` to
`removeObjects` (`inbox`, `studio`, `tasks`, `conductor`) — they now throw instead.
`tools/qa-lint/teardown-lint.mjs` makes the unsafe shape mechanically detectable: selection by
index, or a query with no throw guard. Both linters exit non-zero on failure, so CI and hooks
fail on a bad manifest rather than passing while printing errors.

## Contract mismatches

Verified against source, not inferred. (Generic parameters written in words below, since the
API strips angle brackets from PR bodies.)

- **`plugin-linear`, `plugin-trello`, `plugin-slack`** — six operation declarations took an
  `integration` ref. All three import `ConnectorSpec.GetSyncTargetsInput` / `SyncInput`, which
  take a `connection` ref instead. The same defect the review caught in `plugin-bluesky`; it
  did not report these three.
- **`plugin-bluesky`** — as above, plus a `failed` output field the runtime never returns;
  `test T-8` asserted it.
- **`plugin-discord`** — `SyncInput` has no `channel` field; sync is account-level. Needed the
  real `materializeTarget` operation documented to turn a remote target into a local channel
  ref, and its `remoteTarget` made required to match the handler's `invariant`.
- **`plugin-inbox`** — `addMailbox` places whatever it is given, but only `Mailbox.make`
  provisions the feed, tags and filters that `draftEmail` then requires.
- **`plugin-markdown`** — `op:update` takes `doc` as a ref; passing the live object fails with
  `ref.tryLoad is not a function`. QA-2 had the same bug as QA-1.
- **`plugin-sheet`** — `op:create` returns only an id and files the sheet in the **space root**
  (traced through `CollectionModel.add`), so the flow's collection expectation and its cleanup
  target were both wrong; `getRange` returns *stored* content, so the assert claiming
  `[[10],[20],[30]]` could never pass.
- **`plugin-table`** — `op:create` is not invocable through the invoker (`Cannot read
  properties of undefined (reading 'addType')`); marked `blocked:` per Rule 9.
- **`plugin-commerce`** — the `ResultMapping` omitted required `responseType` and
  `itemLocator` and used a CSS `selector` against a JSON response.
- **`plugin-lingo`, `plugin-chess-com`, `plugin-connector`** — the invoker passes inputs
  straight to handlers without decoding through the schema, so a handler calling
  `Database.load` needs a real ref, not an object or an id.

## Findings rejected, with evidence

Three review findings were wrong. Complying with the third would have broken a correct flow:

1. **`plugin-trello`** — "the assert reads `$result` with no `capture:`". Field Rule 4:
   `$result` *is* the step's own invoke result. `sync.ts` returns a `pulled` struct, so
   `$result.pulled.added` is valid.
2. **`plugin-brain`** — "`queryFacts` needs a `query` wrapper". The handler takes a flat
   struct; the flow already matched the runtime.
3. **`plugin-meeting`** — "replace `$created.object` with `$created`". The operations genuinely
   return an object wrapper. The *spec block* was the stale half, so that was fixed instead.

## Also corrected

`objects: []` teardowns that deleted nothing (`plugin-terra`, `plugin-tasks`); five asserts
calling an array method directly on `query().run()`, which returns an object and would have
thrown; a 404ing avatar fixture and a 4-byte non-decodable PNG (`plugin-crm`, `plugin-file`);
an assert that passed whenever any Journal existed (`plugin-tasks`); a single `RecordReview`
call asserting three reviews (`plugin-lingo`); a `do:`/`invoke:` disagreement about modifier
keys, where only `modifiers.shift` exercises the path the instruction describes
(`plugin-deck`); `covers:` entries naming requirements the flows never exercise; a missing
Extensions table (`plugin-sandbox`); and `op@1.0` appendix blocks under `op@1.1` tables.

## Demos attached to plugin assets

The six flows recorded against a live Composer now appear in their plugins' `dx.config.ts`
screenshots, so the plugin details surface shows the feature working rather than only a still.
`markdown` and `sheet` gain an entry alongside existing assets; `thread`, `illustrator`,
`support` and `search` had no screenshots at all. All six URLs were checked to return 200 and
to honour range requests, so the player can seek.

Full reel (23 chapters):
https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-08-27-qa-flows/all-flows.webm

`dx.config.ts` is listed in each package's `files` and is the manifest's metadata source of
truth, so this is consumer-observable — hence the changeset.

## Test plan

- `node tools/qa-lint/qa-lint.mjs` — 0 unbound references (was 38), exits 0.
- `node tools/qa-lint/teardown-lint.mjs` — 0 unsafe teardowns, exits 0.
- Structural sweep over all 82 specs: one `## QA` each, every step carries `do:` + `expect:`,
  every operation key resolves to a real `DXN.make(...)` in source. 79/82 clean; the remaining
  three are the unbalanced fences in `plugin-studio` / `plugin-devtools` and a parser artifact
  on `plugin-markdown`, all pre-existing on `main`.
- `pnpm format` clean; branch synced with `main`.

https://claude.ai/code/session_0176YFWdREFPgiZyoB1GyjQb

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QA flow setup, validation, synchronization, cleanup, and media checks.
  * Corrected handling of connections, object references, query results, and operation outputs.
  * Clarified blocked scenarios and separated specification defects from test-driver errors.

* **New Features**
  * Added automated checks for unbound references and unsafe cleanup steps.
  * Added dark-mode demo videos for several plugins.

* **Documentation**
  * Updated QA instructions, prerequisites, coverage, and cleanup guidance to reflect actual behavior.


dmaretskyi:
Two of the three findings on `221d54d` are fixed in `515f039`; the third is already handled.

**Fixed — `plugin-commerce` `searchSchema` declared as `JsonSchema`.** Correct: `SetProviderTemplate.input.searchSchema` is `Schema.String` (`src/types/SearchOperation.ts:106`) and the handler parses it, so the manifest declaration contradicted the serialized value the flow already sends. Declaration is now `string`, with a note on why, and T-2 now says the stored value is the *parsed* input. `Provider.searchSchema` stays `JsonSchema` — that is the persisted field, not the operation input.

**Fixed — `plugin-lingo` QA-2 non-determinism.** Correct: the second step asserts `$analyzed.cached === false`, which a leftover `Analysis` from an earlier run would falsify. Added a precondition requiring no `Analysis` for this document and "English". The `after` stage already deletes the Analysis, so a clean run leaves the fixture in that state.

**Skipped — `tools/qa-lint/teardown-lint.mjs` "apply the unguarded-query rule only when `hasQuery`".** Already the behaviour on this head. The loop skips any step whose body contains no `db.query` before the rule is reached:

```js
if (!/db\.query/.test(step.join('\n'))) {
  continue; // removes a bound value.
}
```

Direct-reference cleanups such as `refs: [$given.materialized.target]` therefore never reach the `guarded`/`byIdentity` check and are not flagged — the linter reports `0` on this branch, with no valid flow rejected.

Verified on the new head: `qa-lint` → `0 flows with unbound references; 0 references total`; `teardown-lint` → `0 teardown step(s) that could remove an object the run did not create`; both exit 0. `pnpm format` clean.

---

PR #12825, opened 2026-08-28
client-services: replicate halo documents between devices

dmaretskyi:
Follow-up to #12774. Closes the last two gaps in the automerge-credential path on the client side: HALO documents had no replication between an identity's devices, and a legacy space grew one root per device.

## 1. HALO documents replicate between devices

`IdentityManager._constructSpace` registered exactly one extension on an authorized device session — `dxos.mesh.teleport.gossip`. Data spaces additionally register `dxos.mesh.teleport.automerge`; the HALO never did.

So HALO's space root and the credentials document it names had no replication path between an identity's own devices. A joining device was handed the root URL through the invitation, could not fetch it, and gave up — leaving it on a feed-only credential chain.

- `IdentityManager` takes a `MeshEchoReplicator`, authorizes the peer device for the HALO `space.id`, and registers the automerge extension on halo sessions. The replicator's share policy and collection sync are already generic over space id.
- Adoption retries with exponential backoff — the root arrives over the mesh without emitting an identity state update, so nothing else would retry.
- A successful adopt falls through to opening the credentials document instead of returning early.
- `service-stack.ts`: `MeshEchoReplicatorLayer()` sits below `IdentityManagerLayer` so the service is visible to it, while `registerReplicator(MeshEchoReplicatorService)` stays above, where the echo host is reachable. This ordering is load-bearing: without it the identity manager logs `p2p automerge replication disabled for the halo space` and skips, and the halo-root test below times out rather than failing.

## 2. A legacy space adopts the root edge mints

A legacy space is held by **every** one of an identity's devices, and none was told a root, so each called `EchoHost.migrateSpaceToRootDocument` and minted its own over the same space. Edge's record decided only whose report landed first — which no client ever read, and `adoptSpaceRoot` would have thrown on anyway.

DESIGN decision 3 asks that *"the migration is driven by one writer and the others adopt the root they receive."* Edge is now that writer (dxos/edge#945), because it already serializes per space and needs no election.

- `_anchorSpaceOnRootDocument` calls `_requestSpaceRootFromEdge` for a legacy space instead of minting. An unreachable edge leaves it unanchored for that attempt, which the anchor's existing backoff retries; the client never mints, so there is nothing to fork.
- `EdgeHttpClient.recordSpaceRoot` takes `rootDocumentUrl` **optionally**. Omitted asks edge to mint; a root-anchored space still names its own, because its id is the hash of that document's id.
- `adoptSpaceRoot` **replaces** a disagreeing local root instead of throwing, and takes `fetchFromNetwork`. A root edge minted has never been seen locally so it must be fetched, while a root an inviter merely named stays local-only — blocking there cost every device join 5s. Edge's record is write-once and identical for every device, so a local root that disagrees can only be one this device minted before that record existed.

**HALO still mints locally and its multi-device race is still open:** edge learns a directory from an `Epoch` credential and HALO has none, so edge would have nothing to anchor. Recorded in TASKS.md rather than left implied.

All of it sits behind the existing `DX_AUTOMERGE_CREDENTIALS` opt-in.

## Design notes corrected

DESIGN.md and TASKS.md record the resolution and fix the two notes that let this through:

- *"edge cannot IDENTIFY a space root"* (true) was read as *"cannot OWN one"* (does not follow) — identifying and minting are different problems, and only the first was blocked.
- *"There is no discovery problem — the pointer is handed over"* holds for a peer **joining** a space. Every path it cites hands the pointer to a peer being admitted. The peers migration exists for are **existing members**, admitted long before `space_root_url` was a field.

## Testing

`device-invitation-protocol.test.ts` — the weak assertion is replaced with real convergence: the guest adopts the host's exact root URL. Verified in both directions: without the layer ordering above the test fails at the timeout, with it the guest converges in ~1s.

`space-root-migration.test.ts` is the end-to-end scenario: identity A on two devices (A1, A2) sharing two spaces (S1, S2) with identity B on one device (B1).

- A1 migrates S1, and A2 and B1 both converge on the root **edge** minted rather than each anchoring one of their own. One `TestEdge` is shared by all three peers, which is the property under test — it mints once per space, records write-once, and hands every later caller the same url.
- S2 is the control: edge declines to anchor it, so no peer does and its control feed keeps resolving. Migration is per space.
- The HALO is anchored on both devices, and a space A1 creates afterwards still reaches A2 — which it can only do on credentials both devices agree about.

Verified not vacuous: a probe proves every assertion executes, and substituting a bogus root fails against A2's real one.

The four migration tests in `data-space-manager.test.ts` now drive a minting edge stub, since the client no longer anchors on its own.

To run any of this, a test host needs an edge HTTP client, and none could be injected — one exists only with a configured endpoint, and only alongside an edge socket. `ClientPlatformLayer` now takes an `edgeHttpClient` override beside `signalManager`/`transportFactory` and applies it with no endpoint: the socket and the REST client are independent capabilities, and the credential paths need only the latter.

Full `client-services` suite: **192 passed, 3 skipped (33 files)**. Merged with `main`, lint clean, formatted.

https://claude.ai/code/session_011R2pdrg6peV2gcGqT6ov4c

---
🚀 Composer preview: https://pr-12825-composer-dev.dxos.workers.dev


.agents/projects/registry.yml
```diff
@@ -26,6 +26,16 @@ projects:
...
...ator() defaults true, and subduction serves raw bytes, never materialized documents); and acceptSpace took a spaceRootUrl and ignored it, minting a competing root. BLOCKED, the only thing stopping edge#945 from carrying its e2e suite: edge's catalog:dxos pin must advance past #12825, which needs it merged AND a pkg.pr.new build published (branch commits are never built — only main). The repin is fully verified locally at 10defed3: suite 5/5, plus two unrelated breaks the bump surfaced — ComputeGraphModel.builder is gone (use model.createNode/createEdge directly) and @dxos/plugin-registry's #commands imports condition lacks a `workerd` entry, so a workerd bundle resolves `node`, pulls in @dxos/app-framework/vite-plugin and dies on fs; that second one is fixed in #12825. Suite recoverable from `git show 7db04a6:packages/services/edge/test/credential-upgrade.node.test.ts`. NEXT: land #12825, repin edge, restore the suite. THEN the real remaining work — the CLIENT read flip (no CredentialSource behind ControlPipeline yet, so the document is written but never authoritative) and all of Phase 3. NOTE for edge work in this sandbox: moon is NOT preinstalled, run `proto install moon 2.4.5`."
```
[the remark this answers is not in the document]
dmaretskyi: Agreed on the diagnosis — a bare `945` in `prs` would resolve as dxos/dxos#945, since `history` runs `gh pr view <n>` with no `--repo`.

Solved it differently, though. `'edge#945'` isn't a form the registry schema defines or the `history` verb parses, so adopting it would swap a wrong lookup for a failed one and imply support that doesn't exist. Since `prs` is a dxos-repo list, I dropped the entry (with a comment saying why) and left the edge PR named as `dxos/edge#945` in the entry's `resume`, where it already was and where nothing tries to resolve it.

Teaching the registry a qualified cross-repo form is a reasonable follow-up, but it's a change to the task-planning skill rather than to this PR.

---

packages/core/echo/echo-host/src/db-host/echo-host.ts
```diff
@@ -562,20 +562,39 @@ export class EchoHost extends Resource {
...
+    if (existing?.spaceRootDocUrl === spaceRootUrl) {
       return existing;
     }
+    if (existing) {
+      log.warn('replacing a space root this device minted with the one in force', {
+        spaceId,
+        was: existing.spaceRootDocUrl,
+        now: spaceRootUrl,
+      });
+    }
```
dmaretskyi, line 593: **The replacement path never switches the directory.**

A few lines below, the directory update is still guarded:

```ts
if (!this._spaceStateManager.getRootBySpaceId(spaceId)) {
  await this.updateSpaceRoot(ctx, spaceId, root.directory);
}
```

A device that already minted a root necessarily already has a directory, so on the replacement path the guard is false and `updateSpaceRoot` is skipped. The device ends up with `spaceRootRefs` pointing at edge's root while its data keeps using the *old* directory. For a legacy space the two directories should coincide (edge learns the directory from the `Epoch` credential, which is the one this device migrated), but nothing in the code enforces that — if they ever differ, the device silently forks and there is no signal.

Suggest either asserting `root.directory === this._spaceStateManager.getRootBySpaceId(spaceId)` on the replacement path, or calling `updateSpaceRoot` unconditionally.

---

packages/core/echo/echo-host/src/db-host/echo-host.ts
```diff
@@ -562,20 +562,39 @@ export class EchoHost extends Resource {
...
   /**
-   * Adopts a space root minted elsewhere, so a joining peer records the root the space already has
-   * rather than minting a second one over it. Idempotent; the root must name this space.
+   * Adopts a space root minted elsewhere, so a peer records the root the space already has rather
+   * than minting a second one over it. Idempotent; the root must name this space.
+   *
+   * A local root that disagrees is REPLACED, not kept: the root in force is the one EDGE recorded,
+   * write-once and identical for every device, so a local one can only be a root this device minted
+   * before that record existed. Keeping it would leave the device reading a document no one else
+   * writes to.
```
dmaretskyi: **Credentials written into the old credentials document are abandoned by the replacement, and the docstring doesn't say so.**

`refs.credentialsDocUrl` flips to the one carried by edge's root, with no merge of what the previous credentials document held. Anything this device mirrored locally that never reached the control feed is silently dropped.

I think the argument is "the control feed is still the source of truth during the migration window, so the document can always be rebuilt" — if that's right, please say it here, since it's the thing that makes the replacement safe and it isn't obvious from the code.

---

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -633,6 +639,26 @@ export class DataSpaceManager extends Resource {
...
 
+  /**
+   * Asks edge for this space's root, which it mints on the first such call, and returns the url in
+   * force. Undefined when edge is unconfigured or unreachable, or when it has not processed the
+   * space's control feed far enough to know its directory yet.
+   */
+  private async _requestSpaceRootFromEdge(space: DataSpace): Promise<AutomergeUrl | undefined> {
+    if (!this._edgeHttpClient) {
+      return undefined;
+    }
```
dmaretskyi, line 664: **Cross-repo ordering, and a failure mode that is invisible.**

This now posts `POST /db/spaces/:id/root` with an empty body, which only works against an edge that implements minting. That handler isn't in dxos/edge#945, so it has to already be deployed — otherwise every legacy space silently fails to anchor forever.

The catch below makes that indistinguishable from the expected case: a 404/500/auth failure and "edge hasn't processed the control feed yet" both produce the same `log(...)` at debug level. Suggest keeping `log` for the expected "not anchored yet" response and `log.warn` for anything else, so a misconfigured or un-upgraded deployment is actually visible in the field.

---

packages/sdk/client-services/src/packlets/spaces/data-space-manager.ts
```diff
@@ -600,12 +600,18 @@ export class DataSpaceManager extends Resource {
...
+          // them are handed the same root. An unreachable edge leaves the space unanchored for this
+          // attempt, which the anchor's own retry picks up.
+          const settled = await this._requestSpaceRootFromEdge(space);
+          if (!settled) {
             return false;
           }
 
-          log('migrated space to root document', { spaceId: space.id, refs });
+          const refs = await this._echoHost.adoptSpaceRoot(ctx, space.id, settled, { fetchFromNetwork: true });
+          log('adopted the space root edge minted', { spaceId: space.id, refs });
```
dmaretskyi, line 622: **`_reportSpaceRootToEdge(space)` still runs at the end of this function, which on this branch tells edge the root edge itself just minted.**

It's write-once so the response is the same url and nothing breaks, but it's a wasted round trip — and one that arms a background retry ladder (`SPACE_ROOT_REPORT_RETRY_*`) on every anchor. Worth skipping on the edge-minted branch and keeping it only for the branch where this peer named the root.

---

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -509,14 +523,18 @@ export class IdentityManager {
...
+          // document carries the chain, so the joining device takes the one the inviter named.
+          try {
+            await echoHost.adoptSpaceRoot(ctx, spaceId, adopted);
+          } catch (err) {
+            log('halo space root named by the inviting device has not replicated yet', { spaceId, adopted, err });
+            // The root arrives over the mesh without emitting an identity state update, and this
+            // retry is the only path, so it backs off to a ceiling rather than giving up.
+            scheduleTask(ctx, () => this._anchorHaloOnRootDocument(ctx, identity), this._haloAnchorRetryDelay);
+            this._haloAnchorRetryDelay = Math.min(this._haloAnchorRetryDelay * 2, HALO_ANCHOR_RETRY_MAX);
+            return;
```
dmaretskyi, line 542: **This retry never terminates.**

The delay caps at 30s and each attempt reschedules unconditionally, so a HALO root that will genuinely never arrive (inviter gone, document unrecoverable) means a 30s background loop for the lifetime of the ctx. The sibling path in `DataSpaceManager._reportSpaceRootToEdge` gives up at `SPACE_ROOT_REPORT_RETRY_MAX`.

The comment says why it backs off but not why HALO in particular can't give up where the space path does. If "the identity is unusable without it, so there is nothing better to do than keep trying" is the reason, say that — the asymmetry with the space path is the surprising part.

Also `_haloAnchorRetryDelay` is never reset on success; harmless today since there is one HALO per manager, but it makes the field's meaning depend on that.

---

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -538,40 +556,87 @@ export class IdentityManager {
...
+        queueMicrotask(async () => {
+          try {
+            if (!session.isOpen) {
+              return;
+            }
+            session.addExtension(
+              'dxos.mesh.teleport.gossip',
+              gossip.createExtension({ remotePeerId: session.remotePeerId }),
+            );
+            await this._connectEchoMeshReplicator(space, session);
```
dmaretskyi, line 611: **Nit / assumption check: `space` is in TDZ while `constructSpace` is still awaiting.**

The microtask closes over `const space = await this._spaceManager.constructSpace({ ... })`, so a callback that fired before that assignment completes would throw `ReferenceError` rather than doing anything useful. This relies on no authorized connection being possible before `constructSpace` resolves.

`DataSpaceManager._constructSpace` already has the identical shape, so this is the house pattern and presumably unreachable — flagging only so the assumption is a deliberate one rather than inherited.

---

packages/plugins/plugin-registry/package.json
```diff
@@ -26,10 +26,12 @@
     },
     "#commands": {
       "source": {
+        "workerd": "./src/capabilities/commands.browser.ts",
         "node": "./src/capabilities/commands.ts",
         "default": "./src/capabilities/commands.browser.ts"
       },
       "types": "./dist/types/src/capabilities/commands.d.ts",
+      "workerd": "./dist/lib/commands.browser.mjs",
```
dmaretskyi, line 34: **Unrelated to the PR's subject.**

The condition order is right (`workerd` before `node` in both maps, both pointing at the browser build), so no objection to the change itself — but it has nothing to do with HALO replication or edge-minted roots, and the changeset doesn't mention it. Either split it out or add a line saying what broke without it, so `git log` explains it later.

---

packages/sdk/client-services/src/packlets/spaces/space-root-migration.test.ts
```diff
@@ -0,0 +1,187 @@
...
+const spaceOn = (peer: ServiceContext, spaceKey: PublicKey) => {
+  const space = peer.dataSpaceManager!.spaces.get(spaceKey);
+  invariant(space, 'space not present on this peer');
+  return space;
+};
+
+describe('spaces/space-root-migration', () => {
+  test(
+    'one identity on two devices, sharing two spaces with another identity',
+    { timeout: 300_000 },
```
dmaretskyi, line 91: **CI cost.**

300s here, plus the device-invitation test going from its default to 90s with a 60s `waitForCondition`, and every assertion in this file is a poll-to-timeout. A regression in the anchoring path therefore costs minutes of wall clock per shard before it reports, rather than failing fast.

Fine if that's a deliberate trade for covering the real multi-device topology — but if the 120s polls are sized off one observed run rather than a bound you can argue for, they'll be the first thing to flake under CI load. Worth stating where the numbers come from.

---

dmaretskyi (commented):
Read the whole diff. The shape is sound — `registerReplicator` uses `Effect.serviceOption`, so hoisting it above the conditional `MeshEchoReplicatorLayer()` is safe when `disableP2pReplication` is set; the `TestEdge` / `mintingEdge` seams are clean and the shared-single-instance stub is the right way to express "edge is one writer".

Eight comments inline. The one I'd want resolved before merge is the first: `adoptSpaceRoot`'s new replacement path leaves the old directory in place, because the `updateSpaceRoot` call is still guarded on the device having no directory at all. Second is the cross-repo ordering — the empty-body mint endpoint has to be live on edge before this ships, and the current logging can't tell "edge is old" from "edge isn't ready yet".

The rest are smaller: an abandoned credentials document on replacement, a redundant report back to edge, a non-terminating HALO retry, a TDZ assumption inherited from `DataSpaceManager`, the unrelated `plugin-registry` export condition, and the CI budget of the new test.

---

dmaretskyi:
Pushed `90c9ca2` addressing my review comments, and verified it in a full checkout rather than by inspection:

- `moon run echo-host:build client-services:build` — clean.
- `moon run echo-host:lint client-services:lint` — clean.
- `pnpm format-check` — clean (14646 files).
- `moon run client-services:test` — **192 passed, 3 skipped (33 files)**, matching the number in the PR description.
- `moon run echo-host:test` — **361 passed, 2 skipped, 1 todo (24 files)**.

What changed:

1. `adoptSpaceRoot` calls `updateSpaceRoot` unconditionally (it already no-ops on an unchanged url), so a peer replacing a root it minted itself follows the new root's directory rather than silently keeping its own.
2. `_requestSpaceRootFromEdge` distinguishes edge's "not anchored yet" from every other failure of the mint endpoint. Edge now answers the pending case with a typed `EdgeFailure.data.type` (dxos/edge#945), so an un-upgraded or unreachable edge logs at `warn` instead of looking like the expected case at debug. The constant is duplicated on both sides with a note, for the same catalog-pin reason `orderCredentials` is.
3. `_reportSpaceRootToEdge` is skipped when edge is the one that minted the root.
4. Docstrings for what the replaced credentials document costs, and why the halo anchor caps its period rather than its attempt count.

Two of my comments I did **not** act on, since they are yours to decide: the unrelated `plugin-registry` workerd export condition, and the `space`-in-TDZ shape inherited from `DataSpaceManager._constructSpace`.

On the CI-budget comment — a datapoint rather than an objection: `space-root-migration.test.ts` finishes in **3.4s** against its 300s budget, so those numbers are slack rather than a tuned bound, and a regression would still sit on them for five minutes before reporting.

---

PR #12827, opened 2026-08-28
plugin-space: list initializing spaces as placeholders instead of hiding them

mykola-vrmchk:
## Summary

One space that cannot finish initializing must not take the workspace UI down with it. A user reported both prod and preview showing no space list; the profile's main workspace was stuck in `SPACE_INITIALIZING` for the whole session.

**Scope changed since the first review:** the `@dxos/client-services` root-document retry has been **reverted** (see below). This PR is now purely the client-side graceful-degradation fix.

## Why the space was stuck — server side

Traced in SigNoz against `db-service-production` for the reported window (2026-08-27 16:40–16:52 UTC):

- `SubductionAutomergeReplicator` for that space could not complete a **single** process pass: 20 passes attempted, 20 hung, each abandoned by the 60 s watchdog and re-armed 250 ms later.
- Inner failure: `subductionWasmMutex hold (getDocumentsBytes:getBlobs)` timing out at 10 000 ms, with `loadAllCommits failed` 138× across 16 distinct sedimentrees, plus 36 × `Durable Object storage operation exceeded timeout which caused object to be reset`.
- Consequence: **0** `acceptTransport` / handshake completions and **0** `collection-state sent`, versus 14 / 272 for a sibling space of the same identity in the same window. The client's frames arrived (127 enqueued) and were never answered, which is why client logs show zero `collection-query` sent for this space — it was waiting for a handshake, not idle.
- The hang ran at ~1/minute for hours on 25 and 27 August (166 occurrences in 8 h on the 27th, versus 1 apiece for six unrelated spaces), so it is a property of this profile's data rather than a fleet outage — and a wedge can outlast a session by many hours.

Server-side follow-ups belong in the edge repo (bounding `loadAllCommits`, not holding the WASM mutex across DO storage I/O, and serving the handshake ahead of document work) and are not in this PR.

## Changes

**`@dxos/plugin-space`** — the workspace list filtered strictly on `SPACE_READY`, so a stalled space disappeared entirely. Initializing spaces are now listed as **disabled loading placeholders** carrying their cached name (`namesCache`), with `loading: true` on the node. Every `space.properties` read in `constructSpaceNode` was already guarded on `SPACE_READY`, and `getSpaceDisplayName` / `checkPendingMigration` likewise, so a placeholder never touches an unopened db.

**`@dxos/plugin-navtree`** — the L1 panel for a loading workspace renders "Connecting to this workspace…" instead of `L1PanelContent`, on the same 1 s settle delay as the unavailable message so a normal boot stays quiet. Previously a stalled active workspace rendered *"You don't have this workspace, or it no longer exists"* — false for a workspace that exists and is merely unreachable, and the thing the user actually reported. New test id `navtree.workspace.loading`; `navtree.workspace.visible` / `.unavailable` keep their meaning.

**`@dxos/app-toolkit`** — `AppNodeMatcher.whenSpace` no longer matches an initializing space. This is load-bearing rather than cosmetic: every connector keyed on that matcher builds content by querying the space's db, and there are 13 direct call sites across 8 plugins plus `TypeSection`'s default used by 5 more. Listing the space without this guard would newly point all of them at a database that is not open. Only `SPACE_INITIALIZING` is excluded, so behaviour for every state that already produced a node is unchanged.

## Reverted: the root-document load retry

`packages/sdk/client-services/src/packlets/spaces/data-space.ts` is back to `main` verbatim, and `data-space.test.ts` is removed. Trade-off, stated plainly: the space no longer heals itself when the server recovers — and in this incident it *did* recover on its own, at 17:59:58 UTC the same DO logged `collection-state sent` (103 documents) on the same script version, about an hour after the user gave up. Without the retry a stuck space stays stuck for the session, so the graceful UI in this PR is the entire mitigation.

## Testing

Build, test and lint for `app-toolkit`, `plugin-space`, `plugin-navtree` — results in the PR thread once the run completes; the Check workflow is the authoritative gate.

https://claude.ai/code/session_01DMH2XQMyiJE98Q7GMPsNgy

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspaces that are still connecting now remain visible as disabled loading placeholders with their cached names.
  * The workspace panel displays a connecting message instead of treating unavailable workspaces as missing.
  * Workspace-dependent features no longer attempt to access a workspace before it has finished opening.
* **Localization**
  * Added a localized message for workspaces connecting and syncing.


mykola-vrmchk:
Known gap this PR deliberately does not close, so it isn't lost in review:

While a space is retrying, it still has no graph node, so if it happens to be the **active** workspace the panel reads *"Workspace unavailable — You don't have this workspace, or it no longer exists"* (`packages/plugins/plugin-navtree/src/translations.ts:20`). During an edge outage that statement is false: the workspace exists and is connecting.

This PR makes it **transient instead of permanent** — previously the space never left `SPACE_INITIALIZING`, so the message was terminal until an app restart (and a restart during the outage re-wedged it). That is the incident-blocking half.

Left out on purpose, because it is not a copy change: `L1Panel` only knows whether a workspace has a graph node, so distinguishing *initializing* from *never-had-it* requires `plugin-space` to publish initializing space ids to the navtree. That plumbing (and the question of whether such spaces should appear in the sidebar as loading placeholders at all) is a UI behavior decision worth its own PR rather than an amendment to an SDK patch that is green and reviewed.

---

PR #12831, opened 2026-08-28
plugin-claude: rename from plugin-claude-agents, retry 403s, read latest events

dmaretskyi:
### 1. Rename `plugin-claude-agents` → `plugin-claude`
- Directory `packages/plugins/plugin-claude-agents` → `packages/plugins/plugin-claude`.
- Package `@dxos/plugin-claude-agents` → `@dxos/plugin-claude`; plugin key `org.dxos.plugin.claudeAgents` → `org.dxos.plugin.claude`.
- Entry module `ClaudeAgentsPlugin` → `ClaudePlugin` (export subpath, `vite.config.ts`, `moon.yml` module-structure check, composer-app wiring).
- Operation and skill DXNs move too: `org.dxos.operation.claudeAgents.*` → `org.dxos.operation.claude.*`, `org.dxos.skill.claudeAgents` → `org.dxos.skill.claude`. The repo's `operation-key-shape` lint rule requires an operation's domain to match its defining package, so this is mandatory rather than optional — **anything holding the old operation keys needs updating**. The ECHO type names are unchanged, so existing objects and saved references keep resolving.
- `pnpm-lock.yaml` regenerated for the renamed workspace importer.

### 2. Retry 403s
`isRetryable` now treats `403` as retryable on both GET and POST, within the existing 3-retry / 1s fixed-delay budget. The edge proxy rejects intermittently with `{"status":403,"type":"forbidden","message":"Request not allowed"}`; a 403 is an authorization rejection that lands before the control plane acts, so retrying a POST cannot double-create. 5xx remains non-retryable on POST, which is the case that could double-write.

### 3. `list-events` defaults to the last N events
The Anthropic API defaults to `order=asc`, so a bare `limit` returns the *opening* events of a session — on a long run agents were reading the wrong end. `listEvents` now requests `order=desc` and reverses the page, returning the most recent N events in chronological order. `Get Claude Agent Transcript` takes an optional `order: 'first' | 'last'` (default `last`) to override; the skill text and `PLUGIN.mdl` (`req F-4.7`) document the behaviour.

### 4. Drive-by fix from review
`start-session` resolves session credentials before `createVault`, so a secret that fails to resolve can no longer leave an orphaned vault behind (the handler has no cleanup path).

## Testing

Unit tests in `src/api/client.test.ts` cover the 403 predicate and both `listEvents` orderings (request URL and page reversal).

Verified locally:
- `moon run plugin-claude:test` — 22 passed / 4 files
- `moon run plugin-claude:lint` — clean
- `moon run plugin-claude:check-module-structure` — clean (React unreachable from `ClaudePlugin` under node and workerd)
- `pnpm format` (oxfmt)

CI on `ef61afd` is green across the Check workflow, model-fixture, boot-budget and memory; Codecov +0.02% with all modified lines covered.

Preview: https://pr-12831-composer-dev.dxos.workers.dev

---
_Generated by [Claude Code](https://claude.ai/code/session_01FinFE8cY818z5XzWSy4FX6)_


dmaretskyi:
Disposition of the review, plus what the red CI turned up.

**Fixed and pushed** (`ef61afd`):
- `start-session`: `toVaultCredentials` now resolves before `createVault`, so a secret that fails to resolve can no longer orphan a vault. Valid finding.
- Trimmed the `listEvents` JSDoc to its caller-visible contract and moved the test helper below the suites.

**Also pushed** (`bf7e409`), from CI rather than the review: the lockfile was stale after the package rename (`ERR_PNPM_OUTDATED_LOCKFILE`), and the repo's `operation-key-shape` lint rule requires an operation's DXN domain to match its defining package — so `org.dxos.operation.claudeAgents.*` → `org.dxos.operation.claude.*` and `org.dxos.skill.claudeAgents` → `org.dxos.skill.claude` were mandatory, not optional.

**Not changing — 403 retries.** Retrying a 403 is the point of the change, not an oversight. A 403 is an authorization rejection: the request is refused at the edge proxy or the control plane's auth layer, before any agent, session, message or credential is created, so a retried POST cannot double-create. The risk framing would hold for a 5xx after a partial write, which is exactly why `isRetryable` still refuses those on POST.

**Not changing — comment-length nits** in `ClaudeManagedAgent.ts`, `ClaudeAgentSession.ts`, `keys.test.ts`, `credentials.ts`, `mapping.ts`. Those lines are pre-existing and unmodified; they only appear in the diff because the file moved. Rewriting them would widen a rename PR.

Verified locally before pushing: `plugin-claude:test` 22/22, `plugin-claude:lint` clean, `plugin-claude:check-module-structure` clean, `pnpm format` run.

---

PR #12837, opened 2026-08-28
echo: document deletion semantics and dangling-ref handling (docs-first)

wittjosiah:
Docs-first presentation of the dangling-refs design: the `API.md` files for `@dxos/echo` and `@dxos/echo-react` now describe the target API for refs whose target has been deleted. **The new sections are ahead of the implementation** — they are the spec to review; the code lands in follow-up PRs.

## The design

Deleted objects are invisible by default across the entire API, with one opt-in shape: `{ deleted: 'include' }`, the option queries already have.

- **Single refs** keep returning the `Ref` (URIs stay comparable); the target reads as absent: `.target` is `undefined`, `load()` fails, and `load({ deleted: 'include' })` resolves the tombstone.
- **Ref arrays hide entries whose target is known-deleted**, automatically and for every ref-array field. Deleting an object makes it vanish from every array that references it — no holder is swept, which is what currently forces verbs like DeleteTask to hand-sweep membership arrays (and still miss holders they don't know about). Filtering is lazy against the working set; `Ref.loadAll` closes the gap when completeness matters.
- **Writes are literal.** Splice the view (hidden entries unaffected); wholesale reassignment stores exactly what was assigned and is a reactivity anti-pattern besides.
- **One deletion rule**: the array filter uses the same transitive cascade (parent edge, relation endpoints) as queries and `Obj.isDeleted`, so no two surfaces disagree about membership.
- **Cardinality is a query, order is the array**: displayed counts come from `useQuery`/`db.query` (index-answered, no working-set flicker), never from a ref array's length.
- **GC prunes**: `runGarbageCollection({ pruneDanglingRefs: true })` physically removes hidden entries once the target is destroyed and undelete is impossible.

Deliberately absent: any API distinguishing "deleted" from "not yet synced" (state probes, error reasons, skip counts) — dropped in review as API weight without a proven consumer.

## Also in this PR

- Removed the deprecated `useObjects` hook from the echo-react doc.
- Wrapped the Effect snippets in `Effect.gen` — bare `yield *` outside a generator parses as multiplication, which is why oxfmt kept reformatting `yield*` to `yield *`.
- Documented `Annotation.SetParent` as the parenting API (declared on the ref field); `Obj.setParent` is slated for removal and is not documented.

## Not yet implemented

Everything under "Working with Refs" beyond today's `.target`/`.load()` behavior: `load({ deleted: 'include' })`, `Ref.loadAll`, array filtering and its write semantics, the `useObject(ref, { deleted: 'include' })` overload, and `pruneDanglingRefs`. Queries excluding deleted objects and `Query.options({ deleted })` already work today.

Docs-only change; the `pipeline-discord` replay-fixture test failure reproduces on main and is unrelated.

### DIFF
--- a/packages/core/echo/echo-react/API.md
+++ b/packages/core/echo/echo-react/API.md
@@ -5,7 +5,6 @@
 ```ts
 import {
   useObject, // Subscribe to a single Echo object or Ref
-  useObjects, // Subscribe to multiple Refs
   useQuery, // Reactive query subscription
   useType, // Type (schema registry) subscription
 } from '@dxos/echo-react';
@@ -34,11 +33,15 @@ const [name, setName] = useObject(obj, 'name');
 
 ### Subscribing to a Ref
 
-Automatically dereferences the Ref and handles async loading.
+Automatically dereferences the Ref and handles async loading. A deleted target reads as `undefined`,
+the same as a target that has not loaded — deleted objects are invisible by default throughout the
+API.
 
 ```ts
 const [assignee, setAssignee] = useObject(task.assignee);
-assignee?.name; // undefined until the Ref loads
+assignee?.name; // undefined until the Ref loads, and undefined if the target is deleted
+
+const [ghost] = useObject(task.assignee, { deleted: 'include' }); // resolves a tombstoned target
 ```
 
 ### Updating objects
@@ -70,17 +73,6 @@ const [target] = useObject(maybeRef); // Obj.Snapshot<T> | undefined
 
 ---
 
-## useObjects
-
-Subscribe to multiple Refs' target objects. Returns loaded snapshots, filtering out unloaded refs.
-
-```ts
-const snapshots = useObjects(task.watchers);
-snapshots.length; // only includes refs that have loaded
-```
-
----
-
 ## useQuery
 
 Create a reactive subscription to a database query or filter.
@@ -102,6 +94,20 @@ const tasks = useQuery(space.db, Filter.type(Task, { completed: false }));
 - Accepts both `Query` and `Filter` objects (filters are converted to queries internally).
 - The query is memoized based on its AST -- no need to wrap with `useMemo`.
 - Returns an empty array when `resource` is `undefined`.
+- Deleted objects are excluded by default; opt in with `Query.type(Task).options({ deleted: 'include' })`.
+
+### Counting
+
+Cardinality is a query, order is the array: when the UI shows a count, take it from a `useQuery`
+result, never from the length of a ref array. The query count is answered from the index and only
+moves on real writes and replication; a ref array's length varies with which targets happen to be
+loaded locally.
+
+```ts
+const members = useQuery(space.db, Filter.childOf(taskSet));
+members.length; // stable count, deletion respected
+taskSet.tasks.length; // locally visible entries -- not a number to display
+```
 
 ---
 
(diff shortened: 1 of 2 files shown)


packages/core/echo/echo-react/API.md
```diff
@@ -34,11 +33,15 @@ const [name, setName] = useObject(obj, 'name');
...
+Automatically dereferences the Ref and handles async loading. A deleted target reads as `undefined`,
+the same as a target that has not loaded — deleted objects are invisible by default throughout the
+API.
 
 ```ts
 const [assignee, setAssignee] = useObject(task.assignee);
-assignee?.name; // undefined until the Ref loads
+assignee?.name; // undefined until the Ref loads, and undefined if the target is deleted
+
+const [ghost] = useObject(task.assignee, { deleted: 'include' }); // resolves a tombstoned target
```
dmaretskyi, line 44: 👍

packages/core/echo/echo/API.md
```diff
@@ -111,6 +130,51 @@ Relation.getSource(rel);
...
+```ts
+// Create
+const ref = Ref.make(alice);
+
+// Dereference synchronously (working set only).
+task.assignee.target; // undefined until loaded — and undefined if the target is deleted
+
+// Load through the resolver.
+await task.assignee.load(); // throws when the target is unavailable or deleted
+await task.assignee.load({ deleted: 'include' }); // resolves a tombstoned target
```
dmaretskyi, line 144: 👍

packages/core/echo/echo/API.md
```diff
@@ -65,6 +65,25 @@ const Task = Schema.Struct({
...
+### Owned children
+
+Ownership is declared on the ref field with `Annotation.SetParent`. Writing a ref into an annotated
+field sets the target's parent, so the child cascade-deletes and deep-clones with its holder; there
+is no imperative parent API.
+
+```ts
+const Doc = Schema.Struct({
+  content: Ref.Ref(Text).pipe(Annotation.SetParent.set(true)),
+  sections: Ref.Array(Ref.Ref(Section)).pipe(Annotation.SetParent.set(true)),
```
dmaretskyi:
```suggestion
  sections: Ref.Array(Ref.Ref(Section).pipe(Annotation.SetParent.set(true)),
```

packages/core/echo/echo/API.md
```diff
@@ -111,6 +130,51 @@ Relation.getSource(rel);
...
+const program = Effect.gen(function* () {
+  const assignee = yield* Database.load(task.assignee); // fails with EntityNotFoundError
+  const watchers = yield* Ref.loadAll(task.watchers); // order kept, deduped by id, deleted skipped
+});
+```
+
+A ref to a deleted object keeps its URI — id comparison and grouping keep working — but its target
+behaves as absent everywhere.
+
+### Ref arrays
```
dmaretskyi, line 156: this seems like a bigger obfuscation: array index is different depending on how you look at it.
let me think about it

dmaretskyi: also, this wood mean cascade reads into referenced objects

packages/core/echo/echo/API.md
wittjosiah: Applied in 217f5ae84d0 — annotation moved onto the element ref inside Ref.Array.

packages/core/echo/echo/API.md
wittjosiah: Resolved by dropping array hiding entirely (40c44bbc9a2): the array now reads exactly what it stores, so indexes are the same from every angle. Enumerating live members goes through a query instead (childOf / .reference), which already respects deletion — no cascade reads on property access either.

packages/core/echo/echo/src/internal/Obj/atoms.ts
```diff
@@ -55,30 +55,34 @@ const objectFamily = Atom.family(<T extends Obj.Unknown>(obj: T): Atom.Atom<Obj.
...
+        unsubscribeTarget?.();
+        unsubscribeTarget = subscribe(target, () => {
+          get.setSelf(read(target));
+        });
+        // Runs at once when the node was disposed while the target loaded.
+        get.addFinalizer(unsubscribeTarget);
+        return read(target);
+      };
+
+      return loadRefTarget(ref, get, setupTargetSubscription);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d535e366b09 — real bug, and it had no coverage. `loadRefTarget` now takes the load options, since `ref.target` never yields a deleted object and the async load is the only path that reaches one. Added a test that fails without it.

packages/core/echo/echo-client/src/hypergraph.ts
```diff
@@ -728,12 +729,18 @@ export class HypergraphImpl implements Hypergraph.Hypergraph {
...
+    options?: Ref.LoadOptions,
+  ): Promise<Entity.Unknown | undefined> {
     const db = this._databases.get(spaceId);
     if (!db) {
       return undefined;
     }
-    const [obj] = await db.query(Query.select(Filter.id(objectId)).from(db, { includeFeeds: true })).run();
+    const select = Query.select(Filter.id(objectId));
+    const scoped = options?.deleted === 'include' ? select.options({ deleted: 'include' }) : select;
+    const [obj] = await db.query(scoped.from(db, { includeFeeds: true })).run();
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d535e366b09. A merged-away loser is a redirect rather than a tombstone worth handing back, so resolution follows the chain even under `{ deleted: 'include' }`.

packages/core/echo/echo/src/Filter.ts
```diff
@@ -106,6 +106,16 @@ export const id = (...ids: EntityId[]): Any => {
...
 
+/**
+ * Filter by the id of an entity already in hand, keeping its type for the rest of the chain.
+ *
+ * @example
+ * ```ts
+ * db.query(Query.select(Filter.entity(task)).reference('watchers'));
+ * ```
+ */
+export const entity = <T extends Entity.Unknown>(entity: T | Entity.Snapshot): Filter<T> => id(entity.id);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d535e366b09. I had assumed TypeScript could not infer through `Omit`, but it does: the snapshot overload constrained to `Obj.Unknown` keeps the entity type. Pinned by a type annotation in the `accepts a snapshot` test, which fails against the previous signature.

packages/core/echo/echo/API.md
```diff
@@ -111,13 +130,58 @@ Relation.getSource(rel);
...
+
+// Effect forms.
+const program = Effect.gen(function* () {
+  const assignee = yield* Database.load(task.assignee); // fails with EntityNotFoundError
+  const watchers = yield* Ref.loadAll(task.watchers); // order kept, deduped by id, deleted skipped
+});
+```
+
+A ref to a deleted object keeps its URI — id comparison and grouping keep working — but its target
+behaves as absent everywhere.
```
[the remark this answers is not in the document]
wittjosiah: Fixed in d535e366b09 — the line now reads "behaves as absent unless the read opts into deleted entities".

.changeset/deleted-ref-targets.md
```diff
@@ -0,0 +1,5 @@
+---
+'@dxos/echo': minor
+---
+
+Deleted objects are absent by default when resolving references, matching how queries already treat them. `ref.load`, `ref.tryLoad`, `Database.load`, `Obj.atom` and `useObject` each take `{ deleted: 'include' }` to resolve a tombstoned target instead, and a reference carrying an inlined target applies the same rule rather than returning a deleted object indefinitely. Adds `Filter.entity(entity)`, which selects by an entity's id while keeping its type for the rest of the query chain.
```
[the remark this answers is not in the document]
wittjosiah: Fixed in b0b4d661ace. The ambiguity mattered here, since arrays keeping their dangling entries is the point of the design: the line now reads "Resolving a reference whose target has been deleted now yields nothing by default ... a ref array still stores its dangling entries."

PR #12849, opened 2026-08-30
tooling: evaluate Stryker mutation testing for echo-client and compute-runtime

dmaretskyi:
Evaluation only — nothing is wired into CI. StrykerJS 10 + `@stryker-mutator/vitest-runner` was actually run against both packages; the write-up is in `agents/superpowers/specs/stryker-mutation-testing-evaluation.md`. The PR also strengthens three existing compute-runtime test files against the survivors the run exposed.

## Measured results

| Package | Scope | Mutants scored | Score | Wall clock (4 cores) |
| `compute-runtime` | whole `src` minus tests/`testing`/`LayerStack.ts` | 2759 | **42.5%** (59.0% of covered) | 43m11s |
| `echo-client` | `util/migrate-document.ts`, `query/query-result-cache.ts` | 63 | **71.0%** | 2m45s |
| `compute-runtime` | `object-template.ts`, `protocol.ts` | 262 | 24.4% | 1m04s |

Baselines: `compute-runtime:test` = 21s, `echo-client:test` = 65s. (2837 mutants were instrumented; 78 static ones were excluded by `ignoreStatic`.)

Uncovered by any test in `compute-runtime`: `url.ts`, `functions-trace.ts`, `services/s3-host.ts`, `services/credentials.ts`. Tested but weakly asserted: `ProcessManager.ts` 45%, `ProcessHandle.ts` 40%, `trigger-dispatcher.ts` 57%, `trace-buffer.ts` 2.9%. Strong: `SwarmTraceSink.ts`/`RemoteTraceMonitor.ts` 100%, `TriggerMonitor.ts` 85%.

## Caveats found by running it

1. The instrumenter crashes the whole run on `inDegree[i]!++` (`LayerStack.ts:706`, `:730`) — a `@babel/types` bug in `@stryker-mutator/instrumenter@10`. The file is excluded.
2. `inPlace: true` is required (per-package sandboxing breaks the root `vite.base.config.ts` import and pnpm symlinks) and it rewrites the working tree, adding `// @ts-nocheck` to every file. It restores on exit, but a hard kill leaves the tree dirty.
3. Under pnpm the runner plugin must be named by explicit path in `stryker.conf.json`.
4. Installing it at the workspace root makes pnpm rewrite peer-dependency keys across the entire lockfile (~1600 lines), which CI's frozen install rejects with `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY`. Not worth a permanent lockfile change for an on-demand tool, so the dependencies are **not** committed — the doc installs them per session.

Not problems: vitest 4.1.10 is supported, the forced `pool: threads` did not break `node:sqlite` or automerge WASM, and `@effect/vitest` tests are killed normally.

## Test hardening from the survivor list

Suite goes 164 → 176 passing, tests only, no source changes:

- `ProcessManager.test.ts` — list filtering by `parentProcessId`/`state`/`target` (all three branches previously survived as `if(false)`), termination cascading to non-terminal children only, a shutdown→startup→shutdown cycle catching an inverted `#shutDown` guard, and `Process.Info.error` serialization (previously asserted by no test at all).
- `process-store.test.ts` — index de-duplication, sequence-counter resumption from persisted events, strictly-increasing seq with targeted event removal, per-process state freed on delete, missing-record filtering.
- `trigger-dispatcher.test.ts` — `retryPending` transitions around `RunAgainError`, `running`/`errors` lifecycle state, and refusal paths for disabled and no-runnable triggers.

Survivors that only mutate `log()`/`invariant()` arguments were deliberately left alone — they dominate otherwise well-tested files and killing them adds noise, not signal.

## Measured effect of the hardening

Re-running Stryker over just the four affected source files, both runs with `ignoreStatic`:

| File | Before | After | Delta |
| --- | --- | --- | --- |
| `process-store.ts` | 77.27% | **93.18%** | +15.9 |
| `ProcessManager.ts` | 45.00% | **53.48%** | +8.5 |
| `ProcessHandle.ts` | 39.64% | **45.33%** | +5.7 |
| `trigger-dispatcher.ts` | 56.62% | **58.05%** | +1.4 |

(Measured before the review-requested test removals; the dispatcher figure is now lower again.) The spread is the point: a small self-contained store moves a long way on a few assertions, while the dispatcher barely moves because its remaining survivors are mostly `log()`/`invariant()` arguments and private helpers with no public seam.

## Recommendation

Not a CI gate — ~100× the suite cost, and the score is dominated by uncovered files that line coverage reports more cheaply. Useful as an on-demand audit of a specific module before hardening it, scoped with `ignoreStatic` and a narrow `mutate` glob.

## Changes

- `stryker.conf.json` in each of the two packages.
- `.gitignore`: `.stryker-tmp/`, `reports/mutation/`.
- The evaluation document.
- Strengthened assertions in three existing compute-runtime test files.
- No dependency or lockfile change: `package.json` and `pnpm-lock.yaml` are identical to `main`.

https://claude.ai/code/session_01Da5txwu3F3FBMQQhFzGAfG


agents/superpowers/specs/stryker-mutation-testing-evaluation.md
```diff
@@ -0,0 +1,86 @@
...
+## Summary
+
+StrykerJS 10 with `@stryker-mutator/vitest-runner` **works** against this repo's
+vitest 4.1.10 node projects, with three caveats (below). Measured mutation scores are
+low enough to be informative: the surviving mutants point at real assertion gaps rather
+than at tool noise.
+
+| Package           | Scope measured                                            | Mutants | Score                        | Wall clock (4 cores)            |
+| ----------------- | --------------------------------------------------------- | ------- | ---------------------------- | ------------------------------- |
+| `compute-runtime` | whole `src` minus tests/`testing`/`LayerStack.ts`         | 2837    | **42.5%** (59.0% of covered) | 43m11s                          |
```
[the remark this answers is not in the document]
dmaretskyi: Valid — fixed in b68d742e. Both numbers came from the same run: 2837 was the instrumented count and 2759 the scored count. The 78-mutant gap is the static mutants excluded by `ignoreStatic: true`, which the clear-text reporter doesn't show as a column. The table now reports 2759 (matching the score), and the breakdown states both figures and names the gap.

---

agents/superpowers/specs/stryker-mutation-testing-evaluation.md
```diff
@@ -0,0 +1,96 @@
...
+   survivors that dominate otherwise-good files.
+4. Report the `x!++` crash upstream.
+
+## Reproducing
+
+The Stryker packages are deliberately NOT in the workspace's `package.json`: adding them at the
+root made pnpm rewrite peer-dependency keys across the whole lockfile (~1600 lines), which CI's
+frozen install rejects with `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY`. For an on-demand audit tool
+that is not worth a permanent lockfile change — install it for the session and discard the
+lockfile edit afterwards:
```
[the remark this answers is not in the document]
dmaretskyi: Correct — fixed in a11ddfeb. The text said only "discard the lockfile edit"; it now says to revert what the command wrote to both the workspace-root `package.json` and the shared `pnpm-lock.yaml`, keeping any unrelated edits.

---

agents/superpowers/specs/stryker-mutation-testing-evaluation.md
```diff
@@ -0,0 +1,116 @@
...
+The survivor list was then used to strengthen three existing `compute-runtime` suites (tests only,
+no source changes): 164 → 178 passing tests. Re-running Stryker over just those four source files
+gives the before/after, both runs with `ignoreStatic`:
+
+| File                    | Before | After     | Delta |
+| ----------------------- | ------ | --------- | ----- |
+| `process-store.ts`      | 77.3%  | **93.2%** | +15.9 |
+| `ProcessManager.ts`     | 45.0%  | **53.5%** | +8.5  |
+| `ProcessHandle.ts`      | 39.6%  | **45.3%** | +5.7  |
+| `trigger-dispatcher.ts` | 56.6%  | **58.1%** | +1.4  |
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 8cb24505, though the other way round: the delta was right and the displayed scores were lossy. Stryker reported 56.62% → 58.05%, i.e. +1.43 → +1.4; rounding the scores to one decimal first is what made it look like +1.5. The table now shows all four rows at the reporter's own precision, so the arithmetic checks out as displayed.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
```diff
@@ -287,6 +287,90 @@ describe('TriggerDispatcher', () => {
         expect(result).toEqual(Exit.succeed({ tick: 0 }));
       }, Effect.provide(TestLayer())),
     );
+
+    it.effect(
+      'a timer event carries no feed cursor',
```
dmaretskyi: remove

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
```diff
@@ -287,6 +287,90 @@ describe('TriggerDispatcher', () => {
...
+          trigger,
+          event: { data: {} } satisfies TriggerEvent.DirectEvent,
+        });
+
+        expect(Exit.isFailure(result)).toBe(true);
+      }, Effect.provide(TestLayer())),
+    );
+
+    it.effect(
+      'tracks at most the most recent invocations',
```
dmaretskyi: remove

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
```diff
@@ -489,8 +573,15 @@ describe('TriggerDispatcher', () => {
           results = yield* dispatcher.invokeScheduledTriggers({ kinds: ['timer'] });
           expect(results.length).toBe(0);
 
-          // Past cooldown -- runs again (and fails again).
-          yield* dispatcher.advanceTime(Duration.minutes(4));
+          // Exactly at the cooldown boundary -- the cooldown is treated as elapsed (`<=`, not `<`),
+          // so the trigger fires rather than waiting one more tick.
+          yield* dispatcher.advanceTime(Duration.minutes(3));
```
dmaretskyi: we dont need to be that precise

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
dmaretskyi: Removed in c529794a.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
dmaretskyi: Removed in c529794a.

---

packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts
dmaretskyi: Reverted in c529794a — the exact-boundary step is gone and the test is back to main's coarser stepping (within cooldown → skipped, then +4 min → runs again).

---

PR #12857, opened 2026-08-31
codec: extract the transport vocabulary and derive the compat shape from buf types

dmaretskyi:
Two structural prerequisites for deleting `@dxos/codec-protobuf` and `@dxos/protobuf-compiler`. Both are about the **type** surface: migrating runtime call sites was never going to make either package deletable.

## 1. The transport vocabulary → `@dxos/codec`

Of the 24 source files outside `codec-protobuf` that imported it, most imported only `type`s, and those split two ways:

| Group | Symbols | Fate |
| --- | --- | --- |
| **Transport vocabulary** — describes the wire, no protobuf.js in it | `Any`, `Codec`, `EncodingOptions`, `RequestOptions`, `Struct`, `WithTypeUrl`, `TaggedType`, `TypedProtoMessage`, `ServiceBackend`, `ServiceProvider`, `ServiceDescriptorLike` | **must move** — nothing retires them |
| protobuf.js machinery | `compressSchema`, the three `*Substitutions`, the `Codec`/schema/mapping/sanitizer implementations | deleted with the package |

`hypercore/src/crypto.ts` and `feed-store` want a thing with `encode`/`decode`; `messaging` and `blade-runner` want the `Any` envelope; `client-services`&#39; `pipeline/codec.ts` wants `Codec`. **None is a protobuf.js consumer**, so no migration thread ever removes their import — the symbols had to leave instead, or the package stays in the graph forever.

It cannot live in `@dxos/protocols`: `hypercore` and `feed-store` are `common/` packages that do not depend on it, so routing them through `core/` would invert the layering. Hence a `common/`-level types-only package that `codec-protobuf` implements rather than defines.

**Direct dependents of `codec-protobuf`: 11 → 2** — `protobuf-compiler` (`compressSchema`) and `protocols` (the `*Substitutions`), both genuine protobuf.js machinery that dies with the migration. Each thread landing after this imports the vocabulary from its new home, so the dependent list shrinks monotonically instead of being re-established by every new call site.

The split was cleaner than expected: 21 of the 24 files imported **only** moved symbols, so it was a straight module-path swap. No re-export shims — every call site moves in this change.

### The generator was emitting it too

`file-generator.ts` emits `import type { RequestOptions } from …` into every generated service stub, and `CODEC_MODULE` was serving both that and the serializer machinery. It splits: `CODEC_TYPES_MODULE` (`@dxos/codec`) for the vocabulary, `CODEC_MODULE` (`@dxos/codec-protobuf`) for what stays on protobuf.js. Every stub generated from now on points at the new home by itself, which is what makes the monotonic property hold without vigilance. Found by the full build — the seven checked-in fixtures under `test/proto/gen` broke.

## 2. `Compat<T>` — the compat shape, derived from buf

The vocabulary was only half of it. `decodeCompat<V>`/`compatCodec<T>` take the protobuf.js-shaped type as a **caller-supplied type parameter**, and the only thing that could name that type was the protobuf.js generated barrel. So all ~20 compat-codec call sites still imported `@dxos/protocols/proto/*` purely for a type, and `protobuf-compiler`'s output could not be deleted however much runtime had moved.

`Compat<T>` (`protocols/src/buf/compat-types.ts`, exported from `@dxos/protocols/buf-shape-compat`) derives it from the buf generated type instead, mirroring `shape-compat.ts` rule for rule:

- the five substituted messages → `PublicKey` / `Buffer` / `Timeframe` / `Date` / `Struct`
- `google.protobuf.Any` → a `@type`-tagged bag; `JsonObject` → `Struct`
- oneof groups flattened from buf's `{ case, value }` into protobuf.js's sibling optional fields

`compat-types.test.ts` pins it by assigning a legacy-typed value into `Compat<Buf>` across eight shapes: a credential, a presentation, a oneof group, a private key, a timeframe, a `Struct`+`Any` message, a nested message with enums, and a map field. Those assertions compile against the protobuf.js generated types and are deleted along with them.

### It widens presence, and that reprices the remaining threads

Buf cannot recover proto3's `optional` marker from a generated type, so every singular message field arrives optional even where protobuf.js declared it required (repeated and map fields go the other way — buf always materialises the empty collection, so `Compat` marks them optional to match). Legacy-typed values are assignable to `Compat`, not the reverse.

`#5` and `#9d` were filed as import sweeps. What is actually left is **presence handling** at each consumer, and it cannot be discharged with `!` under the repo's no-cast rule. `echo-host`'s `sqlite-heads-store.ts` is converted here as the worked example — it already read `.hashes ?? []`, so it cost one import line. A site that indexes a required submessage costs a real branch, and packages whose public API re-exports a generated type (`keyring`'s `KeyRecord[]`) propagate the widening to their own consumers — which is why the sweep wants one package per change, not one repo-wide commit. Recorded in the audit.

## ⚠️ Blocked on a human before this can go green

The `@dxos/codec` name must be published and npm trusted publishing configured **before** this lands. Both CI package checks fail otherwise, and I verified both halves:

```
private → check-public-dependencies:  10 published packages depend on a private one
public  → check-packages-published:   @dxos/codec has never been published to npm
```

Someone with npm access needs to run `node scripts/publish-package.mjs codec` and set up OIDC trusted publishing for it. The package is left public here, which is the correct end state. This is the same trap the `blob` project recorded in the registry.

I reproduced all nine steps of the `check` job locally; `check-packages-published` is the only one that fails.

## Also

- Registers the migration as a local project (`.agents/projects/protobufjs-to-buf`) with the traps worth carrying: the `Buffer`-view rule, `vitest` not typechecking, and this packaging trap.
- Marks `#8` **done** in the audit — it merged in #12833 but its status row still read `todo`.

## Verification

| | Result |
| --- | --- |
| `moon exec :build` (full repo) | 441 tasks, exit 0, **zero type errors** |
| Tests, every affected package | 81 tasks, exit 0 — echo-client 530, client-services 180, credentials 66, rpc 53, protocols 52, feed-store 24, messaging 20, hypercore 12, gossip 9, codec-protobuf 5, client-protocol 4 |
| `Compat` commit | `protocols` + `echo-host` build, test and lint clean; `keyring` and `client-services` rebuilt clean |
| Lint | clean |
| `pnpm format-check` | clean over 13,886 files |
| Cast/non-null audit on the diff | clean |

`credentials` is included deliberately: it is the byte-identity-sensitive suite, and `Struct`/`Any` moving types is exactly the kind of change that could perturb it.

---


dmaretskyi:
## `test (shard=3)` — what I established, and where I'm stuck

`Check / test (node-version=24.11.1, shard=3)` has now failed on three heads: `d7c1be50`, `05a6e457`, and `08b23e4d` (the last on a freshly merged base, `74f9b30a`). The other three shards, `memory`, `boot-budget` and `model-fixture` are green; `check` fails for the separate, already-documented `@dxos/codec` publish blocker.

**I cannot read the failure.** CI is Depot-orchestrated: the check run's `output` is empty, there is no GitHub Actions run id, and `get_job_logs` on the check id returns 404. So I have no failing test name from CI itself — everything below is from local reproduction, and I want to be explicit that it is not the CI failure, only the nearest thing I can obtain.

### Local reproduction

Running the shard's exact command (`moon exec :test :test-browser :test-storybook :test-workerd --affected remote --job 3 --job-total 4 -- --no-file-parallelism`) surfaced four storybook failures, none in a package this PR touches:

| Suite | Failure |
| --- | --- |
| `plugin-search` | `SearchDialog.stories.tsx > Test` — interaction timed out in 15000ms |
| `plugin-inbox` | `Header.stories.tsx > Spec` — `Unable to find role="button" and name "system-button.unstar.label"` |
| `plugin-terra` | `TerraArticle.stories.tsx > Objects` |
| `plugin-file` | `FileArticle.stories.tsx > Image` |

**All four pass in isolation on this branch** — `plugin-search` 5/5, `plugin-terra` 10/10, `plugin-file` 16/16, `plugin-inbox` 51/51, exit 0. They only fail when 20+ storybook suites run concurrently, which on a 4-core sandbox is contention, so my local failures are an artefact of the box rather than evidence about CI. (`plugin-terra` and `plugin-file` are the same two suites from #12833, where `plugin-terra` was confirmed unrelated and `plugin-file` then passed 4× with that change.)

### Why I don't believe it's this PR, and the check that could have proved otherwise

The one mechanism by which a type-only change could break a test run is a stray runtime import — if the `export { type Compat }` re-export were emitted rather than elided, `@dxos/protocols/buf-shape-compat` would start pulling `@dxos/codec` into browser bundles. It is elided: the built `dist/src/buf/shape-compat.js` contains no reference to `compat-types` or `@dxos/codec`. Shard 3 also failed on `d7c1be50`, before `Compat` existed at all.

What this PR does do is widen the affected set enormously, because it touches `@dxos/protocols`. That is why shard 3 draws every plugin storybook suite here while a narrow PR like #12856 is green on the same shard — its affected set never includes them. So the honest reading is that this PR *surfaces* pre-existing storybook fragility rather than causing it, but I cannot close that out from the outside.

### What's blocking

I have no means to re-run the failed check (no Actions run id, so no `rerun_failed_jobs`) and no way to read its log. Merging the current base in was the closest available action and did not change the outcome. What would settle it, from someone with Depot access:

1. The failing test name from the shard-3 log. If it is one of the four above, `trunk-quarantine` is the route; if it is something in `protocols`/`echo-host`, it is mine and I'll fix it.
2. Whether shard 3 is red on `main` runs, where the shard partitions the full test set rather than an affected subset.

I'm not disabling or quarantining anything to get green, and I'm keeping this PR watched until it is.

---

dmaretskyi:
Closing in favour of a different approach, at @dmaretskyi's direction.

This PR made `codec-protobuf` deletable by **re-homing** its type surface: extracting the transport vocabulary into a new `@dxos/codec` package, and adding `Compat<T>` so call sites could name the protobuf.js-shaped view without the generated barrel. Both preserved the compat layer and kept consumer code unchanged.

The direction instead is to **change the consumers**: depend on buf types directly, adjust upstream code to buf's plain message shapes, and let the `codec-protobuf` dependency disappear by subtraction — it dies with protobuf.js, so nothing of it needs a new home. That also removes this PR's blocker outright, since `@dxos/codec` was the package that needed publishing.

Two findings from this work carry over and are worth keeping on the record:

- **`Compat` established the shape delta precisely.** Buf cannot recover proto3's `optional` marker from a generated type, so every singular message field is optional where protobuf.js declared it required, while repeated and map fields go the other way. That delta is exactly the upstream adjustment the new direction asks for, now made explicitly at each call site rather than papered over by a type.
- **The generator emitted the vocabulary too.** `file-generator.ts` puts `import type { RequestOptions }` into every generated service stub, so anything left pointing at `codec-protobuf` is re-established by every new stub. Whatever replaces it has to fix the generator, not just the call sites.

Superseded by the PR that follows on the same branch. The plan in `docs/audits/protobufjs-to-buf.md` is being rewritten around the new direction.

---

PR #12883, opened 2026-08-31
observability: capture AI generations to PostHog LLM analytics

wittjosiah:
Forwards the OpenTelemetry GenAI spans our AI stack already emits to PostHog as
`$ai_generation` events, so the LLM analytics product shows traces, token counts,
cost, and latency for every model call Composer makes.

## Why this shape

The effect AI provider layers annotate each model-call span with the OTel GenAI
semantic conventions (`gen_ai.system`, `gen_ai.request.model`, `gen_ai.usage.*`),
and PostHog ingests that same taxonomy. So the data mostly exists already — it
just had nowhere to go, because nothing provides an Effect `Tracer` in
production. PostHog's own SDK wrappers wrap the OpenAI/Anthropic clients and do
not apply to `@effect/ai`, so this maps the spans instead of adopting a vendor
SDK.

```
effect-ai spans (gen_ai.* attrs)
  → RuntimeServices           Tracer + span transformer, installed once   @dxos/app-framework
  → AiSpanProcessor           gen_ai.* → $ai_generation, policy + scrub   @dxos/observability
  → observability.events      PostHog capture                            @dxos/plugin-observability
```

The backend installs at the app level rather than through a pluggable seam.
Every conversation happens in a space and with a model, so the harness should do
this on its own: a plugin contributes an Effect layer to the new
`Capabilities.RuntimeServices`, the process manager merges it into the runtime's
base layer, and every fiber on that runtime inherits it. Call sites annotate
spans unconditionally and know nothing about any of it. `plugin.test.ts` proves
the reach — it contributes a tracer only through the capability and asserts the
`LanguageModel.*` spans arrive with the right space.

Emitting through the existing observability extension means AI capture inherits
the telemetry opt-in, `identify()` (HALO DID), release/environment registration,
and worker stubbing rather than initialising a second PostHog client.

## Capture policy

| Observability toggle | Space | Capture |
| --- | --- | --- |
| off | any | nothing |
| on | EDGE has plaintext access (every space today) | content: prompts, responses, tool names |
| on | E2E-encrypted (future) | metadata only |
| on | not declared by the call site | metadata only |

Content that already leaves the device in plaintext for EDGE to replicate is not
newly exposed in kind by telemetry. An E2E-encrypted space promises that
plaintext never reaches infrastructure, and telemetry must not become the side
channel that breaks it. `contentCaptureAllowed()` in
`packages/plugins/plugin-observability/src/capabilities/ai-observability.ts` is
the single decision point. It returns `true` unconditionally today and carries a
comment naming what it must do once E2E spaces exist — gate at the data
boundary, not the conversation's home space, since a turn reading through a
cross-space reference would otherwise leak content into a captured trace.

The last row is the fail-closed default, and it is enforced structurally: the
processor never consults the predicate for a span with no space, so content
capture is opt-in per call site. `AiSession` declares its space; the utility
model calls behind summarization, tagging and extraction do not, and report
metadata only until they do.

Reviewers may want to weigh one thing I flagged while designing this: EDGE and
PostHog are different trust boundaries. EDGE is ours; PostHog is a third-party
processor whose product makes captured content browsable by anyone with project
access, and captured content includes tool results, so slices of space data land
there. Two cheap mitigations, neither in this PR: update the privacy-notice copy
so the toggle's consent language covers conversation content, and set the
PostHog project's retention deliberately.

Independent of tier, the processor reports nothing while the telemetry opt-in is
off — checked against DXOS's own preference rather than the PostHog client's
separate, API-key-keyed opt-out flag — forwards an **allowlist** of attributes,
and reduces span errors to their **exception class name**, since provider error
messages and stack traces can embed request or response fragments. There is a
test asserting a secret in an exception message never reaches the event.

Neither side of the seam can fail the call it observes. Effect invokes the span
transformer on the model call's own fiber with no error handling, and OTel's
`MultiSpanProcessor` does not catch `onEnd` — which runs inside `Span.end()` on
that same fiber. Tool results are arbitrary values, so a cycle or a `BigInt` in
one was enough to fail the call. Each content attribute is now built and stamped
independently, and the sink's whole body is guarded.

## Changes

- **`@dxos/ai`** — new `AiTelemetry`: `makeContentSpanTransformer` stamps prompt,
  response, and tool names onto the model-call span (effect's own hook for
  observability vendors that want message content, since content is deliberately
  absent from `gen_ai.*`). `ATTRIBUTES` is the name contract the sink restates.
- **`@dxos/observability`** — new `AiObservability`: `AiSpanProcessor` maps spans
  to PostHog properties and applies the policy and scrub rules;
  `createAiTracerProvider` bundles it into a standalone provider, deliberately
  not the global one, so the SigNoz exporter and AI capture cannot clobber each
  other. It samples model calls only, so backing an app-wide tracer with it costs
  nothing for the rest of the app's spans.
- **`@dxos/app-framework`** — new `Capabilities.RuntimeServices`; the process
  manager merges contributions into the runtime's base layer, and logs a late
  contribution the way it already did for `LayerSpec`.
- **`plugin-observability`** — new `AiObservability` module holding the policy and
  wiring the halves together. It resolves the `Observability` capability per span
  rather than requiring it, keeping observability initialization (which awaits a
  network fetch) off the path to the process-manager runtime.
- **`ChatCompletionsAdapter`** — annotates spans so Ollama and LM Studio report
  model and token usage like the Anthropic path already did, with `provider`
  separated from the wire dialect so LM Studio is not reported as OpenAI.
- **`AiSession`** — annotates request spans with the conversation feed URI, which
  becomes `$ai_session_id` so PostHog groups a conversation's turns, and the
  space id the policy reads.
- **`@dxos/effect`** — `makeTracer`, an Effect tracer over an explicit OTel
  provider (`layerOtel` only reads the global one).

## Testing

`AiTelemetry.test.ts` covers tracer injection, provider annotations landing as
OTel attributes, truncation and its marker, and an unserializable tool result
costing its own attribute rather than the call. `ai-span-processor.test.ts`
covers property mapping, the opt-in gate, deny-unknown, error scrubbing, and a
throwing capture callback. `ai-observability.test.ts` in plugin-observability —
the one package depending on both halves — drives transformer through processor
to captured event, including the deny branches, and pins that span annotations
survive a parent the sampler dropped.

Green across the touched packages: `app-framework` (266), `plugin-assistant`
(181), `observability` (113), `ai` (98), `assistant` (65),
`plugin-observability` (10), plus builds, lints, repo-wide format, and the boot
budget (22 preload entries / 4.18 MB against 25 / 4.45 MB).

## Not in this PR

Cloudflare functions runtime capture, the Claude Agent SDK path (which bypasses
`AiService` entirely), `$ai_trace_id` on the feedback survey so a complaint links
to the trace that caused it, extra segmentation properties (skill names,
delegation, turn index), and the PostHog-side evaluations setup.

Space attribution for the ~30 model call sites outside `AiSession` — summarize,
tag, extract, chat-name — which would mean annotating in the operation invoker,
where the space is already known. Until then those report metadata only, which is
the safe direction but not the complete one.


packages/core/compute/ai/src/AiTelemetry.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/** Collapses response parts (final or streamed) into a single assistant choice. */
+const serializeResponse = (response: ReadonlyArray<{ readonly type: string }>): unknown[] => {
+  let text = '';
+  let reasoning = '';
+  const content: unknown[] = [];
+  for (const part of response) {
+    switch (part.type) {
+      case 'text':
+        text += field(part, 'text');
```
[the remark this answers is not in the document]
wittjosiah: Narrowed via a `textField()` accessor in fd7d2b29b3 — worth doing so a missing or non-string field can't stringify into captured content. The stated reason was inaccurate though: `string += unknown` is legal TypeScript, and the module type-checks (`tsc --noEmit -p tsconfig.json` exits 0 with the file in the program).

packages/core/compute/ai/src/resolvers/ChatCompletionsAdapter.ts
```diff
@@ -922,6 +926,11 @@ export const make = (model: string) =>
                       }
                       openAiCalls.clear();
 
+                      annotateResponse(options.span, {
+                        inputTokens: parsed.inputTokens,
+                        outputTokens: parsed.outputTokens,
+                        finishReason: parsed.finishReason ?? 'stop',
+                      });
```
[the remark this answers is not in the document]
wittjosiah: Accurate, but pre-existing and out of scope here, so not fixed in this PR. `parseStreamChunk` returns token counts only for the `ollama` branch; on `main` the `finish` part already emits `usage.inputTokens.total = parsed.inputTokens` (undefined for openai-format streams), and this PR's `annotateResponse` mirrors that same value rather than introducing the gap. The fix you describe (`stream_options.include_usage` plus retaining the usage-only chunk) changes the request payload for every OpenAI-compatible provider including LM Studio, so it deserves its own change and its own testing against a real server. Leaving this thread open for that follow-up.

packages/core/compute/compute/src/Process.ts
```diff
@@ -135,11 +135,12 @@ export interface ProcessContext<I, O> {
   submitOutput(output: O): void;
 
   /**
-   * Set an alarm for the process to be woken up later.
+   * Set an alarm for the process to be woken up later. The timer is forked from the calling fiber,
+   * so `onAlarm` runs with the caller's context.
```
dmaretskyi: this shouldn't be the case, alarms shouldn't inherit the context of whoever called `setAlarm`, otherwise you'll get very deep stacks of processes calling setAlarm recursively

wittjosiah: Agreed, and it was worse than deep stacks: the parent span lives in the fiber context, and the agent schedules each turn from inside the previous one, so every turn's spans nested under the last turn's and stayed reachable through the chain for the life of the conversation. Fixed in 22b91a8963: the manager captures the process's own context once at spawn/hydrate (minus whatever span it was spawned under), and the alarm and child-event forks replace the caller's context with it via `Effect.updateContext`. Each alarm is a fresh entry point with the runtime's services, clock, and tracer. `forkIn(scope)` stays for lifetime. There's a test that an alarm scheduled from inside a span starts a fresh root, and the doc line you flagged is gone.

wittjosiah: Fixed in 22b91a8963: alarms and child events now run with the process's own context, captured at spawn/hydrate without the spawning span, so a turn scheduled from inside the previous one starts a fresh root; the doc line is gone and a test pins it.

PR #12893, opened 2026-09-01
repo: migrate relative imports to explicit .ts/.tsx extensions

dmaretskyi:
## Summary

- Rewrites every relative `import`/`export`/dynamic-`import()`/`import(...)`-type/`new URL()`/`vi.mock` specifier repo-wide to carry an explicit filesystem-accurate extension (`./foo.ts`, `./dir/index.tsx`), matching `tsconfig.base.json`'s `rewriteRelativeImportExtensions`. 7118 files changed, 0 unresolved specifiers, idempotent on re-run.
- Adds `scripts/migrate-esm-imports.ts`, the AST-based codemod that did the rewrite (kept as a repo utility; needs `TS_PARSER_PATH` pointed at a classic-TypeScript install, since the workspace's `typescript` catalog entry is TS 7/tsgo, which no longer exposes the classic synchronous parser API).
- Fixes a real bug this migration exposed at scale in three custom oxlint rules (`dxos-subpath-exports`, `dxos-package-imports`, `import-as-namespace` in `packages/common/eslint-plugin-rules/rules/`): their `resolveModule()` always appended an extension to a specifier, so an already-extensioned one (e.g. `./types/index.ts`) resolved to a nonexistent `index.ts.ts` and silently failed. Fixed all three with an early-return when the specifier already carries a known extension.

## Validation

- `moon exec --on-failure continue --quiet :build` across the full repo: clean except one **pre-existing, unrelated** failure in `packages/common/protobuf-compiler/test/types.test.ts` (a `Timestamp`/`Any` protobuf type mismatch) — confirmed pre-existing by reverting that file to its committed version and re-running the same build task, which fails identically.
- `moon exec --on-failure continue --quiet :lint` across all 340 workspace packages: clean after the three rule fixes above (verified `compute:lint`, which surfaced the bug, now passes). One remaining test failure in `eslint-plugin-rules` itself confirmed pre-existing the same way. One non-blocking informational warning left (its suggested-fix message text still shows the old extension-less style — cosmetic, out of scope here).
- Spot-checked `vi.mock('./relative', ...)` call sites (3 across the repo) after migration — all their tests still pass, confirming mock resolution is extension-agnostic.
- Ran the two packages with build-time codegen into `.ts` sources (`protocols`, `react-ui-gameboard`) before the final pass so the codemod could resolve against real generated files rather than flagging them as unresolved.

https://claude.ai/code/session_01VwVJf3BpifWGseYai3woP4

---
_Generated by [Claude Code](https://claude.ai/code/session_01VwVJf3BpifWGseYai3woP4)_

---
🚀 Composer preview: https://pr-12893-composer-dev.dxos.workers.dev


dmaretskyi:
CI failures on this head (`build`, `Check / memory`, `Check / boot-budget`, `Model Fixture / model-fixture`) all trace back to the same root cause, and it is **not this PR's**: `packages/common/protobuf-compiler/test/types.test.ts` fails `tsc` with a pre-existing type mismatch —

```
error TS2353: Object literal may only specify known properties, and ''@type'' does not exist in type 'Any'.
error TS2739: Type 'Date' is missing the following properties from type 'Timestamp': seconds, nanos
```

This blocks `protobuf-compiler:build`, which most of the graph depends on, so it cascades into the jobs above.

Confirmed pre-existing: I reverted that one file to its committed `main` version (no other changes) and re-ran the same build task locally — it fails identically. This PR doesn't touch that file's type-affecting lines (only adds `.ts` extensions to its own imports, reformatted by the same pass).

No fix exists yet for the underlying `Any`/`Timestamp` type drift, so I haven't ported one. Will keep watching for either a fix landing on `main` (to merge in) or further guidance.

---

dmaretskyi:
`Check / check` and `Model Fixture` are now green (commit 33e50e9) — the remaining red is `Check / test (shard=1)`.

I can't fetch Depot's job logs from here (no `job_id`/`run_id` — the `Check` workflow isn't registered as a native GitHub Actions workflow, it's Depot-managed, so I have no way to trigger a re-run of just this job either). Root-caused it instead by running the full repo test suite locally (`moon run :test`) twice, clean, against this exact commit's tree. Both runs turn up exactly one failure, always the same one:

```
echo-host:test | FAIL src/automerge/automerge-repo-subduction-policy.test.ts >
  SubductionPolicy > authorizeFetch + filterAuthorizedFetch invocation conditions >
  authorizeFetch fires on BOTH proactive push and explicit fetch (empirical)
```

This is pre-existing and unrelated to this PR:
- `git log` on that file shows only this branch's own extension-only migration commit touching it — no logic changes.
- The file already documents itself as flaky: `// TODO(mykola): subduction wasm/network tests are flaky on CI runners`, and the failing assertion's own comment acknowledges timing variance ("the bridge may batch or invoke twice per broadcast").
- Re-running just that test in isolation (2 runs) passed both times, with a `subduction_core` `WARN ... request timed out` log line in each — consistent with the file's own documented network/timing flakiness rather than a real regression.

Nothing in this PR touches that file's logic (see the earlier comment on the substitutions-parser/knip/plugin-set fixes — same "extension-only" scope applies here). No fix exists to port; this is out of scope for an import-extension migration. Will keep watching for a green shard=1 on the next push (this branch is still catching up to `main`, which has moved several more times during this PR).

---

dmaretskyi:
Correction to my earlier comment: the `echo-host` test I named there is wrong — that suite is gated `describe.skipIf(process.env.CI)` and never runs on Depot at all, so it can't be shard=1's failure. I'd reproduced it with `CI` unset locally, which isn't representative.

Redid the investigation with `CI=true` set to match the real environment. Two things turned up, neither traceable to this PR (both files' `git log` shows only this branch's own extension-only touch, no logic changes, and several are untouched by this branch at all):

1. `packages/devtools/cli`'s test suite (~23 failures across `chat`, `mcp`, `plugin-install`, `plugin`, `reset`, `space-list`) — these spawn real `bun`-run `dx` subprocesses, and I can't tell whether this is a real signal or an artifact of this sandbox's process/resource limits under a full concurrent `:test` sweep. Notably `Check / cli` is its own, separately-skipped job in this PR's CI, so I don't think this suite is even part of the sharded `:test` matrix `Check / test` runs.
2. `client-e2e`'s `spaces.test.ts > post and listen to messages` — a genuine intermittent failure (fails ~1/3 of isolated reruns). Looked into it: it's not a timeout-budget problem (raising the internal wait from 200ms to 5s still failed at ~1/3 the same rate, with failures either resolving in ~1s or not resolving at all) — it looks like a subscribe-then-publish race between `space2.listen(...)` and the immediately-following `space1.postMessage(...)`, unrelated to timing. I reverted my timeout bump since it didn't fix anything and touches code this PR has no reason to touch.

I don't have a clean way to confirm which (if either) of these is Depot's actual shard=1 failure — no job-log access, and my local full-sweep environment clearly has different resource characteristics (process limits, concurrency) than Depot's runners. What I can say confidently: nothing in this PR's diff touches either failing area's logic, and every previously-identified real regression (protobuf-compiler, knip, composer-app plugin-set, the two oxlint bugs) is fixed and verified — `Check / check` and `Model Fixture` are green. Continuing to watch; will dig further if shard=1 stays red on a build that isn't resource-starved the way my sandbox is.

https://claude.ai/code/session_01VwVJf3BpifWGseYai3woP4

---

dmaretskyi:
Found the actual root cause (superseding my last two comments, which were both wrong guesses).

`packages/devtools/cli`'s test suite has ~23 failures, and every one traces back to a single bug in `dx-plugin gen` (`packages/sdk/app-framework/src/plugin-cli/ts-util.ts`'s `resolveRelativeModule`): it always appends a resolution extension to a relative specifier, so an **already-extensioned** one — exactly what this migration's own codemod produces (`export * from './x'` → `export * from './x/index.ts'`) — resolves to a nonexistent path (`index.ts.ts`) and the generator silently drops that capability from every generated per-environment barrel: neither the real export nor its `undefined` stub. `plugin-space`'s `NavigationHandler` and `plugin-connector`'s `Coordinator` both hit it (also present in `plugin-client` and `plugin-file-system`), and either one alone crashes the CLI's plugin manager at startup for **every** command (`LazyPluginError` / `SyntaxError: Export named 'X' not found`), since it eagerly resolves every plugin.

This is the same bug class as the three oxlint-rule fixes already in this PR (`resolveModule()` always appending an extension) — just a fourth instance, in a different tool, that I hadn't traced through yet. Fixed the same way: early-return when the specifier already carries a known extension.

Verified: `moon run cli:test` went from 23 failing to 0 failing; full-repo `:build`, `:lint`, and `:check-module-structure` are clean. This is very likely the actual, deterministic cause of `Check / test shard=1` staying red on every commit so far (unlike the two probabilistic flakes I chased and ruled out earlier) — it's the kind of failure that reproduces on every run, not intermittently, which fits the observed pattern much better.

https://claude.ai/code/session_01VwVJf3BpifWGseYai3woP4

---

PR #12901, opened 2026-09-02
devtools: move the last convertible protobuf enums to buf, correct #5's scope

dmaretskyi:
## Status: in progress — do not merge

Second pass on **#5**, layered onto the two already-landed commits. The PR's original framing
("2 converted, 15 blocked") was a correct diagnosis under the old mandate; the mandate then changed
to *fix* the blockers. This body tracks current state and is rewritten as work lands.

Commit order is deliberate: upstream/producer moves first, devtools conversions on top, so a
reviewer can see the panels only compile *because* the upstream moved.

---

## #5 — 6 of 15 declarations converted

**Read the carrier column, not the count.** Two of the three landed by moving a *carrier*, not by
editing a panel, and several remaining declarations share one. 12 declarations remain across **8
carrier groups**, so the work lands in jumps rather than one file at a time.

| # | File | Declaration | Carrier | Status |
| 5 | `hooks/useSignal.tsx` | `SignalResponse` | `DevtoolsHost.subscribeToSignal` | ✅ `bufMessage` |
| 15 | `panels/mesh/SignalPanel/SignalMessageTable.tsx` | `SignalResponse` | ↑ same carrier | ✅ |
| 14 | `panels/mesh/NetworkPanel/NetworkPanel.tsx` | `PeerState` | gossip `Presence` (producer) | ✅ buf at producer |
| 10 | `panels/echo/SpaceInfoPanel/PipelineTable.tsx` | `SubscribeToSpacesResponse` | `DevtoolsHost.subscribeToSpaces` | ✅ group A |
| 11 | `panels/echo/SpaceInfoPanel/SpaceProperties.tsx` | `SubscribeToSpacesResponse` | ↑ same carrier | ✅ group A |
| 3 | `hooks/useFeedMessages.tsx` | `SubscribeToFeedBlocksResponse` | `DevtoolsHost.ts:358` | ⬜ group B — embeds `FeedMessage` (#9c) |
| 8 | `panels/echo/FeedsPanel/FeedsPanel.tsx` | `SubscribeToFeedBlocksResponse` | ↑ same carrier | ⬜ group B |
| 4 | `hooks/useMetadata.tsx` | `SubscribeToMetadataResponse` | `DevtoolsHost.ts:363` | ⬜ group C — embeds `EchoMetadata` (#9c) |
| 1 | `components/performance/panels/EdgePanel.tsx` | `QueryEdgeStatusResponse`, `EdgeStatus` | `EdgeAgentService.ts:19` | ⬜ group D — **23 files** via `EdgeStatus` |
| 6 | `hooks/useStats.ts` | `QueryEdgeStatusResponse` | ↑ same carrier | ⬜ group D |
| 7 | `panels/client/LoggingPanel/LoggingPanel.tsx` | `LogEntry`, `QueryLogsRequest` | `LoggingService.queryLogs` | ✅ group E |
| 9 | `panels/echo/FeedsPanel/FeedsPanel.tsx` | `Contact` | `ContactsService.ts:18` | ⬜ group F — ~10 files, crosses `client.halo.contacts` |
| 12 | `panels/echo/SpaceInfoPanel/PipelineTable.tsx` | `Space.PipelineState` | `SpacesService.ts:202` | ⬜ group G — embeds `Credential` + `TimeframeVector` |
| 2 | `hooks/useCredentials.tsx` | `Credential` | `SpacesService.queryCredentials` | ⬜ group H — boundary decode (#9d out of scope) |
| 13 | `panels/halo/CredentialsPanel/CredentialsPanel.tsx` | `Credential` | ↑ same carrier | ⬜ group H |

Probes are tstyche type-tests (`moon run devtools:test-types`), not build output: each asserts
`$typeName`, the substituted-field shape, and that the protobuf.js type is **not** assignable to the
buf one — so a conversion passing only because two types are structurally interchangeable fails.
**8 tests / 14 assertions, EXIT=0.**

---

## Commits

**`a5bd17d081`** — devtools enum stragglers (`cli-util`, `plugin-debug`) + audit corrections.

**`e1a31022c3`** — `PeerState` to buf **at the producer** (#8.4). `preserveAny` on the gossip
extension was tried and **reverted**: it forces every channel to hand over a pre-packed `Any`,
breaking the app-facing `space.postMessage` path (`spaces-service.ts:237`) — caught by
`gossip-extension.test.ts`, not by argument. **Wire format unchanged.** Consumer ripple collapsed
rather than papered over: most sites were re-implementing `getPeersByIdentityKey`, which already
takes a domain `PublicKey`. Three genuine boundaries remain.

**`1699294b54`** — `SignalResponse` to `bufMessage`; 17 `swarmEvent` errors handled as
`{case, value}`. No casts; the pre-existing `response.swarmEvent!` assertions disappear because the
union narrows honestly. `Message.payload` is a real `Any` and **stays packed**. Wire unchanged.
Also fixes `useSignal`, whose `useMemo` returned void and shadowed its accumulator.

**`873a4ae3da`** — #2 fate-split + proto-guard pin derivation.

**`ea3965e9b8`** — cross-codec guard scope and its two divergences.

---

## Cross-codec guard — landed (`17465dce27`, corrected by `276ea8db5d`)

Walks the buf registry programmatically: **258 messages, 518 generated cases**, fully-populated
synthesised values. **483/517 → 518/518** once the real divergences were separated from three
generator artefacts.

**What it is not — read this before citing it.** Both codecs regenerate from the same `src/proto`
tree, so a wire-incompatible edit to a `.proto` moves them together and passes here unnoticed. It is
**not a downgrade or compatibility guard** and must not be described as one, or listed alongside
one. Catching a schema break needs bytes from a build predating the edit; the rejected
historical-pin approach and why it failed are in the audit doc. What this does give is continuous
agreement between the two codecs on every CI run. **It dies with protobuf.js at teardown.**

**A defect in the guard itself, found by `moon build` and not by vitest.** It reported 518/518 while
silently skipping **every repeated field**: the walk branched on `field.repeated`, which does not
exist on buf's descriptors — buf models repeated as `fieldKind: 'list'` with a `listKind`
discriminant, so every list field fell through to `default`. vitest was green because the type error
sat in a branch no generated case reached, and the case count is per message, not per field, so
nothing in the output distinguished a correct walk from a broken one. This is the clearest evidence
in the migration that a green test run is not proof.

**The ledger: 42 entries, all one root cause.** Correcting the walk took it 19 → 42 with **zero new
error classes** — protobuf.js's decoder materialising an *absent* singular message field with
unsubstituted defaults, which its own encoder then rejects, now reaching through repeated members
(`spaces`, `signatures`, `invitations`, `parents`, `credentials`, `contacts`). Not a tolerance
budget: each entry is keyed on message **and exact field paths**, so a listed message diverging
somewhere new goes red (verified by perturbation → exactly one failure). A no-growth assertion pins
the count; a staleness check fails a listed message that *stops* diverging, since a stale entry
would silence a future regression on a message that is currently clean.

`encodeCompat`/`decodeCompat` handled all 258 messages. Recorded beside the above, not instead of
it — the finding is that protobuf.js's decoder and its encoder disagree, not that the compat path
is clean.

---

## #9c — `EchoMetadata`: the codec has already moved; what remains is the exposed type

Both metadata stores are already on `compatCodec`, so the **bytes** are buf today. What `#9c` moves
is the **type** the store exposes.

**Scope, corrected against the tree.** `EchoMetadata` was scoped at 27 importers spanning
`echo-pipeline` and `echo-db`. Neither package exists in this tree — that code now lives under
`client-services/src/packlets/{space,pipeline}`, which is why the boundary sites
(`control-pipeline.ts:148,165`) are `client-services` paths. Enumerated at HEAD with
`grep -rl "@dxos/protocols/proto/dxos/echo/metadata" --include=*.ts --include=*.tsx packages/`:
**29 raw importers, of which 21 import only `EdgeReplicationSetting`** — a top-level enum that moves
independently and was already proven to (commit 1, `cli-util`/`plugin-debug`). The message-type
surface is **8 files, all in `packages/sdk/client-services`**.

Type move applied; `client-services:build` reports **151 errors** to work through, all in that one
package: 77 `PublicKey` (class vs message), 36 optionality, 24 `Timestamp`/`Date`, 12 `Timeframe`
vs `TimeframeVector`. In progress.

---

## #9c — route (b) holds; (c) off the table

`preserveAny` is **inapplicable** to all three `#9c` messages — the mechanism only fires for
`google.protobuf.Any` fields (`shape-compat.ts:206-215`), and these name `Credential` **directly**
(`feed.proto:37`, `metadata.proto:92`, `services.proto:170,173`).

Route (b) for `FeedMessage`: **9 sites**, none a logic change to credentials core.
`EchoMetadata` / `Space.PipelineState` counts still outstanding.

---

## #2 — split by fate; nothing left to migrate

`tools/protobuf-test` reads as a standalone fixture but its header says "Imported by
protobuf-compiler tests" and that is its only reference — it goes at teardown. The eight
`example/testing/rpc` importers are gated on **#8.2**, not #2. Full ledger in the audit doc.

---

## Verification

| Command | Result |
| --- | --- |
| `devtools:test-types` | EXIT=0 — 8 tests / 14 assertions |
| `devtools:build` / `client-services:build` / `client-protocol:build` | EXIT=0 |
| `teleport-extension-gossip:build` | EXIT=0 |
| `teleport-extension-gossip:test` | EXIT=0 — 9 passed, 1 skipped |
| `protocols:test` | EXIT=0 — **52 passed** |
| `client-services:test` | EXIT=0 — **182 passed, 3 skipped** |
| `oxfmt --check` | clean |

## Known unfixed

buf's `Timestamp.seconds` is a **`bigint`** and `JsonView` renders through a `JSON.stringify`
replacer, which throws on `BigInt`. `MetadataPanel` is the first panel to hit it. Lands with the
panel that needs it.

## Summary by CodeRabbit

* **New Features**
  * Improved mesh presence tracking and peer discovery by identity.
  * Enhanced signal monitoring with richer event, timestamp, and message details.
  * Added more reliable live signal updates and cleanup when leaving the view.
  * Improved display of peer identities in the network panel.

* **Bug Fixes**
  * Improved consistency when handling presence, peer-state, and signal data across the application.

* **Documentation**
  * Expanded protocol migration audit documentation with compatibility considerations and testing guidance.

---
🚀 Composer preview: https://pr-12901-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/devtools-buf-enum-stragglers.md
+++ b/.changeset/devtools-buf-enum-stragglers.md
@@ -1,0 +1,6 @@
+---
+'@dxos/cli-util': patch
+'@dxos/plugin-debug': patch
+---
+
+Move the last two top-level protobuf enum imports in devtools-owned packages onto buf: `EdgeReplicationSetting` in `cli-util`'s space helpers and `ConnectionState` in `plugin-debug`'s `DebugStatus`. Both enums are value-identical between the two generators and TypeScript relates enum types by name, so the swap needs no call-site change.
--- a/docs/audits/protobufjs-to-buf.md
+++ b/docs/audits/protobufjs-to-buf.md
@@ -243,6 +243,60 @@ an identically-valued `enum DifferentName` raises `TS2367`. So:
   `EdgePanel` cannot be fixed ahead of its type move, and it generalises to any nested enum
   elsewhere in the migration.
 
+## `#2`: split by fate before converting, and most of it dies at teardown
+
+`#2`'s nominal file list spans four places. Judged per file rather than per directory, almost none
+of it is worth converting -- the packages that own it are what the teardown slice deletes.
+
+**Dies at teardown -- leave on protobuf.js:**
+
+| File                                                                          | Why                                                                                                                                                                                                                     |
+| ----------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `codec-protobuf/test/codec.node.test.ts` + its three `example/testing` protos | Exercises `@dxos/codec-protobuf` itself.                                                                                                                                                                                |
+| `codec-protobuf/src/substitutions/struct.test.ts`                             | Tests the legacy substitution table.                                                                                                                                                                                    |
+| `protobuf-compiler/test/*` (12 files)                                         | Tests the legacy generator's output.                                                                                                                                                                                    |
+| `protobuf-compiler/src/namespaces.test.ts`                                    | Same.                                                                                                                                                                                                                   |
+| **`tools/protobuf-test` (whole package)**                                     | Reads as a standalone fixture and is not: its own header says "Imported by protobuf-compiler tests", and `protobuf-compiler/test/types.test.ts:19` is its only reference in the repo. It goes with `protobuf-compiler`. |
+
+**Gated on `#8.2`, not `#2`:** the `example/testing/rpc` importers -- `rpc/service.test.ts`,
+`rpc/service-type-url.test.ts`, `teleport/testing/test-extension.ts`,
+`teleport/testing/test-extension-with-streams.ts`, `teleport/muxing/muxer.test.ts`,
+`websocket-rpc/e2e.node.test.ts`, `rpc-tunnel-e2e/test-client.ts`, `rpc-tunnel-e2e/test-worker.ts`.
+These already take their _descriptor_ from `getBufService`, but still import the protobuf.js
+**service interface** as the type argument. Deriving that interface from buf's `GenService` is
+`#8.2`'s remaining work; the methods have to move to buf shapes at the same time, because
+`BufServiceDescriptor` still hands back compat-shaped values.
+
+**Nothing left to migrate.** `example/testing/{data,rpc}.proto` are already buf-generated
+(`buf/proto/gen/example/testing/`), `buf/registry.ts` already consumes them, and there are no
+message-only consumers to move. `#2`'s remaining value is the ledger above, not a conversion.
+
+## Proto-guard: the downgrade leg needs its own dependency closure
+
+The dated snapshot proves old bytes decode on new code. The reverse -- bytes this codec writes being
+readable by a build already in the field -- is the direction that strands a rollback, and it is not
+covered.
+
+**Pin: `@dxos/protocols@0.11.1`, published 2026-08-05.** Derived as the last stable release
+preceding the first buf-migration commit on main (`bdb02cd3a1`, "protocols: audit protobuf.js usage
+and start the buf migration", 2026-08-24). The only thing published in between is a
+`1.0.0-next-<sha>` CI snapshot on 2026-08-10, which is not a release. The codec itself did not move
+until `48eb05d613` (2026-08-26), so either reading of "first buf-migration commit" selects the same
+version. Write it as a literal, never resolve "latest before X" at runtime -- that drifts silently.
+
+**It cannot be imported in-process.** `@dxos/protocols@0.11.1` pulls `@dxos/keys@0.11.1`, which is
+built against `effect@3.x` and calls `Schema.filter`; against the workspace's `effect@4.0.0-rc.108`
+that is `TypeError: Schema.filter is not a function`, thrown at import time before any test body
+runs. Aliasing the package into the workspace does not work.
+
+**It does work in an isolated closure**, which is what a downgrade test should use anyway -- a build
+in the field has its own dependency graph. Verified: an `npm install @dxos/protocols@0.11.1` in a
+scratch directory resolves `effect@3.21.4` and its codec round-trips
+(`dxos.echo.query.Heads` -> `0a04613162320a0463336434`). The remaining work is the harness: install
+the pinned version into a gitignored fixture directory and spawn it to decode bytes the current
+codec wrote. Assert shape and round-trip, not byte identity -- protobuf.js writes `nanos: 0` where
+buf omits the proto3 default, which is wire-compatible and expected.
+
 ## `#8` is the last milestone-sized thread, and it is not a rider
 
 Everything above landed behind an interface that hides which codec carried a value, which is why it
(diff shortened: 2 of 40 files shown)


dmaretskyi:
Closing — the work here is now in #12860, which combines both PRs on current `main`.

At @dmaretskyi's direction the two were combined: they were slices of the same teardown, and keeping them apart meant two branches editing `docs/audits/protobufjs-to-buf.md` and two CI cycles for one migration.

Everything from this branch is carried over, not summarised — all 11 commits are in #12860's history:

- `a5bd17d081` devtools enum stragglers + audit corrections
- `e1a31022c3` `PeerState` to buf at the gossip producer (`#8.4`), with the `preserveAny` attempt reverted
- `1699294b54` `SignalResponse` to `bufMessage`, oneof handled as `{case, value}`
- `17465dce27` + `276ea8db5d` the cross-codec guard and its repeated-field correction
- `a485ef1234` `SubscribeToSpacesResponse` (carrier group A)
- `85133b84c9` `LogEntry` / `QueryLogsRequest` (carrier group E)
- the four audit commits (`#2` fate-split, guard scope, the repeated-field finding, `#9c` scope)

Two things changed in the process, both worth knowing:

1. **The two plan revisions were reconciled rather than one overwriting the other.** #12860's slice structure (A value codecs → B the `Any` envelope → C the RPC seam → D the barrel) is authoritative, and this branch's ordering constraint is preserved in it: `#5`'s remainder is **not** an independent sweep and must follow `#9c`/`#9d`. The status-table rows from here survive too, including `#5` as `part`.
2. **One lint failure appeared that this branch's own CI never saw.** `devtools/spaces.ts` had `SubscribeToSpacesResponseSchema` before `SubscribeToSpacesResponse_SpaceInfoSchema`; `perfectionist(sort-named-imports)` only flags it once the branch sits on current `main`. Fixed in `b7c4853e`.

Verified on the combined tree: full-repo build exit 0 with zero type errors, 134 test tasks exit 0 across 9 packages, the tstyche probes at 4 files / 16 tests / **31 assertions** (up from 8/14 here, since two more carrier groups landed), lint clean, `format-check` clean over 14,110 files.

The `Timestamp.seconds` `bigint` / `JsonView` issue is carried forward as known-unfixed, to land with the panel that needs it.

---

---

PR #12908, opened 2026-09-02
assistant: retry requests rejected for insufficient permissions

dmaretskyi:
Anthropic reports a key whose permissions have not yet propagated as an `InsufficientPermissions` authentication error:

```
AnthropicClient.createMessageStream: InsufficientPermissions: Your API key lacks required permissions
  at AiRequest.runAgentTurn (.../request/AiRequest.ts)
```

The same key succeeds moments later, but the failure killed the turn outright and the reader got no reply at all.

## Changes

- `AiRequest.runAgentTurn` retries the model request up to **10 times, spaced 2 seconds apart with jitter** (`Schedule.spaced(...).pipe(Schedule.jittered, Schedule.upTo({ times: 10 }))`).
- The retry is narrowed to `AuthenticationError` with `kind === 'InsufficientPermissions'`, read off the typed reason rather than matched in text. The other authentication kinds (missing, expired, invalid key) need a credential change and still surface immediately — `AuthenticationError.isRetryable` is `false` for all of them, which is why this is opted into explicitly.
- Retry is additionally gated on nothing having been emitted yet: an `emitted` flag is set when the first complete block of the attempt is submitted, after which the request is not re-issued — the messages are already in `_pending` and a second attempt would duplicate them. In practice the provider raises this at stream open, before any block, so the guard is a safety net rather than the normal path.
- Retries are logged at `warn` so a slow-propagating key is visible rather than silent.

## Testing

- `moon run assistant:build` — clean.
- `oxfmt --check` on the changed file — clean.
- `Check` workflow green on this head: `check`, all four `test` shards, `memory`, `boot-budget`, `changeset-reminder`, `build`.
- `Model Fixture` is red, and not this PR's: 4 `assistant-toolkit` planning-fixture tests fail identically with this PR's code change reverse-applied to `origin/main`. Details and the likely cause are in [a comment below](https://github.com/dxos/dxos/pull/12908#issuecomment-5512500761). That workflow is deliberately outside the required-checks set.

---

https://claude.ai/code/session_01Pz6bGSsMg7Bbt1G57Xo6Xm

### DIFF
--- a/.changeset/airequest-retry-insufficient-permissions.md
+++ b/.changeset/airequest-retry-insufficient-permissions.md
@@ -1,0 +1,5 @@
+---
+'@dxos/assistant': patch
+---
+
+Retry a model request the provider rejected with `InsufficientPermissions`. Anthropic returns this while a key's permissions are still propagating, and the failure previously killed the turn outright, leaving the reader with no reply. `AiRequest.runAgentTurn` now re-issues the request up to ten times, spaced two seconds apart with jitter, and only while no block of the turn has been emitted yet — so a retry can never duplicate content. The other authentication kinds (missing, expired, or invalid key) need a credential change and still surface immediately.
--- a/packages/core/compute/assistant/src/request/AiRequest.ts
+++ b/packages/core/compute/assistant/src/request/AiRequest.ts
@@ -9,6 +9,7 @@ import * as Effect from 'effect/Effect';
 import { pipe } from 'effect/Function';
 import * as Option from 'effect/Option';
 import * as Result from 'effect/Result';
+import * as Schedule from 'effect/Schedule';
 import * as Semaphore from 'effect/Semaphore';
 import * as Stream from 'effect/Stream';
 import * as AiError from 'effect/unstable/ai/AiError';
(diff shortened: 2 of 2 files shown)


dmaretskyi:
`Model Fixture / model-fixture` is red on this head, and it is not this PR's failure.

**What fails** — 4 tests in `assistant-toolkit`:

- `src/skills/planning/skill.test.ts` → *reminds the agent to continue while the checklist has open items*, *stays silent when the user asked for only part of the checklist*
- `src/skills/planning/operations/assign-tasks.model.test.ts` → *assigns tasks that already exist to the conversation*, *unassigns a task without deleting it*

All four are `No memoized conversation found for the given prompt`.

**Why it is not this PR's.** I reproduced it locally, then reverse-applied this PR's only code change (`git checkout origin/main -- packages/core/compute/assistant/src/request/AiRequest.ts`), rebuilt, and re-ran: the same 4 tests fail identically. The drift is in `assistant-toolkit`'s planning fixtures, which this diff does not touch — it adds a retry wrapper on an error path that no fixture exercises.

**No fix exists to port.** Regenerating the fixtures needs `DX_UPDATE_MODEL_FIXTURES=1` against a live model, and the saved-vs-new diff looks like a behavior change rather than prompt drift, so a blind regeneration would bake it in:

```diff
-"result": "Assigned 0 task(s), unassigned 1.\nCurrent checklist:\n  1. [ ] Renew the domain"
+"result": "Assigned 0 task(s), unassigned 0.\n\nCurrent checklist:\n  1. [ ] Renew the domain\n  2. [ ] Rotate the API keys"
```

The unassign is a no-op now and the removal DXN shifted (`<memoized-dynamic-2>` → `-3`), which points at the assign-tasks work in #12900 rather than at a stale recording. That belongs in its own change, not widened into this one.

**Re-run.** These tools cannot re-run it — the workflow runs on Depot and does not appear in the GitHub Actions API (job-log and re-run calls 404). A re-run would not tell us anything anyway: the failure is deterministic and reproduces locally on `origin/main`'s code.

`Model Fixture` is deliberately outside the required-checks set (see the comment at the top of `.depot/workflows/model-fixture.yml`), so it does not block merge. The required **Check** workflow is green on this head: `check`, all four `test` shards, `memory`, `boot-budget`, `changeset-reminder`, and `build`.

---

PR #12912, opened 2026-09-02
react-ui-list: derive a tree's drag scope from its path root

wittjosiah:
## What broke

A tree item's `treeId` is the scope a pragmatic-dnd monitor claims its own drags by. #12890 introduced it and set it to the tree's own `id`:

```ts
const renderContext = useMemo(() => ({ treeId: id, ... }))
```

That holds only when a monitor serves exactly one `Tree`. The navtree mounts one per workspace tab (`id={item.id}` in `L1Panel`), so every row's drag payload carried a per-tab id, the monitor watching `GraphNode.RootId` claimed nothing, and reordering a collection and dropping an object into one both became silent no-ops.

`e2e:matrix-0` and `e2e:matrix-1` have failed on every push to main since:

```
✘ collections.spec.ts:28:3 › Collection tests › re-order collections
✘ collections.spec.ts:47:3 › Collection tests › drag object into collection
```

Bisected against Depot: `05da6c793` green, `32584c984a` red, red on all six main pushes after it.

## The fix

The scope is the root of the tree's path, so trees sharing a path root are one drag scope:

```ts
const treeId = treePath[0];
```

Both existing monitors were already written against that value, so neither call site changes:

| consumer | `path` | `id` | scope | monitor watches |
|---|---|---|---|---|
| navtree | `['root']` | tab id | `'root'` | `GraphNode.RootId` |
| task list | — | `'tasks'` | `'tasks'` | `TASK_TREE_ROOT_ID` |

Deriving it rather than passing it also means the payload's `treeId` and `path` cannot disagree, which is the class of bug this was. An earlier revision of this PR added an explicit `treeId` prop plus a shared constant; deriving it needs neither, and leaves `plugin-navtree` untouched.

## Verification

Depot `Check` dispatched with `e2e=true` on this branch, since `check.yml` gates e2e off pull requests. Locally, `collections.spec.ts` is 5/5 against a production bundle, and `react-ui-list` (41) and `plugin-navtree` (2) unit tests pass.

## Still red on main, untouched here

- `basic.spec.ts:94 › Basic tests › logout` fails and Trunk quarantines it, which is why matrix-0 reports green.
- `Model Fixture` fails with `No memoized conversation found for the given prompt` across four `assistant-toolkit` planning tests. Same failure on main at `a112889ee9` and on other open PRs; the fixture cache needs regenerating by whoever changed those prompts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored drag-and-drop behavior in the navigation tree.
  * Reordering collections and dropping objects into collections now work correctly across workspace tabs.
  * Prevented items from unrelated trees from being dropped into navigation tree targets.

* **Documentation**
  * Clarified drag-and-drop scope behavior for tree items.


wittjosiah:
Addressed in d9cb9de949.

`treeId` is now in the registration effect's dependency list. The finding predates the switch to deriving the scope from `treePath[0]`, which narrowed it: `treeId` now changes only when `path[0]` does, and `id` was already a dependency. No consumer changes its path root today, so nothing was broken in practice, but the payload could go stale and `isTreeDataFor` matches on exact equality, so the dependency belongs there.

No regression test: a drop cannot be driven synthetically here, since pragmatic-drag-and-drop uses native HTML5 drag events. The behaviour is covered end to end by `collections.spec.ts` instead.

wittjosiah:
Addressed in e3f400cb32.

The concern survives the rewrite even though the thread is outdated: the guard was source-side only. A row's `canDrop` checked `source.element !== element` plus the consumer's predicate, and `TreeEndDropTarget` had no `canDrop` at all, so a navtree row could land on a task-list target and the navtree's monitor would read it as a graph node. The task list passes no `canDrop`, so nothing else was rejecting it.

Fixed in `Tree` rather than in each monitor, since a target's own scope is what decides which sources it accepts, and this closes both directions at once. It was latent only because navtree drags were being claimed by nobody, which is the bug this PR fixes, so restoring the monitor would have made it live.

Verified: `collections.spec.ts` 5/5 locally against a production bundle; CI e2e re-dispatched on this commit.

PR #12924, opened 2026-09-03
build: run dx-build and dx-compile under bun

dmaretskyi:
Moves the two workspace build tools — `dx-build` and `dx-compile` — off node and onto bun. Runtime-only: no package-manager change, no CI change, no test-runner change.

## What changed

Three moon task files, one line each:

| File | Tag | Projects | Before | After |
|---|---|---|---|---|
| `tag-ts-build.yml` | `ts-build` | 12 | `pnpm exec dx-build && pnpm exec dx-compile` | `bun .../dx-build/cli.js && bun .../dx-compile/cli.js` |
| `tag-ts-compile.yml` | `ts-compile` | 17 | `pnpm exec dx-build` | `bun .../dx-build/cli.js` |
| `tag-vendor.yml` | `vendor` | 3 (2 effective) | `pnpm exec dx-compile` | `bun .../dx-compile/cli.js` |

bun needs no new setup step. `.moon/toolchains.yml` already declares the `bun` toolchain, so `moon setup` installs the `.prototools` pin (1.3.11, unchanged) and puts it on every task's PATH. Verified in a clean `env -i` with no proto shims: a moon task resolves `bun -> <proto>/tools/bun/1.3.11/bun`. This is consistent with `.github/Dockerfile` deliberately shipping no bun of its own so the pinned one is the only one a task can reach.

## Why absolute paths instead of `pnpm exec`

`pnpm exec dx-build` cannot select a runtime. pnpm's generated `node_modules/.bin/dx-build` is a shell script ending in:

```sh
exec node  "$basedir/../@dxos/dx-build/cli.js" "$@"
```

It hardcodes `node` and never consults the target's shebang. Confirmed by falsification: setting `cli.js`'s shebang to `#!/nonexistent/interpreter` makes direct dispatch fail with `No such file or directory`, while `pnpm exec dx-trace-imports --help` still succeeds. This is also why simply flipping these tools' shebangs would have been a no-op — that was the original shape of this slice and it was dropped for exactly this reason.

## Verified

Bun 1.3.11 (the `.prototools` pin, installed via `proto install bun`), against a full `pnpm install`.

- **Byte-for-byte artifact equality.** Built 12 projects spanning all three tags under node, then under bun, and compared sha256 manifests: **889 artifacts identical**, including `.mjs` bundles, `.d.ts`, sourcemaps and esbuild `meta.json`.
  ```
  moon run keys:build node-std:build random-access-storage:build sql-sqlite:build \
    echo-panproto:build teleport:build esbuild-plugins:build process:build \
    protocols:build lit-grid:build vendor-hypercore:build vendor-hyperformula:build --force
  ```
- **Full build of every affected project.** All 32 projects carrying `ts-build`/`ts-compile`/`vendor`, 194 tasks, green. This includes `dx-build` and `dx-compile` building *themselves* under bun.
- **Tests still pass on node.** `moon run keys:test teleport:test` — teleport 14/14 across 4 files. vitest is untouched.

### Two exclusions from the artifact comparison, both proven pre-existing

Neither is caused by this change; both are nondeterministic under **node** on `main` today.

1. `dist/**/tsconfig.tsbuildinfo` — 6 node-only runs of `esbuild-plugins:build` produced **5 distinct hashes** (bun produced 4 in 6, drawing from the same pool). The unstable field is `effect.diagnosticSeverity`, whose key insertion order varies run to run.
2. `protocols/dist/src/proto/gen/**` — 4 node-only runs of `protocols:build` produced **4 distinct hashes**.

I initially read the `tsbuildinfo` variance as a bun-introduced reproducibility regression and was going to drop `dx-build` over it. A 2-run control was too small; widening to 6 runs per runtime showed node is equally nondeterministic. Flagging separately as a pre-existing build-reproducibility issue — it defeats content-addressed caching regardless of runtime.

## Dropped from this slice

Five of the seven tools I set out to convert did not survive triage:

- **`beast`** — no `moon.yml`, no build task, no `dist/`, not a dependency of anything, and `node_modules/.bin/beast` does not exist. The root `"beast": "beast docs"` script is already dead. Unbuildable, so unverifiable.
- **`x`** — broken under node today: `bin/x.js` uses `require()` in a `"type": "module"` package (`ReferenceError: require is not defined in ES module scope`). Bun runs it, but with no obtainable node baseline there is nothing to prove equivalence against. Same class of latent breakage as the geo generator scripts.
- **`dx-killorphans`** — wraps every vitest invocation and manipulates process groups and signal relaying (`process.kill(-pid)`, `taskkill /T`). Too close to the test runner, which is explicitly out of scope, and the semantics most likely to diverge silently.
- **`dx-fixtures`** — developer-only (`runInCI: false`), needs R2 credentials and `wrangler`/`age` to do anything real. Cannot be verified beyond its argument-validation path.
- **`dx-trace-imports`** — sound under bun (byte-identical `--help`, builds clean), but it is invoked from ~50 per-project `moon.yml` files. That is a large mechanical diff belonging in its own PR, not bolted onto this one.

## Not tested

- Nothing was run in real CI. Local moon runs only.
- No Windows or macOS verification — Linux x64 only. `killorphans`' `taskkill` path is untouched, but `$workspaceRoot` path handling on Windows is unexercised.
- No published-package consumer path. All seven tools are `private: true`, so there is no npm surface, but I did not test a package consuming these artifacts from a tarball.
- The 889-artifact comparison covers 12 of the 32 affected projects. The other 20 were verified only as "builds green", not byte-compared.
- First run after merge will be a full rebuild: changing a task's `command` changes its hash, invalidating cached outputs for all 32 projects.

## Summary by CodeRabbit

- **Chores**
  - Updated TypeScript and vendored dependency build tasks to invoke build and compilation tools directly through the pinned Bun runtime.
  - Preserved existing build arguments and output locations.
  - Improved consistency of development and continuous integration builds by ensuring these tasks run with the intended runtime.

---
🚀 Composer preview: https://pr-12924-composer-dev.dxos.workers.dev


dmaretskyi (commented):
approve

dmaretskyi:
Dequeued from the merge queue at 12:22:05Z with `failed_checks` — `Check / check` failed on queue commit `7e8d3366`. Every other job on that commit passed (all three completed test shards, `memory`, `boot-budget`); `Check / check` had passed on the PR head itself.

I could not reach the Depot job log, so I reproduced the merged state locally instead: merged `origin/main` (`092f3be8`) into the branch and re-ran that job's gates.

All green:

- `moon run :build` — 442 tasks
- `moon run :check-module-structure` — 459 tasks
- `pnpm run knip`
- `pnpm run format-check`, `check-cycles`, `check-publish-config`, `check-public-dependencies`, `check-packages-published`, `pkg-lint`, `node scripts/sync-versions.mjs --check`, `node scripts/check-cache-wiring.mjs`
- `pnpm toolbox --lint-package-exports` + `git diff --exit-code -- '**/package.json'`

Two things suggest the eviction was not this PR: #12922 merged at 12:22:06Z, one second after this was evicted, so both were in the same queue batch; and #12904 took a `failed_checks` eviction earlier today (11:36:57Z) and then merged clean on retry (11:49:32Z).

I have pushed a merge of current `main` onto the branch so CI re-runs against the merged state. The diff is unchanged — still the same three `.moon/tasks/*.yml` files. Ready to re-queue once this run is green.

Not verified: I never saw the actual failing log, so I cannot positively rule out a real interaction that my local run does not reproduce. If it fails the same way a second time I will treat it as real and bisect the gate rather than re-queue again.

PR #12930, opened 2026-09-03
assistant: bind a space-authored skill to a chat

dmaretskyi:
Clicking a locally-authored skill in the chat's skill picker did nothing — no tick, no error, no skill in the conversation.

closes DX-1248

## Root cause

A skill the user authors in a space carries no **registry** key (`Obj.getMeta(skill).key` is undefined — keys identify registry entries). Two independent layers then discarded it:

1. `AiContext.Binder._updateBindings` dropped every keyless skill on the way in (`log.warn('dropping skill with no meta key')`), so the binding never took effect. The stated reason — "tool/operation registration calls `Skill.getKey` which throws" — turned out to be only two **diagnostic log lines** that call `getKey` eagerly as an argument.
2. `useSkillHandlers` rendered the row with `skill.id` as its identity but re-looked the skill up **by meta key** on click, so the handler matched nothing and returned silently. `useActiveSkills` also filtered keyless skills out of the active map, so even a successful bind could not tick the row.
3. `AgentService.createSession` resolved supplied skills through `Skill.upsert(Skill.getKey(skill))`, which throws on a keyless skill and, for a fork, substitutes the pristine registry copy for the caller's.

The debug-log upload failed on this submission, so this is derived from the code rather than the reporter's bundle; the mechanism reproduces the symptom exactly, and the `dropping skill with no meta key` warning it emits will show up in the next bundle if some other path is at fault.

## Changes

- **`AiContext`** — carry keyless skills through instead of dropping them. The two `Skill.getKey` calls that made a *log* throw on such a skill now read the meta key directly (`AiContext.ts:161`, `SkillHooks.ts:67`).
- **Skill picker** — one `getSkillId` (key, else URI) for the row identity and the active map; `onUpdateSkill` takes the **skill object** rather than re-resolving a key, which removes the silent-no-op class outright, and logs each toggle.
- A space copy of a registry skill now shadows the registry entry in the list — it carries the user's edits, the same precedence `Skill.resolveAnnotatedSkills` already applies — and unbinding clears either form.
- **`AgentService.createSession`** — a skill already in a database is bound as-is.

## Verification

Each test fails on the pre-fix code:

| Test | Pre-fix failure |
| --- | --- |
| `AiContext.test.ts` — `binds a keyless skill stored in the space DB` | `expected [] to deeply equal [ Array(1) ]`, with three `dropping skill with no meta key` warnings |
| `useSkillRegistry.test.tsx` — toggles a keyless space skill on and off | row never enters the active map |
| `AgentService.test.ts` — `binds a space-authored skill…` (model fixture) | `Error: Skill is missing the meta key.` |

The AgentService fixture is the end-to-end proof: the space skill's instructions arrive in the system prompt as `<skill>The codename of the project is Kingfisher.</skill>` and the model answers from them. Appended last in the file so the shared module PRNG leaves every existing fixture's ids untouched — exactly one new fixture, no re-records.

Green locally: `moon run :lint -- --fix` (326 tasks), `assistant:test` 81, `plugin-assistant:test` 191, `agent-runtime:test` 20 (`DX_RUN_MODEL_FIXTURE_TESTS=1`), all builds, `pnpm format`. The repo-wide `:test` sweep flagged two `echo-host` subduction tests; they pass on rerun (311 passed) and are unrelated to this diff.

## Summary by CodeRabbit

* **New Features**
  * Space-authored skills without registry keys can now be bound to chats and remain available in the context binder.
  * Skills can be enabled or disabled directly from object references.
  * Space copies take precedence over matching registry skills in the skill picker.
* **Bug Fixes**
  * Disabling a skill now removes both its space and registry representations from the conversation.
  * Improved support for preserving and resolving skills stored in a space.

---
🚀 Composer preview: https://pr-12930-composer-dev.dxos.workers.dev


packages/core/compute/agent-runtime/src/agent-service/AgentService.ts
```diff
@@ -62,8 +62,13 @@ export const createSession: (
...
   'createSession',
 )(function* (opts) {
+  // A skill already in a database is bound as-is: it is either space-authored (no registry key at
+  // all) or a fork carrying the user's edits, and resolving it through the registry would substitute
+  // the pristine copy for the one the caller handed us.
   const skills = yield* Effect.forEach(opts?.skills ?? [], (skill) =>
-    Skill.upsert(Skill.getKey(skill)).pipe(Effect.map(Ref.make)),
+    Obj.getDatabase(skill) !== undefined
+      ? Effect.succeed(Ref.make(skill))
+      : Skill.upsert(Skill.getKey(skill)).pipe(Effect.map(Ref.make)),
```
dmaretskyi: remove Skill.upsert -- itse deprecated and we dont wanna be putting registry skiils into database anyway

dmaretskyi: Done — `Skill.upsert` is gone from `createSession`: a skill already in a database binds as-is, anything else is referenced by `Ref.fromURI(Skill.registryURI(key))`, the same form the picker's `onUpdateSkill` uses, so nothing is cloned into the space any more.

That surfaced a second bug worth knowing about: a registry-URI ref resolves through the **database's** registry (production wires that up via `RegistrySync`), not the `Registry.Service` the assistant test layer seeds — so every session created with a registry skill silently lost it, and 5 scripted `agent-runtime` tests plus 5 `assistant-toolkit` fixture tests went red. The test layer now seeds both, and the `assistant-toolkit` fixtures whose recorded tool arguments embedded ids the skill clone used to shift are re-recorded.

Green on the pushed head: Model Fixture and all four `Check / test` shards, plus locally `agent-runtime` 66, `assistant` 81, `assistant-toolkit` 85, `plugin-assistant` 191 and the whole `#model-fixture:test` sweep.

---

.store/conversations/packages_core_compute_assistant-toolkit_src_skills_planning_operations_assign-tasks.model/28163721e398419a02371d2390ef5b20ec8ebe8600212824448e15f9eb1140d8.json
```diff
@@ -0,0 +1 @@
...mqpi1Y649tsLjgPC","delta":"B55UZ7KP5HLD"},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"HVTXUTHE2U/0"},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"1JGFJJZ0"},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"0G0W"},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"KQSJGMAKCNT"},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"CQ\""},{"metadata":{},"type":"tool-params-delta","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC","delta":"]}"},{"metadata":{},"type":"tool-params-end","id":"toolu_01Cnsomtmqpi1Y649tsLjgPC"},{"metadata":{"anthropic":{"container":null,"contextManagement":null,"stopSequence":null,"usage":{"cache_creation":{"ephemeral_1h_input_tokens":0,"ephemeral_5m_input_tokens":4942},"cache_creation_input_tokens":4942,"cache_read_input_tokens":0,"inference_geo":"global","input_tokens":2,"output_tokens":226,"service_tier":"standard"}}},"type":"finish","reason":"tool-calls","usage":{"inputTokens":{"uncached":2,"total":4944,"cacheRead":0,"cacheWrite":4942},"outputTokens":{"total":226}}}]}
```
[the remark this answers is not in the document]
dmaretskyi: Valid, and fixed in 72a8655e — that file is gone.

It was a leftover of regenerating in place: the earlier run recorded turns under the intermediate id sequence, so the removal targeted a URI no live run produces. It was unreachable at replay (the suite was green with it present, because the matcher never selected it), but misleading to read, so I deleted both affected suite directories and re-recorded them from empty. The planning fixtures now remove exactly the URI the prompt names and report `unassigned 1`; the suite dir went from 15 files to 6.

`assistant-toolkit` replays green: 85 passed, 3 skipped.

---

PR #12934, opened 2026-09-03
react-ui: stop clicks falling through toasts onto the UI beneath

wittjosiah:
## Defect

"Status indicators clicking through toasts." With a toast on screen and a status-indicator menu open
in the same bottom-end corner, the toast was painted over the menu but did not consume clicks: a
click on the toast's surface activated the menu item underneath it.

Reproduced in the running app (`composer-app` dev server, 1440×900): show a toast, then open the
`HelpMenu` status indicator (the ⓘ item in the R0 rail, which opens `side='left'` straight over the
toast area) and click a point the toast visually covers.

- `document.elementFromPoint` at that point returned a `role="menuitem"` element, not the toast.
- A real `page.mouse.click` there landed on the menu's **About Composer** item.

## Root cause

Not the toast viewport's own styling — neither the viewport nor the root carried
`pointer-events-none`. The inertness comes from Radix's dismissable-layer coordination:

1. `DropdownMenu` in `@dxos/react-ui` defaults to `modal = true`, so its content layer sets
   `disableOutsidePointerEvents`. That writes `document.body { pointer-events: none }`.
2. A toast (`ToastPrimitive.Root`) is itself a `DismissableLayer.Root`. Every layer *below* the
   topmost layer with `disableOutsidePointerEvents` gets an inline `pointer-events: none`, so an
   already-open toast is marked inert the moment the menu opens.
3. The menu content keeps `pointer-events: auto`, and the toast still paints above it
   (`z-40` viewport vs `z-20` menu) — so the toast was visible, inert, and the interactive menu sat
   directly beneath it. Hence clicks "through" the toast.

Measured on the live page before the fix:
`body: none`, toast inline `none`, computed `none`, viewport `none`, menu `auto`.

The order matters, which is why this looked intermittent: a toast that appears *after* the menu is
the higher layer and behaved correctly; a toast already on screen when the menu opens did not.

## Fix

`packages/ui/react-ui/src/components/Toast/Toast.theme.ts` — in the package that owns the toast, not
at a call site:

- `toast.root`: `pointer-events-auto!`. Importance is required to override the *inline* style Radix
  writes on the layer; a plain utility loses to it. The toast therefore always consumes clicks on its
  own box.
- `toast.viewport`: `pointer-events-none`. The viewport spans the whole stack's bounding box, so
  without this it would swallow clicks in the gaps between stacked toasts. Toasts are non-modal, so
  only a toast's own surface intercepts and the rest of the viewport stays click-through — matching
  Radix's documented viewport/toast pattern.

## Files touched

- `packages/ui/react-ui/src/components/Toast/Toast.theme.ts` — the fix.
- `packages/ui/react-ui/src/components/Toast/Toast.stories.tsx` — new `OverModalMenu` story +
  regression assertion (a toast, then a modal status-style menu over it; asserts the hit test at the
  overlap resolves inside the toast).
- `.changeset/toast-pointer-events.md`.

## Verification

Storybook browser test (`moon run react-ui:test-storybook`, real chromium):

- `OverModalMenu` passes with the fix; reverting only `Toast.theme.ts` makes it fail, and the element
  hit at the overlap is a `DropdownMenu.Item` — i.e. the test reproduces the reported defect.
- Full package suite green: 59 files / 187 tests.

Live app (`composer-app` dev server), after the fix:

- Toast + `HelpMenu` (modal, over the toast): hit test resolves inside the toast; a real mouse click
  lands on the toast, no longer on "About Composer". Toast inline `pointer-events` is still `none`,
  computed is `auto` — the important utility does override it.
- Toast over the `plugin-progress` R0 popover (non-modal, 6 active providers so it overlaps): still
  consumed by the toast — no regression.
- The toast's own close button still dismisses it with the viewport now click-through.
- The 8px gap between two stacked toasts now passes clicks to the popover beneath, and the app
  remains interactive with toasts up (clicked the account item while two toasts were showing) — the
  fix does not make the viewport a click blocker.

Repo checks for the package touched: `react-ui:build` (emits types, so this is the typecheck),
`react-ui:lint`, `react-ui:test`, `react-ui:test-storybook` — all green.

## Not verified

- Touch behaviour: swipe-to-dismiss and hover-to-pause were not exercised. They should be unaffected
  (both listen on the toast `li` / Radix's wrapper, which keep pointer events), but I did not test
  them.
- Behaviour change worth a reviewer's eye: while a **modal dialog** is open, a toast is now clickable
  rather than inert. That follows the stated expectation ("clicks landing on a toast are consumed by
  the toast"), and a toast's action/close is arguably always meant to work — but it is a deliberate
  divergence from Radix's default layer inertness, so say the word if dialogs should instead keep
  toasts inert-but-blocking.
- No mobile/`plugin-mobile` layout run, and no e2e spec added (the regression lives as a storybook
  browser test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Toasts now reliably receive clicks on their visible surface when overlapping an open menu or dialog.
  * Empty areas around toasts remain click-through, preserving interaction with the surrounding app.

* **Tests**
  * Added coverage for toast interaction when displayed over a modal menu.

---
🚀 Composer preview: https://pr-12934-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/toast-pointer-events.md
+++ b/.changeset/toast-pointer-events.md
@@ -1,0 +1,7 @@
+---
+'@dxos/react-ui': patch
+---
+
+Toasts consume clicks over their own surface again: while a modal menu or dialog is open, a toast is
+no longer transparent to hit-testing, so clicks landing on it can no longer fall through to the UI it
+covers. The toast viewport itself stays click-through, so the surrounding app remains interactive.
--- a/packages/ui/react-ui/src/components/Toast/Toast.stories.tsx
+++ b/packages/ui/react-ui/src/components/Toast/Toast.stories.tsx
@@ -4,9 +4,11 @@
 
 import { type Meta, type StoryObj } from '@storybook/react-vite';
 import React, { type ReactNode, useState } from 'react';
+import { expect, userEvent, waitFor } from 'storybook/test';
 
 import { withTheme } from '../../testing';
-import { Button } from '../Button';
+import { Button, IconButton } from '../Button';
+import { DropdownMenu } from '../Menu';
 import { Toast } from './Toast';
 
 type ActionTriggerProps = { altText: string; trigger: ReactNode };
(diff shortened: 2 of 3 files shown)


wittjosiah:
CI: every required **Check** job is green (check/lint/fmt, all four test shards, boot-budget, memory, storybook). The only red is `Model Fixture / model-fixture`, which is a deliberately non-required workflow replaying committed LLM fixtures — it is **already failing on `main`**, including this branch's base commit `121b1b7495` (and `2cad6c048e`), and this PR touches only toast CSS tokens plus a story. Not attributable to this change; not fixed here.

PR #12948, opened 2026-09-04
e2e: close the browser context a Playwright page object created

wittjosiah:
## What broke

Last night's nightly (`Check` on `96f94c25`, schedule 04:00 UTC) failed in `e2e (browser=chromium, shard=composer)` on `halo.spec.ts:77 › HALO tests › deleting a space replicates across devices`. All three attempts failed, but only the third produced a diagnosis:

| attempt | duration | error |
| --- | --- | --- |
| 0 | 3.2m | `RangeError: Invalid string length` → `browserContext._wrapApiCall: Invalid string length` → `End of central directory record signature not found. Either not a zip file, or file is truncated.` |
| 1 | 3.2m | same |
| 2 | 1.3m | `expect(guest.getSpaceItems()).toHaveCount(2)` received `1` |

Two of the three retries burned 3.2 minutes each and reported a broken trace writer instead of the test's failure.

## Why

`setupPage` calls `browser.newContext()` and returns only the page. Playwright never reclaims a context created off the worker-scoped `browser` fixture, and its artifacts recorder walks `_allContexts()` at the start of **every** test (`willStartTest` → `didCreateBrowserContext` → `startChunk`) and stops them all into that test's `trace.zip` (`didFinishTest` → `leftoverContexts` → `_stopTracing`, which is the `browserContext._wrapApiCall` frame in the error above).

Every suite here closed the page and left the context behind, so the set grows for the life of the worker and each trace re-serializes all of it. The recovered artifact from that run makes it concrete — attempt 2's `trace.zip` holds **9** contexts, and 7 of them carry network logs timestamped 04:04:33–04:06:04 (attempts 0 and 1) with zero recorded actions:

```
ctx0 network entries=1073 first=04:04:33 last=04:04:45   (0 trace actions)
...
ctx6 network entries=1085 first=04:05:54 last=04:06:04   (0 trace actions)
ctx7 network entries=1063 first=04:12:33 last=04:12:37   (445 trace lines — the host)
ctx8 network entries=1904 first=04:12:36 last=04:12:46   (528 trace lines — the guest)
```

Past a point the writer exceeds V8's maximum string length and the zip is left truncated.

## The change

`setupPage` returns a `close()` that disposes the context it created (or just the page when it was handed a context it does not own). Every page object and spec that previously closed a page now calls it:

- `@dxos/test-utils` — `setupPage` returns `close`.
- `composer-app` — `AppManager.closePage()` becomes `AppManager.close()`; all 11 call sites updated.
- `todomvc`, `examples` — `AppManager` gained a `close()`, wired into the suites' teardown (neither closed anything before).
- `rpc-tunnel-e2e` — the four `beforeAll` specs gained an `afterAll` teardown.
- `react-ui-table`, `react-ui-mosaic`, `lit-grid`, `plugin-sheet`, `plugin-kanban` — `page.close()` → `close()`.

## What this does not fix

Attempt 2's failure is real and stands: the guest never saw the host's space replicate, and the guest console shows the WebRTC transport closing right after `authenticateDevice` followed by `Automerge root doc load timeout (DataSpace)`. That is the two-peer stall `halo.spec.ts` already documents as endemic and tracked under DX-1152, which is why the suite carries `retries: 2`. This PR does not make that stall go away — it makes the retries actually retry, and leaves a readable trace when they do not.

## Testing

`pnpm format`, plus builds/typechecks for the touched projects.

The `e2e` jobs are gated to `main`, `changeset-release/*`, or an explicit dispatch, so they are **skipped** on this PR and its own `Check` run does not exercise the change. The full matrix was dispatched against this branch's head `2aefac74` instead (`check.yml` with `e2e=true`, which also sets `DX_E2E_RUN_ID` so the suites execute rather than replay a cache hit): [run `h20h4bjsj3`](https://depot.dev/orgs/t8fblrl00n/workflows/qfm9w237s6?repo=dxos%2Fdxos). All six e2e cells and all four test shards exited 0.

| cell | result |
| --- | --- |
| `e2e` chromium/composer | 37/37 pass. `halo.spec.ts:40 join new identity` failed once (1.5m) and passed on retry #1 in 22.6s — the retry the trace writer used to destroy. |
| `e2e` chromium/rest | 42/42 pass, 2 flaky |
| `e2e` firefox/composer, webkit/composer | 37/37 pass, no retries |
| `e2e` firefox/rest | 42/42 pass, 1 flaky |
| `e2e` webkit/rest | 38 pass, 4 quarantined (`todomvc`) |
| `test` shards 0–3 | pass (shard 3 on its second attempt) |

No `RangeError: Invalid string length` and no truncated `trace.zip` anywhere in the run.

The four quarantined `todomvc` webkit tests are unchanged from `main` — the nightly on `96f94c25` reports the same `Total: 42 Pass: 38 Fail: 0 Quarantined: 4`, the same four tests, and the same signature (`space invitation never reached the auth-code step`, the shell's auth-code input stuck hidden at `connectingSpaceInvitation`). That is DX-1152 again, and this PR's only change to that package is its teardown.

https://claude.ai/code/session_01UmQMktYvgwQzepygZVkp6h

### DIFF
--- a/.changeset/playwright-context-close.md
+++ b/.changeset/playwright-context-close.md
@@ -1,0 +1,5 @@
+---
+'@dxos/test-utils': patch
+---
+
+`setupPage` now returns a `close()` that disposes the browser context it created. Closing only the page left the context open, and Playwright re-serialized every live context into each later trace until the writer exceeded V8's string limit and left a truncated `trace.zip` behind.
--- a/packages/apps/composer-app/src/playwright/app-manager.ts
+++ b/packages/apps/composer-app/src/playwright/app-manager.ts
@@ -61,6 +61,7 @@ export class AppManager {
 
   private readonly _inIframe: boolean | undefined = undefined;
   private _initialized = false;
+  private _close?: () => Promise<void>;
   private _invitationCode = new Trigger<string>();
   private _authCode = new Trigger<string>();
   // Rolling tail of console errors: the app reports operation failures generically to the user, and
@@ -80,8 +81,9 @@ export class AppManager {
       return;
     }
 
-    const { page } = await setupPage(this._browser, { url: INITIAL_URL });
+    const { page, close } = await setupPage(this._browser, { url: INITIAL_URL });
     this.page = page;
+    this._close = close;
     this.page.on('console', (message) => this._onConsoleMessage(message));
 
     // Assert boot rather than proceed on a swallowed `false`, so a failed boot fails here instead of as
@@ -95,10 +97,8 @@ export class AppManager {
     this.deck = new DeckManager(this.page);
   }
 
-  async closePage(): Promise<void> {
-    if (this.page !== undefined) {
-      await this.page.close();
-    }
+  async close(): Promise<void> {
+    await this._close?.();
   }
 
   //
--- a/packages/apps/composer-app/src/playwright/basic.spec.ts
+++ b/packages/apps/composer-app/src/playwright/basic.spec.ts
@@ -25,7 +25,7 @@ test.describe('Basic tests', () => {
   });
 
   test.afterEach(async () => {
-    await host.closePage();
+    await host.close();
   });
 
   test('create identity, space is created by default', async () => {
--- a/packages/apps/composer-app/src/playwright/chat.spec.ts
+++ b/packages/apps/composer-app/src/playwright/chat.spec.ts
@@ -36,7 +36,7 @@ test.describe('Chat', () => {
   });
 
   test.afterEach(async () => {
-    await host.closePage();
+    await host.close();
   });
 
   test('sends a message and receives a response', async () => {
--- a/packages/apps/composer-app/src/playwright/collaboration.spec.ts
+++ b/packages/apps/composer-app/src/playwright/collaboration.spec.ts
@@ -51,8 +51,8 @@ test.describe('Collaboration tests', () => {
     // NOTE: `afterEach` even if the test is skipped in the beforeEach!
     // Guard against uninitialized app managers.
     if (host !== undefined && guest !== undefined) {
-      await host.closePage();
-      await guest.closePage();
+      await host.close();
+      await guest.close();
     }
   });
 
--- a/packages/apps/composer-app/src/playwright/collaboration.spec.ts
+++ b/packages/apps/composer-app/src/playwright/collaboration.spec.ts
@@ -51,8 +51,7 @@ test.describe('Collaboration tests', () => {
     // NOTE: `afterEach` even if the test is skipped in the beforeEach!
     // Guard against uninitialized app managers.
     if (host !== undefined && guest !== undefined) {
-      await host.close();
-      await guest.close();
+      await Promise.all([host.close(), guest.close()]);
     }
   });
 
--- a/packages/apps/composer-app/src/playwright/collections.spec.ts
+++ b/packages/apps/composer-app/src/playwright/collections.spec.ts
@@ -15,7 +15,7 @@ test.describe('Collection tests', () => {
   });
 
   test.afterEach(async () => {
-    await host.closePage();
+    await host.close();
   });
 
   test('create collection', async () => {
--- a/packages/apps/composer-app/src/playwright/comments.spec.ts
+++ b/packages/apps/composer-app/src/playwright/comments.spec.ts
@@ -21,7 +21,7 @@ test.describe('Comments tests', () => {
   });
 
   test.afterEach(async () => {
-    await host.closePage();
+    await host.close();
   });
 
   test('create', async () => {
--- a/packages/apps/composer-app/src/playwright/first-run.spec.ts
+++ b/packages/apps/composer-app/src/playwright/first-run.spec.ts
@@ -17,7 +17,7 @@ test.describe.skip('First-run tests', () => {
   });
 
   test.afterEach(async () => {
-    await host.closePage();
+    await host.close();
   });
 
   test('help plugin tooltip displays (eventually) on first run and increments correctly', async () => {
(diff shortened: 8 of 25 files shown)


wittjosiah:
### Two red `test` shards, neither from this diff

| check | failing test | verdict |
| --- | --- | --- |
| `Check / test (shard=0)` (dispatched run `3k3j8t9q2r`) | `client-e2e › src/indexing.test.ts › indexes persists between client restarts` | `Test timed out in 15000ms` — re-run attempt 2 **passed** |
| `Check / test (shard=2)` (this PR's run `9nf0sxt27l`) | `echo-host › src/db-host/space-state-manager.test.ts › SpaceStateManager persists and restores space root mappings` | `Test timed out in 15000ms` — re-run in flight |

Both are timeouts, not assertion failures, and both are out of this diff's reach: this PR touches only Playwright specs plus `setupPage` in `@dxos/test-utils/src/playwright.ts`, and neither `packages/sdk/client-e2e` nor `packages/core/echo/echo-host` contains any Playwright code (`grep -rn 'setupPage\|test-utils/playwright'` on both is empty).

Run locally on this branch's HEAD, both finish far inside the 15s budget:

```
client-e2e:test  ✓ indexing.test.ts › indexes persists between client restarts   1056ms
echo-host:test   ✓ space-state-manager.test.ts (7 tests)                          668ms
```

668 ms against a 15,000 ms timeout is not a logic failure. A likely contributor is self-inflicted: I dispatched a second full `Check` (with `e2e=true`, which widens the affected scope to `--all`) while this PR's own `Check` was still running on the same head SHA, so two complete runs plus six browser e2e cells were competing for the same runner pool. Nothing is being pushed for these two, and no fix exists to port because there is nothing broken to fix.

### What the dispatched run does show

`e2e (browser=chromium, shard=composer)` — the job that failed in last night's nightly — passed first try:

```
✓ 26 halo.spec.ts:42 › HALO tests › join new identity (27.9s)
✓ 27 halo.spec.ts:77 › HALO tests › deleting a space replicates across devices (29.6s)
  20 passed (2.1m)
Total: 37   Pass: 37   Fail: 0   Quarantined: 0
```

No retries consumed, no `RangeError: Invalid string length`, no truncated `trace.zip`.

Two caveats on how far that goes. The halo test passing once is a single sample of an intermittent stall, so it is not evidence DX-1152 is fixed. And because nothing failed, no trace had to be written, so this run does not directly exercise the trace-writer path either — that part rests on the Playwright source (`willStartTest` → `_allContexts()` → `startChunk`; `didFinishTest` → `leftoverContexts` → `_stopTracing`) and on the nightly's 9-context artifact.

---

wittjosiah:
### `Check / boot-budget-comment` is red repo-wide, not from this PR

Failing on head `31786464`, attempts 1 and 2, both with **no logs and no sandbox** — the job never got a runner. Its parent `boot-budget` provisioned and passed in the same run, and the job's only step is `continue-on-error: true`, so there is nothing in it that can fail on merit.

It is not specific to this PR. Surveying recent `Check` runs triggered by `pull_request`:

| run started | head | `boot-budget` | `boot-budget-comment` |
| --- | --- | --- | --- |
| 12:57 | `9a6edee2` (a different PR) | failed | **failed** |
| 12:45 | `31786464` (this PR) | finished | **failed** |
| 12:22 | `458fde5f` | finished | finished |
| 12:11 | `7702302f` | finished | finished |
| 11:59 | `873953ab` | finished | finished |
| …through 09:58 | | finished | finished |

Green on every run up to 12:22, failing on every run from 12:45 onward, across at least two unrelated PRs. Consistent with a Depot provisioning problem rather than a code change — worth noting that `boot-budget-comment` is one of only two jobs in `check.yml` on `depot-ubuntu-latest` rather than `depot-ubuntu-24.04-8`, though I have not proven the runner class is the variable.

My one re-run is spent (attempt 2, same result). Nothing to port — there is no fix for this in the repo, because nothing in the repo is broken. Flagging rather than chasing it; if it does not clear on its own, it needs someone with Depot org access rather than a change to this branch.

This is separate from the two `test`-shard timeouts covered [earlier in this PR](https://github.com/dxos/dxos/pull/12948#issuecomment-5540389026).

---

wittjosiah:
### CodeRabbit review: 5 fixed, 3 declined

Pushed as `2aefac74`.

**Fixed** — all four stability findings were real, and all four were the leak this PR exists to fix surviving on a path I missed:

1. **`setupPage` leaked on setup failure** (`playwright.ts`). If `newPage()`, `setViewportSize()` or `goto()` rejected, it threw before returning `close`, so the caller had nothing to dispose the context with. Setup now closes what it created and rethrows the original error — the cleanup failure does not displace the cause.
2. **`react-ui-table` closed at the end of each test body**, so a failing assertion skipped the close and stranded the context. Cleanup moved to `afterEach`, matching the sibling storybook suites. Also removed 12 trailing `await close()` calls.
3. **Paired-manager teardowns** (`halo`, `collaboration`, `todomvc`) awaited `host.close()` before starting `guest.close()`, so a rejection from the first stranded the second's context. Both closes now start before either is awaited.
4. **`rpc-tunnel` `close` unguarded in `afterAll`** — Playwright runs `afterAll` even when `beforeAll` threw before assigning it, so the call could throw a `TypeError` masking the real failure. Now optional.
5. **JSDoc on `setupPage`** — taken. Its contract genuinely is not obvious: `close()` disposes the context when handed a `Browser`, but only the page when handed a `BrowserContext`.

**Declined** — JSDoc on `AppManager.close()` in the composer-app, todomvc and examples page objects. The suggested text ("Closes the page or browser context created during initialization") restates the method name and its return type on a two-line delegation. `AGENTS.md` asks for JSDoc on public functions, but it also says to delete a comment the code already makes obvious, and that is the more specific rule here. Happy to add them if a maintainer would rather have the coverage.

Validated locally: `pnpm format` clean, and `moon exec test-utils:build composer-app:build todomvc:build rpc-tunnel-e2e:build react-ui-table:build` → `Tasks: 393 completed`, exit 0.

---

PR #12967, opened 2026-09-06
client: move the WebRTC transport bridge to buf types

dmaretskyi:
Chunk 5 of the protobuf.js → buf teardown: `BridgeService`, the group I deliberately held back from #12966.

## The cast that made this dangerous

In #12966 I flipped `BridgeService` to measure the blast radius and **the full build reported zero errors** — the alarming result, not the reassuring one. `bridge-rpc.ts` joined the Effect RPC group to protobuf.js-typed handlers with `handlers as never`, so the codec could be swapped underneath handlers still emitting the old shapes with nothing to catch it. I reverted and deferred.

This PR starts by deleting that cast. **Removing it alone compiles clean** — the handlers already matched structurally, so the cast was masking nothing *at the time*. But with the cast gone, flipping the codec immediately produced 5 real type errors across both legs of the adapter. That is the whole point: the seam is now type-checked, and the failure mode that would have shipped silently is one the compiler refuses.

The comment above it claimed per-method handler types "cannot be expressed statically". They can.

## Approach: bridge at the RPC boundary, not in the transport

The obvious move — convert `RtcTransportService` and `RtcTransportProxy` to buf — is the wrong one. Every bridge request carries `dxos.keys.PublicKey proxy_id`, and `RtcTransportService` looks its proxies up by that value. Under buf that becomes a `{$typeName, data}` message which compares by identity, so a lookup would silently miss — the same failure as the `_matchesInvitationContext` bug in #12963.

Instead the transport keeps the protobuf.js shapes and `bridge-codec.ts` converts at the boundary. `proxyId` stays a real `PublicKey` on the transport side, so the hazard never arises rather than being defended against. Both codecs agree byte for byte, so the encoding is the conversion — the same seam pattern as #12963's metadata and #12966's signal manager. It deletes itself when the transport moves to buf.

## Runtime coverage, and why I added a test

`rtc-transport-proxy.test.ts` — the only suite exercising this bridge — is `describe.skip` in source, 14/14 skipped, along with every WebRTC transport test. So the build proves nothing about the round trip here, and this is exactly the code where a dropped field corrupts a live connection with no type error.

`bridge-codec.test.ts` covers what I wrote: every request/response round-trips, the data payload survives byte for byte, each `BridgeEvent` variant survives, `StatsResponse` survives `google.protobuf.Struct`, and `proxyId` comes back as a `PublicKey` instance — the property the transport's map lookup actually depends on.

I did not un-skip the transport suites; that is a pre-existing decision and its own change.

## Validation

- Full repo build: exit 0, **0 errors**.
- `moon run :lint -- --fix`: exit 0. `pnpm format-check`: clean over 14,169 files — lint first, then format, same tree.
- Cast audit over the diff: the only `as never` is a **deletion**; no new casts, no non-null assertions.
- `client-protocol` bridge codec: 5/5. `network-manager`: 33 passed (47 pre-existing skips).

## Next

35 `protoMessage` calls remain. `IdentityService` (12), `SpacesService` (11), `DevicesService` (4) and `ContactsService` (3) are all blocked behind the credentials fence — via two type-only references to `ProfileDocument`/`DeviceProfileDocument` in `credential-generator.ts`. `DevtoolsHost` (6) and `QueryService` (1) are not.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebRTC bridge communication compatibility while preserving existing message formats and payload data.
  * Maintained reliable handling of connection, signaling, data, event, details, and statistics messages across bridge transports.

* **Tests**
  * Added coverage for message conversion, proxy identifiers, binary data integrity, bridge events, and statistics handling.

---
🚀 Composer preview: https://pr-12967-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/bridge-service-buf-types.md
+++ b/.changeset/bridge-service-buf-types.md
@@ -1,0 +1,10 @@
+---
+'@dxos/client': minor
+---
+
+Move the WebRTC transport bridge to buf types
+
+`BridgeService`'s RPC payloads are buf messages. The transport itself (`RtcTransportService`,
+`RtcTransportProxyFactory`) keeps the protobuf.js shapes, and `bridge-codec.ts` converts at the RPC
+boundary, so consumers of the transport interface are unaffected. Both codecs produce identical
+wire bytes.
--- a/packages/core/protocols/src/BridgeService.ts
+++ b/packages/core/protocols/src/BridgeService.ts
@@ -6,7 +6,18 @@ import * as Rpc from 'effect/unstable/rpc/Rpc';
 import type * as RpcClient from 'effect/unstable/rpc/RpcClient';
 import * as RpcGroup from 'effect/unstable/rpc/RpcGroup';
 
-import { protoMessage, serviceError } from './service-rpc.ts';
+import {
+  BridgeEventSchema,
+  CloseRequestSchema,
+  ConnectionRequestSchema,
+  DataRequestSchema,
+  DetailsRequestSchema,
+  DetailsResponseSchema,
+  SignalRequestSchema,
+  StatsRequestSchema,
+  StatsResponseSchema,
+} from './buf/proto/gen/dxos/mesh/bridge_pb.ts';
+import { bufMessage, serviceError } from './service-rpc.ts';
 
 /**
  * Effect RPC definitions for the WebRTC transport bridge (`dxos.mesh.bridge.BridgeService`).
@@ -17,31 +28,31 @@ import { protoMessage, serviceError } from './service-rpc.ts';
  */
 export class Rpcs extends RpcGroup.make(
   Rpc.make('open', {
-    payload: protoMessage('dxos.mesh.bridge.ConnectionRequest'),
-    success: protoMessage('dxos.mesh.bridge.BridgeEvent'),
+    payload: bufMessage(ConnectionRequestSchema),
+    success: bufMessage(BridgeEventSchema),
     error: serviceError,
     stream: true,
   }),
   Rpc.make('sendSignal', {
-    payload: protoMessage('dxos.mesh.bridge.SignalRequest'),
+    payload: bufMessage(SignalRequestSchema),
     error: serviceError,
   }),
   Rpc.make('sendData', {
-    payload: protoMessage('dxos.mesh.bridge.DataRequest'),
+    payload: bufMessage(DataRequestSchema),
     error: serviceError,
   }),
   Rpc.make('close', {
-    payload: protoMessage('dxos.mesh.bridge.CloseRequest'),
+    payload: bufMessage(CloseRequestSchema),
     error: serviceError,
   }),
   Rpc.make('getDetails', {
-    payload: protoMessage('dxos.mesh.bridge.DetailsRequest'),
-    success: protoMessage('dxos.mesh.bridge.DetailsResponse'),
+    payload: bufMessage(DetailsRequestSchema),
+    success: bufMessage(DetailsResponseSchema),
     error: serviceError,
   }),
   Rpc.make('getStats', {
-    payload: protoMessage('dxos.mesh.bridge.StatsRequest'),
-    success: protoMessage('dxos.mesh.bridge.StatsResponse'),
+    payload: bufMessage(StatsRequestSchema),
+    success: bufMessage(StatsResponseSchema),
     error: serviceError,
   }),
 ).prefix('BridgeService.') {}
--- a/packages/sdk/client-protocol/src/bridge-codec.test.ts
+++ b/packages/sdk/client-protocol/src/bridge-codec.test.ts
@@ -1,0 +1,75 @@
+//
+// Copyright 2026 DXOS.org
+//
+
+import { describe, test } from 'vitest';
+
+import { PublicKey } from '@dxos/keys';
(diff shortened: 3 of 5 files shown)


dmaretskyi:
`Check / test (node-version=24.11.1, shard=2)` failed on `6e29997`. Shards 0, 1 and 3 are green, as are `check`, `build`, `memory`, `boot-budget` and `model-fixture`.

I could not reproduce it and could not read it. What I ran locally on this exact tree:

- **Full repo test sweep** — `MOON_CONCURRENCY=4 moon run :test -- --no-file-parallelism`: exit 0, 253 suite summaries, every one `passed`, zero `failed`.
- **Every direct consumer of this diff** — `client-protocol` 10/10 (including the 5 new bridge-codec round-trip tests), `client` 30 passed, `worker-framework` 24/24, `network-manager` 33 passed, `protocols` and `client-services` green.
- Full build 0 errors, lint 0, format-check clean.

The Depot check run returns empty `output`, and the job is not in the GitHub Actions API (`get_job_logs` → 404), so I have no failing test name to work from and no means to re-run it myself.

I also checked the one CI-only mechanism this diff plausibly introduces — the new test imports `@dxos/keys`, which would resolve locally via hoisting even if undeclared — and both `@dxos/keys` and `@dxos/protocols` are declared `workspace:*` in `client-protocol`, so that isn't it.

I am not calling this a flake and I am not merging past it. If someone can open the Depot job (`s6qphsbq8v` / job `l4nflglpd4`) and paste the failing test, or re-run shard 2, I will take it from there. Keeping the PR watched meanwhile.

---

dmaretskyi:
Reproduced and root-caused. It is not this PR's.

My first sweep used `--no-file-parallelism` (what `AGENTS.md` prescribes) and was green. Re-running **with** file parallelism, which is how CI runs it, fails:

```
FAIL |node| src/cache.test.ts > symbol cache reuse > editing one package invalidates only that package
Error: ENOENT: no such file or directory, lstat
  '.../packages/reflect/introspect/src/__fixtures__/node_modules/.cache/dxos-introspect'
  ❯ src/cache.test.ts:35:5   await cp(FIXTURE_SRC, root, { recursive: true });
```

`introspect` is untouched by this PR — the diff is `BridgeService.ts` plus three files in `client-protocol`. In isolation the suite passes 27/29.

**Mechanism.** `beforeEach` copies the *source* fixture into a temp dir, then clears the cache in the copy. But the introspector also writes its cache into the source fixture: `src/__fixtures__/node_modules/.cache/dxos-introspect` exists on disk (untracked, and its mtime lands during a test run). Under parallelism one test removes that directory while another's `cp` is walking it — `readdir` sees the entry, `lstat` then fails. Serial runs never interleave, which is why this is invisible without parallelism.

**Proposed patch** (in `packages/reflect/introspect/src/cache.test.ts`) — skip the volatile path, which the copy never needs:

```ts
await cp(FIXTURE_SRC, root, {
  recursive: true,
  // The introspector writes its cache under the source fixture too, and a concurrent test can
  // remove it mid-walk.
  filter: (source) => !source.includes(join('node_modules', '.cache')),
});
```

The deeper fix is to stop the introspector writing into the source fixture at all, but the filter makes the copy robust either way.

I have not pushed this: it is an unrelated package, and folding a test-isolation fix into a protocol migration muddies both. Happy to land it as its own PR — say the word and I will. This PR stays red until that lands or someone re-runs shard 2, and I am still watching it.

---

dmaretskyi:
Correcting the "Next" section of the description above: I wrote that `DevtoolsHost` (6) and `QueryService` (1) were **not** blocked. Both claims are wrong, and I found it while trying to start that work.

- **`QueryService` has zero `protoMessage` calls.** The "1" I counted was the word appearing in a comment explaining why its schemas are hand-authored.
- **`DevtoolsHost` has 4, not 6** (I counted the import line and a comment), and all four are blocked:
  - `GetSpaceSnapshotResponse` / `SaveSpaceSnapshotResponse` → `SpaceSnapshot` → the **`Timeframe` proto substitution**, a class with a `frames()` accessor. `DevtoolsHost.ts:152` already documents this as the reason those two stay on `protoMessage`.
  - `SubscribeToFeedBlocksResponse` → `FeedMessage`, which embeds both `TimeframeVector` and `CredentialsMessage`.
  - `SubscribeToMetadataResponse` → `EchoMetadata`, i.e. the metadata group.

Accurate remaining inventory — **32 calls**, of which exactly **one** is reachable:

| Blocker | Calls |
| --- | ---: |
| Credentials fence (`IdentityService` 11, `SpacesService` 9, `DevicesService` 3, `ContactsService` 2, `InvitationsService` 1) | 26 |
| `Timeframe` class substitution (`DevtoolsHost` snapshot + feed, and `SpacesService.createEpoch`) | 3 |
| `EchoMetadata` group (`DevtoolsHost`) | 1 |
| Test fixture (`QueryService.test.ts`) | 1 |
| **Reachable: `SystemService` → `Platform`** | **1** |

`Timeframe` is the new one and it is not a policy fence: buf has no equivalent of protobuf.js's class substitutions, so a `Timeframe` that arrives as a plain message loses `frames()`. It needs the same treatment `PublicKey` got — a bridge at the boundary, or a wrapper — which is a design call rather than a mechanical conversion.

---

PR #12976, opened 2026-09-07
build: eliminate dx-compile and dx-build in favour of raw vite and tsc

dmaretskyi:
Retires both bespoke build tools. Every library package now builds through the shared vite config, and declaration emit is a plain `tsc` invocation.

## `dx-build` → raw `tsc`

`dx-build` was a wrapper that cleaned the outDir, spawned `tsc`, and rewrote diagnostic paths to be repo-relative. Those two behaviours moved into `DxDeclarationsPlugin`, its only real caller; the `ts-compile` moon tag runs `rm -rf dist && pnpm exec tsc` directly.

The clean is load-bearing, not housekeeping: a surviving `tsconfig.tsbuildinfo` makes an incremental `tsc` consider outputs vite already deleted up to date and emit nothing, so the package ships a `types` entry pointing at a missing file.

## `dx-compile` → vite

`defineConfig` gained the capabilities the esbuild pipeline had and vite lacked:

| Option | Replaces | Notes |
| --- | --- | --- |
| `bundle: string[]` | `--bundlePackage` | Drops the named package (and its subpaths) from the `external` predicate, and exempts it from the node-std remap. |
| `alias` | `--alias` | Via a new `DxAliasPlugin`, not `resolve.alias` — see below. |
| `mainFields` | `--mainFields` | |
| `injectGlobals` / `importGlobals` | same flags | Banner import, and a `@rollup/plugin-inject` rewrite of free `global`/`Buffer`/`process`. |
| `declarations: false` | — | For the vendored bundles, which ship hand-written `.d.ts` (and hyperformula has no tsconfig at all). |

Two places where vite needed more than a config option:

**Aliases** needed their own plugin rather than `resolve.alias`: rollup consults the `external` predicate on the raw specifier *before* any plugin resolves it, so vite's alias never saw an externalized bare import and `@effect/wa-sqlite` stayed in sql-sqlite's output under its upstream name.

**`require()` inside inlined CommonJS** needed `DxHoistRequirePlugin`. Rolldown wraps each inlined CJS module in a `__commonJSMin` factory and leaves its `require("pkg")` calls verbatim, behind a shim that throws wherever `require` is absent — so a browser consumer died on the first one. The calls sit inside factory bodies rather than at the top level, which is why rolldown's own `esmExternalRequirePlugin` cannot lift them; `dx-compile` rewrote the emitted output for the same reason.

No `formats` option: a package needing CJS gets its own config, because the two formats have to disagree about `chunkFileNames` and often about a transform, which one shared output object cannot express.

## Migrated packages

All 12 `ts-build` packages and both `vendor` bundles — ui-theme, random-access-storage, node-std, keys, sql-sqlite, teleport, echo-panproto, shell, devtools, plugin-{map-solid,atproto,library}, vendor-{hypercore,hyperformula}.

Their `exports`/`imports` maps are rewritten from the per-platform-slug layout (`dist/lib/{browser,node-esm,neutral}/…`) to the flat `dist/lib/<entry>.mjs` the vite build emits, under the `default` condition.

shell and devtools each already owned a `vite.config.ts` for a different job — the shell bundle and the devtools dev server. Those move to `vite.bundle.config.ts` / `vite.app.config.ts`, so `vite.config.ts` means the library build in every package.

ui-theme's `@dxos/ui-theme/plugin` keeps a standalone config: it is loaded by a consumer's own vite config, so it ships CJS alongside ESM, and the two formats differ in a *transform* (`import.meta.dirname` has no meaning in CJS), which makes them two passes rather than two outputs. Its `ROOT` walk is retuned for the flat output — the old four-level walk overshot the package root and broke every storybook suite that loads the theme plugin.

## Also removed

app-framework's and config's `compile-plugin` tasks emitted a `dist/plugin` tree nothing resolved — both packages' `exports` already named the vite bundles. The `dx-plugin` CLI becomes a `plugin-cli` entry of app-framework's own build.

`.config/knip.ts` learned the build shape from `--entryPoint=` / `--bundlePackage=` arguments on the moon tasks, so it now reads `entry` and `bundle` out of each `vite.config.ts` instead (scoped to those properties, so an app's `src/main.tsx` is not mistaken for a library entry).

## Verification

Repo-wide `moon exec :build` passes, and the working tree is clean afterwards — `dx-plugin gen` reproduces the hand-written `#capabilities` maps byte-for-byte. `:check-module-structure`, `:test-types`, `pnpm knip`, `pnpm toolbox --lint-package-exports` (the exports-in-sync gate), `oxfmt --check` and `moon run :lint` are all clean, and every migrated package passes its own suites.

Each migrated package's output was inspected directly: `varint` inlined into teleport, the wa-sqlite alias landing in sql-sqlite's bundle, `capabilities.{node,workerd}.mjs` emitted for the conditional plugin barrels, both vendor bundles self-contained with no `__require` of an external left, and `dx-plugin --help` running from its new path.

One commit here is outside the migration: `react-ui-form: forward the trigger prop from FormFieldHeader`. `FormFieldHeader` accepted `trigger` and dropped it, so a nested object's disclosure header never became the `Collapsible`'s trigger. It is included at the request of the PR author; see the discussion below for why it does not on its own green `react-ui-form`'s `FormFieldSet` test, whose story renders an empty DOM.

https://claude.ai/code/session_01CjQ2qjeiUk2u72qmkhyUbk

---
_Generated by [Claude Code](https://claude.ai/code/session_01CjQ2qjeiUk2u72qmkhyUbk)_

---
🚀 Composer preview: https://pr-12976-composer-dev.dxos.workers.dev


vite.base.config.ts
```diff
@@ -996,8 +1121,8 @@ export const defineConfig = (options: DxConfigOptions = {}): UserConfig => {
     build: {
       lib: {
         entry: resolvedEntry,
-        formats: ['es'],
-        fileName: (_, name) => `${name}.mjs`,
+        formats,
+        fileName: (format, name) => `${name}.${format === 'cjs' ? 'cjs' : 'mjs'}`,
```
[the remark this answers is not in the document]
dmaretskyi: Confirmed and fixed in 0a87cd62, by removing the `formats` option rather than making `chunkFileNames` format-aware.

You're right that the fixed `chunk-[name].mjs` would have named CommonJS chunks with an ESM extension. It was latent — no package passed `formats`, so nothing has ever built CJS through the shared config — but the option was a trap either way.

I went with removal because a single `rollupOptions.output` object can't express the difference: `build.lib.formats` maps each format onto the *same* output object, so one `chunkFileNames` has to serve both. And the two formats often diverge in a *transform*, not just a filename — `@dxos/ui-theme/plugin` is the one CJS build here, and it needs `define: { 'import.meta.dirname': '__dirname' }` for its CJS pass, which no output option can supply. So it already runs as two passes over its own standalone config, and that stays the pattern for any future CJS consumer. `defineConfig` is now ESM-only with a comment recording why.

---

packages/core/echo/echo-panproto/package.json
```diff
@@ -16,20 +16,17 @@
     ".": {
       "source": "./src/index.ts",
       "types": "./dist/types/src/index.d.ts",
-      "browser": "./dist/lib/browser/index.mjs",
-      "node": "./dist/lib/node-esm/index.mjs"
+      "import": "./dist/lib/index.mjs"
```
dmaretskyi: keep import -> default everywhere accross your changes, so it always matches

dmaretskyi: Done in ec93ccbd — every bundle entry this PR touches now uses `default`.

You were right that the diff disagreed with itself: `teleport`, `plugin-map-solid`, `plugin-atproto`, `plugin-library` and both vendor bundles already used `default` on main, and my rewrite flipped some of them to `import` while leaving others alone. `default` is also the safer key, since it resolves under every condition rather than only an `import` statement.

Two things I deliberately left alone:

- `@dxos/ui-theme/plugin` keeps `import` + `require`. It is the one genuinely dual-format entry (separate `.mjs` and `.cjs` files), so it cannot collapse to a single `default`.
- Entries that name a different file per condition stay as they are — `sql-sqlite`'s `./platform` (browser/bun/node), `node-std`'s per-condition map, and the plugins' `#capabilities` (node/workerd/default).

Worth noting the toolbox agrees: `pnpm toolbox --lint-package-exports` regenerates these maps and leaves `default` untouched, so the `Check package exports in sync` gate is happy with it.

I scoped this to the packages this PR touches. The ~1200 `import` keys elsewhere in the repo are pre-existing, and converting them looked like a separate change rather than something to fold in here — say the word if you'd like it done repo-wide.

---

dmaretskyi:
### `Check / test` shard 2 — a base-branch failure, not this PR's

**What is failing.** `react-ui-form:test` →
`src/components/Form/FormFieldSet/FormFieldSet.test.tsx > FormFieldSet > a nested object folds under its own header`:

```
TestingLibraryElementError: Unable to find role="button" and name "Address"
```

Deterministic — it reproduces at `MOON_CONCURRENCY=1` in isolation, so it is not the browser-runner starvation that accounts for the other intermittent storybook failures in this shard.

**Root cause.** `FormFieldHeader` accepts a `trigger` prop and never forwards it:

```ts
// packages/ui/react-ui-form/src/components/Form/FormField/FormFieldHeader.tsx
export const FormFieldHeader = ({ label, path, required, readonly, classNames, add, actions, trigger }) => (
  <FormFieldLabel standalone classNames={classNames} label={label} … />   // ← no `trigger`
);
```

`FormFieldLabel` (in `FormField/FormRow.tsx`) is what turns the row into the disclosure:

```ts
return trigger ? (
  <Collapsible.Trigger classNames={…}>…</Collapsible.Trigger>
) : (
  …
);
```

With `trigger` dropped, `FormFieldSetContainer`'s header renders as a plain element, so there is no `role="button"` with `aria-expanded`/`aria-controls` for the test to find.

**Proposed patch** (one line):

```diff
   <FormFieldLabel
     standalone
     classNames={classNames}
     label={label}
     required={required}
     readonly={readonly}
     path={path}
+    trigger={trigger}
     button={
```

**Why it is not this PR's.** `react-ui-form` is not in this diff. Both `FormFieldHeader.tsx` and `FormFieldSet.test.tsx` were last modified by b7822a77 (`react-ui: migrate from Radix to Ark UI (Phases 1–4)`, #12961), which is in this PR's base. `@dxos/ui-theme` and `@dxos/react-ui` both keep their `source` export conditions, so this test resolves them from source and none of this PR's export-map or bundle-layout changes can reach it.

I have not pushed the fix, since it is a UI change in an area this PR does not otherwise touch — happy to include it if the owners would rather see it land here than in a follow-up.

I also have no way to re-run a Depot job from here (no `gh`, and these are not GitHub Actions runs), so I have not spent a re-run on this.

---

PR #12978, opened 2026-09-07
client-services: move identity, contacts, devices and spaces RPCs to buf

dmaretskyi:
Chunk 7 of the protobuf.js → buf teardown: the credentials group.

Flips the `ContactsService`, `DevicesService`, `IdentityService` and `SpacesService` RPC definitions from `protoMessage` to `bufMessage` — 25 calls, leaving **7** repo-wide (`DevtoolsHost` 4, `InvitationsService` 1, `QueryService` 1, `SpacesService` 1).

## Credentials core is untouched

A credential is signed over its protobuf.js encoding, so converting it in `@dxos/credentials` would change what is verified. Instead the payloads cross the RPC boundary as their shared wire bytes, through two new bridge modules:

- `client-services/…/services/credentials-codec.ts` — host side, `@dxos/credentials` shapes ⇄ buf.
- `client/src/services/legacy-codec.ts` — client side, so `@dxos/client`'s **public API is unchanged** (notably `Credential.subject.assertion`, which callers index by `'@type'`).

Both codecs agree byte-for-byte, so the encoding *is* the conversion.

## Two deliberate hold-backs

`SpacesService.postMessage` and `subscribeMessages` stay on protobuf.js. The gossip payload is an arbitrary `Any` a caller keys by `'@type'`; buf carries an `Any` as typeUrl + bytes, which changes what a caller can pass and what the registry can resolve. Flipping `subscribeMessages` broke `spaces.test.ts` ("post and listen to messages") and a subduction e2e — reverted rather than worked around.

## A type/runtime mismatch this surfaced

`client-protocol`'s `ClientServices` promise surface still declared the protobuf.js shapes for methods that already carry buf (from the earlier network and platform chunks). Because `bufMessage` payload schemas accept a structurally-compatible literal, the type-checker never saw it — plain object literals reached `toBinary` and failed at runtime (`ForeignFieldError`). Tightening that interface is what surfaced the consumer updates here: devtools, shell, observability, composer-app, plugin-space, plugin-client, plugin-onboarding, react-client.

## Test fixture

`createMockCredential` asserted `example.testing.rpc.MessageWithAny`, a type the buf registry cannot resolve; it now asserts `dxos.halo.credentials.AuthorizedDevice`, so the round-trip is exercised against a type the schema actually declares.

## Validation

- `moon exec :build` — 0 errors.
- `moon run :lint` — clean.
- `MOON_CONCURRENCY=4 moon run :test` **with file parallelism** (not `--no-file-parallelism`, which hides races) — green apart from one flake, below.
- `oxfmt --check` — clean.

## Known flake (not from this change)

`client`'s `dedicated-worker-client-services.test.ts` fails a *different* leader-election case on different sweeps and passes 3/3 in isolation. This branch touches no leader-election code; it looks load-sensitive on the 4-core sandbox. Flagging rather than fixing.

---

## Summary by CodeRabbit

- **Improvements**
  - Improved compatibility for identity, contact, device, space, and credential data exchanged across services.
  - Enhanced credential and presentation handling while preserving the existing public API.
  - Improved device profile updates and identity recovery with stronger request validation.
  - Improved space and synchronization status handling, including safer defaults when status data is unavailable.
  - Updated service integrations and examples to use consistent protocol representations.
  - Increased the default space-wait timeout to improve reliability during slower operations.

---
🚀 Composer preview: https://pr-12978-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/buf-credentials-service-rpcs.md
+++ b/.changeset/buf-credentials-service-rpcs.md
@@ -1,0 +1,9 @@
+---
+'@dxos/protocols': patch
+'@dxos/client': patch
+'@dxos/client-services': patch
+---
+
+Move the identity, contacts, devices and spaces service RPCs to buf messages. The credential
+subsystem keeps its protobuf.js shapes — the payloads cross the RPC boundary as their shared wire
+bytes — so `@dxos/client`'s public API is unchanged.
--- a/.changeset/buf-credentials-service-rpcs.md
+++ b/.changeset/buf-credentials-service-rpcs.md
@@ -1,9 +1,11 @@
 ---
 '@dxos/protocols': patch
+'@dxos/client-protocol': patch
 '@dxos/client': patch
 '@dxos/client-services': patch
 ---
 
-Move the identity, contacts, devices and spaces service RPCs to buf messages. The credential
-subsystem keeps its protobuf.js shapes — the payloads cross the RPC boundary as their shared wire
-bytes — so `@dxos/client`'s public API is unchanged.
+Move the identity, contacts, devices and spaces service RPCs to buf messages, and correct
+`ClientServices` to declare the buf shapes those methods actually carry. The credential subsystem
+keeps its protobuf.js shapes — the payloads cross the RPC boundary as their shared wire bytes — so
+`@dxos/client`'s public API is unchanged.
--- a/packages/apps/composer-app/src/util/halo.ts
+++ b/packages/apps/composer-app/src/util/halo.ts
@@ -3,6 +3,9 @@
 //
 
 import { invariant } from '@dxos/invariant';
+import { buf } from '@dxos/protocols/buf';
+import { decodeCompat } from '@dxos/protocols/buf-shape-compat';
+import { CredentialSchema } from '@dxos/protocols/buf/dxos/halo/credentials_pb';
 import { type Client } from '@dxos/react-client';
 import { type Credential } from '@dxos/react-client/halo';
 
@@ -27,7 +30,8 @@ export const queryAllCredentials = (client: Client) => {
     const credentials: Credential[] = [];
     stream.subscribe(
       (credential) => {
-        credentials.push(credential);
+        // Callers index `subject.assertion` by '@type', which is the protobuf.js Any substitution.
+        credentials.push(decodeCompat(CredentialSchema, buf.toBinary(CredentialSchema, credential)));
       },
       (err) => {
         if (err) {
--- a/packages/core/mesh/messaging/src/signal-manager/utils.ts
+++ b/packages/core/mesh/messaging/src/signal-manager/utils.ts
@@ -7,7 +7,8 @@ import * as Context from 'effect/Context';
 import { invariant } from '@dxos/invariant';
 import { log } from '@dxos/log';
 import { subscribeStream } from '@dxos/protocols';
-import { DeviceKind } from '@dxos/protocols/proto/dxos/client/services';
+import { toPublicKey } from '@dxos/protocols/buf';
+import { DeviceKind } from '@dxos/protocols/buf/dxos/client/services_pb';
 import { type DevicesService, type IdentityService } from '@dxos/protocols/rpc';
 
 export const setIdentityTags = ({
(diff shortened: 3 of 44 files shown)


dmaretskyi:
`Check / test (shard=2)` was red on `e859467`. Depot's check output is empty and its jobs 404 in the Actions API (the `Check` workflow isn't in `.github/workflows`), so I reproduced locally instead. Two distinct failures, and only one of them is this PR's:

**1. Fixed — `client-e2e` delegated invitations timing out at 500ms.** `Error: Timeout [500ms]` from `waitForSpace`'s default in `packages/sdk/client/src/testing/utils.ts`. The space arrives over the `querySpaces` stream, so the wait spans a full RPC round trip; a sub-second budget times out on a contended shard rather than on a real failure. Raised to 5s in `d133ae9` — no call site asserts that the wait times out, so it only changes how long a genuine breakage takes to surface.

I checked whether my own change made that path slower, since I added a wire-bytes round-trip to the `querySpaces` handler. Measured: `fromBufSpace` costs **0.056 ms/call**, so with a handful of spaces per emission it is nowhere near a 500ms budget. Not the cause.

**2. Not this PR's — `react-ui-form` `FormFieldSet.test.tsx` "a nested object folds under its own header".** `TestingLibraryElementError: Unable to find role="button" and name "Address"`, rendering an empty `<div />`. This one is deterministic, not a flake, and it is **red on the base branch too**: I reverted `core/protocols`, `core/mesh`, `sdk/client`, `sdk/client-protocol`, `sdk/client-services` and `sdk/react-client` to `origin/main` in the working tree and it failed identically. The test was added by #12961 (the Radix → Ark UI migration), which is on `main` and predates this PR's base `cff33b7`. No fix for it exists that I'm aware of, so I have not ported one and have not widened this PR to chase it — it needs an owner on the Ark UI migration.

After the fix, `client-e2e` passes 174/174 and the repo build and `moon run :lint` are clean. Two caveats on this 4-core sandbox, stated plainly rather than glossed: the full `client-e2e` package intermittently drops a *different* test per run under load (`spaces.test.ts > post and listen to messages` on one run, green on the next; each passes 3/3 in isolation), and `client`'s `dedicated-worker-client-services.test.ts` behaves the same way. Neither reproduces in isolation and this branch touches no leader-election or gossip-delivery code, so I have not chased them further.

---

dmaretskyi:
Correcting my previous comment, and stating plainly where this is stuck.

`Check / test (shard=2)` is red again on `d133ae9`, with every other check green (`check`, `build`, `memory`, `boot-budget`, `model-fixture`, test shards 0/1/3, CodeRabbit clean).

**What I got wrong:** I offered the `react-ui-form` `FormFieldSet` failure as the explanation, "red on base". The first half holds — it does reproduce with `core/protocols`, `core/mesh`, `sdk/client`, `sdk/client-protocol`, `sdk/client-services` and `sdk/react-client` reverted to `origin/main`, so it is not this PR's. But it does not explain shard 2: #12975's shard 2 is **green** on a comparable base that also contains #12961. So that failure is local to this sandbox's happy-dom environment and evidently does not fire in CI. Shard 2 is failing on something else.

**What shard 2 is actually failing on, I can't see from here.** Depot's check output is empty (`summary`/`text`/`title` all blank), its jobs 404 in the Actions API since the `Check` workflow isn't in `.github/workflows`, `Check / report` completed as `skipped` with no output, the only `PR Build` artifact is the composer bundle, and the `trunk` MCP server (the repo's documented path for exactly this, per `.agents/skills/trunk-quarantine`) isn't authenticated in this session. I also have no way to re-run a Depot job.

**What I can verify locally is green**, all on `d133ae9`: `moon exec :build` 0 errors, `moon run :lint` clean, `oxfmt --check` clean, the full `:test` sweep with file parallelism, and `client-e2e` 4/4 on consecutive full-package runs (it was this package's delegated-invitation tests that `d133ae9` fixed).

**What I need to get further:** the shard-2 job log from the Depot link on that check, or Trunk auth for this session. Either one and I'll root-cause it. Until then I'm not going to invent a cause or push a speculative fix, and I'm keeping the PR watched rather than treating it as someone else's problem.

---

PR #13009, opened 2026-09-09
automerge: bump automerge-repo to subduction.48, automerge 3.3.2, subduction 0.16.1

mykola-vrmchk:
Moves each `@automerge/*` package to the head of the channel it tracks, and rebases both dist-level patches onto the new upstream builds.

| package | from | to |
| --- | --- | --- |
| `@automerge/automerge` | `3.3.0-fragments.2` | `3.3.2` |
| `@automerge/automerge-repo` (+ `-network-broadcastchannel`, `-network-websocket`, `-storage-indexeddb`) | `2.6.0-subduction.40` | `2.6.0-subduction.48` |
| `@automerge/automerge-subduction` | `0.16.0` | `0.16.1` |

This is the deferred lead recorded in `.agents/projects/client-startup-timeout/TASKS.md` — "rebase the dist patch `.40` → `.47`, and land the `automerge-subduction` 0.16.1 bump".

## Why `automerge` leaves the `fragments` channel

`3.3.0-fragments.2` *is* the latest `-fragments.` release — that channel ended and its work landed in the stable `3.3.x` line, which is why `automerge-repo@2.6.0-subduction.48` depends on an exact `3.3.2`. Since `overrides` pins `@automerge/automerge` to the catalog, leaving the catalog on `3.3.0-fragments.2` would have forced automerge-repo's own dependency back down to a build it was not compiled against. `3.4.1` (npm `latest`) is deliberately not used for the same reason.

## Rebasing the `automerge-repo` patch

The patch is 11 files of compiled `dist/*.js` and `dist/*.d.ts` carrying `Repo.getHandle()`, the `subduction-peer-bound` event, the `MAX_IN_FLIGHT_DOC_SYNCS` sync gate, the disconnect-abort race in `#doSync`, `detach()` entry eviction, and the `knownHashes` disk-scan mirror.

It was rebased by three-way merge (old version → patched, against new version). Nine files merged automatically; **all 247 added lines were verified present in the result**, and the rebased patch touches exactly the same 11 files as before. Two hunks conflicted, both where upstream changed the same lines:

- **`dist/index.d.ts`** — upstream added `setSubductionLogLevel` and the websocket-endpoint / worker-port export block adjacent to our two edits. Resolved as a union: upstream's new exports are kept as context, our `SubductionPeerId` / `SubductionPeerBinding` / `OnSubductionPeerBound` additions reapplied on top.
- **`dist/subduction/source.js`** — upstream replaced the websocket URL list with `WebSocketEndpointInterface` objects (`subductionWebsocketEndpoints` now takes `string | WebSocketEndpointInterface`, strings still accepted). Our `onPeerBound` reporter now wraps the endpoint loop rather than the URL loop; the reported `url` still comes from `endpoint.url`, which the merged `SubductionConnections` already derives.

## The `automerge-subduction` patch changes shape

0.16.1 removes the `bundler` wasm-bindgen target and, with it, `__wbg_set_wasm` from the web target — so upstream rewrote its browser entry (`dist/esm/bundler.js`) from a static `.wasm` import into an **async init with a literal top-level `await`**. Two things follow:

- **The old `browser` condition on `./slim` is dropped.** It pointed at a `bundler-slim.js` re-exporting `wasm_bindgen/bundler/…`, a path 0.16.1 does not ship. Retargeting it at `dist/esm/bundler.js` was wrong twice over: that file carries the top-level await, and it exports no `default` — which `composer-app/src/util/subduction-wasm.js` re-exports from `/slim`. `/slim` now resolves to upstream's `slim.js` under every condition: no init, no await, `default` intact. That is already the resolution `slimWasm` forces in `vite.config.ts`, and it is what "slim" means. (For anyone comparing: the earlier, never-merged `origin/claude/sweet-goldberg-twtikd` renamed this patch to 0.16.1 unchanged — it applies, but the re-export dangles.)
- **`bundler.js` is repointed at upstream's own `web.js`**, which performs the same initialization synchronously via `initSync`. `bundler.js` is also the `browser` condition of the package's `.` export, which `echo-host` reaches through a dynamic import, so this is needed independently of `/slim`.

Top-level await is not incidental here — `vite.config.ts` records that it trips WebKit's out-of-order evaluation under concurrent dynamic imports before Safari 27, which is why the app resolves these packages to `slim` at all. `env-tests` encodes the same constraint by building with esbuild's `iife` format, and caught both problems.

## Verification

- `pnpm install` clean; exactly one copy of each package installed (`automerge@3.3.2`, `automerge-repo@2.6.0-subduction.48`, `automerge-subduction@0.16.1`) — no duplicates.
- Both patches confirmed applied in `node_modules`.
- `moon run env-tests:test` — **2/2 passed, zero top-level-await errors** (both tests failed on the first push).
- `moon run echo-host:build` — green, 39 tasks.
- `moon run echo-host:test` — **320 passed / 2 skipped / 1 todo across 22 files**, unchanged before and after the subduction rework, including the full `SubductionPolicy` suite.
- `oxfmt --check` clean across 14,239 files.

## Downstream

`dxos/edge` carries its own copies of these patches; its matching PR is [dxos/edge#1035](https://github.com/dxos/edge/pull/1035). Edge's subduction patch is a different one (wasm re-init primitives for Workers OOM recovery) and does not touch the export map, so it is unaffected by the `browser`-condition change here.

https://claude.ai/code/session_01PCVNxwKa76gP3dQoewfn58

## Summary by CodeRabbit

* **New Features**
  * Improved document synchronization reliability during connection changes and recovery.
  * Added better visibility into peer connection status for integrations.
  * Improved browser support for subduction-based functionality.

* **Bug Fixes**
  * Prevented stale connections and synchronization entries from lingering after disconnects.
  * Improved handling of failed synchronization attempts and reconnection retries.

* **Chores**
  * Updated the Automerge integration to the latest patch releases.

---
🚀 Composer preview: https://pr-13009-composer-dev.dxos.workers.dev


mykola-vrmchk:
Re the CodeRabbit finding on `patches/@automerge__automerge-repo@2.6.0-subduction.48.patch` — "Do not release the sync gate while the sync request is still pending":

**The mechanism is real.** `#syncGate` is `semaphore(MAX_IN_FLIGHT_DOC_SYNCS)` wrapping the whole `#doSync` body, so it releases when `#doSync` returns. When `disconnected` wins the `Promise.race`, `#doSync` returns immediately but `subduction.syncWithAllPeers(...)` stays pending inside wasm with nothing cancelling it. Repeated disconnects can therefore hold more concurrent rounds than the gate nominally allows, until each abandoned call burns its own sync timeout — which is the contention the gate exists to prevent.

**It is not introduced by this PR.** The flagged block — `disconnectedSentinel`, `#disconnectAbortWaiters`, the race, and `#syncGate` itself — is byte-identical in the `2.6.0-subduction.40` patch this PR rebases. The three-way merge carried it across unchanged; the only hunks that moved in `source.js` are the `websocketEndpoints` → `WebSocketEndpointInterface` refactor and the field-declaration block. `git show HEAD~N:patches/@automerge__automerge-repo@2.6.0-subduction.40.patch` shows the same lines.

**So it is not being fixed here.** Neither available fix is a version bump: cancellation would have to be plumbed through the Rust/wasm `syncWithAllPeers`, which upstream does not expose; holding the semaphore slot until the abandoned promise settles changes backpressure behaviour under disconnect and needs its own test coverage, in a hand-maintained patch against compiled `dist/` output. Doing either inside a dependency bump would put an untested behavioural change behind a diff reviewers are reading as a rebase.

Tracked instead, with the mechanism and both candidate fixes, in `.agents/projects/client-startup-timeout/TASKS.md` (7e7276e).


---

PR #13017, opened 2026-09-09
build: enable the React Compiler (oxc) on every React plugin

dmaretskyi:
> **Stacked on dxos/dxos#12977.** Base is that PR's branch, so the diff here is only the rollout. GitHub retargets to `main` once it merges.

## What

dxos/dxos#12977 added the `reactCompiler` option and opted in one package. This turns it on for **all 92 plugins that set `jsx: 'react'`** under `packages/plugins`, and fixes the reads that memoization would otherwise freeze.

`plugin-map-solid` is untouched (`jsx: 'solid'`). Thirteen plugins set no `jsx` at all — mostly headless connectors — and are left alone; four of those (`plugin-ideogram`, `plugin-osrm`, `plugin-s3`, `plugin-typefully`) do contain a single `.tsx` file each, so they are React-ish without being configured as such. Bringing them in is a separate question about their vite config, not part of this rollout.

## The hazard, and what was fixed

The compiler caches render output keyed on prop identity. ECHO objects are **singleton proxies whose identity survives mutation**, so a component that reads a mutable field straight from its render body keeps showing the value from the render that populated the cache — anti-pattern #3 in the `reactivity` skill, and the reason dxos/dxos#12977 was opt-in rather than repo-wide.

Thirteen files now read through `useObject` / `useObjectValue` / `useResolveRef`, which subscribe:

| Plugin | Fix |
| --- | --- |
| `plugin-assistant` | Extracted a `ChatReference` component — the label was read from a `map` body, where no hook can go. |
| `plugin-kanban` (2) | Card title via a snapshot; the card is a memo boundary. |
| `plugin-space` (2) | `RecordArticle` title, `CollectionArticle`'s `ObjectTile` label. |
| `plugin-preview` (3), `plugin-library` | Destructured card fields via a snapshot. |
| `plugin-stack`, `plugin-trip` | Subscribe to the ECHO object **inside** the Mosaic tile wrapper. |
| `plugin-routine`, `plugin-tasks` | `useResolveRef` on the content/instructions ref, whose `.target` is synchronous and never re-renders on resolution (anti-pattern #4). |

Each keeps the **live** object for identity, mutation and actions, and reads only the displayed value from the snapshot.

`plugin-kanban`'s card is worth calling out: it is already `memo`'d today, so the stale title is a **live bug on `main`**, not something the compiler introduces. The compiler just extends that failure mode to everything below it.

## How the 92 were audited — and what the audit cannot see

Three passes, each narrowing on evidence rather than on the pattern:

1. Regex sweep for live-entity reads in render across all 92 → 32 plugins flagged, 67 files.
2. Read every flagged file. Discarded the classes that cannot go stale:
   - `Obj.getDatabase` / `getURI` / `getTypename` / `getMeta` — immutable.
   - `Obj.getIcon` — resolves through `IconAnnotation` on the **schema**, so type-level and immutable. `Obj.getLabel` goes through `SchemaEx.getField` on the **instance**, so it is the one that matters. (`annotations.ts` is the discriminator.)
   - `.target` on its own — mostly noise.
   → 18 files.
3. Kept only reads the compiler **newly** memoizes. A `getLabel` already inside `useMemo`/`useCallback` is cached on the same deps today, so behaviour is unchanged. Story fixtures never mutate. An `extract:` callback that is a pure function of its argument cannot go stale. `RelatedCards` only forwards `subject` to cards that now subscribe themselves.
   → **13 files, all fixed.** Re-running the audit reports only the documented false-positive classes.

**Recall is not proven.** Both regex passes are file-local, so they cannot see an imported prop type — which is exactly how `plugin-kanban` was missed on the first sweep (`KanbanCardProps` lives in `context.ts`) and only turned up when its card was read directly. So "clean" here means *no detected hazard*, not *proven safe*. And no unit test fails on staleness: the failure is a subtree that stops updating, which needs a rendered app and a mutation to see. Treat the plugin list as audited-by-inspection, and expect the residue to surface in Composer rather than in CI.

An adversarial review pass over the diff caught two of the fixes applied to the wrong value — `plugin-stack` and `plugin-trip` passed the plain Mosaic tile wrapper to `useObjectValue` instead of the ECHO object inside it, which would have thrown in render (`Obj.atom` asserts `isEntity`, so the `?? data` fallback never runs). Both are corrected here; worth knowing that this class of mistake is easy to make and invisible to the audit script.

## Verification

```
92/92 plugins with jsx: 'react' carry reactCompiler: true; plugin-map-solid: 0
```

`tsc --noEmit` per package over all nine changed plugins: **0 errors in every file this PR touches**. `oxlint` clean on all 45 changed files; `pnpm format` run.

Compiler output confirmed in dist (first wave of this rollout): `plugin-atproto` 1 file and `plugin-blogger` 3 files import `react/compiler-runtime`; `plugin-map-solid` 0. `plugin-attention` is 0 because it ships no React components.

**Not verified locally:** the full-tree build and the plugin test suites. This container's prebuilt upstream `dist` typings predate the branch base, so a number of untouched packages fail `tsc` on their own test files (`Connection.open` vs `_open`, `EchoTestBuilder.open/close`, `@dxos/test-utils` types absent after moon restored an incomplete cache archive), and those failures skip ~270 downstream tasks including every plugin build. None of it is in the diff — the per-package typecheck above is the conclusive local signal, and Depot CI builds from a clean tree.

## Notes

- The lockfile change is deliberately three lines — the one new `@dxos/echo-react` workspace link for `plugin-library`. `pnpm add` also re-resolved ~160 lines of unrelated vitest/zod peer keys that `main`'s lockfile is merely stale on; those were reverted.
- `packages/plugins/plugin-preview/package.json` lists `@dxos/echo-react` in both `dependencies` and `devDependencies`. Pre-existing, not touched here, but a dependency audit will flag it.
- Storybook remains all-or-nothing (one builder, every package's stories) and stays off. `DX_REACT_COMPILER=1` is how to view stories the way packages now actually ship.

https://claude.ai/code/session_013uTfuAof79R5nyxVRvBJ5T

---
🚀 Composer preview: https://pr-13017-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/react-compiler-all-plugins.md
+++ b/.changeset/react-compiler-all-plugins.md
@@ -1,0 +1,7 @@
+---
+'@dxos/plugin-space': patch
+---
+
+Enable the React Compiler (oxc backend) on every React plugin, and fix the reactive reads it would otherwise freeze.
+
+The compiler memoizes render output keyed on prop identity. ECHO objects are singleton proxies whose identity survives mutation, so a component that read a mutable field directly in its render body — `Obj.getLabel(subject)`, a destructured `subject.name`, a `ref.target` — would keep showing the value from the render that populated the cache. Those reads now go through `useObject`/`useObjectValue`, which subscribe.
--- a/packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatReferences.tsx
+++ b/packages/plugins/plugin-assistant/src/components/ChatPrompt/ChatReferences.tsx
@@ -6,10 +6,11 @@ import React from 'react';
 
 import { type AiContext } from '@dxos/assistant';
 import { type Database, Obj } from '@dxos/echo';
+import { useObjectValue } from '@dxos/echo-react';
 import { Icon, IconButton, type Label, type ThemedClassName, toLocalizedString, useTranslation } from '@dxos/react-ui';
-import { getStyles, mx } from '@dxos/ui-theme';
+import { mx } from '@dxos/ui-theme';
 
-import { useContextObjects } from '#hooks';
+import { type UseContextObjects, useContextObjects } from '#hooks';
 import { meta } from '#meta';
 
 export type ChatReferencesProps = ThemedClassName<{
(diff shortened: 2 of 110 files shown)


dmaretskyi:
`Check / test (node-version=24.11.1, shard=0)` failed on `864cc1a1` with exactly one test:

```
FAIL  |node| src/spaces.test.ts > Spaces > import space archive (JSON) with feed messages
Error: Test timed out in 5000ms.
 ❯ src/spaces.test.ts:717:3
```

Everything else in that shard passed — `client-e2e` was 169 passed / 1 failed, and the shard as a whole was `489 completed (371 cached), 1 failed`.

**Not this PR's.** The test is in `packages/sdk/client-services`, which this diff does not touch: no vite config, no React component, no plugin. It is a two-client space-archive import held to a 5-second wall-clock budget declared on the test itself (`{ timeout: 5_000 }`), which is what a contended runner loses. Nothing in the React Compiler rollout affects `client-services` at all — it has no `jsx` setting and ships no React.

No fix exists to port, so I have not widened the PR. I have re-run the job once (attempt #2). If it fails a second time on the same test, that is real and belongs to `client-services`, not here — the honest fix would be raising or removing that 5s budget in `spaces.test.ts`, which is a separate change.

For the record on this PR's own scope, locally: `393 tasks completed, 0 failed`, with every changed plugin's suite green — assistant 193, space 98, routine 61, tasks 57, trip 56, kanban 14, library 9, preview 1.

---

PR #13021, opened 2026-09-09
composer-app: support vite full-bundle dev mode

dmaretskyi:
Adds a `serve-bundle` task and makes `vite dev --experimentalBundle` (Vite 8's `experimental.bundledDev`) boot Composer.

```
moon run composer-app:serve-bundle
```

Bundled dev serves a Rolldown dev build instead of transforming each module on request: it reuses `build.rolldownOptions` verbatim and runs no dep optimizer. Seven things broke — but they are not all the same kind of thing, so they're split below.

## Our bugs — correct on their own terms

| Failure | Fix |
| --- | --- |
| `slimWasm` keyed its `automerge-repo` redirect on `command === 'build'`; the optimizer chunk it defers to does not exist in bundled dev | gate on "Rolldown bundles the graph" instead |
| Boot-path `codeSplitting` groups are production shaping whose per-build `reset` is `apply: 'build'`, so dev would reapply the first partition to every rebuild forever | dropped under bundled dev |
| `UNRESOLVED_IMPORT` for codegen outputs behind an indirect dependency (`react-ui-gameboard`'s `src/gen`, `react-ui-experimental`'s `src/shaders/glsl/gen`) — `^:#prebuild` stops at the first hop | named on `serve`, which also fixes the same latent failure in plain dev |

## Upstream workarounds — tape, with a cost

| Failure | Workaround |
| --- | --- |
| Rolldown's lazy dev engine drops factories (`MissingFactoryError`) and evaluates modules ahead of their dependencies across Composer's ~270 concurrent plugin imports | `experimental.devMode.lazy = false` — build the graph eagerly, as `build` does |
| `@vitejs/plugin-react can't detect preamble` — the `vite:react-refresh-fbm` hook runs and returns the preamble tag, but it never reaches the served document; a probe injecting both from one `order: 'pre'` hook showed a `<meta>` arrive and an inline `<script type="module">` not | re-serve the same code from a virtual module and reference it by `src`, which the HTML plugin links into the graph (`src/vite/react-refresh-preamble.ts`) |
| Every save failed the rebuild in `hotUpdate`: Rolldown's plugin bridge supplies neither the options object's `server` field nor the context's `environment` (`Cannot read properties of undefined (reading 'environments')`, then `(reading 'name')`) | `@tailwindcss/vite`'s hook and our own theme hook stand down there |

**Upgrading does not remove these.** Verified against latest rather than assumed: vite 8.2.2's `BundledDev` class and inline-module HTML handling are byte-identical to 8.2.1, and `@vitejs/plugin-react` 6.1.1 still injects the preamble as an inline `children` tag exactly as 6.0.5 does. Rolldown 1.2.8 is untested — the lazy engine is in the native binding, so it can't be settled by reading, and that's the one worth revisiting since `lazy: false` disables the mode's headline feature.

## Not really a vite problem

The node-builtin stubs (`src/vite/node-builtin-stubs.ts`) paper over something of ours: `@dxos/cli-util/callback` is reachable from the browser module graph through plugin-client's and plugin-connector's CLI command barrels, and production only survives it because treeshaking drops the subtree before linking. Vite's default-only `__vite-browser-external` is documented behaviour, not a bug. The real fix is narrowing those barrels — the `dxos-subpath-imports` lint's territory — and deleting this file. Left as a follow-up rather than widening this PR.

Every bundled-dev branch is gated, so production and plain `vite dev` are untouched — the one shared change is the two codegen deps added to `serve`.

## Known limitations

1. **`lazy: false`.** Lazy per-entry compilation is most of why full-bundle dev is interesting; with it off, first start is a ~15s eager build. The lazy failures were made to go away, not root-caused.
2. **Tailwind CSS regeneration.** Tailwind's `hotUpdate` is what regenerates CSS when a scanned source file changes, and it cannot run here. A **newly used** utility class needs a server restart in that mode; everything already in the generated CSS behaves normally.

## Verification

- `moon run composer-app:serve-bundle` → `VITE v8.2.1 ready in 2386 ms`, no build errors.
- Chromium boot: shell renders, `composer.startup {totalMs: 2799, …}`, **zero** console errors. Post-boot EDGE RPC timeouts in the sandbox are its lack of egress to `dxos.network`, not this change.
- HMR: editing a component logs `[vite] (client) page reload`; the page keeps rendering.
- `moon run composer-app:bundle` → `✓ built in 1m 5s`, `boot chunking: 3982 modules -> 11 chunks`.
- `moon run composer-app:serve` → boots, and plugin-react's own inline preamble is present, confirming the new plugin is inactive.
- **Boot-graph neutral**, measured rather than argued: `check-boot-budget` on this branch and again with the whole change reverted in the working tree both give `22 preload entries, 4.46 MB`. The bot's `+297 KB` advisory is base drift.
- `composer-app:build` (which is `tsc --noEmit`), `ui-theme:build`, and both packages' lint green; `pnpm format-check` clean.

https://claude.ai/code/session_01ATHu6FpZh2tTuRubpbAsxb

---
🚀 Composer preview: https://pr-13021-composer-dev.dxos.workers.dev


dmaretskyi:
The boot-graph advisory above (+297.3 KB, +7.0%) is **not from this PR** — it's base drift. Measured directly rather than inferred: `composer-app:check-boot-budget` on this branch, and then again with the entire change reverted in the working tree (the three modified files restored from `HEAD~1`, the two new `src/vite/*` files removed), gives the identical result both times.

```
this branch:      boot graph: 22 preload entries, 4.46 MB (budget: 25 entries, 4.55 MB)
change reverted:  boot graph: 22 preload entries, 4.46 MB (budget: 25 entries, 4.55 MB)
```

That is expected: every bundled-dev branch is gated on a flag only `vite dev` sets, so the production config object is unchanged. `boot chunking: 3982 modules -> 11 chunks` in both builds.

The headroom the comment reports is real, though, and worth someone's attention on its own — 0.09 MB and 3 entries against the budget.

---

PR #13038, opened 2026-09-10
plugin-debug: rebuild the runnable space template as a chess MCP server

dmaretskyi:
## What

`plugin-debug`'s one **runnable** sample space — the one a reader executes rather than reads — built a chatroom on Cloudflare Workers. Its definition of done was two browser tabs talking: a screenshot, and an app with no relationship to the Composer that produced it, so the run ended where the demo should have started.

It now builds a **chess engine exposed as an MCP server** on one Worker, and the last verification step is a chat in the same space asking that Worker for a move. The deliverable is a capability the space gained.

`src/sample/chatroom` → `src/sample/stockfish`; `ChatroomSpace` → `StockfishSpace`; preset `org.dxos.plugin-debug.sample.stockfish`, label "Chess MCP on Workers". No compatibility re-export.

## Why the reshape: nothing mid-run needs an account

The old second stage was "satisfy the prerequisites" — five steps, three of them consent screens, all before a line of code. A first run stopped at whichever one the reader did not have. Each dependency is **removed** rather than reordered:

| Old dependency | Replaced by |
| --- | --- |
| Anthropic key | DeepSeek V4 Pro through the DXOS edge — the reader's own identity is enough |
| Claude managed agent | the assistant itself, coding in a remote sandbox (`org.dxos.skill.sandbox`) |
| `wrangler login` | wrangler's unauthenticated deploy, which mints a temporary account |
| GitHub repo + token | still needed, but stage **five** — after the thing already works |

What remains is five stages, each deploying and verifying before the next starts: design → deploy an empty Worker → implement the MCP server → register it and use it from the chess chat → publish. The reader owns exactly three steps: pick the model, and the two GitHub consent screens at the end.

## The skill

The Development skill loses its "Claude managed agents" section and gains four:

- **Where the code is written** — drive the sandbox directly; do **not** delegate to the coding harness the sandbox image ships (`org.dxos.skill.deepseek`). It starts with none of the space's context — no brief, no task tree, no design document — and writes to a container filesystem instead of into the project, so the task list the reader is watching stops moving.
- **Deploying to Cloudflare** — unauthenticated first deploy, and claim the temporary account immediately, since losing it invalidates every later stage.
- **Models and credentials** — DeepSeek through the edge, no Anthropic key anywhere.
- **MCP servers** — register as `http` (Streamable HTTP, not the deprecated `sse`), the key replicates in plaintext, and a chat that doesn't list the tools has failed to *connect* rather than declined to use them.

## The seeded position

The space also seeds a chess `Game` five moves into a Spanish opening — the position the finished server gets pointed at. It is quiet and has no forced tactic on purpose: an answer the model could have produced from its own knowledge is then visibly *not* the engine's, which is exactly what stage four has to distinguish. Stage four's done-condition is a **tool call in the transcript**, not a plausible move.

Adds `@dxos/plugin-chess` and `@dxos/plugin-game` to `plugin-debug` (no cycle; precedent is the existing note at the top of `sample-spaces.ts` — sample content lives here rather than in the plugins whose types it uses).

## Not seeded, deliberately

- the design document — stage one's output; seeding it answers the question the project exists to work through
- the `McpServer` record — stage four; `url` is required and there is no URL until stage two deploys
- a repository — stage five

## Model selection is a setting, not space content

`Instructions` carries `skills` and `objects` but no model; the chat model resolves from `Assistant.Settings.modelDefaults` per provider (`usePresets`). A space therefore **cannot** select DeepSeek V4 Pro for its reader — so it is a reader-assigned task in stage one, and the project instructions state that no Anthropic key is involved anywhere, which makes a wrong model visible as a credential prompt rather than as a silent cost.

## Verification

- `plugin-debug` typechecks and the archive-shape test was updated: five stages, the three reader-assigned steps asserted **in order**, the `Game` and `Chess.State` present, no `McpServer` and no repo.
- `pnpm format` (oxfmt) clean.
- `moon run plugin-debug:test -- src/sample/stockfish` was still building the cold dependency graph when this PR was opened — **CI is the authority on it**, and I'll fix anything it turns up on this branch.
- **Not done:** an end-to-end demo video of the template being run. Executing it needs a signed-in DXOS identity for the edge model, a reachable sandbox service, and outbound access for `wrangler deploy` — the cloud sandbox this was written in has none of the three. Tracked in `.agents/projects/coding-space-template/TASKS.md` Phase 3.

Also fixes the stale "Coding Chatroom" mentions in `stories-assistant` and in the unreleased changeset that announced the old template.

https://claude.ai/code/session_018f1mHo5DitFyanTjpDv9xg

---
_Generated by [Claude Code](https://claude.ai/code/session_018f1mHo5DitFyanTjpDv9xg)_

## Summary by CodeRabbit

- **New Features**
  - Replaced the coding chatroom preset with a runnable “Chess MCP on Workers” sample space.
  - Includes a seeded chess position, project brief, development guidance, and five-stage workflow.
  - Supports building and verifying a chess MCP server on Cloudflare Workers.
  - Added chess and game capabilities to the debug experience.

- **Bug Fixes**
  - Improved handling of streamed parallel tool calls and DeepSeek responses.

- **Documentation**
  - Added design, task-tracking, deployment, and reference documentation.

- **Tests**
  - Added coverage for sample content, workflows, and parallel tool calls.

---
🚀 Composer preview: https://pr-13038-composer-dev.dxos.workers.dev


dmaretskyi:
## Verified by running it, not by reading it

The plan this template describes was executed for real, which corrected nine of its own claims. Details in [`TASKS.md`](https://github.com/dxos/dxos/blob/claude/dreamy-curie-uwzu6i/.agents/projects/coding-space-template/TASKS.md) phases 2a–2c; the Worker that produced the findings is kept in [`reference/chess-mcp/`](https://github.com/dxos/dxos/tree/claude/dreamy-curie-uwzu6i/.agents/projects/coding-space-template/reference/chess-mcp).

**The MCP server, deployed on no Cloudflare account and called over the wire:**

```
$ npx wrangler deploy --temporary        # mints an account, prints a 60-minute claim URL
$ curl -s $URL/mcp -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
{"tools":[{"name":"best_move",…},{"name":"evaluate_position",…}]}

# the exact position the template seeds (1.e4 e5 2.Nf3 Nc6 3.Bb5 a6 4.Ba4 Nf6 5.O-O Be7):
$ curl -s $URL/mcp -d '{…"name":"best_move","arguments":{"fen":"r1bqk2r/1pppbppp/…"}}'
{"bestMove":"Bxc6","scoreCentipawns":36,"depth":2,"nodesSearched":771,"truncated":false}   # 1.4s
```

**The preset in a live Composer** (dev build, `DX_DEV=true` — `plugin-debug` is enabled only under `isDev`, so a plain `serve` shows no templates at all):

| | |
|---|---|
| [Create-space dialog](https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-10-chess-mcp-template/11-template-listed.png) | "Chess MCP on Workers" listed with its description, alongside the other three |
| [The seeded space](https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-10-chess-mcp-template/07-space-created.png) | 41 objects — the project, the root task, the five stages |

`moon run plugin-debug:test -- src/sample/stockfish` → 4 passed (4). `oxfmt` clean.

## What running it changed about the template

Worth calling out because each was wrong in a way reading could not reveal:

- **`--temporary` is not optional.** Plain `wrangler deploy` refuses non-interactively and demands a token; its own error names the flag. The task said wrangler "deploys with no account by minting a temporary one" — true and unactionable.
- **`Date.now()` does not advance during synchronous compute in a Worker**, so the bounded search the brief demands cannot be bounded by wall-clock: the deadline never fires and the call reports zero elapsed. Bounded by node count now.
- **The claim URL is a bearer credential** and the template told the runner to file it as a project artifact — which replicates in plaintext to everyone in the space, contradicting the skill's own rule. Worker URL is filed; claim URL is not.
- **Claiming must come *after* the last redeploy.** A claimed account is one the unauthenticated agent can no longer update, and `--temporary` refuses when logged in — so claiming early would mint a second account under a different URL and break the registration stage four performs.
- **Stage four was unreachable in a default Composer**: `ChessPlugin` is enabled in no default list, so the seeded `Game` has no board to open.
- **`Project.make` builds an `Outline` that none of the three sample spaces declared** — found by replacing two vacuous test assertions (`countOf` substring-matches, so a zero-count passes when the typename is misspelled) with one derived from `definition.schemas`.

## The demo video is not here, and the reason is measured

Composer's shared worker has a **fixed 15-second deadline** to open its leader session. ffmpeg capture takes ~60% of a core, of four, and the browser then misses it:

```
WorkerConnectionError: Worker connection timed out after 15000ms: opening worker leader session
client services failed to open → fatal dialog
```

With no capture running the identical walkthrough boots in 24s and completes — which is how the screenshots above were taken. So it is recording that breaks the app here, not the app and not this change.

Six attempts; three of them found other real causes, fixed on the way: orphaned chromium from killed runs starving the box to load 9, capture competing with a `moon` dependency-graph rebuild, and — after `main` was merged into this branch — a long-running dev server serving a vite pre-bundle of `@dxos/protocols` from before #13036, so the browser reported an export missing from a `dist` that was correct on disk (clearing `node_modules/.vite` fixed it; rebuilding the package would not have).

The recorder and that write-up are in [`reference/demo/`](https://github.com/dxos/dxos/tree/claude/dreamy-curie-uwzu6i/.agents/projects/coding-space-template/reference/demo). A machine with more cores produces the video with no change to the template.

**One thing for a maintainer:** `chess-mcp-demo.dxos.workers.dev` is currently deployed on the DXOS org Cloudflare account — `CLOUDFLARE_API_TOKEN` is set in this sandbox, so my first deploy landed there before I noticed. I deleted that one and confirmed `--temporary` works independently, but the live demo Worker is back on the org account because the temporary account's `workers.dev` host is behind a managed challenge from this container's IP. Delete it whenever you like; nothing in the PR depends on it.

---

PR #13041, opened 2026-09-10
protocols, lint: make every Any producer name its type_url convention

dmaretskyi:
Follow-up to #13036, which fixed the production regression but left the underlying hazard in place.

## The hazard

`anyPack` writes the spec form `type.googleapis.com/<name>`. Both forms are *required* at the EDGE
boundary, by different readers:

- **Credential assertions want bare.** EDGE keys its assertion registry by `schema.typeName` and
  emits bare itself. A prefixed url misses the lookup — that was #13036's 500, 24,149 times.
- **Service messages want prefixed.** EDGE's router (`worker/router.ts`) and `messenger-client`
  both read the type out with `split('/')[1]`, which yields `undefined` for a bare url.

Neither reader errors on the wrong form. One 500s in production; the other silently produces
`undefined`. And buf normalizes on read (`typeUrlToName` slices at the last `/`), so an emitter that
picks the wrong convention keeps working against every buf consumer and fails only against the
hand-rolled ones. There is no type error and no test failure — which is exactly how #13036 shipped.

So the fix is not to ban a *form*. It is to ban the *silent default*.

## Change

`@dxos/protocols/buf` now exports both conventions as named helpers, and raw `anyPack` is a lint
error:

```ts
anyPackBare(desc, message);     // registries keyed by the bare type name
anyPackPrefixed(desc, message); // readers that strip a `type.googleapis.com/` prefix
```

`@dxos/rules/no-raw-any-pack` catches both the bare call and the `bufWkt.anyPack(…)` namespace form.
The single `eslint-disable` is in `packages/core/protocols/src/buf/index.ts`, where the two wrappers
are defined — the one place that legitimately calls through to buf's default.

All nine existing call sites were migrated to whichever helper **preserves current behaviour**.
`anyPackPrefixed` is a pass-through, so there is no wire change anywhere in this PR.

One exception, and it is the reason the rule earns its keep: the rule surfaced a **third credential
assertion emitter** the #13036 audit missed — `plugin-script/src/util/functions.ts:86`, which builds
a `ServiceAccess` assertion. It is now `anyPackBare`, matching the other two emitters. It is reached
only from a story today, so this was latent rather than live.

Also fixes a doc comment #13036 orphaned: the `Struct`/gossip block had drifted off `packJson`.

## Validation

- `oxlint packages/ tools/` — **0** `no-raw-any-pack` violations, no errors in any file touched here.
  (The `tools/qa-lint` errors in that output are pre-existing — identical with these changes stashed.)
- `moon run credentials:test config:test rpc:test teleport-extension-gossip:test` — **147 passed**,
  exit 0.
- `pnpm knip` — exit 0.
- `pnpm format` clean.

## Not in this PR

EDGE normalizing on read (`typeNameOf` in its `getAssertionType`) is the change that makes this class
of bug impossible rather than merely currently-absent, but it lives in `dxos/edge` and needs a deploy.
It is in flight there alongside the pin bump off the pre-migration snapshot.

Context: `.agents/projects/protobufjs-to-buf/EDGE-ASSERTION-TYPE-URL.md`.

---

---
🚀 Composer preview: https://pr-13041-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/any-pack-convention-lint.md
+++ b/.changeset/any-pack-convention-lint.md
@@ -1,0 +1,16 @@
+---
+'@dxos/echo': patch
+---
+
+`Any` producers now name their `type_url` convention explicitly.
+
+`@dxos/protocols/buf` exports `anyPackPrefixed` alongside `anyPackBare`, and a new
+`@dxos/rules/no-raw-any-pack` lint rule rejects raw `anyPack`. Both forms are required at the EDGE
+boundary by different readers — credential assertions resolve by bare type name, while the router and
+messenger strip a `type.googleapis.com/` prefix — and neither reader errors on the wrong form, so the
+choice now has to be made at the call site rather than inherited from buf's default.
+
+Existing call sites moved to whichever helper preserves their current behaviour, so there is no wire
+change. One exception: `@dxos/plugin-script`'s `getAccessCredential` packed its `ServiceAccess`
+assertion with the prefix, which EDGE's registry does not resolve; it now emits the bare form like
+every other assertion emitter.
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -46,6 +46,7 @@
     "@dxos/eslint-plugin-rules/no-bare-dot-imports": "error",
     "@dxos/eslint-plugin-rules/no-dead-tailwind-logical": "error",
     "@dxos/eslint-plugin-rules/no-effect-run-promise": "error",
+    "@dxos/eslint-plugin-rules/no-raw-any-pack": "error",
     "@dxos/eslint-plugin-rules/operation-key-shape": "error",
     "@dxos/eslint-plugin-rules/prefer-sizing-utilities": "warn",
     "@dxos/eslint-plugin-rules/no-empty-promise-catch": "error",
--- a/packages/common/eslint-plugin-rules/index.js
+++ b/packages/common/eslint-plugin-rules/index.js
@@ -16,6 +16,7 @@ import noBareDotImports from './rules/no-bare-dot-imports.js';
 import noDeadTailwindLogical from './rules/no-dead-tailwind-logical.js';
 import noEffectRunPromise from './rules/no-effect-run-promise.js';
 import noEmptyPromiseCatch from './rules/no-empty-promise-catch.js';
+import noRawAnyPack from './rules/no-raw-any-pack.js';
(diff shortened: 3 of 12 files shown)


dmaretskyi:
`Check / check` is red on `d222060077`, and it is not this PR's failure. Standing down on it rather than widening this PR — details and a proposed patch below.

**What is failing.** The `knip` step, on a file this PR does not touch:

```
Unused files (1)
```

**Why it is not this PR's.** `git diff origin/main...HEAD -- packages/reflect/` is empty — this PR touches nothing in that package. The same failure hit #13036 earlier on head `aec0c30` with an entirely unrelated diff, so it is not diff-specific. Neither commit landed on `main` since this branch's base (`a39bab8bf5`, `88e669c777`) touches `packages/reflect/` either.

**Root cause.** `packages/reflect/deus/moon.yml:12` runs `lezer-generator --typeScript ./src/extension/mdl.grammar -o ./src/extension/gen/mdl.ts`. lezer emits **two** files and `moon.yml` lists both as outputs, but only one is reachable: `syntax.ts:7` does `import { parser } from './gen/mdl'`, and nothing imports `mdl.terms.ts`. Both are committed, so knip is correct — it is genuinely an unused file, introduced with the package in #11368.

**Proposed patch** (not applied here — it is a different package and belongs on `main`, not bolted onto a lint-rule PR). Prefer ignoring the generated directory over deleting the file, because `lezer-generator` recreates `mdl.terms.ts` on every codegen run, so a deletion regresses the next time the grammar is rebuilt:

```ts
// .config/knip.ts — generated parser output; lezer emits mdl.terms.ts as a side-artifact of mdl.ts.
ignore: ['packages/reflect/deus/src/extension/gen/**'],
```

**One caveat on my own evidence.** `pnpm knip` passes locally (exit 0, no findings) on this exact head, so I could not reproduce it. CI runs knip after a full build, and `.config/knip.ts` derives entry points from vite configs and build outputs, so my hypothesis is that an unbuilt `deus` is skipped locally — but I have not confirmed that, and I would rather flag the gap than present a local pass as proof.

I have not re-run the job: the failure is deterministic rather than flaky, so a re-run would only reproduce it.

Everything else on this head is green — `changeset-reminder`, `Model Fixture`, `boot-budget`, `memory`, and test shards 1–4.

---

PR #13043, opened 2026-09-10
deps: collapse the duplicate rolldown and align the tailwind catalog

dmaretskyi:
Two catalog entries resolved to versions their consumers should have been sharing. Found while checking whether newer vite/rolldown would remove the workarounds in #13021.

## rolldown — two copies, and one of them was load-bearing

`rolldown` was pinned exactly at `1.1.5` while vite 8.2.1 depends on `~1.2.1` and resolved `1.2.4`, so the workspace carried two copies with separate native bindings:

```
tools/vite-plugin-log → rolldown 1.1.5
packages/sdk/shell    → rolldown 1.1.5
vite 8.2.1            → rolldown 1.2.4
```

That is not just untidy. `@dxos/vite-plugin-log` imports `RolldownMagicString` **at runtime** (`src/plugin.ts`, `src/transform.ts`) and hands those objects into a build driven by vite's rolldown; `@dxos/shell`'s `vite.bundle.config.ts` imports `esmExternalRequirePlugin` from `rolldown/plugins` into the same build. Both crossed the version boundary. I have not seen it fail, so this is a latent hazard rather than a live bug — but it is the same dual-package class that `composer-app/vite.config.ts` already carries an alias workaround for (`@dxos/solid-ui-geo`).

`~1.2.4` intersects vite's own `~1.2.1`, so the lockfile now carries a single `rolldown@1.2.4` and all three importers resolve it. A tilde rather than another exact pin because the constraint being expressed is "stay inside vite's `~1.2.x`" — an exact pin silently re-breaks on the next vite bump.

## Tailwind — a skew that reached consumers

`@tailwindcss/postcss` sat at `^4.3.0` against `@tailwindcss/vite`'s `^4.3.3`, resolving 4.3.0 and 4.3.3 respectively. Both are **direct dependencies** of `@dxos/ui-theme` (not dev), and Tailwind ships them in lockstep, so the skew shipped. Now both `^4.3.3`.

## After

```
lockfile: single rolldown@1.2.4, single @tailwindcss/postcss@4.3.3
vite-plugin-log → 1.2.4    shell → 1.2.4    vite → 1.2.4
ui-theme: @tailwindcss/postcss 4.3.3, @tailwindcss/vite 4.3.3
```

## Verification

- `moon run shell:bundle` → `✓ built in 6.87s` — the task that actually calls `rolldown/plugins`, so the real test of the version change.
- `vite-plugin-log:build` + `vite-plugin-log:test` (the `RolldownMagicString` consumer), `ui-theme:build`, `composer-app:build` — 380 tasks, all green.
- `pnpm format-check` clean.

## Noticed, deliberately not touched

`@tailwindcss/oxide: ^4.3.0` is a catalog entry no workspace package declares, so editing it changes nothing and *removing* it is the actual cleanup — worth its own look rather than riding along here. Likewise `rollup-plugin-visualizer` (two majors behind), `oxc-resolver`, and `vite` 8.2.1→8.2.2 are upgrades rather than pinning defects, each deserving its own decision.

https://claude.ai/code/session_01ATHu6FpZh2tTuRubpbAsxb

---
_Generated by [Claude Code](https://claude.ai/code/session_01ATHu6FpZh2tTuRubpbAsxb)_

---
🚀 Composer preview: https://pr-13043-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/align-tailwind-catalog.md
+++ b/.changeset/align-tailwind-catalog.md
@@ -1,0 +1,5 @@
+---
+'@dxos/ui-theme': patch
+---
+
+`@tailwindcss/postcss` and `@tailwindcss/vite` now resolve to the same Tailwind version. Both are direct dependencies and Tailwind ships them in lockstep, so the previous caret ranges let them drift apart.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -415,8 +415,8 @@ catalogs:
       specifier: ^4.3.0
       version: 4.3.0
     '@tailwindcss/postcss':
-      specifier: ^4.3.0
-      version: 4.3.0
+      specifier: ^4.3.3
+      version: 4.3.3
     '@tailwindcss/vite':
       specifier: ^4.3.3
       version: 4.3.3
@@ -1336,8 +1336,8 @@ catalogs:
       specifier: ^5.1.0
       version: 5.1.0
     rolldown:
-      specifier: 1.1.5
-      version: 1.1.5
+      specifier: ~1.2.4
+      version: 1.2.4
     rollup:
       specifier: ^3.29.5
       version: 3.30.0
@@ -2901,7 +2901,7 @@ importers:
     devDependencies:
       '@tailwindcss/postcss':
         specifier: 'catalog:'
-        version: 4.3.0
+        version: 4.3.3
       '@types/react':
         specifier: ~19.2.17
         version: 19.2.17
@@ -18850,7 +18850,7 @@ importers:
         version: 19.2.7(react@19.2.7)
       rolldown:
         specifier: 'catalog:'
-        version: 1.1.5
+        version: 1.2.4
       vite:
         specifier: '>=8.2.1'
         version: 8.2.1(@types/node@22.10.2)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.9.0)
@@ -23159,7 +23159,7 @@ importers:
         version: 0.5.11(tailwindcss@4.3.0)
       '@tailwindcss/postcss':
         specifier: 'catalog:'
-        version: 4.3.0
+        version: 4.3.3
       '@tailwindcss/vite':
         specifier: 'catalog:'
         version: 4.3.3(vite@8.2.1(@types/node@22.10.2)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.9.0))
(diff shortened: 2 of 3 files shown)


dmaretskyi:
`Check / test (node-version=24.11.1, shard=0)` is red, and it isn't this PR's failure.

Two `@dxos/client-e2e` P2P timeouts (`Tasks: 448 completed (376 cached), 1 failed`):

```
FAIL src/spaces.test.ts > Spaces > post and listen to messages
  Error: Timeout [30,000ms]  at src/spaces.test.ts:161

FAIL src/invitations.test.ts > Invitations > ServiceContext > space > with shared secret
  Error: Test timed out in 15000ms.  at src/invitations.test.ts:132
```

Trunk already quarantined the first as a known flake; the second has its own flaky-test record but wasn't quarantined, so it takes the shard down (`⚠️ Some test failures were not quarantined, using exit code: 1`).

Why it can't be this diff: the change is two catalog version entries — `rolldown` (a devDependency of `@dxos/shell` and the private `@dxos/vite-plugin-log`) and `@tailwindcss/postcss` (a dependency of `@dxos/ui-theme`). `@dxos/client-e2e` depends on none of those, directly or in moon's graph:

```
$ node -e "…client-e2e/package.json…"   → direct hits: (none)
$ moon project client-e2e | grep 'Depends on'  → (empty)
```

These are swarm/invitation timeouts in node tests — no vite build, no Tailwind, no rolldown anywhere in their path.

Spending the one allowed re-run on that shard to confirm. No code change; if it fails a second time I'll treat it as real and dig in rather than re-running again.

---

dmaretskyi:
Dequeued from the merge queue (`CI_FAILURE`) on the same shard, but a **different** `client-e2e` test again. Three runs, three different failures:

| run | unquarantined failure | result |
| --- | --- | --- |
| PR attempt 1 | `invitations.test.ts > … > with shared secret` | red |
| PR attempt 2 | — (only the quarantined `spaces > post and listen`) | **green** |
| queue run `0m1nm00djx` | `spaces.test.ts > Spaces > space properties are reactive` | red → dequeued |

A dependency change that actually broke something fails the *same* test every time. Different P2P timeouts each run is flakiness.

Confirmed locally on this branch:

```
$ moon run client-e2e:test
 Test Files  1 failed | 13 passed | 1 skipped (15)
      Tests  1 failed | 173 passed | 6 skipped (180)
 FAIL src/spaces.test.ts > Spaces > post and listen to messages
```

The only test that fails locally is the one Trunk has **already quarantined** — and quarantine records are repo-wide, so it predates this PR. Both tests that took CI down here pass locally.

So the blocker isn't this diff, it's that `@dxos/client-e2e`'s space/invitation suite is flaky with only one member quarantined, which will bounce anything out of the queue on roughly a coin flip. Re-queuing; the durable fix is quarantining or stabilising `spaces.test.ts > space properties are reactive` and `invitations.test.ts > … > with shared secret`, which needs someone with Trunk access.

---

PR #13048, opened 2026-09-10
assistant-evals: the chess-MCP template as an end-to-end eval, and what running it fixed

wittjosiah:
## What

An eval that delegates the chess-MCP template's three agent-owned stages to a session and grades the deployment, not the transcript: it speaks MCP to the Worker URL the session filed, from the session's own sandbox, and holds the best-move answer to the rules of chess on the seeded position. Both models run through the same harness; DeepSeek V4 Pro goes through preview EDGE authenticated as the run's identity, with no key.

Latest results, one session each, on preview:

| scorer | Claude Opus 5 | DeepSeek V4 Pro |
| --- | --- | --- |
| design filed, with diagram | 1 | 1 |
| shell used | 1 | 1 |
| Worker URL filed, claim URL not | 1 | 1 |
| MCP handshake lists two tools | 1 | 1 |
| best move is legal | 1 | 1 |
| evaluation returns a score | 1 | 1 |
| delegated stages in review | 3 of 3 | 3 of 3 |
| reader and later work left alone | 0.5 | 0.5* |
| no tool errors | 0 | 0* |
| served the engine fast | 0.31 (46 min) | 0.75 (26 min) |

Both models now serve a working engine. Claude took 46 minutes and shipped Stockfish compiled to wasm on the Worker, answering `a4c6` at depth 12. DeepSeek took 26 minutes with a search of its own over chess.js, answering the same `a4c6` at depth 4 after 3,440 nodes. On the seeded position both are Bxc6, legal.

\* Two DeepSeek marks were the eval's fault and are fixed since that run: it registered neither the file nor the blob type the sandbox's download tool persists, so two downloads errored, and it docked the session for marking stage four `blocked` on the reader's two steps, which is what the template's Development skill tells it to do. Rerun, DeepSeek would score 9.25 of 10.

## Speed is scored

The runner records the session's wall clock, and the eval's last scorer, `served-the-engine-fast`, gives full marks for a working best-move tool within 15 minutes, falling linearly to nothing at the 60-minute budget, and zero for a server that does not answer, so a fast failure earns nothing. Fifteen minutes is a guess: the work is a design, an empty deploy and a small MCP server.

Where the time goes, from the timelines the eval records. Claude: about 18 minutes of model thinking across 119 tool calls, the rest in the sandbox, dominated by the Stockfish wasm build and a handful of two-to-four-minute compile and test commands. DeepSeek: 21 minutes of model thinking across 100 calls and five minutes of sandbox time; a search over chess.js needs no build. The two levers that would move Claude most: the engine choice, and the ten calls spent guessing the MCP server's typename before a types query found it. Whether the brief should steer away from Stockfish is a call for the template and is not changed here.

## What running it fixed

Every one of these was found by a session failing and is covered by a test where a test can reach it.

- **`@dxos/ai`**: the chat-completions adapter sends an assistant turn's reasoning back to DeepSeek as `reasoning_content`, on every tool-calling turn, empty included. DeepSeek's thinking mode refuses a request without it, and it ended every DeepSeek session at its first reasoned tool call. The direct testing preset also retries a transient provider response, and a rejected request logs the shape of what was sent.
- **`@dxos/plugin-sandbox`**: `Exec` accepts a quoted `timeout` and defaults it to five minutes rather than the service's two, and a multi-line command travels to the sandbox as base64. The sandbox service loses a heredoc's framing, which left `cat` waiting until the timeout and then held every later command on that sandbox; one session lost 35 of its 50 minutes to it.
- **`@dxos/plugin-projects`**: delegation binds the sandbox skill alongside planning, markdown and project.
- **`@dxos/plugin-debug`**: the chess template's Development skill says how to get a wrangler with `--temporary` under the sandbox's Node 20, where a plain install resolves to a version that predates the flag.
- **`@dxos/assistant-evals`**: the runner takes a client config, routes DeepSeek models through EDGE, builds skills per run so concurrent variants do not share a skill object, lets a query invoke an operation, records the session's wall clock, and grades a session that timed out or failed rather than throwing, with a timeline of its tool calls in the output. The eval tasks are servers to moon, so a session's hour is never cut.

## Workarounds to remove once dxos/edge#1052 is on preview

That PR fixes the sandbox service itself: heredocs, a command that outlives its timeout holding the sandbox, and Node 22 with wrangler in the image. When it lands, the base64 transport in `plugin-sandbox` and the wrangler paragraph in the Development skill come out.

## Still open

- Claude went on into stage four and registered the MCP server, which was not delegated, and spent ten tool calls guessing the server type's name before a types query found it. Whether stage four's agent-owned steps should be delegated is a design call; the typename friction is real.
- `bindDelegationContext` does not bind a project's instructions, so the eval binds the space's Development skill itself; the app's delegated sessions do not get it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved DeepSeek thinking-mode conversations so tool-calling sessions continue reliably.
  - Added retries for transient AI provider errors.
  - Improved sandbox execution of multi-line commands and scripts.
  - Sandbox commands now accept quoted timeout values and default to five minutes.

- **New Features**
  - Delegated project sessions now automatically include sandbox access for build and run tasks.
  - Added named catalog entries for supported AI models.
  - Added richer assistant evaluation scoring and timing results.

- **Documentation**
  - Clarified sandbox timeout behavior and deployment requirements.

---
🚀 Composer preview: https://pr-13048-composer-dev.dxos.workers.dev


packages/core/compute/assistant-evals/src/evals/chess-mcp.eval.ts
```diff
@@ -0,0 +1,433 @@
...
+const VARIANTS = selected ? MODELS.filter(({ name }) => selected.includes(name)) : MODELS;
+
+evalite.each(VARIANTS)('Chess MCP — a delegated session designs, deploys and serves a chess engine over MCP', {
+  data: [{ input: null }],
+  task,
+  scorers: [
+    {
+      name: 'design-filed-with-diagram',
+      description: 'A design artifact is on the project and carries a mermaid diagram of the request path.',
+      scorer: ({ output }) =>
```
dmaretskyi: I hate how scorers leak into the test. you need to make db query in the main body and add a scorer, lets just pass the db hande with the output and let each scorer make it own queries. Also create a reusable scorer dsl (import-as-namespace):

Scorer.make({ ... })

Scorer.database({
  name: ....
  desicription: ....
  query: ...  Query.select(....),
  scorer: results => results.length > 0
}),

...other ideas you can extract: tool calls, duration, etc...

packages/core/compute/assistant-evals/src/evals/chess-mcp.eval.ts
```diff
@@ -0,0 +1,433 @@
...
+    }),
+});
+
+/**
+ * The models a delegated coding session is measured on. DeepSeek V4 Pro is what the template tells
+ * its reader to select, and it is served through EDGE with the run's own identity, so it needs no
+ * key; Opus goes to Anthropic directly and needs `DX_ANTHROPIC_API_KEY`.
+ */
+const MODELS = [
+  { name: 'claude-opus-5', input: { model: DXN.make('com.anthropic.model.claude-opus-5.default') } },
```
dmaretskyi: can this instead import actual model instaces from @dxos/ai?

packages/core/compute/assistant-evals/src/evals/chess-mcp.eval.ts
dmaretskyi: Done — `src/Scorer.ts` is that DSL, and the eval's `dbQuery` is gone.

Each dimension now carries its own query, run against the space while it is still open (the runner scores in-scope, before the harness is disposed, and reports the query's value next to the mark):

- `Scorer.make({ name, description, query, score })` — any effect over the live space.
- `Scorer.database({ name, description, query: Query.select(…), score: (results) => … })` — as you wrote it.
- `Scorer.toolCalls({ … score: (invocations) => … })` — the session's tool calls paired with their results.
- `Scorer.duration({ targetMinutes, budgetMinutes, when, delivered })` — the wall clock, gated on the thing being timed having actually been produced.

`Scorer.once(key, effect)` shares what several scorers need (the artifact text, the MCP handshake with the deployed Worker) — memoized per run via a `Memo` service rather than at module scope, since variants run concurrently in one process.

`createEvalRunner` takes `scorers`, and `Scorer.toEvalite(SCORERS)` is the evalite side, so the list is declared once.

---

packages/core/compute/assistant-evals/src/evals/chess-mcp.eval.ts
dmaretskyi: Yes — `@dxos/ai` now names the models EDGE serves (`Model.claudeOpus5`, `Model.claudeSonnet5`, `Model.claudeHaiku45`, `Model.deepseekV4Flash`, `Model.deepseekV4Pro`), and `Model.all` is built from those entries rather than restating them. The eval's variants are `[Model.claudeOpus5, Model.deepseekV4Pro]`, taking the variant name from `model.backend` and the input from `model.id`.

---

packages/core/compute/assistant-evals/src/evals/chess-mcp.eval.ts
```diff
@@ -0,0 +1,433 @@
...
+        delegatedInReview: delegated.filter(
+          (candidate) => candidate?.status === 'review' || candidate?.status === 'done',
+        ).length,
+        delegatedTotal: delegated.filter(Boolean).length,
+        // The model choice, the claim and the GitHub steps are the reader's, and cannot be done here.
+        readerStepsUntouched: readerSteps.length > 0 && readerSteps.every((step) => step?.status === 'todo'),
+        // The steps of stages four and five, and the stages themselves, were not delegated. Marking
+        // one blocked on the reader is the Development skill's own instruction, not work on it.
+        laterStagesUntouched:
+          later.length > 0 && later.every((step) => step?.status === 'todo' || step?.status === 'blocked'),
```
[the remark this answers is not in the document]
dmaretskyi: Not changing this one: accepting `blocked` here is deliberate. The Development skill tells a session to mark an undelegated stage blocked on the reader and name what unblocks it, so `blocked` on those steps is the instructed outcome, and requiring `todo` would penalize exactly the behaviour being asked for. The steps this predicate looks at are the ones the session was not given — stages four and five and their children — so "it did work first and then blocked it" is not a path that earns marks elsewhere.

---

packages/core/compute/ai/src/resolvers/ChatCompletionsAdapter.ts
```diff
@@ -814,6 +841,12 @@ export const make = (model: string, requestOptions: RequestOptions = {}) =>
                   ) as Effect.Effect<never, any, never>;
                 }
 
+                // A client that rejects a non-2xx response itself lands here with the body in the
+                // error; the shape of what was sent goes next to it, as below.
+                log.warn('chat completions request failed', {
+                  error: describe('request failed', err),
+                  messages: describeMessages(messages),
```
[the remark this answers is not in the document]
dmaretskyi: Not changing this one. These are `log.warn` diagnostics on a failed or rejected provider call, and the rejection body is the only thing that says *why* the provider refused — replacing it with `bodyLength` removes the entire reason this logging was added. Nothing here is a credential: the request is authenticated by a header that is not logged, and the body logged is the provider's error response, capped at 500 characters. Prompt content reaching a dev-facing warn log on a failed request is the same exposure the error thrown from this path already carries.

---

PR #13063, opened 2026-09-11
assistant-evals: sum the turns, and fail on an empty test discovery

dmaretskyi:
Follow-up to #13059, carrying the two review findings that could not be pushed before that PR merged (the branch was locked in the merge queue).

- `claude-harness.ts`: the run's duration was measured from before the scenario body to the end of the last turn, so a `Scorer.duration` graded the harness's own scaffolding and the between-turn queries as the agent's time. Each turn is now timed and the turns summed.
- `assistant-evals/moon.yml`: drop `--passWithNoTests`. The package has exactly one test file, so discovering none is a config regression — the exact failure mode the `ts-test` tag change in #13059 introduced — rather than a shard with nothing to run.

https://claude.ai/code/session_01Bce52MMJNMucFANFASmpLQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bce52MMJNMucFANFASmpLQ)_

---
🚀 Composer preview: https://pr-13063-composer-dev.dxos.workers.dev

### DIFF
--- a/packages/core/compute/assistant-evals/moon.yml
+++ b/packages/core/compute/assistant-evals/moon.yml
@@ -8,7 +8,9 @@ tasks:
   # `--project=node`, and this package's vitest config has to stay flat for evalite's file discovery
   # to find the scenarios at all, so it defines no projects and the filter matches nothing.
   test:
-    command: pnpm exec dx-killorphans vitest run --passWithNoTests
+    # No `--passWithNoTests`, unlike the inherited task: this package has exactly one test file, so
+    # discovering none is a regression in the config rather than a shard with nothing to run.
+    command: pnpm exec dx-killorphans vitest run
     deps:
       - esbuild-plugins:build
       - node-std:build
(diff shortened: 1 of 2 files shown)


dmaretskyi:
`Check / check` failed on `8999670` in the `knip` step, on a file this PR does not touch:

```
Unused files (1)
```

That is `packages/reflect/deus`'s lezer-generated parser output (`prebuild-lezer` in its `moon.yml`), and nothing in the repo imports it — `syntax.ts` imports only `./gen/mdl.ts`. This PR's diff is two files in `assistant-evals` (per-turn duration accumulation, and dropping `--passWithNoTests`), neither of which can make a file in `deus` unreferenced.

Run locally on this exact branch, both knip passes are clean:

```
$ node_modules/.bin/knip --config .config/knip.ts --no-config-hints           # exit 0, no output
$ node_modules/.bin/knip --config .config/knip.ts --no-config-hints \
    --production --strict --exclude binaries,files                            # exit 0, no output
```

So the failure depends on the state of the generated `gen/` directory after `prebuild-lezer` runs in CI, not on this diff. There is no fix elsewhere to port. I have spent the single permitted re-run of the job (attempt #2); if it fails again the root cause is in `deus`'s prebuild and belongs in its own change rather than widening this PR.

---

PR #13065, opened 2026-09-11
mcp: fix typename queries, task-set load timeouts, and getOutline type guard

dmaretskyi:
Fixes three defects in the operation surface the MCP server projects. Each has a regression test that fails without its fix.

## Defect 1 — `space.queryObjects` typename filter under-returned, nondeterministically

`packages/plugins/plugin-space/src/operations/query-objects.ts`

`resolveType` resolved the caller's typename to **one** registered `Type` entity and built the filter from that entity's identity — a versioned `dxn:` for a static declaration, an `echo:` id for a copy persisted in the space. `Filter.type`'s own docs say a bare typename must be wrapped as `DXN.make(typename)`; passing a resolved entity instead pins the filter to a single registration, so objects of the same typename written under any *other* registration simply do not match. Which registration got picked depended on the order the registry query happened to return, which is why the same call answered with every object, some of them, or none, run to run.

Now filters on `Filter.type(DXN.make(typename))`. The registry is still consulted, but only to reject a typename nothing declares — it no longer decides what matches.

The reported correlation with `limit` was coincidence: the query planner's limit pushdown is a plain prefix slice and is sound here. The variable was the registration, not the limit.

Regression test (`query-objects.test.ts`) persists a second registration of a typename that already has a static declaration and asserts the query still returns every object. Without the fix it returns `[]`:

```
AssertionError: expected [] to have a length of 5 but got +0
```

## Defects 2 and 3 — `projects.get` and `tasks.list` time out

`packages/sdk/types/src/types/TaskSet.ts`

Both go through `TaskSet.loadRefs`, which resolved refs **sequentially**, each `Database.load` a separate indexed query. An unresolvable ref does not fail — the resolver runs a query that finds nothing and waits out its own 30s timeout — so one dangling entry stalls the whole read past any caller's deadline.

This is not a cycle: the `parentTask` / `dependsOn` walks (`effectiveMilestoneIds`, `subtree`, `collectSubtree`, `resolveParentTask`) are already cycle-safe, each carrying its own `visited` set. It also explains why an *empty* task set answered instantly — `forEach` over `[]` does no work.

`loadRefs` now reads materialized refs from the working set, resolves the cold remainder at bounded concurrency, and treats a ref that does not resolve within 5s as gone.

Regression test (`get-project.test.ts`) puts an unresolvable member in a task set and asserts `projects.get` completes with the right counts.

## Related — `tasks.getOutline` on a non-outline

`packages/plugins/plugin-tasks/src/operations/get-outline.ts`

A ref to any other object reached `outline.content` as `undefined` and threw `TypeError: Cannot read properties of undefined (reading 'tryLoad')`. Now guarded, failing with a typed `InvalidOperationInput` naming the actual typename.

## Also

Adds `Database.makeRef`, the Effect wrapper around `db.makeRef`, matching `Database.add` / `Database.load`.

## Not addressed

The suggested fourth fix — surfacing a request id from `invokeOperation` on failure — is out of scope here; it belongs in the MCP transport rather than these three operations. Worth doing separately.

## Test output

```
plugin-space:test    Tests  3 passed (3)
plugin-tasks:test    Test Files  18 passed (18)
                     Tests  72 passed (72)
plugin-projects:test Test Files  18 passed (18)
                     Tests  49 passed | 1 skipped (50)
types:test           Test Files  7 passed (7)
                     Tests  54 passed (54)
```

`pnpm format` run before commit.

https://claude.ai/code/session_01FV1B7Ve3LUssU45wNgiXgF

---
_Generated by [Claude Code](https://claude.ai/code/session_01FV1B7Ve3LUssU45wNgiXgF)_

## Summary by CodeRabbit

- **New Features**
  - Added a public API for creating database references.

- **Bug Fixes**
  - Type-filtered object queries now return all matching objects consistently.
  - Project and task reads handle missing or unresolved references without stalling.
  - Invalid outline references now return a clear, typed validation error.
  - Reference loading is more resilient, including timeouts and failed lookups.

- **Tests**
  - Added coverage for type filtering, dangling references, reference loading, and invalid outline inputs.

---
🚀 Composer preview: https://pr-13065-composer-dev.dxos.workers.dev

### DIFF
--- a/.changeset/mcp-operation-surface-fixes.md
+++ b/.changeset/mcp-operation-surface-fixes.md
@@ -1,8 +1,6 @@
 ---
-'@dxos/plugin-space': patch
-'@dxos/plugin-tasks': patch
-'@dxos/types': patch
 '@dxos/echo': patch
+'@dxos/plugin-markdown': patch
 ---
 
 Three fixes to the operation surface the MCP server projects.
--- a/.changeset/mcp-operation-surface-fixes.md
+++ b/.changeset/mcp-operation-surface-fixes.md
@@ -1,0 +1,16 @@
+---
+'@dxos/plugin-space': patch
+'@dxos/plugin-tasks': patch
+'@dxos/types': patch
+'@dxos/echo': patch
+---
+
+Three fixes to the operation surface the MCP server projects.
+
+`space.queryObjects` with a `typename` filter under-returned, nondeterministically: the typename was resolved to one registered `Type` entity and the filter built from that entity's identity — a versioned `dxn:` for a static declaration, an `echo:` id for a copy persisted in the space. Objects of the same typename written under any other registration did not match, and which registration was picked depended on the order the registry query happened to return, so the same call could answer with every object, some of them, or none. The typename now filters as the bare-typename DXN `Filter.type` documents for this case; the registry is consulted only to reject a typename nothing declares.
+
+`projects.get` and `tasks.list` timed out on a task set with members. Loading a set's tasks resolved the refs one at a time, each a separate indexed query, and an unresolvable ref does not fail — it waits out the resolver's own 30s timeout. Materialized refs are now read from the working set, the cold remainder resolves concurrently, and a ref that does not resolve within 5s is treated as gone.
+
+`tasks.getOutline` given a ref to something that is not an outline now fails with a typed `InvalidOperationInput` naming the actual typename, rather than throwing `TypeError: Cannot read properties of undefined (reading 'tryLoad')`.
+
+Adds `Database.makeRef`, the Effect wrapper around `db.makeRef`.
--- a/packages/core/echo/echo/src/Database.ts
+++ b/packages/core/echo/echo/src/Database.ts
@@ -484,6 +484,15 @@ export const load: <T>(ref: Ref<T>) => Effect.Effect<T, Error.EntityNotFoundErro
 export const peek = <T>(ref: Ref<T>): T | undefined => ref.peek();
 
 /**
+ * Makes a reference to an object addressed by URI, resolvable against this database.
+ * @see {@link Database.makeRef}
+ */
+export const makeRef = <T extends Entity.Unknown = Entity.Unknown>(
+  uri: URI.URI,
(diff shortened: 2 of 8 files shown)


dmaretskyi:
`Check / test (node-version=24.11.1, shard=0)` failed on `0702ae07`. This failure is not this PR's, and I've re-run the job once (attempt #2) to confirm.

**What failed** — two tests in `packages/sdk/client-e2e/src/spaces.test.ts`, both pure timeouts with no assertion failure:

```
FAIL src/spaces.test.ts > Spaces > post and listen to messages
Error: Timeout [30,000ms]
  ❯ asyncTimeout ../../common/async/src/timeout.ts:65:68
  ❯ src/spaces.test.ts:161:11

FAIL src/spaces.test.ts > Spaces > getEpochs
Error: Test timed out in 15000ms.
  ❯ src/spaces.test.ts:594:3
```

The stack bottoms out in space bring-up — `DataSpace.initializeDataPipeline` → `DataSpace._createWritableFeeds` → `SqliteKeyring.createKey`.

**Why it isn't this PR's:**

1. This same shard passed on the previous head, `ae76a9c`, which already contained **every** code change in this PR. The only delta between `ae76a9c` and `0702ae07` is `.changeset/mcp-operation-surface-fixes.md` and `packages/plugins/plugin-tasks/src/operations/get-outline.test.ts` — a changeset file and one plugin-tasks test, neither of which can reach client-e2e space initialization.
2. `client-e2e` contains no reference to anything this PR touches (`TaskSet`, `queryObjects`, `getOutline`, `Database.makeRef`) — verified by grep across `packages/sdk/client-e2e/src/`.
3. `Database.makeRef` is purely additive; no existing call site changed behaviour.

**Fix ported:** none — there is no fix to port. This looks like flaky timing in space/keyring bring-up under load rather than a defect with a known remedy, so there is no upstream change to bring in. I have not touched the test: quarantining or re-timing a test I don't own is not this PR's business.

If attempt #2 reproduces the same two timeouts identically, that confirms a pre-existing flake in `spaces.test.ts` that deserves its own issue rather than a change here.

---

PR #13074, opened 2026-09-12
protocols: replace getDocument with a batched getDocuments (breaking)

dmaretskyi:
## Problem

`DataService.updateSubscription` fetched every requested document with its own `getDocument` call, in series — the `// TODO(dmaretskyi): Batch.` at `data-service-impl.ts:65`. Each call is a Durable Object round trip, so hydrating N objects cost N wake latencies end to end.

This is measured, not inferred. Log timeline of a single `projects.get` over MCP in production (trace `53d6a76c…`, total `invokeOperation` 9252 ms):

```
09:18:26.375  request documents count=1 → document loaded
09:18:27.435  request documents count=4
09:18:28.028  document loaded      ← +593 ms
09:18:28.903  document loaded      ← +875 ms
09:18:29.409  document loaded      ← +506 ms
09:18:29.435  ⚠ index object load timed out   (×2)
09:18:29.895  request documents count=1 → document loaded
09:18:30.401  request documents count=1 → document loaded
09:18:30.899  document loaded
```

**~3.5 s of 9.25 s**, and two objects blew the client's 2 s per-object budget (`INDEX_OBJECT_LOAD_TIMEOUT`, `index-query-source-provider.ts:67`) purely from queueing behind the others. `_resolveIndexedObject` swallows that timeout and returns `undefined`, so the query silently returned a **partial result** — a correctness bug as well as a latency one.

## ⚠️ Breaking change

`EdgeFunctionEnv.DataService.getDocument` is **removed**. `getDocuments(ctx, spaceId, documentIds)` replaces it and is required.

The singular is gone rather than kept alongside the batched form: every call is a Durable Object round trip, so a singular read invites exactly the caller loop this PR exists to delete, and leaving it available would let the serial path come back.

A host implementing `EdgeFunctionEnv.DataService` must now provide `getDocuments`. Documents absent on the host are omitted, so the result may be shorter than `documentIds` and is not positionally aligned with it.

## Change

- `EdgeFunctionEnv.DataService`: `getDocument` → `getDocuments`.
- `DataServiceImpl` fetches through a new `_loadDocuments` helper — one batched call, no fallback.

Emission is deliberately unchanged — still one `sub.next` per document, in the order requested — so the client sees exactly the sequence it saw before. The win is entirely in the fetch.

Bytes are still copied before the RPC stub is disposed (`copyUint8Array`), per the [Workers RPC lifecycle](https://developers.cloudflare.com/workers/runtime-apis/rpc/lifecycle/).

## Companion PR — merge order matters

**dxos/edge#1066** implements `getDocuments` on `DataServiceEntrypoint` (a near-rename: `SubductionHandler.getDocuments(ctx, ids)` already existed) and drops the singular from edge's own `DataService.Service`, its `rpcMethods` list and one test caller.

**This PR must merge first.** Edge's `@dxos` catalog still pins a build whose runtime calls `getDocument`, so #1066 is currently red — 34 tests across 4 shards, all `Workers runtime canceled`, because the RPC method it calls no longer exists on the entrypoint. That clears when this merges, `pkg-pr-new` publishes, and #1066 bumps its catalog.

## Verification

- `npx tsc --build packages/core/compute/functions-runtime-cloudflare/tsconfig.json` — exit 0.
  (Needed `buf generate` in `packages/core/protocols` first; the generated `@dxos/protocols/buf/*` modules are absent in a fresh checkout.)
- `vitest run` in `functions-runtime-cloudflare` — 1 test file, 1 test passed, exit 0.
- `oxfmt --check` on both touched trees — clean.

**Not verified here:** the batched path end to end. Nothing in this repo implements `getDocuments`, so that requires #1066.

## Follow-up

The same trace shows `slow AM open ctx.duration=2617` — a separate 2.6 s cost in `entity-manager.ts:282`, not addressed here.

https://claude.ai/code/session_016agvkaFknv5vY6kyZuvf59

## Summary by CodeRabbit

- **Breaking Changes**
  - Document retrieval now uses a required batched `getDocuments` operation instead of the removed single-document method.
  - Hosts must implement batched retrieval; automatic per-document fallback is no longer available.
  - Batched results may omit documents that are not found and are not guaranteed to match the order of requested IDs.

- **Bug Fixes**
  - Subscription updates now load requested documents in batches and avoid unnecessary loading when no document IDs are provided.

---
🚀 Composer preview: https://pr-13074-composer-dev.dxos.workers.dev


packages/core/protocols/src/edge/EdgeFunctionEnv.ts
```diff
@@ -81,6 +81,15 @@ export interface DataService {
...
   getDocument(ctx: TraceContext, spaceId: SpaceId, documentId: string): Promise<RpcResult<RawDocument | undefined>>;
 
+  /**
+   * Batched {@link getDocument}. Documents absent on the host are omitted, so the result may be
+   * shorter than `documentIds` and is not positionally aligned with it.
+   *
+   * Optional because the host implementing this interface ships separately from its callers: one
+   * that predates the method is driven through `getDocument` one id at a time instead.
+   */
+  getDocuments?(ctx: TraceContext, spaceId: SpaceId, documentIds: string[]): Promise<RpcResult<RawDocument[]>>;
```
dmaretskyi: do a breaking change -- remove non-batched version

dmaretskyi: Done in `5063ba9a`. `getDocument` is gone from `EdgeFunctionEnv.DataService`; `getDocuments` is required, and the serial fallback in `DataServiceImpl` is deleted with it — leaving the singular available would have let the caller loop come back.

Changeset rewritten to lead with the break (`minor`, since pre-1.0 breaking rides the minor per `agents/instructions/changesets.md`).

Companion edge PR implements `getDocuments` on `DataServiceEntrypoint` — a near-rename, since `SubductionHandler.getDocuments(ctx, ids)` already existed — and drops the singular from edge's own `DataService.Service`, its `rpcMethods` list and the one test caller. It needs a catalog bump to this PR's published build, which can only happen once this merges and `pkg-pr-new` publishes.

---

dmaretskyi: Valid — fixed in `d11a15fa`, using the angle-bracket form so the period stays out of the URL.

Verified after: `oxfmt --check` clean, `tsc --build` exit 0, tests 1/1.

---

PR #13076, opened 2026-09-12
assistant: separate the session gantt's events by task

dmaretskyi:
The session gantt drew every trace event of a run on one session bar, smooshed together, with the task lanes below it empty — there was no way to tell which task a tool call belonged to.

Nothing in a trace said which task an agent was working on, so both tools that move a task's status now write a new `task.statusChanged` trace event (`@dxos/compute/Trace`, emitted from the planning tool's `update-tasks` and from `plugin-tasks`' `UpdateTask` — the delegated project flow reaches for the latter). `buildSessionTimeline` reads those events to cut a session into one segment per task: the task lane gets a span of its own, nodes where it started and finished, and every marker inside the segment is attributed to it.

### Before / after

The chart region of the project's pipeline, in each case after a live agent run over the four tasks of the `Incident 0516 retrospective` template. **These are two separate runs** — the "before" was recorded on the pre-fix commit, so its totals differ (30 tools / 1,290.0k vs 34 / 939.4k); the flow and the template are the same.

Before — four empty task lanes, every node on the session bar:
https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-12-gantt-task-separation/chart-before.png

After — one bar per task, each carrying the events of its own stretch of the run:
https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-12-gantt-task-separation/chart-after.png

### Demo

[**Live run, 2:04**](https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-12-gantt-task-separation/gantt-task-separation.webm) — recorded from `stories-assistant / Projects / Default` against a real EDGE agent, driven one gesture at a time: pick the template, select all four tasks, show the pipeline chart, assign them to an agent, and watch each task claim its lane as its status events land. The run itself took 5m11s; the recording is trimmed of still frames and ends on a hold of the finished chart. [Full frame](https://pub-39066a86073446d7b77b1c157b660bb5.r2.dev/demos/2026-09-12-gantt-task-separation/43-final.png).

The video shows only the after state — the before is the still linked above, from its own run.

### How the cut is scoped

Four ways it could go wrong, each guarded and pinned by a test:

- **Only tasks on the session's own checklist take part.** An agent is free to move a task belonging to nothing on this chart; such an event must not close the open segment or move the boundary, so the checklist is an allow-set inside `buildTaskSegments` rather than a filter over its result.
- **A task that is only ever closed** — delegation marks everything it hands over `started` before the agent's first turn — is cut from the previous boundary, but only on the transition out of `started`, so a task merely dismissed (`todo` → `blocked`) claims nothing and a second close (`review` → `done`) mints nothing.
- **A close can arrive after the next task has started.** That stretch is the next task's, so the close-only fallback never synthesizes a segment while another task is open.
- **A delegated task is drawn as the child session that worked it**, so its segment is dropped rather than moved onto a bar whose span does not contain those markers. One consequence worth knowing: for a fully delegated task, its `Task started` / `Task done` nodes stay on the supervisor bar, which is where that work (the handover) happened.

A status tool runs as a child process of the agent, so its trace events carry their own pid and reach the session through `parentPid`; segmenting on the session's own pids alone saw none of them.

The pipeline chart also draws its own lane names and per-lane totals now (`Gantt.Legend` + `Gantt.Meta`): a ledger row is several lines tall and a chart row is one, so nothing lined up between them.

### Tests

`session-timeline.test.ts` 10 passed — segmentation, the delegated-run cut, off-checklist scoping, dismissed/re-closed tasks, an interleaved close, and delegation × segmentation. Trace-emission tests in `update-tasks.test.ts` (assistant-toolkit, reading the events back off the trace feed) and `update-task.test.ts` (plugin-tasks). Both, and the timeline suite, now run on `TestTraceService` from `@dxos/compute/testing`: `TestTraceService.layer` provides the collector with the writer and sink that feed it, and `yield* TestTraceService.messages` reads back what a handler wrote.

Full suites green for `compute`, `assistant`, `assistant-toolkit`, `plugin-assistant`, `plugin-tasks`, `plugin-projects`, `react-ui-components`; lint and `oxfmt --check` clean.

The diff was reviewed adversarially by a second agent, which found three defects in the first cut of the segmentation; CodeRabbit found a fourth (the interleaved close). All four are the guards listed above.

### Open question

`Ref<Task.Task>` in the event instead of `taskId` was requested in review. The feed test says the trace path cannot carry a ref correctly today: the ref's target is inlined as a stale snapshot, the DXN loses its space (`echo:///<id>`), a ref read back has no resolver, and a live subscriber sees a `Ref` instance where a feed reader sees the envelope — because `Trace.Event.data` is `Schema.Unknown`. Details and options are on the `update-task.ts` review thread; the branch keeps `taskId` pending that decision.

https://claude.ai/code/session_01XY9A67e56DV9CB1jB9PFGf

---
_Generated by [Claude Code](https://claude.ai/code/session_01XY9A67e56DV9CB1jB9PFGf)_

### DIFF
--- a/.changeset/gantt-task-separation.md
+++ b/.changeset/gantt-task-separation.md
@@ -1,10 +1,6 @@
 ---
-'@dxos/assistant-toolkit': minor
 '@dxos/compute': minor
 '@dxos/plugin-assistant': minor
-'@dxos/plugin-projects': minor
-'@dxos/plugin-tasks': minor
-'@dxos/react-ui-components': minor
 ---
 
 The session gantt now separates a run's events by task.
--- a/.changeset/gantt-task-separation.md
+++ b/.changeset/gantt-task-separation.md
@@ -1,0 +1,12 @@
+---
+'@dxos/assistant': minor
+'@dxos/assistant-toolkit': minor
+'@dxos/plugin-assistant': minor
+'@dxos/react-ui-components': minor
+---
+
+The session gantt now separates a run's events by task.
+
+The planning tool writes a new `assistant.taskStatusChanged` trace event whenever a checklist task's status changes — the only in-trace record of which task an agent was working on. `buildSessionTimeline` reads those events to cut a session into one segment per task: the task lane gets a span of its own, nodes where it started and finished, and every marker inside the segment is attributed to it rather than piling onto the session bar.
+
+A task that is only ever closed — delegation marks everything it hands over `started` before the agent's first turn — is cut from the previous boundary, so a delegated run separates too.
--- a/.changeset/gantt-task-separation.md
+++ b/.changeset/gantt-task-separation.md
@@ -1,12 +1,16 @@
 ---
-'@dxos/assistant': minor
 '@dxos/assistant-toolkit': minor
+'@dxos/compute': minor
 '@dxos/plugin-assistant': minor
+'@dxos/plugin-projects': minor
+'@dxos/plugin-tasks': minor
 '@dxos/react-ui-components': minor
 ---
 
 The session gantt now separates a run's events by task.
 
-The planning tool writes a new `assistant.taskStatusChanged` trace event whenever a checklist task's status changes — the only in-trace record of which task an agent was working on. `buildSessionTimeline` reads those events to cut a session into one segment per task: the task lane gets a span of its own, nodes where it started and finished, and every marker inside the segment is attributed to it rather than piling onto the session bar.
+Both tools that move a task's status — the planning tool's `update-tasks` and `plugin-tasks`' `UpdateTask` — write a new `task.statusChanged` trace event. Nothing else in a trace says which task an agent was working on, so `buildSessionTimeline` reads those events to cut a session into one segment per task: the task lane gets a span of its own, nodes where it started and finished, and every marker inside the segment is attributed to it rather than piling onto the session bar.
 
-A task that is only ever closed — delegation marks everything it hands over `started` before the agent's first turn — is cut from the previous boundary, so a delegated run separates too.
+Only tasks on the session's own checklist take part in the cut, and a task that is only ever closed — delegation marks everything it hands over `started` before the agent's first turn — is cut from the previous boundary, but only on the transition out of `started`: a task merely dismissed claims nothing, and a second close (`review` → `done`) mints no segment overlapping the task then active. A delegated task is drawn as the child session that worked it, so its segment is dropped rather than moved onto a bar whose span does not contain those markers.
+
+The project's pipeline chart also draws its own lane names and per-lane totals: a ledger row is several lines tall and a chart row is one, so nothing lined up between them.
--- a/packages/plugins/plugin-assistant/src/session-timeline/session-timeline.test.ts
+++ b/packages/plugins/plugin-assistant/src/session-timeline/session-timeline.test.ts
@@ -233,6 +233,131 @@ describe('buildSessionTimeline', () => {
     expect(lanesByLabel.get('Write file')).toBe(secondLane?.id);
   });
 
+  test('a task belonging to no lane on this chart cuts nothing', ({ expect }) => {
+    const mine = Task.make({ title: 'Mine', status: 'done' });
+    const other = Task.make({ title: 'Elsewhere', status: 'started' });
+    const chat = makeChat('Scoped', [mine]);
+    const messages = collectTraceEvents(
+      withMeta(
+        { pid: 'agent', conversation: chat.feed },
+        Effect.gen(function* () {
+          yield* Trace.write(AgentRequestBegin, {}); // 1
+          yield* Trace.write(Trace.TaskStatusChanged, { taskId: mine.id, title: 'Mine', status: 'started' }); // 2
+          yield* toolCall('Read file'); // 3
+          // A task on somebody else's list: it must neither close `Mine` nor move the boundary.
+          yield* Trace.write(Trace.TaskStatusChanged, { taskId: other.id, title: 'Elsewhere', status: 'started' }); // 4
+          yield* toolCall('Write file'); // 5
+          yield* Trace.write(Trace.TaskStatusChanged, {
+            taskId: mine.id,
+            title: 'Mine',
+            status: 'done',
+            previousStatus: 'started',
+          }); // 6
+        }),
+      ),
+    );
+
+    const timeline = buildSessionTimeline({ traceMessages: messages, chats: [chat], tasks: [mine, other] });
+    const laneId = `task:${mine.id}`;
+    expect(timeline.lanes.find((lane) => lane.id === laneId)).toMatchObject({ start: 2, end: 6 });
+    const lanesByLabel = new Map(timeline.markers.map((marker) => [marker.label, marker.laneId]));
+    expect(lanesByLabel.get('Read file')).toBe(laneId);
+    expect(lanesByLabel.get('Write file')).toBe(laneId);
+  });
+
+  test('a task dismissed or re-closed claims no stretch of the run', ({ expect }) => {
+    const dismissed = Task.make({ title: 'Dismissed', status: 'blocked' });
+    const worked = Task.make({ title: 'Worked', status: 'done' });
+    const chat = makeChat('Closes', [dismissed, worked]);
+    const messages = collectTraceEvents(
+      withMeta(
+        { pid: 'agent', conversation: chat.feed },
+        Effect.gen(function* () {
+          yield* Trace.write(AgentRequestBegin, {}); // 1
+          // Never started, so it owns nothing — not the reading the agent did before it said so.
+          yield* Trace.write(Trace.TaskStatusChanged, {
+            taskId: dismissed.id,
+            title: 'Dismissed',
+            status: 'blocked',
+            previousStatus: 'todo',
+          }); // 2
+          yield* Trace.write(Trace.TaskStatusChanged, { taskId: worked.id, title: 'Worked', status: 'started' }); // 3
+          yield* Trace.write(Trace.TaskStatusChanged, {
+            taskId: worked.id,
+            title: 'Worked',
+            status: 'review',
+            previousStatus: 'started',
+          }); // 4
+          yield* Trace.write(Trace.TaskStatusChanged, { taskId: dismissed.id, title: 'Dismissed', status: 'started' }); // 5
+          yield* toolCall('Unblock'); // 6
+          // The sign-off on a task closed at 4: it must not stretch that lane over this one's work.
+          yield* Trace.write(Trace.TaskStatusChanged, {
+            taskId: worked.id,
+            title: 'Worked',
+            status: 'done',
+            previousStatus: 'review',
+          }); // 7
+        }),
+      ),
+    );
+
+    const timeline = buildSessionTimeline({ traceMessages: messages, chats: [chat], tasks: [dismissed, worked] });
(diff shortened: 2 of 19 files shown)


packages/plugins/plugin-assistant/src/session-timeline/session-timeline.ts
```diff
@@ -63,6 +63,88 @@ interface SessionSource {
...
+      if (!open) {
+        open = { taskId: data.taskId, laneId: taskLaneId(data.taskId), start: event.timestamp };
+        segments.push(open);
+        boundary = event.timestamp;
+      }
+    } else if (TERMINAL_STATUS.has(data.status)) {
+      const started = segments.findLast((candidate) => candidate.taskId === data.taskId && candidate.end === undefined);
+      if (started) {
+        close(started, event.timestamp);
+      } else if (data.previousStatus === 'started') {
```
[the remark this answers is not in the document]
dmaretskyi: Correct, and fixed in 71461db6 — the close-only fallback now requires `open === undefined`, so a close arriving after the next task started leaves that task the stretch instead of overlapping it. Regression test added for that event order (`a close arriving after the next task started leaves that task the stretch`): the first task's lane ends where the second's begins, and both tool calls in the overlap window are attributed to the second task.

---

.changeset/gantt-task-separation.md
```diff
@@ -0,0 +1,16 @@
+---
+'@dxos/assistant-toolkit': minor
+'@dxos/compute': minor
+'@dxos/plugin-assistant': minor
+'@dxos/plugin-projects': minor
+'@dxos/plugin-tasks': minor
+'@dxos/react-ui-components': minor
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 71461db6 — `agents/instructions/changesets.md` says naming any one member bumps the whole group, so the frontmatter is now `@dxos/compute` (Group A) and `@dxos/plugin-assistant` (Group B). `pnpm check-changesets` passes.

---

packages/plugins/plugin-assistant/src/session-timeline/session-timeline.test.ts
```diff
@@ -143,6 +143,221 @@ describe('buildSessionTimeline', () => {
...
 
+  test('cuts the session into task segments from the status events', ({ expect }) => {
+    const first = Task.make({ title: 'First', status: 'done' });
+    const second = Task.make({ title: 'Second', status: 'started' });
+    const chat = makeChat('Segmented', [first, second]);
+    const messages = collectTraceEvents(
+      withMeta(
+        { pid: 'agent', conversation: chat.feed },
+        Effect.gen(function* () {
+          yield* Trace.write(AgentRequestBegin, {}); // 1
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 71461db6 — all 39 numbered annotations now end with a period. The shared test helpers moved below their suites in the same commit, per the nit on `update-tasks.test.ts`.

---

packages/plugins/plugin-tasks/src/operations/update-task.ts
```diff
@@ -48,7 +49,18 @@ const handler: Operation.WithHandler<typeof TaskOperation.UpdateTask> = TaskOper
 
       // Through `Task.edit`, so the change and the log entry that explains it land together and a
       // no-op patch records nothing. Milestone stays here: it is set membership, not a field edit.
+      const previousStatus = task.status;
       Task.update(task, { title, description, status, priority, estimate, assignee: sessionAssignee ?? assignee });
+      // The resolved status, not the requested one: a task with reviewers lands in `review`. This is
+      // what cuts an agent session's timeline into per-task segments.
+      if (task.status !== undefined && task.status !== previousStatus) {
+        yield* Trace.write(Trace.TaskStatusChanged, {
+          taskId: task.id,
```
dmaretskyi, line 58: change to Ref<..>, make sure (with tests) that trace feed handles refs correctly

packages/plugins/plugin-tasks/src/operations/update-task.test.ts
```diff
@@ -207,6 +230,52 @@ describe('update-task', () => {
...
+        { taskId: task.id, title: 'Draft v2', status: 'done', previousStatus: 'started' },
+      ]);
+    }).pipe(
+      Effect.provide(collectingTrace(messages)),
+      Effect.provide(TestDatabaseLayer({ types: [Milestone.Milestone, Task.Task, TaskSet.TaskSet] })),
+    );
+  });
+
+  /** Collects what the handler wrote to the trace, so a status change can be asserted on. */
+  const collectingTrace = (messages: Trace.Message[]) =>
```
dmaretskyi: add a layer, with new context tag so that inside the test you can access

const messages = yield* TestTraceService.messages;

TestTraceService.layer: Layer<...>

packages/plugins/plugin-assistant/src/session-timeline/session-timeline.test.ts
```diff
@@ -143,6 +143,254 @@ describe('buildSessionTimeline', () => {
     expect(timeline.range).toEqual({ start: 1, end: 100 });
   });
 
+  test('cuts the session into task segments from the status events', ({ expect }) => {
```
dmaretskyi: use the previously mentioned test trace service in those suites to clean them up

packages/plugins/plugin-assistant/src/session-timeline/session-timeline.ts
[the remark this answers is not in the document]
dmaretskyi: I tried the `Ref<Task.Task>` change and wrote the feed test you asked for — it says the trace feed does **not** handle refs correctly today, so this needs a decision before I push it.

What the test measured, with `task: Ref.Ref(Task.Task)` in the event and `Ref.make(task)` at both emission sites:

1. **The target is inlined.** What lands on the feed is
   `{"/":"echo:///01JGF…","target":{"id":"01JGF…","@type":"dxn:org.dxos.type.task:0.5.0","status":"started","title":"Roundtrip"}}`
   — a frozen snapshot of the task inside every status event. It goes stale the moment the task changes, and it grows each event by the size of the task.
2. **The DXN is space-less** — `echo:///<id>`, not `dxn:echo:<space>:<id>`.
3. **A ref read back has no resolver.** `Database.load(event.task)` dies with `invariant violation: Resolver is not set (ref.ts:612)`.
4. **In-process and feed consumers see different shapes.** `Trace.Event.data` is `Schema.Unknown`, so a live subscriber gets a `Ref` instance while a feed reader gets the encoded envelope; one `Schema.decodeUnknownSync(TaskStatusChanged.schema)` cannot read both — decoding the in-memory form returns nothing, which silently dropped every segment in the timeline unit tests.

Root cause is (4): `Event.data: Schema.Unknown` means nothing on the trace path encodes or rehydrates a ref.

Three ways forward:

1. **Keep `taskId`** (what is on this branch now, green) and file the trace-feed ref gap as its own issue. `Task.refEntityId` is all the timeline needs, and the id is what a stale snapshot would have given us anyway.
2. **Make `Event.data` schema-aware** — encode refs as bare DXN envelopes on write and rehydrate them with the reading database's resolver. That is the real fix and it is a change to `@dxos/compute`'s trace infrastructure, well beyond this PR.
3. **Carry a fully-qualified DXN string** instead of a `Ref` — resolvable by a feed reader, no snapshot, no infra change, but not a typed `Ref<Task.Task>`.

My recommendation is 1 now and 2 as a follow-up, since 2 is what makes refs work for every trace event rather than just this one. The `Ref` patch is ready locally either way — say which and I will push it.

---

packages/plugins/plugin-tasks/src/operations/update-task.test.ts
dmaretskyi: Done in 3d802538 — `TestTraceService` now lives in `@dxos/compute/testing`:

```ts
const events = yield* TestTraceService.events; // or .messages
```

- `TestTraceService.layer` provides the collector together with the `TraceService` and `TraceSink` that feed it, so a suite provides one layer instead of assembling a sink.
- `.messages` returns what was written in write order; `.events` returns the same flattened.
- `TestTraceService.withMeta(meta, effect)` carries the pid and conversation a real emitter would, reusing the collector's monotonic counter as the clock so timestamps stay ordered across nested scopes (a sub-agent writing under its own meta).

`update-task.test.ts` dropped its hand-rolled `collectingTrace` for it. 10/10 there; `compute`, `plugin-tasks` full suites green, lint clean.

---

packages/plugins/plugin-assistant/src/session-timeline/session-timeline.test.ts
dmaretskyi: Done in 3d802538 — this suite is off `collectTraceEvents`/`withMeta` and onto `it.effect` + `TestTraceService.layer`, writing through `TestTraceService.withMeta` (including the nested sub-agent scopes) and reading back with `yield* TestTraceService.messages`. The two fixture-driven tests stay synchronous `test(...)` since they write nothing and read the JSON fixture instead.

`collectTraceEvents` stays where it is — `execution-graph.test.ts` and `span-tree.test.ts` still use it; it can move onto the service separately if you want that consolidated.

10/10 in this suite, 204 passed in `plugin-assistant`, lint clean.

---

packages/plugins/plugin-tasks/src/operations/update-task.ts
dmaretskyi: I built the `Ref<Task.Task>` change and wrote the feed test you asked for. It says the trace feed does **not** handle refs correctly today, so this needs your call before I push it. (I first posted this on the wrong thread — CodeRabbit's, above — and it answered; its vote isn't a substitute for yours.)

Measured with `task: Ref.Ref(Task.Task)` in the event and `Ref.make(task)` at both emission sites:

1. **The target is inlined.** What lands on the feed is
   `{"/":"echo:///01JGF…","target":{"id":"01JGF…","@type":"dxn:org.dxos.type.task:0.5.0","status":"started","title":"Roundtrip"}}`
   — a frozen snapshot of the task inside every status event: stale as soon as the task changes, and every event grows by the size of the task.
2. **The DXN is space-less** — `echo:///<id>`, not `dxn:echo:<space>:<id>`.
3. **A ref read back has no resolver.** `Database.load(event.task)` dies with `invariant violation: Resolver is not set` (`ref.ts:612`).
4. **Live and feed consumers see different shapes.** `Trace.Event.data` is `Schema.Unknown`, so an in-process subscriber gets a `Ref` instance while a feed reader gets the encoded envelope; one `Schema.decodeUnknownSync(TaskStatusChanged.schema)` cannot read both — decoding the in-memory form returns nothing, which silently dropped every segment in the timeline unit tests.

(4) is the root cause: nothing on the trace path encodes or rehydrates a ref while `data` is `Schema.Unknown`.

1. **Keep `taskId`** (this branch, green) and file the trace-feed ref gap separately. `Task.refEntityId` is all the timeline needs, and an id is what a stale snapshot would have yielded anyway.
2. **Make `Event.data` schema-aware** — encode refs as bare DXN envelopes on write, rehydrate with the reading database's resolver. The real fix, and a change to `@dxos/compute`'s trace infrastructure well beyond this PR.
3. **Carry a fully-qualified DXN string** instead of a `Ref` — resolvable by a feed reader, no snapshot, no infra change, but not a typed `Ref<Task.Task>`.

I'd take 1 now with 2 as a follow-up, since 2 is what makes refs work for every trace event rather than this one. The patch for the `Ref` version is ready locally — name the option and I'll push it.

---

PR #13081, opened 2026-09-14
assistant: store the selected model on the Chat object

dmaretskyi:
The model a conversation runs on lived in the chat UI's per-mount React state and was passed into `AgentService.getSession`, so it was lost on remount and unknown to any other caller that started a turn on the chat's behalf. It now lives on the chat.

### Changes

- **`Chat.model`** holds the selection as a ref whose URI is the model DXN. There is no ECHO object behind a model yet, so the ref is a stable handle rather than something that resolves; a `TODO` records that registering `Model` in the registry is out of scope here. Call sites read and write the ref directly.
- **`AgentProcess`** reads the model off the chat it is bound to, falling back to its `defaultModel` option. A rehydrated process recovers the model by re-reading the chat, the way it already recovers its steering instructions.
- **`AgentService.getSession`** loses its `model` option. The `model` option on the service layer and the process is now `defaultModel`, applying only to a chat that has not selected one, and the session cache tracks the chat's model ref, so repointing it tears the process down and respawns it on the new selection.
- **`AgentService.createSession`** puts its `model` option on the chat it creates, which is where the process now reads it.
- **The chat prompt's model picker** (`usePresets`) reads and writes the chat's selection instead of per-mount state, subscribed via `useObject` so it follows a change made on another mount. A model the active provider no longer serves is shown as an unavailable entry rather than silently replaced by the fallback, so the picker and the request agree on what the conversation runs on.
- **The processor and `RunPromptInChat`** stamp the picker's default onto a chat that has not selected a model, before the first request.

### Testing

| Check | Result |
| --- | --- |
| `agent-runtime:test` | 46 passed, 22 skipped |
| `assistant:test` | 81 passed |
| `assistant-toolkit:test` | 77 passed, 14 skipped |
| `plugin-assistant:test` | 209 passed, 7 skipped |
| `compute:test` | 76 passed |
| builds + lint (compute, assistant, agent-runtime, assistant-toolkit, plugin-assistant, stories-assistant) | pass |

New coverage: a `Chat` unit test for the model round-tripping through the database and clearing back to unset, and an `AgentService` test that a chat with no model runs the default, that selecting one respawns the process, and that the same selection is a cache hit.

https://claude.ai/code/session_013UNhcsWSFq1uKw3k1uiYsU

---
🚀 Composer preview: https://pr-13081-composer-dev.dxos.workers.dev

## Summary by CodeRabbit

- **New Features**
  - Chat conversations now retain their selected AI model across remounts and rehydration.
  - Model selections travel with the conversation and apply consistently across sessions.
  - The model picker preserves selections that are no longer available and labels them as unavailable.
  - Default models are used only when a conversation has no model selected.

- **Bug Fixes**
  - Prevented selected models from being silently replaced by the configured provider default.


packages/core/compute/agent-runtime/src/agent-service/AgentService.ts
```diff
@@ -111,7 +113,7 @@ export interface AgentServiceOptions {
   makeTurnProducer?: MakeTurnProducer;
 
   /**
-   * Default model used by sessions that don't specify one explicitly.
+   * Default model for a chat that has not selected one (`Chat.model` unset).
    */
   model?: DXN.DXN;
```
dmaretskyi: rename defaultModel

packages/core/compute/assistant/src/types/Chat.ts
```diff
@@ -191,6 +200,23 @@ export const deleteTask = (
   return subtree;
 };
 
+/**
+ * The model DXN a chat's `model` ref points at; `undefined` when the chat has none or the ref does
+ * not carry a DXN (a ref to a database object, which a future registry-backed model would be).
+ */
+export const modelOf = (ref: Ref.Ref<Obj.Unknown> | undefined): DXN.DXN | undefined =>
```
dmaretskyi: kill those 3 helpers -- just do it inline

packages/core/compute/agent-runtime/src/agent-service/AgentService.ts
dmaretskyi: Renamed to `defaultModel` on both `AgentServiceOptions` and `AgentProcessOptions`, since it is the same value threaded through, and updated `AssistantTestLayer` which sets it.

---

packages/core/compute/assistant/src/types/Chat.ts
dmaretskyi: Removed all three. Each call site now reads `chat.model.uri` through `DXN.tryMake` or writes `Ref.fromURI` inside an `Obj.update` directly.

---

packages/plugins/plugin-assistant/src/operations/run-prompt-in-chat.ts
```diff
@@ -49,8 +49,12 @@ const handler: Operation.WithHandler<typeof AssistantOperation.RunPromptInChat>
           yield* Effect.promise(() => db.flush());
         }
         const preset = yield* chatPreset;
+        // As the chat's own UI does before its first request: the process reads the model off the
+        // chat, so a chat without one is stamped with the model its picker would show.
+        if (!chat.model && preset) {
+          Chat.setModel(chat, preset.model);
+        }
```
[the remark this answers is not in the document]
dmaretskyi: Valid, and this path was the worse of the two: `chatPreset` derives from settings alone, so a chat holding a model from one provider was handed whichever provider the settings now name.

Fixed by a new `providerForModel(model, active)` in `processor/presets.ts`: the active provider wins whenever it actually serves the model (the local model ids are served by `built-in`, `ollama` and `lmStudio` alike, so the catalog's first entry is not necessarily the live one), and otherwise it falls back to a provider that does serve it. Applied here and in the processor, and reused for the picker's unavailable entry. Covered by `processor/presets.test.ts`.

The commit is ready but cannot be pushed to this branch: the PR is already in the merge queue, which locks the branch. It will land as an immediate follow-up PR instead.

---

packages/plugins/plugin-assistant/src/processor/processor.ts
```diff
@@ -509,8 +514,10 @@ export class AiChatProcessor {
         // conversation to run.
         return yield* Effect.die(new Error('Chat processor requires a chat.'));
       }
+      if (!chat.model && this._options.model) {
+        Chat.setModel(chat, this._options.model);
+      }
       return yield* AgentService.getSession(chat, {
```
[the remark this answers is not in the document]
dmaretskyi: Valid. This path was partly covered already, since `usePresets` hands the processor the unavailable entry's own provider, but only when the preset list is non-empty, and the coupling was implicit. The provider is now derived from the chat's model at the call site through the shared `providerForModel` helper, so it no longer depends on what the hook happened to pass.

Same caveat as the sibling thread: the commit is ready but the branch is locked by the merge queue, so it lands as an immediate follow-up PR.

---

PR #13089, opened 2026-09-14
plugin-sandbox: bring binary files out of the sandbox intact

dmaretskyi:
## What

`SandboxOperation.DownloadFile` re-encoded the service's `content` as utf-8 text and stamped every File `text/plain`, so an image read from the sandbox arrived as mangled bytes with the wrong type. This makes the download path byte-exact and typed:

- `SandboxClient.readFile` decodes the full read result (`content`, `encoding`, `mimeType`, `size`) and accepts an optional `encoding` request parameter.
- `SandboxClient.readFileBytes` turns a `base64` read back into the original bytes and pairs them with the service-detected MIME type (falling back to `application/octet-stream` for untyped binary, `text/plain` for untyped text). The type is never guessed from the path.
- `download-file.ts` stores those bytes and that type on the File's blob, for both the new-File and `dest` paths.

## Tests

- `src/services/SandboxClient.test.ts` (docker-free, runs in CI): the base64 → bytes decode, utf-8 re-encode, the content-only fallback of an older service, and the `encoding` query parameter.
- `src/testing/SandboxPlugin.test.ts` (`describe.skip`, needs a running sandbox-service): a new `download image file` case writes a 1x1 PNG inside the sandbox via `Exec`, downloads it, and asserts the blob's type is `image/png`, its bytes equal the PNG, and that fetching `Blob.url(blob)` yields a `content-type: image/png` response whose body is the PNG byte for byte.

```
 ✓ |node| src/services/SandboxClient.test.ts (5 tests) 36ms
 ↓ |node| src/testing/SandboxPlugin.test.ts (4 tests | 4 skipped)
 Test Files  4 passed | 2 skipped (6)
      Tests  13 passed | 5 skipped (18)
```

`moon run plugin-sandbox:build plugin-sandbox:lint` and `pnpm format` are clean.

## Needs a matching `dxos/edge` change

The deployed sandbox-service currently strips everything but `content` from the SDK's read result, so the client cannot tell a base64 read from a text one and the image test will not pass against it until the service reports `encoding`/`mimeType`. The client degrades to today's behaviour against that service (content-only → utf-8 text). The service-side change was applied to a local checkout, where `moon run sandbox-service:build sandbox-service:test` passes, but this session cannot push to `dxos/edge`; the patch is:

<details>
<summary><code>packages/services/sandbox-service</code> diff</summary>

```diff
--- a/packages/services/sandbox-service/src/api.ts
+++ b/packages/services/sandbox-service/src/api.ts
@@ -28,6 +28,7 @@ import {
   FilePathQuery,
   ListSandboxesQuery,
   MkdirRequest,
+  ReadFileQuery,
   SandboxParams,
   SpaceParams,
   WriteFileRequest,
@@ -184,10 +185,11 @@ routes.post(
 routes.get(
   '/spaces/:spaceId/sandboxes/:sandboxId/files',
   Rest.validator('param', SandboxParams),
-  Rest.validator('query', FilePathQuery),
+  Rest.validator('query', ReadFileQuery),
   async (c) => {
     const { sandboxId } = c.req.valid('param');
-    return c.json(await getManager(sandboxId).readFile(spaceIdOf(c), c.req.valid('query').path));
+    const { path, encoding } = c.req.valid('query');
+    return c.json(await getManager(sandboxId).readFile(spaceIdOf(c), path, { encoding }));
   },
 );
--- a/packages/services/sandbox-service/src/sandbox-manager-object.ts
+++ b/packages/services/sandbox-service/src/sandbox-manager-object.ts
@@ -27,7 +27,8 @@ import {
 } from './constants.ts';
 import type { Env } from './env.ts';
 import { encodeExecCommand } from './exec-command.ts';
-import type { ExecOptions, ExecResult, FileEntry, SandboxRecord } from './types.ts';
+import type { FileEncoding } from './schema.ts';
+import type { ExecOptions, ExecResult, FileEntry, ReadFileResult, SandboxRecord } from './types.ts';
@@ -142,13 +143,18 @@ export default class SandboxManagerObject extends DurableObject<Env> {
-  async readFile(spaceId: SpaceId, path: string): Promise<{ content: string }> {
+  /**
+   * The SDK reads a binary file as base64 and a text file as utf-8 unless told otherwise; the
+   * encoding it used, and the MIME type it detected, go back with the content so the caller can
+   * reconstruct the bytes — an image read as text is not an image any more.
+   */
+  async readFile(spaceId: SpaceId, path: string, options: { encoding?: FileEncoding } = {}): Promise<ReadFileResult> {
     const record = await this.#touch(spaceId);
     const { sandbox } = await this.#getMounted(record);
     const result = await this.#guard(record, () =>
-      withTimeout(sandbox.readFile(path), DEFAULT_EXEC_TIMEOUT_MS, `readFile ${record.id}`),
+      withTimeout(sandbox.readFile(path, options), DEFAULT_EXEC_TIMEOUT_MS, `readFile ${record.id}`),
     );
-    return { content: result.content };
+    return { content: result.content, encoding: result.encoding, mimeType: result.mimeType, size: result.size };
   }
--- a/packages/services/sandbox-service/src/schema.ts
+++ b/packages/services/sandbox-service/src/schema.ts
@@ -66,6 +66,18 @@ export const FilePathQuery = Schema.Struct({
   path: BoundedString(MAX_PATH_LENGTH),
 });
 
+/**
+ * Wire encoding of a file's `content`: `utf-8` carries text verbatim, `base64` carries bytes. Left
+ * unset on a read, the SDK picks one from the file's MIME type and the response reports which.
+ */
+export const FileEncoding = Schema.Literals(['utf-8', 'base64']);
+export type FileEncoding = Schema.Schema.Type<typeof FileEncoding>;
+
+export const ReadFileQuery = Schema.Struct({
+  path: BoundedString(MAX_PATH_LENGTH),
+  encoding: Schema.optional(FileEncoding),
+});
+
--- a/packages/services/sandbox-service/src/types.ts
+++ b/packages/services/sandbox-service/src/types.ts
@@ -26,6 +26,14 @@ export type ExecResult = {
   success: boolean;
 };
 
+export type ReadFileResult = {
+  content: string;
+  /** How `content` is encoded; absent only when the SDK did not say. */
+  encoding?: 'utf-8' | 'base64';
+  mimeType?: string;
+  size?: number;
+};
+
```

</details>

Not in scope: `UploadFile` still sends the blob as decoded text, so a binary upload is corrupted the same way the download used to be. The image test writes the PNG inside the sandbox via `Exec` for that reason.

https://claude.ai/code/session_0189gbgcr7yvV7sSNKoEcAb3

---
_Generated by [Claude Code](https://claude.ai/code/session_0189gbgcr7yvV7sSNKoEcAb3)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox file downloads now preserve binary content, including images, without corruption.
  * Downloaded files use their detected MIME type instead of defaulting to plain text.
  * Files without a detected type now receive an appropriate fallback type.
  * Text and binary files are decoded according to their reported encoding, improving download reliability across file types.

---
🚀 Composer preview: https://pr-13089-composer-dev.dxos.workers.dev


dmaretskyi:
CI on `713db63e` is red for a reason outside this PR: the Depot runners are getting HTTP 504 from `github.com/*/releases/download/...` while installing the toolchain, so the affected jobs die before any build or test target runs.

- Failing: `Check / test` shards 0–3, `Model Fixture / model-fixture` (first attempt also `Check / memory`, which passed on retry). Every failed log ends in `curl: (22) The requested URL returned error: 504` from `proto.sh 0.60.2` or `moon.sh 2.5.2`.
- Passing on this head: `Check / check`, `memory`, `boot-budget`, the GitHub `build` job. The same diff was fully green on `20f0ebfc` before the `main` sync.
- Retried the failed jobs once via `depot ci retry`; the second attempts failed identically, which confirms it is the download endpoint, not the change. No fix exists to port; nothing in this repo controls that. githubstatus.com reports all systems operational, so it looks like a transient release-CDN issue.

Auto-merge stays enabled; once the toolchain downloads recover, the failed jobs need one more re-run to go green.

---

PR #13094, opened 2026-09-14
client-services: drive the host lifecycle through an Effect event bus

dmaretskyi:
## Summary

Replaces the hand-maintained open/close sequences in `ClientServicesHost` with an event chain. Each stack layer opens its component on the lifecycle fact it depends on and closes it in its layer finalizer, so boot order lives in the event graph and teardown is runtime disposal.

### `@dxos/effect`: new `Event` module

- `Event.make<T>()(id)` defines a typed, `Pipeable` event; `Event.handler` (data-last) and `Event.subscribe` compose as `X.pipe(Event.handler(fn), Event.subscribe)`.
- `Event.emit` awaits every handler, so an emit doubles as a barrier for the work it triggers. Handlers may emit follow-up events.
- Handler failures are defects: an event cannot enumerate the errors of every subscriber.
- `busLayer` is `Layer.sync`, one bus per runtime.

### `@dxos/client-services`: lifecycle as events

`services/events.ts` documents the chain: `StorageReady → IdentityLoaded → NetworkReady`, then `IdentityBound`, `IdentityAvailable → DataSpacesReady`, `StackOpened`, plus `ProfileUpdated`.

- The host runs the storage stage, emits `StorageReady`, joins the network if an identity exists, emits `IdentityAvailable`, then `StackOpened`. `_initialize`, `_setNetworkIdentity` and `_closeStack` are gone; `createIdentity` and `_acceptIdentity` share `_bindIdentity`.
- New `NetworkLifecycleLayer` owns edge/signal/swarm identity binding and open/close.
- Identity manager, space manager, data-space manager, edge-agent manager, cross-device sync, feed syncer and invitations manager each gained two lines: a finalizer that closes the component and a subscription that opens it. `Resource.close()` before open is a no-op, so the finalizer is unconditional.
- Each RPC handler layer now lives next to its `ServiceImpl` instead of in `client-services-layer.ts`.

### Ordering

Open order comes from event causality, never from subscription order (sibling layers under `Layer.mergeAll` build concurrently, so registration order is not stable). Handlers on one event are independent and run in parallel; if two need an order, that is a missing event. Close order comes from layer scope finalizers, which run in reverse build order; the `provideMerge` chain in `ServiceContextLayer` is therefore load-bearing for teardown and is commented as such.

### Fixes a pre-existing import cycle

`identity-service → service-host → client-services-layer → identity-service` bundled to an undefined layer, so `Layer.mergeAll` threw `Cannot read properties of undefined (reading 'build')` and every `client-e2e` test failed to load on this branch. The `ClientServicesHostService` tag now lives in `services/host-service.ts` with a type-only import of the host class.

## Deliberate behaviour changes

1. The redundant second `metadataStore.load()` in the old open path is dropped; the version check already loads it.
2. Edge-agent manager and cross-device sync open concurrently on `DataSpacesReady` rather than in sequence, and `initialized` wakes after both.
3. Feed store and metadata store still close from the host after runtime disposal, not from their layers: the metadata store persists on close, so it must never close from a layer built without the storage stage having loaded it.

## Test plan

- `effect`: 69 passed, `client-services`: 185 passed (full suites, after merging `main`).
- `client-e2e`: 173 passed, 1 failed. `Spaces > post and listen to messages` also fails on `HEAD` with only the cycle fix applied, so it predates this change.
- Composer Playwright: 18 passed, 1 flaky, 1 failed (`guest joins host's space`, fails first attempt and passes on retry). The same spec on the `HEAD`-plus-cycle-fix baseline was strictly worse (1 failed, 2 flaky). Both e2e failures are pre-existing on this branch and unrelated to lifecycle.
- Local Composer e2e needs `DX_ENVIRONMENT=dev`; the ambient hostname value is rejected by the branding config.

## Summary by CodeRabbit

* **New Features**
  * Added a typed event system for coordinating service and application lifecycle events.
  * Improved client-services startup and shutdown sequencing, including readiness handling for identity, networking, spaces, and agents.
  * Added RPC routing that supports multiple service groups, prefix-based routing, and dynamic registration.
  * Expanded developer tools integration for service access, SQLite queries, database exports, and space or metadata subscriptions.

* **Bug Fixes**
  * Added warnings when space message subscriptions encounter stream errors.
  * Fixed bundled client-services RPC build issues.

---
🚀 Composer preview: https://pr-13094-composer-dev.dxos.workers.dev


packages/common/effect/src/Event.ts
```diff
@@ -0,0 +1,128 @@
...
+  Bus.pipe(Effect.flatMap((bus) => bus.emit(event, payload)));
+
+// Erases the payload type so handlers for different events can share one registry.
+type AnyHandler = {
+  readonly event: Any;
+  readonly handler: (payload: any) => Effect.Effect<void, never, Bus>;
+};
+
+export const makeBus = (): BusService => {
+  const subscribers = new Map<string, Set<AnyHandler>>();
```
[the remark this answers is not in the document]
dmaretskyi: Done: the bus registry is now keyed by the event object, and a test checks that two events sharing an id stay apart.

packages/common/effect/src/Event.ts
```diff
@@ -0,0 +1,128 @@
...
+      Effect.gen(function* () {
+        const handlers = subscribers.get(event.id);
+        if (handlers === undefined) {
+          return;
+        }
+        const dispatch = Array.from(handlers, (entry) => entry.handler(payload));
+        if (event.strategy === 'serial') {
+          yield* Effect.forEach(dispatch, (effect) => effect, { discard: true });
+        } else {
+          yield* Effect.all(dispatch, { discard: true });
```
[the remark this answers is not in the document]
dmaretskyi: Done: parallel dispatch now passes concurrency: 'unbounded', with a test whose first handler only completes after the second has run.

packages/sdk/client-services/src/packlets/identity/identity-manager.ts
```diff
@@ -596,13 +598,24 @@ export const IdentityManagerLayer = (
...
         spaceManager,
         edgeConnection: Option.getOrUndefined(edgeConnection),
         ...options,
       });
+
+      yield* Effect.addFinalizer(() => Effect.promise(() => identityManager.close(Context.default())));
+      yield* StorageReady.pipe(
+        EffectEvent.handler(({ ctx }) =>
+          Effect.promise(() => identityManager.open(ctx)).pipe(
+            Effect.andThen(EffectEvent.emit(IdentityLoaded, { ctx, identity: identityManager.identity })),
```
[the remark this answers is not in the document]
dmaretskyi: Done: switched to Effect.flatMap so the identity is read after open completes. This was a real bug: a persisted identity booted with an ephemeral network identity.

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -696,7 +664,7 @@ export class ClientServicesHost {
     this.#resetting = true;
     this.#statusUpdate.emit();
     if (this.#open) {
-      await this._closeStack(this.#ctx ?? Context.default());
+      await this.#disposeStack();
```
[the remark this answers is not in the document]
dmaretskyi: Done: reset() now goes through close(), so #open, the handlers and the logging service are reset consistently before the wipe.

packages/sdk/client-services/src/packlets/identity/identity-lifecycle.ts
```diff
@@ -0,0 +1,108 @@
...
+import { type Identity } from './identity.ts';
+
+/**
+ * Brings a new identity into the running stack: binds it to the network, joins, and announces it so
+ * the identity-bound services open. The persisted identity on boot goes through the same events
+ * from the host's open sequence.
+ */
+export interface IdentityLifecycle {
+  createIdentity(params?: CreateIdentityOptions, ctx?: Context): Promise<Identity>;
+  acceptIdentity(params: JoinIdentityProps): Promise<Identity>;
```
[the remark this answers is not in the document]
dmaretskyi: Done: JSDoc added to IdentityLifecycle.createIdentity and acceptIdentity, and to InvitationsManager.getInvitationHandler.

packages/sdk/client-services/src/packlets/services/events.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Event } from '@dxos/effect';
+import { type Credential, type ProfileDocument } from '@dxos/protocols/buf/dxos/halo/credentials_pb';
+
+import { type Identity } from '../identity/index.ts';
+
+//
+// Lifecycle events of the client stack. Boot order is the chain of these events, each layer
+// subscribing to the fact it needs and emitting the fact it establishes:
+//
+//   Opening → StorageReady → IdentityLoaded → NetworkReady   (host emits Opening, then cascades)
```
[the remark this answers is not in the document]
dmaretskyi: Done: the lifecycle diagram lines now end with periods.

packages/sdk/client-services/src/packlets/services/service-host.ts
```diff
@@ -617,22 +524,22 @@ export class ClientServicesHost {
...
-    // Run the open lifecycle stages (formerly ServiceContext._open).
-    // The identity service impl's open/close lifecycle is owned by its layer scope in
-    // {@link ClientServicesRpcLayer} and runs when the stack runtime is disposed.
-    await this._openStack(ctx);
+    try {
+      await this._openStack(ctx);
+    } catch (err) {
+      // Release the layer-owned components so a failed boot does not leak the runtime; the stores
+      // below the stack stay untouched since nothing guarantees they were loaded.
+      await this.#stackRuntime.dispose();
```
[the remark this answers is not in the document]
dmaretskyi: Done: one rollback boundary now wraps runtime construction, service resolution and the open chain; it disposes the runtime and resets opening, handlers and the stack context. Stores stay untouched on purpose, since the metadata store persists on close.

packages/sdk/client-services/src/packlets/identity/identity-lifecycle.ts
```diff
@@ -0,0 +1,105 @@
...
+      await identity.joinNetwork(ctx);
+      await emit(EffectEvent.emit(IdentityAvailable, { identity }));
+      log('identity created', { identityKey: identity.identityKey });
+      return identity;
+    },
+
+    acceptIdentity: async (params) => {
+      const ctx = defaultCtx;
+      const { identity, identityRecord } = await identityManager.prepareIdentity(params, ctx);
+      await emit(EffectEvent.emit(IdentityBound, { identity, deviceCredential: params.authorizedDeviceCredential! }));
```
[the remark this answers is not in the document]
dmaretskyi: Done: the prepared identity is closed and the error rethrown if binding or joining fails before the manager accepts it; after acceptance the manager owns it. Also cleared every stack-owned field on dispose and in the open rollback, per the outside-diff note.

packages/core/mesh/rpc/src/RpcRouter.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    Effect.gen(function* () {
+      const scope = yield* Scope.fork(yield* Effect.scope, 'sequential');
+      yield* serving.pipe(Effect.provide(router), Scope.provide(scope));
+      return scope;
+    });
+
+  return { client, register };
+});
+
+const closeScope = (scope: Scope.Closeable) => Scope.close(scope, Exit.void);
```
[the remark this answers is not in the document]
dmaretskyi: Not changed: @effect/vitest exports it.live but no test.live (its index.d.ts has only the top-level live and it.live), and it.live is the repo's established form with 14 uses; the helpers stay above the describe that uses them.

packages/core/mesh/rpc/src/RpcRouter.ts
```diff
@@ -0,0 +1,211 @@
...
+  yield* parent.run(route).pipe(Effect.forkScoped);
+  yield* Queue.take(parent.disconnects).pipe(
+    Effect.flatMap((clientId) =>
+      Effect.forEach([fallback, ...routes.values()], (target) => Queue.offer(target.disconnects, clientId), {
+        discard: true,
+      }),
+    ),
+    Effect.forever,
+    Effect.forkScoped,
+  );
```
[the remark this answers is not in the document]
dmaretskyi: Done: the disconnect loop now drops the client's requestRoutes and pendingEnds after forwarding the disconnect to every route.

packages/core/mesh/rpc/src/RpcRouter.ts
```diff
@@ -0,0 +1,211 @@
...
+            routes.delete(prefix);
+            for (const [clientId, requests] of requestRoutes) {
+              for (const [requestId, owner] of requests) {
+                if (owner === route) {
+                  forgetRequest(clientId, requestId);
+                }
+              }
+            }
+          }),
+        );
```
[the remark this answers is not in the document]
dmaretskyi: Done: the route finalizer now settles the closing route's share of every pending end via onRouteEndedClient, so a route closed before its server accepted Eof still lets the last pending route end the transport. The router suite (7 tests) passes.

packages/sdk/client-services/src/packlets/metadata/sqlite-metadata-store.ts
```diff
@@ -118,6 +123,7 @@ export class SqliteMetadataStore implements IMetadataStore {
    */
   @synchronized
   async load(): Promise<void> {
+    this.#loaded = true;
```
dmaretskyi: `#loaded` is set before the rows are actually read, so a throw anywhere below (SQL error, a corrupt row failing CRC/decode) leaves the store `#loaded === true` while `#metadata` is still `emptyEchoMetadata()`. The layer finalizer's `close()` then passes the `#loaded` guard and calls `_save()`, persisting empty metadata over the real space/identity records — the exact case this flag exists to prevent.

Suggest moving `this.#loaded = true;` to the end of `load()`, after `#metadata` and `#spaceLargeMetadata` are populated.

---

packages/sdk/client-services/src/packlets/diagnostics/diagnostics.ts
```diff
@@ -111,8 +142,8 @@ export const createDiagnostics = async (
     (async () => {
       diagnostics.storage = await asyncTimeout(getStorageDiagnostics(), DEFAULT_TIMEOUT).catch(() => undefined);
     })(),
-    async () => {
-      const identity = serviceContext.identityManager.identity;
+    (async () => {
```
dmaretskyi, line 146: This block went from `async () => {…}` (a function handed to `Promise.all`, so never invoked) to a real IIFE, so it now actually executes — but unlike the two blocks above it, it has no `.catch(() => undefined)` and no `asyncTimeout`.

Anything that throws in here (e.g. `getSpaceStats` on a half-open space, or `EffectContext.get(stack, DataSpaceManagerService)` when the tag isn't in the context yet) will now reject the whole `Promise.all` and fail `SystemService.getDiagnostics`, where it previously returned partial diagnostics. Worth wrapping the same way as its siblings.

---

packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -202,31 +214,43 @@ export const run = ({
...
+            await closeSession();
+            break;
+          }
+
+          const tabGone = message.sessionLockKey
+            ? navigator.locks.request(message.sessionLockKey, () => {
+                // Granted once the tab releases it.
+              })
+            : new Promise<never>(() => {});
+          void Promise.race([superseded.wait(), tabGone]).then(closeSession);
```
dmaretskyi: Minor: the `tabGone` lock request carries no abort signal, and when the race is won by `superseded` it is never cancelled. On repeated connect attempts from the same tab, each superseded session leaves a queued `navigator.locks.request` outstanding on its `sessionLockKey`, whose `closeSession` eventually runs against an already-closed scope. Harmless per instance, but it accumulates over a long-lived worker — passing an `AbortSignal` tied to the session would clear it.

---

dmaretskyi (commented):
Reviewed the non-test hunks of the refactor. The event-bus design, the `provideMerge` ordering comment, and the import-cycle fix all read as sound; three findings inline, one of which (the metadata store) is a potential data-loss path.

---

packages/sdk/client-services/src/packlets/metadata/sqlite-metadata-store.ts
dmaretskyi: Fixed in 6d314df: `#loaded = true` moved to the end of `load()`, after the rows are decoded, the large metadata is loaded and the cleanup interval is scheduled, with a comment naming the wipe it prevents. A throw anywhere above now leaves the store unloaded, so `close()` skips `_save()`.

---

packages/sdk/client-services/src/packlets/diagnostics/diagnostics.ts
dmaretskyi: Fixed in 6d314df: the identity block now ends `.catch((err) => log.warn('failed to collect identity diagnostics', { err }))`, so it degrades to partial diagnostics like its siblings rather than rejecting the `Promise.all`.

Logged rather than swallowed with `() => undefined` — unlike the two above it, this block has no single result to fall back to, so the cause is worth a line. `getDiagnostics` is covered by `client-services` (190 files green).

---

packages/sdk/worker-framework/src/Worker.ts
dmaretskyi: Fixed in 6d314df: the request now takes an `AbortSignal` that is aborted when `superseded` wins the race, so a superseded session no longer leaves a request queued on its `sessionLockKey` for the worker's life. The abort rejection is caught and turned into a never-settling promise so the race still resolves from the `superseded` arm; anything other than `AbortError` rethrows.

`worker-framework` is green, node (27) and browser (13).

---

packages/common/effect/src/Event.ts
```diff
@@ -0,0 +1,134 @@
...
+export const emit = <E extends Any>(event: E, payload: Payload<E>): Effect.Effect<void, never, Bus> =>
+  Bus.pipe(Effect.flatMap((bus) => bus.emit(event, payload)));
+
+// Erases the payload type so handlers for different events can share one registry.
+type AnyHandler = {
+  readonly event: Any;
+  readonly handler: (payload: any) => Effect.Effect<void, never, Bus>;
+};
+
+export const makeBus = (): BusService => {
```
wittjosiah, line 101: why is this not a PubSub?

packages/sdk/client-services/src/packlets/services/client-services-stack.test.ts
wittjosiah, line 1: why is this still here if the host is removed? rename?

packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -202,31 +214,54 @@ export const run = ({
...
+            await closeSession();
+            break;
+          }
+
+          // Aborted when a newer attempt supersedes this session: an un-cancelled request stays
+          // queued on the tab's lock for the worker's life, and would run `closeSession` against an
+          // already-closed scope whenever the tab finally released it.
+          const tabGoneAbort = new AbortController();
+          const tabGone = message.sessionLockKey
+            ? navigator.locks
```
dmaretskyi, line 283: Two issues in this race.

**1. Unhandled rejection.** `navigator.locks.request` can reject — `SecurityError` on an opaque/insecure origin, or the lock manager going away — and nothing here catches it. The race rejects with no handler attached, and as a side effect the session never closes. Per the repo's "never suppress unhandled errors" rule this should be an explicit path, not an implicit one:

```ts
void Promise.race([superseded.wait(), tabGone]).then(closeSession, (err) => {
  log.catch(err);
  void closeSession();
});
```

**2. The lock request is never aborted when the session ends another way.** When `superseded` wins, this `navigator.locks.request` stays queued behind the tab's still-held lock for the life of the worker; when the tab finally goes it fires `closeSession` a second time (harmless today thanks to the `attempt` guard, but it is a growing queue of pending requests per superseded session). The tab side already models this correctly with `{ signal: ctx.signal }` — pass an `AbortSignal` tied to the session scope here too.

---

packages/sdk/worker-framework/src/Client.ts
```diff
@@ -411,6 +414,23 @@ export class Connection extends Resource {
...
+          .request(sessionLockKey, { signal: ctx.signal }, async () => {
+            granted.wake();
+            await released.wait();
+          })
+          .catch((err) => {
+            if (!isAbortError(err)) {
+              log.catch(err);
+            }
+          });
+        await granted.wait();
```
dmaretskyi: `connect` hangs forever if the lock request rejects. Every rejection path above is swallowed by the `.catch` (an `isAbortError` is dropped silently, anything else is logged), but `granted` is never woken in either case — so this `await` blocks indefinitely, with no timeout around it.

The reachable case is an already-aborted `ctx.signal`: if the connection is closed concurrently with `open()`, `navigator.locks.request` rejects immediately with `AbortError`, and this attempt then never completes or fails — it just stops.

Wake `granted` (or reject the attempt) from the `catch`, so the failure surfaces as a connect failure rather than a hang.

---

packages/common/effect/src/Event.ts
```diff
@@ -0,0 +1,134 @@
...
+  const subscribers = new Map<Any, Set<AnyHandler>>();
```
dmaretskyi: **Event identity is object reference, not id** — and I think that is a real hazard at this scale, not just a design choice.

`subscribers` is keyed by the `Event` object, and `Event.test.ts` pins the behaviour down ("keeps events with the same id apart": two events both named `'foo'` never see each other). These events now cross package *and* bundle boundaries — `LocalClientServices` dynamically `import()`s `@dxos/client-services` and emits `HostEvents.Opening` onto a bus that stack layers subscribed to from their own import of the same module.

If that module is ever instantiated twice — dual ESM/CJS resolution, a vite dep-optimizer re-bundle, HMR, or the kind of bundling problem the changeset says it just fixed in `@dxos/client-services` — you get two distinct `Opening` objects, the entire boot cascade becomes a no-op, and there is no error anywhere: `emit` on an unsubscribed event returns `Effect.void` by design. That failure is indistinguishable from "storage is slow".

Suggest keying the registry by `event.id` and `invariant`ing when two distinct event objects claim the same id — that turns a silent dead boot into a loud one, and keeps the test's intent (distinct events stay distinct) by making the id the contract.

Minor, same function: the `Map` entry is never deleted when its `Set` empties in the subscribe finalizer. Bounded by the number of event objects, so cosmetic.

---

packages/sdk/client-services/src/packlets/system/system-service.ts
```diff
@@ -93,12 +122,7 @@ export class SystemServiceImpl implements SystemService.Handlers {
   }
 
   ['SystemService.updateStatus']({ status }: SystemService.UpdateStatusRequest): Effect.Effect<void, Error> {
-    return Effect.tryPromise({
-      try: async () => {
-        await this._onUpdateStatus(status);
-      },
-      catch: (error) => error as Error,
-    });
+    return Effect.sync(() => this.statusRequested.emit(status));
```
dmaretskyi, line 129: `updateStatus` is now a no-op in production. `statusRequested` has exactly one subscriber in the tree — `system-service.test.ts`. The old `onUpdateStatus` acquired/released the resource lock (`ACTIVE` → `#resourceLock.acquire()`, `INACTIVE` → `release()`), and that packlet is deleted in this PR.

The public API on top of it is `Client.resumeHostServices()`, which now silently does nothing. Nothing calls it today (I grepped — the only hit is its own definition), so nothing regresses right now, but it is a live exported method whose contract is gone.

Suggest picking one in this PR rather than leaving it: either wire `statusRequested` into the embedders, or delete `resumeHostServices` + the `updateStatus` rpc along with the locks packlet.

---

packages/sdk/client-services/src/packlets/system/system-service.ts
```diff
@@ -4,56 +4,85 @@
...
+   * Runs the reset chain: the embedder closes the stack, wipes its storage, and reloads. Reports
+   * inactive first so the app falls back at once, and that status is never cleared because the app
+   * reloads.
+   */
+  'reset'(): Effect.Effect<void> {
+    return Effect.gen({ self: this }, function* () {
+      log.info('resetting...');
+      this.#resetting = true;
+      this.#statusChanged.emit(SystemStatus.INACTIVE);
+      yield* EffectEvent.emit(Closing, undefined);
```
dmaretskyi, line 80: This `emit` closes the stack out from under the RPC that is calling it. `Closing`'s handler in the worker runtime does `Scope.close(stackScope)`, and `emit` awaits the handler — so by the time this line returns, every component behind the session's server (built over `Layer.succeedContext(stack)`) is closed, while that server is still serving.

DESIGN.md's open question covers the stale-handler half ("existing sessions keep the old handlers until shutdown"). The half it does not cover is concurrent traffic: another session's in-flight request can now observe a half-torn-down stack mid-finalizer, and component finalizers are not written to be re-entrant against live requests.

It is probably fine in practice because a reset shuts the worker down immediately after — but that is exactly the kind of reasoning worth a comment here, or a gate that rejects new requests once `#resetting` is set.

---

packages/sdk/client/src/services/local-client-services.ts
```diff
@@ -182,87 +275,61 @@ export class LocalClientServices implements ClientServicesProvider {
...
+        autoConnect: this._params.autoConnect,
+      }).pipe(
+        Layer.provideMerge(sqliteLayerFromParams(this._params)),
+        Layer.provide(Layer.succeed(ConfigService, config)),
+        Layer.provide(Layer.succeed(EffectEvent.Bus, this._bus)),
         Layer.orDie,
       ),
     );
+    this._runtime = runtime;
+    this._stack = await runtime.context();
```
dmaretskyi: The leak this `TODO(mykola)` described is still here — but the TODO was removed with the old code.

If `runtime.context()` or the boot emit below throws, `_isOpen` stays `false`, so `close()` returns early and never reaches `_closeStack()`. The `ManagedRuntime` is never disposed, and with `opfsWorkerLayer` now owning the OPFS worker inside that layer, the worker is leaked too — same failure mode the deleted comment warned about, just relocated.

Wrapping from here through the handler setup in `try { … } catch { await this._closeStack(); throw; }` covers it.

---

packages/core/mesh/rpc/src/RpcRouter.ts
```diff
@@ -0,0 +1,224 @@
...
+   * Longest prefix wins, so a `''` route is the catch-all. A prefix can be served once at a time.
+   */
+  serve<Rpcs extends Rpc.Any>(
+    prefix: string,
+    group: RpcGroup.RpcGroup<Rpcs>,
+    options?: ServeOptions,
+  ): Effect.Effect<void, never, Scope.Scope | Rpc.ToHandler<Rpcs> | Rpc.Middleware<Rpcs> | Rpc.ServicesServer<Rpcs>>;
+}
+
+export class RpcRouter extends Context.Service<RpcRouter, Service>()('@dxos/rpc/RpcRouter') {}
```
dmaretskyi, line 39: This module has no production consumer — the only references anywhere are `RpcRouter.test.ts` and the changeset. That is 224 lines of non-trivial multiplexing (per-client request→route maps, `Eof` fan-out with pending-end refcounting, longest-prefix dispatch, a fallback group server for unknown tags) landing without any real traffic exercising it, and `knip` flags dead files.

Is this groundwork for a follow-up that wants several groups on one transport, or a leftover from an approach the session RPC server no longer takes? If the former, I would hold it back until the consumer lands — the invariants here are much easier to get right against a real caller than against tests written alongside them.

---

packages/sdk/client-services/src/packlets/identity/identity-lifecycle.ts
```diff
@@ -0,0 +1,111 @@
...
+      await emit(EffectEvent.emit(IdentityAvailable, { identity }));
+      log('identity created', { identityKey: identity.identityKey });
+      return identity;
+    },
+
+    acceptIdentity: async (params) => {
+      const ctx = defaultCtx;
+      const { identity, identityRecord } = await identityManager.prepareIdentity(params, ctx);
+      try {
+        await emit(EffectEvent.emit(IdentityBound, { identity, deviceCredential: params.authorizedDeviceCredential! }));
```
dmaretskyi: `params.authorizedDeviceCredential!` — non-null assertion, against the repo's no-cast/no-`!` rule, and in new code rather than carried over.

If it is genuinely always present on the accept path, that belongs in the type (`JoinIdentityProps` narrowed, or an accept-specific props type). If it can be absent, `IdentityBound`'s payload already types `deviceCredential` as optional, so just pass it through without the assertion.

---

packages/sdk/client-protocol/src/Rpc.ts
```diff
@@ -78,6 +80,45 @@ export const makeClient = <G>(
...
+ * Effect-native server for an {@link RpcGroup}: a layer that serves `group` with `handlers` over the
+ * ambient {@link RpcServer.Protocol} for the life of the layer.
+ */
+/**
+ * Re-tags a handler layer for the same rpcs carrying {@link RpcTiming.Middleware}.
+ *
+ * `Rpc.AddMiddleware` changes the handler tags' type but not the handlers: a timed server dispatches
+ * to the very values the untimed one takes, and effect-rpc offers no way to say that in the type.
+ */
+const asTimedHandlers = <Rpcs extends Rpc.Any, R>(
```
dmaretskyi, line 89: Nit: two JSDoc blocks are stacked here and the first one ("Effect-native server for an `RpcGroup`: a layer that serves `group` with `handlers`…") describes `serverLayer`, not `asTimedHandlers` — it has drifted off its symbol. Move it down to `serverLayer` so editors show the right doc on hover.

---

packages/sdk/worker-framework/src/Client.test.ts
```diff
@@ -118,19 +118,24 @@ const makeConnection = (
     maxLeaderFailures: options.maxLeaderFailures,
     onPersistentFailure: (error) => failures.push(error),
     onConnect: async ({ clientToWorker, workerToClient, isOwner }) => {
+      // The worker's client transport over this port waits, uninterruptibly, for the tab's runner to
+      // acknowledge it; a real tab serves the port, so stand in for that here or the session scope
+      // cannot close.
+      workerToClient.start();
+      workerToClient.postMessage([0]);
```
dmaretskyi: This hand-rolls effect's worker-runner ready frame (`[0]`) in a test, which is the kind of thing that breaks silently and confusingly when the upstream protocol changes — the symptom would be "session scopes stop closing", nowhere near this line.

Worth a named helper (`postRunnerReady(port)`) with the D18 explanation and a link to the upstream issue, so the next person hitting it finds the reason rather than the magic number.

---

packages/sdk/client-services/src/packlets/diagnostics/diagnostics-collector.ts
```diff
@@ -43,13 +30,7 @@ export class DiagnosticsCollector {
...
     };
 
-    const diagnostics =
-      serviceDiagnostics != null
-        ? { client: clientDiagnostics, services: serviceDiagnostics }
-        : {
-            client: clientDiagnostics,
-            broadcast: await this.broadcastSender.broadcastDiagnosticsRequest(),
-          };
+    const diagnostics = { client: clientDiagnostics, services: serviceDiagnostics ?? undefined };
```
dmaretskyi, line 33: Minor: the output shape changed here — where `serviceDiagnostics` is absent the result used to carry a `broadcast` key, and now carries `services: undefined` (which `JSON.stringify` drops entirely). Fine given the broadcast mechanism was dead, but worth confirming no diagnostics consumer (devtools panel, the feedback-form bundle reader) keys off `diagnostics.broadcast`.

The `?? undefined` itself is only load-bearing for the `null` default in the signature; `services: serviceDiagnostics ?? undefined` reads as if it were guarding more than it is.

---

dmaretskyi (commented):
Read the full diff (104 files). The direction is right and the DESIGN.md is unusually good — event causality for boot order, scope-as-lifetime for the worker/session, and layer finalizers for teardown are all a clear improvement on `ClientServicesHost`'s hand-maintained sequences.

Eleven inline comments. The ones I would not merge without addressing:

1. **`Worker.ts` session-lock race** — unhandled rejection on `Promise.race(...).then(closeSession)`, plus the worker's own lock request is never aborted when the session ends via `superseded`.
2. **`Client.ts` connect hang** — every rejection from `navigator.locks.request` is swallowed without waking `granted`, so an already-aborted `ctx.signal` leaves `connect` blocked forever with no timeout.
3. **`Event.ts` keys events by object reference** — a duplicated module instance (dual resolution, dep-optimizer re-bundle, HMR) turns the entire boot cascade into a silent no-op. Keying by `event.id` with an invariant on collisions makes that failure loud.

Medium: `SystemService.updateStatus` is now a production no-op (and `Client.resumeHostServices` with it); `RpcRouter` lands with no consumer outside its own tests; `LocalClientServices.open()` still leaks the runtime and OPFS worker on failure, and the TODO that documented it was deleted; reset closes the stack under in-flight RPCs.

Two questions beyond the inline ones:

1. Should `Event.Bus` and `ConfigService` become layer requirements rather than options (your own Outstanding item) before this lands? The current double-provide in `ClientServicesLayer` is the kind of redundancy the next embedder will copy.
2. Was the loss of per-request readiness errors (D17) checked against the tab's connect-failure UX, or is that still open?

Note on scope: this is a static read of the diff. I did not build or run any tests — no `node_modules` in this environment — so nothing here is verified against a running stack.

---

packages/sdk/worker-framework/src/Client.test.ts
```diff
@@ -448,3 +452,149 @@ describe('Connection multi-client', () => {
...
+    });
+
+    await asyncTimeout(connection.open(), 10_000);
+    await asyncTimeout(worker.sessionOpened(connection.clientId), 5_000);
+    expect(worker.sessionsOpened).toEqual([connection.clientId]);
+    await sleep(200);
+    expect(worker.sessionsClosed).toEqual([]);
+  });
+
+  test('closing a tab releases its session lock, which closes its session scope', async () => {
```
wittjosiah, line 561: how is this guaranteed to release if a tab crashes/otherwise shuts down unexpectedly?

packages/common/effect/src/Event.ts
dmaretskyi: Fixed in 7594f6b: the registry is keyed by `event.id`, so publisher and subscriber find each other across a duplicated module instance.

On the invariant: a duplicated module and a genuine collision both present as two objects with one id, so identity cannot separate them — asserting on it would fail exactly the case keying by id is meant to fix. What can only be a collision is a differing declared shape, so the invariant fires when one id is declared with two dispatch strategies. Tests now pin both halves: delivery across a separate event object with the same id, and a failure on the strategy mismatch.

Map entries are deleted when the last handler unsubscribes.

---

packages/sdk/worker-framework/src/Worker.ts
dmaretskyi: Both fixed, (2) in 6d314df and (1) in 7594f6b.

The rejection path is now explicit, close to your sketch: `.then(closeSession, (err) => { log.catch(err); void closeSession(); })`, with a comment naming what an unhandled one would cost (the session staying open for the worker's life). The abort half passes an `AbortSignal` aborted when `superseded` wins; its `AbortError` becomes a never-settling promise so the race still resolves from the `superseded` arm, and anything else rethrows into the handler above.

---

packages/sdk/worker-framework/src/Client.ts
dmaretskyi: Fixed in 7594f6b: `granted` carries `Error | undefined`, the `catch` wakes it with the error, and the attempt rethrows it. An already-aborted `ctx.signal` now fails the attempt — which the connect task's own `catch` already handles (logs, reschedules) — instead of parking it forever.

---

packages/sdk/client/src/services/local-client-services.ts
dmaretskyi: Fixed in 7594f6b: the body from `runtime.context()` through the RPC bridge is wrapped, and the `catch` calls `_closeStack()` before rethrowing, with a comment restating what the old `TODO(mykola)` warned about — now including the OPFS worker the layer owns.

Also moved `this._isOpen = true` to after the try, so the flag means "fully open" and cannot be observed true on a half-built stack.

---

packages/sdk/client-services/src/packlets/identity/identity-lifecycle.ts
dmaretskyi: Fixed in 7594f6b: the `!` is gone. `IdentityBound`'s payload already types `deviceCredential` as optional, so it passes through as-is — the second of your two options, since `prepareIdentity` accepts the props without it.

---

packages/sdk/client-protocol/src/Rpc.ts
dmaretskyi: Fixed in 7594f6b — my drift when I extracted `asTimedHandlers`. The server doc is back on `serverLayer`.

---

packages/sdk/worker-framework/src/Client.test.ts
dmaretskyi: Fixed in 7594f6b: extracted as `postRunnerReady(port)`, with the D18 reasoning on it — that the frame is `@effect/platform-browser`'s `BrowserWorkerRunner` start-up ack, that the worker's client transport awaits it uninterruptibly, and that if effect changes the frame the symptom is sessions not closing with nothing pointing back here.

No upstream issue to link — I did not open one. Worth doing if we want the uninterruptible wait bounded rather than worked around.

---

packages/sdk/client-services/src/packlets/system/system-service.ts
dmaretskyi: Documented in 7594f6b rather than gated: a comment on the `Closing` emit stating that it tears the stack down under this call and under any request another session has in flight, that component finalizers are not re-entrant against live traffic, and that this holds only because every embedder shuts down or reloads immediately after `Reset` — a reset that left the worker serving would need the gate you describe.

Chose the comment over the `#resetting` gate because the gate is only correct paired with that guarantee, and adding it here would imply the worker is expected to survive a reset, which is the opposite of what the chain does. Happy to add it if you would rather have the invariant enforced than stated.

---

packages/sdk/client-services/src/packlets/diagnostics/diagnostics-collector.ts
dmaretskyi: Confirmed: nothing keys off `diagnostics.broadcast`. Grepped the monorepo for `.broadcast` — the only hits are `plugin-iroh-beacon`'s transport and `BroadcastChannel` usage, neither related to the diagnostics payload. The devtools panel and the feedback-form bundle reader both go through `services`.

Leaving the `?? undefined` as-is: it is doing exactly the small job you describe (normalizing the `null` default), and the shape change it produces is the intended one now that the broadcast mechanism is gone.

---

packages/sdk/client-services/src/packlets/system/system-service.ts
dmaretskyi: Confirmed dead: `statusRequested` has one subscriber (`system-service.test.ts`), the `SystemService.updateStatus` rpc has one caller (`Client.resumeHostServices`), and that method has none anywhere in the repo.

Not picking for you, because the delete crosses a published surface and I would rather you choose than find out: removing `resumeHostServices` and the rpc touches `@dxos/protocols` (`SystemService.Rpcs`), `@dxos/client-protocol`'s deprecated `ClientServices` entry, the impl and its test, and `@dxos/client` — a breaking removal for any out-of-repo consumer (edge) that calls it, which the changeset would have to name.

Say the word and I will do it in this PR; it is about twenty minutes. Wiring `statusRequested` into the embedders instead is the larger of the two, since the behaviour it used to drive (the resource lock) left with the locks packlet, so it would be new design rather than a reconnect.

---

packages/core/mesh/rpc/src/RpcRouter.ts
dmaretskyi: Your read is right: no production consumer. The session server takes the other approach — one merged `ClientServicesRpcs` group served through `layerClientServicesServer`, so nothing multiplexes several groups over one protocol today.

Leaving the call to you rather than deleting 224 lines of yours: it is the kind of thing that is groundwork or leftover depending on intent I do not have. If it is groundwork, holding it until the consumer lands costs nothing and gets the invariants tested against real traffic; if it is leftover, I will drop the module and its tests in this PR and take the `rpc` changeset line with it.

One correction to the knip claim: `RpcRouter` is exported from `packages/core/mesh/rpc/src/index.ts`, so knip sees it as public API, not a dead file — which is also why nothing has flagged it.

---

packages/common/effect/src/Event.ts
dmaretskyi: Because `emit` has to be a barrier: it completes only once every handler has, so the boot cascade can rely on `StorageReady` meaning storage *is* ready, and a handler can emit a follow-up event that the emitter also waits on. `PubSub` is fire-and-forget into subscriber queues — the publisher resumes immediately and the ordering guarantee the lifecycle depends on is gone.

The second reason is failure propagation. A handler that fails must fail the emit (defects, deliberately — an event cannot enumerate the errors of every subscriber). With `PubSub` the failure surfaces in whatever fiber drains the queue, nowhere near the component whose open depends on it.

What is genuinely missing versus `PubSub` is backpressure and replay, neither of which a lifecycle cascade wants: every event here fires once per runtime and late subscribers are a bug, not a case to buffer for. If we later want a true broadcast stream (many consumers, no barrier), that is a different primitive and `PubSub` would be the right one.

---

packages/sdk/client-services/src/packlets/services/client-services-stack.test.ts
dmaretskyi: Right — renamed in 8e5e2ad to `client-services-stack.test.ts`, with the suite renamed `ClientServicesLayer`, which is what it actually builds now. Contents unchanged, 5 tests green.

---

dmaretskyi: Correction to my sha above: the rename is 3822d21, not 8e5e2ad.

---

dmaretskyi:
Nine of the eleven inline comments are fixed and pushed (`7594f6b`, `3822d21`, on top of `6d314df` for the earlier round). Two I have left for you to decide, with the reasoning on their threads: the `updateStatus` / `resumeHostServices` deletion (it crosses a published surface — breaking for any out-of-repo caller) and whether `RpcRouter` is groundwork or leftover.

All three you called blocking are done:

1. **`Worker.ts` session-lock race** — explicit rejection handler, and the request is aborted when `superseded` wins.
2. **`Client.ts` connect hang** — `granted` carries the error, the attempt rethrows it, so an aborted `ctx.signal` becomes a connect failure the task already handles.
3. **`Event.ts` identity** — keyed by `event.id`. One note: the invariant you suggested cannot be on object identity, since a duplicated module and a genuine collision look identical that way; it fires on a differing dispatch strategy instead, which can only be a collision. Both halves are pinned by tests.

Also: the `LocalClientServices.open()` leak (plus `_isOpen` set only on success), the non-null assertion, the reset comment, the runner ready-frame helper, my own JSDoc drift, and — from @wittjosiah — `service-host.test.ts` renamed to `client-services-stack.test.ts`.

**Your two questions.**

1. **`Event.Bus` / `ConfigService` as layer requirements** — already done, in `a541fe8b` and refined since. `ClientServicesLayer` takes them as requirements and re-provides them so the built context is unchanged; the three embedders each provide them once beneath. The double-provide you saw is gone.
2. **D17 readiness errors vs the tab's connect-failure UX** — still open, and not checked. The gate moved from per-request to once before the server is built, so a startup error now fails the session rather than each call; what the tab renders for that is untested. I would rather that be a separate look than a guess here.

Verified on `3822d21`: `client` 34, `client-e2e` 174, `client-services` 190, `effect` 69, `protocols`, `client-protocol`, `rpc`, `worker-framework` 27 node + 13 browser, lint and format clean. CI was green on the previous head and is running on this one.

On your scope note — understood that this was a static read. Everything above is verified against a run.

---

packages/sdk/worker-framework/src/Client.test.ts
dmaretskyi: because its done via a WebLock wich gets vacated by the browser, if the tab holding it crashes

packages/common/effect/src/Event.ts
wittjosiah: I was confused by the `Event` / `Bus` naming here, so I asked Claude to work out why. Its diagnosis:

> The explanation above is a good reason not to use `PubSub`. The same properties are why "event bus" is a confusing name for this.
>
> Most people read "event bus" to mean:
>
> - the emitter announces something and moves on
> - it doesn't know or care who's listening
> - a failing subscriber is that subscriber's problem
>
> This implementation does the opposite. `emit` waits for every handler, including the events those handlers emit. A handler failure fails the emitter. And boot depends on specific subscribers existing: if nothing handles `StorageReady`, the stack never opens. Those are the barrier properties the lifecycle needs, but "bus" gives no hint of them.
>
> The event names are misleading for the same reason. `emit(StorageReady)` sounds like a quick notification, but it doesn't return until the network is up and data spaces are open. `IdentityBound` reads more like a precondition ("must be bound before it joins") than something that happened.
>
> There's also a practical cost: `@dxos/async` already exports `Event`. Several files in this PR import both and alias the new one as `EffectEvent`.
>
> The closest match is webpack's Tapable hooks. `AsyncParallelHook` and `AsyncSeriesHook` map directly onto the `strategy` option, and callers await the hook until every tap has run. Naming it `Hook` would say "wait for all of these" and avoid the `Event` clash. If the names stay, a doc comment on `emit` saying it waits for all handlers and propagates their failures would help.

---
_Diagnosis generated by [Claude Code](https://claude.ai/code)_

dmaretskyi: Agreed on the substance — the properties are barrier properties, not bus properties, and `Hook` (Tapable's `AsyncParallelHook`/`AsyncSeriesHook`) is the closer analogy, `strategy` being exactly that choice.

Doing the cheap half now: the `emit` JSDoc says explicitly that it completes only once every handler has, including the events those handlers emit in turn, and that a handler failure fails the emitter as a defect — so the barrier semantics are stated at the call site rather than inferred from the name.

Leaving the rename to @dmaretskyi rather than doing it unasked: `Event` → `Hook` touches the module, the `Bus`/`BusService` tag (`@dxos/effect/Event.Bus` is a service key, so renaming it is a context-key change), every `EffectEvent` alias, and ~20 call sites across `client-services`. It is mechanical, but it is a public surface of `@dxos/effect` and it would land as churn on top of an already large diff. Two ways to take it:

1. Rename in this PR (`Event` → `Hook`, `Bus` → `Hook.Registry` or similar, dropping the `EffectEvent` aliases since the clash with `@dxos/async`'s `Event` goes away).
2. Land as-is with the documented semantics and do the rename in a follow-up, which keeps this PR reviewable against the review already done.

Say which and I'll do it.

---

dmaretskyi: Taking option 2: @dmaretskyi asked to land this PR, so it goes in with the documented barrier semantics and the `Event` → `Hook` rename follows separately, keeping this diff reviewable against the review already done.

---

PR #13098, opened 2026-09-14
network-manager, teleport: fix sessions that stall or tear down while connecting or closing

wittjosiah:
Four defects in swarm signalling and the teleport muxer that stalled or tore down peer connections.

- **tell the remote when a session dies before it connects.** A session that ended before its transport connected closed silently, so the remote waited out the 10s transport timeout. The swarm protocol gains a Close message that such a peer sends, and the remote closes its matching session and reconnects on backoff. TRANSPORT_CONNECTION_TIMEOUT moves to transport.ts. Teleport.open now throws when aborted mid-open, and Peer ignores errors from connections it no longer holds.
- **keep buffered channel frames ahead of writes made during the flush.** On OpenChannel the muxer set remoteId and then awaited each buffered frame in turn, so a write made during the flush went straight to the balancer ahead of frames still in the buffer. The first OpenChannel now enqueues the whole buffer synchronously before setting remoteId. New writes wait for the flush before starting their own deadline, and a repeated OpenChannel is ignored. _sendCommand keeps its signature.
- **release queued sends when a muxer is destroyed.** Muxer.destroy left the balancer's queued send triggers and the framer's pending callbacks unresolved, so callers waited out full timeouts. Destroy now wakes them, and the courtesy close uses the short destroy timeout.
- **run a rate-limited MMST update later instead of dropping it.** MMSTTopology dropped a connect or disconnect that fell inside its rate-limit interval, and charged the interval even when it had nothing to do. A rate-limited action now schedules one follow-up update, and the interval is charged only when there were peers to act on.

## Evidence

- tell the remote when a session dies before it connects: Regression test in swarm.test.ts: a session the answerer fails before connecting is retried within TRANSPORT_CONNECTION_TIMEOUT. On BASE the initiator's Connection waits its 10s connect timeout. The same test caught B51 (d42a73c72c), so both ship together.
- keep buffered channel frames ahead of writes made during the flush: Regression test in muxer.test.ts, 'a write made after the remote opens the channel waits on the link': one OpenChannel, then in-order delivery of buffer plus write. BASE sets remoteId before awaiting each buffered frame, so that write enters the balancer ahead of frames 1..19. The repeated-OpenChannel test is defensive; BASE never sends OpenChannel twice.
- release queued sends when a muxer is destroyed: Regression test in muxer.test.ts: destroying a muxer releases writes waiting on the link. On BASE, balancer and framer destroy leave their triggers and callbacks pending, so the write waits out its send timeout.
- run a rate-limited MMST update later instead of dropping it: Fake-timer regression tests in mmst-topology.test.ts: a rate-limited connect runs later, empty passes do not charge the interval, and destroy cancels the deferred update.

## Impact

- Real users: a failed WebRTC offer made the initiator wait 10s before retrying, and extensions attached to aborted teleport sessions.
- Real users: a write made while a teleport channel's buffer was flushing could overtake still-buffered frames, corrupting that channel's stream.
- Real users: closing a teleport session left writes hanging for the full send timeout.
- Real users: a swarm could stay unconnected after a rate-limited connect.

## Testing

Each layer was checked out on its own and passed format, lint, the build of every project it touches, and:

```bash
( cd packages/core/mesh/network-manager && CI= pnpm vitest run src/swarm src/signal && cd ../teleport && CI= pnpm vitest run ) &&
( cd packages/core/mesh/teleport && CI= pnpm vitest run src/muxing/muxer.test.ts ) &&
( cd packages/core/mesh/teleport && CI= pnpm vitest run src/muxing/ ) &&
( cd packages/core/mesh/network-manager && CI= pnpm vitest run src/topology/mmst-topology.test.ts )
```

Part of DX-1264. PR 2 of 12 in a stack split out of #13013 by fix area; the workarounds, mitigations and diagnostics stay on #13013 for separate evaluation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer connection reliability by preventing stalls during connection setup and notifying peers when sessions end early.
  * Preserved message ordering and ensured queued sends are released when connections close.
  * Improved handling of transport startup failures, closed transports, and unavailable connection statistics.
  * Rate-limited topology updates are now deferred instead of being dropped.
  * Improved channel and teleport behavior during shutdown or cancellation.

---
🚀 Composer preview: https://pr-13098-composer-dev.dxos.workers.dev


packages/core/mesh/network-manager/src/transport/webrtc/rtc-transport-service.ts
```diff
@@ -185,10 +222,72 @@ export class RtcTransportService implements BridgeService {
...
+   */
+  private _wasClosedHere(proxyId: PublicKey): boolean {
+    return this._closedTransports.some((id) => id.equals(proxyId));
+  }
+
   private async _safeCloseTransport(transport: TransportState): Promise<void> {
     if (this._openTransports.get(transport.proxyId) === transport) {
       this._openTransports.delete(transport.proxyId);
     }
+    this._rememberClosed(transport.proxyId);
```
dmaretskyi: `_rememberClosed` runs unconditionally, but the line above it guards the `_openTransports.delete` with `=== transport` precisely because this can be called for a state that is no longer the live one for that proxy id.

`open()` takes that path: on a re-open it calls `void this._safeCloseTransport(existingTransport)` and then `_produceTransport` registers the new transport. If the `void`ed close lands after the new registration, `_rememberClosed` will (a) wake and delete the *new* transport's registration trigger and (b) push the proxy id onto `_closedTransports` permanently — so any write that arrives in the new transport's pre-registration window is silently dropped instead of held, which is the exact failure this PR is fixing.

Suggest moving `_rememberClosed(transport.proxyId)` inside the `if (this._openTransports.get(transport.proxyId) === transport)` block. Agreed this only bites on the re-open path (which is already `log.error`'d, and proxy ids are random per `RtcTransportProxy`), but the guard is free.

---

packages/core/mesh/teleport/src/muxing/balancer.ts
```diff
@@ -79,6 +79,12 @@ export class Balancer {
     if (this._sendBuffers.size !== 0) {
       log.info('destroying balancer with pending calls');
     }
+    // Nothing queued will go out now; release the waiters instead of leaving them to time out.
+    for (const buffer of this._sendBuffers.values()) {
+      for (const { trigger } of buffer) {
+        trigger?.wake();
```
dmaretskyi, line 86: `wake()` resolves the trigger as *success*, so every caller blocked on a queued send is told its frame went out when the balancer has just dropped it. Same in `Framer.destroy` below. Up the stack `Muxer._awaitCommand` sees a clean resolve, so `_writeToChannel` — and therefore an extension's `port.send` / RPC call — returns as delivered.

Releasing the waiters is clearly right versus hanging for the full timeout; the question is the signal. `trigger.throw(new Error('muxer destroyed'))` would report the truth, at the cost of `_awaitCommand` re-entering `destroy` (already in progress, so it should be a no-op) and of callers seeing rejections they previously saw as timeouts. Worth a deliberate decision either way, and a one-clause comment recording it, since "resolved" here means "dropped".

---

packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -379,6 +401,17 @@ export class Peer {
     log('closed', { peerId: this.remoteInfo, sessionId: connection.sessionId });
   }
 
+  async onClose(_ctx: Context, message: CloseMessage): Promise<void> {
```
dmaretskyi, line 404: `onClose` tears the connection down on any session-id match, but the message is defined (proto comment, `SessionClosedByRemoteError` text, the sender-side `!connected` guard) as "the session ended *before its transport connected*". Nothing here checks that locally.

Consequence: a `Close` that races a successful connect — e.g. the remote's `onClosed` fires for a displaced/aborted attempt while this side has already reached CONNECTED on the same session — kills a live connection. It also hands a peer a cheap way to drop an established session.

Suggest matching the sender's precondition: ignore the close unless `this.connection.state` is still pre-CONNECTED.

---

packages/core/mesh/network-manager/src/swarm/peer.ts
```diff
@@ -388,6 +421,18 @@ export class Peer {
     await this.connection.signal(ctx, message);
   }
 
+  private _announceClose(sessionId: PublicKey, err?: Error): void {
```
dmaretskyi, line 445: Two things:

1. This sends on `this._ctx`, which `safeDestroy` disposes *before* closing the connection. So the one case where the remote most needs telling — this peer going away mid-handshake — is the case where the announcement is silently dropped (it will surface only as the `session close not delivered` debug log). If that is intentional (swarm-leave covers it), say so in the comment; otherwise the send needs a context that outlives `_ctx`.
2. Placement: private method sits between the public `onClose`/`onSignal` and `safeDestroy`. `code-style` wants publics before privates.

---

packages/core/mesh/network-manager/src/topology/mmst-topology.test.ts
```diff
@@ -0,0 +1,100 @@
...
+//
+
+import { describe, expect, onTestFinished, test, vi } from 'vitest';
+
+import { PublicKey } from '@dxos/keys';
+
+import { MMSTTopology } from './mmst-topology.ts';
+import { type SwarmController, type SwarmState } from './topology.ts';
+
+const MIN_UPDATE_INTERVAL = 10_000;
```
dmaretskyi: This duplicates the constant the code under test uses. Change `MIN_UPDATE_INTERVAL` in `mmst-topology.ts` and these tests keep passing against the old number (or start failing for a reason that has nothing to do with the behaviour). Export it from the module and import it here.

---

.changeset/connection-transport.md
```diff
@@ -0,0 +1,5 @@
+---
+'@dxos/network-manager': patch
```
dmaretskyi, line 3: The changeset names only `@dxos/network-manager`, but the PR ships behaviour changes in `@dxos/teleport` (muxer flush ordering, balancer/framer release on destroy, `Teleport.open` now throwing) and a wire-format addition in `@dxos/protocols` (`Close`). Both are consumer-relevant and neither gets a version bump as written — add them to the frontmatter.

---

packages/core/mesh/network-manager/src/transport/webrtc/rtc-transport-service.ts
```diff
@@ -68,103 +78,130 @@ export class RtcTransportService implements BridgeService {
...
   }
 
   async getStats({ proxyId }: StatsRequest): Promise<StatsResponse> {
-    const transport = this._openTransports.get(requirePublicKey(proxyId));
-    invariant(transport);
+    // Sampled on an interval for the life of the connection, so it must never wait on one.
+    const transport = await this._resolveTransport(requirePublicKey(proxyId), { wait: false });
+    if (!transport) {
+      return create(StatsResponseSchema, {
+        stats: { bytesSent: 0, bytesReceived: 0, packetsSent: 0, packetsReceived: 0, rawStats: 'transport closed' },
```
dmaretskyi: Nit: returning all-zero counters for a closed transport makes the sample indistinguishable from a live-but-idle connection in whatever aggregates these — the `rawStats` marker is the only tell, and it is a free-form string. Consider leaving `stats` unset so the caller can skip the sample rather than record a reset to zero.

---

packages/core/mesh/teleport/src/teleport.ts
```diff
@@ -151,6 +151,9 @@ export class Teleport {
     log('open');
     this._setExtension('dxos.mesh.teleport.control', this._control);
     await this._openExtension('dxos.mesh.teleport.control');
+    if (this._aborting || this._destroying) {
+      throw new Error('Teleport closed while opening.');
```
dmaretskyi: Nit: a bare `Error` here is indistinguishable from a real fault to every `catch` upstream — `Peer.onClosed` will now announce a `Close` carrying this string as the reason for what is a perfectly ordinary abort. Prefer a typed error (`CancelledError` from `@dxos/protocols`, as the rest of the abort paths use) so callers can tell "we cancelled" from "it broke".

---

packages/core/mesh/teleport/src/muxing/muxer.ts
```diff
@@ -437,11 +481,13 @@ export class Muxer {
...
       return;
     }
-    await this._sendCommand(
+    const send = this._enqueueCommand(
       create(CommandSchema, { payload: { case: 'data', value: { channelId: channel.remoteId, data } } }),
       channel.id,
-      timeout,
     );
+    // The deadline covers this write's own send, not the backlog queued ahead of it at open.
+    await channel.flushing;
```
dmaretskyi, line 489: The enqueue-then-await-flush split is the right shape, and ordering is safe because `_enqueueCommand` runs before the `await`. Worth noting for the record that the `timeout` argument now bounds strictly less than it used to: a write issued during a long flush waits `flushing` *plus* `timeout`, so the caller's deadline is no longer a bound on total time-to-send. That is deliberate per the comment, but callers that pass a timeout expecting a wall-clock guarantee (`port.send(data, 1_500)` in the new test) will not get one — worth stating in the JSDoc of the public `send`/`createPort` path rather than only here.

---

dmaretskyi (commented):
Reviewed the full diff. The five fixes are well-scoped, each has a regression test that fails on BASE for the stated reason, and the enqueue/await split in the muxer is the right way to get ordering without changing `_sendCommand`'s signature. Comments below — one I'd call a genuine (if narrow) bug, one a correctness question on the new `Close` handling, the rest smaller.

Highlights:

1. **`_safeCloseTransport` marks a proxy id closed unconditionally** — the `=== transport` guard exists one line above for a reason; the re-open path can race a live registration and permanently poison the id.
2. **`Peer.onClose` doesn't check the connection is still pre-CONNECTED**, though every other part of the design says the message only means "died before connecting". A `Close` racing a successful connect kills a live session.
3. **Balancer/framer release-on-destroy resolves waiters as success**, so a dropped frame is reported to the caller as sent. Probably still better than hanging, but it deserves an explicit decision and a comment.
4. Changeset omits `@dxos/teleport` and `@dxos/protocols`, both of which change behaviour/wire format here.

Things I checked and am satisfied with: `sendData` reordering is not possible because `RtcTransportProxy`'s `connectorStream` is a `Writable` with a single in-flight write; `Stream`'s constructor does catch a producer throw, so the `open()` catch → `_rememberClosed` path the third test exercises really is reachable; `OPEN_REGISTRATION_TIMEOUT` (9s) does sit under the proxy's `RPC_TIMEOUT` (10s); old peers parse the new `close` oneof arm as an empty payload and fall through to the `unknown message` warning rather than crashing.

---

packages/core/mesh/network-manager/src/transport/webrtc/rtc-transport-service.ts
wittjosiah: Fixed in fc7212bc4b. `_rememberClosed` now runs only inside the `=== transport` guard. `close()` no longer deletes the entry first, since the guard would otherwise skip the normal close path. A registration also clears the id from the closed list.

No test for this one: `open()` registers the new transport synchronously, so a write can't land in that window today. The guard covers it if that changes.

packages/core/mesh/teleport/src/muxing/balancer.ts
wittjosiah: Kept the release as a resolve and recorded why in both places (fc7212bc4b). `_awaitCommand` never rethrows: a send failure or timeout already becomes `muxer.destroy(err)` and the send resolves. Throwing the trigger would therefore change nothing a caller sees. A caller learns about the dropped frame from the muxer and teleport closing, not from `send`.

packages/core/mesh/network-manager/src/swarm/peer.ts
wittjosiah: Fixed in fc7212bc4b. `onClose` ignores a Close unless the connection is still CREATED, INITIAL or CONNECTING. The new swarm test 'a close for a session that already connected is ignored' fails on the previous code, where the connection went CLOSED.

packages/core/mesh/network-manager/src/swarm/peer.ts
wittjosiah: Intentional. A peer being destroyed is leaving the swarm, and the remote learns that from the swarm leave. The announcement is only for a session that failed while both peers stay. The method's comment now says so, and the method sits below `safeDestroy` (fc7212bc4b).

packages/core/mesh/network-manager/src/topology/mmst-topology.test.ts
wittjosiah: Done in fc7212bc4b: `MIN_UPDATE_INTERVAL` is exported from `mmst-topology.ts`, and the test imports it.

.changeset/connection-transport.md
wittjosiah: Left as is. `@dxos/network-manager`, `@dxos/teleport` and `@dxos/protocols` are in the same `fixed` lockstep group in `.changeset/config.json`, so naming one bumps all three. `agents/instructions/changesets.md` says to name one member and never list the group.

packages/core/mesh/network-manager/src/transport/webrtc/rtc-transport-service.ts
wittjosiah: Done in fc7212bc4b. A closed transport returns no `stats`, and `Transport.getStats()` now returns `TransportStats | undefined`. `Connection` already skips an undefined sample. The proxy also returns nothing when the bridge is unreachable, instead of zeroes.

packages/core/mesh/teleport/src/teleport.ts
wittjosiah: Done in fc7212bc4b: `Teleport.open` throws `CancelledError` from `@dxos/protocols`.

packages/core/mesh/teleport/src/muxing/muxer.ts
wittjosiah: Documented on `RpcPort.send` in fc7212bc4b. The timeout bounds the message's own send. A message written while buffered frames are still flushing waits for them first, outside that bound.

PR #13099, opened 2026-09-14
worker-framework, client: surface worker and client startup failures instead of hanging

wittjosiah:
Six defects where a failure while starting a worker, a client or a space left callers waiting forever.

- **close a leader session that finishes opening after its connection closed.** #watchLeader read this._ctx at each step, so a connection closed while its leader session was opening re-armed on the fresh context close had installed and kept leading. It now captures its starting context and closes a session that finishes opening after that context is disposed. The in-thread test worker now shuts down when its port closes, as a terminated worker would.
- **close the worker when a leader session times out.** The leader handshake timeout wrapped LeaderSession.open from outside, so a timed-out session left its worker running with the storage lock. The timeout now covers the whole handshake inside LeaderSession._open, and a failure closes the worker.
- **reject the connection when the worker runtime fails to start.** A worker whose runtime failed to start still posted ready, so tabs connected to a host that could not serve them. start() now fails with the startup error, and the worker posts init-failed with the encoded error and shuts down. The initial connection rejects with that error instead of re-electing.
- **throw ClientProvider initialization failures to the error boundary.** ClientProvider caught initialization failures and kept rendering its fallback. Every initialization failure, including onInitialized, now goes through setFailure and is thrown in render to the nearest error boundary. todomvc gains an AppError route element, and its e2e page object reports that text when boot fails.
- **fail only the request whose RPC handler has a defect.** The RPC server treated a handler defect as fatal and tore down every stream on the worker port, and DataService turned rejections into defects through Effect.promise. The server now disables fatal defects and logs them, and DataService uses Effect.tryPromise. A status stream that fails after open surfaces as Client.fatalError, which ClientProvider throws to the boundary.
- **settle a SpaceProxy's triggers when its initialization fails.** When SpaceProxy initialization threw, its _databaseInitialized and _initializationComplete triggers never settled, so createSpace timed out with an unrelated error and waitUntilReady hung. Both triggers now reject with the initialization error, and the plugin-space save tracker catches the rejection.

## Evidence

- close a leader session that finishes opening after its connection closed: Regression test in Client.test.ts (@@ -433): a connection closed while its leader session opens closes that session and does not lead again. The dedicated-worker leader hand-off tests hit the same defect (fe32f6380b).
- close the worker when a leader session times out: Regression test in Client.test.ts (+477-512): a leader session that times out closes its worker, so the retry's worker starts.
- reject the connection when the worker runtime fails to start: Regression tests in Client.test.ts, worker-runtime.test.ts and WorkerProtocol.test.ts: a start failure rejects every booting tab without a retry, and the error survives the structuredClone round trip.
- throw ClientProvider initialization failures to the error boundary: Regression tests in ClientProvider.test.tsx: onInitialized and initialize rejections reach the error boundary. BASE caught them and rendered the fallback.
- fail only the request whose RPC handler has a defect: Regression tests in Rpc.test.ts, client-initialize.test.ts and ClientProvider.test.tsx: a handler defect fails only its own request while an open stream keeps delivering, and a later status failure surfaces as fatalError.
- settle a SpaceProxy's triggers when its initialization fails: Regression test in client-e2e spaces.test.ts: create rejects with the database open error. On BASE it rejects with the unrelated 5s timeout, and waitUntilReady never settles. The test passes with BASE's CREATE_SPACE_TIMEOUT still in place, because a thrown Trigger rejects the bounded wait with its own error.

## Impact

- Real users: a tab that closes mid-election keeps a leader session and the storage lock, and every later client waits on it.
- Real users: after a slow handshake timed out, the retry waited forever on the storage lock the old worker still held.
- Real users: a worker whose migration or config failed still replied ready, so tabs connected to a dead host and hung.
- Real users: a failed client boot showed a blank fallback forever, with no error.
- Real users: one throwing DataService handler killed every RPC stream on the worker port and froze the client.
- Real users: a space whose database failed to open left createSpace and waitUntilReady callers pending.

## Testing

Each layer was checked out on its own and passed format, lint, the build of every project it touches, and:

```bash
( cd packages/sdk/worker-framework && CI= pnpm vitest run src/Client.test.ts && cd ../client && CI= pnpm vitest run src/services/dedicated ) &&
( cd packages/sdk/worker-framework && CI= pnpm vitest run src/Client.test.ts ) &&
( cd packages/sdk/worker-framework && CI= pnpm vitest run  && cd ../client-services && CI= pnpm vitest run src/packlets/worker/worker-runtime.test.ts ) &&
( cd packages/sdk/react-client && CI= pnpm vitest run src/client/ClientProvider.test.tsx ) &&
( cd packages/sdk/client-protocol && CI= pnpm vitest run src/Rpc.test.ts && cd ../client && CI= pnpm vitest run src/client/client-initialize.test.ts && cd ../react-client && CI= pnpm vitest run src/client/ClientProvider.test.tsx && cd ../../core/echo/echo-client && CI= pnpm vitest run src/automerge/repo-proxy.test.ts ) &&
( cd packages/sdk/client-e2e && CI= pnpm vitest run src/spaces.test.ts )
```

One existing test in `client-e2e/src/spaces.test.ts`, `post and listen to messages`, times out on this layer. It times out the same way on the main commit this stack is based on, so it does not come from this change; the file's other tests pass, including the new one.

Part of DX-1264. PR 3 of 12 in a stack split out of #13013 by fix area; the workarounds, mitigations and diagnostics stay on #13013 for separate evaluation.

## Summary by CodeRabbit

* **New Features**
  * Startup failures now report clear errors instead of leaving the app or worker unresponsive.
  * TodoMVC displays boot and rendering errors rather than a blank screen.
  * Client providers forward initialization and fatal runtime errors to configured error boundaries.
  * Failed spaces can be retried and reopened after initialization errors.
  * Worker startup errors preserve useful details for improved diagnostics.

* **Bug Fixes**
  * Leader sessions and workers are cleaned up after timeouts or failed connections.
  * Individual RPC failures no longer disrupt unrelated requests or streams.
  * Data and space initialization failures now settle reliably with actionable errors.

---
🚀 Composer preview: https://pr-13099-composer-dev.dxos.workers.dev


dmaretskyi:
claude, this has big conflics with main, resolve them before I review. make sure to analyse the intent behind changes on main

packages/sdk/client/src/echo/space-proxy.ts
```diff
@@ -416,8 +416,16 @@ export class SpaceProxy implements Space, CustomInspectable {
...
-    await this._initializeDb(ctx);
+    try {
+      await this._invitationsProxy.open();
+      await this._initializeDb(ctx);
+    } catch (err) {
+      // A failed initialization settles both triggers so waiters observe it instead of pending.
+      const error = err instanceof Error ? err : new Error(String(err));
+      this._databaseInitialized.throw(error);
+      this._initializationComplete.throw(error);
+      throw err;
```
dmaretskyi, line 446: **Blocking:** this rethrows without restoring state, so `_initializing` stays `true` forever and the space can never initialize again — the hang becomes a permanently dead space rather than a recoverable failure.

Three consequences, all silent:

- `_processSpaceUpdate` computes `isFirstTimeInitializing = space.state === SpaceState.SPACE_READY && !(this._initialized || this._initializing)` — false from here on, so init is never retried.
- The reopen path is dead too: `isReopening` requires `!this._databaseOpen`, but `_initializeDb` sets `this._databaseOpen = true` as its first statement, *before* `await this._db.open(ctx)` can throw.
- `_setParentCtx` stops refreshing `_ctx` (it early-returns on `_initializing`).

`_reset()` already clears exactly the right set for this situation (`_initializing`, `_initialized`, `_databaseOpen`, plus `_initializationComplete.reset()` / `_databaseInitialized.reset()`); the catch should mirror at least the flags:

```ts
} catch (err) {
  // A failed initialization settles both triggers so waiters observe it instead of pending.
  const error = err instanceof Error ? err : new Error(String(err));
  this._databaseInitialized.throw(error);
  this._initializationComplete.throw(error);
  this._initializing = false;
  this._databaseOpen = false;
  throw err;
}
```

Worth deciding explicitly whether a subsequent attempt should also `reset()` the two triggers — a rejected `Trigger` stays rejected, so without that the reopen path would settle new waiters with the stale error.

(The `throw()` calls themselves are safe: `Trigger.throw` is a no-op once resolved, so a failure occurring after `_databaseInitialized.wake()` won't double-settle.)

---

packages/sdk/client/src/client/client.ts
```diff
@@ -571,8 +583,15 @@ export class Client {
...
-          trigger.wake(err);
           this._statusUpdate.emit(null);
+          if (!statusReceived) {
+            trigger.wake(err);
+            return;
+          }
+
+          // Closing interrupts the stream rather than failing it, and nothing resubscribes it.
+          log.error('system status stream failed', { err });
+          this._fatalErrorUpdate.emit(err);
```
dmaretskyi: The comment says *"nothing resubscribes it"* — but something does. `this._services.closed` (in `_open`, a little above) runs `await this._close(); await this._open(this._ctx); this.reloaded.emit();` whenever `!this._resetting`, and `_open` resubscribes the status stream.

So a worker/transport drop surfaces on **both** paths and whichever lands first wins. If the stream's `onError` fires first, `ClientProvider` now throws to the nearest error boundary and kills the app instead of letting the existing reconnect run.

Same concern for `reset()`: `_resetting` exists precisely so consumers can distinguish an intentional teardown from a lost connection — the doc on `Client.resetting` says consumers reacting to disconnect/reconnect "should check this before choosing a [recovery]" — and this emission doesn't.

Minimum fix:

```ts
// An intentional reset tears down services; `_services.closed` reopens everything else.
if (this._resetting) {
  return;
}
```

Better would be to coordinate with the `_services.closed` handler so the fatal error is only raised for a failure that path is *not* going to recover from. As it stands the new fatal path races the recovery path it's meant to complement, and `_close()` unsubscribing first is the only thing keeping `destroy()` from tripping it.

---

packages/sdk/worker-framework/src/WorkerProtocol.ts
```diff
@@ -109,6 +118,55 @@ export type CoordinatorMessage =
...
+    }
+    const path = new Set(ancestors).add(value);
+    return {
+      name: value.name,
+      message: value.message,
+      stack: value.stack,
+      cause: value.cause === undefined || path.has(value.cause) ? undefined : encode(value.cause, path),
+      errors:
+        value instanceof AggregateError
+          ? value.errors.filter((inner) => !path.has(inner)).map((inner) => encode(inner, path))
```
dmaretskyi: Nit: `filter` drops cyclic members and **shifts the surviving indices**, so `decoded.errors[i]` no longer corresponds to `original.errors[i]` — an `AggregateError` can silently lose arity. Encoding a placeholder in place preserves it:

```ts
errors:
  value instanceof AggregateError
    ? value.errors.map((inner) =>
        path.has(inner) ? { name: 'Error', message: '<cycle>' } : encode(inner, path),
      )
    : undefined,
```

The `cause` branch above is fine as-is because it degrades to `undefined` in a single slot, where there is no index to shift.

---

packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
```diff
@@ -130,25 +132,35 @@ export const makeWorkerRuntime = ({
       .pipe(Layer.orDie),
   );
 
+  // Everything `stop` does except `onStop`, which a failed `start` leaves to its caller.
+  const teardown = async (): Promise<void> => {
+    stopped = true;
```
dmaretskyi: `teardown()` sets `stopped = true` as its first statement, so once a failed `start` has run it, a subsequent `stop()` hits the `if (stopped) return;` guard and **`onStop` never fires**.

That matches the new doc comment on `start` ("after tearing down without `onStop`"), but it means the shutdown signal is unavailable to anyone who calls `stop()` expecting to get it — and `stop()` is documented as "run `onStop`. Idempotent.", which now isn't quite true after a failed start.

Either track it separately:

```ts
let onStopPending = true;
// ...
const stop = () => Effect.promise(async () => {
  try {
    if (!stopped) { await teardown(); }
  } finally {
    if (onStopPending) { onStopPending = false; await onStop?.(); }
  }
});
```

…or note on `stop()`'s doc comment that it is a no-op, `onStop` included, once `start` has failed.

---

packages/core/echo/echo-host/src/db-host/data-service.ts
```diff
@@ -18,6 +18,9 @@ import { type AutomergeHost, type DocumentLease, deriveCollectionIdFromSpaceId }
 import { DocumentsSynchronizer } from './documents-synchronizer.ts';
 import { type SpaceStateManager } from './space-state-manager.ts';
 
+// Typed failures keep their class, stack and context over `serviceError`; a defect keeps only name and message.
```
dmaretskyi: Nit: this comment doesn't describe `toError`, which is just an `unknown → Error` coercion — it reads like it was written for a different helper (nothing here mentions `serviceError` or defects). Per the repo comment rule (one load-bearing clause saying *why*), something like:

```ts
// `Effect.tryPromise` needs an `Error`, and a rejection is not guaranteed to be one.
```

The conversion from `Effect.promise` to `Effect.tryPromise` across all the handlers is the right call — `Effect.promise` was turning every rejection, including the `invariant` failures, into a defect.

---

packages/apps/todomvc/src/playwright/app-manager.ts
```diff
@@ -32,17 +32,29 @@ export class AppManager {
     this._close = close;
     this.page.on('console', (message) => this._onConsoleMessage(message));
     this.shell = new ShellManager(this.page);
-    await this.newTodo().waitFor({ state: 'visible' });
-    await this.page.getByTestId('placeholder').waitFor({ state: 'hidden' });
+    await this._waitForBoot(this.newTodo());
+    await this._waitForBoot(this.page.getByTestId('list'));
```
dmaretskyi, line 36: This swaps `getByTestId('placeholder').waitFor({ state: 'hidden' })` for `getByTestId('list')` **visible**, which is not the same condition — "placeholder gone" is satisfied by any terminal state, "list visible" requires the list to actually render.

Worth confirming `list` renders on a cold first load with no space yet; otherwise this trades the hang for a flake in exactly the first-run case the PR is trying to make legible. The `_waitForBoot` race against `app-error` is a nice pattern either way.

---

dmaretskyi (commented):
Solid change overall — every hang path is turned into a settled failure, and the tests exercise the failure rather than just the happy path. Two things I'd want resolved before merge, plus four nits inline.

**1. `SpaceProxy` is permanently wedged after a failed init** (blocking, inline on `space-proxy.ts`). The new `catch` rethrows without clearing `_initializing` / `_databaseOpen`, which disables both the first-time-init and the reopen paths in `_processSpaceUpdate`. A transient DB-open failure — the exact case the new `spaces.test.ts` test simulates with a one-shot `mockRejectedValueOnce` — becomes a permanently dead space across reconnects.

**2. `fatalError` races the existing reconnect path** (inline on `client.ts`). `_services.closed` already does `_close()` → `_open()` → `reloaded.emit()` when `!this._resetting`, and `_open` resubscribes the status stream, so the "nothing resubscribes it" premise doesn't hold for a transport drop. Whichever path lands first wins, and if it's the stream error, `ClientProvider` throws to the error boundary instead of letting the reconnect happen. At minimum, gate the emit on `_resetting`.

Nits inline: `encodeError`'s cycle filter shifts `AggregateError.errors` indices; `teardown()` makes a later `stop()` skip `onStop`; the `toError` comment in `data-service.ts` describes something else; and the todomvc boot wait changed conditions in a way worth double-checking for the cold-start case.

Things I specifically liked: `Effect.promise` → `Effect.tryPromise` across the data-service handlers (those `invariant`s were becoming defects), folding `serve`/`serveOverProtocol` into `makeServerLayer`, binding `#watchLeader` to the context it started under, and `WorkerProtocol.test.ts` covering the cyclic/AggregateError cases directly.

One coverage gap worth noting: `Rpc.test.ts` exercises a defect without timing middleware. Since `makeServerLayer` now composes `DefectLogMiddleware` on top of an already-timing-wrapped group, a case with `timing` enabled would confirm the ordering holds.

---

packages/sdk/client/src/echo/space-proxy.ts
wittjosiah: Fixed in e363eae175. The catch now clears `_initializing`, and `_initializeDb` clears `_databaseOpen` when its open fails, so the reopen path isn't blocked either.

On the triggers: a failed attempt leaves them rejected, so anyone waiting in between sees the failure. The next attempt resets both before it starts.

It retries when the space returns to ready, not on every re-sent READY update. Retrying on those made a space whose `create()` had failed wait forever for a properties object nothing would write, and that hung client teardown.

The new test 'a space whose database failed to open initializes again when it returns to ready' fails on the previous code, where the space stays INITIALIZING.

packages/sdk/client/src/client/client.ts
wittjosiah: Fixed in e363eae175. The status stream's error is no longer fatal while the client is resetting, or when the error is `RpcClosedError`, which is how a lost connection ends the stream. `_services.closed` recovers both. A handler failure on a live connection is still fatal.

`subscribeStream` reports a pure interruption as a close, not an error, so a clean transport close never reached this path anyway.

The new test 'a status stream closed by a lost connection is not fatal' fails on the previous code.

packages/sdk/worker-framework/src/WorkerProtocol.ts
wittjosiah: Done in e363eae175, with a test that fails on the previous code: the aggregated errors came back as two instead of three.

packages/sdk/client-services/src/packlets/worker/worker-runtime.ts
wittjosiah: No longer applies. Main (#13094) replaced `start`/`stop`/`onStop` with a runtime that lives as long as its scope. After the merge, `makeWorkerRuntime` closes its stack and fails with the startup error, and `Worker.run` reports `init-failed` and closes the runtime scope. There is no separate stop signal left to lose.

packages/core/echo/echo-host/src/db-host/data-service.ts
wittjosiah: Done in e363eae175, with your wording.

packages/apps/todomvc/src/playwright/app-manager.ts
wittjosiah: Checked. On a cold first load, `Root` creates the identity and a space and calls `createTodoList(space)` before the app renders, and `Todos` renders `list` once that list loads, so the wait resolves on first run. Every todomvc test starts in a fresh browser context, and the suite ran clean on all three browsers in the 10-run CI batch with this change.

I kept `list` because the race against `app-error` needs something that becomes visible; 'placeholder hidden' can't be raced that way.

wittjosiah:
Resolved the conflicts with main, merging main in from the bottom of the stack (e363eae175 on this branch). What main changed in the worker, and how it relates to the fixes here:

**#13094 moved the worker onto a scoped Effect runtime.** `Worker.run` now owns a scope the runtime lives in, and each session is a child scope. A session closes when its tab's Web Lock (the new `sessionLockKey`) releases, which is now how the worker learns a tab closed or died. The runtime reports startup errors only through its readiness gate.

- **A closed connection that kept leading, and a timed-out leader session that kept its worker:** both are fixed on the tab side and don't conflict with main. Main takes its session lock inside the same connect attempt and releases it with that attempt's context, which fits leader watching bound to the context it started under.
- **A worker whose runtime failed to start:** on main it still advertised `ready`, and tabs only saw the error when a session waited on the gate. I kept main's structure. `makeWorkerRuntime` now closes its stack and fails with the startup error, so `Worker.run` reports `init-failed` and shuts down. The runtime tests are rewritten for the scoped API, since `start`/`stop`/`onStop` are gone.
- **A handler defect failing every request on the connection:** main added `Rpc.serverLayer` for worker sessions, and it skipped this PR's defect middleware. It now goes through the same server builder as `serve`. The RPC test covers `serve` and `serverLayer`, each with and without timing, and the `serverLayer` cases fail without the change.
- **The in-thread test worker:** main shuts it down through `requestShutdown`. I kept this PR's shutdown on port close, which the leader tests need.
- **Other main changes:** the Event → Hook rename, the SqlTransaction removal and the idle-CPU work merged without conflicts.

The review threads are answered inline, and each fix has a test that fails without it. Main was merged into #13097 and #13098 first, and the result was merged up through #13100–#13102. Every layer builds, and its tests pass.

packages/sdk/client-e2e/src/spaces.test.ts
```diff
@@ -62,6 +62,60 @@ describe('Spaces', () => {
...
+    const [client] = await createInitializedClients(1, { storage: true });
+    const error = new Error('Database open failed.');
+    const openSpy = vi.spyOn(DatabaseImpl.prototype, 'open').mockRejectedValueOnce(error);
+    onTestFinished(() => openSpy.mockRestore());
+
+    await expect(client.spaces.create()).rejects.toBe(error);
+  });
+
+  test('a space whose database failed to open initializes again when it returns to ready', async () => {
+    const testBuilder = new TestBuilder();
```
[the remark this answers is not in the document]
wittjosiah: Done in 5d7f645317: the test registers `testBuilder.destroy()` first, so it runs after the host, server and client close.

packages/sdk/worker-framework/src/Client.ts
```diff
@@ -352,14 +359,30 @@ export class Connection extends Resource {
...
+          // A runtime that failed to start ends the initial connection with its error instead of respawning.
+          this.#leaderFailureCount++;
+          this.#leaderPhase = 'session-failed';
+          this.#lastLeaderError = startFailure;
+          log.warn('worker-connection: worker runtime failed to start', {
+            clientId: this.#clientId,
+            error: startFailure,
+          });
+          // A connection whose open failed is never closed, so it stops its own connect task and listeners.
+          await ctx.dispose();
```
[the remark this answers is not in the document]
wittjosiah: Done in 5d7f645317: a terminal start failure closes `#connectTask` after rejecting `#initialConnection`. Closing it first waits on the connect run still in flight and delays the rejection.

packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -86,12 +86,13 @@ export type Options = {
...
    * `requestShutdown` called), and every session scope is a child of it, so shutdown closes the
-   * sessions first and then the runtime. Put the runtime's teardown in its finalizers.
+   * sessions first and then the runtime. Put the runtime's teardown in its finalizers. A failure is reported
+   * to the leader as `init-failed` and shuts the worker down.
    */
   createRuntime: (args: {
     config: Record<string, any> | undefined;
     requestShutdown: () => void;
-  }) => Effect.Effect<RuntimeHandle, never, Scope.Scope>;
+  }) => Effect.Effect<RuntimeHandle, Error, Scope.Scope>;
```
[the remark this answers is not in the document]
wittjosiah: Done in 5d7f645317. `makeWorkerRuntime` and `layerWorkerRuntime` fail with `WorkerRuntimeStartError` from `@dxos/protocols`, which keeps the original message and carries the failure as its cause. `Worker.Options.createRuntime` fails with `BaseError`, because worker-framework does not depend on protocols.

dmaretskyi (approved):

PR #13114, opened 2026-09-15
sql-sqlite: remove the SqlTransaction service (effect rc.115)

dmaretskyi:
## Summary

`@dxos/sql-sqlite`'s `SqlTransaction` service existed because `@effect/sql-sqlite-do`'s client could not transact inside a Durable Object — its `withTransaction` emitted literal `BEGIN` / `COMMIT`, which workerd forbids. So each platform supplied its own implementation, and every store carried `SqlTransaction` alongside `SqlClient` in its requirements.

effect `4.0.0-rc.115` fixes this upstream: a DO client built with `storage: ctx.storage` backs `SqlClient.withTransaction` with `ctx.storage.transaction()`.

**The service, its `layer` and its `clientLayer` are deleted.** Every call site uses `sql.withTransaction`; a store's `migrate` and its transactional methods now require only `SqlClient`.

The mechanical migration is a separate commit from the behaviour change:

1. `effect: bump to 4.0.0-rc.115` — entirely upstream renames
2. `sql-sqlite: remove the SqlTransaction service` — the actual change

Upstream changes commit 1 absorbs:

- `SchemaAST.Union`'s `mode` argument became a `UnionOptions` bag
- `SchemaTransformation.transformOrFail` / `SchemaGetter.transformOrFail` → `transformEffect`
- `onExcessProperty: 'preserve'` removed — `validateSchema` reads only issues, so `'ignore'` preserves behaviour
- CLI and `Config` constructors capitalized (`Flag.string` → `Flag.String`, `Options.integer` → `Options.Int`, `Options.choice` → `Options.Literals`, `Prompt.text` → `Prompt.String`, `Config.redacted` → `Config.Redacted`, …) — ~90 files. `Options.Boolean` also became required by default, so every boolean flag now states `withDefault(false)`.
- AI service types collapsed (`LanguageModel.Service` → `LanguageModel.LanguageModel`, `Chat.Service` → `Chat.Chat`); `streamText` now yields tool parameters in `opaque` mode, so `AiParser.parseResponse` is parameterized over the mode rather than pinned to `decoded`
- `RpcSerialization.msgPack` removed — the `RpcPort` protocols take their format from the ambient `RpcSerialization`; the `layer*` helpers supply `layerSchemaBinary()`, so `client-protocol` is untouched. **The wire format changes**, but both ends of a port build from the same version.
- `@effect/platform-node` declares an optional `redis` peer, added to the existing `ignoreMissing` list

### JSON Schema changes ECHO absorbs

- A struct serializes open by default; ECHO's wire contract states it closed, so generation passes `onExcessProperty: 'error'`.
- `Schema.Number` projects to `number | "NaN" | ±"Infinity"` for JSON and that projection drops the node's checks, so `multipleOf`, `minimum` and the rest vanished. ECHO never stores a non-finite number, so they are re-applied to the encoded node.
- An empty struct serializes as `not: { type: 'null' }` rather than `anyOf: [object, array]`, which read back as `Unknown` instead of a struct.
- A check's keywords are now merged into the node instead of nested under `allOf`, with only colliding ones nested — the upstream characterisation test says so.

### `RpcRouter` under schema-binary

`main`'s new `RpcRouter` needed three fixes to work on this serialization:

- `RpcServer.Protocol` gained `supportsNotifications` and `codecFor`; the router forwards both from the transport's protocol.
- An unknown request tag is the one response v4 builds **structurally** rather than through the rpc's exit codec, and a serialization whose codec returns bytes has no wire form for that shape — it arrived as a connection-level `Defect` instead of a failure of the one request. The port protocol re-encodes it.
- The router's test harness builds the client protocol into the test's scope: the protocol forks its receive loop into the layer's scope, which a per-effect `Effect.provide` closes as soon as the client is constructed.

## Test plan

- `moon exec :build` across the whole monorepo — **clean, 0 errors**
- `echo` — 609/609
- `mcp-server` — 54/54
- `rpc`, `client`, `client-services`, `sql-sqlite` — all green (`client-services` 191/191)
- `feed` — 47/47; `index-core` — 58/58; `keyring` — 15/15
- `pipeline-discord` `answer-questions` — 4/4 (its `LanguageModel` fake was rebuilt on `LanguageModel.make`, which removes two casts)
- Model fixtures regenerated for the rc.115 tool-parameter shape; replay mode 15/15
- Formatted with `pnpm format`

Cross-repo, verified by local linking rather than a publish round trip:

- dxos/edge#1081 builds clean and passes **183/183** db-service tests against this branch
- Its workerd suite proves real Durable Object transactions through the official API: commit, rollback, and refusal when the client is built without `storage`

## Landing order

dxos/edge#1081 is **red until this lands and edge's catalog is bumped** — edge CI resolves `@dxos/*` from published pins whose types still require `SqlTransaction`. That is expected, not a defect in either diff.

https://claude.ai/code/session_017RJzE15GAn2BfBAXDyqER8

---
_Generated by [Claude Code](https://claude.ai/code/session_017RJzE15GAn2BfBAXDyqER8)_

---
🚀 Composer preview: https://pr-13114-composer-dev.dxos.workers.dev


dmaretskyi:
## CI status: two failures fixed, the rest are one upstream decision

**Fixed and pushed**

- `build` / `boot-budget` — `composer-app:bundle` died on `[MISSING_EXPORT] "isDeepStrictEqual"`. rc.115's `testing/TestSchema` imports it from `node:util`, which the browser bundle maps to `@dxos/node-std`; the `util/` polyfill doesn't provide it. Exported it via `deep-equal` in strict mode (verified against `NaN`, `-0`/`0`, Map/Set/TypedArray, circular refs, prototype sensitivity). `composer-app:bundle` now exits 0 locally.
- `Check / check` — two genuine rule violations of mine: two changesets where the repo allows one, and a `major` bump which is not permitted pre-1.0. Consolidated to one entry, `minor`, breaks described in the body. Both `check-changesets.mjs` and `pkg-lint.mjs` pass locally.

**Blocked — all remaining failures trace to effect rc.113+, none to the SQL change**

| failing | cause |
|---|---|
| ~17 `@dxos/echo` (JSON-Schema round-trip, format, projection, 5 snapshots) | empty-struct serialization changed **exactly at rc.113** |
| 3 `model-fixture` tool tests | tool parameters now default to `opaque`, not `decoded` |
| 6 `sql-sqlite` OPFS browser tests | `SqlClient` internals gained a `borrower` / transaction-control path |
| 15 shard-2 (MCP, chat-prompt, mcp serve/watch) | not yet attributed |

Measured, not inferred — `Schema.toJsonSchemaDocument(Schema.Struct({}))`:

```
rc.108 – rc.112   {"anyOf":[{"type":"object"},{"type":"array"}]}
rc.113 – rc.115   {"not":{"type":"null"}}
```

`toEffectSchema` sees no `type: object` and falls through to `Unknown`.

The OPFS failures are **not** this PR's SQL change: that path never referenced `SqlTransaction`, and the entire diff in `@dxos/sql-sqlite` is the four `SqlTransaction` files.

**Why I've stopped here rather than pushing fixes**

The 5 snapshot diffs are a change in the JSON Schema we *emit* — which is persisted in ECHO and sent to the LLM tool surface. Deciding how the bridge should represent an open/empty struct under the new encoding is a wire-contract decision, not a mechanical fix, and it isn't mine to make unilaterally inside a PR about Durable Object transactions.

**`rc.109` already contains the DO transaction fix this PR needs; `rc.113` is where the breakage starts.** So `rc.112` delivers the entire goal — layer deleted, real DO transactions — with none of the above, and also drops the ~90-file CLI/`Config` rename wave, the `@dxos/rpc` wire-format change and the `isDeepStrictEqual` shim.

Awaiting a call on rc.112 vs rc.115 before going further. Keeping the PR watched.

---

dmaretskyi:
## Correction: the rc.113 change is a runtime regression, not just test fallout

My earlier comment framed this as a wire-contract decision plus some snapshots to re-bless. That understated it. Reproduced locally:

```
$ dx mcp serve   # (piped a real initialize handshake)
SchemaError: Missing key
  at ["type"]
    at effect/dist/unstable/ai/McpServer.js:1122
    at packages/devtools/cli/src/commands/mcp/serve.ts:80
```

`McpServer.js:1122` decodes **every tool's `inputSchema`** against `ToolJsonSchema`, which requires a `type` key:

```js
const inputSchema = yield* Schema.decodeUnknownEffect(ToolJsonSchema)(Tool.getJsonSchema(tool)).pipe(Effect.orDie);
```

Since rc.113, an open/empty struct serializes with **no `type`**:

```
rc.108 – rc.112   {"anyOf":[{"type":"object"},{"type":"array"}]}
rc.113 – rc.115   {"not":{"type":"null"}}
```

So the decode dies and the MCP server never answers the handshake.

**Consequences on this branch**

- `dx mcp serve` is **broken at runtime** — no MCP client can connect (`serve.test.ts` and `watch.test.ts` time out awaiting `initialize`, with empty stderr).
- The same path publishes tool schemas to the assistant, which is why `functions/services.test.ts > assistant plugin operation inputs deserialize to struct-like schemas` fails.

This is a shipped command breaking, not a snapshot nit.

**Also corrected:** I previously said the CLI failures in shard-2 were cleared. They went 15 → 7, not to zero; I had misread an incomplete local run. The 7 are the MCP/schema family above. The boolean-flag fix did land as intended (shard-4 went 8 → 6, and the `Terminal`/`CliPanel` story failures are gone).

**Recommendation**

`rc.109` already carries the Durable Object transaction fix this PR needs. `rc.112` is the last version before the serializer change. Retargeting to rc.112 delivers the entire goal — `SqlTransaction` deleted, real DO transactions — and avoids breaking `dx mcp serve`, changing the schemas the assistant receives, invalidating the committed model fixtures, and ~90 files of CLI/`Config` renames.

Staying on rc.115 means either fixing the tool-schema contract against effect's `ToolJsonSchema` or shipping a broken `dx mcp serve` — which belongs in its own PR, not one whose purpose is removing a SQL transaction layer.

---

packages/common/node-std/src/util.js
```diff
@@ -4,7 +4,13 @@
...
 // NOTE: The `util` module depends on process global.
 import './globals.js';
+import deepEqual from 'deep-equal';
 import util from 'util/';
 
 export { callbackify, debuglog, format, inspect, promisify, stripVTControlCharacters } from 'util/';
+
+// The `util/` polyfill omits this one; `deep-equal` in strict mode is the `deepStrictEqual`
+// comparison Node's version performs.
+export const isDeepStrictEqual = (actual, expected) => deepEqual(actual, expected, { strict: true });
```
mykola-vrmchk, line 14: `isDeepStrictEqual` is the one member of this module that is missing from the default export.

Every other member is re-exported *from* `'util/'`, so it is also a property of the object `export default util` returns (line 16). This one is not — the polyfill lacking it is precisely why the shim exists — so the two import styles diverge:

```js
import { isDeepStrictEqual } from 'node:util'; // works
import * as util from 'node:util';
util.isDeepStrictEqual(a, b);                  // works (named exports)

import util from 'node:util';
util.isDeepStrictEqual(a, b);                  // undefined -> TypeError
```

The default-import path fails at runtime rather than at bundle time, because the `[MISSING_EXPORT]` check that caught the original `composer-app:bundle` break only sees named imports. Nothing in the repo takes that path today, so this is latent rather than broken — but it is a silent failure mode in a compat shim, which is the wrong place for one.

One line, keeping the polyfill's identity and its non-enumerable members (a spread or a bare `Object.assign` into a fresh object would drop those):

```js
export default Object.assign(util, { isDeepStrictEqual });
```

Moot if this retargets to rc.112, since the shim only exists to satisfy rc.115's `testing/TestSchema`.

---

mykola-vrmchk (commented):
One latent issue in the `node-std` shim (inline). Everything else in the diff holds up — including the `onExcessProperty: 'preserve'` -> `'ignore'` swap in `validate.ts`, which I checked against the source: the success branch never reads the decoded value, so it is genuinely inert.

---

packages/common/node-std/src/util.js
dmaretskyi: Correct, and fixed in b537e0ee with your suggested line.

Verified the divergence before changing anything — on the polyfill object, `callbackify` is a `function` while `isDeepStrictEqual` is `undefined`, exactly as you describe. And after `Object.assign(util, { isDeepStrictEqual })`: the member resolves and compares correctly (`{a:1}` vs `{a:1}` → true, vs `{a:2}` → false), `dflt === util` holds so identity is preserved, and `callbackify` / `inspect` / `promisify` remain intact and functional.

Good catch on the failure mode specifically — a silent runtime `TypeError` in a compat shim is worse than the bundle-time break it was added to fix, and the `MISSING_EXPORT` check would never have caught it.

`composer-app:bundle` re-verified clean after the change.

And agreed it's moot under rc.112 — the shim only exists for rc.115's `testing/TestSchema`. I've argued for retargeting in the PR thread; the `dx mcp serve` handshake is broken on rc.115 for a related reason (tool `inputSchema` losing its `type` key), so this shim would go away with it.

---

dmaretskyi:
## Correction: rc.115 stays, and everything above is fixed

Two earlier comments recommended retargeting to rc.112. That recommendation is withdrawn — the branch is on rc.115 and each break is now understood and fixed on this branch. Correcting the record item by item:

**`dx mcp serve` — was genuinely broken, now fixed.** My diagnosis of the mechanism was right and the fix was small. rc.115's `McpServer` decodes every tool's `inputSchema` against `ToolJsonSchema`, which requires `type`; v4 emits an empty struct as `{not: {type: 'null'}}`, which states none, so one parameterless tool (`whoami`) killed the server at startup. `ToolJsonSchema` does not exist in rc.108, which is why this only appears here. The repo already had the idiom for it — `Schema.Record(Schema.String, Schema.Never)`, used by the operation projection — so it moved to `@dxos/mcp-server` as `NoParameters` and `whoami` uses it. `serve.test.ts` and `watch.test.ts` pass.

A second, unrelated fault sat behind it: an unresolved optional field on a `queryOperations` row survives encoding as an explicit `undefined`, and MCP's structured content must be JSON, which has no such value — so **every** `queryOperations` call failed as an internal error. The views omit those keys now.

**ECHO JSON Schema — three real changes, all absorbed.** Structs serialize open by default (generation passes `onExcessProperty: 'error'`); `Schema.Number`'s JSON projection drops the node's checks, losing `multipleOf`/`minimum`/etc.; an empty struct reads back as `Unknown`. echo is 609/609.

**OPFS browser tests — a protocol change, not a flake.** rc.115's worker-backed client pairs each row with its own column names (`rows[0][index]`), where our worker sent one shared array — every column came back `undefined`. 34/34 now.

**`RpcRouter`** (new on main) needed `supportsNotifications` + `codecFor` forwarded to its child protocols, and the unknown-tag exit re-encoded: v4 builds that one response structurally rather than through the rpc's exit codec, which schema-binary cannot put on the wire, so a bad tag arrived as a connection-level `Defect` instead of a failure of the one request.

**Also corrected:** an earlier claim of mine that `dx mcp serve` was already broken on `main` was wrong — it was not, and rc.115 broke it.

Model fixtures were regenerated for the new tool schemas; replay is green. The remaining CI signal is what this push produces.

---

PR #13120, opened 2026-09-15
plugins: exclude app-shell-only modules from the generated capability barrels

dmaretskyi:
## Why

Four modules require capabilities that only an app shell provides, but declare no `environments`, so they ship in every generated `#capabilities` barrel — including the workerd one — and fail the dependency graph on every Cloudflare Worker host:

| Module | Requires |
| --- | --- |
| `plugin-space` `NavigationTargetResolver` | `ClientCapabilities.Client` |
| `plugin-inbox` `NavigationTargetResolver` | `ClientCapabilities.Client` |
| `plugin-magazine` `AppGraphBuilder` | `AttentionCapabilities.ViewState` |
| `plugin-assistant` `CompanionChatProvisioner` | `OperationInvoker`, `AppGraph`, `AtomRegistry`, … |

All four raise `MissingProviderError`. On EDGE's `operation-service` this is **192 `plugin failed to activate` warnings per hour**, matched 1:1 by `dependency graph error`.

The noise was the symptom; the real cost is that it made a genuine regression invisible. Because these failures were unconditional, a handler set that silently stopped merging into the registry was indistinguishable from them — `listOperations` would return a short list with `outcome: 'ok'` and nothing would look wrong.

## Change

`environments: []` on each of the four, excluding them from **every** generated barrel.

This is the mechanism the generator already provides, not a new seam:

- `plugin-cli/generate.ts:60` — `carries = member.environments === null || member.environments.includes(env)`. An omitted annotation reads as `null` and ships everywhere; an empty list ships nowhere.
- `core/plugin.ts:405` — `addModule` short-circuits on `undefined`, commented as existing precisely so headless barrels can stub excluded modules and let one canonical plugin entry serve every environment.

`plugin-space` already used this for `Dashboard` (`[]`); these four were simply missed.

The browser is unaffected: it resolves `default` → `capabilities/index.ts`, not a generated barrel. Generated barrels are gitignored (`.gitignore:272`) and produced by `gen-modules` at build time, so the four source edits, the magazine test and the changeset are the whole change.

### Node is in scope, deliberately

An earlier revision used `environments: ['node']`, keeping these active under Node and excluding them only from workerd. @wittjosiah's review called for `[]` instead, so Node is now covered too.

That has a consequence worth stating explicitly, because CI found it the first time `[]` was tried: vitest resolves `#capabilities` → `gen/node.ts`, and `createComposerTestApp` does provide the app-shell capabilities, so these modules really were active under Node. `plugin-magazine` asserted exactly that:

```
MagazinePlugin > modules activate on the expected events   ❌ Check / test (shard=2)
```

That assertion encoded the old semantics, so it moves with them rather than blocking the change:

```ts
expect(harness.manager.getActive()).toEqual(
  expect.arrayContaining([moduleId('schema'), moduleId('OperationHandler')]),
);
expect(harness.manager.getActive()).not.toContain(moduleId('AppGraphBuilder'));
```

The negative assertion is deliberate — dropping the id from the expected list alone would leave the new behaviour unasserted. Magazine was the only one of the four with an activation test, so the other three change without a test witnessing it either way.

## Verification

`moon run plugin-{space,inbox,magazine,assistant}:gen-modules` — exit 0. All four are now stubbed in **both** barrels:

| plugin / module | `gen/node.ts` | `gen/workerd.ts` |
| --- | --- | --- |
| space `NavigationTargetResolver` | `= undefined` | `= undefined` |
| inbox `NavigationTargetResolver` | `= undefined` | `= undefined` |
| magazine `AppGraphBuilder` | `= undefined` | `= undefined` |
| assistant `CompanionChatProvisioner` | `= undefined` | `= undefined` |

The previously failing test passes:

```
$ moon run plugin-magazine:test
 Test Files  12 passed (12)
      Tests  63 passed (63)
```

## Context

Proposed by @wittjosiah in review of dxos/edge#1086, which worked around this host-side with a `pruneHostOnlyModules` step plus an allow-list of upstream module ids. That workaround is the worse fix — it rebuilds plugins by copying a fixed set of module fields (silently lossy as `PluginModule` grows) and makes edge the owner of a list of ids nothing here is obliged to keep stable. He has since confirmed on that PR that it is "good to land after its incorporated changes from #13120", so this is the blocking dependency: once it lands and edge bumps its catalog, #1086 drops the prune and the allow-list entirely and keeps only its activation-failure check, whose expected failure count then becomes zero.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved plugin activation in non-browser hosts by excluding capabilities that require unavailable app-shell services, preventing dependency errors.
  - Worker-hosted applications now avoid failures caused by unsupported plugin dependencies.
  - Node-based composer and test-harness scenarios continue to support the required plugin functionality.
  - Browser behavior remains unchanged.

---
🚀 Composer preview: https://pr-13120-composer-dev.dxos.workers.dev


dmaretskyi:
CI caught a real problem with the first revision, and the fix narrows the change.

**What failed:** `Check / test (node-version=24.11.1, shard=2)` → `MagazinePlugin > modules activate on the expected events` (`plugin-magazine/src/plugin.test.ts`). This PR's own failure, in a plugin it changes — not a flake.

**Why.** I had described these modules as app-shell-only and excluded them with `environments: []`, which drops them from *every* generated barrel. But the `node` condition is not "headless": vitest resolves `#capabilities` → `gen/node.ts`, and `createComposerTestApp` genuinely does provide attention and the app-shell capabilities. So the first revision removed these modules from every node consumer as well as from workerd. `plugin-magazine` was the only one with a test asserting its module is active; the other three would have regressed silently.

**Fix:** `environments: ['node']` instead of `[]`. That excludes them from the workerd barrel — the actual problem on EDGE — and leaves node consumers byte-identical to before. Verified after regeneration:

| plugin / module | `gen/workerd.ts` | `gen/node.ts` |
| --- | --- | --- |
| space `NavigationTargetResolver` | `= undefined` | real export |
| inbox `NavigationTargetResolver` | `= undefined` | real export |
| magazine `AppGraphBuilder` | `= undefined` | real export |
| assistant `CompanionChatProvisioner` | `= undefined` | real export |

The browser is unaffected either way — it resolves `default` → `capabilities/index.ts`, not a generated barrel.

@wittjosiah this is the one place I've deviated from what you proposed: `['node']` rather than `[]`. `[]` is right if the intent is "no headless host should ever get these", but it takes the node test harness with it, and I'd rather not change node behaviour in a PR aimed at workerd. Say the word if you want `[]` plus an updated magazine test instead — that's a coherent alternative, it just asserts something different about what a node host is.


---

dmaretskyi:
`Check / check` is red on `8cd0fb7`, and I don't believe it belongs to this PR. Recording what I established rather than just asserting it.

**The failure**

```
Unused files (1)
 ELIFECYCLE  Command failed with exit code 1.
```

Step 27, `pnpm run knip`. The first Knip invocation exits 1, so the `&&` chain stops before the `--production --strict` pass.

**Why it isn't this PR's**

- This PR's entire diff is five files: four `plugin-*/src/capabilities/index.ts` and one changeset. Nothing under `packages/reflect/`.
- `mdl.terms.ts` was added on 2026-09-11 by #13053 and is tracked, not generated.
- The same check **passed** on this PR's two previous heads, `dfd06fc` and `a56db89`. The only thing that changed is the base: `main` advanced from `66cae5f` to `58b59d7`. Those two commits are a settings change and a CI-only change — neither touches `reflect/deus` either.
- Knip is not wrong about the file: `grep` finds **zero** importers of `mdl.terms` anywhere in `packages/`. It is genuinely unreferenced, and it is equally unreferenced on `main`.

That combination reads as a semantic conflict between two independently-green PRs — the last importer went away in one, the file stayed in another — surfacing on whichever PR merges `main` next. It is not something this PR introduced or can sensibly fix.

**What I could not reproduce, stated plainly**

`pnpm knip` exits 0 locally on both `origin/main` (`58b59d7`) and this branch (`8cd0fb7`), with no report output at all. So my local environment disagrees with CI, and I could not close that gap — most likely stale `dist/` or workspace state here masking it. I'm flagging that rather than claiming a clean local run proves anything, because on the evidence above CI is the one that's right.

I've re-run the job once (attempt #2) to confirm it reproduces. I have **not** touched `packages/reflect/deus` — deleting or Knip-excluding a file in another package to get a plugin PR green would be widening this PR on a guess.

**Proposed patch, for whoever owns it** — one of:

1. Delete `packages/reflect/deus/src/extension/gen/mdl.terms.ts` if it is obsolete, or
2. make it reachable from a configured entry point, or
3. add a Knip exclusion in `.config/knip.ts` if it is intentionally generated-but-unused.

Happy to push any of these if that's wanted here rather than in its own PR — say which and I'll do it.


---

dmaretskyi:
**Correction to the comment above — the re-run passed, so my diagnosis was wrong.**

Attempt #2 of the same job, on the same commit (`199c74c`, the `refs/pull/13120/merge` result), came back `conclusion=success`. `Check / check` is green and the suite is clean on `8cd0fb7`.

So this was **not** a semantic conflict between two landed PRs, as I claimed. That explanation predicted a deterministic failure; a clean re-run on an unchanged commit rules it out. What I got right was that the failure isn't this PR's and that nothing in `packages/reflect/deus` should be touched for it. What I got wrong was the mechanism — and I stated it with more confidence than a single un-reproduced failure supported.

It also resolves the local/CI discrepancy I flagged: `pnpm knip` exiting 0 locally on both `main` and this branch wasn't stale state masking a real problem. Those runs agreed with the truth, and the one red CI attempt was the outlier.

**The part still worth someone's attention** (not this PR's to fix, and no longer blocking anything): `knip`'s unused-file result for `packages/reflect/deus/src/extension/gen/mdl.terms.ts` is evidently not deterministic — same commit, two attempts, two different answers. `mdl.terms.ts` has zero importers repo-wide, so whether it is flagged appears to depend on whether `.config/knip.ts` manages to map that package's entry points, which the config's own comments note is sensitive to whether `dist/` exists. That makes `Check / check` capable of failing spuriously on any PR. Worth a look independently of this change; I'm not filing it here since it's unrelated to the plugin barrels.

No code change was made for any of this — the three fixes I proposed above should be disregarded unless the flakiness turns out to have a real unused file behind it.


---

dmaretskyi:
**Third update, and this one has a reproducible pattern behind it.** My previous comment called this a one-off flake and told you to disregard it. That was premature — it recurred.

`Check / check` has now failed and then passed on two independent commits, always in the same shape:

| `refs/pull/13120/merge` | Attempt #1 | Attempt #2 (same commit, no changes) |
| --- | --- | --- |
| `199c74c` | ❌ knip: unused `mdl.terms.ts` | ✅ pass |
| `b9afd28` | ❌ knip: unused `mdl.terms.ts` | ✅ pass |

Identical failure, identical file, both times cleared by a re-run of the same commit. That is not randomness — 2/2 in the same direction.

**Most likely mechanism: cold vs. warm build cache.**

`.config/knip.ts` derives each package's entry points from its `exports`/`imports`, and for packages that predate the `source` condition it maps those back from `./dist/...`. The config's own comment says it plainly:

> Packages that predate the `source` condition point straight at `dist`, so map those back through the layout the build mirrors — without it the real entry is never registered and **every file in the package reads as unused**.

On a cold moon cache `@dxos/deus` isn't built, that mapping doesn't resolve, and its only unreferenced-on-paper file surfaces as "unused". On the retry the cache is warm and it resolves. This also explains why `pnpm knip` exits 0 for me locally on both `main` and this branch — my `dist/` is warm.

Two caveats I'd rather state than paper over: `mdl.terms.ts` genuinely has **zero importers** repo-wide, so knip isn't inventing anything — it's the entry-point mapping, not the file, that's environment-dependent. And I have not proven the cache link directly; I've shown the failure is deterministic-by-attempt-number and that the config documents exactly this failure mode. Someone who knows `deus` should confirm.

**Why this matters beyond this PR:** any PR whose `Check / check` lands on a cold cache can fail this way, and the only signal is a re-run fixing it. It is worth fixing at the source — most cleanly by giving `@dxos/deus` a `source` condition so its entry mapping never depends on `dist/` existing.

**Status here:** green on `e47007b`. I have made no change for this and have not touched `packages/reflect/deus` — the fix belongs in that package or in the knip config, not in a plugin-barrels PR.


---

packages/plugins/plugin-assistant/src/capabilities/index.ts
```diff
@@ -65,9 +65,12 @@ export const SubjectContext = Capability.lazyModule(
 export const SkillDefinition = AppCapability.skillDefinition(() => import('./skill-definition.ts'), {
   provides: [RoutineCapabilities.AgentDelegationStrategy],
 });
+// Excluded from the workerd barrel: provisions companion chats against deck planks and attention
+// state, which no worker host holds. Node keeps it — the composer test harness runs there.
```
wittjosiah: this comments are not necessary

packages/plugins/plugin-assistant/src/capabilities/index.ts
```diff
@@ -65,9 +65,12 @@ export const SubjectContext = Capability.lazyModule(
...
 export const CompanionChatProvisioner = Capability.lazyModule(
   'CompanionChatProvisioner',
   {
+    environments: ['node'],
```
wittjosiah:
```suggestion
    environments: [],
```

packages/plugins/plugin-inbox/src/capabilities/index.ts
```diff
@@ -56,9 +56,13 @@ export const AutomationTemplates = Capability.lazyModule(
   { provides: [RoutineCapabilities.Template], activatesOn: RoutineEvents.Start },
   () => import('./automation-templates.ts'),
 );
+// Excluded from the workerd barrel: a worker host has no navigation and cannot provide `Client`,
+// so this would fail the dependency graph rather than be skipped. Node keeps it — the composer
+// test harness runs there and does provide the app-shell capabilities.
 export const NavigationTargetResolver = AppCapability.navigationResolver(
   () => import('./navigation-target-resolver.ts'),
   {
+    environments: ['node'],
```
wittjosiah:
```suggestion
    environments: [],
```

packages/plugins/plugin-magazine/src/capabilities/index.ts
```diff
@@ -16,7 +16,10 @@ import { translations } from '#translations';
 // eslint-disable-next-line import/no-relative-packages
 import pluginSpec from '../../PLUGIN.mdl?raw';
 
+// Excluded from the workerd barrel: builds app-graph nodes from attention view state, which no
+// worker host holds. Node keeps it — the composer test harness runs there and provides attention.
 export const AppGraphBuilder = AppCapability.appGraphBuilder(() => import('./app-graph-builder.ts'), {
+  environments: ['node'],
```
wittjosiah:
```suggestion
  environments: [],
```

packages/plugins/plugin-space/src/capabilities/index.ts
```diff
@@ -55,9 +55,13 @@ export const IdentityCreated = Capability.lazyModule(
 );
 export { NavigationHandler } from './navigation-handler/index.ts';
 export type { NavigationHandlerOptions } from './navigation-handler/index.ts';
+// Excluded from the workerd barrel: a worker host has no navigation and cannot provide `Client`,
+// so this would fail the dependency graph rather than be skipped. Node keeps it — the composer
+// test harness runs there and does provide the app-shell capabilities.
 export const NavigationTargetResolver = AppCapability.navigationResolver(
   () => import('./navigation-target-resolver.ts'),
   {
+    environments: ['node'],
```
wittjosiah:
```suggestion
    environments: [],
```

wittjosiah (changes requested):

dmaretskyi:
Applied, in `40e6f19` — `environments: []` on all four and the comments removed.

One thing came with it that your suggestions didn't mention, so flagging it rather than burying it: `[]` is exactly what CI rejected on the first revision of this PR. `plugin-magazine`'s `modules activate on the expected events` asserts `AppGraphBuilder` is active, and vitest resolves `#capabilities` → `gen/node.ts`, so stubbing it in the node barrel breaks that test.

I took that as the test encoding the old semantics rather than a reason not to make the change, so it moved with them:

```ts
expect(harness.manager.getActive()).toEqual(
  expect.arrayContaining([moduleId('schema'), moduleId('OperationHandler')]),
);
expect(harness.manager.getActive()).not.toContain(moduleId('AppGraphBuilder'));
```

The negative assertion is deliberate — dropping the id from the list alone would leave the new behaviour unasserted, and this was the only one of the four with a test, so the other three were already going to change silently.

Verified after regeneration — all four now stubbed in **both** barrels, where before only workerd stubbed them:

| plugin / module | `gen/node.ts` | `gen/workerd.ts` |
| --- | --- | --- |
| space `NavigationTargetResolver` | `= undefined` | `= undefined` |
| inbox `NavigationTargetResolver` | `= undefined` | `= undefined` |
| magazine `AppGraphBuilder` | `= undefined` | `= undefined` |
| assistant `CompanionChatProvisioner` | `= undefined` | `= undefined` |

```
$ moon run plugin-magazine:test
 Test Files  12 passed (12)
      Tests  63 passed (63)
```

The changeset is rewritten too — it claimed "Browser and Node hosts are unaffected", which this change makes false. It now says Node is affected and names the three modules a Node host stops activating; the browser genuinely is unaffected, since it resolves `capabilities/index.ts` rather than a generated barrel.

Worth a second's thought before merge: this now changes Node behaviour for the composer test harness, which is the thing I'd avoided. If any consumer relies on these activating under Node, this is where it surfaces.

---

dmaretskyi:
@coderabbitai All three findings share one premise, and it's wrong: they assume the generated environment union is built only from explicit `environments` annotations. It isn't — it also includes each maker family's default.

Measured on this branch (`40e6f19`), after `rm -rf`ing every `gen/` dir and regenerating with `--force`, so nothing below is a stale artifact:

```
$ ls packages/plugins/plugin-{space,inbox,magazine,assistant}/src/capabilities/gen/
space      node.ts workerd.ts
inbox      node.ts workerd.ts
magazine   node.ts workerd.ts
assistant  node.ts workerd.ts
```

The workerd barrels exist, and every module in question is stubbed in them:

```
plugin-space/gen/workerd.ts:     export const NavigationTargetResolver = undefined;
plugin-inbox/gen/workerd.ts:     export const NavigationTargetResolver = undefined;
plugin-magazine/gen/workerd.ts:  export const AppGraphBuilder = undefined;
plugin-assistant/gen/workerd.ts: export const CompanionChatProvisioner = undefined;
```

And in the node barrels too, which is the point of `[]`:

```
plugin-space/gen/node.ts:        export const NavigationTargetResolver = undefined;
plugin-inbox/gen/node.ts:        export const NavigationTargetResolver = undefined;
plugin-magazine/gen/node.ts:     export const AppGraphBuilder = undefined;
plugin-assistant/gen/node.ts:    export const CompanionChatProvisioner = undefined;
```

The manifest maps the condition, so there is no fallback to the canonical barrel:

```json
"#capabilities": {
  "source": {
    "workerd": "./src/capabilities/gen/workerd.ts",
    "node": "./src/capabilities/gen/node.ts",
    "default": "./src/capabilities/index.ts"
  },
  ...
}
```

**Why the premise fails.** `generate.ts:46` unions `member.environments`, but `barrel.ts:104` resolves that per member as *"its own `environments` literal if it declares one, else the default its maker family declares"* (`maker-defaults.ts`). So members that declare nothing still contribute their family's conditions to the union. `plugin-magazine` is the clean demonstration: its **only** `environments` annotation is the `[]` this PR adds, and it still generates both barrels — which cannot happen under the union-from-annotations-only model the findings assume.

Taking them individually:

- **plugin-inbox** — "the plugin environment union contains only `node`, so the advertised workerd condition falls back to `src/capabilities/index.ts`". The union contains workerd, `gen/workerd.ts` exists, and `NavigationTargetResolver` is `undefined` in it.
- **plugin-space** — "the environment union is currently `['node']`". It isn't; that was the *previous* revision of this PR. On `40e6f19` it's `[]`, and the workerd barrel is generated with the stub.
- **plugin-magazine** — "when the environment union is empty, the generator removes `src/capabilities/gen`". The union isn't empty, for the family-defaults reason above. The suggested fix (annotating `OperationHandler` with `['node', 'workerd']`) would be a no-op restating a family default.

No change made. Marking the three resolved.

One note on method, since it nearly caught me: I first reproduced your result and was about to concede. I'd regenerated while checked out on a *different* branch (`claude/module-loader-span-attribute`), which carries none of these annotations, so `AppGraphBuilder` came back as a real export exactly as predicted. The barrels are gitignored build output, so a stale or wrong-branch `gen/` directory is indistinguishable from a real one by inspection. Worth stating plainly given the 🟠 High merge-risk rating rests on it.

---

wittjosiah (approved):

PR #13136, opened 2026-09-16
feed: read sync state locally instead of round-tripping to edge

wittjosiah:
## Problem

`getSyncState` issued a `peekPull` request to EDGE on every read. During startup, before the socket is up, that blocked for the full 30 s RPC timeout while Composer waited on it. Found while reading Composer logs for a startup stall.

`FeedStore.onNewBlocks` was also unscoped, so a write in any space recomputed the sync-state subscription for every space.

## Change

Sync-state reads are now purely local. `pull` records how many blocks the server still holds as a side effect of a real pull, and `getSyncState` reads that estimate. Change events carry a space id and subscribers filter on it. The `getSyncState` plumbing through `EchoHost` and `FeedSyncer` is deleted. `peekPull` is deprecated rather than removed.

The tradeoff is a cold-start window: before the first pull completes, a space reports a backlog of 0 rather than an unknown one.

## Adjacent bugs fixed

Found while reading the same paths, each with a regression test verified by reverting the fix:

- **`FeedSyncer.#runSerialized` never held its mutex.** A `using` guard around an un-awaited `return run()` releases as soon as the function returns, not when the promise settles, so every caller ran concurrently. Serialization now lives in `SyncClient`, per space and namespace, so `pull` and `push` cannot interleave their rewrites of the same sync state and positions.
- **A pulled batch landed in two commits.** A failure between them left blocks on disk with no record of the position they advanced to, and no notification, so the indexer never saw them. `applyPulledBatch` writes both in one transaction and notifies after it commits.
- **Sealing could run inside a transaction.** `#append` sealed payloads before opening its transaction on purpose, but nothing enforced it, and `applyPulledBatch` wrapping it would have broken that. The split into `#sealForInsert` and `#insertSealed` makes it structural: sealing and validation run before any caller opens a transaction, and `#insertSealed` only writes. This matters because both SQLite clients gate every statement behind one connection permit, so WebCrypto round trips inside a transaction would block the index engine and AutomergeHost.
- **`append` notified too late.** It announced after the cursor-token write that follows its commit, so a failure there left committed blocks unannounced.
- **Teardown zeroed the backlog estimate.** Fiber interruption and the `ContextDisposedError` a closing context raises both mean the read never happened, so they keep what the last pull recorded. A defect is a real failure and still clears it.
- **`setPosition` and `resetSyncState` changed block positions** — part of what `subscribeFeed` serves — but announced only a sync-state change. The unscoped event used to cover this by accident; scoping it per space exposed the gap.

## Testing

`moon run feed:test` 55 passed, `moon run echo-host:test` 363 passed, `moon run client-services:test -- feed-syncer.test.ts` 4 passed. Each new test was verified by reverting its fix and confirming it fails.

## Summary by CodeRabbit

* **New Features**
  * Sync status now reflects remote backlog, pulled blocks, and positioned uploads.
  * Feed updates are scoped to the affected space and namespace.
  * Pull and push operations for the same feed are coordinated to prevent conflicts.

* **Bug Fixes**
  * Sync failures clear stale backlog indicators.
  * Feed changes and sync-state updates are reported consistently after successful operations.

* **Removed**
  * Removed the public feed sync-state handler and related fallback behavior.

---
🚀 Composer preview: https://pr-13136-composer-dev.dxos.workers.dev


packages/core/echo/feed/src/feed-store.ts
```diff
@@ -480,6 +511,7 @@ export class FeedStore {
           `;
         }),
       );
+      this.#emitBlocksChanged(opts.spaceId);
```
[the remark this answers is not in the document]
wittjosiah: Fixed in 8f19f00164. The estimate belonged to the server whose positions the reset just discarded, and with the notification now firing for block changes as well, it was reachable by subscribers. Routed the four uses of the map key through one `backlogKey` helper at the same time, so the delete cannot drift from the write and the read.

The existing `resetSyncState` test now seeds a backlog and asserts it reads 0 afterwards; removing the `delete` line fails it.

packages/core/echo/feed/src/feed-store.ts
```diff
@@ -78,9 +84,18 @@ export class FeedStore {
...
+
+  /**
+   * Emits when a space's sync backlog may have changed: blocks appended, positions assigned or
+   * reset, or a new remote backlog recorded.
    */
-  readonly onNewBlocks = new Event<void>();
+  readonly onSyncStateChanged = new Event<{ spaceId: string }>();
+
+  /** Blocks the sync server still holds past the last pull, keyed by `spaceId:feedNamespace`; not persisted. */
+  readonly #remoteBacklog = new Map<string, number>();
```
dmaretskyi: just wondering if this isn't better in feed-syncer

wittjosiah: It can't go in `FeedSyncer` — the dependency direction runs the wrong way. `client-services` → `echo-host` → `feed`, and the ownership matches: `EchoHost` constructs the `FeedStore`, `FeedSyncer` constructs the `SyncClient` and borrows the store from below, and the only reader is `LocalFeedServiceImpl` in echo-host. Putting the estimate in `FeedSyncer` puts it a layer above its reader, so echo-host would need a callback up into client-services — which is the `getSyncState` handler this PR deletes.

But you're right that it didn't belong where it was, and two reviewers flagged the same smell independently. Moved it to a `blocksToPull` column on `sync_state` in 4a6bd73c1d. That table is already the client-only pull-progress row for a space and namespace, `applyPulledBatch` already writes it in the same transaction, and the reader already reaches `FeedStore`. The map, both accessors and the key helper are gone, and the estimate now commits atomically with the batch it describes.

Two behavior changes came with it:

1. The estimate survives a restart, which closes the cold-start window where a space reported a drained namespace until the first pull landed.
2. A failed pull keeps the last estimate rather than zeroing it. A failure learns nothing about the server, and to a reader 0 means "caught up", not "unknown". That also deleted the predicate that had to sort teardown from real failures.

A pull that brings no blocks goes through `recordPullProgress`, which writes and notifies only when the row would actually change, so a caught-up namespace still polls without a write.

Net 89 insertions, 183 deletions. feed 54, echo-host 363, client-services feed-syncer 4.

dmaretskyi (approved):

PR #13143, opened 2026-09-16
app-framework: record the module id on the activation span

dmaretskyi:
## Why

`ModuleLoader.load` already spans every module activation — `#runActivation` is wrapped in `Effect.withSpan('ModuleLoader.load')`, and because `Effect.tapCause(() => Scope.close(...))` is piped *before* it, failures are recorded inside the span too.

What it doesn't record is **which module**. A cold start activates ~20 modules, so a trace carries that many identically-named spans and cannot say which one was slow or which one failed.

## Change

One line:

```ts
Effect.withSpan('ModuleLoader.load', { attributes: { 'dx.module.id': module.id } }),
```

`module.id` is already in scope in that pipe chain (the sibling long-activation warning uses it).

## Context

Requested by @wittjosiah in review of [dxos/edge#1084](https://github.com/dxos/edge/pull/1084), and split out here at the author's direction rather than folded into that PR.

EDGE's `operation-service` bridged app-framework's activation `PubSub` messages into detached OTEL spans purely to get per-module attribution. That bridge existed for two reasons: the framework's own activation spans were dropped (the effect ran on the default runtime, whose tracer discards spans), and even once traced they could not be told apart.

The first half is now fixed on the edge side — dxos/edge#1084 `ed1a521` deletes `traceModuleActivation` and `startDetachedSpan` (−76/+18) and runs activation through the OTEL tracer. **This PR is the second half.** Until it lands and edge bumps its catalog, a cold start there emits ~20 identically-named `ModuleLoader.load` spans; with it, the framework's own spans carry the attribution the bridge was reconstructing — for every app-framework consumer rather than one worker.

That attribution is what named `plugin.space` and `plugin.magazine` as the modules failing on every cold start, which is what [dxos/edge#1086](https://github.com/dxos/edge/pull/1086) acts on.

## Verification

```
$ moon run app-framework:build app-framework:lint
▮▮▮▮ app-framework:lint  (7s 736ms)
▮▮▮▮ app-framework:build (7s 739ms)
Tasks: 89 completed
```

Both exit 0; `oxfmt` clean on the changed file.

## Changeset

None. This adds a span attribute to internal instrumentation — no public API, type or behaviour change a consumer can observe, so it isn't changelog-relevant.


packages/sdk/app-framework/src/core/plugin-manager/module-loader.ts
```diff
@@ -388,7 +388,10 @@ export class ModuleLoader {
       return expanded;
     }).pipe(
       Effect.tapCause(() => Scope.close(scope, Exit.void)),
-      Effect.withSpan('ModuleLoader.load'),
+      // The module id is the whole diagnostic value of this span: a cold start activates ~20
+      // modules, so without it a trace carries that many identically-named spans and cannot say
+      // which module was slow or which one failed.
```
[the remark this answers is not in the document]
dmaretskyi: Checked this against the convention and I don't think it holds — leaving the comment as is.

The guideline is that a **comment** ends with a period, not that every wrapped `//` line does. This comment is a single sentence soft-wrapped across three lines, with the period on the last one. Its immediate neighbours in the same file are written the same way:

```ts
// One host yield per module keeps activation tasks under the long-task threshold;
// counted as wait, not run, by the split below.          // ← 342-343

// Wait/run split: `module:<id>` spans the whole pipeline, so scheduling delay (in-flight
// providers + requires resolution) is indistinguishable from work without these sub-measures.   // ← 345-346

// Cap activation so a single misbehaving module can't hold the
// event chain open. On timeout the failure is recorded against
// the plugin and surfaced as `PluginTimeoutError`.        // ← 356-358
```

Applying the rule per-line would require breaking each of those into sentence fragments too.

The proposed replacement also moves the wrong way on a second guideline: `CLAUDE.md` asks for the constraint in **one load-bearing clause, not a multi-sentence essay**, and the suggestion splits one sentence into two.

Flagging one thing in your own review info, since it may explain this: *"CodeRabbit could not clone the repository, so clone-backed analysis was skipped and this review may be incomplete."* Without the clone the surrounding comments in this file weren't visible, which is exactly the context that settles the convention here.


---

wittjosiah: drop the comment

wittjosiah (approved):

packages/sdk/app-framework/src/core/plugin-manager/module-loader.ts
dmaretskyi: Dropped in `be444fe`, and merged as `12aaf06`.

---

PR #13158, opened 2026-09-16
echo: bump automerge to 3.5.0

mykola-vrmchk:
Bumps the `@automerge/automerge` catalog pin from 3.3.2 to **3.5.0**.

## What moves

`@automerge/automerge-repo@2.6.0-subduction.48` pins `@automerge/automerge` to `3.3.2` **exactly**, so the catalog bump alone would leave a second copy in the tree. The existing `overrides` entry — `'@automerge/automerge': 'catalog:'` — is what replaces it, so the single catalog edit moves every consumer and **the lockfile references 3.5.0 only** (no 3.3.2 remains anywhere in it).

`automerge-repo` and `automerge-subduction` are unchanged, so their patches are untouched; this repo carries no patch against `@automerge/automerge` itself.

## Lockfile scope

Regenerating the lockfile makes pnpm re-resolve the whole graph, which normalises ~200 lines of unrelated peer-dependency suffixes (`zod`, `yaml`, `@modelcontextprotocol/sdk`) that have nothing to do with this bump. Those were reverted, so the lockfile diff is **exactly the 39 automerge lines**: the `overrides` pin, the 17 importer `specifier`/`version` pairs, the `packages:` key + integrity, and the two `snapshots:` entries.

The trimmed lockfile is self-consistent: `pnpm install --frozen-lockfile` (what CI runs) succeeds and leaves it byte-identical, and `@automerge/automerge@3.5.0` is what `echo-doc` resolves in `node_modules`.

## Verification

- `pnpm install --frozen-lockfile` clean, lockfile unchanged afterwards.
- `moon run echo-doc:test` — **18/18 passing** (`Doc.test.ts`, `AddOnlySet.test.ts`, `store-adapter.test.ts`), i.e. the package that uses `@automerge/automerge` directly.
- Nothing in this repo reaches into automerge proxy internals (no `_is_proxy` / `_clear_cache` / cache access), so upstream's internal `proxies.js` cache change (plain object → `Map`) is not observable here.
- `oxfmt --check` clean on the added changeset.

https://claude.ai/code/session_01L2Dw3nDQXizgTbiEn3A15j

## Summary by CodeRabbit

* **Chores**
  * Prepared a patch release for the Echo package.
  * Updated the underlying synchronization technology to improve compatibility and maintain current platform support.
  * No changes were made to the public API or user-facing functionality.

---
🚀 Composer preview: https://pr-13158-composer-dev.dxos.workers.dev


mykola-vrmchk:
`Model Fixture / model-fixture` is red here, and it is not this PR's.

It fails in `assistant-toolkit:test` → `update-tasks.model.test.ts` on a memoized-conversation mismatch in the checklist task-ref rendering:

```
- "(ref: echo://<memoized-dynamic-0>/<memoized-dynamic-3>)"
+ "(ref: <memoized-dynamic-3>(echo://<memoized-dynamic-0>/<memoized-dynamic-4>))"
```

That is the format #13154 ("assistant: render task refs as mnemonic-labelled links in the checklist") introduced without re-recording the fixture. It reproduces on `main` — `a34b57ff`, `026406979d1c` and `772ee78e` all fail the same check — and this branch is based on `a34b57ff`, so it inherits it. An automerge version pin cannot reach that assertion.

No fix PR exists yet to port, and re-recording the fixture needs live model calls, which is outside this bump. `Model Fixture` is not in `main`'s required checks, so it does not block; the required `Check / test` shards and `Check / check` are green or still running.

---

PR #13159, opened 2026-09-16
tooling: enforce unknownInEffectCatch and lazyPromiseInEffectSync

wittjosiah:
Two of the seven Effect diagnostics turned off in [#12732](https://github.com/dxos/dxos/pull/12732) are now clear, so they come back on and the TODO that tracked them goes away. A full `moon exec --on-failure continue :build` reports zero diagnostics with both enforcing.

Picking these two was not obvious from the old TODO, which grouped three rules as "the cheapest to retire". Two of them were; the third was not, for a reason worth recording rather than re-deriving next time.

## `unknownInEffectCatch` — six sites

Every one was the same shape: the caught value passed straight through, so the error channel stayed `unknown`.

```ts
Effect.tryPromise({
  try: () => getEdgeClient().forceRunCronTrigger(...),
  catch: (error) => error,
})
```

Five of the six drop the `catch` entirely. `Effect.tryPromise` with a bare thunk supplies `Cause.UnknownError`, which is typed, carries the original as its cause, and is already the form used elsewhere in these same files (`EdgeTriggerManager.ts:79` a few lines above the one I changed). In all five the failure is swallowed immediately downstream — `Effect.catch`, `Effect.matchEffect`, `Effect.catchCause` — so nothing observes the change but the log line, which now prints a proper `Error` instead of a bare rejection value.

The sixth is `errors.test.ts`, where the thunk throws deliberately and the assertions read the resulting stack. Dropping the `catch` would have wrapped it, so that one narrows with `instanceof` and returns the thrown `Error` untouched.

## `lazyPromiseInEffectSync` — one site, and a real bug

```ts
yield* Effect.addFinalizer(() => Effect.sync(() => manager.destroy()));
```

`OnboardingManager.destroy` is `async` and awaits `this._ctx.dispose()`. Wrapped in `Effect.sync`, the promise was dropped on the floor, so scope teardown never waited for the context to dispose. It now goes through `Effect.tryPromise` with a warning on failure, the shape `plugin-native` already uses to dispose its Ollama runtime.

This is the case for the rule existing: the diagnostic found an un-awaited teardown that typechecked, passed tests, and nothing else in the toolchain could see.

## `layerMergeAllWithDependencies` stays off

Both of its findings are in `createToolkit`, on the `Layer.mergeAll` of three handler layers. `OpaqueToolkit`'s `any` makes one layer's output look like another's requirement — the analyzer reports `Handler<any>` and `unknown`, neither of which any layer there actually requires. False positives, but clearing them means typing `OpaqueToolkit`, which is its own change. Its TODO now names the cause instead of a count.

## Notes for a reviewer

**Finding them required a full build, twice.** The `effect-language-service diagnostics` CLI drives the JavaScript TypeScript API, which no longer exists here — `node_modules/typescript/lib` is the Go binary shim — so it dies on startup and `moon exec :build` is the only way to see these. That matters for the remaining rules: a failing package skips its dependents, so the first run showed one finding in `@dxos/effect` and hid everything downstream of it. Four of the six `unknownInEffectCatch` sites only appeared after the earlier ones were fixed. Anyone taking on the next rule should expect to iterate rather than get one complete list.

**Both rules default to `warning`, which gates the build**, so the entries are deleted rather than set to a severity — the TODO's own instruction ("clear the findings, flip the rule, delete the line").

Verified: full `moon exec :build` green with both rules enforcing, and the test suites of all six packages touched (`effect`, `edge-compute`, `plugin-doctor`, `plugin-support`, `plugin-assistant`, `plugin-onboarding`) pass.

Remaining, in the order they look cheapest: `multipleEffectProvide` (82), `globalErrorInEffectCatch` (92), `globalErrorInEffectFailure` (248), plus `missingEffectContext` and `layerMergeAllWithDependencies`, both blocked on `any` in the capability and toolkit types rather than on a backlog of call sites.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved onboarding shutdown by waiting for cleanup to complete and recording failures.
  - Improved error handling for scheduled trigger execution, log retrieval, and support log flushing.
  - Improved reliability when validating Anthropic and DeepSeek API keys, including clearer handling of authorization and network failures.
  - Malformed diagnostic log entries continue to be skipped without interrupting log processing.

- **Quality**
  - Strengthened checks for Effect-related error-handling issues during development.

---
🚀 Composer preview: https://pr-13159-composer-dev.dxos.workers.dev


wittjosiah:
`Model Fixture / model-fixture` is red here and it is not this PR's.

Four planning-skill tests fail on `No memoized conversation found for the given prompt`. The diff in the fixture is the checklist's task-reference format — saved `(ref: echo://<id>/<id>; started)` against generated `(ref: <label>(echo://<id>/<id>))` — so something changed how task refs render in that prompt. This PR touches six files across `effect`, `edge-compute`, `plugin-assistant`, `plugin-doctor`, `plugin-onboarding` and `plugin-support`, none of which reach `assistant-toolkit` or any LLM prompt.

The same failure, on the same four tests, is also red on [#13162](https://github.com/dxos/dxos/pull/13162), whose diff shares nothing with this one beyond being stacked on it.

Not porting a fix: regenerating those fixtures means rerunning with `DX_UPDATE_MODEL_FIXTURES=1` against a live model for a prompt change that belongs to whoever made it, which would widen this PR well past its scope. `.depot/workflows/model-fixture.yml` says the workflow is deliberately outside the required-checks set so fixture drift never blocks merge, and `Check` is green here.

---

tsconfig.base.json
```diff
@@ -73,11 +73,10 @@
...
           "missingEffectContext": "off",
-          // TODO(wittjosiah): Re-enable once cleared. 7 findings across the three,
-          // so these are the cheapest to retire.
-          "unknownInEffectCatch": "off",
-          "layerMergeAllWithDependencies": "off",
-          "lazyPromiseInEffectSync": "off"
+          // TODO(wittjosiah): Re-enable once cleared. Both findings are in
+          // `createToolkit`, where `OpaqueToolkit`'s `any` makes one handler layer's
+          // output look like another's requirement. Typing OpaqueToolkit clears it.
+          "layerMergeAllWithDependencies": "off"
```
dmaretskyi, line 79: is this using tsgo? I saw a post recently about effect + tsgo integration

wittjosiah: Yes — the integration is already wired up here, which is why these settings gate the build rather than just the editor.

```
$ pnpm exec tsc --version
Version 7.0.2+effect-tsgo.0.36.5
```

The `prepare` script runs `effect-tsgo patch --typescript --oxlint`. `node_modules/typescript/bin/tsc` is a shim that execs a native binary chosen by `lib/getExePath.js`, and the patch swaps that binary from `@typescript/typescript-linux-x64` to `@effect/tsgo-linux-x64`. So the Effect diagnostics run inside the compiler, not as a language-service plugin — a finding fails `moon run <pkg>:build` for everyone, not just whoever has the extension installed. (`.oxlintrc.json` points its `$schema` at `@effect/tsgo/oxlint-schema.json`, the `--oxlint` half of the same patch.)

Two consequences worth knowing, both of which cost me time on this stack:

**The diagnostics only appear on a real emit.** `tsc --noEmit` reports nothing, and a stale `dist/types/tsconfig.tsbuildinfo` does the same — worse, tsgo then reports live `@effect-diagnostics-next-line` directives as having no effect, so you can talk yourself into deleting a suppression that is doing real work. The measurement that matches the build is `rm -rf dist/types && pnpm exec tsc`, which is what `DxDeclarationsPlugin` runs (`vite.base.config.ts:323`).

**`effect-language-service diagnostics` (the CLI) does not work in this repo.** It drives the JavaScript TypeScript API, and `node_modules/typescript/lib` no longer contains a compiler — just the ~25-line shim above. It dies on startup, so enumerating findings means a full `moon exec :build`. That also means a failing package hides its dependents' findings: the first run of this stack showed one finding in `@dxos/effect` and nothing downstream, and four more `unknownInEffectCatch` sites only surfaced after the earlier ones were fixed. Expect to iterate rather than get one complete list.

If the post you saw covers something beyond the compiler patch and the oxlint preset, link it and I'll take a look — there may be a piece we are not using.

---

PR #13162, opened 2026-09-16
tooling: enforce multipleEffectProvide

wittjosiah:
Stacked on [#13159](https://github.com/dxos/dxos/pull/13159) — review that one first; this PR's diff is only its own commit.

Third of the seven Effect diagnostics turned off in [#12732](https://github.com/dxos/dxos/pull/12732). Every chained `Effect.provide` is collapsed into one, and the rule is on.

**122 findings across 62 files, not the 82 the TODO recorded.** That count came from a `moon exec :build` where a failing package hides its dependents, so most of the repo was never reached. Running `tsc --noEmit` per package instead sidesteps the dependency graph entirely and sweeps all 332 packages in 2m42s, which is how the real number surfaced. Worth knowing for the rules still outstanding.

## Why `Layer.provideMerge` and not `Layer.mergeAll`

The rule ships a quickfix, and it emits `Effect.provide(Layer.mergeAll(A, B))`. I used that first. It is wrong here, and the way it failed is the interesting part of this PR.

`mergeAll` treats the layers as peers. A provide chain does not: in `pipe(Effect.provide(A), Effect.provide(B))`, B is provided to the already-A-provided effect, so B feeds A's own requirements as well as the effect's. The two agree only when A consumes nothing from B.

Sixteen files failed to compile under `mergeAll`, which is the honest outcome — the requirement simply goes unsatisfied. The other failure is the one worth flagging: in `compute-runtime`, two tracer tests went red while still typechecking.

```ts
Effect.provide(TestLayer),
Effect.provide(Layer.succeed(Tracer.Tracer, makeRecordingTracer(recordedSpans))),
```

Under `mergeAll`, `TestLayer` is no longer constructed with the recording tracer in scope, so the spans land on the default tracer and `recordedSpans` comes back empty. Nothing in the toolchain catches this: `Tracer` is a default service, so it never appears in `TestLayer`'s requirement type, which means `layerMergeAllWithDependencies` — the rule that exists precisely to police `mergeAll` — reports zero findings on that file. I checked.

So "it typechecks" is not a safe guard for this rewrite, and neither is the sibling rule. `Layer.provideMerge(A, B)` is the faithful translation of the chain in every case: B feeds A, both stay available to the effect. That is what all 121 sites use.

A follow-up could demote genuinely independent pairs to `mergeAll` for readability, but it needs a way to prove independence that doesn't exist today.

## The one chain that stays

`react-ui-terminal`'s `Terminal` provides against a still-generic `R`:

```ts
layer: Layer.Layer<Exclude<R, XtermContext.Provided>, never, never>;
```

The checker can discharge `Exclude<R, XtermContext.Provided>` one provide at a time but not in a combined form — both `mergeAll` and `provideMerge` leave the requirement unsolved. It keeps its chain under a suppression naming the reason.

## Mechanics

The rewrite was an AST codemod against `@typescript/typescript6`, anchored on the diagnostic's own reported locations so nothing was rewritten that the rule did not report. `Layer` value imports were added to 37 files (one upgraded from `import type`). One thing that cost a cycle and may be worth knowing: parsing `.ts` files as `ScriptKind.TSX` silently misreads `<T>(…)` generic arrows, which drops the pipes inside them — five sites went missing until I keyed the script kind off the extension.

Verified: full `moon exec :build` green with the rule enforcing, and tests pass in all 29 packages touched (2,151 tests). `compute-runtime` is 218 passed — the two tests `mergeAll` broke.

Remaining after this: `globalErrorInEffectCatch`, `globalErrorInEffectFailure`, then `missingEffectContext` and `layerMergeAllWithDependencies`, both blocked on `any` in the capability and toolkit types rather than a call-site backlog.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime dependency composition across HTTP clients, integrations, database-backed operations, credential handling, and workflow execution.
  * Preserved access to combined services in cases where sequential configuration could cause one dependency to override another.

* **Refactor**
  * Consolidated layered service configuration for more consistent behavior across the application and integrations.

* **Documentation**
  * Documented an intentional exception where separate service configuration remains necessary.

* **Chores**
  * Restored the default enforcement behavior for Effect layer diagnostics.

---
🚀 Composer preview: https://pr-13162-composer-dev.dxos.workers.dev


wittjosiah:
`Model Fixture / model-fixture` is red here and it is not this PR's — same four planning-skill tests, same `No memoized conversation found for the given prompt`, as on the parent [#13159](https://github.com/dxos/dxos/pull/13159#issuecomment-5701505907), where I wrote up the details. Two diffs with nothing in common failing identically is the tell.

Nothing in this PR reaches `assistant-toolkit` or an LLM prompt; it collapses `Effect.provide` chains. `.depot/workflows/model-fixture.yml` keeps the workflow outside the required-checks set precisely so fixture drift never blocks merge.

---

PR #13172, opened 2026-09-16
tooling: enforce globalErrorInEffectFailure

wittjosiah:
Fourth in the stack, on [#13167](https://github.com/dxos/dxos/pull/13167) → [#13162](https://github.com/dxos/dxos/pull/13162) → [#13159](https://github.com/dxos/dxos/pull/13159). Review those first; this PR's diff is its own commit.

The last of the three rules from [#12732](https://github.com/dxos/dxos/pull/12732) that was a backlog rather than a type-system problem. **200 findings across 104 files** — not the 248 the TODO recorded, because the earlier PRs in this stack cleared some on the way past: a catch that starts returning a tagged error usually fixes the failure channel it feeds.

## 167 × `Effect.fail(new Error('…'))`

Each gets a tagged error for its area. Where the package already had one that fit, it is reused, and the CLI is the case worth looking at:

```ts
Effect.catch((error) => Effect.fail(new Error(formatAdminError(error))))
```

`AdminApiError` and `HubApiError` already existed — `formatAdminError` opens by checking `error instanceof AdminApiError`. The error was being formatted to a string and then thrown away as an untyped `Error`, one line after the code proved it knew what type it was. Those twelve sites now keep it.

Where a package had nothing, it gets one error to fail with. Same terms as the previous PR: splitting per condition later is a small change, inventing the taxonomy up front is not.

## Nine more classes that were `Error` with extra steps

`NoBackendError`, `UploadNotFoundError` and friends — same finding as the nine in [#13167](https://github.com/dxos/dxos/pull/13167). An empty `extends Error` subclass is structurally the global `Error`, so `catchTag` could never have discriminated it.

## Two annotations, and why they are needed

Tagging both arms of a ternary makes the union real, and two call sites were inferring the channel from the first branch alone:

```ts
Effect.catchTag('BlobNotAvailableError', (error) =>
  error.context.reason === 'not-found'
    ? Effect.fail(new UploadNotFoundError(uploadId))   // ← channel inferred from here
    : Effect.fail(new NoBackendError()),               // ← and this one no longer fits
)
```

`create-from-upload.ts` and `edge-client/http-client.ts` now annotate the handler's return type. Worth knowing for anyone doing this again: the compile error names the second branch, but the fix belongs on the callback.

## Three files keep their plain `Error`

Each under a file-level suppression naming the reason, because in all three a tagged error would remove the thing being tested or shipped:

- `errors.test.ts` asserts a plain `Error`'s stack frames.
- `plugin-manager.test.ts` simulates third-party plugin code throwing an untagged one — that is what the manager has to survive.
- `forex-effect.ts` ships to users as a starting function and has no access to DXOS-internal error classes.

## `devDependencies`, this time

Six of the packages gaining `@dxos/errors` use it only from tests, so it goes in `devDependencies`. `knip` caught that locally before the push — which is the fix for the one thing that turned [#13167](https://github.com/dxos/dxos/pull/13167) red.

Verified: full `moon exec :build` green with the rule enforcing, `pnpm run knip` clean, 2,852 tests pass across the 41 packages touched. One test, `client`'s `dedicated-worker-client-services > connect client`, timed out at its 2s limit during the parallel sweep and passes on its own; it spawns a worker and `invitations/host.ts` (the only file this PR changes in that package) is not on its path.

After this, the two remaining rules are `missingEffectContext` and `layerMergeAllWithDependencies`, both blocked on `any` in the capability and toolkit generics rather than on a backlog. They are next in the stack.

---

---
🚀 Composer preview: https://pr-13172-composer-dev.dxos.workers.dev


wittjosiah:
`Model Fixture / model-fixture` is red here and it is not this PR's — the same four planning-skill tests and the same `No memoized conversation found for the given prompt` as on every PR below it in this stack, starting at [#13159](https://github.com/dxos/dxos/pull/13159#issuecomment-5701505907) where I wrote up the fixture diff. Five PRs with different diffs failing on the same four tests is the tell.

This PR does touch `assistant-toolkit`, so worth naming what: one `Effect.fail(new Error(…))` in the delegation skill's `delegate-tasks.ts` becomes a tagged error. The failures are in `skills/planning/`, and the fixture diff is the task-reference format inside the planning checklist prompt, which nothing here renders. `.depot/workflows/model-fixture.yml` keeps this workflow outside the required-checks set so fixture drift never blocks merge.

---

PR #13173, opened 2026-09-16
tooling: enforce layerMergeAllWithDependencies

wittjosiah:
Fifth in the stack, on [#13172](https://github.com/dxos/dxos/pull/13172). Two findings, both real, both fixed. No suppressions.

## The rule found a soundness hole

`OpaqueToolkit.layer` was declared `Layer<unknown, E, R>`. That `unknown` was meant as opacity: `make` knows the layer provides `Tool.HandlersFor<Tools>`, and the interface erases it so a consumer doesn't have to know which handlers it carries.

But `unknown` is the top type. `Effect.provide(toolkit.layer)` therefore discharged **every** requirement a consumer had, not just the handler ones — the checker would accept a program missing a service entirely. It is also what made three independent handler layers read as though each fed the next, which is the finding.

The erasure only ever needed to hide the tool *names*:

```ts
export type Handlers = Tool.Handler<any>;
```

Bounded and honest. The proof it was unsound is that tightening it broke `mcp-client`, where the old top type had been silently covering a real gap.

With the output typed, `mergeAll` → `provideMerge` (the rule's own suggestion) clears the second finding: same merged context, and it additionally satisfies a dependency if one ever appears.

## Threading the channels

`OpaqueToolkit.Any` declared `any` for the handler layer's `E` and `R`, so every toolkit passing through the constraint lost both, and `createToolkit` cast the widened result back to its declared type:

```ts
}) as Effect.Effect<OpaqueToolkit.OpaqueToolkit, AiToolNotFoundError, ToolResolverService | ToolExecutionService>;
```

The repo forbids casts that silence the checker, and an earlier revision of this PR shipped one with a TODO. `Any<E = any, R = any>` keeps the defaults that make it usable as a constraint (`T extends OpaqueToolkit.Any` is unchanged) while letting a caller name the channels. `createToolkit` is now generic in both and returns `OpaqueToolkit<never, E, R>`. The cast is gone, not explained.

The parameterization has to sit on `Any` rather than on `OpaqueToolkit<any, E, R>`: `OpaqueTools<TR>` appears in both variance positions of `Toolkit`/`WithHandler`, so `any` there is not assignable to `never`. `Any`'s `Toolkit.Toolkit<any>` is what sidesteps it.

## What this corrects

Two earlier revisions of this description were wrong and are worth naming, since the reasoning is the point of the stack:

1. "Two findings, one file, no code change" — the rule was right; the `unknown` was a hole.
2. "Threading the handler type back through `OpaqueToolkit<TR, E, R>` defeats the abstraction and touches every consumer. Not worth it." Threading `E` and `R` touches five call sites, defeats nothing, and is what removed the cast.

## Verification

`moon run :lint :build :test-types`, 761 tasks, exit 0. `ai` 127 passed, `assistant` 81 passed, `errors` 11 passed.

---

## Summary by CodeRabbit

- **Improvements**
  - Improved type support for AI toolkits, including custom error and environment requirements.
  - Toolkit creation now preserves more precise type information across configured toolkits and handlers.
  - Handler layer composition has been refined for more reliable toolkit setup.

- **Bug Fixes**
  - Toolkit layers now report missing service requirements instead of allowing them to pass type checking.

- **Developer Experience**
  - Layer dependency diagnostics are now surfaced instead of being suppressed, helping identify configuration issues earlier.

---
🚀 Composer preview: https://pr-13173-composer-dev.dxos.workers.dev


wittjosiah:
`Model Fixture / model-fixture` is red here and it is not this PR's — the same four planning-skill tests as on every PR below it in this stack, detailed on [#13159](https://github.com/dxos/dxos/pull/13159#issuecomment-5701505907).

Unusually easy to rule out on this one: the entire diff is a comment block, a `skip-file` directive and one deleted line of `tsconfig.base.json`. No executable code changes at all, so nothing here can move a model fixture. `.depot/workflows/model-fixture.yml` keeps the workflow outside the required-checks set for exactly this reason.

On the changeset reminder above: no changeset. The only change to `@dxos/assistant` is the comment and the directive — nothing a consumer can observe, which is the case the reminder says to ignore.

---

PR #13184, opened 2026-09-17
evals: let a preview MCP eval provision its own identity

dmaretskyi:
Follow-up to #13183, which wired the nightly deployed-MCP job as a **token-mode** run against `preview` — needing two hand-minted secrets and scoring only `operations-discovered` + `tool-latency`.

That was based on a wrong premise, stated in `McpTarget.ts`'s own docstring: that only `dev`'s test-account hatch admits a throwaway identity. It doesn't. Edge's `isTestAccountEnvironment` (`packages/sdk/edge-protocol/src/env.ts:43`) is:

```ts
isDevLikeEnvironment(environment) || environment === EdgeEnvironment.PREVIEW
```

So `POST /hub/account/login` with a `test+*@dxos.org` email creates or rebinds an account with no invitation code on preview too — closed only on staging and production (`hub-service` `account-manager.ts:325`).

## Change

- `EDGE_URLS` gains `main: 'https://preview.dxos.network'`.
- `edgeUrl()` keys off the target having an entry rather than `target === 'dev'`, so `prod` stays `undefined` even under `DX_EVAL_EDGE_URL` — the one target whose hatch really is closed.
- Docstrings corrected to cite `isTestAccountEnvironment` instead of the dev-only claim.
- The workflow's two secrets drop from required to optional: they are only for a closed-hatch target such as `prod`. A comment records the trap — setting either on `dev`/`preview` demotes the run to token mode and silently drops the write stages it exists to score.

Nothing in `claude-harness.ts` changed: `replicateToEdge` and `provisionToken` only ever ask for `edgeUrl(target)`, so the provisioning path was already target-agnostic.

## Effect

The nightly preview run becomes `provisioned` rather than `token`: it binds its own account, replicates its own space, mints its own grant, and every scorer runs — including the write stages and `upload-round-trip`, which is the coverage that would have caught dxos/edge#1101 against the surface users actually hit. No secrets required.

## Cost

Each run leaves a throwaway identity, a `test+mcp-eval-<key>@dxos.org` account and a space behind on preview. These accumulate; nothing reaps them today. Accepted deliberately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FQT9CrowxhfyxSwuKUYxTA)_

---
🚀 Composer preview: https://pr-13184-composer-dev.dxos.workers.dev

## Summary by CodeRabbit

- **Improvements**
  - Preview MCP evaluations now use a dedicated test account, replicate seeded spaces, mint access grants, and include write-operation grading.
  - Preview environments now support throwaway test identities alongside development environments.
  - Production evaluations continue to require token-based authentication and omit write stages when the hatch is closed.
  - Secret validation now focuses on the API key required by the selected evaluation variant, reducing unnecessary setup failures.


dmaretskyi:
Verified against the live deployments rather than only against the source, since the whole PR rests on preview's hatch actually being open.

Probe: a throwaway 32-byte identity key, its `did:halo:B…` DID derived exactly as `createDidFromIdentityKey` does, and a `test+mcp-eval-*@dxos.org` email with no invitation code — the same payload `bindTestAccount` sends — `POST`ed to `/hub/account/login`:

| EDGE | response |
| --- | --- |
| `https://preview.dxos.network` | `200 {"admitted":true}` |
| `https://dev.dxos.network` | `200 {"admitted":true}` |
| `https://dxos.network` | `200 {}` |

`admitted` is the only signal that the hatch ran rather than the waitlist flow, so this matches the `EDGE_URLS` table in the diff exactly: preview and dev provision, prod does not and keeps no entry.

`McpTarget.test.ts` passes locally (5/5), covering `main` → `provisioned` at that URL and `prod` staying `token`/`undefined` even under `DX_EVAL_EDGE_URL`.

---

PR #13187, opened 2026-09-17
feed: recover sync from a server that lost acknowledged positions

mykola-vrmchk:
## Summary

On 2026-09-15 a production `FeedSpace` lost rows it had acknowledged to a client, twice: 1443 blocks a client had
pushed earlier were missing when the first `serverToken` made it re-push everything, and 386 of the positions
assigned during that re-push (13750-14135) were gone again within minutes, while the overloaded router was
dropping the replies for them. Another client's pushes hold those positions today. The client of the incident was
spared a collision only because it never applied the dropped acknowledgements. A client that had applied them
holds positions the server has since given to other blocks; before this PR such a client retried the same head
batch forever (`setPosition` tripping its own `(feedPrivateId, position)` index) while every pull asked for a
position the server would never reach again. The token check never fires in this shape: the server keeps its
token.

This makes `@dxos/feed` treat the server's ordering as authoritative and recover on its own:

- **Adopt, evict, replay.** A position the server hands out is adopted; whatever local block held it is
  unpositioned again (and so re-pushed, de-duplicated by `(actorId, sequence)`), and a block the replica already
  held at another position counts the same way. A rolled-back server (re-issued or moved positions, a pushed block
  placed at or below the pull cursor, a `maxPosition` below the cursor, or a different block at the cursor)
  rewinds the cursor and replays the namespace once; the guard against restarting the replay on every page is
  armed only once the rewind is written, and released again when the server shrinks below the cursor or holds a
  different block at it, since within a replay the block at the cursor is the one the previous page wrote and
  either means the server changed again underneath it. `PositionConflictError` is gone: there is no longer a
  conflict to report.
- **`QueryResponse.maxPosition` and `QueryResponse.cursorBlock`** (new, optional) let a reader see a rollback
  without waiting for a collision, including one the server has already been written past by other clients, where
  neither the high-water mark nor the blocks above the cursor reveal it. An older server that reports neither is
  simply not checked.
- **First server token is verified, not acted on.** Adopting the first token over untokened progress used to
  wipe every namespace of every space and re-sync everything at once; hourly Durable Object overload spikes
  in the router line up exactly with those storms. Now the token is recorded and the cursor steps back one
  slot, so the next pull re-fetches the block the cursor was pulled from: the same block de-duplicates and
  progress is kept, a different block is a displacement and replays the namespace. A replacement whose blocks
  above the cursor collide with nothing the client holds is caught this way too.
- **Failing pulls back off** (250 ms doubling to 30 s) without taking the other spaces' polling schedule with
  them, the push back-off is decided once per run so a namespace that pushes fine no longer resets the delay a
  failing one is growing, **`Error` replies carry `requestId`** so a failed request fails at once instead of
  waiting out the 30 s timeout, and a reply carrying fewer positions than blocks fails the push with a typed
  `SyncAppendPositionMismatchError` instead of silently leaving the tail pending.
- **A space the server reports deleted is left out of sync until the next reconnect.** Both sync tasks run over
  every space and end with the slowest one, one page per namespace per run. The incident client tracks four
  spaces the server has deleted, whose frames the router dropped without a reply, so every run waited out a 30 s
  RPC timeout per namespace: it drained 453 blocks at one batch per minute, nine minutes for work the server
  accepted in milliseconds, and its re-push on 09-15 was cut off by a disconnect after 24 minutes at that pace.
  The companion PR has the `FeedSpace` answer such a frame with an `Error` coded `space_deleted`
  (`FeedProtocol.ErrorCode`, and an optional `Error.code` typed as a string so a code a client does not know yet
  decodes as an ordinary error). `SyncClient` fails such a request with a typed `SyncSpaceDeletedError`; the
  `FeedSyncer` then treats the namespace as done, asks no other namespace of that space, and skips the space
  until `onReconnected`, when the report is trusted afresh. Runs are back to milliseconds, and one page per run
  stays: the round trip between pages is where a blocking sync of another space gets in. A client that never
  receives the reply behaves as before.

A client holding positions the server lost heals on its first sync after this lands: it logs
`feed sync replica ordering is out of step with the serving store, replaying namespace` once, re-pulls the
namespace and re-pushes the lost blocks. No server or manual step is needed. Verified against the incident
client's real database replayed onto an emulated server rolled back to the shape the acknowledgements described:
403 stuck blocks positioned, every block present on the server at the server's position, one replay, under a
second, against both the deployed server and one carrying this change. The server-side hardening of the
acknowledgement path, the FeedSpace's coded reply for deleted spaces and the corrected incident audit are in the
companion PR [dxos/edge#1107](https://github.com/dxos/edge/pull/1107).

closes DX-1273

## Testing

- `moon run feed:test` (62), `moon run protocols:test` (33), `moon run client-services:test` (feed-syncer 12): new
  cases cover the high-water mark and the block at the cursor, adopting a re-issued position, counting a moved
  block, a reader whose cursor is beyond `maxPosition`, a server that drops acknowledged blocks and re-issues their
  positions, a server regrown to or past the cursor (with and without `cursorBlock`), verifying a first token
  against the same and a replaced server, a rewind that fails to persist, a replay that restarts when the block at
  the cursor changes underneath it, a reply short on positions, deterministic pull back-off gaps, a failing space
  that does not starve the others' full poll, a failing push next to a healthy namespace backing off instead of
  retrying at the minimum delay, an `Error` reply coded `space_deleted` failing the request with
  `SyncSpaceDeletedError`, and a space the server reports deleted staying out of both sync tasks until the
  connection is re-established while the live space keeps polling.
- `pnpm format`, `moon run feed:lint protocols:lint client-services:lint`, `moon run feed:build client-services:build`.

## Changeset

`.changeset/feed-sync-rolled-back-server.md` (`@dxos/feed`: patch).

https://claude.ai/code/session_011HsRynYQXq6Quoh1yBh1Nj

## Summary by CodeRabbit

- **Bug Fixes**
  - Feed synchronization now recovers from server rollbacks, lost acknowledged positions, and conflicting updates without becoming permanently stuck.
  - Displaced blocks are automatically replayed, helping clients converge with the server after changes from other clients.
  - Synchronization verifies server state more reliably and safely handles incomplete append responses.
  - Retries now use progressive backoff, while healthy spaces continue syncing normally.
  - Backlogged spaces can complete multiple synchronization pages in one run.
  - Failed requests return promptly with clearer request identification.
- **Documentation**
  - Updated feed recovery documentation to describe rollback detection and replay behavior.

---
🚀 Composer preview: https://pr-13187-composer-dev.dxos.workers.dev


packages/core/echo/feed/src/feed-store.ts
```diff
@@ -863,53 +968,50 @@ export class FeedStore {
...
+              continue;
+            }
+            displaced += yield* this.#evictSlot(request.spaceId, block.feedNamespace, block.position, {
+              ...block,
+              feedPrivateId,
+            });
+            yield* sql`
+              UPDATE blocks SET position = ${block.position}
+              WHERE feedPrivateId = ${feedPrivateId} AND actorId = ${block.actorId} AND sequence = ${block.sequence}
+                AND position IS NOT ${block.position}
```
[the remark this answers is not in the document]
mykola-vrmchk: Fixed in 5f5a33ba: `setPosition` now checks the target row exists before evicting the slot's holder, and skips the block otherwise. Covered by `setPosition leaves a slot alone when the block it was assigned to is gone`.

---

packages/core/echo/feed/src/sync-client.ts
```diff
@@ -370,6 +425,15 @@ export class SyncClient {
         rpcTag: 'AppendRequest',
       });
       const response = yield* self.#expectResponse<AppendResponse>(requestId, message, 'AppendResponse');
+      // Pairing a short reply with the batch would leave the tail unpositioned and the push looping
+      // without a diagnostic; a responder that assigns no positions is not a position authority.
+      if (response.positions.length !== unpositioned.blocks.length) {
+        return yield* Effect.fail(
+          new Error(
+            `AppendResponse carried ${response.positions.length} positions for ${unpositioned.blocks.length} blocks (spaceId=${opts.spaceId} feedNamespace=${opts.feedNamespace} requestId=${requestId}).`,
+          ),
```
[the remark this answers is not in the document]
mykola-vrmchk: Fixed in 5f5a33ba: the short reply now fails with `SyncAppendPositionMismatchError` (request id, space, namespace, block and position counts in its context), following the `SyncRpcTimeoutError` pattern in `errors.ts`.

---

mykola-vrmchk:
Re the outside-diff finding on `#reconcileServerToken` (first token retained a positive cursor): valid, and fixed in 5f5a33ba without going back to the full wipe. The first token over untokened progress now records the token, steps the cursor back one slot and discards the batch, so the next pull re-fetches the block the cursor was pulled from. The same block de-duplicates and progress is kept; a different block is a displacement and replays the namespace from the start. `sync.test.ts` covers both the same-server case and a replacement whose blocks above the cursor collide with nothing the client holds, which is the gap the finding describes.

---

packages/core/echo/feed/src/feed-store.ts
```diff
@@ -655,16 +754,18 @@ export class FeedStore {
...
+          // (actorId, sequence): adopt the server's position for it, even over a position the row
+          // already carries. The authority is the only source of positions, so a differing local
+          // one is stale -- kept, it would leave the replica's ordering out of step for good.
           // A position authority must keep DO NOTHING -- the assignment path below distinguishes a
           // fresh row from a duplicate by whether the insert wrote one.
           const onConflict = this.#options.assignPositions
             ? sql`ON CONFLICT(feedPrivateId, sequence, actorId) DO NOTHING`
             : sql`ON CONFLICT(feedPrivateId, sequence, actorId) DO UPDATE SET position = excluded.position
-                  WHERE blocks.position IS NULL AND excluded.position IS NOT NULL`;
+                  WHERE excluded.position IS NOT NULL AND blocks.position IS NOT excluded.position`;
```
mykola-vrmchk, line 824: A pulled block that *moves* position is applied but never counted, so a rollback the server has already regrown past goes unnoticed.

The upsert now moves a held block from its stale position to the server's, which is right, but only `#evictSlot` feeds `displaced`. A move into a slot nobody holds locally is invisible to the client. Within one batch the eviction of the lower slot catches it (batches are position-sorted); across pulls it does not:

1. Client A holds 0..c with cursor c. The server rolls back to r < c.
2. Something appends after the rollback (another client's new blocks, the server's own writer), then a healed client re-pushes the lost blocks. The server regrows past c before A's next pull.
3. A pulls from c: `maxPosition >= c`, so `clearPositionsAbove` clears nothing. Blocks the server re-issued into (c, m] that A already holds at p <= c are moved p -> q by this upsert, `displaced` stays 0, no replay.

A keeps the old ordering for (r, c] and never pulls what the server now holds there. Worse, A's own blocks that the server lost stay positioned locally and are never re-pushed.

The move can be counted in the statement that already runs once per pulled block, by letting `#evictSlot` also clear the holder's own row when it sits elsewhere:

```sql
UPDATE blocks SET position = NULL
WHERE feedPrivateId IN (SELECT feedPrivateId FROM feeds WHERE spaceId = ? AND feedNamespace = ?)
  AND (
    (position = ? AND NOT (feedPrivateId = ? AND actorId = ? AND sequence = ?))
    OR (feedPrivateId = ? AND actorId = ? AND sequence = ? AND position IS NOT NULL AND position != ?)
  )
RETURNING feedPrivateId, actorId, sequence
```

Rows matching the holder are `moved`, the rest `displaced`; the upsert then writes NULL -> position exactly as it does today, and `pull` treats `moved > 0` like `displaced > 0`. Same statement count as now.

That still leaves the sub-case where the re-pushed blocks land at or below c, which A never pulls at all; see the comment on the pull path for covering that.

---

packages/core/echo/feed/src/sync-client.ts
```diff
@@ -218,26 +223,86 @@ export class SyncClient {
...
+          self.#replaying.delete(self.#namespaceKey(opts));
+          yield* self.#replayNamespace(opts, response.serverToken, {
+            requestId,
+            reason: 'cursor beyond the server',
+            lastPulledPosition: basePosition,
+            maxPosition: response.maxPosition,
+            cleared,
+          });
+          return { done: false };
+        }
```
mykola-vrmchk: `maxPosition` only reveals a rollback while the server is still smaller than this client's cursor.

The server re-issues positions from its counter, so the first client to push after the rollback (including the healed one re-pushing what it lost) grows it back past every other client's cursor. From then on those clients see `maxPosition >= cursor`, blocks above the cursor that collide with nothing they hold, and no replay, while their own lost blocks stay positioned locally and are never re-pushed (trace in the `feed-store.ts` comment). In a multi-client space this is the common shape of a rollback rather than the edge case: only the client that happens to pull first sees the low mark.

The `'verify'` step already has the right primitive: re-fetch the block at the cursor and let de-duplication or displacement decide. Doing that on every pull closes the gap for lost blocks (a lost block shifts everything above it, so the block at the cursor changes unless another block lands in exactly its place), for one indexed row per query and no extra round trip. Two shapes:

- Server side, next to `maxPosition`: echo the identity of the block held at the requested `position` (`cursorBlock?: { actorId: string; sequence: number } | null`), and replay when the client holds a different block there. Optional, so an older server is simply not checked.
- Client only: always request from `lastPulledPosition - 1` and count the anchor as expected. No protocol change, but `done` can no longer be `blocks.length === 0` (the anchor is always returned), so the pull loop needs a "nothing above the cursor" condition instead.

With either in place, `maxPosition` becomes the fast path rather than the only signal.

---

packages/sdk/client-services/src/packlets/services/feed-syncer.ts
```diff
@@ -676,6 +691,12 @@ export class FeedSyncer extends Resource {
         { concurrency: this.#syncConcurrency },
       );
 
+      if (hadPullFailure) {
+        this.#schedulePollRetry();
+        return;
+      }
```
mykola-vrmchk: Returning here skips the full-poll bookkeeping, so one namespace that keeps failing stops every other space from being polled.

While `hadPullFailure` holds, the chain below never runs: `#lastFullPoll` is not refreshed and `#resetSpacesToPoll()` is not called, so `#spacesToPoll` shrinks to the failing space and healthy spaces are polled again only through a `FeedAdvanced` hint, or never against a server that does not push hints. Before this change the failing space spun, but the full poll still fired on its interval.

Keeping the chain and swapping only the immediate re-poll for the back-off keeps both behaviours:

```ts
const scheduleNext = hadPullFailure ? () => this.#schedulePollRetry() : () => this.#pollTask.schedule();
if (!hadPullFailure) {
  this.#pullFailureBackoffMs = DEFAULT_FAILURE_BACKOFF_MS;
}
if (this.#lastFullPoll == null || Date.now() - this.#lastFullPoll > this.#currentPollingInterval) {
  this.#resetSpacesToPoll();
  scheduleNext();
} else if (this.#spacesToPoll.size > 0) {
  scheduleNext();
} else {
  // unchanged
}
```

The new back-off test should still hold with this shape: the failing space is re-polled at 250 ms and 500 ms either way.

---

packages/core/echo/feed/src/feed-store.ts
```diff
@@ -863,53 +968,58 @@ export class FeedStore {
...
+            }
+            displaced += yield* this.#evictSlot(request.spaceId, block.feedNamespace, block.position, {
+              ...block,
+              feedPrivateId,
+            });
+            yield* sql`
+              UPDATE blocks SET position = ${block.position}
+              WHERE feedPrivateId = ${feedPrivateId} AND actorId = ${block.actorId} AND sequence = ${block.sequence}
+                AND position IS NOT ${block.position}
+            `;
```
mykola-vrmchk, line 1071: Four statements per block where the previous shape needed one on the common path.

The feed lookup, the existence check, the eviction and the update each round-trip per block, so a 50-block push reply now runs about 200 statements inside the transaction, against one `UPDATE ... RETURNING` before. Two reductions that keep the semantics:

- Resolve `feedPrivateId` once per distinct `feedId` in the request, as `append` already does; a push batch is almost always a single feed.
- Fold the existence check into the eviction: `AND EXISTS (SELECT 1 FROM blocks WHERE feedPrivateId = ? AND actorId = ? AND sequence = ?)`, so a reply for a deleted block still leaves the slot alone without its own round trip.

That leaves eviction + update per block. Not blocking, but this is the reply path of every push.

---

packages/sdk/client-services/src/packlets/services/feed-syncer.test.ts
```diff
@@ -504,6 +530,38 @@ describe('FeedSyncer', () => {
...
+    await syncer.open(new Context());
+    await new Promise((resolve) => setTimeout(resolve, 900));
+
+    // Retries at 250ms and 500ms after the failing rounds fit in the window; an immediate re-poll
+    // would have sent them by the dozen.
+    const tracePulls = sentMessages.filter(
+      (message) => message._tag === 'QueryRequest' && message.feedNamespace === FeedProtocol.WellKnownNamespaces.trace,
+    );
+    expect(tracePulls.length).toBeGreaterThanOrEqual(2);
+    expect(tracePulls.length).toBeLessThanOrEqual(5);
```
mykola-vrmchk: Optional: this measures wall-clock, so it is sensitive to a stalled runner.

Pulls land at about 0, 250 and 750 ms, so the lower bound trips if the process is paused for more than ~650 ms before the second round, which CI shards do under load. Asserting on the scheduled delays instead of the count would make it deterministic, e.g. recording the `delayMs` the syncer logs for `feed sync poll retry scheduled with backoff` (or exposing the back-off through the harness) and expecting `[250, 500]`.

---

mykola-vrmchk (commented):
Read the whole diff against `main` plus the surrounding `FeedStore`, `SyncClient` and `FeedSyncer` code, and the edge consumer (`FeedSpace`, `api.ts`). CI is green on 5f5a33b and both CodeRabbit threads are resolved. I could not run the suites in this sandbox (no toolchain installed), so everything below is from reading the code, with the traces spelled out inline.

The recovery for the incident shape (a single wedged client against a server that stayed small) looks right and is well covered by the new tests. What I would still change:

1. **Detection has a hole once the server regrows.** `maxPosition` only helps while the server is smaller than the cursor, and a block the upsert *moves* is applied but not counted, so any client whose next pull comes after the server has regrown past its cursor keeps the stale ordering and never re-pushes its lost blocks. In a multi-client space that is every client except the one that pulls first. Two inline comments: count the move in `#evictSlot` (same statement count, closes part of it), and check the block at the cursor on every pull, which the `'verify'` step already does once (closes it).
2. **The pull back-off starves the full poll.** The early return in `#pollTask` skips `#resetSpacesToPoll()` while any pull keeps failing, so one bad namespace stops every other space from being polled except via hints. This is a regression in the failure path relative to `main`; the inline suggestion keeps both the back-off and the schedule.
3. `setPosition` went from one statement per block to four; two of them fold away. Optional.
4. The back-off test is wall-clock based; optional tightening.

If you would rather land the incident fix first, 1 can be a follow-up, but 2 is a small change and worth taking here.

---

packages/core/echo/feed/src/feed-store.ts
mykola-vrmchk: Done in ba0713b1. `#evictSlot` now also clears the holder's own row when it sits at a different position and returns `{ displaced, moved }` from the same statement (`RETURNING feedPrivateId, actorId, sequence`, split in JS). `append` and `setPosition` sum both, and `pull` and `push` treat `moved > 0` exactly like `displaced > 0`.

Covered at the store level by `a replica counts a block handed a new position as moved even into an empty slot`, and end to end against a server that does not report `cursorBlock` by `replays when a server that does not report the cursor block moves a held block above the cursor`, which fails without the count.

One note on the within-batch case you mentioned: when the lower slot's eviction has already cleared the block, its later arrival at the new position reads as `displaced`, not `moved`, so the two existing tests assert `moved: 0` there. The replay condition is the sum, so nothing changes for the client.

---

packages/core/echo/feed/src/sync-client.ts
mykola-vrmchk: Done in ba0713b1, server side. `QueryResponse.cursorBlock` is a `BlockKey` (`feedId`, `actorId`, `sequence`), or `null` when the server holds nothing at the requested position; a position authority sets it for any `position >= 0`, read with one indexed lookup next to `maxPosition` in `FeedStore.query`. The client compares it against `getPositionHolders(cursor)` and replays on a different block, no block, or two local holders. An absent field (older server) is not checked, and neither is a reset or a cursor of -1.

I went with the echo rather than requesting from `lastPulledPosition - 1`: the anchor block's payload would ride every poll of every namespace, while the key is a few dozen bytes.

Covered by `a client the server has regrown to its cursor replays and re-pushes what the server lost`: the regrowth lands exactly at the cursor, so the mark is back at it, nothing sits above it, and only the block at the cursor differs. It fails without the check.

---

packages/sdk/client-services/src/packlets/services/feed-syncer.ts
mykola-vrmchk: Done in ba0713b1, in the shape you sketched: a failure only swaps the immediate re-poll for `#schedulePollRetry()`, and the full-poll chain still runs. `a pull that keeps failing does not stop the other spaces from being polled` tracks two spaces with one of them failing and checks that a block appended to the healthy one after the first round still arrives. It times out on the previous shape (verified by reverting locally), and the back-off test still passes.

---

packages/core/echo/feed/src/feed-store.ts
mykola-vrmchk: Done in ba0713b1. `setPosition` resolves `feedPrivateId` once per distinct feed in the request, and the existence check is folded into the eviction through `requireHolder` (`AND EXISTS (SELECT 1 FROM blocks AS existing WHERE ...)`), so the reply path is eviction + update per block plus one feed lookup per feed. `setPosition leaves a slot alone when the block it was assigned to is gone` still passes.

---

packages/sdk/client-services/src/packlets/services/feed-syncer.test.ts
mykola-vrmchk: Done in ba0713b1. The interceptor records `Date.now()` for each failed trace pull, the test waits for three, and asserts the two gaps are at least 250 ms and 500 ms (minus a 10 ms timer slack). Each gap is a lower bound on the delay scheduled before it, so a stalled runner can only widen them, while an immediate re-poll would put the retries milliseconds apart. No log capture needed.

---

mykola-vrmchk:
All four review points are addressed in ba0713b1:

1. **Detection after regrowth**, both inline suggestions: `#evictSlot` counts the holder's own move, and `QueryResponse.cursorBlock` lets the client check the block at its cursor on every pull (server-side echo; an older server is simply not checked). Two new sync tests reproduce the regrown-server shape with and without the field, and each fails with its detection disabled.
2. **Poll back-off** keeps the full-poll bookkeeping; a new test tracks two spaces with one failing.
3. **`setPosition`** is down to eviction + update per block, with one feed lookup per feed.
4. **Back-off test** asserts the scheduled gaps instead of a wall-clock count.

`moon run feed:test` 60/60, `feed-syncer.test.ts` 10/10, lint and format clean, the three packages typecheck. The edge side needs nothing new for this round: `FeedSpace` serves `cursorBlock` through `FeedStore.query` once the catalog pin is bumped, and until then clients fall back to the move count.

---

packages/core/echo/feed/src/feed-store.ts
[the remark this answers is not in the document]
mykola-vrmchk: Reopening the record on this one: 4778f81e removes the `requireHolder` guard again, together with its test, on purpose.

The guard only changes anything when the replica already holds a block at the position the server just assigned to the pushed block. A replica in step with the server cannot: the server assigns `max + 1`, which is above every position the replica holds. So a holder at that slot means the replica's ordering is stale (the server rolled back and re-issued the slot), and evicting it plus replaying the namespace is the right response whether or not the pushed block still exists locally. Left in place, the guard suppressed exactly that replay for the one case it covered. With the pushed block gone and no holder at the slot, the eviction is a no-op and the position `UPDATE` matches nothing, so nothing is lost either way.

---

packages/sdk/client-services/src/packlets/services/feed-syncer.ts
```diff
@@ -704,26 +740,33 @@ export class FeedSyncer extends Resource {
...
+                  .push(this._ctx, {
+                    spaceId,
+                    feedNamespace,
+                    limit: this.#messageBlocksLimit,
+                  })
+                  .pipe(
+                    Effect.tap(() =>
+                      Effect.sync(() => {
+                        this.#pushFailureBackoffMs = DEFAULT_FAILURE_BACKOFF_MS;
+                      }),
```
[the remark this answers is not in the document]
mykola-vrmchk: Fixed in 5cb7e0a7. The per-push reset is gone; each space returns `{ hadFailure, needsMore }` and the push task calls `#schedulePushRetry` once per run with the aggregate, the same shape the poll task uses. Covered by `backs off between push retries while a push keeps failing next to a healthy namespace`, which seeds one unpositioned trace block whose pushes are refused and asserts the retry gaps grow (250 ms, then 500 ms); against the previous shape the second gap measured about 250 ms and the test fails. The `seedBlocks` helper also moved below the suite.

---

mykola-vrmchk:
@coderabbitai review

The paused review's merge-risk note ("fix the shared backoff handling before merging") is addressed in 5cb7e0a7, and the branch has since gone back to one page per sync run (06e6d47d), so a fresh pass over the current head is worth having.

---

packages/core/echo/feed/src/sync-client.ts
```diff
@@ -218,26 +229,83 @@ export class SyncClient {
...
+          reason,
+          lastPulledPosition: basePosition,
+          maxPosition: response.maxPosition,
+          displaced,
+          moved,
+          cleared,
+          cursorBlock: response.cursorBlock,
+        }))
+      ) {
+        return { done: false };
```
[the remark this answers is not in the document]
mykola-vrmchk: Fixed in c52b7f19: the guard is released for `cursorBlockDiffers` as it already was for `beyondServer`. Within a replay the block at the cursor is the one the previous page wrote, so a mismatch there is never the replay's own doing and always a second rollback regrown past the cursor. Covered by `a replay restarts when the block at the cursor changes underneath it`: a replay is under way, the third page reports a different block at the cursor with the high-water mark back above it, and the cursor must rewind to -1 instead of advancing; against the previous shape it advanced to 2.

---

PR #13203, opened 2026-09-17
echo-host: re-drive denied rounds on a scope change instead of re-announcing the peer

wittjosiah:
The HALO device-join e2e tests fail about one run in ten: the new device joins the identity, but a space's root document never arrives within 60s. That failure was previously put down to the production edge. It is a client-side replication stall between the two devices.

- **Re-drive denied rounds on a scope change instead of re-announcing the peer.** Each space has its own teleport connection between two devices, and `MeshEchoReplicator` multiplexes them under one repo peer. When a second space connected, or a device was authorized for another space, `EchoNetworkAdapter` emitted `peer-disconnected` then `peer-candidate` for a peer whose connection was still open. Under Subduction that bound a second connection to the same peer while the first stayed registered, which strands the rounds pending on the first until the 60s round timeout (see the subduction skill's sync-reliability §3). It also reset the peer's collection sync. Subduction authorizes every round, so under Subduction the host now only schedules `shareConfigChanged()`, which re-drives the rounds denied before the wider scope. The classical sync network keeps the re-announce, which is how it re-evaluates its share policy.

## Evidence

Debug traces from failing CI runs (mesh, replicator and collection-sync logs forwarded into the worker) show the new device receiving the host's advertisement of the root document, then "Connection auth scope changed" twice within 100ms as the second space's connection opened, then no further traffic for that document. The existing root-document retry in `DataSpace` fired at 20s and 40s and changed nothing.

HALO spec only, repeated five times per job, retries 0, chromium, against the production edge:
- Without the fix: 4 failures in 45 runs.
- With the fix: 0 failures in 90 runs (two batches of three CI runs).

The `echo-network-adapter.test.ts` cases cover both routes of a scope change and run in CI. There is no in-memory repro of the stranded round: in-memory rounds finish before the scope change lands, so the CI runs above are the discriminating evidence. The echo-host suite and the client-services spaces and identity suites pass.

With #13209 on main, the scope-change kick shares the share-policy throttle with `onConnectionOpen`, so it runs at the end of that window rather than at once: about 1 s for a freshly joined device, up to the 30 s cap with many documents loaded. It stays throttled on purpose. A device reconnecting to its peer fires one scope change per shared space, and resetting the window on each would bring back the fan-out storm #13209 bounds.

## Impact

Real users: a device that joins an identity, or a peer that gains access to a second space, could wait out a 60s round timeout before a space's documents arrived.

Part of DX-1264. PR 1 of 7, split out of #13013 (replaces #13177, which GitHub closed when the stack was reordered); the workarounds, mitigations and diagnostics stay on #13013 for separate evaluation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization when a peer’s access to a space changes.
  - Previously denied synchronization requests are now retried without creating a duplicate connection, preventing requests from remaining stalled until timeout.
  - Preserved reliable peer reconnection behavior for scenarios that require re-announcing a peer.

- **Documentation**
  - Added guidance describing the improved reliability behavior during authentication-scope changes.

---
🚀 Composer preview: https://pr-13203-composer-dev.dxos.workers.dev


mykola-vrmchk:
Notes from landing #13209 (now on main), which touches the same machinery:

- **The scope-change re-drive now shares a throttle.** #13209 rate-limits `_sharePolicyChangedTask`: at most one fan-out per `clamp(loadedDocs × 15 ms, 1 s, 30 s)`. A call inside the window parks one trailing kick; kicks are coalesced, never dropped. So the new `onConnectionAuthScopeChanged` kick shares that budget with `onConnectionOpen` and the collection-sync kicks. In the sequence from your trace, the second space's connection opens and the scope changes about 100 ms later. The open's kick takes the window, and the scope-change re-drive runs at its end. That is about 1 s for a freshly joined device with few loaded documents, but up to about 12 s at 800 loaded documents (the cap is 30 s). It is still well under the 60 s round timeout, but it's worth re-running your HALO batch on top of main. If the delay shows up, a scope change is an explicit policy change, so it could reset `_sharePolicyKickNextAllowedAt` instead of waiting out the window.
- #13209 also skips the kick scheduled from `_handleCollectionSync` when every unconverged document is a diverged one already resynced at the same heads. That gate doesn't apply to your direct `schedule()`.
- **A pre-existing flake in this area, caused by neither PR.** `automerge-host-subduction.test.ts` › "deny→allow flip auto-recovers via AutomergeHost machinery (no manual kick)" hangs in about 5% of local runs: 1 of 20 on main, 2 of 20 with #13209. One host's heal round reports `succeeded` while the other host still lacks its commit, and nothing re-drives it. Your scope-change kick is the right lever when a real scope change triggers the flip; the kick-less path the test exercises stays flaky.
- The new test in `automerge-host-subduction.test.ts` is inside `describe.skipIf(process.env.CI)`, so CI won't run it. You note it passes without the fix anyway. The `echo-network-adapter.test.ts` cases do run in CI.

_🤖 Written with [Claude Code](https://claude.com/claude-code)_

mykola-vrmchk (approved):

wittjosiah:
Re https://github.com/dxos/dxos/pull/13203#issuecomment-5732082006

Thanks. Main (with #13209) is merged into this PR and forward through the stack.

- **Throttle.** Keeping the scope-change kick inside the window. A device that reconnects to its peer fires one scope change per shared space, so resetting `_sharePolicyKickNextAllowedAt` on each would bring back the fan-out burst #13209 bounds. For a freshly joined device the window is about 1 s, well inside the round timeout. The HALO batch on top of main is part of the re-verification run once every open comment on the stack is addressed. If the delay shows up there, I'll revisit.
- **`_handleCollectionSync` gate.** Agreed that it doesn't apply to the direct `schedule()`. It shouldn't: a scope change is an explicit policy change.
- **deny→allow flake.** With this stack on top of #13209 the test passed 20 of 20 full-file runs locally, but that says nothing about the kick-less path you describe, so I'm leaving it as a known flake.
- **The skipped test.** The description was stale. That test isn't in the PR any more, because it passed without the fix and proved nothing. The `echo-network-adapter.test.ts` cases are what run in CI. Description updated.

_🤖 Written with [Claude Code](https://claude.com/claude-code)_

PR #13204, opened 2026-09-17
network-manager: deliver WebRTC frames in order and serialize offers with remote descriptions

wittjosiah:
Two-peer WebRTC sessions could deliver frames out of order or fail negotiation, and WebKit's web process could abort on an empty ICE candidate.

- **Deliver frames in arrival order.** A data channel defaulted to blob frames, and converting a blob is asynchronous, so the frame after it could be pushed to the stream first. Channels now receive ArrayBuffers, and conversion runs in arrival order.
- **Serialize local offers with remote descriptions.** `onnegotiationneeded` created and set a local offer outside the lock that guards applying a remote offer or answer, so the two could interleave. It now takes that lock.
- **Skip empty end-of-candidates signals.** Trickle ICE does not need them, and WebKit's GStreamer backend aborts the web process when one is added.
- **Bound the memory transport's wait for its remote signal by the connection timeout.** Closing the transport ends that wait, so a closed transport no longer reports a timeout later.

## Evidence

- `rtc-data-channel.test.ts`: a blob frame and the frame after it arrive in order.
- `rtc-peer-connection.test.ts`: an empty candidate is never added to the connection.
- `network-manager:test` passes (46 tests).
- Part of the branch that passed 30 of 30 dispatched e2e CI runs with no retries.

## Impact

Real users: replication over a WebRTC connection could receive a frame out of order, and a WebKit peer could crash while connecting.

Part of DX-1264. PR 2 of 7, split out of #13013 (replaces #13194, which GitHub closed when the stack was reordered).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebRTC connection stability by preserving inbound message order.
  - Prevented overlapping offer and answer processing during negotiation.
  - Ignored empty ICE end-of-candidates signals that could cause WebKit crashes.
  - Improved connection timeout and shutdown handling when remote signals are delayed or unavailable.
  - Added more detailed error information when WebRTC negotiation fails.

- **Tests**
  - Added coverage for frame ordering, empty ICE candidate handling, and reliable transport shutdown behavior.

---
🚀 Composer preview: https://pr-13204-composer-dev.dxos.workers.dev


packages/core/mesh/network-manager/src/transport/webrtc/rtc-data-channel.ts
```diff
@@ -103,30 +105,30 @@ export const bindDataChannel = (
...
-        // The only await on this path, so re-read the binding after it: disposal in the meantime
-        // leaves nothing to push to.
-        data = Buffer.from(await data.arrayBuffer());
-        if (!duplex) {
-          return;
-        }
-      }
-      duplex.push(data);
+      const data = event.data;
+      frameOrder = frameOrder.then(async () => {
```
dmaretskyi, line 133: this looks like a promise queue, those are racy, specifically around lifecycle.

Replace with TransformStream, those are reusable, and `transform` is called sequentialyl and preserves order

dmaretskyi (commented):
good to land, after fix is implemented

packages/core/mesh/network-manager/src/transport/webrtc/rtc-data-channel.ts
dmaretskyi: also, try avoiding node Buffer, prefer native web apis

PR #13205, opened 2026-09-17
echo-host: wake Subduction rounds when one peer is lost

wittjosiah:
When one Subduction peer went away, every sync round in flight to it waited out the full request timeout, and so did every local edit queued behind those rounds.

- **Disconnect a lost peer in Subduction.** The transport's disconnect callback leaves the connection registered in `subduction_core`, with its requests pending until they time out. On a bound peer's `peer-disconnected`, `AdapterConnections` now calls `disconnectFromPeer`, which removes the connection and settles those requests. The round finishes against the peers still connected. Nothing keeps running in the Wasm, and only rounds with a request to that peer are affected.
- **A handshake that completes after its peer left is disconnected the same way**, rather than leaving a connection nobody tracks. A reconnecting peer's handshake waits for the old connection's disconnect, because both bind the same subduction PeerId.

## Evidence

- `automerge-repo-subduction.test.ts`, three tests in a block that runs in CI (in-memory adapters, unlike the rest of the suite): "losing one peer mid-round does not hold back edits for the others", "a peer offered again mid-round does not hold back later edits" and "a peer lost before its handshake completes does not hold back edits". All three fail without this patch and pass with it. With `CI=1` the block passed 10 of 10 runs, and the whole file passed 20 of 20 locally.
- The echo-host automerge and edge suites pass (184 tests). `automerge-host-subduction.test.ts` passed 20 of 20 file runs, including "deny→allow flip auto-recovers".

## Impact

Real users: when one device or the edge dropped, edits to documents syncing with other peers could stall for up to the 60s round timeout.

Part of DX-1264. PR 3 of 7, split out of #13013 (replaces #13195, which GitHub closed when the stack was reordered).

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronization now completes when peers disconnect, allowing pending requests and queued edits to proceed with remaining peers.
  * Improved handling for peers that disconnect during connection or handshake.
  * Reconnecting peers and late handshakes no longer stall synchronization rounds.

* **New Features**
  * Added peer-binding events for improved visibility into active synchronization connections.
  * Added efficient document handle lookup.
  * Limited concurrent synchronization activity to maintain stability under heavier workloads.

---
🚀 Composer preview: https://pr-13205-composer-dev.dxos.workers.dev


mykola-vrmchk:
Notes from landing #13209 (now on main):

- **CI won't run the two new Subduction regression tests.** They sit inside `describe.skipIf(process.env.CI)` in `automerge-repo-subduction.test.ts` (line 43). Only the `collection-synchronizer.test.ts` case guards this change in CI. While landing #13209 we soaked the four skipped Subduction suites plus `automerge-host.test.ts` locally: 38 of 40 full-suite runs passed. Both failures were pre-existing hangs:
  - "deny→allow flip auto-recovers via AutomergeHost machinery (no manual kick)", about 5% on main. A heal round reports `succeeded` while one host still lacks the other's commit, and nothing retries it.
  - "authorizeConnect: per-peer denial in star topology": normally about 28 ms, but rarely `Document … is unavailable` for the whole 5 s window.

  Those two are what currently blocks re-enabling the suites in CI. It would be worth checking whether waking rounds on connection loss changes either.
- **The `onConnectionOpen` guard looks right alongside #13209.** `onConnectionClosed` still clears `interestedPeers`, so a real reconnect is re-queried straight away. #13209's diverged-document resync relies on that: it drops its per-peer resync bookkeeping on disconnect, and the next collection pass re-arms it.
- No overlap in files with #13209. It changed `automerge-host.ts`, the SQLite storage adapter and `FeedStore`, not the automerge-repo patch or `collection-synchronizer.ts`.

_🤖 Written with [Claude Code](https://claude.com/claude-code)_

mykola-vrmchk:
The change is correct for the paths it covers, and it lines up with `packages/services/db-service/src/worker/subduction/replication-spec.md` in edge: cases 1, 3, 4 and 5 all reach the wake through `_onRestartRequested` → `_closeReplacedConnection` → `close()` → `peer-disconnected`. Case 2 is the one it does not help — while the edge's `UNKNOWN_CONNECTION_ERROR_MIN_INTERVAL_MS` throttle is in effect (~50s in the run the spec cites) the client's connection is still open, so nothing is woken and the round burns its full timeout anyway.

**The real fix is upstream.** Every awkward part of this patch traces back to one missing thing: `syncWithAllPeers` has no cancellation. We are racing a promise we cannot stop, so the round is *abandoned* rather than *ended*. If `subduction_core` settled pending `RequestId`s on connection removal — or `syncWithAllPeers` took an `AbortSignal` — this collapses to "cancel the round, re-issue against the remaining peers", and `#disconnectAbortWaiters`, the sentinel, and both risks below all disappear. Worth raising with the Subduction/Automerge side rather than carrying more of this shape in the patch.

The two consequences of having no cancel:

1. **The orphaned round overlaps its own replacement.** The abandoned `syncWithAllPeers` keeps running in the WASM and resolves at its own 60s timeout, discarded — but `finally` has already cleared `entry.syncInFlight` and released the `#syncGate` slot, so a second round for the same sedimentree starts immediately. Before this PR the abandonment path (full disconnect) always went to heal backoff, never to an immediate overlapping re-drive, so concurrent same-sid rounds are new. CRDT semantics make corruption unlikely, but the orphans sit outside `MAX_IN_FLIGHT_DOC_SYNCS` accounting: a flapping peer accumulates untracked WASM requests across a 60s window while the gate cheerfully admits 100 fresh ones.

2. **The wake is source-global, not round-scoped.** One `Repo`/`SubductionSource` per client covers every space, so losing one space's edge peer wakes and re-drives in-flight rounds for all of them. The deliberate no-generation-bump keeps idle documents out of it, which is the right call and does contain the blast radius — but it is still cross-space, and only because `syncWithAllPeers` does not tell us which peers a round is actually waiting on.

Plus one concrete nit, independent of the above:

3. **`boundPeers` can strand an entry.** If `peer-disconnected` arrives after `peer-candidate` but before `boundPeers.add(peerId)` — a handshake that completes *after* the peer went away, rather than throwing into the existing `catch` — the `delete` returns false (no wake) and the later `add` inserts an id that is never removed. Since the client mints a fresh `connectionId` per connection, that id is never reused, so it is a slow `Set` leak on the adapter as well as a missed wake. A "disconnected during handshake" flag checked before `add` covers it.

Two smaller notes:

- The `collection-synchronizer.ts` guard is safe but defensive — I could not find a production path where `onConnectionOpen` fires for a peer already in `interestedPeers`. `_onConnectionOpen` asserts `invariant(!this._connections.has(peerId))`; under subduction `onConnectionAuthScopeChanged` is a callback rather than `'reannounce-peer'`; the mesh promote path closes before it opens; and a reconnecting edge peer arrives under a brand-new peer id anyway. Worth keeping, but the "reconnecting peer" framing in the description overstates what it changes.
- `patches/@automerge__automerge-repo@2.6.0-subduction.48.patch` exists in both repos and has already drifted (edge's copy differs from dxos `main` on the `getHandle` hunk). Functionally moot for the deployed edge worker — `db-service` drives raw `Subduction` via `acceptTransport` and never constructs a `Repo`/`SubductionSource` — but edge's *test* peers do use `subductionAdapters`, so an edge test reproducing a one-peer-loss stall stays red until this is ported.

---

wittjosiah:
Re https://github.com/dxos/dxos/pull/13205#issuecomment-5732082342

- **CI coverage.** The connection-loss tests now live in their own block, which runs in CI. They use in-memory adapters, not the signal server that keeps the rest of the suite out. There are three of them now (see the other reply). With `CI=1` the block passed 10 of 10.
- **The two pre-existing hangs.** Neither test loses a connection, so this change can't reach them. On the reworked patch, locally: "deny→allow flip auto-recovers" passed 20 of 20 alone and 20 of 20 at file level, and "authorizeConnect: per-peer denial in star topology" passed 20 of 20. The policy file failed 1 of 20 on a different test, "authorizePut on client does NOT gate outbound to a permissive server", which then passed 15 of 15 alone on both the old and new patch.
- **`onConnectionOpen` guard.** Removed, per your second comment.

_🤖 Written with [Claude Code](https://claude.com/claude-code)_

wittjosiah:
Re https://github.com/dxos/dxos/pull/13205#issuecomment-5732490727

Reworked around a cancellation Subduction already has. `disconnectFromPeer` removes the connection in `subduction_core` and settles its pending requests. The transport's disconnect callback does neither: `disconnectFromPeer` still returns `true` after that callback has fired. On a bound peer's `peer-disconnected`, `AdapterConnections` now calls `disconnectFromPeer`. The sentinel race for a single lost peer, the `connectionLost` plumbing and the immediate re-drive are gone. The full-disconnect race that was already on main stays.

1. **Orphaned rounds.** None now. The round's request to the lost peer settles in the Wasm, so the round ends. Nothing keeps running outside `MAX_IN_FLIGHT_DOC_SYNCS`, and no second round for the same sedimentree starts while the first is still out.
2. **Source-global wake.** Gone. Only requests to the lost peer settle. Rounds for other spaces that had no request to that peer are untouched.
3. **`boundPeers` race.** Fixed. Each repo peer has one tracked attempt. A handshake that completes after `peer-disconnected` is disconnected at once. The entry is dropped when the attempt ends, so nothing leaks. A reconnecting peer's handshake also waits for the old connection's disconnect, because both bind the same subduction PeerId. New test: "a peer lost before its handshake completes does not hold back edits". It fails on the previous version of this patch.

- **Case 2.** Agreed, this doesn't help while the edge throttle keeps the socket open. The client can't tell that connection from a slow one, so the fix belongs on the edge side.
- **Upstream.** The cancellation exists (`disconnect_from_peer`). What's missing upstream is that a transport's `onDisconnect` should do the same thing. I'd raise that with the Subduction side instead of asking for `AbortSignal`.
- **Collection-synchronizer guard.** You're right: every path into `onConnectionOpen` for a known peer goes through `onConnectionClosed` first. Removed, along with its test, and the description no longer claims it.
- **Edge patch drift.** Edge's copy is behind dxos main by more than the `getHandle` hunk. `detach`, the heads-changed listener and the knownHashes mirror are also missing. I'll port the full patch to edge once this lands, so edge tests exercise the same code.

_🤖 Written with [Claude Code](https://claude.com/claude-code)_

PR #13218, opened 2026-09-18
fix(echo-host): stop reading and writing Subduction's remote-heads family

mykola-vrmchk:
Stacked on #13209. Makes Composer work with profiles that are **already bloated**, with no migration.

## Problem

automerge-repo's `SubductionSource` persists a `remote-heads` record per **(sedimentree, remote peer id)** under `[subduction, remote-heads, <sedimentreeId>, <peerId>]`. Its only purpose is to pre-fill a handle's sync info on attach (`loadRemoteHeads` → replay with `persist=false`).

The peer id in the key was the sync server's **per-incarnation** identity (fixed on the server in dxos/edge#1113), so a new set of records accumulated with every server restart and **nothing deleted them**. One user profile:

- **1,035,965** `remote-heads` rows, 98.3% of the automerge store (808 documents × 4,068 server identities in 50 days). Real document data was ~4 MiB.
- A DevTools trace of boot: **106 `remote-heads` range scans = 9,135 ms, 98.8% of all SQLite time**, averaging 86 ms each (306 ms on a freshly imported copy) against 0.3 ms for real document reads. All of it is blocking OPFS reads, and the worker was saturated until every tab→worker RPC timed out and Chrome reported *Page Unresponsive*.

The edge fix stops **new** growth, but users already carrying these rows still can't boot.

## Fix

`SqliteStorageAdapter` skips the family in `load`, `save`, `saveBatch` and `loadRange`:

```ts
const isRemoteHeadsKey = (key: StorageKey): boolean =>
  key[0] === SUBDUCTION_PREFIX && key[1] === REMOTE_HEADS_FAMILY;
```

- **Reads return empty without touching SQL.** An already-bloated profile boots like a clean one, and the dead rows are never scanned.
- **Writes are dropped**, so the family stops growing whatever the server does.
- **Deletes are not skipped**, so `removeDocument`'s sweep of `SUBDUCTION_KEY_FAMILIES` still reclaims a document's legacy rows.

## Why it's safe to stop persisting them

- **Nothing in DXOS reads them.** Grepping every package (`.ts` + `.tsx`) for `remote-heads`, `remoteHeads`, `getSyncInfo` and `subduction-remote-heads` finds only a debug log on write (removed here) and the GC family list.
- **The sync engine never stores them.** The WASM's `SedimentreeStorage` interface has only ids, commits and fragments. `saveRemoteHeads`/`loadRemoteHeads` are JS-only bridge methods called by `SubductionSource`.
- **Losing the replay only affects display.** A handle's remote heads get filled by the next live sync instead of from disk; the heads used for convergence (collection sync, `SqliteHeadsStore`) are unaffected.
- **No broad scan hits the dead rows.** Every `loadRange` the bridge issues names a family (`ids`, `commits`, `blobs`, `fragments`, `fragment-blobs`, `remote-heads`), so the others are separate key ranges.

## Testing

- New adapter tests: the family is neither written nor read while other subduction families round-trip; and a row planted with raw SQL (a legacy bloated profile) reads as empty but is still deleted by `removeRange`.
- **Both fail against the previous adapter** (`expected Uint8Array[…] to be undefined`; `expected [ {…} ] to deeply equal []`).
- Full `echo-host` suite: 365 passed, 2 skipped, 1 todo. Subduction suites (CI-skipped): 80/80.
- `oxlint`, `oxfmt --check`, typecheck clean.

**End-to-end QA:** import the original bloated profile into this PR's deploy. It should boot, and with #13209 the stuck space should reach 16/16.

---
🚀 Composer preview: https://pr-13218-composer-dev.dxos.workers.dev


mykola-vrmchk:
Closing in favour of #13209, so this lands as one dxos PR. Per review, the adapter keeps reading and writing `remote-heads` records; instead, #13209 adds a **Clear Heads** action to `/recovery.html` that deletes them from an already bloated profile. The growth itself is fixed on edge in dxos/edge#1113.

_🤖 [Claude Code](https://claude.com/claude-code)_

PR #13228, opened 2026-09-18
ci: cut e2e job time by fixing cache restores, and run e2e on every Check

wittjosiah:
Each e2e job spent 13 to 21 minutes in moon to run 2 to 4 minutes of tests. The builds weren't missing the cache; restoring them from it was the slow part. This PR fixes three causes and makes e2e part of every Check run.

## What was slow

- **`composer-app:bundle-e2e` took longer to download than to build.** The output is about 320 MB, 179 MB of it sourcemaps. It builds in 25s but took 3 to 11 minutes to restore. Uploads hit moon's 300s flush limit and were dropped, and each job that built it then sat idle for 5 minutes.
- **`moon exec --concurrency=1` restored the ~390 build outputs one at a time,** not only the suites.
- **The `rest` shard ran `composer-app:e2e` too.** `moon query tasks` ignores `project!=composer-app`, so `rest` restored the whole composer build graph. In 4 of 60 jobs it re-ran the full composer suite.

## Changes

- `composer-app:bundle-e2e` uses `cache: local`, and the `e2e-bundle` job is gone.
- e2e suites share a moon `mutex` in place of `--concurrency=1`, so suites stay serial and restores run in parallel.
- The `rest` target list drops `composer-app` in the node filter.
- e2e runs on every trigger, scoped to affected projects the same way `test` is. The `e2e` boolean dispatch input is replaced by `only` (`all`, `check`, `test`, `e2e`; default `all`). A dispatch still forces every suite to run instead of replaying a cached result. `cli` keeps its main/release gate, plus `only: all` on dispatch.

## Measurements

10 dispatched runs of the first commit, compared with 20 runs of #13222's head (`3eeedf7dcf`). Same concurrency, same build hashes.

| median | before | after |
|---|---|---|
| e2e · composer job (p90) | 15.7m (17.0) | 12.5m (13.3) |
| e2e · rest job (p90) | 22.6m (25.7) | 8.2m (8.6) |
| wait before e2e starts | 15.4m | 0.1m |
| whole Check run | 39.4m | 14.3m |
| failed tasks | 2 | 0 |

Every suite ran in every fix run, with no cached replays.

## Not covered by these runs

- **Both sets ran on commits whose builds were already cached.** Without `e2e-bundle`, the six e2e jobs on a fresh commit may build missing libraries alongside `check`. This PR's own CI is the first measurement of that case.
- **The `only` input and the PR-triggered e2e path haven't run yet.** This PR's CI exercises the PR path.
- **Composer still spends about 7 minutes restoring its build graph.** The `test` jobs pay the same cost, 10 to 12 minutes, even when every test is a cache hit. That's a follow-up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Space creation now continues successfully when database initialization takes longer than five seconds.
  - Default spaces open correctly on first launch, even when Home is still loading.
  - Plugin enable/disable settings no longer unexpectedly revert during startup or synchronization.
  - Deck navigation now resolves missing paths before opening them.
  - Dialogs close reliably in WebKit, including during long-running page tasks.
  - WebRTC invitations are more reliable in browser-based end-to-end testing.
- **Improvements**
  - Worker storage now reports unusable OPFS at startup instead of repeatedly failing database operations.


wittjosiah:
Linked into the wrong native stack (#13200); reopened on stack #13208.

PR #13242, opened 2026-09-20
echo: clear entity manager state on close so a reopen does not duplicate ids

dmaretskyi:
## Root cause

`EntityManager.close()` tore down only the repo proxy: it left `_objects` (the object-core registry), `_spaceRootDocHandle`, the document-handle maps and the pending-load sets fully populated. `DatabaseImpl` is a `Resource` that can be reopened, and its `_open` calls `openWithSpaceState` unconditionally when a root URL is known — that path takes every inline object id from the space root directory and calls `_createInlineObjects`, which asserts `invariant(!this._objects.has(id))`. On the second open, every surviving core trips it, and the space fails to open.

The reopen is a real production path, not a test artifact: `SpaceProxy._reset()` closes the database (space destroy, or `SPACE_REQUIRES_MIGRATION`), and `_processSpaceUpdate` then takes the `isReopening` branch (`state → SPACE_READY` while `!this._databaseOpen`) and re-runs `_initializeDb` → `db.open()`. That is exactly the reported stack (`tT._initializeDb → Qf.open → _open$synchronized → openWithSpaceState → _createInlineObjects`).

The invariant violated is the entity manager's identity invariant — one live core per object id per open session — broken by state from a previous session leaking across `close()`.

## How it was reproduced

New test in `packages/core/echo/echo-client/src/core-db/entity-manager.test.ts`: `database reopens after close with inline objects in the space root`. It writes an inline object into the space root doc, loads its core (holding it across the reopen, since the registry holds cores weakly), then `db.close()` / `db.open()`.

Against the unmodified `entity-manager.ts` (fix stashed), `moon run echo-client:test -- entity-manager.test.ts -t 'reopens after close'` fails with the production error verbatim:

```
Error: invariant violation [!this._objects.has(id)] at packages/core/echo/echo-client/src/core-db/entity-manager.ts:1923
 ❯ EntityManager._createInlineObjects src/core-db/entity-manager.ts:1923:7
 ❯ EntityManager.openWithSpaceState src/core-db/entity-manager.ts:293:12
 ❯ DatabaseImpl._open src/proxy-db/database.ts:395:7
 ❯ DatabaseImpl._open$synchronized ../../../common/async/src/mutex.ts:135:14
 ❯ DatabaseImpl.open ../../../common/context/src/resource.ts:130:5
```

With the fix, `moon run echo-client:test` (whole package) is green: `Test Files 36 passed | 1 skipped (37)`, `Tests 581 passed | 7 expected fail | 12 skipped | 2 todo (602)`. `pnpm format` and `moon run echo-client:lint` both clean.

## The fix

`EntityManager.close()` now drops the per-open state it used to keep: it unsubscribes the doc handles, clears the handle references (including `_spaceRootDocHandle`), clears the core registry and the pending-load / pending-update sets, then closes the repo proxy. A reopen therefore reloads the root doc handle and rebuilds cores from the directory, as a first open does. The invariant is untouched — no upsert, no try/catch — and `EntityManager.close()` has exactly one caller (`DatabaseImpl._close`), so nothing else depended on the state surviving.

### On the second signal

The other production error reported at `entity-manager.ts:247` (1 occurrence, same window) could not be mapped to a source location: the deployed line numbers are from the v0.11.1 build and that tag is not present in this (shallow) checkout, while current line 247 is in the constructor's field declarations. No evidence ties it to this root cause, so it is left alone.

Observed in SigNoz: Composer production, 2 occurrences over 7d (2026-09-13 → 2026-09-20). Blocks the affected space from opening.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue that could prevent a database from reopening when inline objects were stored in the space root document.
  - Database state is now properly reset during close, allowing objects to be restored and accessed after reopening.

- **Tests**
  - Added coverage confirming inline objects remain available after closing and reopening a database.

---
🚀 Composer preview: https://pr-13242-composer-dev.dxos.workers.dev


dmaretskyi (commented):
Approved

dmaretskyi (commented):
Approved

PR #13243, opened 2026-09-20
echo: cancel in-flight object loads when the database is torn down

dmaretskyi:
## Root cause

`EntityManager` resolves object documents asynchronously: `_loadObjectDocument` / `_onObjectLinksUpdated` fan out into `_loadLinkedObjects`, which calls `RepoProxy.find` → `_getOrLoadHandle` → `_loadHandle`. `_loadHandle` opens with `invariant(this._lifecycleState === LifecycleState.OPEN)` (`packages/core/echo/echo-client/src/automerge/repo-proxy.ts`).

Nothing tied those loads to the manager's lifetime. When the owning database is torn down — space closed, client reset, navigation — `EntityManager.close()` closes the repo proxy while link resolution is still in flight, and the continuation lands on a closed proxy and raises the invariant. The raising frames are all fire-and-forget (`void _loadHandleForObject`, a void-called space update, a `Ref.load()` nobody awaits), which is why every occurrence surfaced as an unhandled rejection rather than a caught error. Two adjacent defects made the window wider: `close()` left the `change` subscriptions on every doc handle in place, so a change delivered during teardown could start a fresh load; and `_loadHandleForObject`'s catch retried unconditionally, so a document load failing *because* of teardown retried forever against a dead proxy.

## How it was reproduced

Two tests in `packages/core/echo/echo-client/src/core-db/entity-manager.test.ts`, both deterministic — no sleeps, no timing luck.

**1. The load lands after teardown** — "loading a linked object after teardown is cancelled, not an invariant violation". Creates a client database over a space whose object lives in its own **linked** document (so loading it must go through `RepoProxy.find`), closes the database, then asks for the object. Before the fix:

```
 FAIL  src/core-db/entity-manager.test.ts > ... > loading a linked object after teardown is cancelled, not an invariant violation
AssertionError: expected Error: invariant violation [this._lifecyc… to be an instance of ContextDisposedError
```

**2. The load spans the teardown** — "a load pending across close and reopen is cancelled, not resolved from the new lifetime". Advertises a document in the space directory that was minted on an isolated peer, so the load reaches the network wait and stays there; then closes and reopens the database while the load is suspended. Before the fix the load never settles at all:

```
 FAIL  src/core-db/entity-manager.test.ts > ... > a load pending across close and reopen is cancelled, not resolved from the new lifetime
Error: Timeout [5,000ms]
    at asyncTimeout (packages/common/async/src/timeout.ts:65:68)
    at src/core-db/entity-manager.test.ts:513:28
 Tests  1 failed | 25 skipped (26)
```

After the fix the file passes 26/26, and so do the package and its dependent plugin suite:

```
moon run echo-client:test
 Test Files  36 passed | 1 skipped (37)
      Tests  587 passed | 7 expected fail | 12 skipped | 2 todo (608)

moon run plugin-projects:test
 Test Files  19 passed (19)
      Tests  58 passed | 1 skipped (59)
```

`pnpm format` and `moon run echo-client:lint` are clean.

## The fix

Lifecycle-correct cancellation rather than swallowing the rejection or relaxing the invariant. `EntityManager` owns its lifetime state and **every load validates that lifetime after each asynchronous boundary, on the success and failure paths alike** — a check taken only before the first await cannot see a close that happens during the wait.

- **`#generation`**, bumped by `open()`, is the lifetime token. A load captures it on entry and revalidates with `#isStale(generation)` after every await. Re-reading `#closed` alone is ABA-unsafe: `open()` clears it, so a close-and-reopen during a wait would read as live again and hand a pre-close caller a post-reopen result. (`#closed` itself is the flag #13242 added to this class; the two fixes name one concept once. `EntityManager` does not extend `Resource` — `implements IDatabaseBinding` with an ad-hoc `open`/`close` pair — so there is no inherited `isOpen`/`_lifecycleState`. Making it a `Resource` is the most normalized answer but changes the open/close contract for every caller, so it belongs in its own change.)
- **`#lifetimeEnded`**, emitted by `close()`, releases loads that are waiting on an update. A closed manager answers `getObjectCoreById` with `undefined`, so a readiness predicate can never fire once closed — without this the wait would hang forever, which is strictly worse than rejecting. `loadObjectCoreById` races its wait against it and `batchLoadObjectCores` subscribes for the duration of its wait; both settle with `ContextDisposedError`, and both re-check the generation afterwards.
- **`_loadHandleForObject`** validates after `whenSettledOnDisk()` and after `whenReady()` before calling `_onObjectDocumentLoaded`, inside the detached background `whenReady().then(...)` — which is unbounded and routinely outlives the space that started it — and in the catch, where a stale generation abandons rather than retries.
- **`_loadLinkedObjects`** abandons the fan-out when closed, logging why. Nothing awaits this work, so there is nothing to reject.
- **`close()`** unsubscribes from document handles *before* closing the proxy, so a change delivered during teardown cannot start a new load.

The invariant in `repo-proxy.ts` is untouched.

Observed in SigNoz: Composer preview, ~20 occurrences over 7d (2026-09-13 → 2026-09-20), all as unhandled rejections.

## Summary by CodeRabbit

- **Bug Fixes**
  - In-flight object loads are now canceled when a database is torn down.
  - Loading linked objects after teardown now returns a controlled disposal error instead of an unhandled invariant violation.
  - Prevents additional load attempts against closed database connections during shutdown.
  - Pending loads no longer resolve against a newly reopened database lifetime after the original database has been closed.

---
🚀 Composer preview: https://pr-13243-composer-dev.dxos.workers.dev


packages/core/echo/echo-client/src/core-db/entity-manager.ts
```diff
@@ -379,9 +379,22 @@ export class EntityManager implements IDatabaseBinding {
...
     await this._repoProxy.close();
   }
 
+  /**
+   * Whether the repo proxy this manager loads through is gone. Loads are asynchronous and outlive
+   * the space that started them, so every path reaching `RepoProxy.find` checks this first — the
+   * proxy asserts it is open and a load landing after teardown would raise an invariant violation
+   * nobody is left to catch.
+   */
+  private get _isTornDown(): boolean {
```
dmaretskyi: Doesn't it have its one lifecycle via resource base class. I'd rather normalize naming and use own lifetime state

dmaretskyi:
Reworked in a0b82502 — `_isTornDown` is gone.

One correction to the premise: `EntityManager` does not extend `Resource` — it is `export class EntityManager implements IDatabaseBinding` with an ad-hoc `async open(ctx)` / `async close()` pair. The `isOpen` / `_lifecycleState` I was reading come from `RepoProxy extends Resource`, which is exactly your objection: I was deriving this class's liveness from another object's lifecycle.

Your preference stands, so the class now owns the state. I deliberately reused the flag that #13242 is already adding to this same class rather than inventing a second name for it:

```ts
/** Set from {@link close} until the next {@link open}: the working set is gone, not merely unloaded. */
#closed = false;
```

Same declaration, same set/reset sites, so the two PRs converge instead of landing two flavours of "torn down". They will conflict in `close()` and at the declaration; both sides resolve to identical lines.

I considered making `EntityManager extend Resource` for real — the most normalized answer, and what you half-expected to be true — but it changes the open/close contract for every caller of this class and is a refactor in its own right, not something to smuggle into a telemetry bug fix. Happy to do it as a follow-up if you want it.

Verified after the rework: `echo-client:test` 581 passed / 37 files, `plugin-projects:test` 58 passed / 19 files (running dependents since #13242 hit plugin-projects from an echo-client-only change), plus `pnpm format` and `echo-client:lint` clean. The repro test still fails without the fix and passes with it.

---

dmaretskyi (commented):
Approved

dmaretskyi:
Fixed in ad64a127. The finding was right: every guard was check-before-await, so a close landing *during* a load's wait was never seen.

**Mechanism: a generation counter, not a re-read of the flag.** `#generation` is bumped by `open()`; each load captures it on entry and revalidates with `#isStale(generation)` after every asynchronous boundary, success and failure paths alike. Re-reading `#closed` would be ABA-unsafe exactly as described — `open()` clears it, so a close-and-reopen during a wait reads as live again.

Hanging was the other half, and a hang is not an acceptable teardown outcome, so `close()` now emits `#lifetimeEnded`: `loadObjectCoreById` races its wait against it, `batchLoadObjectCores` subscribes for the duration of its wait, and both settle with `ContextDisposedError` instead of pending forever or reporting an inactivity timeout. `_loadHandleForObject` validates after `whenSettledOnDisk()`, after `whenReady()`, inside the detached background `whenReady().then(...)`, and in the catch, where a moved generation abandons rather than retries.

**New regression test**, driving the race itself: a document advertised in the directory but minted on an isolated peer, so the load reaches the network wait and stays there; then close, then reopen, then assert the pre-close load cannot resolve from the new lifetime. Watched it fail first — on the current branch source it does not reject, it never settles:

```
 FAIL  src/core-db/entity-manager.test.ts > ... > a load pending across close and reopen is cancelled, not resolved from the new lifetime
Error: Timeout [5,000ms]
    at asyncTimeout (packages/common/async/src/timeout.ts:65:68)
    at src/core-db/entity-manager.test.ts:513:28
 Tests  1 failed | 25 skipped (26)
```

With the fix, 26/26 in that file — the original teardown test still passes — and both suites green on the final source:

```
moon run echo-client:test       Test Files 36 passed | 1 skipped (37)   Tests 587 passed | 7 expected fail | 12 skipped | 2 todo (608)
moon run plugin-projects:test   Test Files 19 passed (19)               Tests 58 passed | 1 skipped (59)
```

`pnpm format` and `echo-client:lint` clean. Built on `c6e668c3`; no rebase, no force-push. PR body's `## The fix` updated to describe the generation-based validation.

---

dmaretskyi:
Fixed in 003227ee.

**Listener: fixed.** `loadObjectCoreById` now subscribes to `_updateEvent` explicitly, keeps the unsubscribe, and calls it in the same `finally` as `cancellation.dispose()`, so every exit path releases it — cancellation, timeout and success alike. As noted, `Event.waitFor` could not be fixed in place: it unsubscribes only inside its matched branch and hands back a bare promise. The predicate and `isReady()` semantics are byte-for-byte the same, and the listener is still registered *before* `_loadObjectDocument` so no update can be missed in the gap.

**Timer: not touched**, per your correction — `asyncTimeout` clears its timer in its own `finally` and `unref`s it, and nothing about the restructure made it fall out naturally.

**`batchLoadObjectCores`: no change needed.** It already subscribes with an explicit `unsubscribe` and releases it on every exit, including the cancellation path, which calls `unsubscribe?.()` before rejecting.

**The leak is asserted directly**, no privates involved — `db._updateEvent` is a public accessor on `DatabaseImpl` and `Event.listenerCount()` is public, so the pending-across-close test now snapshots the count before the load and compares after. Watched it fail on the pre-fix source:

```
 × a load pending across close and reopen is cancelled, not resolved from the new lifetime
AssertionError: expected 4 to equal 3
 Tests  1 failed | 25 skipped (26)
```

That is the leaked listener, one per interrupted load. With the fix, 26/26 in that file — both teardown tests still pass — and both suites green:

```
moon run echo-client:test       Test Files 36 passed | 1 skipped (37)   Tests 587 passed | 7 expected fail | 12 skipped | 2 todo (608)
moon run plugin-projects:test   Test Files 19 passed (19)               Tests 58 passed | 1 skipped (59)
```

`pnpm format` and `echo-client:lint` clean. Pushed on top of `ad64a127`; no rebase, no force-push.

---

PR #13244, opened 2026-09-20
compute-runtime: model stale trigger references instead of failing the pass

dmaretskyi:
## Root cause

`TriggerDispatcher` assumed every stored trigger reference stays resolvable, and nothing reconciles them when it does not. A trigger outlives what it points at in two ways — its target object is deleted (`EntityNotFoundError`) or the plugin contributing its operation is gone, so no handler is registered (`NoHandlerError`) — and neither was modelled:

- `invokeScheduledTriggers`, feed branch: `Database.load(spec.feed).pipe(Effect.orDie)` turned a deleted feed into a defect that escaped the per-trigger loop and killed the whole dispatch pass. In natural time that surfaces at the `Effect.tapCause` in `start()` as `trigger dispatcher error` — which also sets `enabled: false` and tears the subscription down, so **one stale trigger stops the dispatcher for the whole space**.
- `#subscribeToTriggerSource` (reached from `refreshTriggers` → `#refreshReactiveSources`): the same `orDie` failed the entire refresh pass — `failed to refresh triggers` — leaving every other trigger's cron schedule and reactive source unreconciled. The function's own doc already promised to return `undefined` for "a feed reference that no longer resolves"; `orDie` made that unreachable.
- `invokeTrigger`: an unresolvable runnable or an unregistered handler was classified as a genuine execution failure — `trigger execution failure` — which arms the failure cooldown, so the trigger re-fires and re-throws every 30s forever and never gets flagged as broken.

Shape A (`NoHandlerError`) and Shape B (`EntityNotFoundError`) are **one cause**: an unresolvable reference is not a modelled state. The telemetry confirms it — the same uri `echo:///01KYMGPJCXG398JQYERR2BJ80W` produced both the dispatch-pass abort (`:457`) and the refresh-pass abort (`:1073`), so a fix that hardened only one path would have been incomplete.

## How it was reproduced

Four new tests in `packages/core/compute/compute-runtime/src/triggers/trigger-dispatcher.test.ts` (`describe('Stale References')`), each using a dangling ref of the shape seen in production:

- a deleted feed parks its trigger without aborting the dispatch pass;
- a deleted feed does not fail the refresh pass that reconciles reactive sources;
- an operation with no registered handler parks the trigger instead of arming the cooldown;
- a deleted operation object parks the trigger.

Run with `moon run compute-runtime:test -- src/triggers/trigger-dispatcher.test.ts`:

- against the unfixed dispatcher: `Tests  4 failed | 41 passed (45)` — all four new tests fail, e.g. `EntityNotFoundError: Entity not found: {"uri":"echo:///01KYMGPJCXG398JQYERR2BJ80W"}` thrown out of `TriggerDispatcher.invokeScheduledTriggers`;
- with the fix: `Tests  45 passed (45)`.

Full package suite (`moon run compute-runtime:test`): `Test Files  21 passed (21)`, `Tests  226 passed | 1 skipped (227)`. `pnpm format` and `moon run compute-runtime:lint` are clean.

## The fix

A stale reference becomes a first-class runtime state rather than an error class that escapes:

- `RuntimeTriggerState.staleReference` (`{ error, until }`), published on `TriggerRuntimeStatus` so the UI can say a trigger is broken rather than idle, and distinct from `cooldownUntil`, which still covers "the operation ran and failed".
- `isStaleReferenceCause` classifies a failure by walking the cause's reasons and each error's `cause` chain for `EntityNotFoundError` / `NoHandlerError`, so only these two are treated as stale — every other failure class keeps the existing cooldown path. No blanket try/catch.
- The two `Effect.orDie` feed loads now catch `EntityNotFoundError`, park that one trigger and continue: the dispatch pass skips it and runs the healthy triggers, and the refresh pass records no reactive source (retried on the next refresh) instead of failing.
- Parked triggers are skipped by all three scheduled-dispatch loops via `_skipReason`, and the park self-clears after `staleReferenceRetryInterval` (new option, default 15 minutes, deliberately longer than the failure cooldown) so re-enabling a plugin or restoring an object resumes dispatch without a restart.
- `invokeScheduledTriggers` publishes runtime statuses once the pass ends, so a trigger parked without ever being invoked reaches the observable state.

Observed in SigNoz: Composer production, ~7 occurrences over 7d (2026-09-13 → 2026-09-20). A stale trigger silently stops working and takes its dispatch pass with it.

---

---
🚀 Composer preview: https://pr-13244-composer-dev.dxos.workers.dev


dmaretskyi:
Approved

PR #13245, opened 2026-09-20
echo: model a core whose body has not landed as unavailable

dmaretskyi:
## Root cause

`EntityManager._onObjectDocumentLoaded` treats a ready document handle as proof that the object body arrived. It is not: a linked document can settle ready while empty — the peer holding it has synced the space directory (the `links` entry) but not yet the object payload, and `handle.whenReady()` resolves on the handle, not on the content. The core created there is bound at mount path `objects/<id>`, a path the document does not contain, so `ObjectCore.getObjectStructure()` deep-gets `undefined` while claiming to return an `EntityStructure`.

Everything downstream trusted that claim. `WorkingSetQueryExecutor._execRelationTraversal` (`source-to-relation` / `target-to-relation`) scans **every** loaded core and calls `EntityStructure.getEntityKind(core.getObjectStructure())`, which dereferences `object.system` — hence `TypeError: Cannot read properties of undefined (reading 'system')`. One body-less core takes down every relation query in the space, which is why the stack is a single deterministic variant.

## How it was reproduced

New test in `packages/core/echo/echo-client/src/core-db/entity-manager.test.ts` (`a body-less linked document yields an unavailable core that queries skip`): a two-peer fixture writes an object into a linked document, then blanks that document's `objects` before the second peer opens the space — the "directory arrived, body did not" state. The second peer triggers the load, runs a relation traversal query, then the body lands and the query is run again.

With the four source files checked out from `origin/main` and the test kept:

```
FAIL |node| src/core-db/entity-manager.test.ts > … > a body-less linked document yields an unavailable core that queries skip
TypeError: Cannot read properties of undefined (reading 'system')
Tests  1 failed | 21 skipped (22)
```

On this branch:

```
✓ a body-less linked document yields an unavailable core that queries skip  1487ms
```

Suites, both green on the branch:

```
moon run echo-client:test      → Test Files 36 passed | 1 skipped (37); Tests 581 passed | 7 expected fail | 12 skipped | 2 todo (602)
moon run plugin-projects:test  → Test Files 19 passed (19); Tests 58 passed | 1 skipped (59)
```

`pnpm format`, `moon run echo-client:lint` and a package typecheck (`tsc --noEmit`) are clean.

## The fix

The core is created whether or not the body is there, and holds the absence itself:

- **`ObjectCore.isBodyAvailable`** — true when the bound document actually carries the object at the core's mount path. It is *derived* from the document rather than stored, so it cannot disagree with what a read would find; nothing is duplicated. `EntityManager._unavailableObjects` stays what it was — the load-gating set keyed by object id, which must also cover ids that have no core at all (a `diskOnly` probe of an object absent from the directory) — and `_onObjectDocumentLoaded` marks a body-less object there, while a new `_markBodiesAvailable` clears the mark from the document's change event once the body lands. The two are one notion read at two levels, not two states kept in sync.
- **`getObjectStructure()` now returns `EntityStructure | undefined`**, so the type system names the case instead of hiding it — that is what forced the audit below rather than a `?.` at the crash site.

Read paths audited and made to check (all callers of `getObjectStructure`, plus the core-level lookups that hand a core to a reader):

| Path | Behaviour now |
| --- | --- |
| `WorkingSetQueryExecutor._coreToItem` | Returns `undefined` for a body-less core; a `WorkingSetItem` carries the `structure` it was built from |
| `_execSelectStep` — wildcard/type, id, incoming-reference selectors | Body-less cores never enter the working set |
| `_execFilterStep`, `_execFilterDeletedStep`, ordering, aggregates (`getProperty`, `getParentEid`, `getGroupKey`) | Read `item.structure`, which exists by construction |
| `_execRelationTraversal` (the production stack), `_execReferenceTraversal`, `_execHierarchyTraversal`, `_isChildOfAny` | Enumerate items, not raw cores; a hop whose target has no body is not an edge |
| `SpaceQuerySource._filterCore` (the lightweight invalidation filter) | A body-less core matches nothing |
| `EntityManager.getEntityById` / `loadEntityById` | No entity is materialized over a missing body |
| `_isCoreResolved` / `_coreOrUndefined` (`loadObjectCoreById`) | A load waits on such a core instead of resolving with it |
| `batchLoadObjectCores` | Counts it as still loading |
| Strong-dependency closure walk (`referencedObjectIds`) | Skips a core with no structure to walk |

No optional chaining was added at the crash site: `getEntityKind` is reached only with a structure that exists.

Observed in SigNoz: Composer production, 33 occurrences over 7d (2026-09-13 → 2026-09-20). Largest production error by volume.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of linked objects that are temporarily available without their content.
  - Queries, filters, relationship traversal, entity lookups, and loading now safely skip incomplete objects instead of returning invalid results or causing errors.
  - Object identity is preserved while content is unavailable, and objects automatically become queryable once their content arrives.
  - Added coverage for objects that initially replicate without content and later become complete.

---
🚀 Composer preview: https://pr-13245-composer-dev.dxos.workers.dev


dmaretskyi:
I'd rather have the core hold the unavailable state: that's cleaner.
Create core anyway, but mark it as unavailable. Have the read paths check that instead of assuming

dmaretskyi (commented):
Approved

PR #13246, opened 2026-09-20
echo: bound index-hit hydration so an answered query is not reported as an index timeout

dmaretskyi:
## Root cause

`IndexQuerySource._runOneShot` starts a single 20s timer (`QUERY_SERVICE_TIMEOUT`) *before* dispatching the request and clears it only once `_mapRecords` has resolved. So the budget named "index query" covers two unrelated things: the host round-trip, and the hydration of every hit returned. Hydration goes `_filterMapResult` → `_resolveIndexedObject` → `ObjectLoader.loadObject` → `EchoClient._loadObjectFromDocument` → `db._loadObjectById({ diskOnly: true })`, which waits on the entity manager's `_updateEvent` and **never settles while the object's document has not reached the client**. That is not a theoretical path — `src/query/query-api-stall.test.ts` already asserts it, holding the document synchronizer paused and checking the query does not resolve.

The consequence is the production signature: the host index answers in microseconds, one hit's document is unavailable, the whole query is held open for the full 20s and then fails as `Timeout [20,000ms]: index query` — naming the index, which was never the slow party, and giving a caller no way to tell "the index is slow" from "one document never arrived". Because every assistant operation in a session shares the query path, one stalled document takes them all out in the same millisecond, which matches the SigNoz evidence exactly.

The doc comment on `_resolveIndexedObject` asserted the opposite — "every path through it settles, so a budget here could only convert a slow load into a dropped object". `query-api-stall.test.ts` falsifies that; the comment is corrected here.

Reactive queries had the mirror-image fault: no budget at all, so `_hydrateLoop` simply never completed a pass and the source never became `_answered`, leaving every subscriber waiting forever.

**This confirms hypothesis (b)** — the host responds fast but hydration stalls — and specifically the framing that a bounded index-query budget was being spent on unbounded document hydration. It is not (a), (c) or (d): the reproduction below drives an immediate host response with no teardown and no load, and still produces the exact production error. `entity-manager.ts` is untouched; nothing here overlaps #13242 or #13243.

## How it was reproduced

Two tests in `index-query-source-provider.test.ts`, driven through the existing in-process `QueryService` harness — no sleeps, no real 20s wait. The budgets are now injectable (`queryTimeout`, `hydrationTimeout` on `IndexQuerySourceProps`) purely so a test can compress them; the production constants are unchanged.

The host emits its response on the first microtask, and `loadObject` returns a promise that never settles (the `query-api-stall.test.ts` condition, at the loader boundary). `queryTimeout` is set to 5,000ms and `hydrationTimeout` to 100ms, so a failure attributable to the index-query budget is distinguishable from one attributable to hydration.

Against `main`'s semantics (verified by reverting only the behaviour change and keeping the tests):

```
× the index-query budget is not spent on hydration, and the error names the stall  5026ms
  AssertionError: expected [Function] to throw error matching
  /index query result hydration \(1 of 1…/ but got 'Timeout [5,000ms]: index query'

× reactive queries publish the hits that did hydrate and re-hydrate the stalled one  15003ms
  Error: Test timed out in 15000ms.

Tests  2 failed | 5 passed (7)
```

The first line is the production error, reproduced from an index that answered on the first microtask. The second shows the reactive path never answering at all.

## The fix

1. **The index-query budget ends at the host's response.** `_runOneShot` clears its timer as soon as `onData` arrives, before hydration begins. `Timeout: index query` now means only what it says: the index did not respond.

2. **Each hit is hydrated under its own bound** (`RECORD_HYDRATION_TIMEOUT`, 10s), in a new `_hydrateRecord`. The expiry is compared by identity so a `TimeoutError` raised from inside hydration is not mistaken for the budget's own, and the abandoned hydration's late rejection is explicitly handled rather than left to surface as an unhandled rejection. `_mapRecords` now returns `{ results, stalled }` — a stall is reported separately from a miss, because the two callers must answer it differently.

3. **Each caller answers a stall on its own terms.** A one-shot caller gets no second pass, so a short result would read to it as the whole set: it fails, with `index query result hydration (N of M objects did not load: <ids>)`. A reactive query publishes the hits that did hydrate and logs the stalled ids; they re-hydrate with no host round-trip when their documents arrive, through the existing `_onObjectsUpdated` path (covered by the second new test).

Nothing is dropped silently and no genuine failure is hidden. In particular this deliberately does **not** reinstate "drop the stalled hit" on the one-shot path: `a slow load is waited for rather than dropped` exists precisely to forbid that, and it still passes unchanged (2.5s load, well inside the new budget).

I did not touch `QUERY_SERVICE_TIMEOUT`. The constant was never the problem — it was being charged for work it does not name.

### Verification

`moon run echo-client:test` — full package, green:

```
Test Files  36 passed | 1 skipped (37)
     Tests  582 passed | 7 expected fail | 12 skipped | 2 todo (603)
```

`src/client/index-query-source-provider.test.ts (7 tests)` all pass. The known `RepoProxy > client and host make changes simultaneously` flake did not occur in this run. `moon run echo-client:lint` clean; `tsc --noEmit` on the package clean; `pnpm format` applied.

Observed in SigNoz: Composer preview/dev, 7d (2026-09-13 → 2026-09-20). Blocks AI chat — `assistant.createChat` (12) and `assistant.generateHomeSuggestions` (16) fail on this timeout, along with `lifecycle: failed` (20), `agent hydrate skipped`, and plugin/module activation failures.

---

---
🚀 Composer preview: https://pr-13246-composer-dev.dxos.workers.dev


dmaretskyi (commented):
Approved

Issue #7387, opened 2024-08-07
[composer] re-arrange in navtree e2e test in failing only in CI in firefox


wittjosiah:
currently it is skipped in firefox

PR #13247, opened 2026-09-20
composer-e2e: extract the browser suite, bind it to mdl QA flows, trend it in PostHog

dmaretskyi:
Splits the Composer Playwright suite out of the application, gives every test a specification to point at, and puts the results on a chart.

## 1. A package of its own

`packages/e2e/composer-e2e` (`@dxos/composer-e2e`, private, `layer: automation`) takes the nine behavioural specs, the page objects and the config. Its `e2e` task serves composer-app's `bundle-e2e` through `vite preview`.

The **measurement** harnesses stay in composer-app deliberately: `startup` and `perf-*` record numbers that `check-startup-budget` and the perf nightly gate on, `dev-*` needs `vite serve`, and `welcome-focus` drives Storybook. Only the behavioural tier should ever gate a merge on a pass/fail, and that separation is now structural rather than a `testIgnore` entry in a config.

Nothing may depend on the new package in return — composer-app doing so is what moon rejects as `would_cycle`, since the suite already depends on the app. The one thing they shared was `INITIAL_URL`, now declared in `harness-helpers.ts`.

The package declares **no runtime dependencies and no `exports`**: every file in it is a test, a page object or the publisher, and nothing imports it. Its `@dxos/*` packages are `devDependencies`, matching every other package that imports `@dxos/test-utils/playwright` (`react-ui-table`, `react-ui-mosaic`, `lit-grid`, `testbench-app`).

CI's `composer` e2e shard retargets to `composer-e2e:e2e`; composer-app loses its `e2e` tag.

## 2. Every active test names an mdl QA flow

A `test QA-n` block is the *specification* of a journey; a Playwright spec is one unattended *execution* of it. The binding is declared on both sides:

```ts
test('create document', { tag: ['@QA-1'] }, async () => { … });
```

```mdl
test QA-1: Spaces and documents
  automated:
    - composer-e2e:basic.spec.ts#Basic tests › create document
```

`automated:` is a new field on the `test` ext in `packages/reflect/deus/lang/qa.mdl`. `scripts/check-qa-coverage.mjs` — a `check` job step — fails when either side names something the other lacks, when the tag does not name that flow, or when an entry points at a test that does not run. The duplication is the point: a one-sided link rots silently.

**A skipped test keeps its tag but is not coverage**, and skipping is inherited from `test.describe.skip`. `tables.spec.ts` and `inbox.spec.ts` are whole-suite skips today, so `table:QA-1` and `inbox:QA-1` have **no** automated coverage — which is the true state, and the point of not letting a list claim otherwise. All 38 tests carry a tag; 28 of them are live and claim coverage.

Flows written for the specs that had none: **QA-5** first run and logout, **QA-6** collections, **QA-7** a second device on one identity, **QA-8** two peers in one space, **QA-9** the guided tours — plus `multi-peer` and `onboarding` suites. `comments` binds to the existing `review:QA-1`.

Because they are real Playwright tags, `--grep @QA-1` runs everything that automates a flow, and they ride into the analytics as a dimension — so a chart can name the broken *journey*, not just the file.

## 3. `BEST-PRACTICES.md`

The authoring rules distilled from the existing specs and the defects behind them: testid-only selectors (and the `properties.testId` trap in `react-ui-menu`), page objects, no `waitForTimeout`, budgets claimed with a reason, the `expect.poll` folded-state pattern from `chat.spec.ts`, asserting on the model rather than the rendering, why there are no retries, and what a skip has to say for itself.

## 4 & 5. Results in PostHog

One `ci.e2e-test` event **per test per browser**, built from Playwright's JSON report (richer than the JUnit XML the Trunk uploader consumes) and published through `scripts/ci-event.mjs`, like every other CI trend in this repo — so uuid seeding and property namespacing stay in exactly one place.

- **Only from `main`, and only the `composer` shard.** A PR run measures a branch and a fork PR has no token; the `rest` cells would add unrelated packages' rows to Composer's product-analytics project.
- **The nightly is the authoritative sample.** `Check`'s `0 4 * * 1-5` schedule is a full run with `DX_E2E_RUN_ID` set, so it actually executes every suite instead of replaying a moon cache hit. A push to main that replays the cache restores no report and the publisher no-ops.
- **`always()`**, because the report of a red run is the entire point.
- **Dated by the run, not the commit** — two nightlies days apart routinely sit on one commit.
- **Skipped tests are published on purpose**: a spec that disables itself must not look like one that was deleted.
- The publisher never fails the build; problems are workflow warnings.

**[Composer e2e (nightly)](https://eu.posthog.com/project/126171/dashboard/963444)** carries the two requested tiles:

1. **Test outcomes per run** — stacked bar, passed/failed/flaky/skipped per run date.
2. **Currently failing tests** — the latest result of each `(test, browser)`, kept when it failed, with the QA flow, commit and error. Empty is the good state.

`DASHBOARD.md` is the event contract and the tile semantics, including why the failing-tests query reduces *before* it filters.

The dashboard stays empty until the first nightly publishes from `main` after this merges.

## Verification

CI is green on `51b95495`. Locally: `tsc` clean; `composer-e2e:lint` and `composer-app:lint` pass; `report.test.ts` 8/8 via CI's exact `vitest --project=node` invocation; `check-qa-coverage.mjs` ✓ 38 tests across 9 specs; Playwright discovers 38 tests in 9 files and `--grep` by tag returns `@QA-1` → 2, `@QA-6` → 4, `@review:QA-1` → 6, matching each flow's `automated:` list; the publish pipeline maps a synthetic report to correct statuses, browsers, QA tags and ANSI-stripped errors, with both no-op paths exiting 0.

The suite itself has not been *executed* here — it needs a built app bundle and three browser engines — so CI's `e2e` job is the first real run.

Four commits after the first were CI fixes, all genuine defects in the extraction: the missing vitest project (`No projects matched the filter "node"`), two dependencies orphaned by the move, and — three rounds — the six unused-dependency findings. That last one was mine to own: I twice theorised about knip's traversal (the `exports` field, then the vite/vitest config) instead of checking what comparable packages declare. knip requires *production* usage to justify a `dependencies` entry and this package has none, which is why exactly six were named and `@dxos/async` was not.

No changeset: `@dxos/composer-e2e` is private and no published package changes behaviour. The reminder bot flags `plugin-review` only because its `PLUGIN.mdl` spec document gained an `automated:` block.

## Follow-ups (tracked in `.agents/projects/composer-e2e/TASKS.md`)

- `inbox.spec.ts` is skipped wholesale — mail sync moved to EDGE, so the `page.route` fixture is never reached. Restore by seeding the mailbox rather than syncing it; `inbox:QA-1` gets its `automated:` entries back then.
- `tables.spec.ts` is `describe.skip` behind a bare "Fix table tests" TODO; same for `table:QA-1`.
- Four specs are chromium-only; only one has an attributed cause (#7387).

https://claude.ai/code/session_01Qws2NfWoDUWxpAHmwU1NPv

## Summary by CodeRabbit

- **New Features**
  - Added Composer browser test coverage for onboarding, logout, collections, collaboration, multi-device identity, comments, tables, Inbox, and guided tours.
  - Added automated QA-flow mapping and validation for browser tests.
  - Added CI reporting for individual test outcomes, including browser, status, duration, and failures.

- **Documentation**
  - Added guidance for running, debugging, and authoring Composer browser tests.
  - Added documentation for QA coverage and test-result dashboards.

---
🚀 Composer preview: https://pr-13247-composer-dev.dxos.workers.dev


packages/e2e/composer-e2e/scripts/check-qa-coverage.mjs
```diff
@@ -0,0 +1,197 @@
...
+      continue;
+    }
+    const indent = line.length - line.trimStart().length;
+    while (trail.length > 0 && indent <= trail[trail.length - 1].indent) {
+      trail.pop();
+    }
+    const describe = /^\s*test\.describe(?:\.skip)?\(\s*(['"])(.*?)\1/.exec(line);
+    if (describe) {
+      trail.push({ indent, title: describe[2] });
+      continue;
```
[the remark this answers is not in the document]
dmaretskyi: Correct, and this was the important one — thank you. Fixed in 1cb94866.

Worth recording what it was hiding: with inheritance applied the checker immediately named **ten** entries, all of `tables.spec.ts` and all of `inbox.spec.ts`. Both are whole-suite `test.describe.skip`, so `table:QA-1` and `inbox:QA-1` were reading as automated while nothing ran them — the precise failure this check exists to prevent, sitting inside the check itself.

Those `automated:` blocks are now removed rather than repaired: the honest state is that neither flow has automated coverage today, and both suites are already tracked as follow-ups (inbox needs seeding rather than syncing since mail sync moved to EDGE; tables is skipped behind a bare TODO).

---

packages/e2e/composer-e2e/src/report.ts
```diff
@@ -0,0 +1,180 @@
...
+      const status = toStatus(test);
+      const browser = test.projectName || 'unknown';
+      // The LAST attempt is the outcome; with `retries: 0` there is exactly one, and where a config
+      // overrode that, the final attempt is what `flaky` versus `failed` was decided on.
+      const result = test.results?.[test.results.length - 1];
+      const error = result?.error?.message;
+      events.push({
+        event: EVENT_NAME,
+        timestamp,
+        dedup: [options.packageName, spec.file ?? '', title, browser].join('|'),
```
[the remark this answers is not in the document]
dmaretskyi: Applied in 1cb94866, but the stated rationale isn't right and I don't want the thread to leave a misleading reason in the record.

Two nightlies on one commit would **not** have collapsed. `ci-event.mjs` seeds the uuid from `(event, commitSha, dedup)`, and PostHog deduplicates on the tuple *(uuid, timestamp, event, distinct_id)* — its own comment says so explicitly. These rows are dated by the run (`stats.startTime`), not the commit, so the timestamps already differed and both nights were retained. That is the behaviour `DASHBOARD.md` documents.

I took the change anyway, for a different reason: that guarantee rests entirely on this event staying dated by the run. If someone later switched it to the commit date — which is `ci-event.mjs`'s default, so it's a natural thing to "fix" — two nights would silently become one point, with nothing on the dashboard to show it. Putting the run in the key means uuid uniqueness no longer depends on that choice. Re-publishing a run stays idempotent, since the timestamp is the report's own `stats.startTime` rather than the moment of publishing.

Added a regression test asserting both halves: same run → same dedup key, different run → different.

---

packages/apps/composer-app/src/playwright/globals.d.ts
```diff
@@ -6,6 +6,9 @@
 // casting at every `page.evaluate` site. Everything here is optional and framework-internal: these
 // hooks appear only once the app has mounted (and `composer.manager` only after React does), so the
 // call sites still guard — the declaration buys type-checking, not a presence guarantee.
+//
```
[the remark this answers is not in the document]
dmaretskyi: Half applied in `77e7df03`.

Applied: the terminal period on `collaboration.spec.ts:100`.

Declined: the three "empty separator comments" are not empty comments. In each case the bare `//` sits *inside* a multi-line `//` block and separates two paragraphs of one comment — the same shape as this repo's file header (`//` / `// Copyright … DXOS.org` / `//`). Removing them would run two paragraphs together rather than fix anything.

---

packages/e2e/composer-e2e/scripts/publish-results.mjs
```diff
@@ -0,0 +1,118 @@
...
+    cwd: WORKSPACE_ROOT,
+    encoding: 'utf8',
+    stdio: 'inherit',
+  });
+  if (result.status !== 0) {
+    warn(`publishing e2e results to PostHog failed (exit ${result.status ?? 'signal'}); the run itself is unaffected.`);
+  }
+};
+
+main();
```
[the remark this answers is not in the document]
dmaretskyi: Correct and fixed in `77e7df03` — `parseArgs`, `mkdirSync` and `writeFileSync` all sit outside `main`'s own `try`, so under `always()` a throw there would fail a run this step only observes. Wrapped the call in a `try` that routes through `warn`.

---

.depot/workflows/check.yml
```diff
@@ -577,6 +583,30 @@ jobs:
           run: >-
             moon exec --on-failure continue ${{ steps.shard.outputs.targets }}
 
+      # Trends every test's outcome in PostHog (`ci.e2e-test`), which the "Composer e2e" dashboard
```
wittjosiah: there's a lot of comment slop is this, cleanup before landing

packages/apps/composer-app/src/playwright/startup.spec.ts
wittjosiah, line 1: why are these not included in the e2e tests? would these not be useful results to track over time as well?

packages/e2e/composer-e2e/BEST-PRACTICES.md
wittjosiah, line 1: should these go in a skill?

packages/apps/composer-app/spec/APP.mdl
```diff
@@ -450,6 +464,200 @@ test QA-4: The assistant edits the document
           $snapshot.layout.workspace === `root/${$given.defaultSpace.id}` && $stepErrors.length === 0;
 ```
 
+```mdl
+test QA-5: First run and logout
```
wittjosiah, line 468: 1. is the expectation that you edit these specs instead of the tests themselves?
2. how do you regenerate the tests if you edit the specs?
3. why are the test specs here and the tests in another package?

dmaretskyi: 1. yes
2. agent does it
3. good question, I'm conflicted on this. I see the benefit of putting them together with tests, but also this being the extension of the app spec makes sense too: the test flows declared in the spec can inform how the app is built
4. there's also a skill which can record a video demo based on this spec

packages/e2e/composer-e2e/BEST-PRACTICES.md
dmaretskyi: I put it locally since it currntly only applies to composer testing, but maybe in the future

.depot/workflows/check.yml
dmaretskyi: Fair, and done in `d2c1f154`. The publish step had 15 lines of prose for a one-line condition, all of which `DASHBOARD.md` already says; it's now four lines — the pointer plus the two constraints you can't read off the YAML (`always()` is deliberate, and why it's the `composer` shard only). Trimmed the `vitest.config.ts` header and the `DX_E2E_RUN_ID` note the same way.

---

packages/apps/composer-app/src/playwright/startup.spec.ts
dmaretskyi: Because they measure rather than assert, and the two want opposite things from CI.

`startup.spec.ts` and `perf-*.spec.ts` don't pass or fail on behaviour — they record numbers (`appendRunSample`, `appendBenchmarkRow`, `writeReport`) that a separate gate then judges: `check-startup-budget` re-runs the warm-cold scenario with `--repeat-each=5` and gates on the **median**, and the perf nightly tracks the series. That's why they stayed behind: they're coupled to the build and the runner, not to the app's behaviour. `e2e-dev` needs `vite serve` rather than a bundle, and `e2e-welcome-focus` drives Storybook, so neither can run under the `composer-e2e` config at all.

Keeping them out is what lets the behavioural tier gate a merge. A timing test on a shared CI runner is flaky by construction, which is exactly why `check-startup-budget` is still manual — one of those in the gated suite and the pass/fail signal stops meaning anything.

On tracking them over time: yes, worth it, and I'd argue they deserve **their own** trend rather than a slot in this one. The stacked chart counts pass/fail/skip; a startup measurement has no such outcome — what you want plotted is milliseconds and module counts with a budget line, which is a different tile and a different event shape (`ci.e2e-test` has `ciStatus`/`ciDurationMs`, not `modulesAtReady`). The publisher is generic — `scripts/ci-event.mjs` already carries every other CI trend in the repo — so it's a small follow-up: emit `ci.startup` from the existing report and add a tile. Happy to do it in a follow-up PR; I don't think it should widen this one, since it needs its own event contract to be worth anything.

---

packages/apps/composer-app/spec/APP.mdl
wittjosiah: I think the spec part is the thing I'm most concerned about. I don't find "agent does it" to be a satisfactory answer to how you get from spec to tests, it all feels a bit ill-defined and lacking a process of how to manage it.

for example: its not clear that the tests are intended to be "generated artifacts", they'll end up being edited directly by people and/or agents and drift from the spec. Then if you re-gen from the spec it will break the tests because the spec wasn't actually the source of truth

dmaretskyi: The drift failure you're describing depends on the tests being generated artifacts, and they aren't — nothing regenerates a test from the spec, and nothing in this PR claims to. There's no generator, no `re-gen` step, and no codegen language anywhere in `qa.mdl`, `BEST-PRACTICES.md` or the skill. `automated:` is a **link between two hand-written things**, not an input to a generator.

That's what makes the direction question moot for correctness: both sides are source, so neither can "break" the other by being edited. What the link buys is that drift can't land silently. `check-qa-coverage.mjs` runs in the `check` job and fails on all seven ways the two can diverge:

- a test carries no `@QA` tag at all;
- a test tags a prefix with no registered `.mdl`;
- an `automated:` entry names a test that doesn't exist (so a rename or deletion is a red build, not silent rot);
- an entry names a test that exists but doesn't declare *that* flow's tag (so re-pointing a test from QA-1 to QA-2 fails until QA-1 drops it);
- an entry names a test that is `test.skip`, including inherited `describe.skip`;
- a flow has no `automated:` block while specs tag it;
- a flow's `automated:` omits a test that tags it.

So the "spec wasn't actually the source of truth" scenario surfaces as a failing `check` the moment either side moves, which is the opposite of drift — the duplication is deliberate, precisely because a one-sided link rots quietly.

Where I think you're right, and the PR doesn't answer it: nothing here defines **who writes which side first**, or what to do when a test legitimately needs to change shape. Today the checker enforces that they agree, not how they got there — an agent or a person can start from either end, and the process for "the flow was wrong, update both" is convention rather than tooling. If that's the gap worth closing, it's a documented workflow in the skill (flow first, test second, both in one change), not a generator — and I'd rather write that down than leave it implied. Happy to add it to this PR if you want it before landing.

The placement question — spec beside the tests vs. as an extension of `APP.mdl` — is @dmaretskyi's call; he's said he's genuinely torn, so I'll leave that one to him.

---

wittjosiah: none of those checks check the behaviour of the test, just it existing and being associated properly

dmaretskyi: Correct, and my docs were overclaiming — fixed in `a92dcbc1`.

The check is **referential, not semantic**: it proves the two sides name each other and that the named test actually runs. It cannot read a spec and decide whether it exercises the flow's `steps`, and a test tagged `@QA-1` that asserts nothing passes it clean. `qa.mdl` said "coverage is a claim about the `steps`", which reads as if that claim were verified; it isn't, and both it and `BEST-PRACTICES.md` now say so and put the judgement on the reviewer.

Worth separating the two arms, because the behaviour *is* checked against the spec — just not by this link:

- The flow's steps carry `invoke:` and `assert:` (`assert: return $snapshot.spaces.some(…)`), and the `composer-qa` skill executes them against a live app, judging each step by the UI snapshot. That's the arm where the spec's prose is actually executed and where its assertions have teeth.
- The Playwright spec is the unattended arm — it runs on every CI cell, and what it asserts is whatever it was written to assert.

Nothing cross-checks those two against each other, and short of deriving one from the other I don't think anything can — which lands back on your original question. I'd argue that's the honest boundary rather than a gap to close: the value on offer is that every test points at a documented intent and the pointer can't rot silently, not that the pointer is proof of coverage.

One correction to my previous reply: I said nothing defines who writes which side first, and that was wrong — `BEST-PRACTICES.md` §1 already says **"Write the flow first"**, with the reasoning. What's genuinely undefined is the update path when a test legitimately changes shape, which is the part I'd still be happy to write down before landing if you want it.

---

PR #13257, opened 2026-09-21
edge-client: buffer messages sent while the websocket is connecting

dmaretskyi:
## Root cause

`EdgeWsConnection` extends `Resource`, and `_open()` only constructs the `WebSocket` — it returns while the socket is still in `CONNECTING`, so `isOpen` is true before the handshake completes. `EdgeClient._createConnection` assigns `this._currentConnection = connection` immediately after construction, ahead of `await connection.open()` and ahead of the `_ready` trigger woken from `onopen`. Any caller holding the connection across a reconnect window (swarm `offer` / `initiateConnection` via messenger) therefore reaches `EdgeWsConnection.send()` while the socket is still connecting, and `send()` wrote straight through with no `readyState` gate:

- V0 path: `this._ws.send(binary)` — the DOM spec makes `send()` throw `InvalidStateError` in exactly this one state, which is the ~57-event error family in preview.
- Muxer path: `WebSocketMuxer.send()` calls `this._ws.send(...)` for small messages, and `_sendChunkedMessages` guarded only `CLOSING`/`CLOSED`, never `CONNECTING`. The throw lands in an async function whose only handler is `.catch((e) => log.catch(e))`, so the message is lost silently rather than loudly.

## How it was reproduced

`FakeWebSocket` in `edge-ws-connection.test.ts` now starts at `readyState = 0` and throws `InvalidStateError` from `send()`, matching browser behaviour; the test helper sets `readyState = 1` before dispatching `onopen`, as a real socket does. A new parameterised test sends a message while the socket is connecting and asserts it neither throws nor is dropped, and that it reaches `ws.sent` after `onopen`.

Before the fix, both parameterisations failed — V0 with `Error: Failed to execute 'send' on 'WebSocket': Still in CONNECTING state.`, muxer with `expected [] to have a length of 1 but got +0` (the silent-drop signature). After the fix all 4 tests in the file pass, along with the full `edge-client` suite (81 passed / 4 skipped) and `messaging` (20 passed / 1 skipped).

## The fix

- `EdgeWsConnection.send()` enqueues into a pending queue when `readyState === CONNECTING` and returns; `onopen` flushes the queue in submission order **before** invoking `onConnected()`, so callers reacting to connection see an already-drained socket. The queue is cleared in `_close()`, on `onclose`, and when the socket connects after the resource became inactive.
- The bound is 256 messages, dropping the **oldest** on overflow with a warning. A handshake is one round trip, so the real backlog is a handful of messages; the cap only exists so a server that accepts a socket and never completes the handshake cannot grow the queue without limit. Dropping the oldest rather than the newest is deliberate: these are signalling messages, and during a reconnect the freshest state is the useful one — rejecting the newest would keep a stale backlog and discard the offer that matters.
- `WebSocketMuxer._sendChunkedMessages` now treats `CONNECTING` as backpressure rather than an error: the deferred sender reschedules itself on the existing buffer-full backoff instead of throwing, so segmented sends resume once the socket opens.

No try/catch at the crash site, and no `readyState` check bolted onto the callers — the connection owns the pre-open window.

closes DX-1278

---

---
🚀 Composer preview: https://pr-13257-composer-dev.dxos.workers.dev


dmaretskyi (commented):
land

PR #13260, opened 2026-09-21
echo: retire a closed replicator connection before it is promoted

dmaretskyi:
## Root cause

`MeshEchoReplicator` keeps one list of `MeshReplicatorConnection`s per peer. The first connection is enabled and handed to `EchoNetworkAdapter`; any further teleport session with the same peer queues behind it, disabled, and is promoted to enabled when the enabled one closes.

`MeshReplicatorConnection` only acted on its extension's `onClose` when the connection was *enabled* (`_disconnectIfEnabled`). A queued, disabled connection whose teleport session died was therefore never retired: it stayed in the per-peer list holding an `AutomergeReplicator` whose `_destroy()` had already set `_destroyed = true`. When the enabled connection later closed, `onRemoteDisconnected` promoted that dead connection — `onConnectionOpen` registered it in the adapter's peer map, and from then on every write to that peer reached a destroyed extension:

`queryCollectionState` → `_send` → `writer.write` → `MeshReplicatorConnection` sink → `AutomergeReplicator.sendSyncMessage` → `invariant(!this._destroyed)`.

Collection-state queries are fire-and-forget, so the throw surfaced as the ~37 unhandled `invariant violation [!this._destroyed]` events in the ticket, logged out of `echo-network-adapter.ts` when the errored stream's `abort()`/`cancel()` replayed the stored error.

## How it was reproduced

`packages/core/echo/echo-host/src/automerge/echo-network-adapter.test.ts` gained a test that drives the real lifecycle — two teleport sessions with one peer, each backed by a real `AutomergeReplicator`; the queued session's extension is closed first, then the enabled one, and the adapter is then asked to send.

Before the fix (`moon run echo-host:test -- src/automerge/echo-network-adapter.test.ts`):

```
 × a connection that closed while disabled is not promoted to the peer (DX-1279) 6ms
   error: InvariantViolation [Error]: invariant violation [!this._destroyed] at packages/core/mesh/teleport-extension-automerge-replicator/src/automerge-replicator.ts:140
       at AutomergeReplicator.sendSyncMessage (…/automerge-replicator.ts:140:5)
 FAIL  src/automerge/echo-network-adapter.test.ts > a connection that closed while disabled is not promoted to the peer (DX-1279)
 AssertionError: expected false to be true
   200|     }
   201|     expect(PeerNotFoundError.is(error)).to.be.true;

 Test Files  1 failed (1)
      Tests  1 failed | 10 passed (11)
```

After the fix:

```
 ✓ |node| src/automerge/echo-network-adapter.test.ts (11 tests) 89ms

 Test Files  1 passed (1)
      Tests  11 passed (11)
```

`teleport-extension-automerge-replicator:test` (5 passed), the full `echo-host:test` suite and `client-services:test` were also run; the two failures seen there (`automerge-repo-subduction.test.ts` "authorizeFetch permissive on server allows fetcher", `invitations-handler.test.ts` "invitation success after first guest error") are unrelated timeouts under parallel load and pass on rerun in isolation — neither touches the changed modules.

## The fix

- `MeshReplicatorConnection` records that its extension closed and retires the connection through `onRemoteDisconnected` regardless of whether it was enabled, so a queued connection leaves the per-peer list as soon as it dies.
- `MeshEchoReplicator` filters closed connections out of the per-peer list and never promotes one, so the adapter is only ever handed a peer it can write to.
- The connection's writable sink drops a message for a closed or disabled connection instead of asserting, so a send racing a teardown is a no-op rather than a rejection thrown at a fire-and-forget caller. A send to a peer the adapter no longer knows remains `PeerNotFoundError`, which `CollectionSynchronizer` already handles.

closes DX-1279

---

---
🚀 Composer preview: https://pr-13260-composer-dev.dxos.workers.dev


dmaretskyi (commented):
land

PR #13261, opened 2026-09-21
client-services: build the stack from LayerSpecs

dmaretskyi:
Builds the client-services stack from `LayerSpec`s aggregated by a `LayerStack`, instead of the hand-composed `provideMerge` chain in `ServiceStack`.

## The specs

Each part of the stack is its own top-level spec declaring the tags it requires and the tags it provides, so build order — and which specs are built at all — follows from the graph:

```ts
export const EchoHostSpec = (options: ServiceStackServices) =>
  LayerSpec.make(
    { affinity: 'application', requires: [IdentityManagerService, SpaceManagerService, SqlClient.SqlClient], provides: [EchoHostService] },
    () => echoHostLayer({ useSubduction: options.edgeFeatures?.subductionReplicator }),
  );
```

`clientServiceSpecs(options)` is the list of them with the option-driven choices applied. Two conventions carry what the chain used to encode:

- A spec whose point is a side effect provides no tag for anyone to request — a lifecycle subscription, an rpc registration, a replicator attaching itself to the echo host — so it is `eager`.
- Conditionality lives in the spec list: a spec that needs the edge requires the edge tags, and one that cannot be built is pruned rather than failing.

The platform inputs (edge clients, signal manager, transport factory) are specs too, which is what lets the stack declare its ambient requirements as tags: config, the hook controller and the two SQL services, all of which are always there.

## What builds it

`layerClientServices` is a plain `Layer<LayerStack.Service | ServiceResolver, never, …>`, and every embedder composes it — the worker runtime, `LocalClientServices`, and the test `ServiceContext`. They resolve services through the stack's `ServiceResolver` rather than holding a context of everything, so adding a service touches no embedder. `ServiceStack`, `client-services-layer.ts` and `ClientPlatformLayer` are gone.

## `compute` / `compute-runtime`

- `LayerSpec.eager` — build the spec when its slice initializes rather than when one of its tags is requested.
- `LayerStack.init` — initialize a slice without asking for a tag, which is how a stack whose point is its side effects starts.
- `LayerStack.layer({ layers, services: Tag[] })` — ambient services declared as tags, so a missing one is a compile error rather than a silently pruned spec; provides the `ServiceResolver` alongside the stack.
- Ambient `services` on the stack, and batch runtimes disposed newest first.
- The module's surface is effects: `destroy` returns an `Effect`, so callers compose it rather than wrapping a promise.

## Bugs this surfaced

- **Replicators never reached `DataSpaceManager`.** It reads both replicator tags with `Effect.serviceOption`; the first spec version built them into `registerReplicator` and provided nothing, so enabled replication was silently skipped.
- **Subduction enabled without an endpoint pruned six rpc services.** `memoryConfig` sets `subductionReplicator` with no edge URL, so requiring the edge replicator on the feature flag alone pruned `DataSpaceManager` and everything built on it.
- **`@dxos/schema` dev-depended on `@dxos/react-client`** for a single story, which closed a cycle preventing anything under `client-services` from depending on `@dxos/compute`. The story moved to `@dxos/react-client`.

## Testing

`compute-runtime:test` 240 passed, `client-services:test` 220 passed / 3 skipped, `client:test` 42 passed. `client-services`, `client`, `client-e2e`, `composer-app` and `app-framework` build; lint, `oxfmt --check` and `pnpm knip` clean.

https://claude.ai/code/session_01MTDp7Kb4Yh4t7tiaqmRoC3


packages/sdk/client-services/src/packlets/services/layer-specs.ts
```diff
@@ -0,0 +1,554 @@
...
+    {
+      affinity: application,
+      requires: [EchoHostService, Hook.Controller, EdgeConnectionService, EdgeHttpClientService],
+      provides: [],
+      eager: true,
+    },
+    () =>
+      options.edgeFeatures?.subductionReplicator
+        ? registerReplicator(EdgeAutomergeReplicatorService).pipe(Layer.provide(EchoEdgeSubductionReplicatorLayer()))
+        : Layer.empty,
```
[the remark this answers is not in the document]
dmaretskyi: Confirmed and fixed in 57e8eca5 — thanks, this was a real regression against the chain these specs replace. `meshReplicatorLayer` used `Layer.provideMerge`, so the tag stayed in context for `data-space-manager.ts:1220-1221` to read; my spec used `Layer.provide` with `provides: []`, so it did not.

Each replicator is now a provider spec (`provides: [MeshEchoReplicatorService]` / `[EdgeAutomergeReplicatorService]`) with a separate eager registration spec requiring that tag, and the `DataSpaceManagerService` spec requires whichever tags the options enable — so a disabled replicator is modelled by the same pruning that already models edge absence, rather than by a null.

`client-services:build`, `pnpm knip` and `client-services:test` (220 passed, 3 skipped) are clean.

---

packages/core/compute/compute-runtime/src/LayerStack.ts
dmaretskyi, line 880: add a type-safe layer api for layer stack

LayerStack.layer = (opts: {...., services: Tag[] }): Layer<LayerStack, never, Tag>

dmaretskyi: Added in 0a43bdfe:

```ts
export const layer = <const Tags extends readonly Context.Key<any, any>[]>(opts: {
  readonly layers: LayerSpec.LayerSpec[];
  readonly services: Tags;
}): Layer.Layer<Service, never, Context.Service.Identifier<Tags[number]>>
```

`LayerStack.Service` is the tag for the built stack; the layer picks the declared tags out of its context (`Context.pick`) so the stack sees exactly what the type promised, and destroys the stack when the layer's scope closes. Two tests cover it: ambient services arriving through the layer context, and teardown on scope close.

The type safety is the point — an ambient requirement that goes missing is currently invisible, since a spec whose requirement is absent is pruned rather than failing. This PR has one bug that was exactly that shape (subduction enabled in config with no edge endpoint pruned `DataSpaceManager` and the six rpc services built on it), so the compile error is worth having.

`makeClientServicesStack` does not use it yet: its ambient set includes the edge clients, which exist only with a configured endpoint, and `services: Tag[]` is all-or-nothing. Making those specs rather than ambient services would fix that — happy to do it here or as a follow-up, whichever you prefer.

---

packages/sdk/client-services/src/packlets/services/client-services-stack.ts
```diff
@@ -63,45 +56,111 @@ export const runtimePropsFromConfig = (
...
- * beneath it. Build it with `ManagedRuntime`, then emit `Opening` and `StackOpened` to boot; disposing the
- * runtime tears everything down in reverse.
+ * The running client services stack: the {@link clientServiceSpecs} graph aggregated by a
+ * {@link LayerStack}, over the ambient services the embedder supplies.
+ *
+ * Specs are built on demand — resolving a tag builds what that tag needs and nothing else — except
+ * the `eager` ones (rpc registrations, lifecycle subscriptions, replicators), which the slice builds
+ * as soon as it initialises, since nothing would ever ask for them.
+ */
+export class ClientServicesStack {
```
dmaretskyi: I dont like that this class needs to exist.

1. surroudiong code is effect. doing pomises here just makes it messy with conversions
2. using LayerStack's new layer api would make it clean compositional fit

packages/sdk/client-services/src/packlets/services/layer-specs.ts
```diff
@@ -0,0 +1,712 @@
...
+  echoHostLayer,
+  identityProviderLayer,
+  registerReplicator,
+  storageLifecycleLayer,
+  storageMigrationLayer,
+} from './service-stack.ts';
+import { HypercoreStorageDirectoryLayer, SqliteStorageLayer, SqliteStorageService } from './sqlite-storage.ts';
+import { StackReadinessLayer, StackReadinessService } from './stack-readiness.ts';
+
+const application = 'application' as const;
```
dmaretskyi: useless const, just use the string everywhere

packages/core/compute/compute-runtime/src/LayerStack.ts
```diff
@@ -26,15 +26,67 @@ import { LayerDependencyCycleError } from './errors.ts';
...
+export class Service extends Context.Service<Service, LayerStack>()('@dxos/compute-runtime/LayerStack') {}
+
+/**
+ * A {@link LayerStack} whose ambient services are declared as tags rather than handed over as an
+ * opaque {@link Context.Context}: the returned layer requires exactly those tags, so an embedder
+ * that fails to provide one is a compile error rather than a spec silently pruned at runtime.
+ *
+ * The stack is destroyed when the layer's scope closes.
+ */
+export const layer = <const Tags extends readonly Context.Key<any, any>[]>(opts: {
```
dmaretskyi, line 53: make it also provide service resolver tag

packages/core/compute/compute-runtime/src/LayerStack.ts
```diff
@@ -714,9 +795,15 @@ class Slice {
     );
   }
 
+  /**
+   * Disposes the batch runtimes newest first: a batch materialized later may hold services from an
+   * earlier one, and Effect only orders finalizers within a single runtime.
+   */
   async destroy() {
```
dmaretskyi: covert all apis in this module to be effect, not promises, update consumers

packages/core/compute/compute-runtime/src/LayerStack.ts
dmaretskyi: Done in f608bacc — `LayerStack.layer` returns `Layer<Service | ServiceResolver.ServiceResolver, ...>` and provides both from one `Layer.effectContext`, so a consumer can `ServiceResolver.resolve(tag, {})` without reaching through the stack. The client-services test builder stops threading a resolver by hand (`Effect.provide(stackContext)` now), and there is a test covering resolution through the layer.

---

packages/core/compute/compute-runtime/src/LayerStack.ts
dmaretskyi: Done in f608bacc. `LayerStack.destroy` and `Slice.destroy` return `Effect<void>` rather than promises, so the module's surface is effects throughout.

Consumers compose them instead of wrapping: the layer's finalizer is `Effect.addFinalizer(() => stack.destroy())`, the slice teardown loop yields directly, and `app-framework`'s process-manager finalizer drops its `Effect.promise` wrapper (`Effect.andThen(layerStack.destroy())`). The only remaining `Effect.promise` inside the module is around `ManagedRuntime.dispose()`, which is promise-returning upstream.

`compute-runtime` 240 passed, `client-services` 220 passed, `client` 42 passed; builds, lint and knip clean.

---

PR #13262, opened 2026-09-21
errors: report a dismissed passkey prompt below error severity

dmaretskyi:
## Root cause

WebAuthn reports a dismissed passkey prompt as a `NotAllowedError` `DOMException` — its normal "the user cancelled" signal, also used for a timeout — and an aborted `AbortSignal` as `AbortError`. Neither is a defect, but both reached the production error stream at ERROR severity from two places:

- `ProcessHandle.ts` — `logFailure` is the single point every FAILED process transition passes through, and it reported unconditionally at `log.error`. The `RedeemPasskey` operation runs as a process, so a dismissal crashed the process and logged as a crash (9 of the 14 production events).
- `WelcomeScreen.tsx` — the passkey handler called `log.catch(redeemError)` before classifying it, so even a failure it then rendered as "dismissed" had already been logged at error (5 events).

Together that was roughly half of Composer's remaining error-severity production logs, which hid real regressions.

A typed error for this already existed and was already being raised: `PasskeyError.Dismissed` (`PasskeyDismissedError`), produced by `PasskeyError.fromAssertion` on exactly these two `DOMException` names. Neither call site consulted it when choosing a severity — the welcome screen used it only to pick the user-facing message, after logging; the process runtime is a layer below `plugin-client` and could not reference it at all. So the discrimination existed and the reporting simply ignored it; nothing raises `PasskeyDismissedError` at error severity at the point it is constructed.

## How it was reproduced

Both changes are covered by tests asserting the *level* a dismissal reports at, not merely the state it leaves behind — reusing the `captureLogEntries` helper added for DX-1250.

`packages/core/compute/compute-runtime/src/ProcessManager.test.ts` — spawns a process that fails with a domain error wrapping a `NotAllowedError`, an `AbortError`, a `NotSupportedError` (control) and a marker-carrying domain error, and asserts the lifecycle outcome levels. Before the fix (14 = ERROR, 12 = INFO):

```
 FAIL  src/ProcessManager.test.ts > … > a user-dismissed prompt reports below error level (DX-1281)
AssertionError: expected 14 to deeply equal 12
 FAIL  src/ProcessManager.test.ts > … > an aborted ceremony reports below error level, other DOMExceptions do not (DX-1281)
AssertionError: expected [ 14, 14 ] to deeply equal [ 12, 14 ]
 FAIL  src/ProcessManager.test.ts > … > a domain error that declares itself a cancellation reports below error level (DX-1281)
AssertionError: expected [ 14 ] to deeply equal [ 12 ]

 Test Files  1 failed | 1 passed (2)
      Tests  3 failed | 76 passed (79)
```

`packages/plugins/plugin-client/src/types/PasskeyError.test.ts` — 5 further failures before the fix, covering the welcome screen's path.

After the fix:

```
 Test Files  2 passed (2)
      Tests  79 passed | 1 skipped (80)        # compute-runtime ProcessManager
 ✓ src/types/PasskeyError.test.ts (11 tests)   # plugin-client
```

Full suites green for the changed packages and their dependents: `errors`, `compute-runtime` (21 files / 229 tests), `plugin-client` (12 files / 28 tests), `plugin-onboarding` (2 files / 9 tests). `lint` clean for all four; `pnpm format` run.

## The fix

- `@dxos/errors` gains `isCancellation`, which walks an error's `cause` chain (depth-bounded, self-reference-safe) and matches either a `DOMException` named exactly `NotAllowedError` or `AbortError`, or the new `Cancellation` marker interface. The chain walk is what makes it usable from a layer below the domain error: the `DOMException` sits on the wrapper's `cause`. No other `DOMException` matches, and nothing is swallowed — only the severity changes.
- `PasskeyError.Dismissed` now implements `Cancellation`, so the existing typed error is what the generic reporter recognises, including on the paths where the platform reported a dismissal without a `DOMException` (the native Tauri bridge, a null credential). `classify` also accepts a raw cancellation, so a `NotAllowedError` that reaches it unwrapped is no longer reported as a generic failure.
- `ProcessHandle.logFailure` reports a cancelled process at `log.info` under `lifecycle: cancelled`; every other failure keeps `lifecycle: failed` at `log.error`, with the same fields.
- `WelcomeScreen` calls the new `PasskeyError.report`, which logs at `info` for a dismissal and `log.catch` otherwise, then returns the same `Failure` the UI already switched on.

No UI behaviour changes: the same error state is set and the same message rendered in every case.

closes DX-1281

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Dismissing or cancelling a passkey prompt is now recorded as an informational cancellation instead of an error.
  * Genuine passkey failures continue to be reported as errors.
  * Cancellation detection now covers aborted ceremonies, dismissed prompts, and cancellations wrapped within related errors.
  * No changes were made to passkey or onboarding UI behavior.

---
🚀 Composer preview: https://pr-13262-composer-dev.dxos.workers.dev


dmaretskyi (commented):
land

PR #13265, opened 2026-09-21
plugin-mobile: don't assume the scroll event target is an element

dmaretskyi:
## Root cause

`useAutoScroll` in `MobileLayout.tsx` registers a capture-phase `scroll` listener on `document` for debug logging and treats every `event.target` as an `HTMLElement`:

```ts
const el = event.target as HTMLElement;
if (el === document.documentElement || el === document.body) { return; }
dbg(`scroll: ${el.tagName}.${Array.from(el.classList)...}`);
```

Scrolling the viewport dispatches a `scroll` event whose target is the `Document`, not an element. The guard only excluded `documentElement` and `body`, so the `Document` fell through, `document.classList` is `undefined`, and `Array.from(undefined)` threw `TypeError: Array.from requires an array-like object - not null or undefined`. The throw happens inside a listener installed by the layout root's effect, taking down the whole mobile shell. The unchecked `as HTMLElement` cast is what let a non-`Element` target reach `classList` in the first place.

This empty value is a legitimate state — viewport scroll is already handled by the window scroll reset — so the handler now ignores it rather than logging a bogus entry; nothing renders differently.

## How it was reproduced

The target description was extracted into `scroll-target.ts` (`describeScrollTarget`) and covered by `scroll-target.test.ts`, which attaches a capture-phase listener on `document` and dispatches a real `scroll` event on `document` (happy-dom), exactly as the viewport does.

Against the pre-fix logic:

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
 ❯ describeScrollTarget src/components/MobileLayout/scroll-target.ts:12:33
     12|   return `${el.tagName}.${Array.from(el.classList).slice(0, 2).join('.…
 ❯ HTMLDocument.listener src/components/MobileLayout/scroll-target.test.ts:12:54
 Tests  1 failed | 2 passed (3)
```

After the fix:

```
 ✓ |node| src/components/MobileLayout/scroll-target.test.ts (3 tests) 6ms
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

## The fix

- `describeScrollTarget(target: EventTarget | null)` narrows with `target instanceof Element` before touching `classList`, and returns `undefined` for the `Document` and the scrolling roots — no cast, no `[]` fallback.
- `useAutoScroll` calls it and only logs when there is something to report.
- `plugin-mobile` node tests now run under `happy-dom` so DOM behaviour is testable.

`@dxos/plugin-mobile` is private, so no changeset.

closes DX-1291

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile debug scroll logging to ignore viewport-level scroll events and focus on meaningful scrollable elements.
  * Scroll details now include the element type, select classes, and rounded scroll position for clearer diagnostics.

* **Tests**
  * Added coverage for viewport scrolling and scrollable container descriptions.

---
🚀 Composer preview: https://pr-13265-composer-dev.dxos.workers.dev


dmaretskyi (commented):
land

PR #13266, opened 2026-09-21
ui-editor: cancel the pending-text busy update on teardown

dmaretskyi:
## Root cause

`pendingText()` flags the editor busy while a pending (streaming) session is open. It did so from a bare `EditorView.updateListener`, deferring the dispatch to a `queueMicrotask` "to avoid dispatching mid-update". A plain update listener has no lifetime: the microtask it queues survives a reconfigure that drops the extension. When the editor is reconfigured while a session is open — the assistant chat prompt rebuilds its extension list, the transcription plugin's extension provider comes and goes — the microtask then ran `update.view.state.field(pendingTextState)` against a configuration that no longer contains the field, and CodeMirror threw `RangeError: Field is not present in this state`. On the token-streaming path that throw lands mid-response.

Every other read of the field (`commitPending`, `cancelPending`, the plugin call sites) already passed `false`; this deferred one was the only unguarded read that could outlive its configuration, which matches the stack in DX-1290.

## How it was reproduced

`packages/ui/ui-editor/src/extensions/streaming/pending/pending-text.test.ts` — "tearing the extension down cancels the queued busy update": open a pending session (which queues the busy microtask), reconfigure the view to drop the extension in the same tick, then flush microtasks and assert nothing was thrown. Before the fix:

```
 FAIL  src/extensions/streaming/pending/pending-text.test.ts > pendingText extension > tearing the extension down cancels the queued busy update
AssertionError: expected [ …(1) ] to deeply equal []

- []
+ [
+   RangeError {
+     "message": "Field is not present in this state",
+   },
+ ]
```

After the fix the whole package is green (`Test Files 37 passed`, `Tests 451 passed`), as is `plugin-transcription` (the other consumer of the extension).

## The fix

Option 2 — cancel the callback rather than read defensively. The busy reporter is now a `ViewPlugin`, so it has the lifetime the deferred work needs: CodeMirror destroys it when the extension leaves the configuration, and `destroy()` marks it so any queued microtask abandons its dispatch. The read inside the microtask stays strict, so a genuine lifecycle bug would still surface rather than being silently swallowed. The one defensive read added is on `update.startState`, where the previous configuration legitimately may not have carried the field.

closes DX-1290

---

---
🚀 Composer preview: https://pr-13266-composer-dev.dxos.workers.dev


dmaretskyi (commented):
land

dmaretskyi (commented):
land

PR #13269, opened 2026-09-21
worker-framework: displace the previous worker before queueing on the storage lock

dmaretskyi:
## Root cause

`Worker.run` runs its entire loop inside the `storageLockKey` Web Lock, and only posts `listening` once that lock is granted. The 15s budget that fails as `opening worker leader session` covers exactly `spawn worker → listening → init → ready`, so anything that keeps a worker from taking the storage lock spends the whole budget.

Displacement is the mechanism that frees that lock: a starting worker broadcasts a stop signal so the previous worker for the same storage lock tears down (`SPEC.mdl` F-3.1). But the broadcast was made *inside* the lock callback — i.e. only after the lock had already been granted. A worker queued behind a live incumbent therefore never sent the signal that would release the lock it was waiting on, and never reached `listening`.

That matters because a previous worker routinely outlives the tab that spawned it from the new leader's point of view: `LeaderSession` tears its worker down on the ex-leader's **main thread**, which is precisely what is unavailable in the cases this ticket lists (a wedged or saturated leader evicted by a steal, a stale mixed-generation worker). Worker-to-worker displacement is the path that does not depend on that main thread, and it was inert.

`packages/sdk/client/src/testing/storage-lock.ts` already does it in the right order (`Worker.displace(...)` before `navigator.locks.request(...)`), which is the same handshake the worker startup path was getting backwards.

A second hang surfaced while verifying the fix: the connect task's port-wait promise was left unresolved when its context was disposed (the timer was cleared, nothing resolved), and `Connection.close()` joins that task — so a tab closing while it waited for a port never finished closing.

## How it was reproduced

Two tests in `packages/sdk/worker-framework/src/Client.test.ts`, both fast when the bug is absent:

1. `a worker starting while a previous one holds the storage lock displaces it` — worker level.
2. `a leader session opens against a storage lock a previous worker still holds` — connection level, and it fails with the exact reported error.

Before the fix:

```
 × a worker starting while a previous one holds the storage lock displaces it 5018ms
 × a leader session opens against a storage lock a previous worker still holds 18006ms

WorkerConnectionError: Worker connection timed out after 15000ms: opening worker leader session.
    at lockOrRpcTimeoutError (src/internal/locks.ts:8:89)
    at waitWithLockOrRpcTimeout (src/internal/locks.ts:9:113)
    at LeaderSession._open (src/Client.ts:758:30)
```

After the fix, both pass in 37ms, and the full package suite is green in both runtimes:

```
 ✓ |browser-chromium (chromium)| src/Client.test.ts (20 tests)
 Test Files  7 passed (7)
      Tests  40 passed (40)

 Test Files  6 passed (6)
      Tests  36 passed (36)
```

`client:test` (9 files, 42 tests) and `client:build` / `worker-framework:build` are green as well.

## The fix

- `Worker.run` opens the displacement channel, installs its listener and broadcasts the stop signal **before** requesting the storage lock, so the incumbent hears it while the newcomer queues.
- A displace that arrives while this worker is still queued is recorded and checked once the lock is granted: such a worker does not serve, since the worker that displaced it is waiting behind it and would otherwise be stranded for the full leader-session budget.
- The connect task's port-wait promise is now settled on disposal, and the run returns early when its context is disposed — so `close()` cannot hang on a tab that was waiting for a port.

closes DX-1289

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed a client startup hang when a new worker replaces an existing worker.
  - Improved worker handoff when another worker is still holding the storage lock.
  - Fixed connections that could remain open while waiting for a port during shutdown.
  - Ensured pending connection attempts stop promptly when their context is disposed.

- **Tests**
  - Added coverage for worker displacement and recovery from stranded workers.

---
🚀 Composer preview: https://pr-13269-composer-dev.dxos.workers.dev


packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -166,15 +179,19 @@ export const run = ({
...
+    onDisplaced = () => {
+      log('displaced by newer worker, shutting down');
+      void shutdown();
     };
+    if (displaced) {
+      // A newer worker broadcast while this one queued: it is waiting behind us on the storage lock,
+      // and serving here would strand it for the leader session's whole budget.
+      log('displaced while waiting for the storage lock, not serving');
+      channel.close();
+      return;
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in `6fd6eec2`. Correct and worth catching — `channel.close()` alone left the liveness-lock callback awaiting `livenessLockHeld` forever, so a displaced queued worker kept holding `${storageLockKey}/liveness/…`. Clients watch that key to detect termination, so the very worker that stood down would have continued to read as alive; `endpoint.close?.()` was also skipped. Now routed through `await shutdown()`, which releases both locks and closes the endpoint.

Verified before pushing: `worker-framework:test-browser` 40/40, `worker-framework:test` 36/36, lint clean.

---

packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -166,15 +179,21 @@ export const run = ({
...
-    // BroadcastChannel queues a delivery task and never replays it to a listener attached after an
-    // await, which would leave this worker holding the storage lock a displacer is waiting on.
-    channel.onmessage = (event) => {
-      if (event.data?.action === DISPLACE_MESSAGE.action) {
-        log('displaced by newer worker, shutting down');
-        void shutdown();
-      }
+    // Set before the first await, so a displace arriving while this worker starts up cannot land in a
+    // gap and leave it holding the storage lock a displacer is waiting on.
+    onDisplaced = () => {
```
dmaretskyi: does displacement also terminate the displaced worker? the tab that spawned it could do that. that will work in cases where previous worker is stuck.

packages/sdk/worker-framework/src/Worker.ts
```diff
@@ -120,6 +120,23 @@ export const run = ({
   displaceChannel = displaceChannelFor(storageLockKey),
   createRuntime,
 }: Options): void => {
+  // Displacement is a worker-to-worker handshake, so the channel listens and the stop signal goes out
+  // before this worker queues on the storage lock: the incumbent releases that lock only when it is
+  // displaced, and a signal broadcast after the grant can never reach the worker being waited on.
+  const channel = new BroadcastChannel(displaceChannel);
```
dmaretskyi: wrap this is a small utility class with `on<action>` callbacks and methods to post actions

packages/sdk/worker-framework/src/Worker.ts
dmaretskyi: No — displacement is purely cooperative, and you've put your finger on its limit.

**What happens today.** The displaced worker has to be running its event loop to hear the `BroadcastChannel` message and run `shutdown()`. Hard termination exists but lives on the other side: only the tab holding the `Worker` handle can call it, in `LeaderSession.#closeWorker()` → `this.#worker.terminate()` (`Client.ts:777`), reached from `_close()` or a failed handshake.

Those two do chain, for a *responsive* worker: `shutdown()` releases the liveness lock → the ex-leader tab's `navigator.locks.request(livenessLockKey, …)` callback fires (`Client.ts:735`) → `onClose.emit('Dedicated worker terminated.')` → the connection's handler (`Client.ts:315`) drops the session → `_close()` → `terminate()`. So in the orderly case the tab does terminate its worker; it just does so *after* and *because of* the cooperative shutdown.

**Which means the stuck case is exactly the gap.** A worker wedged badly enough not to service the channel never runs `shutdown()`, never releases the liveness lock, so its tab is never told anything is wrong and never reaches `terminate()`. The new worker cannot terminate it either — cross-worker there is no handle, only the channel it is already ignoring. It keeps the storage lock, and we are back to the 15s budget.

**So your suggestion is the right next step, with one caveat.** Having the spawning tab terminate on displacement covers a genuinely different failure class from this PR: the worker is wedged but *its tab is alive*. It does not help when the tab itself is gone or saturated — which is the case this PR addresses, and why worker-to-worker displacement had to work at all.

Concretely it would need a tab-side signal that does not depend on the wedged worker: the displacing worker broadcasts, and the ex-leader **tab** (not its worker) hears it and calls `terminate()`. That needs a channel the tab listens on, plus a decision about how long to wait for a graceful shutdown before terminating — terminating immediately would cut off a worker that was merely slow, mid-flush.

Happy to do it, but it is a design change beyond this PR's scope rather than a fix to it, and this PR is green as it stands. Two options:

1. Land this, and I open a follow-up issue for tab-side termination on displacement.
2. I extend this PR with it now.

Your call — tell me which and I'll do it.

---

packages/sdk/worker-framework/src/Worker.ts
dmaretskyi: Done in `94b7dd87` — `DisplaceChannel`:

```ts
class DisplaceChannel {
  readonly #channel: BroadcastChannel;

  /**
   * Called when another worker posts `stop`. Assignable because what a stop means to this worker
   * changes as it starts up — before the storage lock is granted there is nothing to stand down from.
   */
  onStop: () => void = () => {};

  constructor(channelName: string) { … }
  postStop(): void { … }
  close(): void { … }
}
```

The action-name check against `DISPLACE_MESSAGE` now lives inside the class, so `run` reads as `channel.onStop = …` / `channel.postStop()` and never touches `postMessage` or `event.data`. The exported `displace()` helper goes through the same class rather than building its own `BroadcastChannel`, so the wire format has one definition.

Verified: `worker-framework:test-browser` 40/40, `worker-framework:test` 36/36, `tsc --noEmit` clean, lint clean.

---

packages/sdk/worker-framework/src/Worker.ts
dmaretskyi: land this PR as-is, then repro the case where the worker is stuck on busy cpu loop. and build a 2-nd level mechanism where the non-cooperative worker is killed forcefully if it doesn't wanna displace itself

dmaretskyi:
Both nitpicks from the latest review are declined, with reasons — the PR lands as-is per the author's instruction on the termination thread.

**"Add JSDoc for the changed public functions" (`displace`, `run`) — already present.** Both have had JSDoc since the original commit, covering the exact contracts the finding asks for:

```ts
/**
 * Shuts down whichever worker currently holds `storageLockKey`, so its storage can be taken over.
 * A worker does this to its predecessor on startup; a caller rewriting the storage from outside
 * does it to free the lock, and must still wait for the lock itself — this only asks.
 */
export const displace = …

/**
 * Runs the generic dedicated-worker message loop: storage lock, liveness lock, displacement of prior
 * workers, the init/ready/session protocol, session deduplication, and owner detection.
 *
 * Liveness and displacement are owned by the framework: the worker holds a dedicated liveness Web Lock
 * for its whole lifetime (released on shutdown so clients observe termination), and broadcasts a stop
 * signal on startup so any previous worker for the same storage lock tears down.
 */
export const run = …
```

Nothing to add.

**"Validate displacement payloads with Effect Schema" — declined.** The cited guideline ("Use Effect-TS for schema validation") does not appear in this repo's `.agents/skills/*`, `AGENTS.md`/`CLAUDE.md`, or `.coderabbit.yaml`; I looked before deciding.

On the substance: `BroadcastChannel` is same-origin, and the only writers are `DisplaceChannel.postStop()` and the `displace()` helper, both of which now go through the one class in this file. The payload is a single constant, `{ action: 'stop' }`. The existing check already ignores anything that is not that action, which is the entire validation surface — a decoder would add a dependency and a failure path to reject messages that, if they ever appeared, would mean something inside the same origin was posting to our private channel. That is not a boundary Schema makes safer.

If the message ever grows fields, or the channel becomes something third parties post to, this changes and the decoder earns its place.

---

PR #13273, opened 2026-09-21
worker-framework: recover from hostile callers and browsers instead of wedging

dmaretskyi:
Treats the framework's callers and the browser as hostile — handles that throw, hang or resolve late; workers that die, hang, or cannot build a session; coordinator links that die; leader locks stolen mid-handshake — and fixes what each of those did to the worker connection. Every defect below has a regression test that fails on `main`, and a new seeded model-based walk that found the two worst ones.

## Defects fixed (all in `@dxos/worker-framework`)

| # | Symptom on `main` | Fix |
|---|---|---|
| 1 | **Leader lock stolen while `LeaderSession.open()` is in flight orphans the session and its worker.** The steal path calls `session.close()`, but `Resource.close()` is a no-op on a resource still opening, so the session finishes opening later, nobody closes it, and its worker keeps the **storage lock** forever. Every later leader's worker waits on that lock and never posts `listening`; each handshake times out at 15s, backs off, and repeats. Any tab that blocks its main thread ≥5s during the worker handshake (a second tab opening while the first boots a big profile) triggers it. Found by the model walk. | The lock callback re-checks after `open()` whether it is still the current hold and closes the session if not; per-hold state (`#leaderDone`, `#holdsLeaderLock`, phase) is guarded so a stale callback cannot clobber the next hold's. |
| 2 | **Reconnect after leader/worker death waits behind an in-flight `onConnect`.** `AsyncTask` serialises runs, so the retry queued behind the abandoned attempt for up to the 15s handle timeout; the late handle was then installed over the live one and fired a spurious `reconnected`. | The attempt's context is disposed first and the wait races it; a handle that resolves after abandonment or timeout is closed, never installed. |
| 3 | **Stale handle closed after the replacement connects.** `handleLeaderStopped` rescheduled before `await oldHandle.close()`, so a consumer keeping one handle's resources (e.g. `DedicatedWorkerClientServices.#services`) could have the old `close` null out the new proxy. | Dispose, close the old handle (bounded), then schedule. |
| 4 | **A throwing `onReconnect` callback loops the reconnection with no backoff** (hundreds of worker sessions per second, `reconnected` never fires). | Callbacks are isolated and logged. |
| 5 | **A failing connect handle retries in a tight loop** (measured 153 attempts/s in the test) and never escalates. | Exponential capped jittered backoff on `retryBackoff`; `maxLeaderFailures` consecutive failures fire `onPersistentFailure`. |
| 6 | **A late `provide-port` for an abandoned attempt is accepted.** The worker supersedes that session when it serves the new attempt, so the tab connects through ports nobody serves and hangs another 15s. Plausible whenever the worker is slower than `portTimeout`. | `session`/`provide-port` carry `attempt`; the tab ignores mismatches (optional, compatible with older workers). |
| 7 | **A session the worker cannot build leaves the tab holding dead ports for the full handle timeout.** Ports are posted before the session is built (the build blocks on the tab's side), so a `createSession` failure was invisible to the tab. | New `session-failed` message worker → leader → coordinator (broadcast, so old coordinators relay it) → tab; the attempt fails at once with the decoded error, before or after the ports arrive, even after the handle opened, and retries with backoff. |
| 8 | **A leader steals its own lock.** A leader whose own port request times out with no heartbeat heard (e.g. a half-dead SharedWorker port) passed the stale test — the lock holder was itself — and evicted its own worker, up to the steal budget, restarting a worker other tabs were using. | Never steal while holding the lock or once the coordinator reported its link dead; count the attempt towards the escalation budget instead. |
| 9 | **A dead coordinator link (SharedWorker script 404 after a redeploy removed old chunks) was only logged.** The tab spent the whole 30s boot budget and up to `maxLeaderFailures` steals of a healthy leader's worker before failing. | `WorkerCoordinator.onError` (optional; `Coordinator.SharedWorker` emits it). `open()` rejects with it at once; stealing is suppressed. |
| 10 | **A `LeaderSession` whose worker dies is never closed** — its coordinator listener leaks, one per dead worker. | The catch path closes it instead of the `onClose` handler nulling the field. |
| 11 | **`open()` that times out leaves everything running.** `Resource.close()` is a no-op on a resource that never opened, so the lock, worker, timers and election loop outlived the caller's error handling. | The timeout path tears down; `open()` also rejects at the escalation threshold (leader-session or connect failures) instead of at the end of the 30s budget. |
| 12 | Worker-side: `start-session` while shutting down forked a session from the closing runtime scope; a stolen storage lock surfaced as an unhandled rejection. | Ignored while shutting down; lock loss shuts the worker down. `Worker.run` gains `signal` to terminate an in-process worker the way the browser would. |

Also: a dedicated worker's uncaught error after the handshake is now logged (it was silently dropped).

## Exploratory testing added

- **`src/Client.model.test.ts` — seeded model-based walk.** 2–4 in-process tabs, one recording worker, a hub coordinator. Commands: open/close/reload tab, kill worker, hang worker (buffers messages, keeps locks, then floods), block a tab's main thread (heartbeats stop, inbound queues), break a tab's link (inbound lost), drop `request-port`, delay `provide-port` past the port timeout, fail the next handle, fail the next session. Invariants after convergence: every open tab connected through exactly one handle; one live worker holding its liveness lock; all tabs agree on worker and leader; the worker's live sessions are exactly the open tabs; no leaked coordinator listeners; exactly one leader/storage lock holder. `DX_MODEL_SEED=n` replays a seed; `DX_MODEL_SEEDS`/`DX_MODEL_COMMANDS` soak. Runs 4 seeds in CI (~4s); a 30×40 soak passes.
- **`src/Client.hostile.test.ts`** — one deterministic regression test per defect above.
- **`src/testing/harness.ts`** — the in-process hub (per-tab link modes, held/dropped messages), a pausable/terminable `Worker.run` host, and the recording worker, shared by the three test files (extracted from `Client.test.ts`).

## Gaps not addressed here

- A worker that is CPU-wedged **forever** while its leader tab stays alive is still undetectable: heartbeats come from the tab, so followers never steal, and there is no leader→worker liveness probe. A watchdog would kill legitimate long hydrations (the 18s case in `client-startup-timeout/DESIGN.md`), so it needs a generous or opt-in timeout; left for a follow-up.
- `Connection.closed` is declared but never emitted; `Client._open` in `@dxos/client` subscribes to it. Dedicated-worker mode recovers through `onReconnect` instead, so nothing observes it today.
- `Resource.close()` during a pending `open()` is a no-op (base class); the framework now tears itself down on failed boots, but a caller closing a still-opening connection is still ignored.

## Testing

- `moon run worker-framework:test` — 48 node tests (8 files) pass; the 10 hostile tests and the 4 default model seeds all fail on `main`.
- `moon run worker-framework:test-browser` — Chromium suite passes (52 tests, including the real dedicated-worker + SharedWorker counter suite).
- `moon run client:test -- src/services` — 7 pass, 1 expected fail (pre-existing).
- `DX_MODEL_SEEDS=30 DX_MODEL_COMMANDS=40` soak: 30/30 converge.
- lint, `oxfmt --check`, `tsc --noEmit` clean.

https://claude.ai/code/session_01HqPFZufYyEWNDetj2bxQDS

---
_Generated by [Claude Code](https://claude.ai/code/session_01HqPFZufYyEWNDetj2bxQDS)_

## Summary by CodeRabbit

* **New Features**
  * Improved worker connection recovery during failed, interrupted, or abandoned connection attempts.
  * Added retry backoff and persistent-failure escalation with detailed connection diagnostics.
  * Workers can now shut down when their lock is lost or an abort signal is triggered.
  * Coordinator link failures are detected and reported promptly.

* **Bug Fixes**
  * Prevented stale handles, sessions, and port responses from being installed or leaked.
  * Improved recovery from failed worker sessions and unresponsive coordinator links.
  * Reconnect callbacks that throw no longer interrupt reconnection.

---
🚀 Composer preview: https://pr-13273-composer-dev.dxos.workers.dev


dmaretskyi:
`Check / test (shard=0)` is red on one test that this PR does not reach:

```
FAIL |node| client-e2e src/spaces.test.ts > Spaces > import JSON into existing space
Error: Test timed out in 5000ms.
```

That test builds its client with `createLocalClientServices` (in-process services, no dedicated worker, no coordinator), so none of the `@dxos/worker-framework` changes here are on its path; the shard's other 463 targets passed. It passes locally on this branch (`moon run client-e2e:test -- src/spaces.test.ts -t "import JSON into existing space"`), and `spaces.test.ts` timeouts under CI load are a known pattern (#13099 noted `post and listen to messages` timing out on `main` the same way). No fix for it exists to port yet; retrying the shard once.

---

PR #13288, opened 2026-09-21
echo: evaluate queries in SQLite, behind a flag

wittjosiah:
Adds a second query executor that evaluates a query inside SQLite, and leaves it off. The host compiles a `QueryPlan` into one statement over the index tables and loads no Automerge document to answer it. The in-memory executor is untouched and remains the default, so this lands inert.

Turn it on per host with `EchoHost({ queryExecutor: 'sql' })`, or for a process with `DX_ECHO_QUERY_EXECUTOR=sql`.

## Why

The in-memory executor uses SQLite only to find candidates. Everything after candidate selection runs in JS over documents it loads for the purpose, so a 20-result `order + limit` over 5,000 tasks loads all 5,000 documents, and a cold 1,000-result query on a 10,300-object store cannot complete at all: loading 5,000 documents first exceeds the client's 20 s budget.

## How

`SqlPlanCompiler` (`echo-host/src/query/sql/compile.ts`) turns each plan step into a CTE over the previous one, producing a single `WITH RECURSIVE` statement over `objectMeta`, `objectSnapshot`, `reverseRef` and `ftsIndex`. Transitive deletion, `child-of` and strong-dependency resolution are recursive CTEs; aggregates use window functions at two CTE levels. The index seeks the plans depend on are pinned with `INDEXED BY`, because SQLite's planner has no `ANALYZE` statistics and mis-picks without them, and `explain.test.ts` asserts the plan shape so a planner regression fails a unit test rather than a benchmark.

The object store is the one [#13272](https://github.com/dxos/dxos/pull/13272) just landed. This branch originally added its own body store; that is deleted in favour of `objectSnapshot`.

**Plans the compiled path declines.** `objectSnapshot` drops `@meta` for document rows, so any filter reading it (foreign keys, a registry key or version, tags) cannot be evaluated in SQL. `planReadsObjectMeta` detects those plans and the executor runs them in memory, whatever the flag says. `QueryExecutor.mode` reports the path actually taken.

## Storage

Written on the existing indexing pass whichever executor is selected, so enabling the flag needs no migration of its own:

- `objectMeta` gains `parentId`, `sourceId` and `targetId`, the normalized ids the compiled joins seek on.
- `reverseRef` gains `propPathNormalized`; its cursor name bumps so existing rows re-present once and fill it.
- SQLite 3.45 or newer is required and asserted when the index opens, because `json_type(x, path)` returns NULL for a missing path only from that version.
- While the snapshot store is still filling after upgrade, a query awaits indexing before its first execution. That gate only applies to the compiled path, which reads the store directly.

## Results

N = 5,000 tasks in a 10,300-object store, both executors measured in the same harness (`echo-client-e2e/BENCHMARKS.md`, "Query executor: memory vs sql"):

| row | memory | sql | speedup |
| --- | ---: | ---: | ---: |
| type (5,000 results) | 678.7 ms | 460.6 ms | 1.47× |
| type + property (1,000) | 404.9 ms | 159.2 ms | 2.54× |
| reference traversal (10) | 469.8 ms | 118.3 ms | 3.97× |
| order + limit (20) | 293.8 ms | 19.0 ms | 15.46× |
| reactive first result (1,000) | 359.5 ms | 105.5 ms | 3.41× |
| cold reload + query (1,000) | times out | 9,064.6 ms | |

Host execution per query goes from 130–337 ms to 9–38 ms, with document loads from 2,500–5,000 to zero. The end-to-end gain scales inversely with result size: a 5,000-result query is only 1.5× faster because both paths pay the client's hydration of 5,000 objects, while a 20-result query is 15× faster because the old path still loaded the whole type.

Memory follows the same shape. One cold property query retains 57 MB of heap and 35 MB of RSS on the memory path against 12 MB and 1.6 MB on SQL.

## Testing

| suite | result |
| --- | --- |
| index-core | 67 passed |
| echo-host | 398 passed |
| echo-client | 587 passed |
| echo-client-e2e | 365 passed |
| echo-client-e2e query suite, forced to `sql` | 185 passed |

A differential test (`sql/differential.test.ts`) checks 200 randomly generated filters against the JS matcher. It found a real three-valued-logic bug where a `NULL` predicate dropped rows the matcher kept, fixed by coalescing every type test.

## Behaviour differences under `sql`, deliberate

Ordering by a string property uses code-unit (BINARY) order instead of `localeCompare`. Range comparisons do not coerce across types, so a number property never matches a string operand. The client working-set executor's natural order and `child-of` depth (now 10) were aligned with the host, which affects both paths.

## Not in this change

Phase 4 (the EDGE indexer, in `dxos/edge`) and phase 5 (property expression indexes, contentless FTS). The design doc is `echo-host/docs/SQL_QUERY_EXECUTION.md`.

**Known follow-up, unrelated to the executor.** Seeding cost grows with store size: a 200-object flush takes 5.5 s at 1,000 objects and 27.8 s at 9,000, hitting the 30 s RPC timeout around 15,000. The growing call is the client's `updateSubscription` on each batch. This bounded the benchmark at N = 5,000.

https://claude.ai/code/session_019tvpn6rzPaoGfFx9zgNtpB

---
_Generated by [Claude Code](https://claude.ai/code/session_019tvpn6rzPaoGfFx9zgNtpB)_

---
🚀 Composer preview: https://pr-13288-composer-dev.dxos.workers.dev


wittjosiah:
Two checks went red on `1e52e12` for the same reason, and it is not this diff.

- `Check / test (node-version=24.11.1, shard=4)`
- `Check / e2e (browser=firefox, shard=composer)`

Both died during toolchain setup, before any test body ran. The logs show repeated `curl: (22) The requested URL returned error: 504` fetching the moon CLI release tarball from GitHub, so the shard exited on the download rather than on anything it was asked to test:

```
curl: (22) The requested URL returned error: 504
failed to download https://github.com/moonrepo/moon/releases/download/v2.5.2/moon_cli-x86_64-unknown-linux-gnu.tar.xz
##[error]action failed with exit code 1
```

I have re-run both jobs. If other shards on this run fail the same way, it is this outage rather than a second cause; I will not repeat this comment for each one. Anything that fails after reaching a test body is a real failure and I will treat it as this PR's.

The same suites pass locally on this commit, in both executor modes: index-core 67, echo-host 398, echo-client 587, echo-client-e2e 365, and the e2e query suite forced to `DX_ECHO_QUERY_EXECUTOR=sql` at 185.

---

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -728,8 +770,69 @@ export class QueryExecutor extends Resource {
...
+   * indexed snapshot.
+   */
+  async #execCompiled(): Promise<QueryExecutionResult> {
+    const trace: ExecutionTrace = { ...ExecutionTrace.makeEmpty(), beginTs: performance.now() };
+    const plan = this._plan;
+    const { rows, sql, explain } = await this._runInRuntime(
+      Effect.gen(function* () {
+        const compiled = yield* compilePlan(plan);
+        const rows = yield* compiled.statement;
+        const explain = TRACE_QUERY_EXECUTION
```
dmaretskyi: isn't it a perf drain to explain every query

packages/core/echo/echo-host/src/query/sql/compile.ts
```diff
@@ -0,0 +1,1506 @@
...
+  /** Distinct `in-query` subqueries already compiled this plan, keyed as the executor caches them. */
+  readonly #subqueries = new Map<string, string>();
+  /** Resolved `metaVersion` literal sets, keyed by `key\0range`. */
+  readonly #metaVersions: Map<string, readonly string[]>;
+  /** Day boundaries for `timestamp` group keys in a named zone; SQLite has no zone tables of its own. */
+  readonly #dayStarts: DayStarts;
+  /** Whether any select in the plan scopes a space with its feeds, which lets a traversal reach feed items. */
+  #includeAllFeeds = false;
+
+  constructor(
```
dmaretskyi, line 142: Add SQL plan step kind, and make this emit the plan. sql plan step is just another step the executor can run

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: It never runs unless you ask for it: `TRACE_QUERY_EXECUTION` is `!!import.meta.env?.DX_TRACE_QUERY_EXECUTION`, read once at module scope, and the `EXPLAIN QUERY PLAN` round trip is inside that branch (along with the `console.log` that consumes it). With the flag unset a query issues exactly one statement.

Left as is, now in `_execSqlStep` after the restructure below.

---

packages/core/echo/echo-host/src/query/sql/compile.ts
dmaretskyi: Done in 9880fdcd.

`QueryPlan.SqlStep` is now a step kind:

```ts
export type SqlStep = {
  _tag: 'SqlStep';
  sql: string;
  params: readonly unknown[];
  /** The steps the statement stands for, kept so traces and scope analysis still see them. */
  steps: readonly Step[];
};
```

`SqlPlanCompiler.compile` emits `plan: QueryPlan.Plan.make([{ _tag: 'SqlStep', ... }])` alongside the statement, and `QueryExecutor` runs it through the ordinary loop — `_execStep` gained a `case 'SqlStep'` calling `_execSqlStep`, which is a source step (it ignores the incoming working set, since the statement stands for every step that would have produced one) and returns working-set items like any other.

Consequences worth noting:
- `#execCompiled` is now compile → `_execPlan` → the shared change detection; the separate `#compiledResults` cache is gone, and results flow through `getResults()` with everything else.
- `get plan()` reports the compiled plan once it exists, so a `sql` query's plan names the SQL it runs.
- The step-level trace carries `sql`/`explain`, nested under the root trace rather than replacing it.

The step is data-only (text + params), so nothing in `plan.ts` depends on the SQL layer, and a future partial compile can emit a `SqlStep` followed by in-memory steps without further changes to the executor.

Verified: index-core 67 passed, echo-host 398 passed, echo-client 588 passed, echo-client-e2e 365 passed, and the e2e query suite forced to `sql` 185 passed.

---

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -640,10 +677,13 @@ export class QueryExecutor extends Resource {
     this._id = options.queryId;
     this._query = options.query;
     this._reactivity = options.reactivity;
+    this.#mode = resolveQueryExecutorMode(options.executor);
 
     const queryPlanner = new QueryPlanner();
     this._plan = queryPlanner.createPlan(this._query);
     this.#scopes = extractScopes(this._plan);
+    // `objectSnapshot` drops `@meta` for document rows, so a plan reading it cannot be compiled.
+    this.#compilable = !planReadsObjectMeta(this._plan);
```
dmaretskyi: this belongs in query-planner. just make it do sql-compile too. config parameters in constructor

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -728,8 +776,46 @@ export class QueryExecutor extends Resource {
...
+    return this.mode === 'sql' ? this.#execCompiled() : this.#execInMemory();
+  }
+
+  /**
+   * The compiled path: the plan is compiled into a single `SqlStep` and run through the same step
+   * loop as any other plan. One statement over the index tables, no document loads — document rows
+   * ship identity only, since the client hydrates them from the document itself; feed rows carry
+   * the indexed snapshot.
+   */
+  async #execCompiled(): Promise<QueryExecutionResult> {
```
dmaretskyi: no, its just another step, like all other steps

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -42,6 +42,25 @@ import { QueryError } from './errors.ts';
...
+
+/**
+ * Which evaluation path answers queries: `memory` loads the objects a plan touches and evaluates
+ * every step in JS; `sql` compiles the plan into one SQLite statement over the index tables and
+ * loads nothing. `memory` is the default — the compiled path is opt-in per host through
+ * {@link QueryExecutorOptions.executor}, else by the `DX_ECHO_QUERY_EXECUTOR` environment variable.
+ */
+export type QueryExecutorMode = 'sql' | 'memory';
+
+export const resolveQueryExecutorMode = (explicit?: QueryExecutorMode): QueryExecutorMode => {
```
dmaretskyi: at this level this should be a constructor param, and hiegher up -- a @dxos/config option

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: do we need this method if we already have execSqlStep? just route everything through there, no? remove

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -656,13 +696,18 @@ export class QueryExecutor extends Resource {
   }
 
   get plan(): QueryPlan.Plan {
-    return this._plan;
+    return this.#compiledPlan ?? this._plan;
```
dmaretskyi: smelly -- just have one plan. compiled plan is the one that has the sql step. I wanna remove distinquition as much. aspossible in the executor

packages/core/echo/index-core/src/index-engine.ts
```diff
@@ -178,6 +200,15 @@ export class IndexEngine {
    * Query snapshots by recordIds.
    * Used to load queue objects from indexed snapshots.
    */
+  /**
+   * True once every `objectMeta` row has a snapshot. False while the store is still filling after
+   * its introduction, during which the compiled query path would read an incomplete database and
+   * must await indexing first.
+   */
+  hasCompleteSnapshots(): Effect.Effect<boolean, SqlError.SqlError, SqlClient.SqlClient> {
```
dmaretskyi, line 208: explain why we need this

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: Done in efc67f70. `QueryPlanner` takes the path as constructor config and owns the compile decision:

```ts
export type QueryPlannerOptions = {
  defaultTextSearchKind: QueryPlan.TextSearchKind;
  /** Evaluation path `QueryPlanner.compilePlan` targets. */
  executor?: QueryExecutorMode;
  noIndexes?: boolean;
};

compilePlan(query): Effect<QueryPlan.Plan, QueryError | SqlError, SqlClient>
```

`compilePlan` builds the steps, declines a plan that reads `@meta`, and otherwise returns a plan whose single `SqlStep` stands for them. `QueryExecutorMode` and `resolveQueryExecutorMode` moved here too, since the path is now planning-time config; everything outside imports them from `query/index.ts` as before, so no call site changed.

One thing I did not do, and want to flag rather than bury: `createPlan` stays pure and synchronous. Compiling reads the store — `metaVersion` literal sets come from `objectSnapshot` and the day boundaries of a named zone from `objectMeta` — so folding it into `createPlan` would have made it effectful, and the compiler itself calls `createPlan` recursively for subqueries (`compile.ts:604`, `compile.ts:1325`), as does `_resolveInQueryFilter`. Making it effectful would have rippled through those recursions and all ~20 planner tests for no gain. So compiling is a second method, and the executor resolves it once on first execution rather than in its constructor. Say the word if you want it folded in anyway.

---

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: Agreed — it is, now. efc67f70 deletes `#execCompiled` and `#execInMemory` entirely; there is one `execQuery` and no branch on the mode:

```ts
if (!this.#planned) {
  this._plan = await this._runInRuntime(this.#planner.compilePlan(this._query));
  this.#planned = true;
}
const { workingSet, trace } = await this._execPlan(this._plan, []);
```

A `SqlStep` is reached through `_execStep` like every other step. `mode` is no longer state — it reports whether the plan the planner produced contains a `SqlStep`. The `#compilable` flag and the separate compiled-result cache are gone, so both paths share one change-detection and one `getResults()`.

---

packages/core/echo/echo-host/src/query/sql/compile.ts
```diff
@@ -0,0 +1,1513 @@
...
+          break;
+        }
+        const conditions: Fragment[] = [fts.condition, scope];
+        if (step.selector.typename !== null) {
+          conditions.push(
+            step.selector.typename.length > 0 ? sql`(${buildTypeDxnCondition(sql, step.selector.typename)})` : sql`0`,
+          );
+        }
+        const rank = fts.ranked ? sql`-bm25(ftsIndex)` : sql`1.0`;
+        base = sql`SELECT m.recordId, m.objectId, m.spaceId, ${rank} AS rank FROM ftsIndex f JOIN objectMeta m ON m.recordId = f.rowid WHERE ${sql.and(conditions)}`;
```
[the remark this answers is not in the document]
dmaretskyi: Confirmed and fixed in efc67f70 — this one is real. `buildFtsCondition` builds its predicate on alias `f`, and `FtsIndex.query` aliases a different table per branch: `ftsIndex AS f` when ranked, `objectSnapshot AS f` for the `LIKE` fallback, with a comment giving the reason. The compiler aliased `ftsIndex` unconditionally, so a query whose shortest term is under the trigram minimum ran `f.snapshot LIKE …` against the debounce-flushed full-text table and missed writes the in-memory path matches.

Fixed as suggested, selecting the table from `fts.ranked`.

---

packages/core/echo/echo-host/src/query/query-planner.ts
```diff
@@ -27,8 +32,28 @@ const queryTooComplexError = (query: QueryAST.Query | null): QueryError => {
...
 
+/**
+ * Which evaluation path a plan targets: `memory` leaves every step for the executor to evaluate in
+ * JS over loaded objects; `sql` compiles the steps into a single {@link QueryPlan.SqlStep} over the
+ * index tables. `memory` is the default — the compiled path is opt-in per host through
+ * {@link QueryPlannerOptions.executor}, else by the `DX_ECHO_QUERY_EXECUTOR` environment variable.
+ */
+export type QueryExecutorMode = 'sql' | 'memory';
+
+export const resolveQueryExecutorMode = (explicit?: QueryExecutorMode): QueryExecutorMode => {
```
dmaretskyi: why is env thing here? this one just cares about constructor params

packages/core/echo/echo-host/src/query/query-planner.ts
```diff
@@ -67,6 +93,27 @@ export class QueryPlanner {
...
+  /**
+   * The plan as the executor will run it. Under `sql` the steps are compiled into one
+   * {@link QueryPlan.SqlStep}; otherwise, and for a plan the compiler declines, the steps are
+   * returned as {@link createPlan} built them.
+   *
+   * Separate from `createPlan` because compiling reads the store — `metaVersion` literal sets and
+   * the day boundaries of a named zone are resolved against it — while `createPlan` stays pure so
+   * the compiler can recurse into subqueries with it.
+   */
+  compilePlan(
```
dmaretskyi, line 83: we dont need new public APIs. the constructor param declares what result you get

packages/core/echo/echo-host/src/query/sql/compile.ts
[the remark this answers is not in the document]
dmaretskyi: Correcting my previous reply: the fix I described is not what landed, because main moved underneath it.

`0005_rebuild_fts_text` rebuilt the full-text table with a `text` column in place of `snapshot`, and `FtsIndex.query` now reads `ftsIndex AS f` on **both** branches — the index stores the extracted text, so a term below the trigram minimum scans that column rather than another store. Pointing the unranked branch at `objectSnapshot` was right against the old code and wrong against this one; it failed every full-text test in the e2e suite with `no such column: f.snapshot`.

The live bug the merge exposed is that `buildFtsCondition` — the helper the compiler shares with the index — still named `f.snapshot` after that rename. Fixed in f3ae4a9b: the helper builds on `f.text`, and the compiler reads `ftsIndex` on both branches, matching `FtsIndex.query`. The e2e query suite forced to `sql` is back to 185 passed.

---

packages/core/echo/echo-host/src/db-host/query-service.ts
```diff
@@ -161,7 +186,7 @@ export class QueryServiceImpl extends Resource implements QueryService.Handlers
       );
       scheduleMicroTask(ctx, async () => {
         await queryEntry.executor.open();
-        if (queryEntry.feedScoped) {
+        if (queryEntry.feedScoped || !(await this.#snapshotsComplete())) {
```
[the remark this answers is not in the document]
dmaretskyi: Confirmed and fixed in f3ae4a9b — this was real. `0008_normalized_ids` only does `ADD COLUMN` plus `CREATE INDEX`, `hasCompleteSnapshots()` counts missing snapshots and nothing else, and `reverseRef`'s cursor bump does not help here because `objectMeta` is written by the primary pass, whose cursors no migration resets. So an upgraded database kept `parentId`/`sourceId`/`targetId` NULL until each object next changed, and the compiled path silently omitted dependency, `child-of` and relation results for everything untouched.

Fixed with a backfill rather than a wider gate, since the gate would only delay the same missing data:

- `0009_backfill_normalized_ids` walks `objectMeta` in batches of 500 and fills the three columns for every row with a `parent`, `source` or `target` set.
- It derives them with `localEntityId`, the same function the indexing pass uses, so a backfilled row is indistinguishable from a rewritten one. That meant moving `localEntityId` into `entity-ids.ts` — importing it from the index would have made the migration module depend on the index that imports the migrations.
- As you noted, `NULL` is valid for a cross-space or non-entity reference, so rows are selected on the EID column being set rather than on the id column being NULL. The migration runs once, tracked in the migrations table, so that is idempotent.

Covered by a test in `entity-meta-index.test.ts` that inserts a pre-0008 row and a cross-space one: the local reference normalizes to its bare id, the cross-space one stays NULL. It fails without the migration and passes with it.

Doc claims in `SQL_QUERY_EXECUTION.md` corrected in the same commit, including your other thread's four items.

---

packages/core/echo/echo-host/src/db-host/query-service.ts
```diff
@@ -36,6 +36,16 @@ export type QueryServiceProps = {
...
+
+  /**
+   * True once every indexed object has a snapshot. The compiled executor reads that store rather
+   * than loading documents, so while it is still filling after upgrade a query awaits indexing
+   * before its first execution. Ignored on the in-memory path, which loads documents itself.
+   */
+  hasCompleteSnapshots?: () => Promise<boolean>;
+
+  /** Evaluation path for every query this service creates; see {@link QueryExecutorMode}. */
+  executor?: QueryExecutorMode;
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in f3ae4a9b, as suggested:

```ts
if (queryEntry.feedScoped || (queryEntry.executor.mode === 'sql' && !(await this.#snapshotsComplete()))) {
  await this._params.updateIndexes({ reason: 'feed-scoped-query' });
}
```

One thing this needed: after the planner refactor, `mode` was resolved from the compiled plan and so only became accurate after the first execution — too late for a gate that runs before it. `QueryPlanner.compiles(plan)` now answers the same question synchronously (configured mode is `sql` and the plan does not read `@meta`), and `mode` reads from that, so the gate gets the right answer up front.

---

packages/core/echo/echo-host/docs/SQL_QUERY_EXECUTION.md
```diff
@@ -0,0 +1,532 @@
...
+  bumped `reverseRef2` to `reverseRef3` for `propPathNormalized`, with tracker migration 0004
+  dropping the orphaned rows. `IndexEngine.update` reads the source once and writes `objectMeta`
+  once for all three dependent indexes when their cursor sets agree (the steady state); only while
+  one index lags, as during this backfill, does each diff against its own cursors in its own pass.
+- **The query gate** is `QueryServiceProps.hasCompleteBodies`; `QueryServiceImpl` awaits
+  `updateIndexes()` before a query's first execution while it is false and caches `true` once seen.
+- **The in-memory path is deleted.** It went once both benchmark runs were recorded (the `memory`
+  column in `BENCHMARKS.md` comes from `c5294281`). `QueryExecutor` now depends only on the
+  `SqlClient` runtime: it compiles the plan, runs the statement and diffs the rows for `changed`,
+  with no `IndexEngine`, `AutomergeHost` or `SpaceStateManager` dependency.
```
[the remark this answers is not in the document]
dmaretskyi: All four corrected in f3ae4a9b. Each was right:

- There is no `objectData`/`objectData1` — the compiler reads `objectSnapshot`, and `INDEX_NAMES` carries `objectSnapshot`, `reverseRef3`, `fts7`. The bullet also had the backfill mechanism wrong, which turned out to matter: `objectMeta` is not filled by a cursor reset at all, which is the bug in your other thread. Rewritten to say how each store is actually backfilled.
- The tracker migration is `0006_retire_reverse_ref2_cursors` as of this commit — it was `0005`, and renumbered in the merge because main's `0005_rebuild_fts_text` took that id.
- `hasCompleteBodies` → `hasCompleteSnapshots`.
- "The in-memory path is deleted" was the reverse of what ships: it is still the default, `QueryExecutor` still holds `IndexEngine`, `AutomergeHost` and `SpaceStateManager`, and `resolveQueryExecutorMode` defaults to `memory`. That paragraph described an earlier iteration of the branch.

---

packages/core/echo/echo-client/src/query/working-set-executor.ts
```diff
@@ -620,13 +621,16 @@ const _compareValues = (valueA: unknown, valueB: unknown): number => {
...
-    return valueA.localeCompare(valueB);
+    return valueA < valueB ? -1 : valueA > valueB ? 1 : 0;
   }
   if (typeof valueA === 'number' && typeof valueB === 'number') {
     return valueA - valueB;
   }
-  return String(valueA).localeCompare(String(valueB));
+  const stringA = String(valueA);
+  const stringB = String(valueB);
+  return stringA < stringB ? -1 : stringA > stringB ? 1 : 0;
```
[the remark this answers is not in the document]
dmaretskyi: Not taking this one, and it is worth saying why rather than silently skipping it.

The divergence is real and described correctly. But the suggested fix changes ordering on the **default** path: `localeCompare` is what every user gets today, and swapping it for code-unit comparison would reorder mixed-case and accented results for everyone, whether or not they ever enable the flag. That is a user-visible behaviour change well outside a PR that is meant to land inert.

The PR takes the opposite position deliberately — "Behaviour differences under `sql`, deliberate" documents code-unit ordering as a property of the compiled path. Between the two paths, the one that should move is the compiled one, by teaching the compiler a collation that matches `localeCompare`; that is a real piece of work (SQLite has no locale-aware collation without registering a custom one) and belongs with turning the flag on, not here.

Leaving the memory path as is and keeping the difference documented. Happy to be overruled if you would rather take the reorder now — @wittjosiah / @dmaretskyi, your call.

---

packages/core/echo/index-core/src/index-engine.ts
dmaretskyi: Explained in the code rather than only here:

> `objectSnapshot` is filled by the indexing pass, so a database that predates it holds rows without one, and the store fills over several passes after upgrade. The compiled query path reads that store directly instead of loading documents, so until it is complete a query there would silently return fewer objects than exist — not stale data, missing data. The query service gates a compiled query's first execution on this and caches `true` once seen, since it never goes back to false. The in-memory path loads documents itself and is not gated.

Short version: without it, the first queries after an upgrade quietly under-report while the store is still filling, and the failure is invisible — the query succeeds, it just misses rows.

While there I also put `querySnapshotsJSON`'s doc comment back; an earlier edit had left it stranded above this method.

---

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: Both done, in 3d4e4f85.

Constructor param: `QueryExecutor` takes the mode and hands it to `QueryPlanner` as constructor config; nothing below reads the environment. `resolveQueryExecutorMode` now lives in `EchoHost`, the one place the option arrives, so `DX_ECHO_QUERY_EXECUTOR` is resolved at the boundary and everything downstream just takes the value it is given.

The `@dxos/config` option is the part I have not done — this stays an `EchoHostProps` field plus the env var. Happy to wire it into config as a follow-up, but it reaches beyond this PR and would want a config schema change.

---

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: There is one plan now. `QueryExecutor` holds a single `_plan` that the planner produces in the constructor, and under `sql` it already contains the `SqlStep` — no second field, no `compiled ?? uncompiled`, no per-run recompile. `execQuery` runs `_execPlan(this._plan)` with no branch on the mode, and `mode` is just a report of which path the plan took.

The distinction that is left in the executor is one `case 'SqlStep'` in the step loop, which is the same kind of case every other step has.

---

packages/core/echo/echo-host/src/query/query-planner.ts
dmaretskyi: Done in 2fbf3217 — `createPlan` is the only public method again, and it is synchronous.

What made the second method necessary was that compiling read the store, and per your other note we now exclude those cases instead. `planDeclinedByCompiler` sends three shapes to the in-memory executor: a filter reading `@meta`, a `metaVersion` semver range (resolvable only against the versions present), and a `timestamp` day group in a named zone (local day boundaries need the store's timestamp range). With those gone, compilation is steps in, SQL out, and:

```ts
new QueryPlanner({ executor, sql }).createPlan(query): QueryPlan.Plan
```

The constructor param decides what comes back, as you said it should. `compilePlan` and `compiles` are deleted, the executor's per-run recompile is deleted, and the planner tests needed no changes because nothing became effectful.

Excluding the named-zone day group also fixed a real bug: its boundaries were resolved at compile time and cached with the plan, so once the store grew past the last boundary, objects updated on a later day were grouped into the wrong day and older replicated ones got a NULL key.

---

packages/core/echo/echo-client/src/query/working-set-executor.ts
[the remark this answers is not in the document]
dmaretskyi: Reversing my earlier decline — you were right, and I was wrong about the stakes.

I argued this was order-only and not worth changing the default path. It is not: under `limit` the collation decides which rows are *cut*, so the two executors returned different objects, not the same objects in a different order. A repro is in `echo-client-e2e/src/sql-ordering.test.ts`:

| query | `sql` | `memory` |
| --- | --- | --- |
| `orderBy(property('title'))` | `Banana, Date, apple, cherry, elderberry` | `apple, Banana, cherry, Date, elderberry` |
| `.limit(2)` | `Banana, Date` | `apple, Banana` |

Fixed in 2fbf3217 by normalizing everything to SQLite's collation, per @dmaretskyi: `_compareByProperty` and `GroupBy.compareScalar` now use a shared `compareCodeUnits`. The client working-set executor was already code-unit, so it needed no change.

The test pins the shared order as a literal rather than just asserting the two paths agree, so a comparator drifting back to `localeCompare` fails it instead of quietly moving both.

---

packages/core/echo/echo-host/src/query/sql/compile.ts
```diff
@@ -0,0 +1,1527 @@
...
+const aggregateColumn = (name: string): string => `agg_${sanitizeIdentifier(name)}`;
+
+/** Result field names come from the query author; only a safe subset can be a column name. */
+const sanitizeIdentifier = (name: string): string => {
+  const safe = name.replace(/[^A-Za-z0-9_]/g, '_');
+  if (safe !== name) {
+    throw new QueryError({ message: `Aggregate name is not a valid identifier: ${name}`, context: { name } });
+  }
+  return safe;
+};
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 2fbf3217, by encoding rather than by positional columns.

Every site already goes through `keyColumn`/`aggregateColumn`, and the identifier only has to be unique within the statement, so the change is contained to one helper: an already-safe name is kept as `s_<name>` for readable SQL, anything else becomes `u_<hex of its code units>`. The distinct prefixes keep the two classes from colliding, and the fixed-width encoding stays injective for astral characters.

`aggregate({ 'last-at': Aggregate.max('created') })` now compiles instead of throwing; covered by a test in `compile.test.ts`.

---

packages/core/echo/echo-host/src/query/query-executor.ts
```diff
@@ -725,53 +760,49 @@ export class QueryExecutor extends Resource {
   async execQuery(): Promise<QueryExecutionResult> {
     invariant(this._lifecycleState === LifecycleState.OPEN);
 
+    // Compiling reads the store, so the plan resolves here rather than in the constructor. Its
+    // steps do not change between runs, so this happens once.
+    if (!this.#planned) {
+      this._plan = await this._runInRuntime(this.#planner.compilePlan(this._query));
+      this.#planned = true;
+    }
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 2fbf3217, though not by recompiling per run.

The diagnosis was right: the day starts came from the store's timestamp range at compile time and were then cached with the plan, so an object updated past the last boundary fell into the previous day's group and one older than the first got a NULL key. Recompiling every run would have papered over it while keeping compilation impure.

Instead the compiler declines these plans and the in-memory executor answers them, which recomputes `GroupBy.truncateTimestamp` per run and never had the bug. `planDeclinedByCompiler` covers it along with `@meta` filters and `metaVersion` ranges — the three shapes that needed a store read before the statement existed. UTC day groups need no boundary table and still compile.

That also made compilation pure, so `QueryPlanner.createPlan` is synchronous again and the plan is built once in the constructor; there is no `#planned` flag left to go stale.

---

packages/core/echo/echo-host/src/query/sql/compile.ts
```diff
@@ -0,0 +1,1527 @@
...
+        return sql`COALESCE(${column}, 0) ${direction}`;
+      }
+      case 'property': {
+        if (grouped && grouped.aggregateNames.includes(order.property)) {
+          return sql`${sql.literal(aggregateColumn(order.property))} ${direction} NULLS LAST`;
+        }
+        const at = jsonPathLiteral(sql, [order.property]);
+        // Only scalars order; anything else sorts with the nulls, last.
+        return sql`CASE WHEN json_type(d.snapshot, ${at}) IN ('text', 'integer', 'real', 'true', 'false') THEN json_extract(d.snapshot, ${at}) END ${direction} NULLS LAST`;
+      }
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 2fbf3217, taking your option 2 rather than option 1 — @dmaretskyi's call was to normalize everything to SQLite rather than decline these plans, since `orderBy(property).limit(n)` is exactly the shape the compiled path exists to make fast.

Your analysis held up, including the part I had pushed back on: the `limit` case returns different rows, not just a different order. `_compareByProperty` and `GroupBy.compareScalar` (which drives `max`/`min`) now share a `compareCodeUnits` helper; the client working-set executor was already code-unit. `echo-client-e2e/src/sql-ordering.test.ts` pins the shared order.

You were also right that `differential.test.ts` only covers filtering — that is why this needed its own test rather than showing up there.

---

packages/core/echo/echo-host/docs/SQL_QUERY_EXECUTION.md
```diff
@@ -0,0 +1,534 @@
...
+| timestamp filter under `or`/`not`             | `QueryError` from the planner                                               | unchanged                                           | out of scope; the compiler could lift it later                                                                                                                       |
+| result universe                               | candidates from the index; bodies and traversal targets from live documents | everything from the index                           | reactive re-execution is triggered by the indexer after it commits, so the two universes cannot disagree there; a one-shot query reads the universe it selected from |
+| `documentJson` on document rows               | sent, discarded by the client                                               | omitted                                             | `_resolveIndexedObject` loads the document itself; feed rows keep the body                                                                                           |
```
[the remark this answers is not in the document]
dmaretskyi: Fixed in 2fbf3217, though the resolution went the other way from what you proposed.

Rather than label the table SQL-only, the ordering row is now true of both executors: the in-memory comparator moved to code-unit order, so the paths agree. That was the right fix because the divergence changed which rows a `limit` returned, not only their order.

The table now says so explicitly:

> "After" is the behaviour of **both** executors, not of the compiled path alone. The in-memory executor remains the default and answers every plan the compiler declines, so a row that held only for `sql` would be a divergence between the two.

The other two rows you named were already true of both paths: mixed-type comparison is the planner's typed rule, and `documentJson` omission is what the client's `_resolveIndexedObject` expects regardless of executor.

---

packages/core/echo/echo-host/src/query/query-executor.ts
dmaretskyi: The config half is done too now, in 4760f4d4 — I said it was a follow-up; @dmaretskyi asked for it in this PR, so it is here.

```proto
/// Which path evaluates ECHO queries on the host.
enum QueryExecutor {
  UNSPECIFIED_QUERY_EXECUTOR = 0;
  MEMORY = 1;
  SQL = 2;
}

optional QueryExecutor query_executor = 18 [ (env_var) = "DX_ECHO_QUERY_EXECUTOR" ];
```

It reaches the host by the same route `enableVectorIndexing` takes: `runtimePropsFromConfig` maps the enum to the mode, the service stack threads it through `echoHostLayer` to `EchoHostLayer`, and `EchoHost` takes it as the `queryExecutor` prop it already had.

One deliberate choice: an unset field maps to `undefined`, not `memory`. Defaulting in the config mapper would shadow the `DX_ECHO_QUERY_EXECUTOR` fallback in `EchoHost` and quietly break the env override, so the default stays in one place. Covered by a test in `client-services-stack.test.ts`, including that an explicit override beats config.

The generated `config_pb.ts` is gitignored and rebuilt by `protocols:gen-buf`, so the `.proto` is the whole change here.

---

